PR #31, opened 2015-01-27
Case insensitive should in example_wording.rb

### LINKED ISSUE #30: The Example Wording cop check for should is case sensitive.
It would be nice if the Example Wording cop also caught wordings like this:

```rspec
it 'Should do something' do 
```

and not only

```rspec
it 'should do something' do  
```

EsterYtterbrinkAtTrialbee:
example wording cop now also looks for should when spelled with an capital S. 

Closes #30.


geniou:
Sorry for the long delay. I changed the implementation (f4797cee4ffcb2106eba2077fb7d8b893c93e4b1) a little bit so that for example `SHOULD` would be found as well. Thanks for the support!

PR #39, opened 2015-06-18
Fix the build with a shallow clone of RuboCop

bquorning:
Running the specs depends on being able to load RuboCop's spec helpers. Since the RuboCop gem no longer (as of [v0.30.0](https://github.com/bbatsov/rubocop/commit/5e8199ed0b3f6cf50cb63a59311278eebc8683fa)) contains the spec files, we have to vendorize RuboCop.

We can make a git submodule, or we can make a git clone, and I have only heard bad things about using submodules in git. And making a shallow clone is relatively fast.

Potential issues:

 - The vendorized RuboCop and the RuboCop gem dependency may be out of sync, in a number of ways:
   - Travis will always checkout the master branch of RuboCop.
   - A locally vendorized RuboCop becomes outdated without warning.

### DIFF
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -2,8 +2,11 @@
 
 require 'rubocop'
 
-rubocop_gem_path = Gem::Specification.find_by_name('rubocop').gem_dir
-Dir["#{rubocop_gem_path}/spec/support/**/*.rb"].each { |f| require f }
+rubocop_path = File.join(File.dirname(__FILE__), '../vendor/rubocop')
+unless File.directory?(rubocop_path)
+  fail "Can't run specs without a local RuboCop checkout. Look in the README."
+end
+Dir["#{rubocop_path}/spec/support/**/*.rb"].each { |f| require f }
 
 RSpec.configure do |config|
   config.order = :random
--- a/.gitignore
+++ b/.gitignore
@@ -49,3 +49,5 @@ pkg
 
 # For rubinius:
 #*.rbc
+
+/vendor
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,6 +10,8 @@ matrix:
   allow_failures:
     - rvm: ruby-head
     - rvm: rbx-2
-before_install: gem update --remote bundler
+before_install:
+ - gem update --remote bundler
+ - git clone --depth 1 git://github.com/bbatsov/rubocop.git vendor/rubocop
 script:
   - bundle exec rake
--- a/README.md
+++ b/README.md
@@ -72,6 +72,22 @@ RSpec/FilePath:
 ```
 
 
+## Contributing
+
+1. Fork it
+2. Create your feature branch (`git checkout -b my-new-feature`)
+3. Commit your changes (`git commit -am 'Add some feature'`)
+4. Push to the branch (`git push origin my-new-feature`)
+5. Create new Pull Request
+
+For running the spec files, this project depends on RuboCop's spec helpers.
+This means that in order to run the specs locally, you need a (shallow) clone
+of the RuboCop repository:
+
+```bash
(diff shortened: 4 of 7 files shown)


spec/spec_helper.rb
```diff
@@ -2,8 +2,11 @@
 
 require 'rubocop'
 
-rubocop_gem_path = Gem::Specification.find_by_name('rubocop').gem_dir
-Dir["#{rubocop_gem_path}/spec/support/**/*.rb"].each { |f| require f }
+rubocop_path = File.join(File.dirname(__FILE__), '../vendor/rubocop')
+unless File.directory?(rubocop_path)
+  fail "Can't run specs without a local RuboCop checkout. Look in the README."
```
andyw8: There doesn't seem to be any README in this PR?

bquorning: True. I was hoping to get feedback on the code before writing README (and PR description).

bquorning:
Added README into and PR description.

bquorning:
Ping @andyw8

geniou:
Thanks for the pull request! As soon as I have the time I'm having a look. Sorry that I have to let you wait, but its not forgotten..

bquorning:
No problem :-)

spec/spec_helper.rb
bquorning: The README has been changed to include information about RuboCop’s spec helpers.

backus:
ping @geniou. Might make sense to make @bquorning a maintainer or point to his fork or something

bquorning:
Rebased on master.

@geniou @andyw8 Could one of you please find time to look at this PR? I would like to have some feedback on whether this approach makes sense, of if I should try looking for other solutions to the build problem.

andyw8:
My gut feeling is that having to clone RuboCop is not so good. But I'm not really familiar enough with RuboCop to suggest a better approach.

bquorning:
The problem is that in order to run the specs for `rubocop-rspec`, the spec helpers from `rubocop` are needed. And they are not shipped with the gem, only with the source.

andyw8:
Perhaps @bbatsov can suggest a better approach?

bquorning:
We could add the spec helpers back into the gem?

geniou:
Sorry for my silence.. I'm really sorry for that and thanks for the pull request!

I get the problem but I also think the need of having to clone of rubocop is not perfect.. I see some options:

* include rubocop clone (current proposal)
* provide the rubocop test cases somehow - maybe by extention in a separate gem we and rubocop is using.. Therefore we would need an opinion of @bbatsov
* find a solution to have our own test classes and don't need to rely on external once..

I personally would like the second one.. bur maybe for now we should go for the first one and then try to replace it by a better (long term) solution.. opinions?

geniou:
Just for reference the commit on rubocop that caused that problem: https://github.com/bbatsov/rubocop/commit/5e8199ed0b3f6cf50cb63a59311278eebc8683fa

We could also - as work around - use a version of rubocop that still delivers the required files..

bquorning:
Depending on an old version of RuboCop is not a good idea, I think. Wouldn't that make this gem useless for people using a newer version of RuboCop?

Another (temporary) solution would be to simply copy the RuboCop spec helpers into this project.

geniou:
I don't know if there is a solution where you have development dependencies different from the one on production..

I think it's more than only one file so coping is not a good solution as well.

geniou:
I'm going to merge this even in the knowledge that its not a perfect solution to have this `vendor` folder. But for now it looks like the (good) way to get the test green for now. I'm also creating a new issue to find a long term solution.

@bquorning thanks for the PR and all you afford!

bquorning:
:heart: Thanks.

backus:
:smiley:

PR #49, opened 2015-09-16
Add RSpec/AnyInstance

mlarraz:
Because `allow_any_instance_of` and `expect_any_instance_of` are bad

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 * Update to rubocop 0.37.1. ([@nijikon][])
 * Update ruby versions we test against. ([@nijikon][])
 * Add `RSpec::NotToNot` cop. ([@miguelfteixeira][])
+* Add `Rspec/AnyInstance` cop. ([@mlarraz][])
 
 ## 1.3.1
 
@@ -73,3 +74,4 @@
 [@nijikon]: https://github.com/nijikon
 [@pstengel]: https://github.com/pstengel
 [@miguelfteixeira]: https://github.com/miguelfteixeira
+[@mlarraz]: https://github.com/mlarraz
--- a/config/default.yml
+++ b/config/default.yml
@@ -1,3 +1,7 @@
+RSpec/AnyInstance:
+  Description: 'Check that instances are not being stubbed globally'
+  Enabled: true
+
 RSpec/DescribeClass:
   Description: 'Check that the first argument to the top level describe is the tested class or module.'
   Enabled: true
--- a/lib/rubocop-rspec.rb
+++ b/lib/rubocop-rspec.rb
@@ -9,6 +9,7 @@ require 'rubocop/rspec/top_level_describe'
 RuboCop::RSpec::Inject.defaults!
 
 # cops
+require 'rubocop/cop/rspec/any_instance'
 require 'rubocop/cop/rspec/describe_class'
 require 'rubocop/cop/rspec/describe_method'
 require 'rubocop/cop/rspec/described_class'
--- a/lib/rubocop/cop/rspec/any_instance.rb
+++ b/lib/rubocop/cop/rspec/any_instance.rb
@@ -1,0 +1,45 @@
+# encoding: utf-8
+
+module RuboCop
+  module Cop
+    module RSpec
+      # Pefer instance doubles over stubbing any instance of a class
+      #
+      # @example
+      #   # bad
+      #   describe MyClass do
+      #     before { allow_any_instance_of(MyClass).to receive(:foo) }
+      #   end
+      #
+      #   # good
+      #   describe MyClass do
+      #     let(:my_instance) { instance_double(MyClass) }
+      #
+      #     before do
+      #       allow(MyClass).to receive(:new).and_return(my_instance)
+      #       allow(my_instance).to receive(:foo)
+      #     end
+      #   end
+      class AnyInstance < Cop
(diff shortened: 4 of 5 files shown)


andyw8:
This would be a nice addition! :+1:

spec/project_spec.rb
```diff
@@ -1,6 +1,6 @@
 # encoding: utf-8
 
-describe 'RuboCop Project' do # rubocop:disable RSpec/DescribeClass
+describe 'RuboCop Project' do
```
andyw8: This change seems be causing Travis to fail.

mlarraz: Yeah interestingly I actually need that to make it pass locally, due to
different versions of Rubocop I guess.

mlarraz:
I believe this is ready to go now

pd:
:+1: I would love to see this make it into master, I am just itching to properly ban `any_instance` from the projects I work on without having to scan for it in every pull request.

packrat386:
:+1: This would be great to have. I've seen too many bad tests with it.

nijikon:
@mlarraz Please rebase on top of master

nijikon:
@mlarraz Can you please rebase again, add changelog entry and squash to one commit? Sorry for the inconvenience.

nijikon:
Still missing the changelog entry and commits need to be squashed to one.

nijikon:
Nearly there, please fix tests :)

mlarraz:
Thanks, don't think I've ever seen a project with tests for their changelog before

nijikon:
@mlarraz This is taken from rubocop :) https://github.com/bbatsov/rubocop/blob/master/spec/project_spec.rb

maciejmajewski:
Could you explain why it is a bad practice?

andyw8:
https://relishapp.com/rspec/rspec-mocks/docs/working-with-legacy-code/any-instance

> This feature is sometimes useful when working with legacy code, though in general we
> discourage its use for a number of reasons:
> * The rspec-mocks API is designed for individual object instances, but this feature operates on entire classes of objects. As a result there are some semantically confusing edge cases. For example, in expect_any_instance_of(Widget).to receive(:name).twice it isn't clear whether a specific instance is expected to receive name twice, or if two receives total are expected. (It's the former.)
> *Using this feature is often a design smell. It may be that your test is trying to do too much or that the object under test is too complex.
> *It is the most complicated feature of rspec-mocks, and has historically received the most bug reports. (None of the core team actively use it, which doesn't help.)

maciejmajewski:
@andyw8 thanks

nijikon:
:+1:

dpisarewski:
Could you explain why stubbing of the `new` method is better? In my opinion it's much worse than stubbing a needed method of the class.

bquorning:
> In my opinion it's much worse than stubbing a needed method of the class.

What `any_instance` does is not stubbing a needed method of the class. It’s stubbing a needed method of *an instance of* the class. I like that by stubbing `new` you now at least know which object instance you are working with.

But, as mentioned above:

> Using this feature is often a design smell. It may be that your test is trying to do too much or that the object under test is too complex.

Stubbing `new` obviously doesn’t fix this problem.

backus:
Here is an example of why people usually say that `any_instance` should be the very last tool you reach for to test your code:

```ruby
require 'rspec'

class Foo
  def bar
    :original_implementation
  end
end

$global_instance = Foo.new

RSpec.describe Foo do
  context 'any instance of' do
    before do
      allow_any_instance_of(Foo)
        .to receive(:bar)
        .and_return(:stubbed_implementation)
    end

    it 'stubs method for instance created in life cycle' do
      expect(Foo.new.bar).to be(:stubbed_implementation)
    end

    it 'hijacks instances created without Foo.new' do
      expect(Class.instance_method(:new).bind(Foo).call.bar)
        .to be(:stubbed_implementation)
    end

    it 'stubs preexisting instance created outside of test' do
      expect($global_instance.bar).to be(:stubbed_implementation)
    end
  end

  context 'stubbing .new' do
    before do
      stubbed_instance = Foo.new

      allow(stubbed_instance)
        .to receive(:bar)
        .and_return(:stubbed_implementation)

      allow(Foo)
        .to receive(:new)
        .and_return(stubbed_instance)
    end

    it 'stubs method for instance created in life cycle' do
      expect(Foo.new.bar).to be(:stubbed_implementation)
    end

    it 'does not hijack objects created without Foo.new' do
      expect(Class.instance_method(:new).bind(Foo).call.bar)
        .to be(:original_implementation)
    end

    it 'does not stub preexisting instance created outside of test' do
      expect($global_instance.bar).to be(:original_implementation)
    end
  end
end
```

```
$ rspec --format documentation any_instance_madness.rb

Foo
  any instance of
    stubs method for instance created in life cycle
    hijacks instances created without Foo.new
    stubs preexisting instance created outside of test
  stubbing .new
    stubs method for instance created in life cycle
    does not hijack objects created without Foo.new
    does not stub preexisting instance created outside of test

Finished in 0.00695 seconds (files took 0.07858 seconds to load)
6 examples, 0 failures
```

Summarizing what this code example shows:

1. Stubbing `.new` ensures that the object you are stubbing is created during the life cycle of your test.
2. Stubbing `.new` specifies that your object is created through `.new`.

dpisarewski:
Thank you for the explanation.

PR #54, opened 2015-11-20
'RSpec' class should translate to 'rspec' file path instead of 'r_spec'

rickypai:
'RSpec' is typically translated into 'rspec' instead of 'r_spec' in the underscore notation. `RSpec::FilePath` cop should also conform to this.

### DIFF
--- a/.rspec
+++ b/.rspec
@@ -1,1 +1,2 @@
 --require spec_helper
+--color
--- a/lib/rubocop/cop/rspec/file_path.rb
+++ b/lib/rubocop/cop/rspec/file_path.rb
@@ -19,6 +19,9 @@ module RuboCop
 
         MESSAGE = 'Spec path should end with `%s`'
         METHOD_STRING_MATCHER = /^[\#\.].+/
+        EXCEPTIONS = {
+          'RSpec' => 'rspec'
+        }
 
         def on_top_level_describe(node, args)
           return unless single_top_level_describe?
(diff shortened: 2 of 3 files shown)


andyw8:
It might be useful to provide a mapping as a configuration option, but I don't think this should be part of the gem itself. Wouldn't that imply that any other special case that someone might need should be added to the gem too?

rickypai:
That's a good suggestion. Do you think RSpec -> rspec should be part of the default? Else the gem itself would be violating its own rule if we also assert for directory paths with RSpec (https://github.com/nevir/rubocop-rspec/tree/master/spec/rubocop/cop/rspec)

nijikon:
@rickypai @andyw8 what is your conclusion on this?

andyw8:
Just got around to looking this again, and it seems to me that the desired behaviour is already available by using the `CustomTransform` config, e.g.:

```ruby
  context 'with a CustomTransform configuration' do
    let(:cop_config) { { 'CustomTransform' => { 'RSpecHelper' => 'rspec_helper' } } }

    it 'handles exception for class names including RSpec' do
      inspect_source(cop,
                    'describe RSpecHelper do; end',
                    'rspec_helper_spec.rb')
      expect(cop.offenses).to be_empty
    end
  end
```

rickypai:
yeah, i ended up finding the configuration. thanks!

PR #60, opened 2016-01-14
Upgrade rubocop to 0.36

jawshooah:
Also fix offenses introduced by the upgrade.

[Review on Reviewable](https://reviewable.io/reviews/nevir/rubocop-rspec/60)

PR #63, opened 2016-01-20
Fix Travis test/builds

shaneog:
#### Fix Travis test
Test against Ruby 2.2 and 2.3
Also speed up tests by caching the bundler directory, and prevent test
suites which are allowed to fail from delaying the reporting of success.

This relies on #60 and #61, which introduce fixes for Ruby >2.1 and
having the correctly versioned Rubocop source available.

#### Switch from Coveralls to CodeClimate
CodeClimate is already used for code analysis so we can use it for test
coverage quite easily.

## NOTE: 
@nevir You need to replace the `XXXX` in `.travis.yml` with the actual CodeClimate repo token, which you can get by following the instructions [here](https://docs.codeclimate.com/docs/setting-up-test-coverage).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,8 +1,11 @@
+cache: bundler
 language: ruby
 rvm:
   - 1.9.3
   - 2.0.0
   - 2.1
+  - 2.2
+  - 2.3
   - ruby-head
   - jruby-19mode
   - rbx-2
@@ -10,6 +13,7 @@ matrix:
   allow_failures:
     - rvm: ruby-head
     - rvm: rbx-2
+  fast_finish: true
 before_install:
  - gem update --remote bundler
  - git clone --depth 1 git://github.com/bbatsov/rubocop.git vendor/rubocop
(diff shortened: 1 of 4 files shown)


.travis.yml
```diff
@@ -1,8 +1,11 @@
+cache: bundler
 language: ruby
 rvm:
   - 1.9.3
   - 2.0.0
   - 2.1
+  - 2.2
+  - 2.3
```
jawshooah: I believe this needs to be 2.3.0

shaneog: Thanks, changed

spec/spec_helper.rb
```diff
@@ -2,6 +2,11 @@
 
 require 'rubocop'
 
+if ENV['TRAVIS']
```
andyw8: I would suggest using `ENV['CI']` here rather than making it Travis-specific.

nijikon:
This was fixed in master, thanks!

PR #73, opened 2016-03-04
Add cop to check for focused specs

renanborgescampos:
Detect offenses whenever focus: true arg is found


danielnc:
:+1:

andyw8:
@renanborgescampos did you mean to delete this? Seems like it would be useful.

lib/rubocop/cop/rspec/focus.rb
```diff
@@ -0,0 +1,41 @@
...
+      #
+      #   # good
+      #   describe MyClass do
+      #   end
+      class Focus < Cop
+
+        MESSAGE = "Focused spec found.".freeze
+
+        FOCUSABLE_BLOCKS = [
+          :describe, :context, :it, :feature, :scenario
```
andyw8: What about `specify`? Also, the pending variants such as `xit`, `xdescribe` etc.

renanborgescampos:
@andyw8 Soon I'll open another pull request with this cop, and some code improvements.

PR #83, opened 2016-04-25
Add cop to check for predicate matcher

haeky:
Favor `expect(foo).to be_valid` over `expect(foo.valid?).to be_truthy`


lib/rubocop/cop/rspec/predicate.rb
```diff
@@ -0,0 +1,77 @@
+# encoding: utf-8
+
+module RuboCop
+  module Cop
+    module RSpec
+      # Prefer predicate matcher over
+      #
+      # @example
+      #   # bad
+      #     expect(food.valid).to be_truthy
```
andyw8: Should this say `foo.valid?` ?

lib/rubocop/cop/rspec/predicate.rb
```diff
@@ -0,0 +1,77 @@
+# encoding: utf-8
+
+module RuboCop
+  module Cop
+    module RSpec
+      # Prefer predicate matcher over
```
andyw8: Incomplete description?

spec/rubocop/cop/rspec/predicate_spec.rb
```diff
@@ -0,0 +1,20 @@
...
+
+    expect(cop.offenses.size).to eq(1)
+    expect(cop.highlights).to eq(['expect(foo.valid?).to be_truthy'])
+
+    message = 'Use predicate matcher `expect(foo).to be_valid`'
+    expect(cop.messages).to eq([message])
+  end
+
+  it 'autocorrects an offense' do
+    replacement = autocorrect_source(cop, ['expect(foo.valid?).to be_truthy'])
```
andyw8: Minor issue, but is there a chance the autocorrected code spec could stricter than the original, since `be_truthy` is different than `be_true`?

haeky: I'm not sure to understand. The cop should change `be_truthy` or `be_true` to a predicate matcher altogether as long as you call a boolean method in the expectation?

andyw8: You're right, I misunderstood how the `be_` matcher works.

haeky: This is where the magic happens 

https://github.com/rspec/rspec-expectations/blob/master/lib/rspec/matchers/built_in/be.rb#L239-L241

lib/rubocop/cop/rspec/predicate.rb
```diff
@@ -0,0 +1,77 @@
+# encoding: utf-8
+
+module RuboCop
+  module Cop
+    module RSpec
+      # Prefer using predicate matcher instead of truthiness
+      #
+      # @example
+      #   # bad
+      #     expect(foo.valid).to be_truthy
```
andyw8: Should be `foo.valid?`?

nijikon:
@haeky what is the status of this?

haeky:
I think this is good to go @nijikon 👍

nijikon:
@haeky Nice! Can you please add a changelog entry?

haeky:
Done @nijikon!

haeky:
Done @nijikon!

haeky:
@andyw8 What do you think of this?

andyw8:
Looks good to me. It would be nice to also handle the inverse (falsey) but it's not critical.

backus:
Yeah can we make this configurable @haeky? I often write

```ruby
expect(foo.success?).to be(false)
```

because being falsey can mean returning `nil` and I want to be explicit about exactly what I want type an important public method like `success?` returns.

Could you also run your cop through [mutant](https://github.com/mbj/mutant) and kill any mutations you encounter? If you add `mutant-rspec` to your Gemfile temporarily then you can run `bundle exec mutant --use rspec -- 'RuboCop::Cop::RSpec::Predicate*'`. I'm happy to answer any questions you have after using mutant

haeky:
What do you mean by configurable? I thought you could deactivate any cop by modifying `rubocop.yml`

backus:
@haeky some cops have allow you to specify a style. For example:

* [`BarePercentLiterals` Config](https://github.com/bbatsov/rubocop/blob/master/config/default.yml#L194-L199)
* [`BarePercentLiterals` implementation which takes configuration into account](https://github.com/bbatsov/rubocop/blob/master/lib/rubocop/cop/style/bare_percent_literals.rb#L9)

So we are saying that it would be nice if we could write something like

```yaml
RSpec/Predicate:
  Description: 'Prefer using predicate matcher instead of truthiness'
  EnforcedStyle: inflected
  SupportedStyles:
      - inflected # enforces expect(foo).to be_bar
      - explicit  # enforces expect(foo.bar?).to be(true)
  Enabled: true
```

(idk about the naming for the `SupportedStyles` thing but you get the idea)

backus:
Hey @haeky are you interested in making these changes? If not that is fine. I'd be happy to pickup where you left off at some point. Just want to know whether I should keep pinging you

haeky:
@backus I will take care of it!

shepmaster:
Running this PR against my code has this error on one file:

```
An error occurred while RSpec/Predicate cop was inspecting .../app/models/client.rb.
undefined method `source' for nil:NilClass
.../rubocop-rspec/lib/rubocop/cop/rspec/predicate.rb:50:in `correction'
.../rubocop-rspec/lib/rubocop/cop/rspec/predicate.rb:18:in `on_send'
```

shepmaster:
Additionally, this cop changed a migration of mine - perhaps he cop is not properly scoped to the spec like the rest of them?

db/migrate/20141029092439_remove_logs_with_missing_loggables.rb

```diff
-    Log.all.includes(:loggable).to_a.select { |log| log.loggable.nil? }.each(&:destroy)
+    Log.all.includes(:loggable).to_a.select(&:loggable).each(be_nil)
```

haeky:
I will take a look at it @shepmaster

shepmaster:
It also improperly negated some checks:

```diff
-      expect(company.active?).to eq false
+      expect(company).to be_active
```

shepmaster:
It should also respect the special case for `has_` predicates:

```diff
-      expect(project.has_dependent_tasks?).to eq true
+      expect(project).to be_has_dependent_tasks
```

It should be 

```diff
+      expect(project).to have_dependent_tasks
```

shepmaster:
And to be clear, I'm very excited for this cop to be added, this code base has a slew of issues it would help with ;-)

haeky:
Bare with me, I didn't think of all those use-cases, I'll refactor a little bit.

shepmaster:
> Bare with me,

Absolutely! I just found this project, so I'm more than happy to help provide some more test cases 😈

shepmaster:
> can I have an example of where the code crashed?

This seems to reproduce it:

```ruby
class Client < ActiveRecord::Base
  def self.fetch_import(columns)
    columns.delete_if { |_, v| v.blank? }
      .each { |k, v| columns[k] = v&.strip }
  end
end
```

```
0.42.0 (using Parser 2.3.1.2, running on ruby 2.3.1 x86_64-darwin15)
```

And I'm at ccc61bb54d11da08cbd3b61e5dfb64e5a1dad185 from this branch.

haeky:
I made a quick PoC, I'll clean up the code when I get back home.

shepmaster:
Appears that the error now occurs

```
undefined method `source' for nil:NilClass
rubocop-rspec/lib/rubocop/cop/rspec/predicate.rb:66:in `predicate_correction'
rubocop-rspec/lib/rubocop/cop/rspec/predicate.rb:48:in `correction'
rubocop-rspec/lib/rubocop/cop/rspec/predicate.rb:20:in `on_send'
```

shepmaster:
It got a bit overzealous:

With nested calls:

```diff
-        expect(controller.send(:connected_and_access_allowed?)).to eq false
+        expect(controller).not_to be_send(:connected_and_access_allowed)
```

With question marks in strings (also note that the `1` disappears ⚠️):

```diff
-    expect(Delayed::Job.where(['handler = ? and failed_at is null', ExtendProjectJob.handler(project_id)]).count).to eq 1
+    expect(Delayed::Job).to be_where(['handler =  and failed_at is null', ExtendProjectJob.handler(project_id)])
```

`respond_to` should be special-cased:

```diff
-      expect(Client.respond_to?(:save_import)).to eq true
+      expect(Client).to be_respond_to(:save_import)
```

It lost some code and generated invalid code ⚠️ 

```diff
-        it { expect(IntegrationPolicy
-                      .user_is_allowed?(user: owner,
-                                        integration: integration)).to be_truthy }
+        it { expect(IntegrationPolicy).to be_user_is_allowed(user: owner, }
```

haeky:
Thanks for all those specs 👍

backus:
@haeky @shepmaster thanks for helping this PR move forward :)

shepmaster:
Haha, it's all @haeky; I'm just being an effective troublemaker! 😇

haeky:
@shepmaster I added tons of spec and made them all works. I'll refactor the code to make it cleaner but let me know what you think about it.

shepmaster:
This is probably outside of the scope of this Cop, but maybe there should be special Capybara knowledge somewhere. Right now, this is the change:

```diff
-    expect(page.has_no_text?("Quick Start")).to eq true
+    expect(page).to have_no_text("Quick Start")
```

But It could be:

```diff
+    expect(page).to_not have_text("Quick Start")
```

shepmaster:
Nice! You handle the `is_a?` specially as well! 👍 

```diff
-      expect(client.data.is_a?(Hash)).to eq true
+      expect(client.data).to be_a(Hash)
```

haeky:
@shepmaster Let's make this merge first and we can increment after 😄

shepmaster:
I think this works, but I'd normally write it as `have_key`...

```diff
-      expect(project.staff_summary.key?("manager_color")).to eq true
+      expect(project.staff_summary).to be_key("manager_color")
```

haeky:
Is `key?` the norm or `has_key?`?

shepmaster:
> Let's make this merge first and we can increment after

Oh, certainly. I don't even know if such a thing should exist (the original example there was a misuse of Capybara, afterall).

shepmaster:
Still getting this error:

```
undefined method `source' for nil:NilClass
rubocop-rspec/lib/rubocop/cop/rspec/predicate.rb:66:in `predicate_correction'
rubocop-rspec/lib/rubocop/cop/rspec/predicate.rb:48:in `correction'
rubocop-rspec/lib/rubocop/cop/rspec/predicate.rb:20:in `on_send'
```

And a migration (outside of the spec directory!) is still being changed.

shepmaster:
> Is key? the norm or has_key??

Rubocop [changes it to `key?`](Use Hash#key? instead of Hash#has_key? and Hash#value? instead of Hash#has_value?), but it reads strange for the spec. It's a special case that you could choose to ignore.

haeky:
No worry, I'll make it handle both.

haeky:
@shepmaster @backus, I've refactor most of my code to handle multiple style and multiple use-cases. Let me know what you think of it.

backus:
@haeky one failing test remaining that just asserts that the alphabetical order of cop requires is maintained. I'll do a code review either tonight or this weekend.

haeky:
Thanks, I'll fix the rubocop errors in the meanwhile.

spec/rubocop/cop/rspec/predicate_spec.rb
```diff
@@ -0,0 +1,116 @@
+# encoding: utf-8
+
+describe RuboCop::Cop::RSpec::Predicate, :config do
+  subject(:cop) { described_class.new(config) }
+
+  context 'when EnforcedStyle is boolean' do
+    let(:cop_config) { { 'EnforcedStyle' => 'equality' } }
+
+    TESTS = {
```
backus: Using a constant here ends up emitting warnings. If you use a local variable you'll get the same reuse perk without the warnings

spec/rubocop/cop/rspec/predicate_spec.rb
```diff
@@ -0,0 +1,116 @@
...
+      'expect([]).to respond_to(:foo)' =>
+        'expect([].respond_to?(:foo)).to eq(true)',
+      'expect([]).not_to be_a(Array)' =>
+        'expect([].is_a?(Array)).to eq(false)',
+      'expect([:foo]).to include(:foo)' =>
+        'expect([:foo].include?(:foo)).to eq(true)'
+    }.freeze
+
+    TESTS.each do |offense, suggestion|
+      it "flags `#{offense}`" do
```
backus: This came after you originally opened this PR, but I highly recommend using this matcher that I added which makes tests like these way more clear and readable. This is what these tests would look like:

```ruby

it 'flags violations' do
  expect_violation(<<-RUBY)
    expect(foo.bar).to be_valid
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^ Use predicate matcher `expect(foo.bar.valid?).to eq(true)`
    expect({ foo: :bar }).to have_key(:foo)
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ Use predicate matcher `expect({ foo: :bar }.key?(:foo)).to eq(true)`
    expect({ foo: :bar }).to have_attributes(:foo, :baz)
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ Use predicate matcher `expect({ foo: :bar }.has_attributes?(:foo, :baz)).to eq(true)`
    expect([]).to respond_to(:foo)
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ Use predicate matcher `expect([].respond_to?(:foo)).to eq(true)`
    expect([]).not_to be_a(Array)
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ Use predicate matcher `expect([].is_a?(Array)).to eq(false)`
    expect([:foo]).to include(:foo)
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ Use predicate matcher `expect([:foo].include?(:foo)).to eq(true)`
  RUBY
end
```

lib/rubocop/cop/rspec/predicate.rb
```diff
@@ -0,0 +1,182 @@
...
+            matcher.gsub('have', 'has')
+          else
+            matcher.gsub('be_', '')
+          end
+        end
+        # rubocop:enable Metrics/MethodLength
+
+        def matcher
+          if action == :to
+            requires_truthiness? ? 'be_truthy' : 'eq(true)'
```
backus: Booleans can be compared by identity so this should be `be(true)` instead

lib/rubocop/cop/rspec/predicate.rb
```diff
@@ -0,0 +1,182 @@
...
+          else
+            requires_truthiness? ? 'be_falsey' : 'eq(false)'
```
backus: Instance of https://github.com/nevir/rubocop-rspec/pull/83/files#r74694683

CHANGELOG.md
```diff
@@ -20,6 +20,7 @@
 * Add `BeEql` cop which looks for expectations that can use `be(...)` instead of `eql(...)`. ([@backus][])
 * Add autocorrect support for `BeEql` cop. ([@backus][])
 * Add `MessageExpectation` cop for enforcing consistent style of either `expect(...).to receive` or `allow(...).to receive`. ([@backus][])
+* Add `RSpec::Predicate` cop. ([@haeky][])
```
backus: Can you make this just `Predicate`? All of the other cops do not include the `RSpec` prefix in the changelog

lib/rubocop/cop/rspec/predicate.rb
```diff
@@ -0,0 +1,182 @@
...
+
+        private
+
+        def requires_predicate?
+          style == :predicate
+        end
+
+        def correction(node)
+          if requires_predicate?
+            PredicateCorrector.new(node, style).correction
```
backus: The predicate corrector can just be initialized with `node` since it doesn't use `style` internally

lib/rubocop/cop/rspec/predicate.rb
```diff
@@ -0,0 +1,182 @@
...
+          else
+            BooleanCorrector.new(node, style).correction
+          end
+        end
+      end
+
+      # Corrector base class for offense detection
+      class Corrector
+        attr_reader :node, :style, :object, :args, :action
```
backus: Can you make these private?

lib/rubocop/cop/rspec/predicate.rb
```diff
@@ -0,0 +1,182 @@
...
+          if requires_predicate?
+            PredicateCorrector.new(node, style).correction
+          else
+            BooleanCorrector.new(node, style).correction
+          end
+        end
+      end
+
+      # Corrector base class for offense detection
+      class Corrector
```
backus: Can you make the Corrector (and its subclasses) private constants? What you can do is write `private_constant(*constants(false))` beneath `BooleanCorrector` which will make all of the correctors private

lib/rubocop/cop/rspec/predicate.rb
```diff
@@ -0,0 +1,182 @@
...
+        def autocorrect(node)
+          lambda do |corrector|
+            corrector.replace(node.loc.expression, correction(node))
+          end
+        end
+
+        private
+
+        def requires_predicate?
+          style == :predicate
```
backus: Can you write this as `style.equal?(:predicate)`. This is more strict (it signals that `style` will ONLY be that symbol)

config/default.yml
```diff
@@ -125,3 +125,12 @@ RSpec/NamedSubject:
 RSpec/SubjectStub:
   Description: 'Checks for stubbed test subjects'
   Enabled: true
+
+RSpec/Predicate:
+  Description: 'Prefer using predicate matcher instead of truthiness'
```
backus: Can you update this to simply convey that it enforces one of several styles. Checkout the descriptions of the other `EnforcedStyle` cops for examples of what I mean

lib/rubocop/cop/rspec/predicate.rb
```diff
@@ -0,0 +1,182 @@
...
+      class Predicate < Cop
+        include ConfigurableEnforcedStyle
+
+        MSG = 'Use predicate matcher `%s`'.freeze
+
+        def on_send(node)
+          expectation = correction(node)
+          return if expectation.nil?
+
+          add_offense(
```
backus: Can you put this on one line? It is well within the column limit

lib/rubocop/cop/rspec/predicate.rb
```diff
@@ -0,0 +1,182 @@
+# encoding: utf-8
+
+module RuboCop
+  module Cop
+    module RSpec
+      # Prefer using predicate matcher instead of truthiness
```
backus: Ditto here https://github.com/nevir/rubocop-rspec/pull/83/files#r74694733

lib/rubocop/cop/rspec/predicate.rb
```diff
@@ -0,0 +1,182 @@
...
+      #
+      # @example
```
backus: Ditto for the example section. Can you add an example for each style? See https://github.com/nevir/rubocop-rspec/pull/83/files#r74694733

lib/rubocop/cop/rspec/predicate.rb
backus: FWIW once #174 is merged this will always be the same as what is in the config file

lib/rubocop/cop/rspec/predicate.rb
backus: By the way I like that you extracted these corrector classes that is much better than stuffing corrections into the cop which people tend to do

lib/rubocop/cop/rspec/predicate.rb
```diff
@@ -0,0 +1,182 @@
...
+          node.source
+            .gsub(object.source, expectation)
+            .gsub(args.source, matcher)
+            .gsub(".#{action}", actionable)
+        end
+
+        def correctable?
+          return false unless object
+          return false unless args
+          return false if args.source =~ /^(be_truthy|be_falsey)$/
```
backus: Can you extract this regular expression to a constant?

lib/rubocop/cop/rspec/predicate.rb
backus: Can you nest these within `Predicate`? Right now this is `RuboCop::Cop::RSpec::Corrector`

lib/rubocop/cop/rspec/predicate.rb
backus: Also could you use `\A` and `\z` instead of `^` and `$`. They are different and you probably want `\A`

lib/rubocop/cop/rspec/predicate.rb
```diff
@@ -0,0 +1,182 @@
...
+        def correction
+          return unless correctable?
+
+          node.source
+            .gsub(object.source, expectation)
+            .gsub(args.source, matcher)
+            .gsub(".#{action}", actionable)
+        end
+
+        def correctable?
```
backus: Could you rewrite this method as simply

```ruby
def correctable?
  object && args && expectation && args.source !~ /^(be_truthy|be_falsey)$/
end

**Comment by backus** [comment]
I wonder if it would be possible to hook into RSpec for some of the rewriting rules here. It might be tricky to 1-to-1 match their inflection although it looks like you've done a good job so far

File: lib/rubocop/cop/rspec/predicate.rb [hunk 20]
```diff
@@ -0,0 +1,182 @@
...
+        end
+        # rubocop:enable Metrics/MethodLength
+      end
+
+      # Corrector for boolean and truthiness style
+      class BooleanCorrector < Corrector
+        private
+
+        def requires_truthiness?
+          style == :truthiness
```
backus: instance of https://github.com/nevir/rubocop-rspec/pull/83/files#r74694726

lib/rubocop/cop/rspec/predicate.rb
```diff
@@ -0,0 +1,182 @@
...
+          when /^(have_)\w+/
+            matcher.gsub('have', 'has')
+          else
+            matcher.gsub('be_', '')
+          end
+        end
+        # rubocop:enable Metrics/MethodLength
+
+        def matcher
+          if action == :to
```
backus: https://github.com/nevir/rubocop-rspec/pull/83/files#r74694726

backus:
@haeky I did a pass over this code. You're doing a great job with this cop. The nature of this cop is complex so I'm sorry this PR is moving slowly. 

I've left some preliminary style and testing flags. Can you checkout mutant as well? I've been making sure I have 100% coverage with mutant before I merge new cops and I'd like to hold contributions to the same standard. To run mutant add this to your `Gemfile.local`

```ruby
gem 'mutant-rspec', github: 'mbj/mutant'
```

and then run this:

```sh
bundle exec mutant --use rspec -- 'RuboCop::Cop::RSpec*::Predicate' 'RuboCop::Cop::RSpec::Corrector*' 'RuboCop::Cop::RSpec::PredicateCorrector*' 'RuboCop::Cop::RSpec::BooleanCorrector*'
```

backus:
@haeky can you also please include the `SpecOnly` mixin into your cop?

backus:
@haeky if you add me to your fork I can help fix some of the smaller flags. This should reduce the time this spends in review

lib/rubocop/cop/rspec/predicate.rb
bquorning: Wouldn’t it be clearer if this were in two lines?
```rb
return false if args.source == "be_truthy"
return false if args.source == "be_falsey"
```

lib/rubocop/cop/rspec/predicate.rb
```diff
@@ -0,0 +1,182 @@
...
+        end
+      end
+
+      # Corrector for the predicate style
+      class PredicateCorrector < Corrector
+        private
+
+        def actionable
+          if args.source.include?('false')
+            '.not_to'
```
bquorning: This should probably take into account the `EnforcedStyle` configuration for `RSpec/NotToNot`.

backus: Hm I don't think it needs to. RuboCop will keep doing autocorrect sweeps when one cop changes something (and it looks out for infinite loops). I believe it does this so that unrelated cops do not have to account for a huge amount of style variance

lib/rubocop/cop/rspec/predicate.rb
```diff
@@ -0,0 +1,182 @@
...
+          method_args = object.method_args.map(&:source).join(', ')
+          matcher += "(#{method_args})" unless method_args.empty?
+
+          matcher
+        end
+
+        # rubocop:disable Metrics/MethodLength
+        def matcher_for_method(method)
+          case method
+          when /^(respond_to|include)$/
```
bquorning: Wouldn’t it be easier to read using strings instead of regular expressions?
```rb
when 'respond_to', 'include'
```
etc.

lib/rubocop/cop/rspec/predicate.rb
bquorning: That makes sense – thanks for the explanation.

backus: Sure. This is just my observation as a user of the auto correct feature by the way. I'm fairly certain this is the design but just be aware that in this case I'm not familiar with the actual implementation details

andyw8:
Why did you choose to close this @haeky? Do you need help with anything?

Darhazer:
:+1: for having such cop

dgollahon:
@haeky: Did you decide you didn't want to work on this? If not, please let us know so someone else can take over the effort as there seems to be some interest in the cop. If you'd like to resume working on it, that's fine also.

haeky:
@dgollahon I don't have time anymore for personal reason. Feel free to take it up from here.

dgollahon:
@haeky: Alrighty, thanks for the heads up.

Issue #36, opened 2015-04-12
RSpec/VerifiedDoubles requires instance_double in cases where duck typing is employed


CJBridges:
Duck typing - double stubs a method, but could be one of many classes *with no shared base class*

```ruby
def Runner
  def initialize(worker)
    @worker = worker
  end

  def call
    worker.call
  end
end
```

A spec for this could look something like...
```ruby
describe 'Runner' do
  let(:return_val) { 'RetVal' }
  let(:my_worker) { double(:sample_worker, call: return_val) } # ack, verified double checker

  it 'returns the result of my_worker.call'
    expect(Runner.new(my_worker).call).to be(return_val)
  end
end
```

*Proposed fix*: ??? unknown here.

*Workaround*: Define something like this instead in the spec file:
```ruby
class Spec::ExampleWorker
  def call; end
end
```

Then change the reference to:
```ruby
instance_double('Spec::ExampleWorker')
```

CJBridges:
P.S. These were real world problems encountered while trying to enable the verified double checker (which I heartily approve of!)

andyw8:
I would be inclined to choose an arbitrary existing class that implemented the particular interface.

CJBridges:
I like the view that tests can serve documentation of the interface of the class and not really tied to any particular caller (which may vanish completely separate from the class).

I tried a couple approaches including OpenStruct for some simple cases or creating a real instance that behaves like I wanted to... something like... `Object.new.tap { |o| def o.method; 'foo'; end }`  (this is tough since the body of the method doesn't evaluate in rspec context, harder to vary values)

Overall, I haven't found anything that resolves the issue any better than the workaround I noted in the description, unfortunately.  The joys of duck typing!

backus:
@andyw8 similar to #35 it should be alright to close this, correct?

andyw8:
Yes, resolved by #99.

Issue #35, opened 2015-04-12
RSpec/VerifiedDoubles is overaggressive when nothing stubbed


CJBridges:
Note: no tests actually run.  May be dumb error in code.

I'm writing a class which is used for plumbing a single value.  In the real world, this class takes 4 values, processes 3 of them, and passes one of them through unchanged.

```ruby
def MyService
  def initialize(opaque_obj)
    @opaque_obj = opaque_obj
  end

  def call
    my_internal_service.new(@opaque_obj).call
  end

  private

  def my_internal_service  # for stubbing
    MyInternalService
  end
end
```

A spec for this could look something like...
```ruby
describe 'MyService' do
  let(:opaque_obj) { double(:opaque_obj) }  # uh oh, verified double checker....
  let(:my_internal_service) { class_double('MyInternalService') }
  let(:retval) { true }
  subject { MyService }

  before do
    allow(subject).to_receive(my_internal_service).and_return(my_internal_service) # stub external dependency
  end

  it 'passes var to MyInternalService'
    # assert that input made it to correct spot in output.
    expect(my_internal_service).to receive('new').with(opaque_obj).and_return(retval)

    #and verify that my service passes the retval back
    expect(subject.new(opaque_obj).call).to eq(retval)
  end
end
```

So we have a case where we want to pass a class through, but treat it as an opaque object, never calling anything on it, modifying it, or even looking at it.  From a double perspective, we shouldn't care about being a verifying double or what type it is at all.  Keeping a notation in the spec that we're expecting a non-specific class is valuable, but there's no value in verifying it's an instance of a specific class (could be a variety, perhaps unrelated. Go duck typing!)

*Proposed fix*:
  - if no stubbed methods, expects, or allows applied to the double, let it be!  No errors.

*Workaround*:
  - Use '' (or similar constant) as the value instead.  This loses the information that it's really a class typically being passed through.  Can also mess with equality checks later if you're fast and loose with reference vs. value equality.

CJBridges:
P.S. These were real world problems encountered while trying to enable the verified double checker (which I heartily approve of!)

andyw8:
Hi @CJBridges 

I was the author of that cop, but I've only just seen your message today.

I agree fully that there is no point in using a verified double if nothing is being stubbed.

To permit a non-verifying double, we'd need to ensure several things:

* There is no hash supplied when creating the double, e.g. `double(:widget, foo: "bar")`
[...]

PR #87, opened 2016-05-04
Enhance VerifiedDoubles cop

andyw8:
Based on the discussions in #35 and #36.

With these changes, I believe it's appropriate to have this cop enabled by default.

cc: @CJBridges

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,12 +5,13 @@ language: ruby
 rvm:
   - 1.9.3
   - 2.0.0-p648
-  - 2.1.8
-  - 2.2.4
-  - 2.3.0
+  - 2.1.10
+  - 2.2.5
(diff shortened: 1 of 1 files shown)


CJBridges:
Thanks @andyw8 - looks good to me.  I can see an argument for `double` with no arguments should be an error, but it never comes up in practice for us, so I'm far from picky about it.

You get the :+1: from me that it meets my requirements/needs.

andyw8:
I think it would good to have a separate cop to check for unnamed doubles, both normal and verifying.

nijikon:
@andyw8 please rebase on top of master.

andyw8:
I messed up the rebase and GitHub now thinks this is already merged - I'll open a new PR.

PR #85, opened 2016-05-02
Add ExampleLength cop

andyw8:
This was suggested by @agrimm in issue #40

A long example is a form of the Obscure Test smell:
http://xunitpatterns.com/Obscure%20Test.html

I've chosen a default maximum of 5 lines on the basis that a typical
example would contain 3 lines of setup, one line of execution and one
line of assertion. However, this is quite dependent on the style used in
the codebase. Suites which rely more on `let` and `before` blocks would
likely have shorter example blocks.

Somewhat ironically, I had to set the configuration for rubocop-rspec
itself to quite a high value (12 lines).

PR #99, opened 2016-05-15
Enhance VerifiedDoubles cop

andyw8:
This replaces #87 which was accidentally closed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## master (unreleased)
 
+* Expand `VerifiedDoubles` cop to check for `spy` as well as `double`. ([@andyw8][])
+* Enable `VerifiedDoubles` cop by default. ([@andyw8][])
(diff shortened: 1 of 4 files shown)


nijikon:
@andyw8 do we want ```IgnoreSymbolicName``` be ```true``` by default?

andyw8:
@nijikon I believe so, for the reasons mentioned by @CJBridges in #35 and #36.

CJBridges:
I'll be using the option, but I suggest defaulting it to "false" so the cop defaults to its strongest state and behaves obviously.  The rule with its subtleties can be turned on intentionally.

andyw8:
Ok, I've updated the PR to remove that default.

nijikon:
@andyw8 please rebase on top of master

andyw8:
I'll do this after #85 is merged, otherwise I think one of these will end up conflicted again.

nijikon:
@andyw8 done, rebase please

andyw8:
@nijikon rebased.

PR #104, opened 2016-05-24
Add DefaultHookArgument cop

tsigo:
This cop will add offenses when an RSpec hook method (`before`, `after`,
`around`) receives the default `:each` or `:example` argument.

### DIFF
--- a/config/default.yml
+++ b/config/default.yml
@@ -2,6 +2,10 @@ RSpec/AnyInstance:
   Description: 'Check that instances are not being stubbed globally'
   Enabled: true
 
+RSpec/DefaultHookArgument:
+  Description: 'Omit the default `:each` argument for RSpec hooks.'
+  Enabled: true
+
 RSpec/DescribeClass:
   Description: 'Check that the first argument to the top level describe is the tested class or module.'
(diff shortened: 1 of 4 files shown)


mockdeep:
Was just thinking this would be nice to have. We actually use `before(:each)` rather than `before`, so it would be nice if it could be configured to enforce one way or the other.

backus:
Thanks @tsigo! This is in master now: https://github.com/nevir/rubocop-rspec/commit/c60a86c3b24a9287fe1c591d349a3240c142e228

I changed up a few things:

* I renamed this cop to `HookArgument`
* I added a few more tests before changing any of your code
* I added support for `EnforcedStyle` (so now you can either enforce `before { }` syntax, `before(:each) { }` syntax, or `before(:example) { }` syntax)
    * This also makes this compatible with rubocop's `--auto-gen-config` flag which will build up a TODO file and infer what a projects preference is.
* I added autocorrection support so now you can just run `--auto-correct` with rubocop and it will rewrite all violations

Thanks for your work on this!

tsigo:
@backus Fantastic, thanks!

PR #105, opened 2016-05-24
Prefer usage of be_valid_for

Darhazer:
Using `be_valid` and `be_invalid` leads to fragile tests. When you add new validation, all tests that use `be_valid` should be updated to add the new property; `be_invalid` may fail for validation fails different that the one in the example


andyw8:
I'm not familiar with this matcher. Is it added by a gem? I can't find any native `ActiveModel::Validations#valid_for?`method in the API docs.

Darhazer:
Sorry, I was completely unaware that it's not a standard one (and it's a very useful one). I guess I should provide separate gem.

Anyway, I'm just starting writing cops for tests so I'll try few more... I'll just make sure that it's part of the core rspec before the next PR.

Thanks for looking at this.

PR #107, opened 2016-05-31
New cop RSpec/DocString to check Example/Example Group quoting

tjwp:
We prefer to use double-quoted strings for our example documentation, since those strings may contain contractions or interpolation. Elsewhere we prefer single-quoted strings. As a result, we have Style/StringLiterals disabled for our spec directory.

This change adds a cop to check string quoting just for examples and example groups.

Even better would be to also have a cop that does the complement -- only checking string literals that are not doc strings. I may work on that as a follow-up to this, but this seems like a good first step.

I've added the cop as disabled by default.


backus:
Hey @tjwp thanks a ton for the PR. @nevir gave me permission to take over maintaining this project so I'm going through the open PRs right now in order to get things moving again.

With that said, I'm not sure how I feel about this particular cop. I understand this is a policy that your dev team follows but I don't know if anyone else would really benefit from how specific this cop is. I think you probably agree to some extent given that you opted to make it disabled by default. 

With this said, I appreciate the work you put into this cop, but I don't know if it is right for this project. I'm going to close this PR for now. I am always open to having my opinion changed though so feel free to continue discussion here.

tjwp:
No problem. My inclination was to share it just in case anybody else found it useful, and additional eyes from the community are always helpful.

But if it is too obscure I'm happy to maintain it in our own gem, along with the opposite cop that only checks non-doc strings.

tjwp:
For anybody that stumbles across this discussion, the cop is publicly available as part of https://github.com/salsify/salsify_rubocop under the name `Salsify/RspecDocString`.

PR #108, opened 2016-06-07
Add Before/After(:all) cop

cfabianski:
As described in the cop, making use of `before(:all)` or `after(:all)` might lead to flakiness when your specs are being split out / parallelised. That's also an issue as the `before/after(:all)` aren't rolled back as per the RSpec documentation https://relishapp.com/rspec/rspec-rails/docs/transactions

This rule is turned off by default because not everyone would face this issue 😸


lib/rubocop/cop/rspec/before_after_all.rb
```diff
@@ -0,0 +1,42 @@
+# encoding: utf-8
+# frozen_string_literal: true
+
+module RuboCop
+  module Cop
+    module RSpec
+      # Check that before/after(:all) isn't being used.
+      # This leads to flaky tests when being split out.
```
andyw8: By 'split out' do you mean run in parallel?

cfabianski: Yes!

lib/rubocop/cop/rspec/before_after_all.rb
```diff
@@ -0,0 +1,42 @@
...
+      #     after(:all) { do_something_else }
+      #   end
+      #
+      #   # good
+      #   describe MyClass do
+      #     before(:each) { do_something }
+      #     after(:each) { do_something_else }
+      #   end
+      class BeforeAfterAll < Cop
+        MESSAGE = 'Avoid the use of before/after(:all) to avoid specs flakyness when split out'.freeze
```
andyw8: 'flakiness' seems to be the more common spelling. (500,000 vs 8000 matches on Google)

cfabianski: Updated! Thanks 👍

config/default.yml
```diff
@@ -23,6 +23,9 @@ RSpec/ExampleWording:
     not: does not
   IgnoredWords: []
 
+RSpec/BeforeAfterAll:
+  Description: 'Do not use before/after(:all).'
```
andyw8: I think a little more explanation would be helpful, e.g. `since data created in before(:all) is not rolled back`. https://relishapp.com/rspec/rspec-rails/docs/transactions

cfabianski: Oh yeah definitely. Thanks!

lib/rubocop/cop/rspec/before_after_all.rb
andyw8: It applies in non-parallel tests too, e.g. if you forget an `after(:all)` to clean up data, then it may stick around between test runs.

cfabianski: I'll update the description to make it clearer based on your previous feedback yeah! 💯

cfabianski:
What do you think about the latest changes?

config/default.yml
```diff
@@ -23,6 +23,9 @@ RSpec/ExampleWording:
     not: does not
   IgnoredWords: []
 
+RSpec/BeforeAfterAll:
+  Description: 'Avoid the use of before/after(:all) as they are not rolled back and leads to flakiness.'
```
andyw8: "and may lead" ?

Also, "flakiness" feels a little vague – how about something like "and may lead to database state leaking between examples"?

cfabianski: Agree. Sorry for the feedback loop I should have been more careful about the description. Thanks for taking the time to review and suggest improvements... ❤️

config/default.yml
```diff
@@ -23,6 +23,9 @@ RSpec/ExampleWording:
     not: does not
   IgnoredWords: []
 
+RSpec/BeforeAfterAll:
+  Description: 'Avoid the use of before/after(:all) as they are not rolled back and nd may lead to database state leaking between examples.'
```
andyw8: missing character – "nd may lead"

cfabianski: Updated. I've made some changes to comply with the rubocop rules as well.

cfabianski:
@andyw8 The build is now 💚 . What do you think?

andyw8:
Looks good, no other issues I can see.

cfabianski:
@andyw8 when is the next planned release, if any, please?

andyw8:
Sorry, I'm not a maintainer, only a user and contributor.

cfabianski:
@andyw8 oh ok! @nijikon might be the best to answer I guess :smile_cat:

backus:
@cfabianski I can review soon. Ping me this weekend if you haven't seen any further notifications from me. I'm trying to revive this project

config/default.yml
```diff
@@ -23,6 +23,9 @@ RSpec/ExampleWording:
     not: does not
   IgnoredWords: []
 
+RSpec/BeforeAfterAll:
+  Description: 'Avoid the use of before/after(:all) as they are not rolled back and may lead to database state leaking between examples.'
+  Enabled: false
```
backus: Why make this disabled by default? I doubt many people at all will benefit from this if it is disabled by default.

config/default.yml
```diff
@@ -23,6 +23,9 @@ RSpec/ExampleWording:
     not: does not
   IgnoredWords: []
 
+RSpec/BeforeAfterAll:
+  Description: 'Avoid the use of before/after(:all) as they are not rolled back and may lead to database state leaking between examples.'
```
backus: I don't think this should talk about the database. rubocop-rspec is for any ruby code. Maybe talk about global state and how test state should be set up and torn down before and after each test?

backus:
@cfabianski a few thoughts:

* what about `before(:suite)` and `after(:suite)`?

* do you think this should flag `before(:all)` and `after(:all)` within RSpec configuration?

config/default.yml
cfabianski: I agree. I will update to talk about global state instead of database.

config/default.yml
cfabianski: My thought was that it would ease the transition but I'd be happy to enable it by default :)

lib/rubocop/cop/rspec/before_after_all.rb
```diff
@@ -0,0 +1,45 @@
+# encoding: utf-8
+# frozen_string_literal: true
+
+module RuboCop
+  module Cop
+    module RSpec
+      # Check that before/after(:all) isn't being used
+      # as they are not rolled back.
+      # See https://relishapp.com/rspec/rspec-rails/docs/transactions
```
backus: Related to my comment about this best practice being useful for more than just database configuration, could you either remove this link (or preferably) find a non-rails specific link to cite?

config/default.yml
backus: Cool. I think everyone benefits more from it being enabled by default.

backus:
@cfabianski the main rubocop project uses this [node_pattern DSL](https://github.com/bbatsov/rubocop/blob/master/lib/rubocop/node_pattern.rb) which seems useful. It isn't used in this project yet but would you like to look into it? It might make it even cleaner to assert match the nodes we care about. As an example of what it is useful for, here is how one of their cops use it:

```ruby
def_node_matcher :double_negative?, '(send (send _ :!) :!)'

def on_send(node)
  return unless double_negative?(node) && node.loc.selector.is?('!')
  add_offense(node, :selector)
end
```

config/default.yml
backus: Thanks @cfabianski!

lib/rubocop/cop/rspec/before_after_all.rb
```diff
@@ -0,0 +1,45 @@
...
+
+        BEFORE_AFTER_METHODS = [
+          :before,
+          :after
+        ].freeze
+
+        ALL_PAIR = s(:sym, :all)
+
+        def on_send(node)
+          _receiver, method_name, *args = *node
```
backus: You could write this as `_, selector, (first_arg,) = *node` and then in your conditional below could be more precise by writing `args.first.eql?(ALL_PAIR)`

lib/rubocop/cop/rspec/before_after_all.rb
```diff
@@ -0,0 +1,45 @@
...
+        MESSAGE = 'Avoid the use of before/after(:all) '\
+          'as they are not rolled back and may lead to database state '\
+          'leaking between examples'.freeze
+
+        BEFORE_AFTER_METHODS = [
+          :before,
+          :after
+        ].freeze
+
+        ALL_PAIR = s(:sym, :all)
```
backus: What do you think of `ALL_SYM`? I wasn't sure at first what you meant by `ALL_PAIR`

lib/rubocop/cop/rspec/before_after_all.rb
```diff
@@ -0,0 +1,45 @@
...
+
+        def on_send(node)
+          _receiver, method_name, *args = *node
+          return unless BEFORE_AFTER_METHODS.include?(method_name)
+          return unless args.include?(ALL_PAIR)
+
+          add_offense(node, :expression, MESSAGE)
```
backus: If you rename the `MESSAGE` constant to `MSG` then you can omit this parameter (`add_offense` will automatically use `MSG`)

spec/rubocop/cop/rspec/before_after_all_spec.rb
```diff
@@ -0,0 +1,63 @@
...
+      ]
+    end
+
+    let(:expected_error) do
+      'Avoid the use of before/after(:all) '\
+        'as they are not rolled back '\
+        'and may lead to database state leaking between examples'
+    end
+
+    it 'reports 2 offenses' do
```
backus: Thoughts on removing this test? The test below (`"reports the lines for these offenses "`) is asserting that there are offenses on lines 2 and 3. Unless I'm missing something that should be equivalent to asserting the number of cops

spec/rubocop/cop/rspec/before_after_all_spec.rb
```diff
@@ -0,0 +1,63 @@
...
+    end
+
+    it 'reports the lines for these offenses' do
+      inspect_source(cop, code)
+      expect(cop.offenses.map(&:line).sort).to eq([2, 3])
+    end
+
+    it 'describes the offenses' do
+      inspect_source(cop, code)
+      expect(cop.messages).to eq([expected_error, expected_error])
```
backus: Not saying that this is better but you could write

```ruby
expect(cop.messages).to all(eq(expected_error))
```

backus:
Hey @cfabianski I finished doing a pass over this PR. Let me know when you would like to review again. (@andyw8 thanks for helping)

config/default.yml
andyw8: I think it would be helpful to cover both situations, Rails and non-Rails, since using Rails is such a common use case. For example:

> Avoid the use of before/after(:all) as it prevents examples from being fully isolated, meaning state may leak between tests. (For Rails apps, this means the database may not be cleared between tests').

lib/rubocop-rspec.rb
```diff
@@ -21,3 +21,4 @@
 require 'rubocop/cop/rspec/multiple_describes'
 require 'rubocop/cop/rspec/not_to_not'
 require 'rubocop/cop/rspec/verified_doubles'
+require 'rubocop/cop/rspec/before_after_all'
```
andyw8: It looks the existing convention is to list the cops alphabetically.

andyw8:
@cfabianski are you still interested in working on this? I think it would very useful.

lib/rubocop/cop/rspec/before_after_all.rb
andyw8: We could link to this discussion: https://github.com/rspec/rspec-core/issues/573

andyw8:
Superceded by https://github.com/backus/rubocop-rspec/pull/252

PR #119, opened 2016-07-23
Drop support for Ruby 2.0

backus:


bbatsov:
I believe you meant to say 2.0. :-) For what is worth RuboCop will continue supporting 2.0 for a while.

backus:
@bbatsov whoops thanks for catching that.

> For what is worth RuboCop will continue supporting 2.0 for a while.

Yeah I considered matching what RuboCop supports but the maintenance of this gem has been someone lacking so I figured it would make it easier (if it is mainly just me) to reduce to number of versions I support.

andyw8:
I guess this means all the magic encoding comments can be removed?

backus:
> I guess this means all the magic encoding comments can be removed?

Good catch yeah I'll remove them

PR #120, opened 2016-07-23
Remove magic file encoding comments

backus:


backus:
Thanks @andyw8 for pointing this out.

Some files have `# frozen_string_literal: true`. I think this isn't helpful as long as this gem supports versions below 2.3. I'm not certain on this though. Thoughts?

andyw8:
Since CI is running 2.3, then I would say there's still an advantage to having the `frozen_string_literals` check. It helps to ensure that strings aren't inadvertently mutated, which presumably will become the default behaviour in some upcoming version of Ruby.

backus:
> It helps to ensure that strings aren't inadvertently mutated

Good point

PR #121, opened 2016-07-24
Refactor verified doubles

backus:


.rspec
```diff
@@ -1 +1,3 @@
 --require spec_helper
+--color
```
andyw8: Was it intentional to commit this config change as part of the PR?

backus: Yeah I could have put it in another PR or on master but I figured it was fine here

backus:
@andyw8 would you like to review these changes? My goals:

1. 100% mutation coverage of `VerifiedDoubles` cop
2. Subsequent refactor to use `def_node_matcher` after full mutation coverage

backus:
@andyw8 I'm going to merge this for now so I don't get too backed up with PRs. Would still love feedback on the changes

andyw8:
Sorry, I didn't get a chance to look at this until now.

I hadn't come across the `def_node_matcher` approach before, so had to skim through [node_pattern.rb](https://github.com/bbatsov/rubocop/blob/master/lib/rubocop/node_pattern.rb) to understand what's going on. Even though it's more concise, I do wonder whether it might put off casual contributors, than if everything was in plain Ruby?

PR #125, opened 2016-07-27
Order selectors in Focus cop

andyw8:
Just something I spotted when reviewing #123.

I think it's easier to visually scan, and to avoid accidental duplicates, when they are ordered.


backus:
Nice! Thanks @andyw8 :)

PR #131, opened 2016-08-01
Change minimum ruby version to 2.2 and move jruby to allowed failures on travis

backus:


andyw8:
It doesn't affect me, but I'm curious why you chose to drop 2.1 support?

backus:
Lower maintenance burden basically. 2.1 will be EOL soon and I imagine people using this gem do a decent job of keeping up to date. I want to add a lot to this gem so having to support a lot of stuff I don't use is a huge drag

atz:
While undertaking the maintenance burden of legacy support, having builds break for including a rubocop check is also a huge drag.

backus:
Sorry @atz? Are you saying your build broke because I removed support for ruby 2.1 or because of a new cop that I added?

tsigo:
This was a problem for us @gitlab as well. We want to use the latest rubocop-rspec to get some of the new cops enabled, but we still need to support Ruby 2.1.

Do we know how long RuboCop itself will support 2.1? It might make sense to follow its lead.

backus:
I'd be open to supporting it again now that it won't impose a cost on CI.

tsigo:
Thanks, @backus.

For now we've [reverted back to 1.5.0](https://gitlab.com/gitlab-org/gitlab-ce/merge_requests/6444) but I hope 2.1 can be supported again soon.

bbatsov:
@rspeicher We typically support Rubies until they reach their EOL. We've decided to make an exception with 2.0 and 2.1 and supported them a bit longer than needed, but the support will be dropped in RuboCop in its next release.

PR #135, opened 2016-08-01
Improve coverage for FilePath cop

backus:


spec/rubocop/cop/rspec/file_path_spec.rb
```diff
@@ -110,8 +139,8 @@
   it 'handles alphanumeric class names' do
     inspect_source(
       cop,
-      'describe IPV6 do; end',
-      'ipv6_spec.rb'
+      'describe IPv4AndIPv6 do; end',
+      'i_pv4_and_i_pv6_spec.rb'
```
bquorning: I don’t see this as an improvement…

backus: Oh whoops I don't think I did that intentionally.

bquorning: That’s what I thought :-) See #195 for a fix.

PR #139, opened 2016-08-01
Improve test coverage for DescribeClass cop

backus:


spec/rubocop/cop/rspec/describe_class_spec.rb
```diff
@@ -23,6 +23,11 @@
                                 'the class or module being tested.'])
   end
 
+  it 'checks describe statements after a require' do
```
andyw8: Should this description be different from the one above?

backus: Good catch thanks for pointing that out

PR #140, opened 2016-08-01
Improve coverage of DescribeMethod cop

backus:


lib/rubocop/rspec/util.rb
```diff
@@ -0,0 +1,19 @@
...
+
+module RuboCop
+  module RSpec
+    # Utility methods
+    module Util
+      # Error raised by `Util.one` if size is less than zero or greater than one
+      SizeError = Class.new(IndexError)
+
+      # Return only element in array if it contains exactly one member
+      def one(array)
```
andyw8: Hey @backus, I just stumbled across this snippet and I really like it. It always feel awkward to call `first` or `last` on a collection when you know it should contain one element. It would be nice to have this in ActiveSupport, or maybe even Ruby itself. Did you borrow this from somewhere else or just come up with it?

PR #143, opened 2016-08-02
Add RSpec/NamedSubject cop

### LINKED ISSUE #53: Name `subject` when referenced explicitly
When referencing subject explicitly is necessary it should be given a name

**good**

```ruby
RSpec.describe UserCreator do
  subject(:create) { described_class.new(path) }

  it 'creates a user' do
    expect { create }.to change(User, :count).by(1)
  end
end
```

**bad**

```ruby
RSpec.describe UserCreator do
  subject { described_class.new(path) }

  it 'creates a user' do
    expect { subject }.to change(User, :count).by(1)
  end
end
```

thoughts?  Happy to do a PR

backus:
Close #53


lib/rubocop/cop/rspec/named_subject.rb
```diff
@@ -0,0 +1,68 @@
+# encoding: utf-8
```
andyw8: No longer needed?

lib/rubocop-rspec.rb
```diff
@@ -21,3 +21,4 @@
 require 'rubocop/cop/rspec/multiple_describes'
 require 'rubocop/cop/rspec/not_to_not'
 require 'rubocop/cop/rspec/verified_doubles'
+require 'rubocop/cop/rspec/named_subject'
```
andyw8: Let's keep this in alphabetical order.

lib/rubocop/cop/rspec/named_subject.rb
```diff
@@ -0,0 +1,68 @@
...
+# frozen_string_literal: true
+
+module RuboCop
+  module Cop
+    module RSpec
+      # Avoid referencing `subject` explicitly in specs
```
andyw8: Perhaps the message should explain *why* this bad?

andyw8: Oops, I mean the `MSG` constant, not the comment.

lib/rubocop/cop/rspec/named_subject.rb
backus: Nice catch

lib/rubocop/cop/rspec/named_subject.rb
backus: Hm well the other messages don't explain why the code is bad but many do provide links. Maybe I'll try to provide a link in the YARD docs

lib/rubocop/cop/rspec/named_subject.rb
```diff
@@ -0,0 +1,68 @@
...
+          unnamed_subject(node, &block)
+
+          node.children.each do |child|
+            subject_usage(child, &block)
+          end
+        end
+      end # NamedSubject
+    end # Devtools
+  end # Cop
+end # RuboCop
```
andyw8: I don't think this `end` commenting is used elsewhere in the project?

spec/rubocop/cop/rspec/named_subject_spec.rb
```diff
@@ -0,0 +1,63 @@
+# encoding: utf-8
+# frozen_string_literal: true
+
+RSpec.describe RuboCop::Cop::RSpec::NamedSubject do
+  subject(:cop) { described_class.new }
+
+  it 'checks it for explicit subject usage' do
+    expect_violation(<<-SRC)
```
andyw8: The other cop specs use `<<-RUBY` rather than `<<-SRC`, probably worth sticking to that?

lib/rubocop/cop/rspec/named_subject.rb
backus: Damn good catch. This is leftover from transferring this from rubocop-devtools

spec/rubocop/cop/rspec/named_subject_spec.rb
backus: Yeah good point. Using `<<-RUBY` also buys you syntax highlighting in many places

lib/rubocop/cop/rspec/named_subject.rb
backus: I haven't explicitly explained why but I've updated this comment to make it more obvious. Now it says

> Give `subject` a descriptive name if you reference it directly

backus:
Thanks for the feedback @andyw8! Keep it coming :smile:

spec/rubocop/cop/rspec/named_subject_spec.rb
```diff
@@ -0,0 +1,62 @@
+# frozen_string_literal: true
+
+RSpec.describe RuboCop::Cop::RSpec::NamedSubject do
+  subject(:cop) { described_class.new }
+
+  it 'checks it for explicit subject usage' do
```
andyw8: Should this be "checks it and specify for explicit subject usage" ?

(also, perhaps enclose the keywords in backticks for emphasis and readability?).

andyw8:
I don't see any other issues, so :+1: from me.

spec/rubocop/cop/rspec/named_subject_spec.rb
backus: Great point

PR #147, opened 2016-08-03
Add SkipBlocks option for DescribedClass cop

backus:
Now the following code will **never** be flagged:

```ruby
describe MyClass do
  Class.new  { foo = MyClass }
  Module.new { bar = MyClass }
end
```

With the `SkipBlocks` option enabled the following code would receive the following flags:

```ruby
describe MyClass do
  controller(ApplicationController) do
    bar = MyClass
  end

  before do
    MyClass
    ^^^^^^^ Use `described_class` instead of `MyClass`

    Foo.custom_block do
      MyClass
    end
  end
end
```

With `SkipBlocks` disabled (the default option) the code would receive the following flags:

```ruby
describe MyClass do
  controller(ApplicationController) do
    bar = MyClass
          ^^^^^^^ Use `described_class` instead of `MyClass`
  end

  before(:each) do
    MyClass
    ^^^^^^^ Use `described_class` instead of `MyClass`

    Foo.custom_block do
      MyClass
      ^^^^^^^ Use `described_class` instead of `MyClass`
    end
  end
end
```

### DIFF
--- a/lib/rubocop/cop/rspec/described_class.rb
+++ b/lib/rubocop/cop/rspec/described_class.rb
@@ -30,7 +30,15 @@ module RuboCop
             (args) $_)
         PATTERN
 
-        def_node_matcher :scope_change?, '{def class module}'
+        def_node_matcher :scope_changing_syntax?, '{def class module}'
+
+        def_node_matcher :common_instance_exec_closure?, <<-PATTERN
+          (block
+            (send
+              (const nil {:Class :Module}) :new
+              ...)
+            ...)
+        PATTERN
 
         def on_block(node)
           describe, described_class, body = described_constant(node)
@@ -58,6 +66,10 @@ module RuboCop
           node.children.each do |child|
             find_constant_usage(child, described_class, &block)
           end
+        end
+
+        def scope_change?(node)
+          scope_changing_syntax?(node) || common_instance_exec_closure?(node)
         end
       end
     end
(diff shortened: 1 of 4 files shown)


backus:
@andyw8 would love your feedback here

lib/rubocop/cop/rspec/described_class.rb
```diff
@@ -20,41 +20,94 @@ module RSpec
...
+          (block (send (const nil {:Class :Module}) :new ...) ...)
+        PATTERN
 
-          _receiver, _method_name, object = *method
-          return unless object && object.type.equal?(:const)
+        def_node_matcher :rspec_block?, <<-PATTERN
+          (block (send nil {#{RSPEC_BLOCK_METHODS}} ...) ...)
+        PATTERN
+
+        def_node_matcher :scope_changing_synax?, '{def class module}'
```
andyw8: Typo (synax -> syntax).

backus: Thank you!

andyw8:
I would need to spend some more time to better understand this, but I don't see any blockers, so feel free to merge if you're happy with it.

backus:
Thanks @andyw8.

After today's opened issue about semver I'm considering just always releasing a minor version bump unless I'm releasing a bugfix. Think that sounds reasonable?

andyw8:
Yeah, seems the right approach.

Issue #56, opened 2015-12-15
Contexts without any tests


agrimm:
In the following code, I accidentally created a context without any tests, because the `it 'is chunky'` is after the context has ended.

    describe Bacon do
      let(:bacon) { Bacon.new(chunkiness) }
      let(:chunkiness) { false }
      context 'extra chunky' do
        let(:chunkiness) { true }
      end

      it 'is chunky' do
        expect(bacon.chunky?).to be_truthy
      end
    end

Contexts without any tests should generate an offense.

nijikon:
:+1:

backus:
Interesting. I wish it was easy for us to have a namespace of our own like RuboCop does. Then this could be classified as `RSpec/Lint/EmptyContext`

backus:
This is in master now

PR #152, opened 2016-08-04
Add EmptyExampleGroup cop

backus:


backus:
@andyw8 and @agrimm could you review this? This is an implementation of #56

lib/rubocop/cop/rspec/empty_example_group.rb
```diff
@@ -0,0 +1,76 @@
+# frozen_string_literal: true
+
+module RuboCop
+  module Cop
+    module RSpec
+      # Checks if test is focused.
```
andyw8: Wrong comment?

lib/rubocop/cop/rspec/empty_example_group.rb
```diff
@@ -0,0 +1,76 @@
...
+      #
+      # @example
```
andyw8: Should this document the `CustomIncludeMethods` option?

lib/rubocop/cop/rspec/empty_example_group.rb
backus: Ah shucks. Thanks

lib/rubocop/cop/rspec/empty_example_group.rb
backus: Probably. I need to look into how this is usually documented then

PR #156, opened 2016-08-05
Add RSpec/NestedGroups cop

backus:


lib/rubocop-rspec.rb
```diff
@@ -21,6 +21,7 @@
 require 'rubocop/cop/rspec/instance_variable'
 require 'rubocop/cop/rspec/example_length'
 require 'rubocop/cop/rspec/multiple_describes'
+require 'rubocop/cop/rspec/nested_context'
```
andyw8: Should be after `named_subject`.

andyw8:
If I'm reading this correctly, it allows no nesting at all? I wonder if it should be configurable.

backus:
> If I'm reading this correctly, it allows no nesting at all? I wonder if it should be configurable.

Yeah I agree

andyw8:
Nice. I love this kind of cop – it's a very simple rule, but it can really help push code in the right direction since it highlights the smell of complex setup.

PR #157, opened 2016-08-06
Add RSpec/ExpectActual cop

### LINKED ISSUE #17: Check `expect` to provide actual value
`expect` should contain the actual value, not the expected value. An idea would be to check for thinks like array, string and hash inside of `expect`.

https://www.relishapp.com/rspec/rspec-expectations/v/3-0/docs/built-in-matchers

backus:
```ruby
# bad
expect(5).to eq(price)
expect(/foo/).to eq(pattern)
expect("John").to eq(name)

# good
expect(price).to eq(5)
expect(pattern).to eq(/foo/)
expect(name).to eq("John")
```

Fixes #17


backus:
@andyw8 could you review this?

backus:
Also @andyw8 I'd be interested in your thoughts on whether this should have an autocorrect.

backus:
@andyw8 I'm going to merge this as is now but I'm not releasing so any feedback you have will still be implemented

lib/rubocop/cop/rspec/expect_actual.rb
```diff
@@ -0,0 +1,77 @@
...
+          str
+          sym
+          complex
+          rational
+          regopt
+        ).freeze
+
+        COMPLEX_LITERALS = %i(
+          array
+          hash
```
andyw8: Should `set` be here too?

lib/rubocop/cop/rspec/expect_actual.rb
```diff
@@ -0,0 +1,77 @@
...
+          pair
```
andyw8: I haven't come across this before, how would it look in code?

config/default.yml
```diff
@@ -29,6 +29,10 @@ RSpec/EmptyExampleGroup:
   Enabled: true
   CustomIncludeMethods: []
 
+RSpec/ExpectActual:
+  Description: 'Checks for `expect(...)` with literal values'
```
andyw8: Perhaps clarify this to be "with literal values on the left hand side"?

andyw8:
Re. the name – best I can think of is `ExpectationLiteralOnLeftHandSide` but it's not great. Happy to leave it as is.

config/default.yml
backus: Good idea

backus:
@andyw8 yeah I think that might be too verbose for my taste.

config/default.yml
backus: I'm just going to change it to

> Checks for `expect(...)` calls containing literal values

I don't think "left hand side" makes things very clear

PR #161, opened 2016-08-06
Add SubjectStub cop

backus:


backus:
@andyw8 not ready for prime time at all but this should give you an idea of the bare minimum complexity for a cop like this

backus:
I think this is actually pretty stable now. Running this against a lot of local projects yielded good results.

andyw8:
@turino As this PR is already merged, would you mind creating an issue for these enhancements? I agree they'd both be useful.

turino:
@andyw8 I'd be happy to

PR #168, opened 2016-08-08
Add LeadingSubject cop

backus:


config/default.yml
```diff
@@ -67,6 +67,10 @@ RSpec/InstanceVariable:
 RSpec/LetSetup:
   Description: 'Checks for `let!` being used for test setup.'
 
+RSpec/LeadingSubject:
+  Description: 'Checks for `subject` definitions that come after `let` defintions.'
```
andyw8: typo (defintions -> definitions)

backus: Thanks will fix

PR #169, opened 2016-08-08
Add MessageExpectation cop

### LINKED ISSUE #23: Prefer spies over mocks
I think a cop to verify that spies are used instead of stubs would be useful.

See http://robots.thoughtbot.com/spy-vs-spy

I'll try writing a PR for this if you agree with the principle.

backus:
Enforces preferred styles of either

    expect(foo).to receive(:bar)

or

    allow(foo).to receive(:bar)

Fixes #23


andyw8:
I'm not sure about this practicality of this one. Won't the vast majority of specs in any project need to use `expect`?

See https://relishapp.com/rspec/rspec-mocks/docs/basics/spies for some examples of the situations I had in mind for #23.

backus:
@andyw8 I've run this on two projects now and it actually seems very effective. Basically this should be one piece of the puzzle in enforcing one of these two styles:

```ruby
allow(foo).to receive(:bar)

# exercise system under test

expect(foo).to have_received(:bar)
```

```ruby
expect(foo).to receive(:bar)

# exercise system under test
```

Note that this cop will never flag `have_received`. It simply says either "add spies before exercising your code" or "add a message expectation before exercising your code."

The second piece of the puzzle (which I haven't implemented yet) would be to find tests where spies are setup but not verified within that test. Basically it could detect (and autocorrect) the following:

```ruby
allow(foo).to receive(:bar)
^^^^^^^^^^ Message expectation specified but not verified

thing.do_something(foo)
```

which could be autocorrected to

```ruby
allow(foo).to receive(:bar)

thing.do_something(foo)

expect(foo).to have_received(:bar)
```

PR #170, opened 2016-08-08
Fix config inconsistencies

backus:


backus:
@andyw8 this should fix some inconsistencies/typos in the config file that you pointed out. Any thoughts?

PR #171, opened 2016-08-08
HookArgument cop

backus:
This is extracted from #104. I'm just rebasing, improving the tests a bit using the new matcher, and I'm going to change this to support `EnforcedStyle` and `NodePattern`


lib/rubocop/cop/rspec/hook_argument.rb
```diff
@@ -0,0 +1,112 @@
+# encoding: utf-8
```
andyw8: Can be removed.

lib/rubocop/cop/rspec/hook_argument.rb
```diff
@@ -0,0 +1,112 @@
...
+# frozen_string_literal: true
+
+module RuboCop
+  module Cop
+    module RSpec
+      # Omit the default :each argument for RSpec hooks.
+      #
+      # @example when configuration is `EnforcedStyle: :implicit`
```
andyw8: `:implicit` shouldn't be a symbol here, since it's illustrating the config.yml?

backus: Correct

lib/rubocop/cop/rspec/hook_argument.rb
backus: Thanks.

Issue #173, opened 2016-08-09
Deduplicate documentation


backus:
Right now every cop has a yarddoc description as well as a different description in the config/default.yml file. I'd like to unify these by enforcing a short YARD summary for all cops and then adding a build step which extracts this documentation and writes it to the `config/default.yml` file.

PR #174, opened 2016-08-09
Deduplicate documentation

backus:


backus:
@andyw8 could you review these changes? Details in #173

Rakefile
```diff
@@ -27,4 +30,19 @@ task :internal_investigation do
   abort('RuboCop failed!') unless result.zero?
 end
 
-task default: [:spec, :internal_investigation]
+desc 'Build config/default.yml'
+task :build_config do
+  sh('bin/build_config')
+end
+
+desc 'Confirm config matches is unchanged'
```
andyw8: "are unchanged"?

bin/build_config
```diff
@@ -0,0 +1,18 @@
+#!/usr/bin/env ruby
+
+$LOAD_PATH.unshift(__dir__ + '/../lib')
```
andyw8: Should this use `File.join` for Windows compatibility?

andyw8:
(Incomplete review, will look again later today).

bin/build_config
backus: Good idea

Rakefile
```diff
@@ -27,4 +30,19 @@ task :internal_investigation do
...
+task confirm_config: :build_config do
+  _, stdout, _, process =
+    Open3.popen3('git diff --exit-code config/default.yml')
+
+  if ENV.key?('CI') && !process.value.success?
```
andyw8: Any reason to limit this to CI? Might be helpful to have it available locally?

config/default.yml
```diff
@@ -46,82 +45,82 @@ RSpec/ExampleWording:
...
-  Description: 'Checks the file and folder naming of the spec file.'
+  Description: Checks that spec file paths are consistent with the test subject
   Enabled: true
   CustomTransform:
     RuboCop: rubocop
     RSpec: rspec
 
 RSpec/VerifiedDoubles:
-  Description: 'Prefer using verifying doubles over normal doubles.'
+  Description: Prefer using verifying doubles over normal doubles.
```
andyw8: Looks like some of these end with periods, others don't. Worth unifiying?

lib/rubocop/cop/rspec/let_setup.rb
```diff
@@ -3,7 +3,7 @@
 module RuboCop
   module Cop
     module RSpec
-      # Detect unreferenced `let!` calls being used for test setup
+      # Checks for `let!` being used for test setup.
```
andyw8: I think the previous message was better. `let!` is ok for test setup as long as the name is referenced.

config/default.yml
backus: Sure I was actually going to and got bogged down by other changes in this PR which I ended up abandoning (tl;dr splitting up config like rubocop does into `config/enabled.yml` and `config/default.yml` is impossible with the current monkey patching we do)

Rakefile
backus: Hm well I just didn't want it to fail the rake task unless we were on CI. It still dumps the new config locally so you would see it while staging your code

Rakefile
backus: Good point I just ended up rewording it

Rakefile
backus: Alright I just made it fail everywhere if it is out of sync

lib/rubocop/cop/rspec/let_setup.rb
backus: Updated. FYI the main thing I was going for was trying to standardize messages using "Check" or "checks" when possible. You are right though

config/default.yml
backus: Unified now

backus:
@andyw8 ready for another pass

backus:
@andyw8 any further comments?

andyw8:
All looks good to me.

Issue #185, opened 2016-08-23
Release 1.7


backus:
@andyw8 @nijikon @geniou Any objections before I release 1.7? I ask since I've added a lot of new functionality since 1.6: https://github.com/nevir/rubocop-rspec/blob/master/CHANGELOG.md

I'll aim to release on Wednesday otherwise

geniou:
Fine for me. Thanks for the big effort.

backus:
Released!

PR #187, opened 2016-08-23
Add MessageChain cop disallowing stubbed chains

### LINKED ISSUE #186: Suggestion: Cop that disallows stubbing message chains
As [the RSpec documentation](https://relishapp.com/rspec/rspec-mocks/v/3-5/docs/working-with-legacy-code) says (with my emphasis):

> RSpec provides a few features that, while not generally recommended, can be useful when you are getting legacy code under test (or in similar situations). *Usage of these features should be considered a code smell*.
>
> 1. Any Instance
> 2. Message Chains

While rubocop-rspec already includes a cop checking against `allow_any_instance_of` etc., there is no cop checking against `stub_chain` and `receive_message_chain`.

bquorning:
According to [the RSpec documentation](https://relishapp.com/rspec/rspec-mocks/v/3-5/docs/working-with-legacy-code), stubbing chains of method “…should be considered a code smell.”

Fixes #186.


lib/rubocop/cop/rspec/message_chain.rb
```diff
@@ -0,0 +1,35 @@
...
+      #
+      # Prefer instance doubles over stubbing any instance of a class
+      #
+      # @example
+      #   # bad
+      #   allow(foo).to receive_message_chain(:bar, :baz).and_return(42)
+      #
+      #   # better
+      #   thing = Thing.new(baz: 42)
+      #   allow(foo).to receive(bar: thing)
```
bquorning: It’s hard to suggest a “better” solution, since it will often include refactoring the code under test to avoid Demeter violations.

backus: Sometimes I'll give an example like this:

```
# bad
...

# okay
... (intermediary step)

# good
... (ideal outcome)
```

where the "okay" step is something minimally better (here that might be `allow(foo).to receive(bar).and_return(instance_double(Thing, baz: 42))`) while "good" might be an example that actually shows an improved design under test.

WDYT?

lib/rubocop/cop/rspec/message_chain.rb
```diff
@@ -0,0 +1,35 @@
...
+      #
+      class MessageChain < Cop
+        include RuboCop::RSpec::SpecOnly
+
+        MESSAGE = 'Avoid stubbing using `%{method}`'.freeze
```
backus: Small point but I like using `%<method>s` instead of `%{method}` because it codifies the type of the variable that should be interpolated.

backus:
@bquorning great idea! I left two small notes. After those are resolved I'm happy to merge this and include it in 1.7 (see #185)

lib/rubocop/cop/rspec/message_chain.rb
bquorning: I don’t know what I’d put under the “good” section. I can however rename “better” to“okay” :-)

bquorning: It looks like the existing cops are using 

* bad
* good
* also good
* better
* allowed

So I’d prefer to leave my example as “better”.

lib/rubocop/cop/rspec/message_chain.rb
bquorning: Fixed.

There should be a cop for that ;-)

andyw8:
Nice complement to the AnyInstance cop :+1:

lib/rubocop/cop/rspec/message_chain.rb
```diff
@@ -0,0 +1,35 @@
+module RuboCop
+  module Cop
+    module RSpec
+      # Check that chains of messages are not being stubbed.
+      #
+      # Prefer instance doubles over stubbing any instance of a class
```
andyw8: Accidentally leftover from copy and paste?

bquorning: Yes, thanks for pointing that out. I’ve removed it now.

lib/rubocop/cop/rspec/message_chain.rb
backus: alrighty

lib/rubocop/cop/rspec/message_chain.rb
backus: Yeah that would be nice

spec/rubocop/cop/rspec/message_chain_spec.rb
```diff
@@ -0,0 +1,21 @@
...
+                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ Avoid stubbing using `receive_message_chain`
+      end
+    RUBY
+  end
+
+  it 'finds old `should_receive` syntax' do
+    expect_violation(<<-RUBY)
+      before do
+        foo.stub_chain(:one, :two).and_return(:three)
+        ^^^^^^^^^^^^^^^^^^^^^^^^^^ Avoid stubbing using `stub_chain`
```
backus: @bquorning maybe this should *only* highlight `stub_chain(:one, :two)`?

lib/rubocop/cop/rspec/message_chain.rb
backus: I'll try to make it happen 😉 https://github.com/bbatsov/rubocop/issues/3438

spec/rubocop/cop/rspec/message_chain_spec.rb
bquorning: I can call `add_offense` with `:selector` instead of `:expression` and get highlighting on only `stub_chain` and `receive_message_chain` – without the arguments.

bquorning: Fixed.

We should consider also using `:selector` when adding offense in the AnyInstance cop.

backus: Yup I think in general these should be scrutinized I just haven't done a full review

bquorning: I’ll squash the 2 commits into 1, then this branch should be ready.

backus: Great

spec/rubocop/cop/rspec/message_chain_spec.rb
```diff
@@ -0,0 +1,21 @@
...
+  it 'finds `receive_message_chain`' do
+    expect_violation(<<-RUBY)
+      before do
+        allow(foo).to receive_message_chain(:one, :two) { :three }
+                      ^^^^^^^^^^^^^^^^^^^^^ Avoid stubbing using `receive_message_chain`
+      end
+    RUBY
+  end
+
+  it 'finds old `should_receive` syntax' do
```
backus: Sorry @bquorning just noticed that this should probably read `stub_chain` not `should_receive`. I'll merge after this is fixed

bquorning: Fixed. Thanks for spotting that one.

bquorning:
Removed `RSpec/` from the changelog. That was the last change, I swear.

backus:
Thanks @bquorning!

PR #190, opened 2016-08-27
disable MultipleExpectations where multiple expectations are expected

edslocomb:


backus:
@edslocomb I think these belong in the projects configuration files and not in rubocop-rspec's. For example, my company uses `spec/integration`. We also aren't on rails and we don't use cucumber. I respect what you are aiming for here but I don't think it makes sense to try to disable this globally for all users. Instead I just encourage users to maintain their own rubocop config so that it suits their needs.

edslocomb:
Alrighty.

Sounds like a rubocop-rspec_rails fork might be in order, but I'm certainly not going to bell that cat :)

backus:
Haha I don't think a fork makes sense for that but a richer config for rails users would be a fine little config file gem. I hope you understand why I personally don't want to add that to this gem. In all seriousness you could make that gem and make it just be a config file with these excludes. I believe that is the point of `inherit_gem`

PR #196, opened 2016-09-01
Optionally ignore methods

bquorning:
Personally I have never named a spec file after the name of the method it’s testing. Perhaps I’m a minority?

With this PR, I can configure `RSpec/FilePath` with `'IgnoreMethods' => true` and it will only look at the described class/module name when recommending a path for this spec file.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Master (unreleased)
 
+* Optinally ignore method names in the `describe` argument when running the `FilePath` cop. ([@bquorning][])
+
 ## 1.7.0 (2016-08-24)
 
 * Add support for checking all example groups with `ExampleLength`. ([@backus][])
--- a/config/default.yml
+++ b/config/default.yml
@@ -90,6 +90,7 @@ RSpec/FilePath:
   CustomTransform:
     RuboCop: rubocop
     RSpec: rspec
+  IgnoreMethods: false
 
 RSpec/VerifiedDoubles:
   Description: Prefer using verifying doubles over normal doubles.
--- a/lib/rubocop/cop/rspec/file_path.rb
+++ b/lib/rubocop/cop/rspec/file_path.rb
@@ -20,7 +20,6 @@ module RuboCop
         include RuboCop::RSpec::SpecOnly, RuboCop::RSpec::TopLevelDescribe
 
         MESSAGE = 'Spec path should end with `%s`'.freeze
-        METHOD_STRING_MATCHER = /^[\#\.].+/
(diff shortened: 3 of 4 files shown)


backus:
This looks great @bquorning thanks for doing this. Can you add documentation for this config option in the class documentation? Sort of like this cop's docs: https://github.com/nevir/rubocop-rspec/blob/master/lib/rubocop/cop/rspec/hook_argument.rb#L13-L59

bquorning:
> Can you add documentation for this config option in the class documentation?

Sure, I’ll do that.

backus:
> Personally I have never named a spec file after the name of the method it’s testing.

By the way, while I don't do this all of the time, I would recommend it. When you limit yourself to one describe per file (and you try to only test through one public method per describe) I think having per-method spec files helps reenforce some pain you should feel when your public API for an object is getting fat

bquorning:
> When you limit yourself to one describe per file (and you try to only test through one public method per describe) I think having per-method spec files helps reenforce some pain you should feel when your public API for an object is getting fat

Valid point. But the pain of managing a single, 300-line spec file also tells me that my public API is too large :-) If I have a class with a small public API – 3 methods, say – wouldn’t it be preferable to test them in one file instead of in three different files?

bquorning:
By the way, I have fixed the documentation in the class.

One thing I forgot to mention: While `'IgnoreMethods' => true` *allows* naming the spec file only after the class, it doesn’t *enforce* it. This is because of the `*` in the generated regex.

backus:
> If I have a class with a small public API – 3 methods, say – wouldn’t it be preferable to test them in one file instead of in three different files?

Yeah this is why I don't require myself to follow a hard rule for this. Basically, if I'm writing more than two or three tests for a method, I'll consider moving it to a file. Otherwise small class can stay in one file

backus:
Thanks @bquorning. I've rebased this branch on top of master and merged them

backus:
Rebase and merge was in #202. I did the rebase myself since we're in different time zones and it is easier this way to manage the this way. Let me know if you're interested in joining as a collaborator in the future. That way you could just push to a branch on this repo and I could rebase directly

bquorning:
> Rebase and merge was in #202. I did the rebase myself

No problem – and thank you.

> Let me know if you're interested in joining as a collaborator in the future. 

Sure.

bquorning:
> I did the rebase myself since we're in different time zones and it is easier this way to manage the this way.

~~Just~~ Almost in time: https://github.com/blog/2247-improving-collaboration-with-forks

backus:
haha

PR #200, opened 2016-09-03
Add ImplicitExpect cop

### LINKED ISSUE #194: Enforce either `it { should ... }` OR `it { is_expected.to }`

backus:
Closes #194


backus:
@andyw8 and @bquorning could you review this cop?

bquorning:
As I understand it (from reading http://rspec.info/blog/2014/05/notable-changes-in-rspec-3/#new-api-for-one-liners-isexpected), setting `config.syntax = :expect` has no influence on the support for `it { should ... }` or `it { is_expected.to ... }` – since this particular `should` is not a monkey patch on `Object`. Is that correct?

config/default.yml
```diff
@@ -113,6 +113,14 @@ RSpec/ExampleLength:
   Enabled: true
   Max: 5
 
+RSpec/ImplicitExpect:
+  Description: Check that a consist implict expectation style is used.
```
andyw8: "consistent implicit"

lib/rubocop/cop/rspec/implicit_expect.rb
```diff
@@ -0,0 +1,89 @@
+# frozen_string_literal: true
+
+module RuboCop
+  module Cop
+    module RSpec
+      # Check that a consist implict expectation style is used.
```
andyw8: "consistent implicit"

backus: Good eye!

config/default.yml
backus: Thanks! For future reference you only need to flag something in the description on the class. The build enforces that this description is the same

backus:
@bquorning correct!

spec/rubocop/cop/rspec/implicit_expect_spec.rb
```diff
@@ -0,0 +1,94 @@
...
+    end
+
+    it 'flags it { is_expected.to }' do
+      expect_violation(<<-RUBY)
+        it { is_expected.to be_truthy }
+             ^^^^^^^^^^^^^^ Prefer `should` over `is_expected.to`.
+      RUBY
+    end
+
+    it 'flags it { is_expected.to_not }' do
```
bquorning: Perhaps also add an example for `it { is_expected.not_to }`?

backus: Sure

bquorning:
:+1:

PR #205, opened 2016-09-05
Use buildkite elastic CI + Docker for faster builds

backus:
This PR adds

1. A docker container for running builds in an isolated context with different
   ruby versions

2. A set of bash scripts for setting up this docker container and for easily
   running rake tasks inside of the container

3. Initial configuration for the build matrix docker container which mirror
   the ruby versions currently tested on Travis

4. Buildkite configuration (a pipeline definition and pre-command hook) that
   replicates our build on buildkite

Using this setup should make it much quicker to test builds of rubocop-rspec.
I currently have a Buildkite's [elastic-ci](https://github.com/buildkite/elastic-ci-stack-for-aws) hooked up to a personal AWS account of mine. This setup
has a single EC2 instance always online waiting for builds. If some build activity
picks up then the elastic CI configuration will spin up as many as 10 instances
in order to satisfy the demand for build instances.

This buildkite setup is *fast*! Take a look at the difference:

![travis][travis-build]
![buildkite][buildkite-build]

[travis-build]: https://cloud.githubusercontent.com/assets/2085622/18332207/8fbbac96-7519-11e6-986e-86384902e632.png
[buildkite-build]: https://cloud.githubusercontent.com/assets/2085622/18332231/c89bca96-7519-11e6-8570-6bb2a798872c.png

If it isn't obvious from the picture:

* Travis CI: 3 minutes 15 seconds
* Buildkite: 23 seconds

So roughly **8x faster** than Travis.

Travis runs one ruby build at a time per ruby version. Buildkite can afford to
run each step in parallel. Buildkite also doesn't need to download and install ruby for every build which saves a lot of time. This means I can scale up the supported rubies (or
the independent build steps) and incur almost not additional cost in build time.

---

That was the motivation for this PR. Here is what makes it tough:

By default each build step runs directly on the AWS instance. Therefore, if this
buildkite setup is open to outside contributors (which I would like it to be), someone
could make a PR like this

```ruby
# anywhere in the codebase that is exercised by tests
system('rm -rf /')
```

or any other arbitrary command and this would affect the CI server permanently. This
is only one issue of course and there are plenty other malicious things people could
do. This is where the docker container comes in. Each build step is run like so:

```sh
$ build/run 2.2.5 spec
```

Inside the Docker container this is setting the ruby version to 2.2.5 and running
`rake spec`. The project directory is mounted inside of the docker container as a
read only filesystem (so the code running inside the container is not able to modify
the outside world) so the code executing inside cannot make changes to the CI server.

These commands are specified in the `.buildkite/pipeline.yml` file like so:

```yaml
steps:
  - command: "build/run 2.2.5 spec"
    label: ":rspec: RSpec (2.2)"
    agents:
      - queue=elastic

  - command: "build/run 2.3.1 spec"
    label: ":rspec: RSpec (2.3)"
    agents:
      - queue=elastic

  - command: "build/run 2.2.5 internal_investigation"
    label: ":rubocop: RuboCop (2.2)"
    agents:
      - queue=elastic

  - command: "build/run 2.3.1 internal_investigation"
    label: ":rubocop: RubCop (2.3)"
    agents:
      - queue=elastic

  - command: "build/run 2.3.1 confirm_config"
    label: ":scroll: Configuration"
    agents:
      - queue=elastic
```

This docker isolation is nice and all but that alone doesn't secure the build
process. Any third party collaborator could just edit the `pipeline.yml` file and
add a step that does `rm -rf /` or whatever. Same for any of the scripts or the
dockerfile itself. To work around this, the buildkite pipeline is also configured
via the dashboard (which only I can control right now) and the first build step
is this:

```sh
gpg --keyserver pgp.mit.edu --recv 1B3C10C4

git_signature=$(git --no-pager show --quiet --pretty='%G?: %GS %GK' HEAD)
expected_signature="U: John Backus <johncbackus@gmail.com> 9A91898D0B0B2FBE"

if [[ "$git_signature" != "$expected_signature" ]]; then
  echo "Expected signature: $expected_signature"
  echo "Actual signature:   $git_signature"
  echo

  git diff --exit-code origin/master -- build .buildkite
fi
```

This is doing the following:

1. Download the PGP key I use to sign commits
2. Check the signature of the current `HEAD` commit
3. If the current head commit is not signed by me then fail the build if any
   files were changed in the `build/` or `.buildkite` directories

The buildkite settings then halts the build early if a commit not issued by
me altered the build steps.

With this leading check I am confident enough that a build for a 3rd party
PR can't just arbitrarily hijack an AWS box of mine :)


build/docker-bootstrap.sh
```diff
@@ -0,0 +1,85 @@
...
+# Install all ruby versions specified in our ruby matrix file
+for ruby_version in "${ruby_versions[@]}"; do
+  install_ruby_version "$ruby_version"
+done
+
+# Create a new unprivileged user named "ci"
+adduser -DS -s /bin/bash ci
+
+# Create our separate bundle config directory which ci is allowed to write to
+mkdir "$BUNDLE_APP_CONFIG" && chown --recursive "ci:" -- "$BUNDLE_APP_CONFIG"
```
dkubb: Given that the file is being create here is it necessary to use `--recursive` like this?

build/docker-bootstrap.sh
```diff
@@ -0,0 +1,85 @@
...
+
+# Create a source directory which files are copied to so writes don't impact the host
+mkdir /src && chown --recursive "ci:" -- /src
```
dkubb: Instance of https://github.com/backus/rubocop-rspec/pull/205/files#r77936248

build/docker-exec
```diff
@@ -0,0 +1,30 @@
+#!/bin/bash
+
+if [[ $# -ne 2 ]] ; then
+    echo "build/docker-exec - run a rake command on a specified ruby version"
+    echo "Usage:"
+    echo "  build/docker-exec 2.2.5 spec"
+    exit 1
```
dkubb: You were using 2 space indent in other bash scripts but this uses 4.

build/docker-exec
```diff
@@ -0,0 +1,30 @@
...
+# Disable strict settings that break chruby
+set +o nounset
+set +o noglob
+
+source "/usr/local/share/chruby/chruby.sh"
+
+chruby "$ruby_version"
+
+# Reenable strict settings
+set -o nounset
```
dkubb: What about doing the above stuff inside parenthesis so that you don't have to remember to re-enable things again?

build/docker-bootstrap.sh
backus: True! I guess I cargo culted that in the beginning and missed it when scrutinizing

build/docker-exec
backus: Darn good catch

build/run
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash
+
+# Mount the project as a read only filesystem to /app and pass
+# along any arugments that were provided to this script
+docker run -v"$(pwd):/app:ro" backus/rubocop-rspec "$@"
```
dkubb: I'd recommend using `--volume` instead of `-v` in scripts.

I think it's fine, maybe even preferable, when you are working on shell to use the short letter options. In scripts though I think that if a longer option exists it should be preferable. While it is verbose it helps later on when you are trying to understand some function when you can more easily search for it in docs and online.

build/docker-exec
backus: So I wasn't actually sure if doing the `source` in a subshell would impact things after the parens ended but I'll check to see if that works

backus: So I wasn't actually sure if doing the `source` in a subshell would impact things after the parens ended but I'll check to see if that works

build/strict-mode.sh
```diff
@@ -0,0 +1,14 @@
...
+# http://www.davidpashley.com/articles/writing-robust-shell-scripts/
+# http://kvz.io/blog/2013/11/21/bash-best-practices/
+# http://redsymbol.net/articles/unofficial-bash-strict-mode/
+
+set -o errexit    # Exit when an expression fails
+set -o pipefail   # Exit when a command in a pipeline fails
+set -o nounset    # Exit when an undefined variable is used
+set -o noglob     # Disable shell globbing
+set -o noclobber  # Disable automatic file overwriting
+
```
dkubb: No umask? ;)

build/run
backus: I agree regarding the long flags (I tried to catch all instances where I could eliminate the opaque shorthand) so good catch! Also, I think there is a `--read-only` flag which I could use instead of doing `:ro`

build/docker-exec
```diff
@@ -0,0 +1,30 @@
...
+set -o noglob
+
+bundle install --jobs 8 --path /bundle
```
dkubb: Technically the `install` keyword is unnecessary since it's the default when you run `bundle`.

build/docker-exec
```diff
@@ -0,0 +1,30 @@
...
+rake_task="$2"
+
+cp -r /app /src
+cd /src/app || exit 1
+
+# Disable strict settings that break chruby
+set +o nounset
+set +o noglob
+
+source "/usr/local/share/chruby/chruby.sh"
```
dkubb: I would tend to use single quotes on any strings where I am not doing variable interpolation.

build/docker-exec
backus: Oh neat I'll drop it

build/docker-exec
```diff
@@ -0,0 +1,30 @@
+#!/bin/bash
+
+if [[ $# -ne 2 ]] ; then
+    echo "build/docker-exec - run a rake command on a specified ruby version"
+    echo "Usage:"
+    echo "  build/docker-exec 2.2.5 spec"
```
dkubb: These lines are instances of https://github.com/backus/rubocop-rspec/pull/205/files#r77936617

build/docker-exec
dkubb: Actually are any quotes even needed here at all?

build/strict-mode.sh
backus: Hehe no posix either. I'm open to it later I think but I didn't want to sink the extra time into whitelisting permissions yet.

build/docker-bootstrap.sh
dkubb: I've since begun to only use `--` when my positional arguments are defined by variables. This prevents them from unintentionally being confused with switches or other arguments. When the position argument is static this can't really happen so I don't bother using `--` in those cases.

build/docker-bootstrap.sh
```diff
@@ -0,0 +1,85 @@
...
+# Use ruby-install to install a ruby version
+install_ruby_version(){
+  local ruby_version="$1"
+
+  # Execute inside of a subshell so that environment changes for ruby-install
+  # don't have a global impact
+  (
+    unset_strict;
+    # Skipping doc installation saves space and time during `docker build`
+    ruby-install --jobs=8 ruby "$ruby_version" -- --disable-install-doc --disable-install-rdoc --disable-install-capi;
```
dkubb: Instance of https://github.com/backus/rubocop-rspec/pull/205/files#r77936748

dkubb: Is the `=` necessary between `--job` and `8`? If not I would probably drop it.

build/docker-exec
backus: They're probably not good catch

build/docker-bootstrap.sh
backus: I think in this case the `--` delineates what flags are for `ruby-install` and what flags are being passed down to the ruby compiler

backus: @dkubb yeah it is for some reason. Otherwise it fails to parse

build/docker-bootstrap.sh
dkubb: I would probably change `"ci:"` to `'ci:'`. I assume it's necessary to quote it? If not then just drop the quotes.

build/docker-bootstrap.sh
dkubb: Instance of https://github.com/backus/rubocop-rspec/pull/205/files#r77937086

build/docker-exec
```diff
@@ -0,0 +1,30 @@
...
+    exit 1
+fi
+
+ruby_version="$1"
+rake_task="$2"
+
+cp -r /app /src
+cd /src/app || exit 1
```
dkubb: What is the reason for copying the files over? Is it because one of your later commands is trying to write to the path?

build/docker-exec
backus: Yeah it doesn't work if you subshell:

```
$ git --no-pager diff
diff --git i/build/docker-exec w/build/docker-exec
index 7ea461d..45e8afc 100755
--- i/build/docker-exec
+++ w/build/docker-exec
@@ -13,17 +13,15 @@ rake_task="$2"
 cp -r /app /src
 cd /src/app || exit 1

-# Disable strict settings that break chruby
-set +o nounset
-set +o noglob
+(
+  # Disable strict settings that break chruby
+  set +o nounset
+  set +o noglob

-source "/usr/local/share/chruby/chruby.sh"
+  source "/usr/local/share/chruby/chruby.sh"

-chruby "$ruby_version"
-
-# Reenable strict settings
-set -o nounset
-set -o noglob
+  chruby "$ruby_version"
+)

 bundle install --jobs 8 --path /bundle

$ build/run 2.2.5 spec
build/docker-exec: line 26: bundle: command not found
```

build/Dockerfile
```diff
@@ -0,0 +1,61 @@
...
+ADD $RUBY_INSTALL_URL "$RUBY_MANAGERS_DIR"
+
+# Download chruby
+ADD $CHRUBY_URL "$RUBY_MANAGERS_DIR"
+
+RUN docker-bootstrap.sh
+
+USER ci
+
+WORKDIR /app
```
dkubb: Maybe create the `/src/app` path here and set that to the WORKDIR? You don't really do any work that is required for you to be in `/app`, you can just copy `/app/*` to `.` in the docker-exec script.

build/docker-bootstrap.sh
dkubb: Oh you're right. I noticed `--` used in another case below and I think it's what triggered me.

build/docker-exec
backus: Yeah. For now I didn't want to deal with the back pressure of making the commands not write to the system.

build/Dockerfile
backus: Good point

build/docker-bootstrap.sh
backus: Good idea

build/docker-exec
dkubb: Really? It doesn't work. I am surprised. I know other `set` options only persist for the duration of the subshell in which they were set.

backus: Well I think the subshell isolates the environment variables (so setting them inside the subshell doesn't affect the parent) and all chruby is doing is setting a handful of environment variables

build/Dockerfile
backus: So the `app/*` doesn't work quite right since it does not copy hidden files.

backus: I can do `cp -r /app ..` though

build/run
backus: `--read-only` does not work! That makes the entire FS read only compared to just the volume.

backus:
@dkubb flags addressed if you'd like to do another pass

backus:
@lukasmartinelli any chance you could help review this PR? This is the work that led me to hadolint :smile:

backus:
@andyw8 and @bquorning could you review also? It is fine if you aren't super familiar with bash / docker / buildkite. tl;dr is that buildkite is much faster and should allow us to add other checks to CI like jruby and mutation testing. One downside here is that buildkite doesn't yet have publicly viewable build pages so I'm going to keep travis on as a build step that 3rd party contributors can view if their build fails. Travis shouldn't be required for the build to be merged though.

bquorning:
The 2nd commit is named `Fixup! ...` and was probably intended to have been sqashed into the 1st commit :-)

.buildkite/pipeline.yml
```diff
@@ -0,0 +1,25 @@
...
+    agents:
+      - queue=elastic
+
+  - command: "build/run 2.2.5 internal_investigation"
+    label: ":rubocop: RuboCop (2.2)"
+    agents:
+      - queue=elastic
+
+  - command: "build/run 2.3.1 internal_investigation"
+    label: ":rubocop: RubCop (2.3)"
```
bquorning: Typo: `s/RubCop/RuboCop/`

backus:
@bquorning correct! You caught me fixing a tiny blemish

.buildkite/pipeline.yml
backus: lol... Thank you

build/Dockerfile
```diff
@@ -0,0 +1,65 @@
...
+ADD $RUBY_INSTALL_URL "$RUBY_MANAGERS_DIR"
+
+# Download chruby
+ADD $CHRUBY_URL "$RUBY_MANAGERS_DIR"
+
+RUN mkdir -p /src/app
+
+WORKDIR /src/app
+
+COPY build/docker-exec build/docker-exec
```
bquorning: I’m wondering what this line does. The source and destination are identical.

backus: Copies `build/docker-exec` from the host environment to `/src/app/build/docker-exec` inside of the container

build/Dockerfile
```diff
@@ -0,0 +1,65 @@
...
+  # Create ruby managers directory for downloads
+  && mkdir "$RUBY_MANAGERS_DIR"
+
+# Download ruby-install
+ADD $RUBY_INSTALL_URL "$RUBY_MANAGERS_DIR"
+
+# Download chruby
+ADD $CHRUBY_URL "$RUBY_MANAGERS_DIR"
+
+RUN mkdir -p /src/app
```
lukasmartinelli: For what do you need the `/src/app` directory? Nothing is added at build time - is a volume mounted there later on? Then perhaps add a `VOLUME` instruction to mark it.

backus: Yeah the volume is mounted in `build/app`. The `VOLUME` instruction seems like a good idea I need to read up on that

bquorning:
:+1:

build/strict-mode.sh
```diff
@@ -0,0 +1,14 @@
+# Taken from dkubb/dockerfiles: https://git.io/viGZZ
```
lukasmartinelli: Add a Shebang? perhaps unnecessary here..

backus: Will do

build/run
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash
+
+# Mount the project as a read only filesystem to /app and pass
+# along any arugments that were provided to this script
+docker run --volume "$(pwd):/app:ro" backus/rubocop-rspec "$@"
```
lukasmartinelli: Do you want the container to exist after running it?
Otherwise specify `docker run --rm --volume "$(pwd):/app:ro" backus/rubocop-rspec "$@"`

backus: Oh interesting. I guess I don't.

backus:
Also @bquorning and @andyw8 buildkite is free for OSS so I can send you an invite so you can checkout the dashboard

build/Dockerfile
backus: Ah @lukasmartinelli so I mount the current build to `/app` then I copy it over to `/src/app` so I'm making the volume `/app`. Seem fine?

lukasmartinelli: Get it now. yep makes sense 👍

backus:
@lukasmartinelli I addressed your flags. Thanks for reviewing :smile:

backus:
@bquorning did you get your buildkite invite?

backus:
@dkubb I'm going to merge this for now so that I can use the new CI setup for other PRs. I'm still waiting on getting review feedback here from the buildkite team but I think this is safe enough to use for the next few days if they don't have time to get to it immediately. If you have any other feedback I'll happily address it in a followup PR.

build/Dockerfile
```diff
@@ -0,0 +1,67 @@
...
+
+WORKDIR /src/app
+
+VOLUME /app
```
dkubb: Given that this is very unlikely to change maybe it should move above the commands with version numbers, near the top of the dockerfile?

I generally list my commands from least to most likely to change so I can re-use as many layers from previous builds as possible.

build/docker-bootstrap.sh
```diff
@@ -0,0 +1,85 @@
...
+# Install all ruby versions specified in our ruby matrix file
+for ruby_version in "${ruby_versions[@]}"; do
+  install_ruby_version "$ruby_version"
+done
+
+# Create a new unprivileged user named "ci"
+adduser -DS -s /bin/bash ci
+
+# Create our separate bundle config directory which ci is allowed to write to
+mkdir "$BUNDLE_APP_CONFIG" && chown 'ci:' "$BUNDLE_APP_CONFIG"
```
dkubb: Are the single quotes in `'ci:'` needed?

build/Dockerfile
backus: Good idea

build/docker-bootstrap.sh
backus: Hm probably not

build/docker-exec
```diff
@@ -0,0 +1,29 @@
...
+  echo 'build/docker-exec - run a rake command on a specified ruby version'
+  echo 'Usage:'
+  echo '  build/docker-exec 2.2.5 spec'
+  exit 1
+fi
+
+ruby_version="$1"
+rake_task="$2"
+
+cp -r /app ..
```
dkubb: Is the reason you copy things into your current directory because of how `bundler` creates the `.bundle` directory? If so I wonder if there's a way to turn that off.

backus: I also want the rake task to write be able to write a file and I make the volume read only

dkubb: Actually, what if you run bundler like this:

```bash
BUNDLE_APP_CONFIG=/tmp/.bundle/config bundle
```

It appears to be [able to control](https://github.com/bundler/bundler/blob/master/lib/bundler.rb#L180-L186) where the `.bundle` directory is being written.

PR #211, opened 2016-09-14
Alphabetize config/default.yml

tsigo:
This eliminates any potential confusion about where to add configuration for
new cops, and makes it easier for a project with a custom `.rubocop.yml`
containing rubocop-rspec configuration to compare its own configuration against
the default.


backus:
Thanks! I should add a test for this. We already assert many other things are properly alphabetized

backus:
Thanks!

PR #214, opened 2016-09-21
Use Config#[] instead of removed Config#fetch

bquorning:
In RuboCop v0.43.0, the `Config` class no longer inherits from `Hash`, and thus no longer implements all of `Hash`'s API.

### DIFF
--- a/spec/project/default_config_spec.rb
+++ b/spec/project/default_config_spec.rb
@@ -20,7 +20,7 @@ describe 'config/default.yml' do
 
   def cop_configuration(config_key)
     cop_names.map do |cop_name|
-      cop_config = default_config.fetch(cop_name)
+      cop_config = default_config[cop_name]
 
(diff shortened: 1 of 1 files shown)


backus:
Nice. I thought this was a good change you made upstream

bquorning:
Thank you.

Do you know why the build is failing?

backus:
First build step is supposed to pull down my key from MIT's PGP servers but its failing for some reason (fails locally as well) and I think it might be their issue. Bizarre. I'll look into it

backus:
Yeah weird its just an issue with mit's servers I think. Updating to `gpg --keyserver keyserver.ubuntu.com --recv 1B3C10C4` works

backus:
Heh the new CI is worth it I swear 😆

backus:
Thanks @bquorning!

PR #219, opened 2016-10-05
Recommend 'be(nil)' over 'eql(nil)'

### LINKED ISSUE #217: Add `nil` to `BeEql`

bquorning:
Fixes #217.

I couldn’t find any documentation on how to use `def_node_matcher`, and I don’t know if the `nil_type?` makes sense. But the specs are passing…


backus:
Awesome! Thank you.

PR #222, opened 2016-10-09
Ignore .DS_Store

backus:
This has been bugging me :P


bquorning:
You should add it to your global (`~/.gitignore`) file instead :-)

backus:
I didn't know about that... Thanks!

PR #224, opened 2016-10-18
Add MessageSpies cop

bquorning:
Add `MessageSpies` cop for enforcing consistent style of either `expect(...).to have_received` or `expect(...).to receive`. This new cop is intended as a replacement for the `MessageExpectation` cop.

This relates to various issues:
- https://github.com/backus/rubocop-rspec/issues/23
- https://github.com/backus/rubocop-rspec/pull/169
- https://github.com/backus/rubocop-rspec/issues/218
- https://github.com/backus/rubocop-rspec/issues/229

See also discussion in https://gitter.im/rubocop-rspec/Lobby

This PR is a work in progress, and we need to discuss:
- [x] Add specs (and code?) for different cases of chaining.
- [x] Is it OK to change an existing cop, or should this be a new cop?
- [x] Changelog
- [x] Why is there so little documentation for `def_node_matcher` and `def_node_search` :trollface: 

cc/ @backus @andyw8


rubocop-rspec.gemspec
```diff
@@ -3,7 +3,7 @@
 $LOAD_PATH.unshift File.expand_path('../lib', __FILE__)
 require 'rubocop/rspec/version'
 
-Gem::Specification.new do |spec|
+Gem::Specification.new do |spec| # rubocop:disable Metrics/BlockLength
```
backus: Interesting. IMO RuboCop should just ignore gemspecs for this cop

bquorning: I agree. Fixed.

backus:
> Is it OK to change an existing cop, or should this be a new cop?

I think this should be a new cop and the old cop should be disabled by default

backus:
@bquorning I think you need to adapt the node matchers a bit to handle for different chaining cases. For example, we should probably enforce this cop for all of the following

```ruby
expect(foo).to have_received(bar)
expect(foo).to have_received(bar).with(baz)
expect(foo).to ordered
expect(foo).to have_received(bar).at_most(3).times
# etc
```

this is approximately what the node pattern and corresponding node search were doing before with

```ruby
def_node_matcher :message_expectation, <<-PATTERN
  (send $(send nil {:expect :allow} ...) :to #receive_message?)
PATTERN

def_node_search :receive_message?, '(send nil :receive ...)'
```

Otherwise these changes look good!

bquorning:
> I think this should be a new cop 

Done.

> and the old cop should be disabled by default

That should probably be a separate PR.

> I think you need to adapt the node matchers a bit to handle for different chaining cases.

Working on it.

bquorning:
Better now @backus? Will squash commits before merge.

backus:
This looks great @bquorning. Thanks for working on this. Feel free to squash and merge

bquorning:
I added a commit using different error messages for the two styles. Please let me know if I should change the text or remove the commit again.

lib/rubocop/cop/rspec/message_spies.rb
```diff
@@ -0,0 +1,75 @@
...
+      #   # good
+      #   expect(foo).to receive(:bar)
+      #
+      class MessageSpies < Cop
+        include RuboCop::RSpec::SpecOnly, ConfigurableEnforcedStyle
+
+        MSG_RECEIVE = 'Prefer `receive` for setting message ' \
+          'expectations.'.freeze
+        MSG_HAVE_RECEIVED = 'Prefer `have_received` for setting message ' \
+          'expectations. You will need to set up `%s` as a spy, using either ' \
```
backus: Small nit pick: How about "Setup `%s` as a spy using `allow` or `instance_spy`". This seems better for two reasons:

1. A bit more concise (but I don't feel that strongly about this so if you disagree then don't feel a need to change)
2. Defaults to encouraging `instance_spy` instead of `spy` (same benefit as using `instance_double` instead of `double`)

lib/rubocop/cop/rspec/message_spies.rb
```diff
@@ -0,0 +1,75 @@
...
+        def_node_matcher :message_expectation, %(
+          (send (send nil :expect (send nil $_)) :to ...)
+        )
+
+        def_node_search :receive_message, %(
+          $(send nil {:receive :have_received} ...)
+        )
+
+        def on_send(node)
+          message_expectation(node) do |receiver|
```
backus: I usually try to avoid multiple levels of block nesting like this (happy to elaborate on why if you are curious). I usually opt to create a private method which more clearly conveys what nodes I'm filtering down to. In this case I might do something like this:

```ruby
def on_send(node)
  receive_message_matcher(node) do |receiver, message_matcher|
    return correct_style_detected if preferred_style?(message_matcher)

    add_offense(message_matcher, :selector, error_message(receiver)) do
      opposite_style_detected
    end
  end
end

private

def receive_message_matcher(node)
  return unless (receiver = message_expectation(node))

  receive_message(node) { |match| yield(receiver, match) }
end
```

lib/rubocop/cop/rspec/message_spies.rb
```diff
@@ -0,0 +1,75 @@
...
+            receive_message(node) do |match|
```
backus: I would probably name `match` something like `message_matcher`. I usually try to dig deep into rspec's source to find proper naming for the parts of their API that I'm dealing with. In this case [`have_received`](https://github.com/rspec/rspec-mocks/blob/c0bf45b5ceb179c51b0d38d7c91faf60a7a6048f/lib/rspec/mocks/example_methods.rb#L281-L283) and [`received`](https://github.com/rspec/rspec-mocks/blob/c0bf45b5ceb179c51b0d38d7c91faf60a7a6048f/lib/rspec/mocks/syntax.rb#L115) both are dealing with messages and setting matchers so `message_matcher` seems good enough

bquorning:
@backus I ended up using all three of your suggestions. I think this is ready for the final review.

backus:
Feel free to merge

lucascaton:
Before:
```ruby
before { expect(account).to receive(:errors).and_return(amount_cents: [error]) }
```

After:
```ruby
before { expect(account).to have_received(:errors).and_return(amount_cents: [error]) }
```

Error:
```sh
NoMethodError: undefined method `and_return' for #<RSpec::Mocks::Matchers::HaveReceived:0x007fe9ccb409e8>
```

lucascaton:
@backus will do, thanks.

PR #226, opened 2016-10-19
Don’t stub when you’re mocking

### LINKED ISSUE #221: Cop idea: Don’t stub when you’re mocking
Sometimes, I see people defining a mock and tacking a stub onto the end it:

``` rb
expect(foo).to receive(:bar).with(42).and_return("hello world")

subject.exercise
```

Now RSpec will expect `foo` to receive `:bar` with the argument `42`.

If you read the code aloud, it very much sounds like `"hello world"` should also be part of the expectation. It isn’t.

I can think of at least two ways to rewrite this:

``` rb
allow(foo).to receive(:bar).with(42).and_return("hello world")
expect(foo).to receive(:bar).with(42)

subject.exercise
```

or

``` rb
allow(foo).to receive(:bar).with(42).and_return("hello world")

subject.exercise

expect(foo).to have_received(:bar).with(42)
```

bquorning:
Fixes #221.
### TODO
- [x] Must also work for `and_raise`, `and_yield`, `and_throw`, `and_call_original`, and `and_wrap_original`.
- [x] Changelog entry.
- [x] Rename cop?
- [x] Better cop description.

cc @backus 


spec/rubocop/cop/rspec/mock_not_stub_spec.rb
```diff
@@ -0,0 +1,37 @@
...
+    expect_violation(<<-RUBY)
+      expect(foo).to receive(:bar).with(42).and_return("hello world")
+                                                       ^^^^^^^^^^^^^ Don't stub your mock.
+    RUBY
+  end
+
+  it 'flags expect(...).to receive(...).with(...) { } ' do
+    expect_violation(<<-RUBY)
+      expect(foo).to receive(:bar).with(42) { "hello world" }
+                                              ^^^^^^^^^^^^^ Don't stub your mock.
```
bquorning: I really want to highlight the entire block `{ "hello world" }`.

spec/rubocop/cop/rspec/mock_not_stub_spec.rb
```diff
@@ -0,0 +1,37 @@
+# frozen_string_literal: true
+
+describe RuboCop::Cop::RSpec::MockNotStub do
+  subject(:cop) { described_class.new }
+
+  it 'flags expect(...).to receive(...).with(...).and_return' do
+    expect_violation(<<-RUBY)
+      expect(foo).to receive(:bar).with(42).and_return("hello world")
```
bquorning: I really want to highlight the entire call to `.and_return("hello world")`.

spec/rubocop/cop/rspec/mock_not_stub_spec.rb
backus: Do that you would want something like this:

```ruby
source_map = match.loc

source_range = Parser::Source::Range.new(
  source_map.expression.source_buffer,
  source_map.begin.begin_pos,
  source_map.end.end_pos
)

add_offense(match, source_range)

Comment (backus) [inline spec/rubocop/cop/rspec/mock_not_stub_spec.rb]: also you would have to change the node matcher like so

```diff
--- i/lib/rubocop/cop/rspec/mock_not_stub.rb
+++ w/lib/rubocop/cop/rspec/mock_not_stub.rb
@@ -27,8 +27,8 @@ module RuboCop
           (send
             (send nil :expect ...) :to
             {
-              (send #receive :and_return $_)
-              (block #receive (args) $_)
+              $(send #receive :and_return _)
+              $(block #receive (args) _)
             }
           )
         PATTERN

**Comment by backus** [comment]
@bquorning I played with the branch locally to get the highlighting you are interested in. This should work:

```diff
diff --git i/lib/rubocop/cop/rspec/mock_not_stub.rb w/lib/rubocop/cop/rspec/mock_not_stub.rb
index 286df07..aba2985 100644
--- i/lib/rubocop/cop/rspec/mock_not_stub.rb
+++ w/lib/rubocop/cop/rspec/mock_not_stub.rb
@@ -27,8 +27,8 @@ module RuboCop
           (send
             (send nil :expect ...) :to
             {
-              (send #receive :and_return $_)
-              (block #receive (args) $_)
+              $(send #receive :and_return _)
+              $(block #receive (args) _)
             }
           )
         PATTERN
@@ -42,7 +42,24 @@ module RuboCop
 
         def on_send(node)
           message_expectation_with_return_block(node) do |match|
-            add_offense(match, :expression)
+            source_map = match.loc
+
+            offending_range =
+              if match.send_type?
+                Parser::Source::Range.new(
+                  source_map.expression.source_buffer,
+                  source_map.selector.begin_pos,
+                  source_map.end.end_pos
+                )
+              else
+                Parser::Source::Range.new(
+                  source_map.expression.source_buffer,
+                  source_map.begin.begin_pos,
+                  source_map.end.end_pos
+                )
+              end
+
+            add_offense(match, offending_range)
           end
         end
       end
```

I'd refactor it a bit of course to make it not just one big method.

backus:
Nice job with the node pattern stuff btw. I wish people would use it more on rubocop proper

bquorning:
Thanks for the code snippet there, @backus. I think I’ll use `source_map.dot.begin_pos` instead of `source_map.selector.begin_pos` and get highlight on the `.` preceding `and_return` as well.

config/default.yml
```diff
@@ -103,6 +103,10 @@ RSpec/MessageExpectation:
   - allow
   - expect
 
+RSpec/MockNotStub:
+  Description: Checks that stubs aren't added to mocks.
```
bquorning: I’d like some feedback on the cop’s name and description.

backus:
Should this cop be configurable? It seems like `expect(...).to receive(...).and_return(...)` should be a valid style.

bquorning:
Configurable? You can just disable it. Perhaps I misunderstand you?

backus:
No I guess I was curious if there was an alternate style that this sort of thing could also enforce if we're just saying "do not use this part of the RSpec API."

I look forward to hearing outside input for @jcoyne. I'm wondering how users will respond to this cop.

backus:
Should this cop also reject things like `expect(...).to receive(...).and_raise(...)`, `expect(...).to receive(...).and_yield(...)`, `expect(...).to receive(...).and_throw(...)`?

bquorning:
> Should this cop also reject things like …

Yes, good point. All three cases should be policed by this cop.

bquorning:
Also `and_call_original`, `and_wrap_original`, `yield_with_args`…

jcoyne:
Can you say why this is good style? What's the justification?

bquorning:
The argumentation / justification is in #221.

jcoyne:
It sounds like you want to change the rspec-mocks API via rubocop. Have you considered this change as a deprecation in rspec-mocks?

bquorning:
> Have you considered this change as a deprecation in rspec-mocks?

Yes :-) I thought it would be easier to start here.

I’ve also considered mentioning someone from rspec-mocks in this PR.

jcoyne:
@bquorning I think that would be a good idea.  If you want to influence the usage of a library, it makes sense to consult the authors thereof.

bquorning:
After a brief discussion in rspec/rspec-mocks#1123 I’m convinced that rspec-mocks will not change its DSL. Which makes this a personal style preference – perhaps suitable for RuboCop-RSpec?

backus:
> > @JonRowe would you consider deprecating this behavior?
>
> Maybe if there was a broad community consensus. But users like to use different parts of RSpec in different ways and I would expect a lot of push back on such a deprecation. In general, we don't deprecate or disallow things just because something is a code smell.

I think seeing myron call this a code smell is more than enough to convince me that this is worth making a cop

backus:
I'll try to do another review tonight

JonRowe:
Just a personal opinion (which may or may not be shared with my other rspec team members) but this cop should be disabled by default it's very much a personal opinion that creates extra code for people, and rubocop tends to give the impression that something is "recommended best practise" which this isn't.

backus:
@JonRowe thanks I'll keep that in mind. If you have a minute, would you be able to look at the other cops and say whether you think this is true for the others? I'm curious how broad you view things as style vs. best practice. 

On an unrelated note, I'm not sure how much rubocop is "recommended best practice" anymore given that most cops with more than one possible style tend to be configurable in different ways.

JonRowe:
> On an unrelated note, I'm not sure how much rubocop is "recommended best practice" anymore given that most cops with more than one possible style tend to be configurable in different ways.

I see people using the default config a fair bit, and just accepting them as "the way"

lib/rubocop/cop/rspec/mock_not_stub.rb
```diff
@@ -0,0 +1,73 @@
...
+            $(send #receive :and_return _)
+          )
+        PATTERN
+
+        def_node_matcher :message_expectation_with_return_block, <<-PATTERN
+          (send
+            (send nil? :expect ...) :to
+            $(block #receive (args) _)
+          )
+        PATTERN
```
pirj: Wondering if those two can be joined together like:
```
          (send
            (send nil? :expect ...) :to
            ${
              (send #receive :and_return _)
              (block #receive (args) _)
            }
          )
```

lib/rubocop/cop/rspec/mock_not_stub.rb
```diff
@@ -0,0 +1,73 @@
...
+
+        def_node_matcher :receive, <<-PATTERN
+          {
+            (send nil? :receive ...)
+            (send (send nil? :receive ...) :with ...)
```
pirj: Wondering if it will work to change this to `def_node_search` so that it supports arbitrary nesting.
Due to the wide range of allowed qualifiers (e.g. `once`, `at_least`, `ordered`, ...) it's near impossible to match them with a `def_node_matcher`.

I've faced a similar problem [here](https://github.com/rubocop-hq/rubocop-rspec/pull/726/files#diff-faa46202d8b79d8bf965118100cff4d2R73), and using `def_node_search` solved it.

dgollahon:
I kind of lean with the disable-by-default crowd here. This feels a little obscure/confusing to me, though @bquorning motivating example in the linked issue seems reasonable and I would probably write the specs the same way he would. I strongly suspect we'll get very confused users asking what we mean by this cop though and why it's better to do it another way, so my intuition is that it's a little too surgical for the average user.

If this gets merged and enabled by default, the explanation/motivation should probably be documented more clearly, imo.

pirj:
> the explanation/motivation should probably be documented more clearly, imo

All for adding this to the style guide 👍 

@bquorning This would make an incredibly useful cop. I'd love to see it making to 2.0. Do you need a hand with it?

bquorning:
> Do you need a hand with it?

Oh, yes please. I haven’t looked at this PR in years.

pirj:
@bquorning @Darhazer I believe it's more or less ready for the prime time.

Let the cop name discussion begin! ⚔️ 
The [guidelines for cop names have not been set yet](https://github.com/rubocop-hq/rubocop/issues/7077).

As I understand it, cop nameы should tell to what the offence is. Basing on that my proposal is `RSpec/StubbedMock`.

pirj:
Regarding the implementation:
 - I feel ashamed for introducing a pending example in this spec. However, it seemed to me like a lot of work
 - documentation for sure can be extended, but spec for it is quite comprehensive
 - there is some redundancy in node patterns. They can be refactored, but it works quite well as is, and hope reads too

lib/rubocop/cop/rspec/mock_not_stub.rb
```diff
@@ -0,0 +1,98 @@
...
+      # Checks that message expectations are not combined with message stubs.
+      #
+      # @example
+      #
+      #   # bad
+      #   expect(foo).to receive(:bar).with(42).and_return("hello world")
+      #
+      #   # good (without spies)
+      #   allow(foo).to receive(:bar).with(42).and_return("hello world")
+      #   expect(foo).to receive(:bar).with(42)
```
Darhazer, line 15: even better `expect(foo).to have_received(:bar).with(42)`

lib/rubocop/cop/rspec/mock_not_stub.rb
```diff
@@ -0,0 +1,98 @@
...
+      #
+      class MockNotStub < Base
+        MSG = 'Do not stub your mock.'
+
+        def_node_matcher :message_expectation?, <<-PATTERN
+          {
+            (send nil? { :receive :receive_message_chain } ...)
+            (send (send nil? :receive ...) :with ...)
```
Darhazer, line 23: not sure this would work in all possible chains, like `receive.once.with()`
We need a helper for catching message expectations, see ReceiveCounts as an example of cop dealing with chains

spec/rubocop/cop/rspec/mock_not_stub_spec.rb
```diff
@@ -0,0 +1,99 @@
...
+  it 'flags stubbed message expectation' do
+    expect_offense(<<-RUBY)
+      expect(foo).to receive(:bar).and_return('hello world')
+                                  ^^^^^^^^^^^^^^^^^^^^^^^^^^ Do not stub your mock.
+    RUBY
+  end
+
+  it 'flags stubbed message expectation with a block' do
+    expect_offense(<<-RUBY)
+      expect(foo).to receive(:bar) { 'hello world' }
```
Darhazer, line 15: what if the block checks the yielded argument - quite common on exception matching (even though we do not recommend having expectation inside that block)

spec/rubocop/cop/rspec/mock_not_stub_spec.rb
```diff
@@ -0,0 +1,99 @@
...
+
+  it 'flags with order and count constraints', :pending do
+    expect_offense(<<-RUBY)
+      expect(foo).to receive(:bar) { 'hello world' }.ordered
+                                   ^^^^^^^^^^^^^^^^^ Do not stub your mock.
+      expect(foo).to receive(:bar).ordered { 'hello world' }
+                                           ^^^^^^^^^^^^^^^^^ Do not stub your mock.
+      expect(foo).to receive(:bar).with(42).ordered { 'hello world' }
+                                                    ^^^^^^^^^^^^^^^^^ Do not stub your mock.
+      expect(foo).to receive(:bar).once.with(42).ordered { 'hello world' }
```
Darhazer, line 74: what about this example, but with `and_return`. e.g. all count constraint examples are with a block

spec/rubocop/cop/rspec/mock_not_stub_spec.rb
```diff
@@ -0,0 +1,99 @@
...
+                                   ^^^^^^^ Do not stub your mock.
+      expect(foo).to receive(:bar).with(42, &canned)
+                                            ^^^^^^^ Do not stub your mock.
+      expect(foo).to receive_message_chain(:foo, :bar, &canned)
+                                                       ^^^^^^^ Do not stub your mock.
+    RUBY
+  end
+
+  it 'ignores message allowances' do
+    expect_no_offenses(<<-RUBY)
```
Darhazer, line 96: empty example :)

Darhazer:
`StubbedMock` is a great name indeed

lib/rubocop/cop/rspec/mock_not_stub.rb
pirj: The title says "without spies". Both would work with a partial double, but still `have_received` reminds of a spy-style.

lib/rubocop/cop/rspec/mock_not_stub.rb
pirj: It for sure won't. Those count and ordering constraints will add complexity that I don't really want to deal with right now. I fully acknowledge this cop is not perfect.

spec/rubocop/cop/rspec/mock_not_stub_spec.rb
pirj: To my best knowledge, `with` supports argument matchers. If it's not possible to use `with`, `and_wrap_original` should be used (only to check arguments and pass to the original).

spec/rubocop/cop/rspec/mock_not_stub_spec.rb
pirj: There's an example with everything packed in two lines below.
I doubt that it makes sense to cover all possible permutations.

This is a pending example, I let the future improvement to extend it. For now, it's just an indication, that this is not detected.

spec/rubocop/cop/rspec/mock_not_stub_spec.rb
pirj: Ooops!

lib/rubocop/cop/rspec/mock_not_stub.rb
Darhazer: Somehow missed that, perhaps I should avoid such late evening reviews :)

lib/rubocop/cop/rspec/stubbed_mock.rb
```diff
@@ -0,0 +1,98 @@
...
+      #
+      #   # bad
+      #   expect(foo).to receive(:bar).with(42).and_return("hello world")
+      #
+      #   # good (without spies)
+      #   allow(foo).to receive(:bar).with(42).and_return("hello world")
+      #   expect(foo).to receive(:bar).with(42)
+      #
+      class StubbedMock < Base
+        MSG = 'Do not stub your mock.'
```
bquorning, line 18: We need a better message here.

config/default.yml
```diff
@@ -536,6 +536,12 @@ RSpec/SingleArgumentMessageChain:
   VersionChanged: '1.10'
   StyleGuide: https://www.rubydoc.info/gems/rubocop-rspec/RuboCop/Cop/RSpec/SingleArgumentMessageChain
 
+RSpec/StubbedMock:
+  Description: Checks that message expectations are not combined with message stubs.
+  Enabled: true
```
bquorning, line 541: When I wrote the first version of this cop, there were a few comments saying that it should probably be disabled by default. What do you think @pirj? How many offenses are encountered in your real world specs collection?

pirj: I like this method more and more. Just spotted `undefined method `end_pos' for nil:NilClass` for several files there.

47162 files inspected, 7321 offenses detected. Quite a lot.
Just a couple of randomly picked ones reveal issues:

```ruby
# 24pullrequests/spec/services/popularity_scorer_spec.rb:25:42:
  let(:project) { FactoryBot.create(:project) }

    it 'returns a score of 3 if there are 5 recent issues' do
      issue = double(:issue)
      expect(project).to receive(:repo).and_return(double(updated_at: 6.months.ago))
      expect(project).to receive(:issues).and_return([issue] * 5)
      expect(popularity_scorer.score).to eq(3)
    end
```
Should be `allow`.

```ruby
    it "does not download the Cask per default" do
      allow(Cask::CaskLoader).to receive(:load).and_return(cask)
      expect(Cask::Auditor).to receive(:audit)
        .with(cask, audit_download:        false,
                    audit_appcast:         false,
                    audit_token_conflicts: false,
                    audit_new_cask:        false,
                    audit_online:          false,
                    audit_strict:          false,
                    quarantine:            true)
        .and_return(result)

      described_class.run("casktoken")
    end
```
`.and_return` is very confusing here.

lib/rubocop/cop/rspec/stubbed_mock.rb
pirj: 'Do not stub methods of the mock object.'?

config/default.yml
pirj: There are several offences in `rspec-expectations` itself.
Will you be happy to merge this as enabled by default if I'll manage to push through the pull request that fixes those offences?

lib/rubocop/cop/rspec/stubbed_mock.rb
bquorning: I think many people don’t immediately know the differences between stubs and mocks, and wouldn’t understand where we are trying to nudge.

I think the message needs to explain that the cop is “offended” when people combine `expect` with `and_return`, and that they should use `allow` instead.

pirj: What do you prefer:
 - "Do not use configure response with message expectations, use message allowances instead."
 - "Use message allowances instead of message expectations when configuring a response."
 - any other option?

Darhazer: * "Prefer `allow` over `expect` when configuring a response". 
This one may need to be dynamic (as there are different methods for expectation); or it might be "Prefer allowance over expectation when configuring responses"

lib/rubocop/cop/rspec/stubbed_mock.rb
```diff
@@ -81,7 +81,7 @@ def offending_argument_range(source_map)
           Parser::Source::Range.new(
             source_map.expression.source_buffer,
             source_map.dot.begin_pos,
-            source_map.end.end_pos
+            source_map.end&.end_pos || source_map.expression.end_pos
```
bquorning, line 84: This seems to work for all cases:
```suggestion
            source_map.expression.end_pos
```
…also for `#offending_block_range` below.

lib/rubocop/cop/rspec/stubbed_mock.rb
bquorning: > * "Prefer `allow` over `expect` when configuring a response".

I like the explicitness of that one, @Darhazer. 

I am not sure “allowance” can be used in this context. I think we need a native English speaker to help us out 😄

As for dynamic, do you mean `expect`, `is_expected`, `expect_any_instance_of` (`Language::Expectations::ALL`)? The current implementation only searches for `expect`…

lib/rubocop/cop/rspec/stubbed_mock.rb
pirj: It does, thanks!

lib/rubocop/cop/rspec/stubbed_mock.rb
pirj: Nice catch on other expectation methods. 👍 
Loving the "Prefer ... to ..." wording as well.

pirj: Wondering if should still highlight the stubbing part if we put the offence message this way:
```ruby
      is_expected.to receive(:bar).and_return(:baz)
                                  ^^^^^^^^^^^^^^^^^ Prefer `allow` to `is_expected` when configuring a response.
```
or should it become
```ruby
      is_expected.to receive(:bar).and_return(:baz)
      ^^^^^^^^^^^ Prefer `allow` to `is_expected` when configuring a response.
```

With `expect` and `expect_any_instance` it's clear. For a plain `allow`, should we advise `is_expected` as a replacement?

pirj: `is_expected`'s replacement goes directly against `SubjectStub`. Going to recommend `allow(subject)` and let `SubjectStub` deal with it.

pirj:
@bquorning @Darhazer I'm pretty happy with it now. Please take another look.

Except for the HEREDOC going over the 80 chars margin. RuboCop somehow misses that.

pirj:
Rebased.

pirj:
Squashed commits, rebased.

pirj:
Wondering is there such a CHANGELOG format that doesn't induce conflicts on all other pull requests when one gets merged?

lib/rubocop/cop/rspec/stubbed_mock.rb
```diff
@@ -0,0 +1,97 @@
...
+            (send (send nil? :receive ...) :with ...)
+          }
+        PATTERN
+
+        def_node_matcher :configured_response?, <<~PATTERN
+          { :and_return :and_raise :and_throw :and_yield
+            :and_call_original :and_wrap_original }
+        PATTERN
+
+        def self.expectation_with(matcher)
```
Darhazer, line 33: Nice :1st_place_medal: 
Up to now I'd have a matcher `expectation` that caputres `matcher` and then some code like
```
expectation(node) do |matcher|
  next unless configured_response?(matcher)
...
end
```
Now there's one more tool in my toolbox

pirj: Now when I think of it, I like your variant more.

```diff
-        def_node_matcher :expectation_with_configured_response,
-                         expectation_with(<<~PATTERN)
-                           (send #message_expectation? #configured_response? _)
-                         PATTERN
+        def_node_matcher :matcher_with_configured_response, <<~PATTERN
+          (send #message_expectation? #configured_response? _)
+        PATTERN
```
and the following becomes slightly more generic:
```
def on_send(node)
  expectation(node, &:on_expectation) # this can become a concern!
end

def on_expectation(expectation)
  matcher_with_configured_response(expectation) do |matcher, method_name|
    add_offense(matcher, message: msg(method_name))
  end

  matcher_with_return_block(expectation) do |matcher, method_name|
    add_offense(matcher, message: msg(method_name))
  end

  matcher_with_blockpass_or_hash(expectation) do |matcher, method_name|
    add_offense(matcher, message: msg(method_name))
  end
end
```

I'll give it a shot later today.

config/default.yml
```diff
@@ -543,6 +543,12 @@ RSpec/SingleArgumentMessageChain:
   VersionChanged: '1.10'
   StyleGuide: https://www.rubydoc.info/gems/rubocop-rspec/RuboCop/Cop/RSpec/SingleArgumentMessageChain
 
+RSpec/StubbedMock:
+  Description: Checks that message expectations do not have a configured response.
+  Enabled: true
```
bquorning, line 548: I guess this has to be changed, right?
```suggestion
  Enabled: pending
```

pirj: Not necessarily. I guess we'll start adhering to that policy starting from 2.0. This wasn't the case for any of the 1.x releases at least.

Darhazer: Let's try to start adhering to the policy (our minimum rubocop version requirement already supports pending, doesn't it) - users will be happier

lib/rubocop/cop/rspec/stubbed_mock.rb
pirj: Worked! I like it much better this way.

Thanks, and I think there are cases when my old approach would work, but in this specific cop `on_expectation` worked quite nice.

I also had this crazy idea https://github.com/rubocop-hq/rubocop-ast/issues/97, but it ended up being even more pleasant to look at.

config/default.yml
pirj: Agreed. At least we'll be able to see if pending works for extension cops.

pirj:
@bquorning @Darhazer Ready to take another look?

I won't have time to take over and finish #464 soon. So maybe 1.44 time after merging this, and then time to merge and refactor challenge for #956 ?

pirj:
Even though it's your own pull request, I'd love to hear your thoughts on the recent changes before merging it @bquorning

lib/rubocop/cop/rspec/stubbed_mock.rb
```diff
@@ -0,0 +1,95 @@
...
+      #   allow(foo).to receive(:bar).with(42).and_return("hello world")
+      #   expect(foo).to receive(:bar).with(42)
+      #
+      class StubbedMock < Base
+        MSG = 'Prefer `%<replacement>s` to `%<method_name>s` when ' \
+              'configuring a response.'
+
+        def_node_matcher :message_expectation?, <<-PATTERN
+          {
+            (send nil? { :receive :receive_message_chain } ...)
```
bquorning, line 23: In some cops, we also match `:stub_chain`. Should we do the same here? Or, is it soon time to remove compatibility with the deprecated RSpec syntax?

lib/rubocop/cop/rspec/stubbed_mock.rb
```diff
@@ -0,0 +1,95 @@
...
+      #
+      #   # bad
+      #   expect(foo).to receive(:bar).with(42).and_return("hello world")
+      #
+      #   # good (without spies)
+      #   allow(foo).to receive(:bar).with(42).and_return("hello world")
+      #   expect(foo).to receive(:bar).with(42)
+      #
+      class StubbedMock < Base
+        MSG = 'Prefer `%<replacement>s` to `%<method_name>s` when ' \
```
bquorning, line 18: We usually use “Prefer foo *over* bar” in the offense messages.
```suggestion
        MSG = 'Prefer `%<replacement>s` over `%<method_name>s` when ' \
```

lib/rubocop/cop/rspec/stubbed_mock.rb
```diff
@@ -0,0 +1,95 @@
...
+          }
+        PATTERN
+
+        def on_send(node)
+          expectation(node, &method(:on_expectation))
+        end
+
+        private
+
+        def on_expectation(expectation, method_name, matcher)
```
bquorning, line 62: Maybe I’m having a bad day, but I have a _really_ hard time understanding this code. It’s not immediately obvious what the node matchers actually match. The names are easy enough to understand (maybe except `matcher_with_blockpass_or_hash` which I’m not sure I get), but… Maybe we should start adding adding comments to our node matchers, with one or two examples of what the would match?

It can also be that I lack training in reading node patterns (and `parser` output). I wonder how others feel about this?

spec/rubocop/cop/rspec/stubbed_mock_spec.rb
```diff
@@ -0,0 +1,131 @@
...
+
+  it 'flags stubbed message expectation with argument matching and a block' do
+    expect_offense(<<-RUBY)
+      expect(foo).to receive(:bar).with(42) { 'hello world' }
+      ^^^^^^^^^^^ Prefer `allow` to `expect` when configuring a response.
+    RUBY
+  end
+
+  it 'ignores `have_received`' do
+    expect_no_offenses('expect(foo).to have_received(:bar)')
```
bquorning, line 35: Just for consistency (making the spec file easier to read), could you please use a heredoc in this example too?

bquorning:
> Even though it's your own pull request, I'd love to hear your thoughts on the recent changes before merging it @bquorning

Absolutely. Sorry for sometimes taking several days before responding; it feels like there are fewer hours in a day than there used to be 😕

Perhaps the spec for `pending` state can stand alone, but you can squash the other commits (just chang me to be [co-author](https://docs.github.com/en/github/committing-changes-to-your-project/creating-a-commit-with-multiple-authors)).

lib/rubocop/cop/rspec/stubbed_mock.rb
pirj: You are quite right that this cop is getting really hard to grasp. I'm inclined not to add support for a rather obscure and deprecated syntax, further complicating it.

If that wouldn't be complicating the cop, I'm all for supporting the deprecated syntax, even when RSpec 4 is out.
Users might still stick with RSpec 3, and I thought we may still support it for some period of time, just like [`rspec-rails` supports Rails 4.2](https://github.com/rubocop-hq/rubocop-rails/blob/159eef021279532312af9b7b89331bd44a7eeb9d/spec/support/shared_contexts.rb#L4).

lib/rubocop/cop/rspec/stubbed_mock.rb
pirj: Thanks for this fresh view on the cop.
It's quite common to miss the complexity when working on it, but as time goes, this complexity becomes apparent.

In addition to node matcher YARD docs that I believe now help a lot to understand [`RSpec/EmptyExampleGroup`'s node patterns](https://github.com/rubocop-hq/rubocop-rspec/blob/f1048eb8389d44d47afeff60e6db797487bbc46a/lib/rubocop/cop/rspec/empty_example_group.rb#L68), there's also [Node pattern comment feature](https://docs.rubocop.org/rubocop-ast/node_pattern.html#comments).
I intend to add both to this cop's node patterns.

lib/rubocop/cop/rspec/stubbed_mock.rb
```diff
@@ -0,0 +1,172 @@
...
+        #
+        #   @example source that matches
+        #     receive(:foo).with('bar', &canned)
+        #
+        #   @param node [RuboCop::AST::Node]
+        #   @yield [RuboCop::AST::Node] matcher
+        def_node_matcher :matcher_with_blockpass, <<~PATTERN
+          {
+            (send nil? { :receive :receive_message_chain } ... block_pass)  # receive(:foo, &canned)
+            (send (send nil? :receive ...) :with ... block_pass)            # receive(:foo).with('foo', &canned)
```
pirj, line 132: That's kind of odd that `LineLength` ignores HEREDOC

pirj, line 132: Wondering if you guys find that kind of node pattern comments helpful @bquorning @Darhazer

bquorning: I found them very useful, thank you @pirj

marcandre:
```ruby
# bad
expect(foo).to receive(:bar).with(42).and_return("hello world")
# good (without spies)
allow(foo).to receive(:bar).with(42).and_return("hello world")
expect(foo).to receive(:bar).with(42)
```

I strongly disagree that an expressive and DRY form should be broken into two repetitive forms for no benefit.

I really dislike when cops complain about *valid* code, especially when the solution is not clearly better (i.e. clearly worse for many, me included).

Of course, I fully realize that yes, we can disable it. That mere fact does not imply that the cop should be enabled by default.

I believe having cops like these, enabled by default, give RuboCop a bad name.

Note: I'm sorry, I didn't read the whole thread as it contains comments on the code as well as discussion about the usefulness of the cop itself.

bquorning:
There has been a lot of discussion for and against whether this cop should be enabled by default. I am actually not sure if we _really_ made any decision either way.

@pirj I remember you revived this PR in https://github.com/rubocop-hq/rubocop-rspec/pull/226#issuecomment-667683897, but I do not recall if we really decided to let this cop be enabled or disabled by default. I think it just … happened.

marcandre:
>  I think it just … happened.

Maybe disabling it and issuing a 2.0.1 release is the way to go then.

pirj:
@
```
~/source/rubocop $ rubocop --only RSpec/StubbedMock

Offenses:

spec/rubocop/result_cache_spec.rb:178:11: C: RSpec/StubbedMock: Prefer allow over expect when configuring a response.
          expect(team).to(
          ^^^^^^^^^^^^
spec/rubocop/result_cache_spec.rb:191:11: C: RSpec/StubbedMock: Prefer allow over expect when configuring a response.
          expect(team).to(
          ^^^^^^^^^^^^
spec/rubocop/result_cache_spec.rb:332:9: C: RSpec/StubbedMock: Prefer allow over expect when configuring a response.
        expect(FileUtils).to receive(:mkdir_p).with(start_with(cache_root))
        ^^^^^^^^^^^^^^^^^

... 3 offenses detected
```

Let's take a closer look:
```ruby
      context 'when team external_dependency_checksum changes' do
        it 'is invalid' do
          cache.save(offenses)
          expect(team).to(
            receive(:external_dependency_checksum).and_return('bar')
          )
          cache2 = described_class.new(
            file, team, options, config_store, cache_root
          )
          expect(cache2.valid?).to eq(false)
        end
      end
```

If I would be reading this, I would ask myself:
1. Why is `expect(team)` an `expect` and not `allow`, is there a risk that the SUT won't send an `external_dependency_checksum` message to `team`?
2. Why, if we have:
 - properly balanced examples
 - we check the output
we didn't limit ourselves to set the input and rather decided to increase specs' brittleness by tying ourselves to the internal implementation?

You may also notice that:
```ruby
  before do
    #...
    allow(team).to receive(:external_dependency_checksum).and_return('foo')`
```

This means we're double-stubbing `external_dependency_checksum` 😱 
We should have received RSpec's:
```
            "You're overriding a previous stub implementation of `#{@message}`. " \
            "Called from #{CallerFilter.first_non_rspec_line}."
```
warning, but we don't, not sure why exactly, probably because it's only implemented for pure doubles and doesn't work on partial doubles, or warnings are muted somehow.

The fact that this spec works doesn't mean it's good. I can lend you a hand with fixing it and turning `StubbedMock` back to `Enabled: true` if you like.

I don't have other specs on hand to reason about, but I'm all ears to know when `RSpec/StubbedMock` doesn't make sense at all.

marcandre:
> we're double-stubbing external_dependency_checksum

The first `allow` is called once to build the cache, then the method is stubbed again for the second call to check the cache validity.

> increase specs' brittleness

The spec assumes an implementation detail. If that implementation detail, the spec needs to be changed. That's why we stub when there is no better alternative. I see the spec failing as a virtue.

pirj:
Just for the record, I take back my argument that double-stubbing is an offence, it's my false belief.

PR #234, opened 2016-11-06
Add RuboCop::Cop::RSpec::Cop parent class

backus:
- Works around the RuboCop::Cop::Cop.inherited hook
 - Inlines the spec only behavior so it doesn't need to be included
   everywhere
 - Inlines the language modules so the constants are available
   everywhere


lib/rubocop/cop/rspec/cop.rb
```diff
@@ -0,0 +1,64 @@
...
+      def self.<(other)
+        other.equal?(RuboCop::Cop::Cop) || super
+      end
+    end
+    private_constant(:WorkaroundCop)
+
+    module RSpec
+      # The criteria for whether rubocop-rspec analyzes a certain ruby file
+      # is configured via `AllCops/RSpec`. For example, if you want to
+      # customize your project to scan all files within a `test/` directory
```
bquorning: `test/` directory? That's not what the example code says.

lib/rubocop/rspec/config_formatter.rb
```diff
@@ -19,6 +19,7 @@ def dump
 
       def unified_config
         cops.each_with_object(config.dup) do |cop, unified|
+          next if cop.eql?('RSpec/Cop')
```
bquorning: I don't understand this line.

lib/rubocop/cop/rspec/cop.rb
backus: Good catch

lib/rubocop/rspec/config_formatter.rb
backus: This config formatter grabs the configuration for each cop. It receives a set of keys from YARD under the `RuboCop::Cop::RSpec` namespace. It skips `RSpec/Cop` since it is not a cop and doesn't appear in our config file

backus: I found a better way to handle this actually. Good fit for the yard `@abstract` tag

backus:
@bquorning ready for another pass

bquorning:
Looks good :+1:

Issue #232, opened 2016-10-27
Add additional namespace to cops


backus:
After @JonRowe's feedback in #228 I'd like to start by at least adding an additional namespace for our cops. This change can be pretty simple:

``` ruby
diff --git i/lib/rubocop/cop/rspec/leading_subject.rb w/lib/rubocop/cop/rspec/leading_subject.rb
index c9b813a..d87e927 100644
--- i/lib/rubocop/cop/rspec/leading_subject.rb
+++ w/lib/rubocop/cop/rspec/leading_subject.rb
@@ -31,6 +31,10 @@ module RuboCop

         MSG = 'Declare `subject` above any other `let` declarations'.freeze

+        def self.cop_name
+          "RSpec/Style/LeadingSubject"
+        end
+
         def_node_matcher :subject?, '(block $(send nil :subject ...) args ...)'

         def on_block(node)
```

we could also add another namespace (might be too much nesting idk) and define a `cop_name` method which just defines `name.split('::').last(3).join('/')`. This might be a mess though since, as I've noted before, we can't just do `class RuboCop::RSpec::Cop < RuboCop::Cop::Cop` because `RuboCop::Cop::Cop` mutates some class level state on the `inherited` hook which I don't know how to avoid.

The point of this name spacing would be to start creating an end-user visible differentiation between stylistic and opinionated cops which should be treated differently from linting and best practice cops

backus:
@bquorning interested what you think here. I know there are a few parts of the rubocop codebase which assume a cop name has a single namespace but, as far as I can tell, deviating from this contract doesn't break anything

bquorning:
I’m not completely sold on the idea. This feels like something that belongs inside RuboCop proper.

They have the Rails cops (which really should be a separate project) – they ought to be namespaced as well, I guess.

pirj:
FactoryBot and Rails namespaces were added in #1019
Speaking of separating cops by their importance or purpose seems to be a task nobody is willing to take.

PR #238, opened 2016-11-06
Refactor DescriptionExtractor to decorate yardoc

backus:
I found that the description extractor was mainly speaking in terms of predicates that passed around the YARD code object. This seems like a smell so this is an attempt to wrap this logic up into an instance.

### DIFF
--- a/lib/rubocop/rspec/description_extractor.rb
+++ b/lib/rubocop/rspec/description_extractor.rb
@@ -2,36 +2,71 @@ module RuboCop
   module RSpec
     # Extracts cop descriptions from YARD docstrings
     class DescriptionExtractor
-      COP_NAMESPACE = 'RuboCop::Cop::RSpec'.freeze
-      COP_FORMAT    = 'RSpec/%s'.freeze
-
       def initialize(yardocs)
-        @yardocs = yardocs
+        @code_objects = yardocs.map(&CodeObject.public_method(:new))
       end
 
       def to_h
-        cop_documentation.each_with_object({}) do |(name, docstring), config|
-          config[format(COP_FORMAT, name)] = {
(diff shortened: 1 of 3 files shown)


backus:
@bquorning what do you think of this refactor?

lib/rubocop/rspec/description_extractor.rb
```diff
@@ -2,36 +2,71 @@ module RuboCop
...
+        #
+        # @return [Hash]
+        def configuration
+          { cop_name => { 'Description' => description } }
+        end
+
+        private
+
+        def cop_name
+          Object.const_get(documented_constant).cop_name
```
backus: I was conflicted about this line. The upside is that we now don't duplicate the `cop_name` rule here (which would make #232 for example). The concern is that using `const_get` could be fragile

backus:
ping @bquorning any thoughts on this refactor?

lib/rubocop/rspec/description_extractor.rb
```diff
@@ -2,36 +2,71 @@ module RuboCop
...
+        end
+
+        def description
+          yardoc.docstring.split("\n\n").first.to_s
```
bquorning: I don't think it's necessary to call `to_s` at the end. Calling `split` makes me confident we are working with a string.

bquorning:
This is definitely an improvement. I'm fine with the `const_get`. I think the errors it would cause are "good" errors, that with the old `format(COP_FORMAT, name)` code wouldn't surface.

bquorning:
:+1:

lib/rubocop/rspec/description_extractor.rb
backus: It is necessary for cases where there are no docs. I didn't add a test case for this though so good catch

PR #241, opened 2016-11-12
Do not flag top level describe in a shared example

### LINKED ISSUE #216: DescribeClass cop should not flag top level describe in a shared example
Example false positive:

```
# frozen_string_literal: true

RSpec.shared_examples 'Common::Interface' do
  describe '#public_interface' do
    it 'conforms to interface' do
      # ...
    end
  end
end
```

clupprich:
Fixes #216.

These were my first 15min _ever_ with cops and AST, hope this is somewhat functional and usable :)

### DIFF
--- a/lib/rubocop/cop/rspec/describe_class.rb
+++ b/lib/rubocop/cop/rspec/describe_class.rb
@@ -39,12 +39,12 @@ module RuboCop
           (sym {:request :feature :routing :view}))
         PATTERN
 
-        def_node_matcher :shared_examples?, <<-PATTERN
-          (block (send (const nil :RSpec) :shared_examples ...) ...)
+        def_node_matcher :shared_group?, <<-PATTERN
+          (block (send {(const nil :RSpec) nil} {#{SharedGroups::ALL.to_node_pattern}} ...) ...)
         PATTERN
 
         def on_top_level_describe(node, args)
-          return if shared_examples?(root_node)
+          return if shared_group?(root_node)
           return if valid_describe?(node)
 
           describe_with_metadata(node) do |pairs|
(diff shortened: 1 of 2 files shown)


lib/rubocop/cop/rspec/describe_class.rb
```diff
@@ -39,7 +39,12 @@ class DescribeClass < Cop
           (sym {:request :feature :routing :view}))
         PATTERN
 
+        def_node_matcher :shared_examples?, <<-PATTERN
+          (block (send (const nil :RSpec) :shared_examples ...) ...)
```
backus: We should probably support the following cases

```ruby
RSpec.shared_examples do
end

shared_examples do
end

RSpec.shared_context do
end

shared_context do
end
```

So I think the node pattern like this would look like so

```
(block (send {(const nil :RSpec) nil} {:shared_examples :shared_context : shared_examples_for} ...) ...)
```

I have `RuboCop::RSpec::Language` which defines a lot of the RSpec public interface: https://github.com/backus/rubocop-rspec/blob/master/lib/rubocop/rspec/language.rb

with that you can do

```ruby
<<-PATTERN
(block (send {(const nil :RSpec) nil} {#{SharedGroups::ALL.to_node_pattern} ...) ...)
PATTERN
```

backus:
Thanks for the help with this! I left one comment on other cases this should cover. I'm good to merge this after that :smile:

clupprich:
Thanks! 😀 I've updated the PR with the proposed use of `SharedGroups::ALL`, and also added some more specs, which should cover all cases.

backus:
Thanks @clupprich!

PR #243, opened 2016-11-15
Add SingleArgumentMessageChain cop

### LINKED ISSUE #240: Cop idea: Detect receive_message_chain/stub_chain with just one argument
Sometimes, when code change, I have seen specs change from e.g. `allow(foo).to receive_message_chain(:bar, :baz)` to `allow(foo).to receive_message_chain(:bar)`. Since `receive_message_chain` (and `stub_chain`) [is a code smell](https://github.com/backus/rubocop-rspec/pull/187), we might have a cop that suggests replacing with `allow(foo).to receive(:bar)`.

I know that it clashes a little bit with our MessageChain cop from #187, but I think that it would be useful anyway. Perhaps it could even be part of that cop?

bquorning:
Suggests replacing `receive_message_chain(:bar)` with `receive(:bar)`.

Fixes #240.


lib/rubocop/cop/rspec/single_argument_message_chain.rb
```diff
@@ -0,0 +1,53 @@
...
+      #   allow(foo).to receive("bar.baz")
+      #
+      class SingleArgumentMessageChain < Cop
+        MESSAGE = 'Use `%<recommended_method>s` instead of calling ' \
+          '`%<called_method>s` with a single argument'.freeze
+
+        MESSAGE_CHAIN_METHODS = [
+          :receive_message_chain,
+          :stub_chain
+        ].freeze
```
backus: This is in `MessageChain` too. Think we can add a new chain entry to the `RSpec::Language` definitions file?

bquorning: In `RuboCop::Cop::RSpec::DescribedClass` I see `RuboCop::RSpec::Language::ALL` being refered to as `RSPEC_BLOCK_METHODS`. I am not sure whether `:receive_message_chain` belongs in that module of block methods.

backus: I meant something like adding

```diff
diff --git i/lib/rubocop/rspec/language.rb w/lib/rubocop/rspec/language.rb
index d44e2b7..be2b92f 100644
--- i/lib/rubocop/rspec/language.rb
+++ w/lib/rubocop/rspec/language.rb
@@ -31,6 +31,10 @@ module RuboCop
         attr_reader :selectors
       end

+      module Matchers
+        MESSAGE_CHAIN = SelectorSet.new(%i(receive_message_chain stub_chain))
+      end
+
       module ExampleGroups
         GROUPS  = SelectorSet.new(%i(describe context feature example_group))
         SKIPPED = SelectorSet.new(%i(xdescribe xcontext xfeature))
```

to https://github.com/backus/rubocop-rspec/blob/master/lib/rubocop/rspec/language.rb

`Matchers::MESSAGE_CHAIN.include?(method_name)` below

For context: I added this language file because most of the cops duplicated knowledge about part of the RSpec public API and many cops had different holes depending on what the original cop author remembered to use. Unifying the language definitions made it easier to make sure that all cops were in sync

bquorning: Done and pushed.

bquorning:
@backus Will you review and merge, please?

backus:
Looks great. Nice job

PR #252, opened 2016-12-17
Add Before/After(:all) cop

andyw8:
This is a continuation of https://github.com/backus/rubocop-rspec/pull/108, as there was no response from @cfabianski.


andyw8:
@backus let me know what you think. I think this could be one of the more confusing cops when users encounter it, but it's difficult to write a clear description.

lib/rubocop/cop/rspec/before_after_all.rb
```diff
@@ -0,0 +1,49 @@
...
+          'to leak between tests. If you are using rspec-rails, and '\
+          '`use_transactional_fixtures` is enabled, then records created in '\
+          '`before(:all)` are not rolled back.'.freeze
+
+        BEFORE_AFTER_METHODS = [
+          :before,
+          :after
+        ].freeze
+
+        ALL_PAIR = s(:sym, :all)
```
bquorning: https://www.relishapp.com/rspec/rspec-core/v/3-5/docs/hooks/before-and-after-hooks says 

> Note: the `:example` and `:context` scopes are also available as `:each` and
`:all`, respectively. Use whichever you prefer.

We should also check for before/after `:context`.

andyw8: @bquorning Good point. Added in ed5049b.

backus: Might be useful to add these to the language definition and then generate a node pattern https://github.com/backus/rubocop-rspec/blob/master/lib/rubocop/rspec/language.rb#L57

backus:
@andyw8 thoughts on special casing spec_helper? I think it is usually reasonable to do some general setup with a `before(:all)` hook in those specs.

backus:
I think the message is probably fine

andyw8:
@backus good point. `rails_helper` too probably? And anything in `spec/support` ?

andyw8:
Actually, won't these excluded already since they don't end in `_spec.rb` ?

backus:
@andyw8 I believe the global rule for our cops is either in the `spec` directory or ending in `_spec.rb`

bquorning:
I was gonna write an argument for why I use `before(:all)` in my specs, but then I noticed that I use `after(:all)` in all the same places to reset the global state 😳  So, now I use `around(:all)` instead.

I now have only one instance of `before(:all)` left in my spec suite, and it already had the comment `# FIXME: This is slow, and this changes global state.` ;-)

lib/rubocop/cop/rspec/before_after_all.rb
```diff
@@ -0,0 +1,50 @@
...
+      # Check that before/after(:all) isn't being used.
+      #
+      # @example
+      #   # bad
+      #   #
+      #   # Faster but risk of state leaking between examples
+      #   #
+      #   describe MyClass do
+      #     before(:all) { Widget.create }
+      #     after(:all) { Widget.delete_all }
```
bquorning: How do we like `around`?
```rb
around(:all) do |example|
  Widget.create
  example.run
  Widget.delete_all
end
```

andyw8: For `around(:each)`, yes, that should be covered too.

I don't think `around(:all)` exists yet: https://github.com/rspec/rspec-core/issues/1031#issuecomment-189409723

andyw8: Oh, just saw your comment above, so it does exist?

andyw8:
@backus any pointers for where to look in the code for how to exclude particular paths?

backus:
@andyw8 I believe you can exclude them in the config file. Here is an example from rubocop proper https://github.com/bbatsov/rubocop/blob/master/config/default.yml#L725-L733

andyw8:
@backus thanks, I've added them to the config file (3b2285a), but that doesn't seem to picked up by the tests, is that expected? (the spec fail)

backus:
@andyw8 The `Exclude` configuration is used by rubocop to determine whether or not to even run a cop against the code in a given file. The tests fail because `inspect_source` applies the cop directly I believe. If you look at rubocop it doesn't actually test the excludes in its config.

backus:
Thanks for the help @andyw8! @trevorcreech wrapped up this cop for #326

andyw8:
@backus @trevorcrrech thanks for taking over.

PR #253, opened 2016-12-18
Simplify `TopLevelDescribe` using `#each_child_node`

Drenmi:
I was digging through the source code, doing research for the extension system for RuboCop, and I noticed `TopLevelDescribed` added some things that are already handled quite elegantly in `RuboCop::Node`, so I thought I'd create a PR. 😀


backus:
Awesome, thank you!

PR #255, opened 2016-12-20
Refactor selector set

backus:
Trying to make the language namespace resolve more duplication between cops


backus:
@bquorning or @andyw8 could you review this?

bquorning:
I’m a bit unsure about the naming of `block_pattern` and `send_pattern`, but I cannot come up with any better names. Code sure looks cleaner :+1:

backus:
Cool. I can't come up with better names now either so I'll leave it for now and we can see how the code evolves

PR #259, opened 2016-12-26
Detect repeated descriptions in example groups

backus:


backus:
@bquorning @andyw8 this is reviewable

lib/rubocop/cop/rspec/repeated_description.rb
```diff
@@ -0,0 +1,82 @@
...
+
+        private
+
+        def repeated_descriptions(node)
+          hooks_in_group(node)
+            .group_by(&:method_args)
+            .reject { |args, _| args.empty? }
+            .values
+            .reject(&:one?)
+            .flatten
```
bquorning: Would this be more readable using local variables?

```rb
hooks_with_arguments = hooks_in_group(node).group_by(&:method_args).reject { |args, _| args.empty? }
hooks_with_identical_arguments = hooks_with_arguments.values.reject(&:one?)
hooks_with_identical_arguments.flatten
```

If I got the names wrong, it’s because I didn’t understand the code.

spec/rubocop/cop/rspec/repeated_description_spec.rb
```diff
@@ -0,0 +1,19 @@
+# frozen_string_literal: true
+
+describe RuboCop::Cop::RSpec::RepeatedDescription do
+  subject(:cop) { described_class.new }
+
+  it 'registers an offense for repeated descriptions' do
```
backus: Yeah good catch. I actually thought I had more than one test here but I guess not. I'll add a few more

lib/rubocop/cop/rspec/repeated_description.rb
backus: I tried to clarify with inline comments. I prefer a functional style like this when possible but only if the code is still clear. Is the updated code clear?

lib/rubocop/cop/rspec/repeated_description.rb
```diff
@@ -0,0 +1,82 @@
...
+        def hooks_in_group(node, &block)
+          return to_enum(__method__, node) unless block_given?
+
+          node.each_child_node { |child| find_hooks(child, &block) }
+        end
+
+        def find_hooks(node, &block)
+          return if scope_change?(node)
+
+          example(node, &block)
```
backus: Sorry about that. I added some doc comments and I also renamed the misleading name... It was a holdover from code I borrowed from #254. Didn't realize I forgot to change the name sorry about that

spec/rubocop/cop/rspec/repeated_description_spec.rb
backus: @bquorning the tests should be sufficient now

lib/rubocop/cop/rspec/repeated_description.rb
backus: This method does a recursive search until it hits a scope changing node. RuboCop's node pattern methods will yield (unless you defined a predicate method ending in `?` I believe) if you give it a block and the provided node matches. By writing

```ruby
def_node_pattern :desired_node, '...'
def_node_pattern :node_should_stop_recursion?, '...'

def desired_nodes(node, &block)
  node.each_child_node { |child| desired_node_search(node, &block) }
end

def desired_node_search(node, &block)
  return unless node_should_stop_recursion?(node)

  desired_node(node)
  desired_nodes(node)
end
```

you get a recursive node search that stops when it hits a match for `node_should_stop_recursion?`. The end result is that I can write

```ruby
desired_nodes(node) do |match|
  add_offense(match, ...)
end
```

which I like. I follow this pattern in a handful of cops

backus:
@bquorning ready for another pass.

lib/rubocop/cop/rspec/repeated_description.rb
```diff
@@ -0,0 +1,87 @@
...
+        # @!method scope_change?(node)
+        #
+        #   Detect if the node is an example group or shared example
+        #
+        #   Selectors which indicate that indicate we should stop searching
+        #
+        def_node_matcher :scope_change?,
+                         (ExampleGroups::ALL + SharedGroups::ALL).block_pattern
+
+        MSG = "Don't repeat descriptions withhin an example group.".freeze
```
andyw8: Typo in "withhin".

andyw8:
Ran it over a large codebase and it caught 74 issues, all appear to be genuine. Nice work on this!

bquorning:
This looks good. Add a CHANGELOG entry (and perhaps copy it to the PR description for future reference), and it’s a 👍 from me.

lib/rubocop/cop/rspec/repeated_description.rb
backus: Good catch thanks. Fixed now

backus:
@bquorning whoops good catch on the changelog. Added. Will merge when build is green :smile:

PR #261, opened 2016-12-28
Add RepeatedExample cop

### LINKED ISSUE #260: Detect repeated example bodies
Similar to #258, we should detect identical block bodies. For example,

```ruby
it "does x" do
  expect(foo).to be(bar)
end

it "does y" do
  expect(foo).to be(bar)
end
```

should be a violation

backus:
Fixes #260


config/default.yml
```diff
@@ -141,6 +141,10 @@ RSpec/RepeatedDescription:
   Enabled: true
   Description: Check for repeated description strings in example groups.
 
+RSpec/RepeatedExample:
+  Enabled: true
+  Description: Detect repeated examples within example groups.
```
andyw8: Minor point, but should we follow the "Check..." sentence convention used by the other cops?

backus: Yeah good point I'll change that

lib/rubocop/cop/rspec/repeated_example.rb
```diff
@@ -0,0 +1,41 @@
...
+    module RSpec
+      # Check for repeated examples within example groups.
+      #
+      # @example
+      #
+      #    it 'is valid' do
+      #      expect(user).to be_valid
+      #    end
+      #
+      #    it 'is valid' do
```
bquorning: If we instead use two different descriptions, I think it will be more clear what this cop is checking.

spec/rubocop/cop/rspec/repeated_example_spec.rb
```diff
@@ -0,0 +1,65 @@
+# frozen_string_literal: true
+
+describe RuboCop::Cop::RSpec::RepeatedExample do
+  subject(:cop) { described_class.new }
+
+  it 'registers an offense for repeated example' do
+    expect_violation(<<-RUBY)
+      describe 'doing x' do
+        it "does x" do
+                    ^^ Don't repeat examples within an example group.
```
bquorning: I know we can't highlight the entire block contents, but just just highlighting `do` looks a bit strange. Could we highlight the entire line instead?

lib/rubocop/cop/rspec/repeated_example.rb
backus: Good point

spec/rubocop/cop/rspec/repeated_example_spec.rb
backus: Yeah. I need to make the `ExpectViolation` thing support this. RuboCop currently will highlight the whole expression but only display the first line in its output. The `ExpectViolation` thing doesn't know how to handle multiple lines yet

spec/rubocop/cop/rspec/repeated_example_spec.rb
```diff
@@ -0,0 +1,65 @@
...
+  end
+
+  it 'does not register a violation if rspec keyword magic is involved' do
+    expect_no_violations(<<-RUBY)
+      describe 'doing x' do
+        it "does x" do
+          expect(foo).to be(bar)
+        end
+
+        it "does y", :magic do
```
bquorning: Nitpick: These are tags, not keywords. And I'd prefer using e.g. `:focus` instead of `:magic` to make it more "realistic".

backus: Fair. I used `:magic` to make it a bit more clear that what we're avoiding is tests that magically import special behavior using RSpec's tags.

Also this is definitely not a nitpick (I want this feedback). Matching rspec's language in this codebase is pretty helpful

lib/rubocop/cop/rspec/repeated_example.rb
backus: Updated

spec/rubocop/cop/rspec/repeated_example_spec.rb
backus: Updated

backus:
@bquorning @andyw8 any other flags here?

bquorning:
I still have concerns over the readability of that 7-line method chain.

Meh, let's merge this. 👍🏼

backus:
Thanks @bquorning. I'm finding a lot of cases recently where I have this sort of chaining for duplicate detection or code that isn't separated by a newline. I'll try to find a more clean implementation if this pattern keeps showing up

alexjfisher:
@backus Should this cop also ignore repeated examples if they occur in different case matches?
eg https://github.com/voxpupuli/puppet-squid/blob/192c27213b00f2beb1a7fea63fb109da97f4adc2/spec/classes/init_spec.rb has 28 offenses now.

https://travis-ci.org/voxpupuli/puppet-squid/jobs/193205490

Thanks.

backus:
Hm I've never seen specs written like that before. I'll take a closer look later today

backus:
@alexjfisher I'm taking a closer look at the moment but I'm going to open a new issue for your question and resume the discussion there. I personally find it much harder to track these discussions when they're in the implementing pull request

backus:
See https://github.com/backus/rubocop-rspec/issues/309

PR #262, opened 2016-12-28
Extract RuboCop::RSpec::{Example,ExampleGroup}

backus:
These objects wrap RuboCop::Node objects in order to unify traversal and also clarify how certain nodes are being accessed.

I've extracted this refactor from #261 which needed a lot of the same code that `RepeatedDescription` depended on.

### DIFF
--- a/lib/rubocop/cop/rspec/repeated_description.rb
+++ b/lib/rubocop/cop/rspec/repeated_description.rb
@@ -28,26 +28,13 @@ module RuboCop
       #     end
       #
       class RepeatedDescription < Cop
-        def_node_matcher :example, <<-PATTERN
-          (block $(send _ #{Examples::ALL.node_pattern_union} str ...) ...)
-        PATTERN
-
-        # @!method scope_change?(node)
-        #
-        #   Detect if the node is an example group or shared example
-        #
-        #   Selectors which indicate that we should stop searching
-        #
-        def_node_matcher :scope_change?,
-                         (ExampleGroups::ALL + SharedGroups::ALL).block_pattern
-
         MSG = "Don't repeat descriptions within an example group.".freeze
 
         def on_block(node)
           return unless example_group?(node)
 
-          repeated_descriptions(node).each do |repeated_hook|
-            add_offense(repeated_hook, :expression)
+          repeated_descriptions(node).each do |repeated_description|
+            add_offense(repeated_description, :expression)
           end
         end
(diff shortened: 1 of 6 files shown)


backus:
@andyw8 or @bquorning any thoughts here? I did this refactor to help me avoid painful duplication in #261

spec/rubocop/rspec/example_spec.rb
```diff
@@ -0,0 +1,62 @@
...
+
+    it 'can be used as a key in a hash' do
+      hash = {}
+
+      hash[example('it("foo")')] = 123
+
+      expect(hash[example('it("foo")')]).to be(123)
+    end
+
+    it 'computes #hash based on class and node' do
```
bquorning: All examples test the AST, except this one where we test the `Example` itself. Should it perhaps be in a different context/describe block?

lib/rubocop/rspec/example.rb
```diff
@@ -0,0 +1,56 @@
...
+
+      def implementation
+        extract_implementation(node)
+      end
+
+      def ==(other)
+        node.eql?(other.node)
+      end
+
+      alias eql? ==
```
bquorning: I think aliasing the other way around, and using `eql?` inside `#eql?` will make the code a tiny bit clearer. What do you think?
```rb
def eql(other)
  node.eql?(other.node)
end

alias == eql?
```

lib/rubocop/cop/rspec/repeated_description.rb
```diff
@@ -28,58 +28,30 @@ module RSpec
...
-
-          example(node, &block)
-          examples_in_scope(node, &block)
+          grouped_examples =
+            RuboCop::RSpec::ExampleGroup.new(node)
+              .examples
+              .group_by(&:doc_string)
+
+          grouped_examples
+            .reject { |description, group| !description || group.one? }
```
bquorning: I get a bit confused by `reject`, `not` and `or` in the same line. How about this instead:
```rb
.select { |description, group| description && group.size > 1 }
```

backus: Much better

lib/rubocop/rspec/example.rb
backus: Good idea

bquorning:
:+1:

PR #264, opened 2016-12-28
Fix expect violation to match rubocop's offense highlighting behavior

backus:
Extracted from #261


backus:
@bquorning can you review? This enables what you were asking about in highlighting the first line for #261

bquorning:
I don't know the difference between `offense.location` and `offense.highlighted_area`, but it looks like it works fine :+1:

backus:
@bquorning basically `offense.highlighted_area` is code from rubocop that is used to generate the highlight displayed to users

PR #266, opened 2016-12-29
Autocorrect in SingleArgumentMessageChain

bquorning:
When running the new `SingleArgumentMessageChain` cop I realised that autocorrect should be really simple to implement. It was.

1st commit cleans up a couple of copy/paste errors in the specs.


backus:
Looks good

PR #271, opened 2017-01-02
Fix bad NestedGroups top level describe behavior

### LINKED ISSUE #270: NestedGroups counts levels differently since version 1.9.0
With version 1.8.0, the following does not cause a NestedGroups offense:

```ruby
RSpec.describe Foo do
  context 'bar' do
    it 'baz' do
      pass
    end
  end

  describe '#qux' do
    context 'when quuz' do
      it 'zyxxy' do
        pass
      end
    end
  end
end
```

With 1.9.0, it seems the nested is counted differently, since it registers an offense, even thought the MaxNesting parameter default has stayed the same, at `2`.

This seems to be a bug or regressions, since this changes is not marked in the change log.

backus:
The changes in bd88c07 changes the node pattern for NestedGroups from

    (block (send nil [SELECTORS UNION PATTERN] ...) (args) ...)

to

    (block (send _ [SELECTORS UNION PATTERN] ...) ...)

which led to issue #270 being reported. While this did introduce a
change in behavior, it helped reveal an inconsistency in the
implementation. With the change mentioned above, the following code
generated an offense when it didn't in 1.8.0

    RSpec.describe Foo do
      context 'bar' do
        it 'baz' do
          expect(foo).to_not be(1)
          expect(foo).not_to be(1)
        end
      end

      describe '#qux' do
        context 'when quuz' do
          it 'zyxxy' do
            pass
          end
        end
      end
    end

the following nearly equivalent code *did* generate an offense in 1.8
though:

    describe Foo do
      context 'bar' do
        it 'baz' do
          expect(foo).to_not be(1)
          expect(foo).not_to be(1)
        end
      end

      describe '#qux' do
        context 'when quuz' do
          it 'zyxxy' do
            pass
          end
        end
      end
    end

The problem was that the previous node pattern would only start
counting nesting at the first bare `describe` or `context` and would
not count `RSpec.describe`. It was also probably much less efficient
since it was searching through all nodes inside every example group
instead of only searching from top level describes.

This change fixes #270 by starting a search at the top level describe
and properly handle `RSpec.describe` like `describe`.

This change also convinced me that the current default for NestedGroups
is too aggressive. Almost all of the specs that I write use
`RSpec.describe` which means that rubocop-rspec was more lenient for my
tests. The default MaxNesting is now 3 instead of 2.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## Master (unreleased)
 
+* Fix unintentional regression change in `NestedGroups` reported in #270. ([@backus][])
+* Change `MaxNesting` for `NestedGroups` from 2 to 3. ([@backus][])
+
 ## 1.9.0 (2016-12-29)
 
 * Add `MessageSpies` cop for enforcing consistent style of either `expect(...).to have_received` or `expect(...).to receive`, intended as a replacement for the `MessageExpectation` cop. ([@bquorning][])
--- a/config/default.yml
+++ b/config/default.yml
@@ -127,7 +127,7 @@ RSpec/NamedSubject:
 RSpec/NestedGroups:
   Description: Checks for nested example groups.
   Enabled: true
-  MaxNesting: 2
+  MaxNesting: 3
 
 RSpec/NotToNot:
   Description: Checks for consistent method usage for negating expectations.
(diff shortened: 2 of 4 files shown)


backus:
@mvz does this fix your issue?
@andyw8 or @bquorning could you review?

bquorning:
:+1:

backus:
Thanks @bquorning. I'll do a quick release with this now

Issue #189, opened 2016-08-24
Add expected and actual maximum to offense message for MaxExpectations


backus:
Good point from @mvz on https://github.com/troessner/reek/issues/1036#issuecomment-242225265:

> @backus thanks for the demo. I can see how this will be very useful in making our specs consistent. One thing I see is that there are several cops that mention some exceeded maximum, without mentioning what the value actually is ('Maximum example group nesting exceeded', and 'Too many expectations.'). It would be useful to have the values shown, as 'Example has too many lines. [10/5]' does.

PR #282, opened 2017-01-08
Show actual/max value in multiple expectations cop

Darhazer:
Following #189 I've added the total/max to `MultipleExpectations`, using the same format as in `ExampleLength`. This leaves `NestedGroups` the only cop not to provide this information, but as it adds the offence as soon as the max value is exceeded, it can't use the same format. If there is a suggestion how to implement this for `NestedGroups`, I'd be happy to open a PR for it as well


andyw8:
LGTM

backus:
Awesome. Thanks @Darhazer

PR #285, opened 2017-01-13
Add .gitattributes to improve CHANGELOG merging

backus:


bquorning:
If this works like I think it does (no more rebase because someone else added a line to the changelog) this is the PR of the year! Thank you!

Do you know if same change is implemented in the RuboCop repository?

backus:
Github doesn't respect it but it means that, locally, we can rebase without a merge conflict. Definitely improvement but would be much nicer if Github used it.

> Do you know if same change is implemented in the RuboCop repository?

Yeah I actually learned this trick from @pocke a while ago. See this PR https://github.com/bbatsov/rubocop/pull/3588#issuecomment-252417890

PR #287, opened 2017-01-13
Deprecate MaxNesting in favor of Max

### LINKED ISSUE #286: Suggestion: Rspec/NestedGroups maximum option
This is just a suggestion from my part since I got a bit confused while trying to set custom maximum value for nested groups example.

The maximum value set for most of the cops use key `Max` but for `NestedGroups` it is `MaxNesting`.
https://github.com/backus/rubocop-rspec/blob/master/config/default.yml#L130

Perhaps I might be wrong or I might be too late to suggest but I think it would be a consistent behaviour among cops if we can use just `Max` as key to set maximum values. Thanks for your time and efforts. 🙇

backus:
Fixes #286

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 * Fix internal dependencies on RuboCop to be compatible with 0.47 release. ([@backus][])
 * Add autocorrect support for `SingleArgumentMessageChain` cop. ([@bquorning][])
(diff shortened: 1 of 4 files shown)


backus:
@andyw8 or @bquorning can you review? This is in response to #286 which points out that we use a different Max variable for no good reason for `NestedGroups`

PR #289, opened 2017-01-14
Add 2.4.0 to the ruby matrix

backus:


andyw8:
Maybe we can drop 2.2 and 2.3? i.e. if it works with 2.1 and 2.4 then it's highly likely it'll work with everything in between.

backus:
@andyw8 I would definitely do that if we were talking about travis. I don't mind with buildkite though since I can execute all of these different tests in parallel which is why I put a lot of work into that build system :smile:

PR #290, opened 2017-01-14
Add rubocop master build step

backus:


.buildkite/pipeline.yml
```diff
@@ -29,7 +29,18 @@ steps:
     agents:
       - queue=elastic
 
+
```
andyw8: Accidental?

backus: Yes

dgollahon:
This seems like the kind of thing where it'd just be easy enough to resolve any new bugs whenever `rubocop` proper does a release. Since they don't adhere to semver, there's not too much we can do about compatibility anyway and this PR has gone stale.

PR #302, opened 2017-01-22
Add a preferred style for not using described_class. Fixes #193

### LINKED ISSUE #193: Add a preferred style for not using described_class
Using `described_class` isn't definitively better. I prefer it but many teams prefer the small amount of duplication. There should be a preferred_style that lets you enforce _not_ using `described_class` and autocorrects `described_class` references to use the class name.

See https://github.com/troessner/reek/pull/1035#issuecomment-243418073

Darhazer:
@backus 

Few notes:
* I had no idea how to name the supported styles. I've chosen `explicit` as that's what you are doing when not using desrcibed_class - explicitly refer the class - but maybe that's not best as there's no `implicit` option. Unless I rename described_class to implicit.
* The implementation is as ugly as hell. Any suggestions how to improve it are more than welcome

### DIFF
--- a/lib/rubocop/cop/rspec/described_class.rb
+++ b/lib/rubocop/cop/rspec/described_class.rb
@@ -6,10 +6,11 @@ module RuboCop
       # Checks that tests use `described_class`.
       #
       # If the first argument of describe is a class, the class is exposed to
-      # each example via described_class - this should be used instead of
-      # repeating the class.
+      # each example via described_class.
       #
-      # @example
+      # This cop can be configured using the `EnforcedStyle` option
+      #
+      # @example `EnforcedStyle: described_class`
       #   # bad
       #   describe MyClass do
       #     subject { MyClass.do_something }
@@ -19,11 +20,24 @@ module RuboCop
       #   describe MyClass do
       #     subject { described_class.do_something }
       #   end
+      #
+      # @example `EnforcedStyle: explicit`
+      #   # bad
+      #   describe MyClass do
+      #     subject { described_class.do_something }
+      #   end
+      #
+      #   # good
+      #   describe MyClass do
+      #     subject { MyClass.do_something }
+      #   end
+      #
       class DescribedClass < Cop
         include RuboCop::RSpec::TopLevelDescribe
+        include RuboCop::Cop::ConfigurableEnforcedStyle
 
         DESCRIBED_CLASS = 'described_class'.freeze
-        MSG             = "Use `#{DESCRIBED_CLASS}` instead of `%s`".freeze
+        MSG             = 'Use `%s` instead of `%s`'.freeze
 
         def_node_matcher :common_instance_exec_closure?, <<-PATTERN
           (block (send (const nil {:Class :Module}) :new ...) ...)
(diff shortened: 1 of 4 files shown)


backus:
Thanks for the PR @Darhazer. I'm reviewing now. The build is failing currently but I'm sure you are aware of that

lib/rubocop/cop/rspec/described_class.rb
```diff
@@ -38,27 +52,42 @@ def on_block(node)
           describe, described_class, body = described_constant(node)
           return unless top_level_describe?(describe)
 
-          find_constant_usage(body, described_class) do |match|
-            add_offense(match, :expression, format(MSG, match.const_name))
+          @described_class = described_class
```
backus: I usually try to avoid setting instance variables in the `on_*` methods that are used by private methods. I think it makes it a bit easier to reason about if the various cop methods aren't depending on state that may or may not be set. I think part of the reason I prefer this (and it seems to be the style in most regular rubocop cops) is because it is sort of unclear what the lifecycle of a cop is. It seems like it gets initialized once per file but I actually haven't checked to confirm this before :laughing:

lib/rubocop/cop/rspec/described_class.rb
```diff
@@ -38,27 +52,42 @@ def on_block(node)
...
-            corrector.replace(node.loc.expression, DESCRIBED_CLASS)
+            corrector.replace(node.loc.expression, replacement)
           end
         end
 
         private
 
-        def find_constant_usage(node, described_class, &block)
-          yield(node) if node.eql?(described_class)
+        def find_constant_usage(node, &block)
```
backus: I would rename this method now that it is looking for constant usage OR `described_class` usage.

lib/rubocop/cop/rspec/described_class.rb
Darhazer: @backus I completely agree, but the alternative is in the autocorrect method to travel the nodes up to the top-level describe, in order to find what should be used. I can attempt such implementation of course, but if there are any other alternatives, I would like to explore them. Any ideas?

backus: Yeah I actually have sometimes bitten the bullet before and just called the node traversal calls twice. That is a good reason though. Maybe a comment above the ivar assign then would be helpful?

Also, although we have a cop that discourages this, what happens if we have something like this?

``ruby
RSpec.describe Foo do
  it 'blah' do
    described_class
  end
end

RSpec.describe Bar do
  it 'blah' do
    described_class
  end
end
```

**Comment by backus** [comment]
@Darhazer mind rebasing this and adding a changelog entry? I think as long as the instance variable doesn't result in buggy behavior with multiple top level describes then I'm onboard with this. The code looks good to me otherwise. I usually run my code through mutant if you'd like to try that. To do that you would add `gem 'mutant-rspec'` to the `Gemfile.local` (you'll have to create this) and then run:

```sh
$ bundle exec mutant --use rspec 'RuboCop::Cop::RSpec::DescribedClass'
```

that should show you how strong the tests are for your changes

lib/rubocop/cop/rspec/described_class.rb
Darhazer: Added a test case for this. It works with the multiple describes. Also added a comment

Darhazer:
PR updated. Added a test for the multiple top-level describes, comment for the instance variable and a changelog entry.

backus:
Thanks @Darhazer!

PR #306, opened 2017-01-26
Add it_has_behavior to Includes::ALL

jdelStrother:


backus:
@jdelStrother I don't think this is a builtin include

jdelStrother:
Yeah, sorry - you're right.   I think I confused myself 😕

backus:
No problem! Was there a particular use case you were trying to fix? This made me think about whether we should add a configuration option for globally specifying example group aliases.

jdelStrother:
It looks like someone added the alias from https://www.relishapp.com/rspec/rspec-core/v/2-0/docs/example-groups/shared-example-group#aliasing-"it-should-behave-like"-to-"it-has-behavior" to our project config because it read better in some cases. I didn't realise it wasn't built in - I might try and reword things back to the default methods.

backus:
Cool. In the meantime that will give you better static analysis via rubocop-rspec. The `RSpec/EmptyExampleGroup` cop has the `CustomIncludeMethods` but realistically this should be a global config!

PR #317, opened 2017-02-06
Add AroundBlock cop. Fixes #315

### LINKED ISSUE #315: Detect invalid `around`
```
# bad
around do
  blah
end

# bad
around do |test|
  blah
end

# good
around do |test|
  blah
  test.call
end

# good
around do |test|
  blah
  test.run
end

# ignore this since it is hard to know statically
around do |test|
  blah
  method_invocation(&test)
end
```

Darhazer:
@backus 
Unfortunately I didn't find an easy way to avoid the instance var in this cop as well.
Please review the error messages.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Master (unreleased)
 
+* Add `AroundBlock` cop. ([@Darhazer][])
 * Add `EnforcedStyle` configuration for `RSpec/DescribedClass` cop. ([@Darhazer][])
 * Fix false positive for `RSpec/RepeatedExample` cop. ([@redross][])
 
--- a/config/default.yml
+++ b/config/default.yml
@@ -9,6 +9,10 @@ RSpec/AnyInstance:
   Description: Check that instances are not being stubbed globally.
   Enabled: true
 
+RSpec/AroundBlock:
+  Description: Checks that around blocks actually run the test.
+  Enabled: true
+
 RSpec/BeEql:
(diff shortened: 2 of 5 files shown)


backus:
@Darhazer What about just defining a node matcher for `around` blocks? In the cases where it yields an lvar we could just do a search and avoid using the variable force which I find sort of complicated (definitely worth it for RuboCop but I personally haven't implemented anything with it). Thoughts?

Darhazer:
I'll attempt this re-implementation over the weekend :+1: 
Btw I tried this initially (as the hooks argument uses it) but then realized that I don't know how to check block arguments and switched to VariableForce instead. But with a node matcher it will look simpler, so I'll give another try.

backus:
@Darhazer Great. Also make sure to look into `def_node_search`. Works exactly the same way but makes it really easy to just recursively search starting at a given node

Darhazer:
PR updated

backus:
Looking good! You have some style violations but after that I think I'm good to merge

Darhazer:
PR updated

backus:
@Darhazer almost there! We have a build step which asserts that the documented description is the same as the one in the config. Can you update them so they are the same? You can just run `bin/build_config` and commit the result

Darhazer:
Fixed. Any idea why it doesn't fail when running `rspec`? How can I test the entire build locally?

backus:
I just don't have it implemented with RSpec. If you run `bundle exec rake` it should be identical to CI

PR #319, opened 2017-02-10
Fix false positives of RepeatedExample with `its` examples

redross:
RepeatedExample cop incorrectly labels different examples as
repeated since the argument passed to its is ignored, ex.:

its(:x) { is_expected.to be_present }
its(:y) { is_expected.to be_present }

Fixed by adjusting the key used for grouping if `its` is used.

cc @backus

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## Master (unreleased)
 
 * Add `EnforcedStyle` configuration for `RSpec/DescribedClass` cop. ([@Darhazer][])
+* Fix false positive for `RSpec/RepeatedExample` cop. ([@redross][])
 
 ## 1.10.0 (2017-01-15)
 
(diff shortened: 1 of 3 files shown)


backus:
Great fix. Thank you @redross. Will merge when build is green :smile:

backus:
Looks like you've got a few style violations @redross. Could you actually add a changelog entry too?

redross:
will update

redross:
ping @backus

CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 ## Master (unreleased)
 
 * Add `EnforcedStyle` configuration for `RSpec/DescribedClass` cop. ([@Darhazer][])
+* Fix false negative for `RSpec/RepeatedExample` cop. ([@redross][])
```
backus: This would be a "false positive", right?

redross: Of course :D Sorry, updating.

backus:
Thanks again @redross!

redross:
Thanks for the quick response and review :+1:

backus:
> Thanks for the quick response and review 👍

Happy to do so. Feel free to open more PRs if any other improvements stand out to you :smile:

PR #323, opened 2017-02-14
Extend RSpec/DescribeMethod cop

redross:
## Description

Makes the `DescribeMethod` cop to enforce that when describing a class, the second level describes should specify method names (`#instance` or `.class`).

Allows whitelisting predefined describe names that will be ignored.

## Reasoning

The codebase I'm working on tries to follow at least some parts of the [rspec-style-guide](https://github.com/reachlocal/rspec-style-guide) and one part of it that we really like and think it helps keep specs consistent is the "Describe block naming" [rule](https://github.com/reachlocal/rspec-style-guide#describe-block-naming), which this cop tries to enforce.

This can be used to check the whole spec structure and make it as close as possible to:
```ruby
describe Class
  describe .class_method
    context X
      specs..
    context Y
  describe #instance_method
    ...
  describe #other_instance_method
    ...
``` 

Any tips on how this can be improved are greatly appreciated. cc @backus

### DIFF
--- a/config/default.yml
+++ b/config/default.yml
@@ -39,7 +39,7 @@ RSpec/DescribedClass:
   - explicit
 
 RSpec/DescribeMethod:
-  Description: Checks that the second argument to `describe` specifies a method.
+  Description: Checks that describes specify methods.
   Enabled: true
 
 RSpec/DescribeSymbol:
--- a/lib/rubocop/cop/rspec/describe_method.rb
+++ b/lib/rubocop/cop/rspec/describe_method.rb
@@ -3,33 +3,124 @@
 module RuboCop
   module Cop
     module RSpec
-      # Checks that the second argument to `describe` specifies a method.
+      # Checks that describes specify methods.
       #
       # @example
       #   # bad
-      #   describe MyClass, 'do something' do
+      #   describe MyClass do
+      #     describe 'flowers' do; end
+      #   end
+      #
+      #   describe MyClass, 'flowers' do
       #   end
       #
       #   # good
+      #   describe MyClass do
+      #     describe '#my_instance_method' do; end
+      #   end
+      #
+      #   describe MyClass do
+      #     describe '.my_class_method' do; end
+      #   end
+      #
       #   describe MyClass, '#my_instance_method' do
       #   end
       #
       #   describe MyClass, '.my_class_method' do
       #   end
+      #
+      # @example configuration
+      #
+      #   # .rubocop.yml
+      #   RSpec/DescribeMethod:
+      #     IgnoredDescribes:
+      #     - flowers
+      #     - stars
+      #
+      #   # bad
+      #   describe MyClass do
+      #     describe 'not_flowers_or_stars' do; end
+      #   end
+      #
+      #   describe MyClass, 'not_flowers_or_stars' do
+      #   end
+      #
+      #   # good
+      #   describe MyClass do
+      #     describe 'flowers' do; end
+      #     describe 'stars' do; end
+      #     describe '#some_method' do; end
+      #     describe '.some_class_method' do; end
+      #   end
+      #
+      #   describe MyClass, 'flowers' do
+      #   end
+      #
       class DescribeMethod < Cop
         include RuboCop::RSpec::TopLevelDescribe
         include RuboCop::RSpec::Util
 
-        MSG = 'The second argument to describe should be the method '\
-              "being tested. '#instance' or '.class'.".freeze
-
         METHOD_STRING_MATCHER = /\A[\#\.].+/
+
+        MSG = 'The second level describes should be the method '\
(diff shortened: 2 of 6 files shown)


backus:
This looks like a great cop, thanks @redross.

You asked for a few tips so here are some high level pointers (I have only had a minute or so to scan the code here).

* I heavily use the `describe Foo, '#bar'` style. I think that should be supported too
* Take a look at `RuboCop::RSpec::ExampleGroup` and grep for some of its usage https://github.com/backus/rubocop-rspec/blob/master/lib/rubocop/rspec/example_group.rb
  * Specifically, you should be able to look at top level describes and use this class to then just get all of the subsequent example groups in scope. Feel free to add a method to the class if you need to do so something like `example_group.example_groups.select(&:describe?)` or something like that.
* Can this build or be added to `DescribeMethod`? https://github.com/backus/rubocop-rspec/blob/master/lib/rubocop/cop/rspec/describe_method.rb

redross:
Thanks for the tips, will try to find time this week to update the PR.

redross:
Updated the PR, by merging `DescribeNaming` cop with `DescribeMethod`. I did try to use the `ExampleGroup` class, but it's probably far from great yet.

Could you elaborate a bit on the second tip you made? I'm following the idea of checking second level describes that are inside a described class block, but maybe this all could be simplified, by just picking out all of the describes and checking for names that don't match `#smth`/`.smth`?

backus:
Thanks @redross. I'll take another look later today after and try to explain. The `ExampleGroup` class (and its companions) are very new and I'm still trying to figure out how useful they are so sorry about them being a bit confusing

redross:
Rebased on latest master. No pushing, but if you have any tips on how to improve this I'm here :) cc @backus

lib/rubocop/cop/rspec/describe_method.rb
```diff
@@ -3,32 +3,123 @@
...
       #   end
       #
       #   describe MyClass, '.my_class_method' do
       #   end
+      #
+      # @example configuration
+      #
+      #   # .rubocop.yml
+      #   RSpec/DescribeMethod:
+      #     IgnoredDescribes:
```
backus: I'm not sure how useful this configuration is. Are there instances where configuring an `IgnoreDescribes` entry would buy me more than just using `rubocop:disable` once or twice?

config/default.yml
```diff
@@ -39,7 +39,7 @@ RSpec/DescribedClass:
   - explicit
 
 RSpec/DescribeMethod:
-  Description: Checks that the second argument to `describe` specifies a method.
+  Description: Checks that arguments and sub-describes of `describe` specify a method.
```
backus: "Checks that secondary arguments to top level describes and nested describes specify methods

lib/rubocop/cop/rspec/describe_method.rb
```diff
@@ -3,32 +3,123 @@
...
+      #   end
+      #
       class DescribeMethod < Cop
         include RuboCop::RSpec::TopLevelDescribe,
                 RuboCop::RSpec::Util
 
-        MESSAGE = 'The second argument to describe should be the method ' \
-                  "being tested. '#instance' or '.class'".freeze
         METHOD_STRING_MATCHER = /^[\#\.].+/
+        MESSAGE = 'The second level describes should test described'\
```
backus: How about "nested describes should test.."? I imagine if someone has a third level describe it is also going to flag that?

lib/rubocop/cop/rspec/describe_method.rb
```diff
@@ -3,32 +3,123 @@
...
+                  " class methods: '#instance' or '.class'".freeze
```
backus: How about `a method (either '#instance' or '.class')`. Saying "class method" might be confusing

backus:
@redross this is looking pretty good. I am skeptical about the usefulness of the `IgnoreDescribes` and I'd also like to cleanup the message wording a bit. Otherwise I think I am onboard with these changes :smile:

lib/rubocop/cop/rspec/describe_method.rb
redross: We have a couple of predefined describe blocks that are added to internal framework component tests, it looks pretty readable and I wanted to preserve the option of doing such things. It's kind of from my own personal need, so it might not be used very often, but then again - this shouldn't hurt performance too much and is quite isolated.

redross:
Sorry for taking so long to respond, will try to find time next week to update the PR.

lib/rubocop/cop/rspec/describe_method.rb
redross: It actually checks for second level describes of constants, so third level describes are ignored (unless second level describe is of a constant/class (can happen with namespaced classes))

redross:
@backus any tips on how I could improve this even more?

PR #326, opened 2017-02-22
Add Before/After(:all) cop (take 3)

trevorcreech:
Trying to push #252 over the line (which started with #108).

* Removed broken `Exclude` specs per https://github.com/backus/rubocop-rspec/pull/252#issuecomment-268933620
* Rebased on master

Thanks to @andyw8 and @cfabianski


backus:
Hurray! Thanks for wrapping this up @trevorcreech :smile:

Issue #233, opened 2016-11-02
Add option for MultipleExpectation to special case aggregate_failures


backus:
As @JonRowe mentioned in #228, the general guideline is to have "one business rule expectation" per spec. Sometimes this isn't always possible with a single `expect(...)`. It would be nice to support something like

```ruby
aggregate_failures("verifying response") do
  expect(response.status).to eq(200)
  expect(response.headers).to include("Content-Type" => "text/plain")
  expect(response.body).to include("Success")
end
```

being treated as one expectation

JonRowe:
Great way to represent it.

Darhazer:
I think there should be an option for `MaxAggregatedExpectations` as even with `aggregate_failures` 10 expectations might be too much.

Also an option may determine if `RSpec/MultipleExpectation` recommends using `aggregate_failures` when there is more then whatever the maximum expectations are set. With such recommendation, it is possible to implement an autocorrect as well.

backus:
This was implemented in https://github.com/backus/rubocop-rspec/pull/329

PR #329, opened 2017-02-25
Count aggregate_failures block as single expectation

Darhazer:
Basic implementation of #233

Let me know if you want me to implement some of the options suggested in the #233

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## Master (Unreleased)
 
 * Add `RSpec/SharedContext` cop. ([@Darhazer][])
+* `Rspec/MultipleExpectations`: Count aggregate_failures block as single expectation. ([@Darhazer][])
 
 ## 1.12.0 (2017-02-21)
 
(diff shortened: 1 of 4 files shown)


backus:
Is it possible to instead redefine the `expect` matcher to `(send _ {:expect :aggregate_failures} ...)`?

backus:
It isn't as simple as modifying the node pattern, but how do you feel about instead altering the behavior of the recursive example search to stop at `aggregate_failures` too? Then we don't have to count and then search again to decrement nodes within `aggregate_failures`?

Darhazer:
Re-implemented. Not sure if this is what you meant though

backus:
Looks great! Thanks @Darhazer

backus:
@Darhazer I really appreciate your recent contributions. Would you be interested in joining as a collaborator? You'd be able to push to the repo directly, I'd ping you for reviews occasionally, and you could help with things like cutting releases.

Darhazer:
@backus I'd be honored

backus:
@Darhazer invited! Thanks for helping out and I'm excited to what you PR next :smile:

PR #330, opened 2017-02-25
Implement SharedContext Cop. Fixes #293

### LINKED ISSUE #293: Recommend shared_examples or shared_context instead depending on contents
I think a nice cop would be something that enforces the following

```ruby
# This should be `shared_examples` since it only defines examples
RSpec.shared_context 'only examples here' do
  it 'does x' do
  end

  it 'does y' do
  end
end

# This should be `shared_context` since it only does setup
RSpec.shared_examples 'only setup here' do
  subject(:blah) { hello }

  let(:foo) { bar }

  before do
    blah
  end
end
```

If setup and examples are both defined inside the definition then the cop should ignore it. Same if it uses any rspec keywords that might include other behavior. Same if it does `include_examples` or `include_context`.

Darhazer:
@backus


lib/rubocop/cop/rspec/shared_context.rb
```diff
@@ -0,0 +1,103 @@
...
+      #
+      # @example
+      #   # bad
+      #   RSpec.shared_examples 'only setup here' do
+      #     subject(:foo) { :bar }
+      #
+      #     let(:baz) { :bazz }
+      #
+      #     before do
+      #       somethind
```
backus: probably should be `something`

spec/rubocop/cop/rspec/shared_context_spec.rb
```diff
@@ -0,0 +1,128 @@
+describe RuboCop::Cop::RSpec::SharedContext do
+  subject(:cop) { described_class.new }
+
+  context 'shared_context' do
+    it 'registers an offense for shared_context with only examples' do
+      expect_violation(<<-RUBY)
+        shared_context 'foo' do
+        ^^^^^^^^^^^^^^^^^^^^ Use `shared_examples` when you don't define context
+          it 'perorms actions' do
```
backus: Typo here. Should be `performs`

lib/rubocop/cop/rspec/shared_context.rb
```diff
@@ -0,0 +1,103 @@
...
+      class SharedContext < Cop
+        MESSAGE_EXAMPLES = "Use `shared_examples` when you don't define context"
+          .freeze
+        MESSAGE_CONTEXT = "Use `shared_context` when you don't define examples"
+          .freeze
+
+        EXAMPLES = (Examples::ALL + Includes::EXAMPLES)
+        def_node_search :examples?, EXAMPLES.send_pattern
+
+        SUBJECT = SelectorSet.new(%i(subject))
```
backus: Would be nice to add this to the Language module

lib/rubocop/cop/rspec/shared_context.rb
```diff
@@ -0,0 +1,103 @@
...
+        def context_with_only_examples(node)
+          _receiver, method, _args = *node.children.first
+          return unless method == :shared_context
+
+          yield if examples?(node) && !context?(node)
+        end
+
+        def examples_with_only_context(node)
+          _receiver, method, _args = *node.children.first
+          return unless %i(shared_examples shared_examples_for).include?(method)
```
backus: If you split out `SharedGroups` like you did for `Includes::EXAMPLES` then you could do something like `SharedGroups::EXAMPLES.include?` here

lib/rubocop/cop/rspec/shared_context.rb
```diff
@@ -0,0 +1,103 @@
...
+              corrector.replace(node.loc.selector, 'shared_context')
+            end
+          end
+        end
+
+        private
+
+        def context_with_only_examples(node)
+          _receiver, method, _args = *node.children.first
+          return unless method == :shared_context
```
backus: Same applies here as https://github.com/backus/rubocop-rspec/pull/330/files#r103094142

backus:
@Darhazer thanks so much for this! Nice job

backus:
Thanks again @Darhazer!

PR #331, opened 2017-02-26
Fix FilePath not flagging non spec files

### LINKED ISSUE #324: Detect specs missing a `_spec` ending in filename
I discovered some specs in a test suite which were not being executed in CI because they didn't end in `_spec.rb`.

I thought the `RSpec/FilePath` would have caught this, but these files were ignored because they are excluded by the filter:

https://github.com/backus/rubocop-rspec/blob/master/config/default.yml#L5

Any thoughts on how this could be handled? I imagine it would need to ignore `spec/support/*.rb`, `spec_helper.rb`, etc.

backus:
Fixes #324


backus:
@Darhazer could you review?

PR #332, opened 2017-02-26
Fix ExpectActual flagging routing specs

### LINKED ISSUE #198: RSpec/ExpectActual false positive against routing specs?
One of our apps has a number of routings specs, that do things like the following (in line with https://www.relishapp.com/rspec/rspec-rails/docs/routing-specs):

```
expect(get: "/some/path").to_not be_routable
expect(get:"/another/path").to route_to(...)
```

The RSpec/ExpectActual cop reports these as offences and I suspect it probably shouldn't:

```
spec/routing/some_spec.rb:6:12: C: RSpec/ExpectActual: Provide the actual you are testing to expect(...)
    expect(get: "/some/path").to route_to(
           ^^^^^^^^^^^^^^^^^
```

For now I've just configured this cop to exclude `spec/routing/**/*`

backus:
Fixes #198

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 * Add `RSpec/SharedContext` cop. ([@Darhazer][])
 * `Rspec/MultipleExpectations`: Count aggregate_failures block as single expectation. ([@Darhazer][])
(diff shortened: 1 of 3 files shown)


backus:
@bquorning could you review?

PR #335, opened 2017-03-03
Add ItBehavesLike cop

dgollahon:
- This cop enforces exclusively using `it_behaves_like` or
  `it_should_behave_like`, instead of allowing both.


lib/rubocop/cop/rspec/it_behaves_like.rb
```diff
@@ -0,0 +1,52 @@
...
+
+        def on_send(node)
+          if example_inclusion_offense?(node)
+            add_offense(node, :expression, message)
+          end
+        end
+
+        def autocorrect(node)
+          lambda do |corrector|
+            if example_inclusion_offense?(node)
```
backus: Do you need this? It will only call the corrector for offenses

spec/rubocop/cop/rspec/it_behaves_like_spec.rb
```diff
@@ -0,0 +1,30 @@
...
+  end
+
+  shared_examples 'an it_behaves_like cop' do |preferred, alternate|
+    let(:enforced_style) { preferred }
+
+    it 'flags a violation for the alternate term' do
+      inclusion = "#{alternate} 'a foo'"
+      expect_violation(<<-RUBY)
+        #{inclusion}
+        #{'^' * inclusion.length} Prefer `#{preferred}` over `#{alternate}` when including examples in a nested context.
```
backus: If you are going to use a shared context then I'd just use `inspect_source`. Grep the other examples for that usage.

I'd actually prefer the duplication of just writing the violation twice in the different contexts

dgollahon: Ok, I wondered if you'd prefer that. I almost did it that way, but then decided to try this and see what you thought.

backus: heh well good prediction

lib/rubocop/cop/rspec/it_behaves_like.rb
dgollahon: Oh, ok. Wasn't sure how it worked.

spec/rubocop/cop/rspec/it_behaves_like_spec.rb
```diff
@@ -0,0 +1,54 @@
...
+      RUBY
+    end
+
+    it 'does not flag a violation for it_behaves_like' do
+      expect_no_violations("it_should_behave_like 'a foo'")
+    end
+
+    include_examples(
+      'autocorrect',
+      "foo(); it_behaves_like 'a foo'",
```
dgollahon: I added `foo();` to kill mutations. Is there another way you'd prefer I do that in case `foo();` seems confusing to someone who doesn't know I'm appeasing `mutest`?

backus:
Looks great! Can you add a changelog entry?

dgollahon:
> Looks great! Can you add a changelog entry?

Sure!

spec/rubocop/cop/rspec/it_behaves_like_spec.rb
backus: Nah this is fine.

PR #336, opened 2017-03-03
Remove extraneous namespace

dgollahon:
- I happened to notice these Rubocop:: prefixes when I added the last
  cop. They are not necessary.


backus:
Thanks!

PR #337, opened 2017-03-03
Improve rspec config

dgollahon:
- Stops RSpec from monkey patching Object
  * Required updating all describes to use the explicit RSpec.describe
    format
- Raises errors on deprecations
- Raises errors instead of warnings for misusing RSpec


backus:
Awesome! I've been wanting to do this for a while

PR #340, opened 2017-03-03
Implement RSpec/IteratedExpectation cop. Fixes #278

### LINKED ISSUE #278: Cop Idea: Prefer using the `all` matcher instead of iterating over an array
The [`all` matcher](https://www.relishapp.com/rspec/rspec-expectations/docs/built-in-matchers/all-matcher) is pretty cool. We can't infer a lot of use cases statically of course but I bet we could probably detect enough cases for it to be useful. For example, I've seen some specs like this

```ruby
it 'validates users' do
  [user1, user2, user3].each { |user| expect(user).to be_valid }
end
```

this could be written as

```ruby
it 'validates users' do
  expect([user1, user2, user3].to all(be_valid)
end
```

Darhazer:
@backus 

Please look at the examples and let me know if you can think of other examples that we can detect, or some possible false positives.

TODO:
* [ ] Implement autocorrect
* [x] Multiple expectations can also be converted to `all` with an `and`


spec/rubocop/cop/rspec/each_spec.rb
```diff
@@ -0,0 +1,80 @@
...
+        end
+      end
+    RUBY
+  end
+
+  it 'ignore `each` when the body does not contain only expectations' do
+    expect_no_violations(<<-RUBY)
+      it 'validates users' do
+        [user1, user2, user3].each do |user|
+          allow(Somethig).to receive(:method).and_return(user)
```
bquorning: Somethig -> Something

spec/rubocop/cop/rspec/each_spec.rb
```diff
@@ -0,0 +1,80 @@
...
+        [user1, user2, user3].each { |user| expect(user).to be_an(User) }
+        ^^^^^^^^^^^^^^^^^^^^^^^^^^ Prefer using the `all` matcher instead of iterating over an array.
+      end
+    RUBY
+  end
+
+  it 'ignores `each` with proc' do
+    expect_no_violations(<<-RUBY)
+      it 'validates users' do
+        [user1, user2, user3].each(&:to_s)
```
bquorning: Since this proc also doesn’t set an expectation, isn’t this case covered in the following spec (line 30)?

spec/rubocop/cop/rspec/each_spec.rb
```diff
@@ -0,0 +1,80 @@
...
+        [user1, user2, user3].each { |user| expect(user).to be_valid }
+        ^^^^^^^^^^^^^^^^^^^^^^^^^^ Prefer using the `all` matcher instead of iterating over an array.
+      end
+    RUBY
+  end
+
+  it 'flags `each` when expectation calls method with arguments' do
+    expect_violation(<<-RUBY)
+      it 'validates users' do
+        [user1, user2, user3].each { |user| expect(user).to be_an(User) }
```
bquorning: be_an -> be_a

lib/rubocop/cop/rspec/each.rb
```diff
@@ -0,0 +1,53 @@
...
+      #   # bad
+      #   it 'validates users' do
+      #     [user1, user2, user3].each { |user| expect(user).to be_valid }
+      #   end
+      #
+      #   # good
+      #   it 'validates users' do
+      #     expect([user1, user2, user3]).to all(be_valid)
+      #   end
+      class Each < Cop
```
bquorning: I agree that `Each` is too generic a name. `IteratedExpectation` is the best I can think of right now, and I don’t really like that either. Perhaps @backus can think of a better name?

lib/rubocop/cop/rspec/each.rb
```diff
@@ -0,0 +1,53 @@
...
+        PATTERN
+
+        def_node_matcher :expectation?, <<-PATTERN
+          (send (send nil :expect (lvar %)) :to ...)
+        PATTERN
+
+        def on_block(node)
+          each?(node) do |arg, body|
+            if single_expectation?(body, arg) || only_expectations?(body, arg)
+              @arg = arg
```
bquorning: What does setting `@arg` do?

lib/rubocop/cop/rspec/each.rb
backus: Maybe `AllMatcher`? I usually like looking at the rspec internal naming but it isn't super helpful in this case heh https://github.com/rspec/rspec-expectations/blob/82ab90b513dfcd0b52064f56be4b2c6218adeded/lib/rspec/matchers.rb#L638-L655

dgollahon: I think `IteratedExpectation` is actually a pretty reasonable name. Since we may possibly have cops in the future relating to the `all` matcher, `AllMatcher` may be too general also.

backus: If our intention is to have multiple cops that recommend the `all(...)` matched then I agree

dgollahon: Yeah, i just think it's definitely possible if not likely. There's already a few cops (for instance the `ItBehavesLike` cop i just added...) that cover lots of potential recommendations (who's to say we won't have more corrections for using `it_behaves_like` in the future) so I think we should start being more careful about how we name new cops.

At the risk of going slightly off topic, there's also https://github.com/backus/rubocop-rspec/issues/232 to consider at some point. Idk if I'm sold on it exactly, but another area where this was slightly awkward recently was adding a new feature to the `ExampleWording` cop. It didn't make sense to add a second wording cop but the two features inside there now would probably be in separate cops and individually configurable, in an ideal world.

lib/rubocop/cop/rspec/each.rb
Darhazer: left over from some initial attempts to implement autocorrect. Its not needed, as arg can be retrieved from the node itself, but I have dropped the autocorrect support completely. I'll clean up

Darhazer:
PR updated

backus:
Looks great! Thanks @Darhazer

PR #341, opened 2017-03-03
Improve AroundBlock cop

dgollahon:
- Significantly improves test coverage. This cop is now 100% mutation
  tested.
- Covers additional around scopes such as `:example` or `:context`.
- Allows string versions of scopes such as `'example'`.
- Checks that `#run` or `#call` is called on the test object, not just that
  the test argument was referenced.
- Refactors a bit of naming and test descriptions to be (hopefully) a
  bit clearer.

### DIFF
--- a/lib/rubocop/cop/rspec/around_block.rb
+++ b/lib/rubocop/cop/rspec/around_block.rb
@@ -24,53 +24,42 @@ module RuboCop
       #     test.run
       #   end
       class AroundBlock < Cop
-        MSG_NO_ARG = 'Test object should be passed to around block'.freeze
+        MSG_NO_ARG     = 'Test object should be passed to around block'.freeze
         MSG_UNUSED_ARG = 'You should call `%<arg>s.call` ' \
-                          'or `%<arg>s.run`'.freeze
+                         'or `%<arg>s.run`'.freeze
 
-        def_node_matcher :scoped_hook, <<-PATTERN
-        (block (send nil :around (sym {:each :example})) $(args ...) ...)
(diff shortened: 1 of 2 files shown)


Darhazer:
Good job! Thank you!

I was not aware that `around` can be used with string. IMO rubocop-rspec should discourage such usage (which requires a new cop)

dgollahon:
> I was not aware that around can be used with string. IMO rubocop-rspec should discourage such usage (which requires a new cop)

Yeah, I was thinking that would be a good cop to add. I didn't know either, but I decided to try it out and... you can! o_O

dgollahon:
@backus: Based on our earlier discussion today, I tried to account for cases when the test might be called outside of the around block. What do you think of the newest implementation? If you agree, I'll squash it down and can merge it. Otherwise, I can revert to simpler behavior.

dgollahon:
On a related note, I discovered that you can do `test_proxy.call(any, number, of, args)` which is surprising. I couldn't tell what it did, if anything. My current implementation will not recognize those cases and flag them because of the way the node matcher works (which assumes the last part of the `send` is `:call` or `:run`). Seemed too obscure to fix, but it's sort of interesting in general.

dgollahon:
@Darhazer, @backus: Actually I realized I was wrong about the string qualifiers! `rspec` actually accepts any garbage in hook methods. You can give it `nil` or `Object.new` or whatever you want and it seems to be equivalent to `around {}`. It will only validate symbols (so it will at least complain about something like `around(:garbage)`.

I'll undo my support for strings in a moment.

PR #342, opened 2017-03-03
Take into account aggregate_failures meta data

Darhazer:
@backus


backus:
Very nice! Could you rebase and add a changelog entry?

Darhazer:
Rebased and moved the check if aggregate_failures meta data is true to the node matcher.
As for the changelog entry, this is an addition to the still-not-released functionality, I don't see need for a separate changelog entry

backus:
Good point regarding the changelog. Thanks for the this!

PR #343, opened 2017-03-03
Require empty lines after subject and latest let

Darhazer:


backus:
Nice! A few comments:

* I love this kind of cop and I've really been wanting something like this. I think this cop should add autocorrect though before we merge it. My private codebase I ran it on is pretty polished and it found 66 offenses. Manually correcting a bunch of newlines is a pain

* I think I found some false positives. Try testing this case:
    ```ruby
    RSpec.shared_context 'foo' do
      let(:bar) { baz }
      let(:baz) { qux }
    end

    RSpec.describe Foo do
      it_beahves_like 'x' do
        let(:bar) { baz }
        let(:baz) { qux }
      end
    end
    ```

bquorning:
I have identical observations. Ran the cops on some 1200 files, got 105 offenses. A few of them were false positives where `let` is the last line in a block.

Darhazer:
Added autocorrect
Fixed the false positives
Added changelog entries

spec/rubocop/cop/rspec/empty_line_after_let_spec.rb
```diff
@@ -0,0 +1,110 @@
...
+
+  it 'ignores empty lines between the lets' do
+    expect_violation(<<-RUBY)
+      RSpec.describe User do
+        let(:a) { a }
+
+        subject { described_class }
+
+        let!(:b) { b }
+        ^^^^^^^^^^^^^^ Add empty line after the last `let` block.
```
bquorning: “… and group all `let` lines together” – but that’s for another cop :-)

lib/rubocop/cop/rspec/empty_line_after_let.rb
```diff
@@ -0,0 +1,64 @@
...
+      #
+      #   # good
+      #     let(:foo) { bar }
+      #     let(:something) { other }
+      #
+      #     it do
+      #       ...
+      #     end
+      class EmptyLineAfterLet < Cop
+        MSG = 'Add empty line after the last `let` block.'.freeze
```
backus: Can you change this to `"Add an empty line"` so it uses proper grammar?

backus:
I think we should probably rename `EmptyLineAfterLet` to `EmptyLineAfterFinalLet`

backus:
Wow. I ran this cop on a few codebases I work with an it did a great job! This is the kind of cop I've wanted for a long time I'm glad you put in the time and energy :smile:

backus:
@Darhazer I have another small bug for you:

```
$ cat ex_spec.rb
RSpec.describe 'silly heredoc syntax' do
  let(:foo) { <<-BAR }
  hello
  world
  BAR

  it 'has tricky syntax' do
    expect(foo).to eql("  hello\n  world\n")
  end
end

$ bundle exec rubocop --require rubocop-rspec --only RSpec/EmptyLineAfterLet ex_spec.rb
Inspecting 1 file
C

Offenses:

ex_spec.rb:2:3: C: RSpec/EmptyLineAfterLet: Add empty line after the last let block.
  let(:foo) { <<-BAR }
  ^^^^^^^^^^^^^^^^^^^^

-> [1]

$ bundle exec rubocop --require rubocop-rspec --only RSpec/EmptyLineAfterLet -a ex_spec.rb
Inspecting 1 file
C

Offenses:

ex_spec.rb:2:3: C: [Corrected] RSpec/EmptyLineAfterLet: Add empty line after the last let block.
  let(:foo) { <<-BAR }
  ^^^^^^^^^^^^^^^^^^^^

1 file inspected, 1 offense detected, 1 offense corrected

$ cat ex_spec.rb
RSpec.describe 'silly heredoc syntax' do
  let(:foo) { <<-BAR }

  hello
  world
  BAR

  it 'has tricky syntax' do
    expect(foo).to eql("  hello\n  world\n")
  end
end
```

I found this while running your cops on the main rubocop codebase

backus:
@Darhazer I'm not sure if there are others but I recommend checking by running this on the rubocop codebase when you fix this bug. It is a good way to fuzz your implementation and it also can become a very easy and fun PR you can offer up to the rubocop team when you run autocorrect :smile:

spec/rubocop/cop/rspec/empty_line_after_let_spec.rb
Darhazer: Nice idea. I will work on this next ;)

Darhazer:
PR updated

Darhazer:
Tested on rubocop and one more code base for false positives.
Is there anything else to be added/changed here?

CHANGELOG.md
```diff
@@ -2,6 +2,9 @@
 
 ## Master (Unreleased)
 
+* Add `RSpec/EmptyLineAfterSubject` cop. ([@Darhazer][])
+* Add `RSpec/EmptyLineAfterLet` cop. ([@Darhazer][])
```
backus: This should probably also say `EmptyLineAfterFinalLet`

backus:
@Darhazer these cops are great! Sorry it took me a bit longer to get to the final reviews. If you could just update the cop name in the changelog you can push and then merge assuming CI passes :smile:

backus:
Great work @Darhazer!

PR #348, opened 2017-03-04
Improve DescribeMethod cop

dgollahon:
Tiny refactor.

- Slightly strengthens the regex that checks for #/. prefixed method
  names.
- Uses the convenient str_type? method.

### DIFF
--- a/lib/rubocop/cop/rspec/describe_method.rb
+++ b/lib/rubocop/cop/rspec/describe_method.rb
@@ -22,10 +22,10 @@ module RuboCop
 
         MESSAGE = 'The second argument to describe should be the method ' \
                   "being tested. '#instance' or '.class'".freeze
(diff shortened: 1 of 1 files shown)


backus:
You updated the regex which is good. We could also actually just do `one(second_arg.children).start_with?('.', '#')` which is nice. Thoughts?

dgollahon:
> You updated the regex which is good. We could also actually just do one(second_arg.children).start_with?('.', '#') which is nice. Thoughts?

I think it's probably better to leave as is because this way we at least guarantee there's a method name involved of some kind. So `#s` would be valid, but `#` would not be. I could also change it to just `.` instead of `.+` but I thought the `+` might make the intention a bit clearer.

backus:
Heh I didn't think of the case where someone might write `describe Foo, '#'`... alright!

PR #351, opened 2017-03-04
Refactor ExampleWording cop

dgollahon:
- Replace manual destructuring with a much simpler node matcher.
- Breaks offense logic into a helper function.


backus:
This looks great

PR #355, opened 2017-03-04
Refactor ItBehavesLike cop

dgollahon:
- Uses the node matcher instead of direct destructuring.

_refactors thing he wrote yesterday_

Node matching is neat!


backus:
> Node matching is neat!

I agree!

PR #361, opened 2017-03-05
Standardize cop messages

dgollahon:
This tries to bring a bit more consistency to the cop messages. Another
pass should be done to improve the wording, but this is something.

- All messages now end in a period, `.`.
- ALl cops with a single message constant are named MSG (I chose MSG
  over MESSAGE because it's what rubocop expects).
- Adjusted some message alignment and chose a consistent line
  continuation pattern of not having an extra space before `\`.
- Removed explicit MSG arguments where they were no longer needed.
- Lifted one message into a frozen MSG constant.
- Verified much of the above was correct with a script.

### DIFF
--- a/lib/rubocop/cop/rspec/any_instance.rb
+++ b/lib/rubocop/cop/rspec/any_instance.rb
@@ -21,7 +21,7 @@ module RuboCop
       #     end
       #   end
       class AnyInstance < Cop
-        MESSAGE = 'Avoid stubbing using `%{method}`'.freeze
+        MSG = 'Avoid stubbing using `%{method}`.'.freeze
 
         def_node_matcher :disallowed_stub, <<-PATTERN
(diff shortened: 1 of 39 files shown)


dgollahon:
Quick and dirty script I used to check my work:

```
RuboCop::Cop::RSpec.constants.flat_map do |cop_name|
  cop = RuboCop::Cop::RSpec.const_get(cop_name)
  cop.constants
    .grep(/\A(MSG|MESSAGE)/)
    .map { |msg_const| ["#{cop}::#{msg_const}", cop.const_get(msg_const)] }
end.each do |const, str|
  puts "#{const} did not end in '.'" unless str =~ /\.\z/
  puts "#{const} should use MSG instead of MESSAGE '.'" if const =~ /MESSAGE/
end
```

dgollahon:
:)

PR #364, opened 2017-03-05
Refactor FilePath cop

dgollahon:
Converts specs to use `expect_violation` as well as refactors the cop itself in the following ways:

- Use `File.fnmatch?` instead of generating a file matching regex.
- Leverage node mappers instead of manual destructuring.
- Use `#fetch` with defaults instead of `a[b] || c`.
- Reorder methods so they appear in roughly the order they are used.
- Rename various methods and variables to improve readability.

### DIFF
--- a/lib/rubocop/cop/rspec/file_path.rb
+++ b/lib/rubocop/cop/rspec/file_path.rb
@@ -44,70 +44,67 @@ module RuboCop
       class FilePath < Cop
         include RuboCop::RSpec::TopLevelDescribe
 
-        MSG          = 'Spec path should end with `%s`.'.freeze
-        ROUTING_PAIR = s(:pair, s(:sym, :type), s(:sym, :routing))
+        MSG = 'Spec path should end with `%s`.'.freeze
+
+        def_node_search :const_described?,  '(send _ :describe (const ...) ...)'
+        def_node_search :routing_metadata?, '(pair (sym :type) (sym :routing))'
 
         def on_top_level_describe(node, args)
+          return unless const_described?(node) && single_top_level_describe?
           return if routing_spec?(args)
 
-          return unless single_top_level_describe?
-          object = args.first.const_name
-          return unless object
+          glob = glob_for(args)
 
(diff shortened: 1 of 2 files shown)


lib/rubocop/cop/rspec/file_path.rb
```diff
@@ -45,69 +45,66 @@ class FilePath < Cop
...
         end
 
         def ignore_methods?
           cop_config['IgnoreMethods']
         end
 
-        def custom_transform
-          cop_config['CustomTransform'] || {}
+        def filename_ends_with?(glob)
+          File.fnmatch?("*#{glob}", processed_source.buffer.name)
```
dgollahon: We could actually simplify this slightly by making `glob_for` return a leading `*` but  that changes the error message too, so Idk if that's maybe a bit noisy (or possibly helpful?) to display.

lib/rubocop/cop/rspec/file_path.rb
```diff
@@ -45,69 +45,66 @@ class FilePath < Cop
...
 
-        def matcher(object, method)
-          path = File.join(parts(object))
-          if method && method.type.equal?(:str) && !ignore_methods?
-            path += '*' + method.str_content.gsub(/\W+/, '')
+        def glob_for((described_class, method))
+          path = path_for(described_class)
+
+          if method && method.str_type? && !ignore_methods?
+            path += '*' + method.str_content.gsub(/\W/, '')
```
dgollahon: Removed the `+` from the previous `\W+` because it's equivalent with a `#gsub`.

lib/rubocop/cop/rspec/file_path.rb
dgollahon: Oddly, not having the '*' passed all the specs, but then broke when running on itself. I added https://github.com/backus/rubocop-rspec/pull/364/commits/2061189f599e8cd11eb913e39db15c1bc97f3944#diff-b856527dd5957e11027e754a894e1b09R80 to ensure the behavior is preserved.

PR #366, opened 2017-03-06
Flag repeated 'it' in `it` docstrings

dgollahon:
- Checks for things like `it 'it does something' do; end'
- Actually fixed a few instances in our own codebase, which is
  satisfying!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Master (Unreleased)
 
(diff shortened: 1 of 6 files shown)


dgollahon:
Another option would be to make this a separate cop since the existing config options only apply to the `should` part of the behavior. But we seem to have reserved a lot of broad cop names like `ExampleWording`.

PR #368, opened 2017-03-06
Consistently add 2 space indentation to HEREDOCs

bquorning:
It seems we have an inconsistent choice of style 😱  regarding indentation of HEREDOCs. Because the spacing inside HEREDOCs is *actually* spaces in a string, I don’t think we can check them using, well, Rubocop.

In e.g. Ruby code and AST patterns, spaces don’t matter, so we can add or remove indentation as we like. I prefer 2 extra spaces of indentation, but won’t force anything on anyone. Except for consistency, of course.

### DIFF
--- a/lib/rubocop/cop/rspec/describe_class.rb
+++ b/lib/rubocop/cop/rspec/describe_class.rb
@@ -23,20 +23,20 @@ module RuboCop
               'the class or module being tested.'.freeze
 
         def_node_matcher :valid_describe?, <<-PATTERN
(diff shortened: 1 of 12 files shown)


backus:
Nice. @pocke did actually add a cop recently for heredoc indentation that isn't released yet. I wish we could just use `<<~` but we support rubies lower than 2.3.

Thanks for cleaning this up!

bquorning:
> @pocke did actually add a cop recently for heredoc indentation that isn't released yet.

Interesting. I hadn’t noticed that change.

dgollahon:
Nice! This was actually bugging me also.

PR #369, opened 2017-03-06
Accept :aggregate_failures as aggregate_failures: true

Darhazer:
@backus 
The original fix didn't account that you can specify `it "something", :aggregate_failures` which default to true. The boolean option is actually to overwrite the value when you specify globally that failures have to be aggregated.


backus:
Nice fix! Thanks

PR #371, opened 2017-03-09
Add RSpec/OverwritingSetup cop. Fixes #328

### LINKED ISSUE #328: Detecting overriding same let in same rspec scope
Avoid copy-paste / grammar errors by detecting overriding same rspec `let` / `let!` / `subject`, e.g.

```ruby
let(:a1) { 1 }
let(:a1) { 2 }
```

--------

## Expected behavior

This should create an offense:

```ruby
let(:a1) { 1 }
let(:a1) { 2 }
```

Correct code could be:

```ruby
let(:a1) { 1 }
let(:a2) { 2 }
```

## Actual behavior

It doesn't create an offense

## RuboCop version

```
0.47.1 (using Parser 2.3.3.1, running on ruby 2.3.3 x86_64-darwin16)
```

(mostly just a memo for myself to create this cop, originally filed there: https://github.com/bbatsov/rubocop/issues/4057)

Darhazer:


bquorning:
Nitpick: There’s a typo in the PR title.

lib/rubocop/cop/rspec/overwriting_setup.rb
```diff
@@ -0,0 +1,55 @@
...
+
+          find_duplicates(body) do |duplicate, name|
+            add_offense(duplicate, :expression, format(MSG, name: name))
+          end
+        end
+
+        private
+
+        def find_duplicates(node)
+          already_defined = []
```
bquorning: Could we name this collector after its contents? Perhaps `setup_expressions = []`?

lib/rubocop/cop/rspec/overwriting_setup.rb
```diff
@@ -0,0 +1,55 @@
...
+
+          find_duplicates(body) do |duplicate, name|
+            add_offense(duplicate, :expression, format(MSG, name: name))
+          end
+        end
+
+        private
+
+        def find_duplicates(node)
+          setup_expressions = []
```
dgollahon: Not that I imagine it would influence the performance much either way, but `Set.new` here might be more appropriate.

(idk if it actually leads to better code / if you want to use it, but there's also an `#add?` method on `Set` which returns nil if the element was already in the set so you could do `yield child, name unless setup_expressions.add?(name)`.)

spec/rubocop/cop/rspec/overwriting_setup_spec.rb
```diff
@@ -0,0 +1,46 @@
...
+    RUBY
+  end
+
+  it 'finds overwriten `subject`' do
+    expect_violation(<<-RUBY)
+      RSpec.describe User do
+        subject(:a) { a }
+
+        let(:a) { b }
+        ^^^^^^^^^^^^^ `a` is already defined.
```
dgollahon: This is pretty cool! Neat that it checks for overriding `subject`s and `let`s and vice versa rather than just one of each kind.

backus:
Since it looks like @bquorning and @dgollahon both already approved this I'm going to merge. Thanks @Darhazer!

dgollahon:
Oh, I just realized: this needs a changelog entry.

backus:
lol. Can you add that @dgollahon?

dgollahon:
@backus: Sure.

PR #373, opened 2017-03-18
Leading subject autocorrect

Darhazer:


lib/rubocop/cop/rspec/leading_subject.rb
```diff
@@ -37,14 +37,37 @@ def on_block(node)
...
               break add_offense(node, :expression)
             end
           end
         end
 
+        def autocorrect(node)
+          lambda do |corrector|
+            first_let = find_first_let(node)
+            first_let_position = first_let.loc.expression
+            ident = "\n" + ' ' * first_let.loc.column
```
backus: This should probably say `indent`

lib/rubocop/cop/rspec/leading_subject.rb
```diff
@@ -37,14 +37,37 @@ def on_block(node)
           node.parent.each_child_node do |sibling|
             break if sibling.equal?(node)
 
-            if sibling.method_name.equal?(:let)
+            if [:let, :let!].include?(sibling.method_name)
```
backus: This also shows up on line 60. Can you extract a method?

lib/rubocop/cop/rspec/leading_subject.rb
```diff
@@ -37,14 +37,37 @@ def on_block(node)
...
 
+        def find_first_let(node)
+          node.parent.each_child_node do |sibling|
+            return sibling if [:let, :let!].include?(sibling.method_name)
+          end
+        end
+
+        def node_range(node)
+          range = node.source_range
+          range = range_with_surrounding_space(range, :left, false)
```
backus: Neat! I didn't know about this method

Darhazer:
PR updated. Autocorrect is tested against the rubocop codebase.

btw it doesn't leave empty line after the new subject location, as there is another cop for requiring the empty line

backus:
> btw it doesn't leave empty line after the new subject location, as there is another cop for requiring the empty line

Good. It isn't the concern of this cop I agree

backus:
Another great cop, @Darhazer :smile:

PR #374, opened 2017-03-18
Add RSpec/ScatteredLet cop

Darhazer:
@bquorning I have promised you this one ;)

btw it seems related to ScatteredSetup - @backus what do you think about combining them?


lib/rubocop/cop/rspec/scattered_let.rb
```diff
@@ -0,0 +1,58 @@
...
+      #   describe Foo do
+      #     subject { Foo }
+      #     before { prepare }
+      #     let(:foo) { 1 }
+      #     let(:bar) { 2 }
+      #     let!(:baz) { 3 }
+      #   end
+      #
+      class ScatteredLet < Cop
+        MSG = 'Group all let/let! blocks in the example group togetger.'.freeze
```
backus: together

backus:
@Darhazer they are similar so I am glad you also used the word `Scattered` but this should be its own cop I think.

Darhazer:
Fixed the typo

backus:
🎈

PR #375, opened 2017-03-19
Fix suggestion when the argument is a hash. Fixes #267

### LINKED ISSUE #267: Bug in SingleArgumentMessageChain
If you (for some strange reason) are calling `receive_message_chain` with a hash, you should substitute it with `receive_messages` instead of just `receive`. E.g.:

```rb
allow(foo).to receive_message_chain(bar: 42, baz: 42)
```
is suggested by the cop to be replaced with
```rb
allow(foo).to receive(bar: 42, baz: 42)
```
while the actually working code would be
```rb
allow(foo).to receive_messages(bar: 42, baz: 42)
```

Darhazer:
@bquorning

### DIFF
--- a/lib/rubocop/cop/rspec/single_argument_message_chain.rb
+++ b/lib/rubocop/cop/rspec/single_argument_message_chain.rb
@@ -19,12 +19,16 @@ module RuboCop
               '`%<called>s` with a single argument.'.freeze
 
         def_node_matcher :message_chain, <<-PATTERN
-          (send _ #{Matchers::MESSAGE_CHAIN.node_pattern_union} $...)
+          (send _ #{Matchers::MESSAGE_CHAIN.node_pattern_union} $_)
         PATTERN
 
+        def_node_matcher :single_key_hash?, '(hash pair)'
+
         def on_send(node)
-          message_chain(node) do |(first, *remaining)|
-            return if first.to_s.include?('.') || remaining.any?
+          message_chain(node) do |arg|
+            return if arg.to_s.include?('.')
+
+            return if arg.hash_type? && !single_key_hash?(arg)
 
             add_offense(node, :selector)
           end
@@ -33,10 +37,26 @@ module RuboCop
         def autocorrect(node)
           lambda do |corrector|
             corrector.replace(node.loc.selector, replacement(node.method_name))
+            message_chain(node) do |arg|
+              autocorrect_hash_arg(corrector, arg) if single_key_hash?(arg)
+            end
           end
         end
 
         private
+
+        def autocorrect_hash_arg(corrector, arg)
+          key, value = *arg.children.first
+
+          corrector.replace(arg.loc.expression, key_to_arg(key))
+          corrector.insert_after(arg.parent.loc.end,
+                                 ".and_return(#{value.source})")
+        end
+
+        def key_to_arg(node)
+          key, = *node
+          node.sym_type? ? ":#{key}" : node.source
+        end
 
         def replacement(method)
           method.equal?(:receive_message_chain) ? 'receive' : 'stub'
--- a/spec/rubocop/cop/rspec/single_argument_message_chain_spec.rb
+++ b/spec/rubocop/cop/rspec/single_argument_message_chain_spec.rb
@@ -47,6 +47,45 @@ RSpec.describe RuboCop::Cop::RSpec::SingleArgumentMessageChain do
         end
       RUBY
     end
+
+    context 'with single-key hash argument' do
+      it 'reports an offence' do
+        expect_violation(<<-RUBY)
+          before do
+            allow(foo).to receive_message_chain(bar: 42)
+                          ^^^^^^^^^^^^^^^^^^^^^ Use `receive` instead of calling `receive_message_chain` with a single argument.
+          end
+        RUBY
+      end
+
+      include_examples(
+        'autocorrect',
+        'before { allow(foo).to receive_message_chain(bar: 42) }',
+        'before { allow(foo).to receive(:bar).and_return(42) }'
+      )
+
+      include_examples(
+        'autocorrect',
+        'before { allow(foo).to receive_message_chain("bar" => 42) }',
+        'before { allow(foo).to receive("bar").and_return(42) }'
+      )
+
+      include_examples(
+        'autocorrect',
+        'before { allow(foo).to receive_message_chain(:"#{foo}" => 42) }',
+        'before { allow(foo).to receive(:"#{foo}").and_return(42) }'
+      )
+    end
+
+    context 'with multiple keys hash argument' do
+      it "doesn't report an offence" do
+        expect_no_violations(<<-RUBY)
+          before do
+            allow(foo).to receive_message_chain(bar: 42, baz: 42)
+          end
+        RUBY
+      end
+    end
   end
 
   describe 'stub_chain' do
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Master (Unreleased)
 
+* Fix false positive in `RSpec/SingleArgumentMessageChain` cop when the single argument is a hash. ([@Darhazer][])
+
 ## 1.15.0 (2017-05-24)
 
 * Add `RSpec/DescribeSymbol` cop. ([@tsigo][])


backus:
@Darhazer I'm not sure these are equivalent?

```
>> require 'rspec/mocks/standalone'
=> true
>> dbl = double(:foo)
=> #<Double :foo>
>> allow(dbl).to receive_message_chain(foo: 1, bar: 2)
=> []
>> dbl.foo
=> #<Double (anonymous)>
>> dbl.foo.bar
=> 2
>> dbl.bar
RSpec::Mocks::MockExpectationError: #<Double :foo> received unexpected message :bar with (no args)
from /Users/johnbackus/.gem/ruby/2.4.0/gems/rspec-support-3.6.0.beta2/lib/rspec/support.rb:87:in 'block in <module:Support>'
>> dbl2 = double(:foo)
=> #<Double :foo>
>> allow(dbl2).to receive_messages(foo: 1, bar: 2)
>> dbl2.foo
=> 1
>> dbl2.bar
=> 2
>> dbl2.foo.bar
NoMethodError: undefined method `bar' for 1:Integer
from (pry):13:in `__pry__'
```

Darhazer:
Then #267 seems invalid

backus:
Yeah. @bquorning can you confirm?

bquorning:
I stand corrected, my suggestion in #267 was incorrect. But the cop is still incorrect in suggesting 
`receive` as a substitute for `receive_message_chain` in
```rb
allow(foo).to receive_message_chain(bar: 42, baz: 42)
```

When a singular argument to `receive_message_chain` is a hash, `RSpec/SingleArgumentMessageChain` should report no offense.

Darhazer:
If that's the case, it's even simpler to fix and I updated the PR.

btw, what is the reason for `first.to_s.include?('.')` check?

bquorning:
> btw, what is the reason for `first.to_s.include?('.')` check?

That’s to allow `allow(foo).to receive_message_chain("one.two") { :three }`. It’s not a syntax that I’d use, but I guess we have to support it.

bquorning:
You should probably add an entry to the changelog.

Darhazer:
Changelog entry added.

@bquorning should we handle the case when the argument is a hash with single key, like
`allow(foo).to receive_message_chain(bar: 42)`

bquorning:
> should we handle the case when the argument is a hash with single key

That sounds like a good idea. These two are equivalent:
```rb
allow(foo).to receive_message_chain(bar: 42)
allow(foo).to receive(:bar) { 42 }
```

dgollahon:
> That sounds like a good idea. These two are equivalent:
>
> ```
> allow(foo).to receive_message_chain(bar: 42)
> allow(foo).to receive(:bar) { 42 }
> ```

```ruby
allow(foo).to receive(:bar) { 42 }
```

is also equivalent to

```ruby
allow(foo).to receive(:bar).and_return(42)
```

which I think is more intention revealing / nicer if we're going to correct it.

For that matter, it'd probably make a decent cop (though correcting from `receive(foo) { bar }` to `receive(foo).and_return(bar)` is not always truly equivalent due to evaluation order).

Darhazer:
PR updated

spec/rubocop/cop/rspec/single_argument_message_chain_spec.rb
```diff
@@ -47,6 +47,49 @@
...
+          end
+        RUBY
+      end
+    end
+
+    context 'with empty hash' do
+      it "doesn't report an offence" do
+        expect_no_violations(<<-RUBY)
+          before do
+            allow(foo).to receive_message_chain({})
```
bquorning: I’m not sure what `receive_message_chain({})` does. Intuitively, I’d think it doesn’t do anything.

spec/rubocop/cop/rspec/single_argument_message_chain_spec.rb
```diff
@@ -47,6 +47,49 @@
...
+      include_examples(
+        'autocorrect',
+        'before { allow(foo).to receive_message_chain(bar: 42) }',
+        'before { allow(foo).to receive(:bar) { 42 } }'
+      )
+
+      include_examples(
+        'autocorrect',
+        'before { allow(foo).to receive_message_chain("foo bar" => 42) }',
+        'before { allow(foo).to receive("foo bar") { 42 } }'
```
bquorning: Can we do that, with a string including whitespace?

Darhazer: lol, I was so concerned to handle all hash cases that I forgot they are used in a `receive` call. Guess all this safety is unnecessary

Darhazer:
PR updated

spec/rubocop/cop/rspec/single_argument_message_chain_spec.rb
```diff
@@ -47,6 +47,39 @@
...
+      include_examples(
+        'autocorrect',
+        'before { allow(foo).to receive_message_chain(bar: 42) }',
+        'before { allow(foo).to receive(:bar) { 42 } }'
+      )
+
+      include_examples(
+        'autocorrect',
+        'before { allow(foo).to receive_message_chain("bar" => 42) }',
+        'before { allow(foo).to receive(:bar) { 42 } }'
```
dgollahon: I wonder if we should preserve the key type.

lib/rubocop/cop/rspec/single_argument_message_chain.rb
```diff
@@ -22,22 +22,42 @@ class SingleArgumentMessageChain < Cop
           (send _ #{Matchers::MESSAGE_CHAIN.node_pattern_union} $...)
         PATTERN
 
+        def_node_matcher :hash?, '(hash ...)'
+
         def on_send(node)
           message_chain(node) do |(first, *remaining)|
             return if first.to_s.include?('.') || remaining.any?
 
+            return if hash?(first) && first.children.count != 1
```
dgollahon: You could, alternatively, just match a hash with a single pair. I think it might just be `'(hash pair)'`

lib/rubocop/cop/rspec/single_argument_message_chain.rb
```diff
@@ -22,22 +22,42 @@ class SingleArgumentMessageChain < Cop
           (send _ #{Matchers::MESSAGE_CHAIN.node_pattern_union} $...)
         PATTERN
 
+        def_node_matcher :hash?, '(hash ...)'
+
         def on_send(node)
           message_chain(node) do |(first, *remaining)|
             return if first.to_s.include?('.') || remaining.any?
```
dgollahon: I think you could also just match a single argument in the node pattern instead of having to explicitly check if `remaining` is empty.

dgollahon: Though i guess that code wasn't introduced in this PR.

spec/rubocop/cop/rspec/single_argument_message_chain_spec.rb
```diff
@@ -47,6 +47,39 @@
...
+            allow(foo).to receive_message_chain(bar: 42)
+                          ^^^^^^^^^^^^^^^^^^^^^ Use `receive` instead of calling `receive_message_chain` with a single argument.
+          end
+        RUBY
+      end
+
+      include_examples(
+        'autocorrect',
+        'before { allow(foo).to receive_message_chain(bar: 42) }',
+        'before { allow(foo).to receive(:bar) { 42 } }'
```
dgollahon: Do you prefer this syntax to `.and_return(42)`? If so, just curious as to why (i'm not asking you to change anything, just wondering).

lib/rubocop/cop/rspec/single_argument_message_chain.rb
dgollahon: You could also capture the key and the value in this expression to make the autocorrect clearer, if you wanted.

spec/rubocop/cop/rspec/single_argument_message_chain_spec.rb
dgollahon: If so, it would actually make https://github.com/backus/rubocop-rspec/pull/375/files#diff-91336a4d8c49f0937c8d89c6ea457d1cR58 simpler by just being `#to_s`, i think.

Darhazer:
Thank you for the review @dgollahon 
I'll update it soon

spec/rubocop/cop/rspec/single_argument_message_chain_spec.rb
Darhazer: Actually I prefer the `.and_return`. I left the block one just for consistency with the other test cases. On the other hand `and_return` is closer to the original code (no lazy evaluation), so I will replace it with `and_return`

Darhazer:
@dgollahon I couldn't actually apply most of your advice, but made the node matchers to match only single arguments / hash with single key; and also changed the autocorrect to `and_return`

dgollahon:
@Darhazer: Alrighty. I'll take a glance after work 👍

lib/rubocop/cop/rspec/single_argument_message_chain.rb
```diff
@@ -33,11 +37,27 @@ def on_send(node)
...
+          key, value = *arg.children.first
+
+          corrector.replace(arg.loc.expression, key_to_arg(key))
+          corrector.insert_after(arg.parent.loc.end,
+                                 ".and_return(#{value.source})")
+        end
+
+        def key_to_arg(node)
+          key, = *node
+          ":#{key.to_sym}"
```
dgollahon: Pretty sure this `#to_sym` is unnecessary.

What do you think about

`node.sym_type? ? ":#{key}" : node.source`

to preserve the string or symbol type of keys? I think it would be better if the autocorrect were as true to the original as possible and the choice between string and symbol keys could be left to another cop.

dgollahon: At least, I don't think there's any case where `#to_sym` and then an implicit `#to_s` is different than just the implicit `#to_s`.

I've realized the code as it stands actually has a bug though.

Try adding:

```
include_examples(
  'autocorrect',
  'before { allow(foo).to receive_message_chain(:"#{foo}" => 42) }',
  'before { allow(foo).to receive(:"#{foo}").and_return(42) }'
)
```

to the test suite (you can also observe the same issue with an interpolated string rather than a symbol).

My suggestion fixes these issues, but I don't know if there are potentially any other corner cases.

We could also, of course, limit the node pattern to only match against sym, dsym, str, and dstr. The way i have it now seems to handle sends and constants though.

ex:

```
receive_message_chain(foo => 42)
```
receive_message_chain(FOO => 42)
```

I can't think of any other semi-reasonable case at the moment.

Darhazer:
@dgollahon thanks for the help. I've applied `node.sym_type? ? ":#{key}" : node.source` and added the test case you've mentioned

PR #381, opened 2017-03-23
Add cop to disallow leaky class definitions in specs

### LINKED ISSUE #197: Suggestion: Check for constants being defined in specs
When you define a constant (modules, classes, constants) in a spec, it’s defined in global scope. That may lead to confusing spec errors:

``` rb
context "one" do
  class Foo
    def self.bar
      1
    end
  end

  it "should be 1" do
    expect(Foo.bar).to eq(1)
  end
end

context "forty two" do
  class Foo
    def self.bar
      42
    end
  end

  it "should be 42" do
    expect(Foo.bar).to eq(42)
  end
end
```

With the output

```
» rspec class_scope_spec.rb
F.

Failures:

  1) one should be 1
     Failure/Error: expect(Foo.bar).to eq(1)

       expected: 1
            got: 42

       (compared using ==)
     # ./class_scope_spec.rb:9:in `block (2 levels) in <top (required)>'

Finished in 0.01052 seconds (files took 0.11059 seconds to load)
2 examples, 1 failure

Failed examples:

rspec ./class_scope_spec.rb:8 # one should be 1
```

alexgerstein:
Closes #197


backus:
@alexgerstein thanks for the PR! We enforce that the config description and the first line of the doc description should be the same. If you run `bin/build_config` it should write the documented description to the config file

alexgerstein:
Very nifty. Thanks, @backus!

backus:
This cop will probably need to be a bit more sophisticated. AFAICT, it is legitimate to do something like

```ruby
before do
  stub_const('Foo', Class.new)
  class Foo
    # define the class now
  end
end
```

in specs and this is a style that I follow a lot.

By the way I realize you just opened the PR so if it isn't ready for review yet sorry for the premature comment.

alexgerstein:
No worries! I appreciate the fast pace.

I'll rework the cop to allow for that pattern. Are there any other common exceptions I should keep in mind?

backus:
> Are there any other common exceptions I should keep in mind?

@alexgerstein Nothing else that immediately comes to mind right now. What I do though when writing cops is I'll do my best to handle for the cases I know, and then once it is working I'll run it against the main rubocop project as well as a few of my private repos and see if I disagree with anything flagged. I'd recommend you do that as well.

alexgerstein:
Thanks for the feedback, @backus. I took your advice and tested out my changes on more repos. It caught some important edge cases, like defining two classes under the same namespace.

This should be ready for review now whenever you have the time.

Thanks again!

spec/rubocop/cop/rspec/leaked_const_spec.rb
```diff
@@ -0,0 +1,100 @@
...
+        ^^^^^^^^^ Opening a class to define methods can pollute your tests. Instead, try using `stub_const` with an anonymized class.
+        end
+        stub_const('Foo', Class.new)
+      end
+    RUBY
+  end
+
+  it 'finds a basic test set-up using modules to open the describe block' do
+    expect_no_violations(<<-RUBY)
+      module TestedNamespace
```
dgollahon: Perhaps this should be a different cop, but we should be flagging this kind of test too, imo. This is probably not a good idea because the interior of the module (or class) is open/mutable and you are now scoped in such a way that you can reference private implementation details of the module that should not be tested directly.

Regardless of whether we should do that here, we should definitely flag module declarations inside of the describe like this:

```ruby
RSpec.describe MyClass do
  module TestModule
    def self.foo
      'bar'
    end
  end

  it 'somehow uses the module' do
    expect(TestModule.foo).to eql('bar')
  end
end
```

as it has the same global setup problem that a class declaration does.

lib/rubocop/cop/rspec/leaked_const.rb
```diff
@@ -0,0 +1,61 @@
...
+      #   end
+      #
+      #   # good
+      #   describe MyClass do
+      #     before do
+      #       stub_const('Foo::BAR', 1)
+      #     end
+      #   end
+      class LeakedConst < Cop
+        MSG = 'Opening a class to define methods can pollute your tests. '\
```
dgollahon: We might be able to improve this message and explain why they pollute the test. Additionally, it might be nice to mention using `Class.new` since I (personally) think that's simpler than / preferable to using `stub_const`. I'm not sure how to put it succinctly yet though (also, i'm not saying this message is bad--it's not, just thinking we may be able to make it better).

lib/rubocop/cop/rspec/leaked_const.rb
```diff
@@ -0,0 +1,61 @@
...
+              'Instead, try using `stub_const` with an anonymized class.'.freeze
+
+        def_node_matcher :stub_const, <<-PATTERN
+          (begin (send nil :stub_const $_ ...) ...)
```
dgollahon: What is the begin node controlling for?

spec/rubocop/cop/rspec/leaked_const_spec.rb
```diff
@@ -0,0 +1,100 @@
...
+      end
+    RUBY
+  end
+
+  it 'finds a new class in a module that is already marked with stub_const' do
+    expect_no_violations(<<-RUBY)
+      before do
+        stub_const('Namespace', Module.new)
+        module Namespace
+          class Foo
```
dgollahon: What happens when you put another class inside this class? Or A module inside this module? Or if the `module Namespace` was `class Namespace` instead?

spec/rubocop/cop/rspec/leaked_const_spec.rb
alexgerstein: That's a good point, @dgollahon. Maybe that's best left for a future update though to keep this PR from getting too big?

lib/rubocop/cop/rspec/leaked_const.rb
alexgerstein: I agree about the message. I had a hard time coming up with good wording. This was the best I could do, but I'll keep trying to re-work it.

To your comment about using stub_const, maybe we shouldn't suggest one over the other in the message since people may have different preferences?

lib/rubocop/cop/rspec/leaked_const.rb
alexgerstein: Since we're checking for usages of stub_const on the class definition's parent, the begin is required to catch the before block.

spec/rubocop/cop/rspec/leaked_const_spec.rb
alexgerstein: Good question. I just tested it out and it seems to behave correctly. So long as the parent class or module is stubbed, the cop does not raise a violation.

alexgerstein:
Thanks for the feedback, @dgollahon! I rebased with master and tried cleaning up the messaging. Let me know if you think it's cleaner.

dgollahon:
Hey @alexgerstein sorry I didn't follow up on this--it completely slipped my mind. I'll try to take a look after work tomorrow.

spec/rubocop/cop/rspec/leaked_const_spec.rb
dgollahon: Yeah, I'm ok with scoping this to classes for now if you find it easier. I don't think we should make this an explicit spec though because I think this should be flagged in the future.

lib/rubocop/cop/rspec/leaked_const.rb
```diff
@@ -0,0 +1,60 @@
...
+          namespace = class_name(node)
+
+          stubbed_const = node.each_ancestor(:begin).any? do |block_ancestor|
+            stub_const(block_ancestor) do |stubbed_const_name|
+              _const_type, stubbed_const_name = *stubbed_const_name
+              namespace.include?(stubbed_const_name.to_s)
+            end
+          end
+
+          add_offense(node, :expression, MSG) unless stubbed_const
```
dgollahon: You can remove `, MSG)` here since that is the default.

lib/rubocop/cop/rspec/leaked_const.rb
```diff
@@ -0,0 +1,60 @@
...
+      #   end
+      #
+      #   # good
+      #   describe MyClass do
+      #     before do
+      #       stub_const('Foo::BAR', 1)
+      #     end
+      #   end
+      class LeakedConst < Cop
+        MSG = 'Ensure that new constants are cleaned up between specs'.freeze
```
dgollahon: Please add a '.' to the end of this.

spec/rubocop/cop/rspec/leaked_const_spec.rb
```diff
@@ -0,0 +1,100 @@
...
+
+  it 'finds a basic test set-up using modules to open the describe block' do
+    expect_no_violations(<<-RUBY)
+      module TestedNamespace
+        describe Subject do
+        end
+      end
+    RUBY
+  end
+end
```
dgollahon: Another failing test case you may find helpful

```ruby
  it 'handles separate stubbing and class creation' do
    expect_no_violations(<<-RUBY)
      RSpec.describe Subject do
        before { stub_const('Foo', Class.new) }
        
        it 'does stuff' do
          class Foo
            def thing
            end
          end
          
          expect(Foo.new.thing).to be(nil)
        end
      end
    RUBY
  end
```

I would also suggest trying cases where you are in different `describe`/`context` nestings.

dgollahon:
Incidentally, I checked `SimpleCov` and this cop is "100%" covered. This is a perfect example of something that can appear well tested using a line coverage metric while actually missing a lot of important test cases--the best example in "the wild" I've run across in a while. Even branch coverage would fall short here. This is why mutation testing is so cool (feel free to check out the fork of `mutant` that @backus and I maintain now: https://github.com/backus/mutest)!

@alexgerstein: btw, please know that I'm not trying pick on you at all here! It's easy to think you've covered more than have and this is a tricky cop. I just think this is a cool example of the failings of line coverage. If you've never tried out mutation testing, you might enjoy giving `mutest` or `mutant` a whirl :) (but don't feel obligated of course--I think it's super cool though I'm obviously biased).

dgollahon:
@alexgerstein: Are you interested in updating / finishing this cop? Or would you like me or someone else to take it over?

alexgerstein:
Sorry for the delay, @dgollahon! Your comments made sense to me. I took a look at Mutest. It looks really interesting.

I was planning to address your suggestions, but I probably won't be able to get to them for a while, so if you or someone else wants to take over to get this cop finished sooner, please feel free.

bquorning:
> Are you interested in updating / finishing this cop? Or would you like me or someone else to take it over?

@dgollahon Are you still interested in taking over?

PR #382, opened 2017-03-24
Add CHANGELOG entry for OverwritingSetup cop

dgollahon:
https://github.com/backus/rubocop-rspec/pull/371#issuecomment-288943181

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Master (Unreleased)
 
+* Add `RSpec/OverwritingSetup` cop. ([@Darhazer][])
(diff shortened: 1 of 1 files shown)


Darhazer:
Thanks. By the way it seems that I've introduced the practice to prefix the cops with `RSpec/` in the changelog. I can clean the changelog and open a PR to have it consistent with the earlier entries.

dgollahon:
Ah, interesting. I think I must have followed your lead then since I've used it a couple of times. Actually, it might be better to just update the other entries to have the `RSpec/` prefix since that's how they're referenced in the config and what they're actually named.

PR #384, opened 2017-03-24
Add DescribeSymbol cop

tsigo:
This cop will add offenses when a Symbol is used as the first argument
to `describe`.

Re-submission of https://github.com/backus/rubocop-rspec/pull/106.


lib/rubocop/cop/rspec/describe_symbol.rb
```diff
@@ -0,0 +1,37 @@
...
+      class DescribeSymbol < Cop
+        MSG = 'Avoid passing a Symbol as the first argument to ' \
+              '`describe`.'.freeze
+
+        def_node_matcher :describe_symbol?, <<-PATTERN
+          (send {(const nil :RSpec) nil} :describe $sym ...)
+        PATTERN
+
+        def on_send(node)
+          match = describe_symbol?(node)
```
dgollahon: I think you can just do `describe_symbol?(node) { add_offense(node, :expression }` since node and match should be the same and the block will only trigger when you match.

lib/rubocop/cop/rspec/describe_symbol.rb
```diff
@@ -0,0 +1,37 @@
...
+      #     ...
+      #   end
+      #
+      # See https://github.com/rspec/rspec-core/issues/1610
+      class DescribeSymbol < Cop
+        MSG = 'Avoid passing a Symbol as the first argument to ' \
+              '`describe`.'.freeze
+
+        def_node_matcher :describe_symbol?, <<-PATTERN
+          (send {(const nil :RSpec) nil} :describe $sym ...)
```
dgollahon: Doesn't seem like you need to capture the `sym` using `$`.

lib/rubocop/cop/rspec/describe_symbol.rb
```diff
@@ -0,0 +1,37 @@
...
+      #   end
+      #
+      #   # good
+      #   describe '#my_method' do
+      #     ...
+      #   end
+      #
+      # See https://github.com/rspec/rspec-core/issues/1610
+      class DescribeSymbol < Cop
+        MSG = 'Avoid passing a Symbol as the first argument to ' \
```
dgollahon: It might be cleaner to just say 'Avoid describing symbols.'

spec/rubocop/cop/rspec/describe_symbol_spec.rb
```diff
@@ -0,0 +1,34 @@
...
+  it 'flags violations for `describe :symbol` with multiple arguments' do
+    expect_violation(<<-RUBY)
+      describe(:some_method, "description") { }
+               ^^^^^^^^^^^^ Avoid passing a Symbol as the first argument to `describe`.
+    RUBY
+  end
+
+  it 'flags violations for `RSpec.describe :symbol`' do
+    expect_violation(<<-RUBY)
+      RSpec.describe(:some_method, "description") { }
```
dgollahon: Might be good to have a case with a nested describe, ex:

```ruby
RSpec.describe Foo do
  describe :to_s do
           ^^^^^ Avoid passing a Symbol as the first argument to `describe`.
  end
end
```

(this case also demonstrates that `RSpec.describe` is not always flagged and a constant is accepted. maybe not totally necessary, but couldn't hurt.)

lib/rubocop/cop/rspec/describe_symbol.rb
tsigo: Definitely much cleaner, but it changes the highlighted bit:

Before:

```
      describe(:some_method) { }
               ^^^^^^^^^^^^ Avoid passing a Symbol as the first argument to `describe`.
```

After:

```
      describe(:some_method) { }
      ^^^^^^^^^^^^^^^^^^^^^^ Avoid passing a Symbol as the first argument to `describe`.
```

I'm fine with that if you are.

lib/rubocop/cop/rspec/describe_symbol.rb
dgollahon: Oh, you were using it, i see^

lib/rubocop/cop/rspec/describe_symbol.rb
dgollahon: Oh, in that case you can also just do `describe_symbol?(node) { |symbol| add_offense(symbol, :expression }`.

tsigo: Yep, will do. 👍

dgollahon:
@tsigo: Can you add a changelog entry for this? You can add yourself as a contributor at the bottom of the document and link to yourself so the format is uniform.

tsigo:
Should be all set, thanks for the quick review @dgollahon!

dgollahon:
Thanks and no problem, @tsigo. :)

PR #394, opened 2017-04-02
Add LetBeforeExamples cop

Darhazer:


config/default.yml
```diff
@@ -137,6 +137,10 @@ RSpec/LeadingSubject:
   Description: Checks for `subject` definitions that come after `let` definitions.
   Enabled: true
 
+RSpec/LetBeforeExamples:
+  Description: Checks for `let` definition that come after an example or example group.
```
dgollahon: I think you want `definitions`, plural, rather than singular here.

(I might also say `an example group` but i think the way you have it is also grammatically correct).

PR #403, opened 2017-04-26
Fix example wording autocorrect

### LINKED ISSUE #402: Auto correct of RSpec/ExampleWording crashes
Hello! I have been using rubocop-rspec without any issues so far and like it a lot. However, today I started to use it on a bigger project and I found the following example that crashes the auto correct:

```ruby
# model_spec.rb
it 'It does something' do
  # ^ Note the uppercase i
end
```

If I run it without the autocorrect, the offense is registered correctly:

```
$ rubocop --only RSpec/ExampleWording model_spec.rb 
Inspecting 1 file
C

Offenses:

model_spec.rb:1:5: C: Do not repeat 'it' when describing your tests.
it 'It does something' do
    ^^^^^^^^^^^^^^^^^
```

However, it breaks with the `--auto-correct` flag:

```
$ rubocop --only RSpec/ExampleWording --auto-correct model_spec.rb
Inspecting 1 file


0 files inspected, no offenses detected
no implicit conversion of nil into String
/home/phansch/.rvm/gems/ruby-2.3.1/gems/parser-2.3.3.1/lib/parser/source/rewriter.rb:156:in `[]='
/home/phansch/.rvm/gems/ruby-2.3.1/gems/parser-2.3.3.1/lib/parser/source/rewriter.rb:156:in `block in process'
/home/phansch/.rvm/gems/ruby-2.3.1/gems/parser-2.3.3.1/lib/parser/source/rewriter.rb:152:in `each'
/home/phansch/.rvm/gems/ruby-2.3.1/gems/parser-2.3.3.1/lib/parser/source/rewriter.rb:152:in `process'
/home/phansch/.rvm/gems/ruby-2.3.1/gems/rubocop-0.48.1/lib/rubocop/cop/corrector.rb:63:in `rewrite'
/home/phansch/.rvm/gems/ruby-2.3.1/gems/rubocop-0.48.1/lib/rubocop/cop/team.rb:143:in `autocorrect_all_cops'
/home/phansch/.rvm/gems/ruby-2.3.1/gems/rubocop-0.48.1/lib/rubocop/cop/team.rb:80:in `autocorrect'
/home/phansch/.rvm/gems/ruby-2.3.1/gems/rubocop-0.48.1/lib/rubocop/cop/team.rb:108:in `block in offenses'
/home/phansch/.rvm/gems/ruby-2.3.1/gems/rubocop-0.48.1/lib/rubocop/cop/team.rb:125:in `investigate'
/home/phansch/.rvm/gems/ruby-2.3.1/gems/rubocop-0.48.1/lib/rubocop/cop/team.rb:104:in `offenses'
/home/phansch/.rvm/gems/ruby-2.3.1/gems/rubocop-0.48.1/lib/rubocop/cop/team.rb:54:in `inspect_file'
/home/phansch/.rvm/gems/ruby-2.3.1/gems/rubocop-0.48.1/lib/rubocop/runner.rb:248:in `inspect_file'

```

I guess that's because [`replacement_text`](https://github.com/backus/rubocop-rspec/blob/master/lib/rubocop/cop/rspec/example_wording.rb#L77) returns `nil` in that case. I probably have time to look into a fix later today.

dgollahon:
Fix example wording autocorrect

- Fixes #402
- Previously the `#on_block` method would correctly flag cases of leading
  `it` and `should` case-insensitively by downcasing the docstring but
  in the autocorrect phase the message casing  was left untouched which
  caused a nil bug for upper cased `Should`s, `IT`s, etc.
- Prevents words starting with `should` from being flagged
  * ex: `it 'shoulders the burden' {}`
- Improves `shouldn't be` handling
  * original docstring: `shouldn't be true`
  * correction before: `does not be true`
  * correction after: `is not true`
- Fixes the pluralizer so that it is not confused by uppercase letters
  * original docstrings: `should WISH ME LUCK`, `should WORRY`
  * corrections before: `WISHs ME LUCK`, `WORRYs`
  * corrections after: `WISHES ME LUCK`, `WORRIES`
- If a whole word being corrected is completely uppercase, the suffix
  appended will also be uppercase. If the word is mixed case, the
  lowercase suffix will be appended.
- Performs general refactoring.
- Adds pluralization support for words ending in `z`
    * ex: `buzz` -> `buzzes`

### DIFF
--- a/lib/rubocop/cop/rspec/example_wording.rb
+++ b/lib/rubocop/cop/rspec/example_wording.rb
@@ -33,8 +33,8 @@ module RuboCop
         MSG_SHOULD = 'Do not use should when describing your tests.'.freeze
         MSG_IT     = "Do not repeat 'it' when describing your tests.".freeze
 
-        SHOULD_PREFIX = 'should'.freeze
-        IT_PREFIX     = 'it '.freeze
+        SHOULD_PREFIX = /\Ashould(?:n't)?\b/i
+        IT_PREFIX     = /\Ait /i
 
         def_node_matcher(
           :it_description,
@@ -43,20 +43,16 @@ module RuboCop
 
         def on_block(node)
           it_description(node) do |description_node, message|
-            text = message.downcase
-
-            if text.start_with?(SHOULD_PREFIX)
+            if message =~ SHOULD_PREFIX
               add_wording_offense(description_node, MSG_SHOULD)
-            elsif text.start_with?(IT_PREFIX)
+            elsif message =~ IT_PREFIX
               add_wording_offense(description_node, MSG_IT)
             end
           end
         end
 
         def autocorrect(range)
-          lambda do |corrector|
-            corrector.replace(range, replacement_text(range))
-          end
+          ->(corrector) { corrector.replace(range, replacement_text(range)) }
         end
 
         private
@@ -77,13 +73,13 @@ module RuboCop
         def replacement_text(range)
           text = range.source
 
-          if text.start_with?('should')
+          if text =~ SHOULD_PREFIX
             RuboCop::RSpec::Wording.new(
               text,
               ignore:  ignored_words,
               replace: custom_transform
             ).rewrite
-          elsif text.start_with?(IT_PREFIX)
+          else
             text.sub(IT_PREFIX, '')
           end
         end
--- a/lib/rubocop/rspec/wording.rb
+++ b/lib/rubocop/rspec/wording.rb
@@ -4,6 +4,11 @@ module RuboCop
   module RSpec
     # RSpec example wording rewriter
     class Wording
+      SHOULDNT_PREFIX    = /\Ashould(?:n't| not)\b/i
+      SHOULDNT_BE_PREFIX = /#{SHOULDNT_PREFIX} be\b/i
+      ES_SUFFIX_PATTERN  = /(?:o|s|x|ch|sh|z)\z/i
+      IES_SUFFIX_PATTERN = /[^aeou]y\z/i
+
       def initialize(text, ignore:, replace:)
         @text         = text
         @ignores      = ignore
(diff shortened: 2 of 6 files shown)


lib/rubocop/rspec/wording.rb
```diff
@@ -4,44 +4,78 @@ module RuboCop
...
-          return "#{word[0..-2]}ies"
+        words.join(' ')
+      end
+
+      def ignored_word?(word)
+        ignores.any? { |ignore| ignore.casecmp(word).zero? }
+      end
+
+      def substitute(word)
+        # NOTE: Custom replacements are case sensitive.
```
dgollahon: Not totally sure how we should handle this. As a temporary hack I just added uppercase versions of our replacements in the default yml. We could also transform the mapping into lower case and do the same uppercasing logic I have elsewhere but that could potentially be weird in a case or two. Any thoughts?

dgollahon:
@backus, @bquorning, or @Darhazer if one of you has a minute, this is ready for review.

FYI: The new changes are mutation covered (including inlining the regexes so that the mutator could see them).

lib/rubocop/cop/rspec/example_wording.rb
```diff
@@ -33,8 +33,8 @@ class ExampleWording < Cop
         MSG_SHOULD = 'Do not use should when describing your tests.'.freeze
         MSG_IT     = "Do not repeat 'it' when describing your tests.".freeze
 
-        SHOULD_PREFIX = 'should'.freeze
-        IT_PREFIX     = 'it '.freeze
+        SHOULD_PREFIX = /\Ashould(?:n't|not)?\b/i
```
bquorning: There should probably be a space in front of `not`.

dgollahon: By using a boundary character instead of a space it will detect single word cases like `it 'shouldn't' {}`.

dgollahon:
Obviously there are still plenty of edge cases we're not covering, but we should consider checking into, at least briefly situations like

`it 'should X and Y'`
`it 'should X, Y, and Z`
`it 'should X then Y'`

where it will only pluralize X and the verb tenses will disagree.

Even if some of those might be too tricky to fix, we should consider detecting that we can't fix them and skip the example.

lib/rubocop/cop/rspec/example_wording.rb
dgollahon: Oh, _in front of_. Sorry, I misread that.

bquorning: Yeah :-)
```
>> "shouldn't be"[SHOULD_PREFIX]
"shouldn't"
>> "shouldnot be"[SHOULD_PREFIX]
"shouldnot"
>> "should not be"[SHOULD_PREFIX]
"should"
```

dgollahon: Weirdly, it works because it flags the regular should followed by a boundary character and then the `Wording` class has a correct regex. Technically doing `|not` or `| not` is redundant since it includes `\b`. I think I'll just delete that side of the alternation since it's effectively unreachable code.

dgollahon: Good catch.

bquorning: But since we don’t capture anything, I think we can just remove the entire `|not` part.

bquorning: There seems to have been a race condition in our agreement 😉

dgollahon: > There seems to have been a race condition in our agreement 😉

haha!

When I ran `mutest`, it actually caught this case. But I forgot to inline this particular regex (i did all the ones in the `Wording` class). I need to add static constant mutation support to `mutest` one of these days...

dgollahon:
> We have no ambitions of replacing shouldn't with doesn't, right?

Hm, I don't think so. Probably more work than its worth to detect--especially since it's an equivalent statement.

PR #406, opened 2017-04-30
Bump version to 1.15.1

Darhazer:
Bug fix release - there are 2 crashes fixed, so it's good to release a new version
Also fixed the date of the previous release


dgollahon:
I was thinking we should do a release after #404 got merged the other day, nice 👍

dgollahon:
I don't think this made it to rubygems yet though. I'll push it out.

PR #407, opened 2017-05-01
Rename ExpectViolation to ExpectOffense

backus:
I'd like to upstream this to RuboCop soon!

 - Drops dependency on concord
 - Drops dependency on anima
 - Drops dependency on adamantium

See also bbatsov/rubocop#3905

### DIFF
--- a/rubocop-rspec.gemspec
+++ b/rubocop-rspec.gemspec
@@ -39,8 +39,5 @@ Gem::Specification.new do |spec|
   spec.add_development_dependency 'rake'
   spec.add_development_dependency 'rspec', '>= 3.4'
   spec.add_development_dependency 'simplecov'
-  spec.add_development_dependency 'anima'
-  spec.add_development_dependency 'concord'
-  spec.add_development_dependency 'adamantium'
   spec.add_development_dependency 'yard'
 end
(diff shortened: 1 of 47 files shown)


backus:
@dgollahon WDYT? I'd be open to a PR that makes the classes a bit bigger if you are interested but I personally don't want to do that right now 😆.

dgollahon:
> @dgollahon WDYT?

I think the renaming makes sense. Why'd you decide to remove the dev dependencies? Was it for moving to `rubocop`? Because otherwise I think having lightweight dev dependencies is pretty harmless.

> I'd be open to a PR that makes the classes a bit bigger if you are interested

WDYM?

If you're talking about general changes to this, I think one of the biggest things to do would be to make error messages more reasonable (maybe using some rspec features would be helpful?). It can be a bit frustrating to debug some cases.

backus:
@dgollahon yeah I'm trying to eventually upstream this to RuboCop. See bbatsov/rubocop#3905. That is the reason for the rename and the dev dependency removal.

> > I'd be open to a PR that makes the classes a bit bigger if you are interested
> WDYM?

> If you're talking about general changes to this, I think one of the biggest things to do would be to make error messages more reasonable (maybe using some rspec features would be helpful?). It can be a bit frustrating to debug some cases.

Well I figured that you'd probably agree this has "tiny class syndrome." Was just preempting discussion about reducing that

dgollahon:
> Well I figured that you'd probably agree this has "tiny class syndrome." Was just preempting discussion about reducing that

Ah, yeah. Since this wasn't new code or a directed refactor I didn't really think nitpicking the implementation was terribly important. The object could probably be improved but that doesn't have to be done right now.

Issue #393, opened 2017-03-31
Cop Idea: Forbid non static sentences declared statically on factories


jonatas:
Hi, I'm going to work in a cop for factories (using factory girl), complaining against nonstatic attributes declared statically.

The idea is cover cases like:

```ruby
# bad
kind [:active, :rejected].sample
# good
kind { [:active, :rejected].sample }

# bad
closed_at 1.day.from_now
# good
closed_at { 1.day.from_now }

# good
kind :static

# good
comments_count 0

# good
type User::MAGIC
```

Maybe the `rubocop-rspec` project is a good place to put this Cop. Are you interested in receiving a contribution with a Cop like this?

Darhazer:
I'm planning on implementing some rails and capybara specific cops, so FactoryGirl extension is also welcome. I think however that all those not-core extensions have to be in their own namespaces (`RSpec/FactoryGirl/....`) and to have global on/off switch, e.g. disabling all factorygirl checks at once

PR #409, opened 2017-05-04
Forbids dynamic attributes defined statically on factories

jonatas:
I proposed it in #393.

It complains against nonstatic attributes declared statically without being wrapped in a block.

The idea is to cover cases like:

```ruby
# bad
kind [:active, :rejected].sample
# good
kind { [:active, :rejected].sample }

# bad
closed_at 1.day.from_now
# good
closed_at { 1.day.from_now }

# good
kind :static

# good
comments_count 0

# good
type User::MAGIC
```

### DIFF
--- a/config/default.yml
+++ b/config/default.yml
@@ -4,6 +4,10 @@ AllCops:
     Patterns:
     - _spec.rb
     - "(?:^|/)spec/"
+  RSpec/FactoryGirl:
+    Patterns:
+    - spec/factories/**/*.rb
+    - features/support/factories/**/*.rb
 
 RSpec/AnyInstance:
   Description: Check that instances are not being stubbed globally.
@@ -224,3 +228,7 @@ RSpec/VerifiedDoubles:
   Description: Prefer using verifying doubles over normal doubles.
   Enabled: true
   IgnoreSymbolicNames: false
+
+FactoryGirl/DynamicAttributeDefinedStatically:
+  Description: Prefer declaring dynamic attribute values in a block.
+  Enabled: true
--- a/lib/rubocop/cop/rspec/factory_girl/dynamic_attribute_defined_statically.rb
+++ b/lib/rubocop/cop/rspec/factory_girl/dynamic_attribute_defined_statically.rb
@@ -1,0 +1,84 @@
+# frozen_string_literal: true
+
+module RuboCop
+  module Cop
+    module RSpec
+      module FactoryGirl
+        # Prefer declaring dynamic attribute values in a block.
+        #
+        # @example
+        #   # bad
+        #   kind [:active, :rejected].sample
+        #
+        #   # good
+        #   kind { [:active, :rejected].sample }
+        #
+        #   # bad
+        #   closed_at 1.day.from_now
+        #
+        #   # good
+        #   closed_at { 1.day.from_now }
+        #
+        #   # good
+        #   kind :static
+        #
+        #   # good
+        #   comments_count 0
+        #
+        #   # good
+        #   type User::MAGIC
+        class DynamicAttributeDefinedStatically < Cop
+          MSG = 'Use a block to set a dynamic value to an attribute.'.freeze
+
+          def_node_matcher :dynamic_defined_statically?, <<-PATTERN
+          (send nil _ (send ... ))
+          PATTERN
+
+          def_node_search :factory_attributes, <<-PATTERN
+          (block (send nil {:factory, :trait} ...) _ { (begin $...) $(send ...) } )
+          PATTERN
+
+          def on_block(node)
+            return if node.method_name == :trait
+            factory_attributes(node).to_a.flatten.each do |attribute|
+              if dynamic_defined_statically?(attribute)
+                add_offense(attribute, :expression)
+              end
+            end
+          end
+
+          def autocorrect(node)
+            if method_uses_parens?(node.location)
+              autocorrect_replacing_parens(node)
+            else
+              autocorrect_without_parens(node)
+            end
+          end
+
+          private
+
+          def method_uses_parens?(location)
+            return false unless location.begin && location.end
+            location.begin.source == '(' && location.end.source == ')'
+          end
+
+          def autocorrect_replacing_parens(node)
+            lambda do |corrector|
+              corrector.replace(node.location.begin, ' { ')
+              corrector.replace(node.location.end, ' }')
+            end
+          end
+
+          def autocorrect_without_parens(node)
+            lambda do |corrector|
+              arguments = node.descendants.first
+              expression = arguments.location.expression
+              corrector.insert_before(expression, '{ ')
+              corrector.insert_after(expression, ' }')
+            end
+          end
+        end
+      end
+    end
+  end
+end
--- a/lib/rubocop/rspec/config_formatter.rb
+++ b/lib/rubocop/rspec/config_formatter.rb
@@ -4,7 +4,7 @@ module RuboCop
   module RSpec
     # Builds a YAML config file from two config hashes
     class ConfigFormatter
-      NAMESPACE = 'RSpec'.freeze
+      NAMESPACES = /^(#{Regexp.union('RSpec', 'FactoryGirl')})/
 
       def initialize(config, descriptions)
         @config       = config
@@ -12,7 +12,7 @@ module RuboCop
       end
 
       def dump
-        YAML.dump(unified_config).gsub(/^#{NAMESPACE}/, "\n#{NAMESPACE}")
+        YAML.dump(unified_config).gsub(NAMESPACES, "\n\\1")
       end
 
       private
@@ -24,7 +24,7 @@ module RuboCop
       end
 
       def cops
-        (descriptions.keys + config.keys).uniq.grep(/\A#{NAMESPACE}/)
+        (descriptions.keys | config.keys).grep(NAMESPACES)
       end
 
       attr_reader :config, :descriptions
--- a/lib/rubocop/rspec/description_extractor.rb
+++ b/lib/rubocop/rspec/description_extractor.rb
@@ -19,7 +19,7 @@ module RuboCop
 
       # Decorator of a YARD code object for working with documented rspec cops
       class CodeObject
-        COP_NAMESPACE = 'RuboCop::Cop::RSpec'.freeze
+        RSPEC_NAMESPACE = 'RuboCop::Cop::RSpec'.freeze
 
         def initialize(yardoc)
           @yardoc = yardoc
@@ -54,7 +54,7 @@ module RuboCop
         end
 
         def rspec_cop_namespace?
-          documented_constant.start_with?(COP_NAMESPACE)
+          documented_constant.start_with?(RSPEC_NAMESPACE)
         end
 
         def documented_constant
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Master (Unreleased)
 
+* Add `RSpec/FactoryGirl` namespace including the first cop for factories: `FactoryGirl/DynamicAttributeDefinedStatically`. ([@jonatas][])
+
 ## 1.15.1 (2017-04-30)
 
 * Fix the handling of various edge cases in the `RSpec/ExampleWording` cop, including one that would cause autocorrect to crash. ([@dgollahon][])
@@ -218,3 +220,4 @@
 [@cfabianski]: https://github.com/cfabianski
 [@dgollahon]: https://github.com/dgollahon
 [@tsigo]: https://github.com/tsigo
+[@jonatas]: https://github.com/jonatas
--- a/bin/build_config
+++ b/bin/build_config
@@ -8,7 +8,8 @@ require 'rubocop-rspec'
 require 'rubocop/rspec/description_extractor'
 require 'rubocop/rspec/config_formatter'
 
-glob = File.join(__dir__, '..', 'lib', 'rubocop', 'cop', 'rspec', '*.rb')
+cops = '{*.rb,factory_girl/*.rb}'
+glob = File.join(__dir__, '..', 'lib', 'rubocop', 'cop', 'rspec', cops)
 YARD.parse(Dir[glob], [])
 
 descriptions = RuboCop::RSpec::DescriptionExtractor.new(YARD::Registry.all).to_h
--- a/lib/rubocop-rspec.rb
+++ b/lib/rubocop-rspec.rb
@@ -16,6 +16,7 @@ require 'rubocop/rspec/example_group'
 require 'rubocop/rspec/example'
 require 'rubocop/rspec/hook'
 require 'rubocop/cop/rspec/cop'
+require 'rubocop/rspec/factory_girl'
 
 RuboCop::RSpec::Inject.defaults!
 
@@ -35,6 +36,7 @@ require 'rubocop/cop/rspec/example_length'
 require 'rubocop/cop/rspec/example_wording'
 require 'rubocop/cop/rspec/expect_actual'
 require 'rubocop/cop/rspec/expect_output'
+require 'rubocop/cop/rspec/factory_girl/dynamic_attribute_defined_statically'
 require 'rubocop/cop/rspec/file_path'
 require 'rubocop/cop/rspec/focus'
 require 'rubocop/cop/rspec/hook_argument'
--- a/lib/rubocop/rspec/factory_girl.rb
+++ b/lib/rubocop/rspec/factory_girl.rb
@@ -1,0 +1,7 @@
+module RuboCop
+  module RSpec
+    # RuboCop FactoryGirl project namespace
+    module FactoryGirl
+    end
+  end
+end
--- a/spec/project/default_config_spec.rb
+++ b/spec/project/default_config_spec.rb
@@ -4,14 +4,18 @@ RSpec.describe 'config/default.yml' do
   end
 
   let(:cop_names) do
-    glob = SpecHelper::ROOT.join('lib', 'rubocop', 'cop', 'rspec', '*.rb')
-
+    namespaces = {
+      'rspec' => 'RSpec',
+      'factory_girl' => 'FactoryGirl'
+    }
+    glob = SpecHelper::ROOT.join('lib', 'rubocop', 'cop',
+                                 'rspec', '{,factory_girl/}*.rb')
     cop_names =
       Pathname.glob(glob).map do |file|
         file_name = file.basename('.rb').to_s
         cop_name  = file_name.gsub(/(^|_)(.)/) { Regexp.last_match(2).upcase }
-
-        "RSpec/#{cop_name}"
+        namespace = namespaces[file.dirname.basename.to_s]
+        "#{namespace}/#{cop_name}"
       end
 
     cop_names - %w[RSpec/Cop]
--- a/spec/rubocop/cop/rspec/factory_girl/dynamic_attribute_defined_statically_spec.rb
+++ b/spec/rubocop/cop/rspec/factory_girl/dynamic_attribute_defined_statically_spec.rb
@@ -1,0 +1,87 @@
+# frozen_string_literal: true
+
+# rubocop:disable Metrics/LineLength
+RSpec.describe RuboCop::Cop::RSpec::FactoryGirl::DynamicAttributeDefinedStatically do
+  subject(:cop) { described_class.new(config) }
+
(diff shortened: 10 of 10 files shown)


jonatas:
I logged in buildkite website but I don't have access to the build links. 🤔

dgollahon:
@jonatas: You can look at the Travis build (it's the "public" CI). The BK build is something @backus setup for collaborators.

jonatas:
Ok! Thanks, @dgollahon! looking the Travis is enough 👍

config/default.yml
```diff
@@ -4,6 +4,10 @@ AllCops:
     Patterns:
     - _spec.rb
     - "(?:^|/)spec/"
+  FactoryGirl:
```
Darhazer: IMO it should still be prefixed `RSpec/` e.g. having nested namespace here. Also it should have an option to disable the whole namespace, similar to rubocop's railscops

lib/rubocop/cop/factory_girl/dynamic_attribute_defined_statically.rb
```diff
@@ -0,0 +1,58 @@
...
+      #   # good
+      #   comments_count 0
+      #
+      #   # good
+      #   type User::MAGIC
+      class DynamicAttributeDefinedStatically < Cop
+        MSG = 'Use a block to set a dynamic value to an attribute.'.freeze
+
+        def_node_matcher :dynamic_defined_statically?, <<-PATTERN
+          (send nil _method (send ... {:ago :current :from_now :now :sample}))
```
Darhazer: IMO any method call is a dynamic, so you can match any send type node. Can't think of a case when this would fail

lib/rubocop/cop/factory_girl/dynamic_attribute_defined_statically.rb
```diff
@@ -0,0 +1,58 @@
...
+        PATTERN
+
+        def on_send(node)
+          return if already_defined_in_a_block?(node)
```
Darhazer: although the cop is configured to check only spec/factories/, maybe it's good to check if you're in a FactoryGirl.define block in the first place

lib/rubocop/cop/factory_girl/dynamic_attribute_defined_statically.rb
```diff
@@ -0,0 +1,58 @@
...
+            expression = node.descendants.first.location.expression
+            corrector.insert_before(expression, '{ ')
+            corrector.insert_after(expression, ' }')
+          end
+        end
+
+        private
+
+        def already_defined_in_a_block?(node)
+          node.sibling_index.zero? && node.parent && node.parent.block_type?
```
Darhazer: is there specific reason to check only the first sibling?

spec/rubocop/cop/factory_girl/dynamic_attribute_defined_statically_spec.rb
```diff
@@ -0,0 +1,63 @@
...
+        title "Static"
+        description FFaker::Lorem.paragraph(10)
+        tag Tag::MAGIC
+      end
+    end
+  RUBY
+
+  it 'registers an offense for offending code' do
+    inspect_source(cop, bad)
+    expect(cop.offenses.size).to eq(3)
```
Darhazer: use the expect_violation and expect_no_violations helpers. There are very convenient and make tests very descriptive

Darhazer:
I have run on a local project and overall looks good. It does not catch cases where the dynamic call is in a hash/array/other structure though. Also, the list of dynamic methods can be expanded. I think catching all `send` nodes will do better job, but I'm not sure if this won't lead to some false positives. I'll be happy to test each implementation against my code base to see for such false positives

lib/rubocop/rspec/config_formatter.rb
```diff
@@ -24,7 +24,9 @@ def unified_config
       end
 
       def cops
-        (descriptions.keys + config.keys).uniq.grep(/\A#{NAMESPACE}/)
+        (descriptions.keys + config.keys)
+          .uniq
```
DNNX: `(a + b).uniq` -> `a | b`

A good candidate for a new cop, by the way.

lib/rubocop/cop/factory_girl/dynamic_attribute_defined_statically.rb
jonatas: I observed that the value is the first sibling when it's a block. I'll recheck if I can improve it.

backus:
Since it sounds like @Darhazer approves, could you squash down the commits into one nice passing commit and rebase @jonatas? I'll merge after that

jonatas:
@Darhazer, I'm not sure about how to keep it using a simple `send`. Got some issues with the implementation because constants are also a `send`.

config/default.yml
jonatas: Do you mean make the namespace like `RSpec/FactoryGirl`?

jonatas: I'm not sure where to put the option to disable the whole namespace. Should it be when it's requiring the files or a kind of verification that all factory girl cops should follow?

dgollahon:
>  Got some issues with the implementation because constants are also a send.

WDYM, exactly? Do you have an example of a case you found problematic? Referencing a const is not the same as a send (unless there's some case I'm not aware of).

```shell
$ ruby-parse -e 'closed_at { HISTORICAL_DATE }'

(block
  (send nil :closed_at)
  (args)
  (const nil :HISTORICAL_DATE))
```

jonatas:
I was tryng a example with Faker:

```
$ ruby-parse -e 'description FFaker::Lorem.paragraph(10)'                                 17:07:35
(send nil :description
  (send
    (const
      (const nil :FFaker) :Lorem) :paragraph
    (int 10)))
```

dgollahon:
So in that case it is non-static and you would want to flag it. The AST is only showing you that `(const (const nil :FFaker))` is the receiver of the `send` node--the constant lookup itself does not involve a `send`.

You should be able to capture the children of the factory-related send (in this case, `description`) and just recurse over its children with another matcher that just looks for any `send` node.

That's assuming the point is to enforce no non-static, non-block factories (which may or may not fit everyone's use cases).

config/default.yml
Darhazer: Regarding prefix: yes
Regarding the option, haven't really thought about this. Guess disabling the requires is easier. @backus any thoughts?

jonatas:
Thanks, @dgollahon! I implemented your suggestion and now it catches all `send` 🎉

lib/rubocop/cop/factory_girl/dynamic_attribute_defined_statically.rb
```diff
@@ -0,0 +1,73 @@
...
+
+        def trait_node?(node)
+          node.block_type? && node.method_name == :trait
+        end
+
+        def into_factories?(node)
+          node.ancestors.find(&method(:into_factory_block?))
+        end
+
+        def already_defined_in_a_block?(node)
```
Darhazer: Seems that this is not used anymore

spec/rubocop/cop/factory_girl/dynamic_attribute_defined_statically_spec.rb
```diff
@@ -0,0 +1,74 @@
...
+RSpec.describe RuboCop::Cop::FactoryGirl::DynamicAttributeDefinedStatically do
+  subject(:cop) { described_class.new(config) }
+
+  let(:config) { RuboCop::Config.new }
+
+  it 'registers an offense for offending code' do
+    expect_offense(<<-RUBY)
+      FactoryGirl.define do
+        factory :post do
+          trait :published do
```
Darhazer: I would separate the tests for better readability (registers an offense in a trait as an own test case)

lib/rubocop/cop/factory_girl/dynamic_attribute_defined_statically.rb
```diff
@@ -0,0 +1,73 @@
...
+          end
+        end
+
+        private
+
+        def trait_node?(node)
+          node.block_type? && node.method_name == :trait
+        end
+
+        def into_factories?(node)
```
Darhazer: not used :scissors:

config/default.yml
jonatas: Thanks, @Darhazer.

Should I also move the Cop class to `lib/rubocop/cop/rspec/factory_girl` and make the module with 4 levels like `RuboCop::Cop::RSpec::FactoryGirl`?

lib/rubocop/cop/rspec/factory_girl/dynamic_attribute_defined_statically.rb
```diff
@@ -0,0 +1,75 @@
...
+          (send nil _ (send ... ))
+          PATTERN
+
+          def_node_search :factory_attributes, <<-PATTERN
+          (block (send nil {:factory, :trait} ...) _ { (begin $...) $(send ...) } )
+          PATTERN
+
+          def on_block(node)
+            return if trait_node?(node)
+            factory_attributes(node).to_a.flatten.each do |attribute|
```
jonatas: I was checking the original documentation, Looks like the library supports [inheritance](https://github.com/thoughtbot/factory_girl/blob/master/GETTING_STARTED.md#inheritance). I never used it but I believe that the `.to_a.flatten` will do the job for us. 😌

lib/rubocop/rspec/config_formatter.rb
```diff
@@ -4,15 +4,15 @@ module RuboCop
...
+      NAMESPACES = ['RSpec'.freeze, 'FactoryGirl'.freeze].freeze
 
       def initialize(config, descriptions)
         @config       = config
         @descriptions = descriptions
       end
 
       def dump
-        YAML.dump(unified_config).gsub(/^#{NAMESPACE}/, "\n#{NAMESPACE}")
+        YAML.dump(unified_config).gsub(/^(#{NAMESPACES.join('|')})/, "\n\\1")
```
backus: I wonder if `Regexp.union` would have been usable here. Not requesting the change just curious

lib/rubocop/rspec/description_extractor.rb
```diff
@@ -54,7 +55,7 @@ def class_documentation?
         end
 
         def rspec_cop_namespace?
-          documented_constant.start_with?(COP_NAMESPACE)
+          documented_constant.start_with?(RSPEC_NAMESPACE, FACTORIES_NAMESPACE)
```
backus: Is `FACTORIES_NAMESPACE` dead code now that it lives inside of `RuboCop::Cop::RSpec`?

backus:
This is neat! I ended up doing a quick look over the code since it is significant that we are adding a new namespace. Can you add a CHANGELOG entry?

lib/rubocop/rspec/config_formatter.rb
jonatas: Cool @backus!
<details>
  <summary> Looks like work in the same way.</summary>

```ruby
simple =  /^(#{["a","b"].join("|")})/ # => /^(a|b)/
with_union = /^(#{Regexp.union("a","b")})/ # => /^((?-mix:a|b))/
puts "a first \nb second".gsub(simple, "\n\\1")

a first

b second
=> nil
puts "a first \nb second".gsub(with_union, "\n\\1")

a first

b second
=> nil
```
</details>

config/default.yml
```diff
@@ -224,3 +228,7 @@ RSpec/VerifiedDoubles:
   Description: Prefer using verifying doubles over normal doubles.
   Enabled: true
   IgnoreSymbolicNames: false
+
+FactoryGirl/DynamicAttributeDefinedStatically:
+  Description: Prefer to declare dynamic attribute values in a block.
```
dgollahon: Minor wording tweak: I might start the sentence with "Prefer declaring" or just "Declare". Not a requirement, just a suggestion.

lib/rubocop/rspec/description_extractor.rb
jonatas: Oh yeah.. I'll remove it now.

lib/rubocop/cop/rspec/factory_girl/dynamic_attribute_defined_statically.rb
```diff
@@ -0,0 +1,75 @@
...
+              expression = node.descendants.first.location.expression
+              corrector.insert_before(expression, '{ ')
+              corrector.insert_after(expression, ' }')
+            end
+          end
+
+          private
+
+          def trait_node?(node)
+            node.block_type? && node.method_name == :trait
```
dgollahon: Do you need to check `#block_type?` since this node came from `#on_block`?

lib/rubocop/cop/rspec/factory_girl/dynamic_attribute_defined_statically.rb
```diff
@@ -0,0 +1,75 @@
...
+        #   # good
+        #   comments_count 0
+        #
+        #   # good
+        #   type User::MAGIC
+        class DynamicAttributeDefinedStatically < Cop
+          MSG = 'Use a block to set a dynamic value to an attribute.'.freeze
+
+          def_node_matcher :dynamic_defined_statically?, <<-PATTERN
+          (send nil _ (send ... ))
```
dgollahon: If you want to be thorough, you could go a little farther with this. This will not detect cases like:

```ruby
foo do
  "hi, i'm not a send"
  bar.baz
end
```

You could do `(send nil _ $...)` instead and perform a search for just `(send)` over its children.

dgollahon: Although, you may have to be careful since some factory components might be nested. There are probably some edge cases here already.

lib/rubocop/cop/rspec/factory_girl/dynamic_attribute_defined_statically.rb
```diff
@@ -0,0 +1,75 @@
...
+              if dynamic_defined_statically?(attribute)
+                add_offense(attribute, :expression)
+              end
+            end
+          end
+
+          def autocorrect(node)
+            lambda do |corrector|
+              expression = node.descendants.first.location.expression
+              corrector.insert_before(expression, '{ ')
```
dgollahon: Will this break if the attribute is defined using `do ... end` syntax.

lib/rubocop/cop/rspec/factory_girl/dynamic_attribute_defined_statically.rb
jonatas: Makes sense :) I'll even drop the method.

lib/rubocop/cop/rspec/factory_girl/dynamic_attribute_defined_statically.rb
jonatas: No. We're filtering for nodes that are missing the block.

dgollahon: > We're filtering for nodes that are missing the block.

I don't mean nodes that don't have a block, i mean nodes that use the multiline block format:

```ruby
factory :post do
  status do
    x = 1
    x + x
  end
end
```

dgollahon: Oh, right, nevermind. I misinterpreted what was happening here.

dgollahon: What if the attribute was defined with parenthesis around the call?

ex:
```ruby
status(foo.bar)
```

Do the parens get in the way?

lib/rubocop/cop/rspec/factory_girl/dynamic_attribute_defined_statically.rb
jonatas: I don't know if I get your point. The cop is looking only for calls that are not using a block. if it's using a block, it's already dynamic right?

dgollahon: Yup, you're right I'm sorry. I flipped the problem in my mind. You _could_ still have a case like

```ruby
status((x=1; baz(x)))'
```

but I think that's so rare we shouldn't really worry about it.

lib/rubocop/cop/rspec/factory_girl/dynamic_attribute_defined_statically.rb
jonatas: Good question 😄 

I didn't think about this alternative syntax. Do you think there are people declaring factories using parens in syntax?

dgollahon: I'm sure someone is. I tend to use parens in DSLs more than the average rubyist.

jonatas: Ok. I'll cover it!

jonatas:
Rebased again @backus! Is there anything else I can do now?

config/default.yml
```diff
@@ -224,3 +228,7 @@ RSpec/VerifiedDoubles:
   Description: Prefer using verifying doubles over normal doubles.
   Enabled: true
   IgnoreSymbolicNames: false
+
+FactoryGirl/DynamicAttributeDefinedStatically:
```
Darhazer: I just wonder if this shouldn't be under `RSpec/` namespace as well, e.g. does it work right now

lib/rubocop/cop/rspec/factory_girl/dynamic_attribute_defined_statically.rb
```diff
@@ -0,0 +1,92 @@
...
+            if method_uses_parens?(node.location)
+              autocorrect_replacing_parens(node)
+            else
+              autocorrect_without_parens(node)
+            end
+          end
+
+          private
+
+          def into_factories?(node)
```
Darhazer: I don't see where this is used?

lib/rubocop/cop/rspec/factory_girl/dynamic_attribute_defined_statically.rb
```diff
@@ -0,0 +1,92 @@
...
+            node.ancestors.find(&method(:into_factory_block?))
+          end
+
+          def already_defined_in_a_block?(node)
```
Darhazer: I don't see where this is used?

lib/rubocop/cop/rspec/factory_girl/dynamic_attribute_defined_statically.rb
```diff
@@ -0,0 +1,92 @@
...
+            node.sibling_index.zero? && node.parent && node.parent.block_type?
+          end
+
+          def method_uses_parens?(location)
+            return false unless location.begin && location.end
```
Darhazer: what is the case here?

lib/rubocop/cop/rspec/factory_girl/dynamic_attribute_defined_statically.rb
jonatas: Sorry, I forgot to remove. Removing now.

lib/rubocop/cop/rspec/factory_girl/dynamic_attribute_defined_statically.rb
jonatas: Another method disappearing 🎉

lib/rubocop/cop/rspec/factory_girl/dynamic_attribute_defined_statically.rb
jonatas: I observed that when the node location has `begin` and `end` filled when the method call has parenthesis. I just observed that this is true and for some reason when we don't put parenthesis the `begin/end` location is not filled. Then I used it as condition. Is this verification weird?? Should I documented it in somehow? 🤔

backus:
Quick note (I can explain more after work if this isn't clear0 I wrote the badge stuff for rubocop and I'm pretty sure it will only ever sample the last two names in the class definition to build the cop badge (so even if it is `RuboCop::Cop::RSpec::FactoryGirl::Thing` it will never generate a badge like `RSpec/FactoryGirl/Thing`. I'm fine with just shipping `FactoryGirl/CopName` and I think it is also reasonable to have the `RSpec/FactoryGirl` key under `AllCops` to help clarify.

jonatas:
@backus I think now it's ready to go. Feel free to merge it 🎉

backus:
🎈 nice work!

PR #410, opened 2017-05-07
Refactor ExpectOffense implementation

backus:
I have been reworking this code so that I can upstream it to RuboCopproper. See bbatsov/rubocop#4337

This implementation is much simpler. We simply split out the annotation lines from the source and initialize an `AnnotatedSource` object. We can provide an array of offenses to the `AnnotatedSource` instance and get back a new `AnnotatedSource` instance representing the real offenses. `AnnotatedSource` instances then generate an annotated source string (just like what is provided to the matcher) so we can compare the strings and get a nice RSpec diff in the output if they don't match.

Way simpler AND way nicer to work with thanks to RSpec's diffs.

### DIFF
--- a/spec/support/expect_offense.rb
+++ b/spec/support/expect_offense.rb
@@ -1,189 +1,168 @@
 # frozen_string_literal: true
 
+# Mixin for `expect_offense` and `expect_no_offenses`
+#
+# This mixin makes it easier to specify strict offense expectations
+# and a declarative and visual fashion. Just type out the code that
+# should generate a offense, annotate code by writing '^'s
+# underneath each character that should be highlighted, and follow
+# the carets with a string (separated by a space) that is the
+# message of the offense. You can include multiple offenses in
+# one code snippet.
+#
+# @example Usage
+#
+#     expect_offense(<<-RUBY.strip_indent)
+#       a do
+#         b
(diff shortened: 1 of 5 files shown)


spec/support/expect_offense.rb
```diff
@@ -1,189 +1,163 @@
...
-  def expect_offense(source, filename: DEFAULT_FILENAME)
-    expectation = Expectation.new(source)
-    inspect_source(cop, expectation.source, filename)
-    offenses = cop.offenses.map(&method(:to_assertion)).sort
-
-    if expectation.assertions.empty?
-      raise 'Use expect_no_offenses to assert no violations'
+  def expect_offense(source, filename = DEFAULT_FILENAME)
+    expected_annotations = AnnotatedSource.parse(source)
+    if expected_annotations.plain_source == source
```
dgollahon: I think i'd put a blank line on either side of this `if` and possibly another one before the final expectation as this method feels very claustrophobic.

Additionally, you might consider adding an `annotations?` (or similar) predicate to `AnnotatedSource` and push this logic down so that this method is slightly simpler. Just a suggestion though.

backus: Done!

PR #415, opened 2017-05-26
Fix style for RuboCop 0.49.

reitermarkus:


lib/rubocop/cop/rspec/multiple_expectations.rb
```diff
@@ -48,7 +48,7 @@ module RSpec
       class MultipleExpectations < Cop
         include ConfigurableMax
 
-        MSG = 'Example has too many expectations [%{total}/%{max}].'.freeze
+        MSG = 'Example has too many expectations [%<total>s/%<max>s].'.freeze
```
dgollahon: I think these should be `%<..>d`

reitermarkus: Ah, right! I was too focused on fixing the error than on actually using it right.

PR #413, opened 2017-05-22
Add cops for aligning let braces

backus:

PR #418, opened 2017-05-28
Depend on latest RuboCop

backus:
The latest cops (from #413) depend on the new `Cop.autocorrect_incompatible_with` method so that autocorrect does not immeditately clash with the `ExtraSpacing` autocorrect and create an infinite loop. Kinda unfortunate that I need to add such an intense version constraint but we're basically just inheriting the cost of being an extension of a gem that isn't `1.0` 😆 :cry:

I also finished upstreaming our `ExpectOffense` matcher which lets me delete this code from our repo entirely! I do have to add a small patch though to make it nicer to use.


rubocop-rspec.gemspec
```diff
@@ -34,7 +34,7 @@ Gem::Specification.new do |spec|
   spec.test_files = spec.files.grep(%r{^spec/})
   spec.extra_rdoc_files = ['MIT-LICENSE.md', 'README.md']
 
-  spec.add_runtime_dependency 'rubocop', '>= 0.42.0'
+  spec.add_runtime_dependency 'rubocop', '>= 0.49.0'
```
dgollahon: We might consider making this `~> 0.49` or have a `< 1.0` constraint also. Not that it's likely to happen soon, but would be slightly more correct.

Gemfile
```diff
@@ -4,7 +4,6 @@ gemspec
 
 group :test do
   gem 'codeclimate-test-reporter', '~> 1.0.0'
-  gem 'rubocop',                   '~> 0.47'
```
dgollahon: Nice. Weird that we have different requirements. I bet we've had some incompatibilities in the real released gem due to this.

spec/support/expect_offense.rb
```diff
@@ -1,168 +1,17 @@
...
-#     )
-#
-# If you do not want to specify an offense then use the
-# companion method `expect_no_offenses`. This method is a much
-# simpler assertion since it just inspects the source and checks
-# that there were no offenses. The `expect_offenses` method has
-# to do more work by parsing out lines that contain carets.
+# This mixin is the same as rubocop's ExpectOffense except the default
+# filename ends with `_spec.rb`
 module ExpectOffense
```
dgollahon: Great :)

spec/support/expect_offense.rb
```diff
@@ -1,168 +1,17 @@
...
+  include RuboCop::RSpec::ExpectOffense
```
dgollahon: I'm pretty surprised this is exported with the gem, actually.

dgollahon: I wonder if they'll ever make it non-public.

Gemfile
backus: Yeah

rubocop-rspec.gemspec
backus: Possibly but I'm lazy 😛

spec/support/expect_offense.rb
backus: I'd protest if they do. We do have to explicitly `require 'rubocop/rspec/support'` to get it

dgollahon: Oh, ok. That's a good compromise then.

PR #420, opened 2017-05-29
Add MultipleSubjects cop

### LINKED ISSUE #419: Detect multiple subjects inside of example group
Saw something like this recently

```ruby
describe Foo do
  subject(:config) { ... }
  subject(:cop) { ... }
end
```

backus:
This cop checks if example group defines `subject` multiple times.

```ruby
# bad
describe Foo do
  subject(:user) { User.new }
  subject { dead_code }
  subject(:post) { Post.new }
end

# good
describe Foo do
  let(:user) { User.new }
  subject(:post) { Post.new }
end
```

The autocorrect behavior for this cop depends on the type of
duplication:

  - If multiple named subjects are defined then this probably indicates
    that the overwritten subjects (all subjects except the last
    definition) are effectively being used to define helpers. In this
    case they are replaced with `let`.

  - If multiple unnamed subjects are defined though then this can *only*
    be dead code and we remove the overwritten subject definitions.

  - If subjects are defined with `subject!` then we don't autocorrect.
    This is enough of an edge case that people can just move this to
    a `before` hook on their own

Closes #419


backus:
@bquorning I haven't pinged you for review in a while. Could you review this?

bquorning:
Yep, this looks useful :+1:

Darhazer:
@backus I think I found a false positive:

```
it_behaves_like 'user' do
  subject { described_class.new(user, described_class) }

  it { expect(subject).not_to be_accessible
end

it_behaves_like 'admin' do
  subject { described_class.new(user, described_class) }

   it { expect(subject).to be_accessible
end
```

backus:
Good catch you're right

Issue #94, opened 2016-05-12
Cop suggestion: Detect unnecessary usage of `let!`


andyw8:
I often come across code which uses `let!(...)` for setting up state, but then never makes any direct reference them to those identifiers. This setup should be in a `before` block, or inlined into the example's `it` block:

**Bad**

```ruby
let!(:my_widget) { create(:widget) }

it "..." do
  expect(Widget.count).to eq(1)
end
```

**Good**

```ruby
it "..." do
  create(:widget)
  expect(Widget.count).to eq(1)
end
```

```ruby
before { create(:widget) }

it "..." do
  expect(Widget.count).to eq(1)
end
```

nijikon:
👍

backus:
@andyw8 this is implemented in master

backus:
@andyw8, @jcoyne asked:

> Why is this bad? Can you provide some justification? Is this just personal preference?

Care to weigh in? I'll be able to provide context later potentially but I'll let you respond since you originally proposed this cop

jcoyne:
If this is a problem, why hasn't it be proposed for deprecation/removal in rspec itself?

backus:
@jcoyne we aren't flagging all violations. The cop detects when you use `let!` and don't reference the defined method. The cop is encouraging you to only use `let!` when you need to reference the value later and otherwise use `before` for setup

jcoyne:
I guess I don't see what the harm/benefit of that is.  If anything the name is giving some context to the reader of the spec.

backus:
Before I defend the cop further I'll note that these cops are supposed to be opinionated. This is why many are configurable. You're always free to disable particular cases, disable cops for a directory, or disable them entirely.

Consider the following snippet of code

```ruby
def write_user
  user = database.persist(user_info)

  do_some_work
  do_some_other_work

  true
end
```

I think this is similar to the argument you are making about an unreferenced `let!`. I'd say that this local variable assignment to `user` should be removed because it isn't being used in the rest of the method. If I wanted to add some context for that write then I'd add a comment above it. 

Likewise, consider this spec:

```ruby
let!(:user) { database.write_user(user_info) }

before do
  do_some_work
  do_some_other_work
end

it "asserts something" do
  expect(something_that_depends_on_db_state).to be(true)
end
```

[...]

PR #421, opened 2017-06-01
Disable RSpec/LetSetup cop by default

dpisarewski:
See issue #94

### DIFF
--- a/config/default.yml
+++ b/config/default.yml
@@ -156,7 +156,7 @@ RSpec/LetBeforeExamples:
 
 RSpec/LetSetup:
   Description: Checks unreferenced `let!` calls being used for test setup.
-  Enabled: true
+  Enabled: false
 
 RSpec/MessageChain:
   Description: Check that chains of messages are not being stubbed.


backus:
I don't agree with disabling this by default right now but I'm open to further discussion. Basically, in the example you gave

```
describe Product do
  subject { product }

  let!(:product) { create(:product) }

  it { is_expected.to be }

  context 'when product does not exist' do
    let!(:product) { nil }

    it { is_expected.to be_nil }
  end
end
```

1. You can use `subject!` here and `LetSetup` wont flag it
2. This shorthand style (`is_expected.to`) is no longer recommended style and it looks like you are using that as part of the demonstration here for why overriding a `let!` with `nil` is useful

So if you want to do this `let!` style in your own code then that is cool you can just disable it in your config but right now it still seems like a smell compared to just doing something in a `before` hook.

Also, I realize your example was contrived to demonstrate the core of the point. If you could share a real code example that might help more.

dpisarewski:
@backus 
1. Using `subject!` can't really replace `let!`, since only one subject can be defined. Rubocop also doesn't like stubbing on subject or using it explicitly. So, this won't work:

```ruby
describe Product do
  subject! { create(:product) }
  it { is_expected.to be }

  context 'when product does not exist' do
    let!(:product) { nil }

    it { is_expected.to be_nil }
  end

  context 'having invalid data' do
    let(:invalid_data) { #some_data }

    before { allow(product).to receive(:data).and_return(invalid_data) }

    it 'raises an error' do
      expect { product.save! }.to raise_error(SomeValidationError)
    end
  end
end
```

2. I didn't know `is_expected` is considered as bad style. It is recommended by (BetterSpecs)[http://www.betterspecs.org/#let] and is also used in `rubocop-rspec` itself. But `is_expected` wasn't the of the discussion.

3. There is one more reason of preferring lets over before hooks generally: They are lazy-evaluated, that's why better for performance. https://stackoverflow.com/a/5359979/1305200

4. The major problem of having this cop enabled by default is that many developers/teams consider defaults as best practice. It makes any deviation from defaults pretty difficult.

I see no reason to make it enabled by default, IMHO.

dgollahon:
I also have a few thoughts on this:

> Using subject! can't really replace let!, since only one subject can be defined.

You can still override the original subject later on. Probably a little weird for this purpose, but an option.

> There is one more reason of preferring lets over before hooks generally: They are lazy-evaluated, that's why better for performance. https://stackoverflow.com/a/5359979/1305200

This is an argument for `let`, not an argument for using `let!`. `let!` is actually just [`let` with a `before` hook](https://github.com/rspec/rspec-core/blob/f4dc5ef5c6bded2b1ec348e856352b60a3e072e9/lib/rspec/core/memoized_helpers.rb#L355). Its entire purpose is to be a `let` that is _not_ lazily evaluated.

> The major problem of having this cop enabled by default is that many developers/teams consider defaults as best practice.

Well, I think I still agree with @backus that doing a named `let!` is probably less communicative and using more features than you need (if only the `before` is relevant and all you're doing is causing a side effect and not reference the object). So it seems like a good default to say "only use `let!(:thing)` if you are going to reference `thing` by name somewhere.

> It makes any deviation from defaults pretty difficult.

If changing one config entry to false is too difficult, then `rubocop-rspec` may not be for you/your team. We aren't going to get everything right all the time, but we try to have reasonable defaults, but for teams that disagree with our opinions, they're free to configure things as they wish :)

I'm still open to being convinced that this is a bad default, but I am currently not leaning that way.

That said... I don't think your example from https://github.com/backus/rubocop-rspec/issues/94#issuecomment-305413335 should have been flagged by this cop. I would just write `subject!(:product)` before I'd write a `let!(:product)` and then alias it to `subject`. My preference there is beside the point though, since `product` was actually referenced in the `subject` helper meaning that the `let!(:name)` was referenced by name. I think we should probably be more conservative and just look for any matching send in the file since it's hard to track the actual execution of the various helpers statically.

@backus: I think we should consider filing this as a bug^. Thoughts?

Further thoughts on the test design: I think if you have a test where in one case you need a side effect before the example and then you don't in another case, I would use two contexts and only put the before hook in one of the two instead of having a global `let!` and then another `let!` overriding the original `let!` to be a no-op. Also, if you are running into this cop frequently, you might consider re-evaluating your test suite. The projects I work on have few or no `let!` because usually a lazily evaluated `let` is preferable. If you need to have a lot of global side effects to get your test suite to work, you might want to consider more isolated unit tests / refactoring your system.

dgollahon:
Maybe we should improve the error message for `LetSetup` instead. It sounds like we're saying "never use this", but I think it is only supposed to flag when it is used as if it were a `before`. We could say something more like "Do not use `let!` for test setup. Either reference the value by name, or move the setup logic into a `before` hook." (wording could probably still be improved).

backus:
> @backus: I think we should consider filing this as a bug^. Thoughts?

Yes now that I look at it with that perspective, not recognizing the reference in `let!` is a bug.

backus:
@dpisarewski I'm going to close this for now and open a separate bug report for the case your example points out. I'm not quite sure what else there is to discuss but it seems that you might also be misunderstanding the use case of `let!` or exactly what `LetSetup` is talking about. The bug will be fixed and if you still don't like the cop you can disable it :smile:

nijikon:
@backus can you point me to a discussion/documentation in which it is written that `This shorthand style (is_expected.to) is no longer recommended style`?

backus:
@nijikon see https://github.com/backus/rubocop-rspec/issues/228

> > ImplicitExpect
> The default style for this should be to not use it at all, it's a legacy hang around.

halilim:
A little late to the party. I usually encounter this when I want to test that a method does not return something. Right now I'm trying to test an Active Record scope (renamed and simplified):

```ruby
describe '.active' do
  let!(:inactive) { FactoryGirl.create(:thing, active: false) }
  let!(:active) { FactoryGirl.create(:thing, active: true) }

  it 'returns only active records' do
    expect(described_class.active).to eq([active])
  end
end
```

Putting the creation of the inactive record into a `before` block would fix this; but that way it doesn't look as descriptive to me, i.e. with one of the records among `let` blocks and another in the `before` block and without a name.

I'd appreciate any suggestion that would fix this in a descriptive and compliant way.

mikegee:
@halilim I prefer using fewer RSpec features:

```ruby
describe '.active' do
  it 'returns only active records' do
    inactive = FactoryGirl.create(:thing, active: false)
    active   = FactoryGirl.create(:thing, active: true)

    expect(described_class.active).to eq([active])
  end
end
```

halilim:
@mikegee Dangers of my simplification :) There might be more than one example using some of those records (and Rubocop says "Useless assignment to variable - \`inactive\`", but it can be fixed by prefixing with `_`).

RobinDaugherty:
I've written plenty of specs where there are examples sharing records exactly as @halilim describes. Placing them in the example is a bad idea, in addition to the fact that it's a useless assignment to a variable. Please disable this cop by default. While there is one argument for this rule, there are numerous reasons against it, and even [the PR itself](https://github.com/rubocop-hq/rubocop-rspec/issues/94) has nothing but downvotes.

Issue #392, opened 2017-03-28
Two equivalent stubs using different syntax


bquorning:
I just came across a spec that had (the equivalent of) the following lines:
```rb
allow(:foo).to receive(:bar) { "baz" }
allow(:foo).to receive(:bar).and_return("baz")
```

It would be nice if I had a cop to detect this kind of duplication. Note that there’s a subtle difference between the two lines, but in this case it doesn’t matter.

backus:
Yes I was thinking this recently. It would be nice to also enforce one style (except in cases where the value inside of the block in the first style is dynamic).

Darhazer:
I can implement such. The one of the supported option is obviously `block`; how should be called the other one?

Drenmi:
Can this be closed as of the merging of #422? 🙂

PR #422, opened 2017-06-03
Add a cop to enforce style for defining stub's return value. Ref #392

Darhazer:
@bquorning @backus  please review.

I'm open for better names as well.
Note: Didn't implement autocorrect, as I had trouble removing the whitespace, e.g.
`allow(Foo).to receive(:bar) { 42 }` to change into `allow(Foo).to receive(:bar).and_return(42)` (I got `allow(Foo).to receive(:bar). and_return(42)` instead)


lib/rubocop/cop/rspec/return_from_stub.rb
```diff
@@ -0,0 +1,84 @@
...
+          return unless style == :block
+
+          and_return_value(node) do |args|
+            add_offense(node, :expression, MSG_BLOCK) unless dynamic?(args)
+          end
+        end
+
+        private
+
+        def dynamic?(node)
```
backus: We should probably have this method instead check to see if the value only contains literals. The `YodaCondition` cop had a lot of issues because it tried to blacklist dynamic nodes instead of whitelist static nodes: https://github.com/bbatsov/rubocop/issues?q=is%3Aissue+YodaCondition+label%3Abug

Darhazer:
Updated

backus:
🎈

lib/rubocop/cop/rspec/return_from_stub.rb
```diff
@@ -0,0 +1,83 @@
+# frozen_string_literal: true
+
+module RuboCop
+  module Cop
+    module RSpec
+      # Checks for consistent style of stub's return setting.
+      #
+      # Enforces either `and_return` or block-style return in the cases
+      # where the returned value is constant. Ignores dynamic returned values
+      # are the result would be different
```
Darhazer: Fixed in the next commit

PR #426, opened 2017-06-07
Add auto-correct support to ReturnFromStub cop

Darhazer:
With little digging (and copying) in the Rubocop source, I've managed to implement auto-correct support for the new ReturnFromStub cop.

### DIFF
--- a/lib/rubocop/cop/rspec/return_from_stub.rb
+++ b/lib/rubocop/cop/rspec/return_from_stub.rb
@@ -7,9 +7,9 @@ module RuboCop
       #
       # Enforces either `and_return` or block-style return in the cases
       # where the returned value is constant. Ignores dynamic returned values
-      # are the result would be different
+      # as the result would be different.
       #
-      # This cop can be configured using the `EnforcedStyle` option
+      # This cop can be configured using the `EnforcedStyle` option.
       #
       # @example `EncorcedStyle: block`
       #   # bad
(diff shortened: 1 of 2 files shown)


lib/rubocop/cop/rspec/return_from_stub.rb
```diff
@@ -49,7 +49,7 @@ class ReturnFromStub < Cop
 
         def_node_matcher :and_return_value, <<-PATTERN
             (send
-              (send nil :receive (...)) :and_return $(...)
+              (send nil :receive $(...)) :and_return $(...)
```
Darhazer: It would be better if I could match only the and_return() part, this way offense won't mark `receive(:msg)` part and the code would be cleaner. If anyone have suggestion how to do so, please share

backus: Could you pass in `:selector` instead of expression to your add_offense call?

Darhazer:
Updated

lib/rubocop/cop/rspec/return_from_stub.rb
```diff
@@ -77,6 +87,36 @@ def dynamic?(node)
...
+          block_range = range_between(begin_pos_for_replacement(node),
+                                      node.loc.end.end_pos)
+          range_with_surrounding_space(block_range, :left)
+        end
+
+        def begin_pos_for_replacement(node)
+          block_send_or_super, _block_args, _block_body = *node
+          expr = block_send_or_super.source_range
+
+          if (paren_pos = (expr.source =~ /\(\s*\)$/))
```
backus: Are we able to do this without regexp matching on the source?

Darhazer:
Closing in favor of #495

PR #430, opened 2017-06-21
[Feature] Add style guide links

maschwenk:
Add style guide links for each cop to the RubyDoc documentation site.

This is great for IDE/text editor integration and means that users don't have to copy/paste `RSpec/CopName` into Google to find documentation. Because you guys already use RubyDoc, we basically get this for $free.99.

Rubocop uses the same approach:

https://github.com/bbatsov/rubocop/blame/master/config/disabled.yml#L29-L30
____

#### On the CLI:
```ruby
bundle exec rubocop --display-style-guide
-- > C: RSpec/SubjectStub: Do not stub your test subject. 

(http://www.rubydoc.info/gems/rubocop-rspec/RuboCop/Cop/RSpec/SubjectStub)
```

#### In editor (the link icon takes you to the Rubydoc documentation):
image

### DIFF
--- a/config/default.yml
+++ b/config/default.yml
@@ -12,22 +12,27 @@ AllCops:
 RSpec/AnyInstance:
   Description: Check that instances are not being stubbed globally.
   Enabled: true
+  StyleGuide: http://www.rubydoc.info/gems/rubocop-rspec/RuboCop/Cop/RSpec/AnyInstance
 
 RSpec/AroundBlock:
   Description: Checks that around blocks actually run the test.
   Enabled: true
+  StyleGuide: http://www.rubydoc.info/gems/rubocop-rspec/RuboCop/Cop/RSpec/AroundBlock
 
 RSpec/AlignLeftLetBrace:
   Description: Checks that left braces for adjacent single line lets are aligned.
   Enabled: false
+  StyleGuide: http://www.rubydoc.info/gems/rubocop-rspec/RuboCop/Cop/RSpec/AlignLeftLetBrace
 
 RSpec/AlignRightLetBrace:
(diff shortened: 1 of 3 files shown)


arcwhite:
👍

arcwhite:
👍

backus:
This is a neat idea, thank you! I'm curious though if this clobbers the main `StyleGuideBaseURL` value for RuboCop. Do you know?

maschwenk:
@backus You are right, it does clobber. Let me think of a workaround (if any)

backus:
I believe the workaround would be to just have the base url in all of our cop docs. We could have a buildstep that checks this I guess

maschwenk:
@backus looks like if you just define the entire URL, then base url config from Rubocop is ignored 😄 . 
image

backus:
Great!

maschwenk:
Gonna open/search for issue in Rubocop, wonder if there's a cleaner way to do it, but I hope that's not a blocker!

backus:
Not a blocker at all. This whole gem is technically a violation of public API heh. As long as something in the build confirms that all the links are using the proper base url. Seems like it would fit in nicely with the config generator

backus:
Looks great! Could you squash your commits down @maschwenk? I'll merge after that

maschwenk:
@backus done. You know you can configure that repo-wide? That's what I do at work because I'm very lazy ;)

backus:
@maschwenk Yeah I should consider just doing that at some point. I also know some people care about squashing down their work in a certain way that gives it a clean description.

Thanks for the contribution!

PR #434, opened 2017-07-18
Drop comma in def_node_search

pocke:
In bbatsov/rubocop's master branch, comma is not allowed in node pattern.

See. https://github.com/bbatsov/rubocop/pull/4499


backus:
Oh good catch. Thanks!

PR #437, opened 2017-07-21
Use eval_gemfile instead of eval

bquorning:
Bundler has a method `eval_gemfile` that allows to *not* call `eval` in our Gemfile.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -10,5 +10,5 @@ end
 local_gemfile = 'Gemfile.local'
 
 if File.exist?(local_gemfile)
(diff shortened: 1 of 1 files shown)


backus:
Just so you know @bquorning: https://github.com/bbatsov/rubocop/issues/3903#issuecomment-272712779

backus:
I'm fine with using the method though

bquorning:
I’ll leave the decision to you 😄  Like you, I have long considered the (public) method a part of Bundler’s API. But seeing on of the maintainers saying that it’s not (and therefore undocumented) makes me question my beliefs.

PR #438, opened 2017-07-21
Fix changed RuboCop methods

### LINKED ISSUE #435: Test is failed with RuboCop HEAD
A method signature of `autocorrect_source` has been changed  by https://github.com/bbatsov/rubocop/pull/4510 .
So the test for rubocop-rspec is failed with next version RuboCop.

This problem can be avoided by one of the following methods.

1. Drop rubocop 0.49.1 or older support if rubocop 0.50.0 is released
1. Pin rubocop version with 0.49.1 or older
1. Check `Method#arity` for `autocrrect_source` method.
  - This way supports both rubocop versions, but it's a workaround.

bquorning:
Fixes #435 :crossed_fingers:

### DIFF
--- a/lib/rubocop/cop/rspec/repeated_example.rb
+++ b/lib/rubocop/cop/rspec/repeated_example.rb
@@ -40,7 +40,7 @@ module RuboCop
           key_parts = [example.metadata, example.implementation]
 
           if example.definition.method_name == :its
-            key_parts << example.definition.method_args
+            key_parts << example.definition.arguments
           end
 
           key_parts
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -11,7 +11,7 @@ Metrics/BlockLength:
     - '**/*.rake'
     - spec/**/*.rb
 
-Style/FileName:
+Naming/FileName:
   Exclude:
     - lib/rubocop-rspec.rb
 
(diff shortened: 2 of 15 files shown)


lib/rubocop/cop/rspec/repeated_example.rb
```diff
@@ -40,7 +40,7 @@ def example_signature(example)
           key_parts = [example.metadata, example.implementation]
 
           if example.definition.method_name == :its
-            key_parts << example.definition.method_args
+            key_parts << [example.definition.first_argument]
```
bquorning: I’m not too sure the new code is equivalent to the old.

backus: if this is a block type I think you could also destructure and use the `method_args` method I think it is still there on `SendNode` types

bquorning: `example.definition.arguments` seems to work.

backus: Nice

bquorning:
Green build, now using RuboCop v0.50.0.

backus:
Great. If you'd like to do another release go ahead

Issue #378, opened 2017-03-22
Cop idea: detect not_to have_selector capybara matchers


Darhazer:
The `have_*` collection of capybara matchers has the same set of `have_no` matches, for checking that there is no content matching. This creates the option to write the expectation in two forms:

`expect(page).to have_no_selector`
`expect(page).not_to have_selector`

For consistency, it would be nice to have a cop that enforces only one of the styles.

This goes for all of:
```
have_button
have_checked_field
have_css
have_field
have_link
have_select
have_selector
have_table
have_text
have_unchecked_field
have_xpath
```

Additionally the cop could detect usage of the node_matchers with eq false:
`expect(has_selector?('#something')).to eq false`

timrogers:
Are we sure this is a problem? The Capybara readme says the following:

```

Capybara's RSpec matchers, however, are smart enough to handle either form.
The two following statements are functionally equivalent:

```ruby
expect(page).not_to have_xpath('a')
expect(page).to have_no_xpath('a')
```
```

Darhazer:
I can't prove (although I'm pretty sure they have different behavior), but even if that is the case, it still good to have cop enforcing consistent usage. Maybe it should be configurable if one prefers the `not_to` style

timrogers:
I'm not going to have time to jump on this for now I'm afraid!

Darhazer:
Anyway your input was very valuable, thank you for looking at this

ybiquitous:
This is a very nice suggestion and must be useful for so many developers!

I found a similar warning on the ["Capybara cheatsheet"](https://devhints.io/capybara):

> Use should have_no_* versions with RSpec matchers because should_not have_* doesn’t wait for a timeout from the driver.

![image](https://user-images.githubusercontent.com/473530/47366698-32215680-d719-11e8-9858-c431c5d807f0.png)

ybiquitous:
A similar warning on the Codeship blog:

> ...
> That brings us to a common mistake: flipping the assertion instead of the finder, like this: `refute page.has_content?("Dashboard")`. 
> ...
> Which means that this is 15 seconds (the default) slower than `assert page.has_no_content?("Dashboard")`.

https://blog.codeship.com/faster-rails-tests/

badosu:
Just for reference I tested a big codebase refactor from one form to another and did not see big difference improvements.

[...]

PR #439, opened 2017-08-05
Add a cop to check for consistent method usage in feature specs.

rspeicher:


tsigo:
I started down the path of making this `ConfigurableEnforcedStyle`, with `rspec` and `capybara` options.

If someone were to set `capybara`, we'd have to make sure to only check:

1. Files in `spec/features`
1. Files with `type: :feature` (and possibly just `:feature`?) metadata

I wasn't sure of the best way to handle those two conditions, so I'm open to feedback on making this configurable.

lib/rubocop/cop/rspec/feature_methods.rb
```diff
@@ -0,0 +1,65 @@
...
+      #     end
+      #
+      #     it 'with OAuth' do
+      #       # ...
+      #     end
+      #   end
+      class FeatureMethods < Cop
+        MSG = 'Use `%s` instead of `%s`.'.freeze
+
+        # https://git.io/v7rLu
```
backus: If you hit `y` on your keyboard when visiting this link then github will give you a link to a sha that will never break. https://help.github.com/articles/getting-permanent-links-to-files/

tsigo: Sure, updated.

bquorning:
I’ve been trying to find out *why* capybara provides these aliased methods. [The README](https://github.com/teamcapybara/capybara#the-dsl) just says that “Capybara also comes with a built in DSL for creating descriptive acceptance tests […] `feature` is in fact just an alias for `describe ..., :type => :feature`, `background` is an alias for `before`, `scenario` for `it`, and `given`/`given!` aliases for `let`/`let!`, respectively.” without any explanation for why this DSL should be preferable.

Also, I just want to note that these aliases are only available if Capybara (or rspec-rails) is loaded. Which puts this cop in the same category as `RSpec::FactoryGirl::DynamicAttributeDefinedStatically` as “not really an RSpec cop, but close”. Perhaps we should have an `RSpec::Capybara` module?

rspeicher:
@bquorning I believe they're just syntactic sugar to make feature specs read more like [Cucumber](https://github.com/cucumber/cucumber/blob/master/docs/gherkin.md) features.

Darhazer:
Yup, this should be in RSpec::Capybara module. There are few other things we could implement for capybara (like #378 for example)

rspeicher:
Moved the copy to a new `Capybara` namespace.

spec/project/default_config_spec.rb
```diff
@@ -6,10 +6,11 @@
   let(:cop_names) do
     namespaces = {
       'rspec' => 'RSpec',
+      'capybara' => 'Capybara',
       'factory_girl' => 'FactoryGirl'
     }
     glob = SpecHelper::ROOT.join('lib', 'rubocop', 'cop',
-                                 'rspec', '{,factory_girl/}*.rb')
+                                 'rspec', '{capybara/,,factory_girl/}*.rb')
```
rspeicher: Requiring these two changes might get messy quickly if you anticipate adding more namespaces, but I went with the simplest thing for now.

rspeicher:
Can't see what the build failures are since they're private.

Darhazer:
@rspeicher you can run `rake` on your local machine to check the build. I think you should add `Capybara` to the namespaces in `ConfigFormatter` and fix the build_config script to include capybara
```
-cops = '{*.rb,factory_girl/*.rb}'
+cops = '{*.rb,factory_girl/*.rb,capybara/*.rb}'
```

bquorning:
Ahh, I think you need to add `,capybara/*.rb` to bin/build_config for the `confirm_config` test to pass.

bin/build_config
```diff
@@ -8,7 +8,7 @@ require 'rubocop-rspec'
 require 'rubocop/rspec/description_extractor'
 require 'rubocop/rspec/config_formatter'
 
-cops = '{*.rb,factory_girl/*.rb}'
+cops = '{*.rb,capybara/*.rb,factory_girl/*.rb}'
```
rspeicher: This could be simplified to just `**/*.rb`.

rspeicher:
Back to green, thanks y'all!

rspeicher:
Rebased since 1.16.0 was released.

lib/rubocop/rspec/config_formatter.rb
```diff
@@ -22,7 +22,7 @@ def unified_config
         cops.each_with_object(config.dup) do |cop, unified|
           unified[cop] = config.fetch(cop)
             .merge(descriptions.fetch(cop))
-            .merge('StyleGuide' => STYLE_GUIDE_BASE_URL + cop)
+            .merge('StyleGuide' => STYLE_GUIDE_BASE_URL + cop.sub('RSpec/', ''))
```
rspeicher: Would you prefer an anchored regex here?

Darhazer:
I think it's ready for merge. @backus final words?

backus:
LGTM. @rspeicher or @Darhazer (whoever does the merge) can you squash down the commits?

PR #440, opened 2017-08-10
Add new `RSpec/VoidExpect` cop

pocke:
This cop checks void `expect()` usage.

If `expect()` is used without `.to()` or `.not_to()`, the `expect()` does not make sense. But RSpec does not raise error / warning.

For example:

```ruby
it 'something' do
  # bad: it does nothing.
  expect(something)
end
```

This cop checks the problem.


backus:
Nice!

PR #441, opened 2017-08-11
Add new `RSpec/InvalidPredicateMatcher` cop

pocke:
RSpec defines predicate matchers automatically.

```ruby
expect(something.available?).to eq true
expect(something).to be_available
```

I sometimes write 'expect(something).to be_available?` by mistake. This cop detects the mistake.


bquorning:
> I sometimes write 'expect(something).to be_available?` by mistake.

In that case, won’t RSpec tell you about the mistake? Something like
```
expected something to respond to `available??`
```

pocke:
> In that case, won’t RSpec tell you about the mistake? Something like

Yes, but I can notice this problem more sooner by this cop, because I run RuboCop more frequently than test. 
And this cop tells me about mistake more descriptively.

lib/rubocop/cop/rspec/invalid_predicate_matcher.rb
```diff
@@ -0,0 +1,42 @@
...
+          invalid_predicate_matcher?(node) do |predicate|
+            add_offense(predicate, :expression)
+          end
+        end
+
+        private
+
+        def predicate?(name)
+          name = name.to_s
+          name.start_with?('be_') && name.end_with?('?')
```
backus: I think it can also start with `have_`

lib/rubocop/cop/rspec/invalid_predicate_matcher.rb
```diff
@@ -0,0 +1,42 @@
...
+      #   # bad
+      #   expect(foo).to be_something?
+      #
+      #   # good
+      #   expect(foo).to be_something
+      class InvalidPredicateMatcher < Cop
+        MSG = 'Omit `?` from `%<matcher>s`.'.freeze
+
+        def_node_matcher :invalid_predicate_matcher?, <<-PATTERN
+          (send (send nil :expect ...) {:to :not_to} $(send nil #predicate?))
```
backus: should also have `to_not`

bquorning:
👎 IMHO, Rubocop is not the right tool for solving this problem. The extra `?` at the end of the predicate matcher is a typo, and as all (well, most) other typos it will be caught during runtime.

dgollahon:
I'm not completely opposed to this, but it does seem like `rspec` already has this covered as @bquorning pointed out. Whatever you think though, @backus.

pocke:
@backus I fixed about `to_not` and `have_...?`.


I created a pull request about this on rspec-expectations also. https://github.com/rspec/rspec-expectations/pull/1008
So this cop may not be needed if this pull-request is merged. WDYT?

backus:
I still think it is probably a net positive to have small helpful cops like this. For example, I imagine it is still helpful for people with editor integrations. Thanks @pocke

PR #443, opened 2017-08-17
Change HookArgument cop to detect when hook has a receiver

pocke:
RSpec hook has a receiver in `spec_helper.rb`

```ruby
RSpec.configure do |config|
  config.before :each do
    something
  end
end
```

But currently this cop does not check hook that has a receiver.
This cop will analyse such hooks by this change.


backus:
Nice!

PR #444, opened 2017-08-18
Drop a comma from node pattern

pocke:
Ref: #434 

This comma is unnecessary. Sorry it is my mistake.

### DIFF
--- a/lib/rubocop/cop/rspec/predicate_matcher.rb
+++ b/lib/rubocop/cop/rspec/predicate_matcher.rb
@@ -154,7 +154,7 @@ module RuboCop
         def_node_matcher :predicate_matcher_block?, <<-PATTERN
           (block
             (send
-              (send nil :expect, $!nil)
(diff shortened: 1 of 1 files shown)


bquorning:
:+1: We should have a cop checking this.

backus:
Would be great to get an internal_investigation cop for this in rubocop proper that we could use too

bquorning:
> Would be great to get an internal_investigation cop for this in rubocop proper that we could use too

This has already been fixed in Rubocop master; see bbatsov/rubocop#4499 and bbatsov/rubocop#4501 by @Drenmi. Commas will now result in a pattern compile error.

backus:
Oh great. We probably don't need a cop then once that version of rubocop is release

Drenmi:
Sorry. Should have done a quick sweep through `rubocop-rspec` when I removed the support for arbitrary commas. 🙂

PR #446, opened 2017-08-20
Count expect_any_instance_of towards expectations count

Darhazer:
Inspired by recent @pocke contributions, I've checked where Language::Expectations could be useful.


backus:
Awesome! Thanks for helping keep things DRY :smile:

PR #450, opened 2017-09-01
Add FactoryGirl/CreateList cop. Fixes #411

### LINKED ISSUE #411: Cop idea: [FactoryGirl] replaces n.times { create } with create_list
Now that we have the first candidate for a FactoryGirl specific cop, here's another one
I often see people use `10.times { create :user }`, while it can be done with `create_list :user, 10`
Note: in the case the block parameter of time is used, create_list is no longer valid approach, like in
`10.times { |n| create :user, registered_at: n.days.ago }`

Darhazer:
Fixes #411 

create_list should be preferred over `n.times { :create }` (also over any loop that creates the same kind of object, but that's not implemented; n.times is the most common way I've encountered) as:
* it's provided by the FactoryGirl DSL
* it actually runs map, hence returning the created instances
* Just to enforce consistency

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 * Fix `RSpec/NestedGroups` cop support --auto-gen-config. ([@walf443][])
 * Fix false positives in `Capybara/FeatureMethods` when feature methods are used as property names in a factory. ([@Darhazer][])
 * Allow configuring enabled methods in `Capybara/FeatureMethods`. ([@Darhazer][])
+* Add `FactoryBot/CreateList` cop. ([@Darhazer][])
 
 ## 1.24.0 (2018-03-06)
 
--- a/config/default.yml
+++ b/config/default.yml
@@ -380,6 +380,15 @@ Capybara/FeatureMethods:
   EnabledMethods: []
   StyleGuide: http://www.rubydoc.info/gems/rubocop-rspec/RuboCop/Cop/RSpec/Capybara/FeatureMethods
 
+FactoryBot/CreateList:
+  Description: Checks for create_list usage.
+  Enabled: true
+  EnforcedStyle: create_list
+  SupportedStyles:
+  - create_list
+  - n_times
+  StyleGuide: http://www.rubydoc.info/gems/rubocop-rspec/RuboCop/Cop/RSpec/FactoryBot/CreateList
+
 FactoryBot/DynamicAttributeDefinedStatically:
   Description: Prefer declaring dynamic attribute values in a block.
   Enabled: true
--- a/lib/rubocop/cop/rspec/factory_bot/create_list.rb
+++ b/lib/rubocop/cop/rspec/factory_bot/create_list.rb
@@ -1,0 +1,148 @@
+# frozen_string_literal: true
+
+module RuboCop
+  module Cop
+    module RSpec
+      module FactoryBot
+        # Checks for create_list usage.
+        #
+        # This cop can be configured using the `EnforcedStyle` option
+        #
+        # @example `EnforcedStyle: create_list`
+        #   # bad
+        #   3.times { create :user }
+        #
+        #   # good
+        #   create_list :user, 3
+        #
+        #   # good
+        #   3.times { |n| create :user, created_at: n.months.ago }
+        #
+        # @example `EnforcedStyle: n_times`
+        #   # bad
+        #   create_list :user, 3
+        #
+        #   # good
+        #   3.times { create :user }
+        class CreateList < Cop
+          include ConfigurableEnforcedStyle
+
+          MSG_CREATE_LIST = 'Prefer create_list.'.freeze
+          MSG_N_TIMES = 'Prefer %<number>s.times.'.freeze
+
+          def_node_matcher :n_times_block?, <<-PATTERN
+            (block
+              (send (int _) :times)
+              ...
+            )
+          PATTERN
+
+          def_node_matcher :factory_call, <<-PATTERN
+            (send ${(const nil? {:FactoryGirl :FactoryBot}) nil?} :create (sym $_) $...)
+          PATTERN
+
+          def_node_matcher :factory_list_call, <<-PATTERN
+            (send ${(const nil? {:FactoryGirl :FactoryBot}) nil?} :create_list (sym $_) (int $_) $...)
+          PATTERN
+
+          def on_block(node)
+            return unless style == :create_list
+            return unless n_times_block?(node)
+            return unless contains_only_factory?(node.body)
+
+            add_offense(node.send_node,
+                        location: :expression, message: MSG_CREATE_LIST)
+          end
+
+          def on_send(node)
+            return unless style == :n_times
+
+            factory_list_call(node) do |_receiver, _factory, count, _|
+              add_offense(
+                node,
+                location: :selector,
+                message: format(MSG_N_TIMES, number: count)
+              )
+            end
+          end
+
+          def autocorrect(node)
+            if style == :create_list
+              autocorrect_n_times_to_create_list(node)
+            else
+              autocorrect_create_list_to_n_times(node)
+            end
+          end
+
+          private
+
+          def contains_only_factory?(node)
+            if node.block_type?
+              factory_call(node.send_node)
+            else
+              factory_call(node)
+            end
+          end
+
+          def autocorrect_n_times_to_create_list(node)
+            block = node.parent
+            count = block.receiver.source
+            replacement = factory_call_replacement(block.body, count)
+
+            lambda do |corrector|
+              corrector.replace(block.loc.expression, replacement)
+            end
+          end
+
+          def autocorrect_create_list_to_n_times(node)
+            replacement = generate_n_times_block(node)
+            lambda do |corrector|
+              corrector.replace(node.loc.expression, replacement)
+            end
+          end
+
+          def generate_n_times_block(node)
+            receiver, factory, count, options = *factory_list_call(node)
+
+            arguments = ":#{factory}"
+            options = build_options_string(options)
+            arguments += ", #{options}" unless options.empty?
+
+            replacement = format_receiver(receiver)
+            replacement += format_method_call(node, 'create', arguments)
+            "#{count}.times { #{replacement} }"
+          end
+
+          def factory_call_replacement(body, count)
+            receiver, factory, options = *factory_call(body)
+
+            arguments = ":#{factory}, #{count}"
+            options = build_options_string(options)
+            arguments += ", #{options}" unless options.empty?
+
+            replacement = format_receiver(receiver)
+            replacement += format_method_call(body, 'create_list', arguments)
+            replacement
+          end
+
+          def build_options_string(options)
+            options.map(&:source).join(', ')
+          end
+
+          def format_method_call(node, method, arguments)
+            if node.parenthesized?
+              "#{method}(#{arguments})"
+            else
+              "#{method} #{arguments}"
+            end
+          end
+
+          def format_receiver(receiver)
+            return '' unless receiver
+            "#{receiver.source}."
+          end
+        end
+      end
+    end
+  end
+end
--- a/spec/rubocop/cop/rspec/factory_bot/create_list_spec.rb
+++ b/spec/rubocop/cop/rspec/factory_bot/create_list_spec.rb
@@ -1,0 +1,140 @@
+RSpec.describe RuboCop::Cop::RSpec::FactoryBot::CreateList, :config do
+  subject(:cop) { described_class.new(config) }
+
+  let(:cop_config) do
+    { 'EnforcedStyle' => enforced_style }
+  end
+
+  context 'when EnforcedStyle is :create_list' do
+    let(:enforced_style) { :create_list }
+
+    it 'flags usage of n.times with no arguments' do
+      expect_offense(<<-RUBY)
+        3.times { create :user }
+        ^^^^^^^ Prefer create_list.
+      RUBY
+    end
+
+    it 'flags usage of n.times when FactoryGirl.create is used' do
+      expect_offense(<<-RUBY)
+        3.times { FactoryGirl.create :user }
+        ^^^^^^^ Prefer create_list.
+      RUBY
+    end
+
+    it 'flags usage of n.times when FactoryBot.create is used' do
+      expect_offense(<<-RUBY)
+        3.times { FactoryBot.create :user }
+        ^^^^^^^ Prefer create_list.
+      RUBY
+    end
+
+    it 'ignores create method of other object' do
+      expect_no_offenses(<<-RUBY)
+        3.times { SomeFactory.create :user }
+      RUBY
+    end
+
+    it 'ignores create in other block' do
+      expect_no_offenses(<<-RUBY)
+        allow(User).to receive(:create) { create :user }
+      RUBY
+    end
+
+    it 'flags n.times with argument' do
+      expect_offense(<<-RUBY)
+        3.times { |n| create :user, created_at: n.days.ago }
+        ^^^^^^^ Prefer create_list.
+      RUBY
+    end
+
+    it 'ignores n.times when there is no create call inside' do
+      expect_no_offenses(<<-RUBY)
+        3.times { do_something }
+      RUBY
+    end
+
+    it 'ignores n.times when there is other calls but create' do
+      expect_no_offenses(<<-RUBY)
+        used_passwords = []
+        3.times do
+          u = create :user
+          expect(used_passwords).not_to include(u.password)
+          used_passwords << u.password
+        end
+      RUBY
+    end
+
+    it 'flags FactoryGirl.create calls with a block' do
+      expect_offense(<<-RUBY)
+        3.times do
+        ^^^^^^^ Prefer create_list.
+          create(:user) { |user| create :account, user: user }
+        end
+      RUBY
+    end
+
+    include_examples 'autocorrect',
+                     '5.times { create :user }',
+                     'create_list :user, 5'
+
+    include_examples 'autocorrect',
+                     '5.times { create(:user, :trait) }',
+                     'create_list(:user, 5, :trait)'
+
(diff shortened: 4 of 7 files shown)


lib/rubocop/cop/rspec/factory_girl/create_list.rb
```diff
@@ -0,0 +1,67 @@
...
+
+          def_node_matcher :times_block?, <<-PATTERN
+          (block
+            n_times
+            ...
+          )
+          PATTERN
+
+          def_node_matcher :factory_call?, <<-PATTERN
+            (send nil :create (sym $_) $...)
```
mikegee: Some apps I've worked on don't [include `FactoryGirl::Syntax::Methods `](https://github.com/thoughtbot/factory_girl/blob/master/GETTING_STARTED.md#rspec) to enable this shorthand syntax and call `FactoryGirl.create` instead.

Darhazer:
PR updated

mikegee:
Should this cop enforce usage of the other [`*_list`](http://www.rubydoc.info/gems/factory_girl/FactoryGirl/Syntax/Methods) methods, too?

Darhazer:
I don't think anyone uses `build` inside of n.times as without reference to the built object you can't do anything useful. That's one difference between using n.times and create_list - the later returns the created objects

mikegee:
Good point

Darhazer:
Rebased on master

Darhazer:
@bquorning can you review this?

bquorning:
If you add a PR description I will :-)

config/default.yml
```diff
@@ -6,6 +6,7 @@ AllCops:
     - "(?:^|/)spec/"
   RSpec/FactoryGirl:
     Patterns:
+    - _spec.rb
```
bquorning: What does this do?

lib/rubocop/cop/rspec/factory_girl/create_list.rb
```diff
@@ -0,0 +1,72 @@
...
+        #   create_list :user, 3
+        #
+        #   # good
+        #   3.times { |n| create :user, created_at: n.months.ago }
+        class CreateList < Cop
+          MSG = 'Prefer create_list.'.freeze
+
+          def_node_matcher :n_times, '(send (int $_) :times)'
+
+          def_node_matcher :times_block?, <<-PATTERN
```
bquorning: This matcher isn’t being used?

lib/rubocop/cop/rspec/factory_girl/create_list.rb
```diff
@@ -0,0 +1,72 @@
...
+            ...
+          )
+          PATTERN
+
+          def_node_matcher :factory_call?, <<-PATTERN
+            (send ${(const nil :FactoryGirl) nil} :create (sym $_) $...)
+          PATTERN
+
+          def on_block(node)
+            receiver, args, body = *node
```
bquorning: Should we check that this is a `times` block? I guess that’s what `times_block?` was meant for…

config/default.yml
Darhazer: The first FactoryGirl cop checked just the factories, so it ignored the actual spec files. I need it to test them.

lib/rubocop/cop/rspec/factory_girl/create_list.rb
Darhazer: Nice catch. I'll add a test for this

config/default.yml
Darhazer: Actually seems unneeded

Darhazer:
PR updated

lib/rubocop/cop/rspec/factory_girl/create_list.rb
```diff
@@ -0,0 +1,74 @@
...
+        #
+        #   # good
+        #   3.times { |n| create :user, created_at: n.months.ago }
+        class CreateList < Cop
+          MSG = 'Prefer create_list.'.freeze
+
+          def_node_matcher :n_times, '(send (int $_) :times)'
+
+          def_node_matcher :times_block_without_args?, <<-PATTERN
+          (block
```
bquorning: Indentation is missing. Isn’t `Layout/IndentHeredoc` supposed to catch this?

Darhazer:
> Besides create, shouldn’t we enforce the same style for build, build_stubbed and attributes_for?

Only create could be used in this way in a loop. For the rest you need to store the result of a method somewhere, otherwise it just doesn't have a side effect. It's possible that someone does this in a loop and uses a variable, like `3.times { data << attributes_for(:user) }` but those are much harder to catch.

nevertheless, thanks for pointing the documentation again, I will add some tests for create used with a block.

bquorning:
How about e.g. `3.times.map { build(:user) }`?

Darhazer:
Wow, I've never seen such code.

Thanks for the feedback @bquorning , I will work on improving the cop in the next couple of weeks. It will need a betetr name as well, as soon as it's not only for create_list.

bquorning:
> Wow, I've never seen such code.

Neither have I 😃  And I don’t know if we should cover every possible case in this PR, or if we should wait and see if our users report an actual need. Probably the latter is fine.

backus:
Also I'm pretty sure `n.times.map { x }` is already autocorrected to `Array.new(n) { x }` by rubocop

backus:
Also I'm pretty sure `n.times.map { x }` is already autocorrected to `Array.new(n) { x }` by rubocop

lib/rubocop/cop/rspec/factory_girl/create_list.rb
```diff
@@ -0,0 +1,74 @@
...
+            #n_times
+            (args)
+            ...
+          )
+          PATTERN
+
+          def_node_matcher :factory_call?, <<-PATTERN
```
bquorning: Since this matcher returns three captures (receiver, factory, and options) I would prefer we didn’t suffix it with `?` – that should be reserved for methods returning true/false.

lib/rubocop/cop/rspec/factory_girl/create_list.rb
```diff
@@ -0,0 +1,74 @@
...
+
+          def generate_replacement(block)
+            receiver, _args, body = *block
+            count = n_times(receiver)
+            factory_call_replacement(body, count)
+          end
+
+          def factory_call_replacement(body, count)
+            receiver, factory, options = *factory_call?(body)
+            replacement = receiver ? "#{receiver.source}." : ''
```
bquorning: I would prefer to split this in two lines:
```rb
replacement = ''
replacement += "#{receiver.source}." if receiver
```

lib/rubocop/cop/rspec/factory_girl/create_list.rb
```diff
@@ -0,0 +1,74 @@
...
+            replacement += "create_list :#{factory}, #{count}"
+            if options.count > 0
+              additional_options = options.map(&:source).join(', ')
+              replacement += ", #{additional_options}"
+            end
```
bquorning: It seems this block could be simplified a bit:
```rb
options.each do |option|
  replacement += ", #{options.source}"
end
```

spec/rubocop/cop/rspec/factory_girl/create_list_spec.rb
```diff
@@ -0,0 +1,64 @@
...
+      3.times { do_something }
+    RUBY
+  end
+
+  it 'ignores n.times when there is other calls but create' do
+    expect_no_offenses(<<-RUBY)
+      used_passwords = []
+      3.times do
+        u = create :user
+        expect(userd_passwords).not_to include(u.password)
```
bquorning: Typo – should be `used_passwords`.

Darhazer:
PR updated:
* Rebased on master
* Addessed bquorning comments, except for using just one matcher
* Added tests for possible false positive, if FactoryGirl.create is used with a block
* Preserved parens if the original code uses ones

lib/rubocop/cop/rspec/factory_girl/create_list.rb
```diff
@@ -0,0 +1,92 @@
...
+              ...
+            )
+          PATTERN
+
+          def_node_matcher :factory_call, <<-PATTERN
+            (send ${(const nil :FactoryGirl) nil} :create (sym $_) $...)
+          PATTERN
+
+          def on_block(node)
+            return unless times_block_without_args?(node)
```
Drenmi: Technically with arguments would also be an offense? Since there is a sequence helper in FactoryGirl. It just can't be easily auto-corrected. 🙂

lib/rubocop/cop/rspec/factory_girl/create_list.rb
```diff
@@ -0,0 +1,92 @@
...
+
+          private
+
+          def generate_replacement(block)
+            receiver, _args, body = *block
+            count = n_times(receiver)
+            factory_call_replacement(body, count)
+          end
+
+          def method_uses_parens?(node)
```
Drenmi: The node extension for parameterised nodes has a `#parenthesized?` method. 🙂

walf443:
FactoryGirl renamed to FactoryBot https://github.com/thoughtbot/factory_bot

walf443:
I created  a PR that rename to FactoryBot https://github.com/backus/rubocop-rspec/pull/490

spec/rubocop/cop/rspec/factory_girl/create_list_spec.rb
```diff
@@ -0,0 +1,76 @@
...
+  it 'flags usage of n.times with no arguments' do
+    expect_offense(<<-RUBY)
+      3.times { create :user }
+      ^^^^^^^ Prefer create_list.
+    RUBY
+  end
+
+  it 'flags usage of n.times when FactgoryGirl.create is used' do
+    expect_offense(<<-RUBY)
+      3.times { FactoryGirl.create :user }
```
pirj: FactoryBot as well.

spec/rubocop/cop/rspec/factory_girl/create_list_spec.rb
```diff
@@ -0,0 +1,76 @@
...
+        expect(used_passwords).not_to include(u.password)
+        used_passwords << u.password
+      end
+    RUBY
+  end
+
+  it 'ignores FactoryGirl.create calls with a block' do
+    expect_no_offenses(<<-RUBY)
+      3.times do
+        create(:user) { |user| create :account, user: user }
```
pirj: But why? It could be `create_list(:user, 3) { |user| create :account, user: user }`

Darhazer: The documentation doesn't mention this

pirj: It's [in the spec](https://github.com/thoughtbot/factory_bot/blob/c716ce01b448ce4e0bf855c5a2c63ecb9206322e/spec/acceptance/create_list_spec.rb#L45). Tried in the console, works as well.

CHANGELOG.md
```diff
@@ -16,6 +16,7 @@
 ## 1.19.0 (2017-10-18)
 
 Compatibility release so users can upgrade RuboCop to 0.51.0. No new features.
+* Add `FactoryGirl/CreateList` cop. ([@Darhazer][])
```
bquorning: Please move this line further up in the Changelog :-)

lib/rubocop/cop/rspec/factory_bot/create_list.rb
```diff
@@ -0,0 +1,92 @@
...
+          def_node_matcher :times_block_without_args?, <<-PATTERN
+            (block
+              #n_times
+              (args)
+              ...
+            )
+          PATTERN
+
+          def_node_matcher :factory_call, <<-PATTERN
+            (send ${(const nil? {:FactoryGirl :FactoryBot}) nil?} :create (sym $_) $...)
```
bquorning: Couldn’t `{:FactoryGirl :FactoryBot}` be extracted to somewhere in lib/rubocop/rspec somehow?

spec/rubocop/cop/rspec/factory_bot/create_list_spec.rb
```diff
@@ -0,0 +1,83 @@
...
+  subject(:cop) { described_class.new }
+
+  it 'flags usage of n.times with no arguments' do
+    expect_offense(<<-RUBY)
+      3.times { create :user }
+      ^^^^^^^ Prefer create_list.
+    RUBY
+  end
+
+  it 'flags usage of n.times when FactgoryGirl.create is used' do
```
bquorning: `s/Factgory/Factory/g` – 2 places needs fixing.

lib/rubocop/cop/rspec/factory_bot/create_list.rb
```diff
@@ -0,0 +1,92 @@
...
+
+          private
+
+          def generate_replacement(block)
+            receiver, _args, body = *block
+            count = n_times(receiver)
+            factory_call_replacement(body, count)
+          end
+
+          def method_uses_parens?(node)
```
bquorning: Since it’s only called from `#format_method_call`, I’d prefer to have this method definition at the bottom of the class.

lib/rubocop/cop/rspec/factory_bot/create_list.rb
```diff
@@ -0,0 +1,92 @@
...
+        #   3.times { |n| create :user, created_at: n.months.ago }
+        class CreateList < Cop
+          MSG = 'Prefer create_list.'.freeze
+
+          def_node_matcher :n_times, '(send (int $_) :times)'
+
+          def_node_matcher :times_block_without_args?, <<-PATTERN
+            (block
+              #n_times
+              (args)
```
bquorning: Help me: does `args` have a special meaning here?

Darhazer: means no arguments passed to the blog (otherwise (args will be followed by the args yielded)

lib/rubocop/cop/rspec/factory_bot/create_list.rb
Darhazer: No longer used

Darhazer:
PR updated:
* Rebased on master
* Made the cop configurable, since create_list is slower than n.times and some users may prefer to enforce this style

Darhazer:
TODO: handle create_list with block

spec/rubocop/cop/rspec/factory_girl/create_list_spec.rb
Darhazer: 10x. Will handle it

lib/rubocop/cop/rspec/factory_bot/create_list.rb
bquorning: I’m asking because I don’t see it documented in https://github.com/bbatsov/rubocop/blob/master/lib/rubocop/node_pattern.rb#L5-L91

Darhazer: I'm just using `ruby-parse` to generate the ASTs. `(args) is used in some other cops too

Darhazer: No longer needed, as now n.times with arguments is also flagged

Darhazer:
Rebased and added handling for create with block

lib/rubocop/cop/rspec/factory_bot/create_list.rb
```diff
@@ -0,0 +1,154 @@
...
+          def_node_matcher :factory_list_call, <<-PATTERN
+            (send ${(const nil? {:FactoryGirl :FactoryBot}) nil?} :create_list (sym $_) $...)
+          PATTERN
+
+          def on_block(node)
+            return unless style == :create_list
+
+            return unless n_times_block?(node)
+
+            return unless contains_only_factory?(node.body)
```
bquorning: Nitpick: I prefer no empty lines between these three guard clauses.

lib/rubocop/cop/rspec/factory_bot/create_list.rb
```diff
@@ -0,0 +1,154 @@
...
+              arguments = ":#{factory}"
+              arguments += build_options_string(options.drop(1))
+
+              replacement = format_receiver(receiver)
+              replacement += format_method_call(node, 'create', arguments)
+              "#{count}.times { #{replacement} }"
+            end
+          end
+
+          def generate_replacement(block)
```
bquorning: I would be fine with inlining this method into `#autocorrect_n_times_to_create_list`.

lib/rubocop/cop/rspec/factory_bot/create_list.rb
```diff
@@ -0,0 +1,154 @@
...
+
+            arguments = ":#{factory}, #{count}"
+            arguments += build_options_string(options)
+
+            replacement += format_method_call(body, 'create_list', arguments)
+            replacement
+          end
+
+          def build_options_string(options)
+            options.map { |option| ", #{option.source}" }.join
```
bquorning: That leading comma feels a bit wrong here. How about e.g. `options.map(&:source).join(', ')`, and then something like `arguments += ', ' if options.any?` in the callers?

lib/rubocop/cop/rspec/factory_bot/create_list.rb
```diff
@@ -0,0 +1,154 @@
...
+              ...
+            )
+          PATTERN
+
+          def_node_matcher :factory_call, <<-PATTERN
+            (send ${(const nil? {:FactoryGirl :FactoryBot}) nil?} :create (sym $_) $...)
+          PATTERN
+
+          def_node_matcher :factory_list_call, <<-PATTERN
+            (send ${(const nil? {:FactoryGirl :FactoryBot}) nil?} :create_list (sym $_) $...)
```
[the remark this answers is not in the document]
bquorning: …or even `(sym $_) (int $_) $...`? Then `count.source` would just become `count`.

Drenmi:
Thank you for working on this @Darhazer. 🙇

PR #455, opened 2017-09-06
Bump version to 1.16.0

Darhazer:
We didn't have a release for a while and there's a lot of cops added.

@backus @dgollahon as I don't have access to RubyGems, can you please push the new version there?


bquorning:
I’ve pushed v1.16.0 to RubyGems. Thank you for the release @Darhazer.

backus:
Thanks guys

PR #458, opened 2017-09-09
Fix false positives in ReturnFromStub. Fixes #457

### LINKED ISSUE #457: ReturnFromStub on strings with interpolation
Running rubocop on this file
```rb
RSpec.describe 'bug' do
  let(:name) { 'Maxim' }

  before do
    allow(Foo).to receive(:bar) { "Hello #{name}" }
  end
end
```
tells me to use `and_return`:
```
spec/test.rb:5:19: C: RSpec/ReturnFromStub: Use and_return for static values.
    allow(Foo).to receive(:bar) { "Hello #{name}" }
```

But, is a string with interpolation a static value?

Darhazer:


Darhazer:
Updated. Not as pretty as it was before, but works better. I can't think of a situation when checking for the parent block will yield false positive

PR #459, opened 2017-09-09
Fix for JRuby

zverok:
JRuby support was broken at 1.16.0, and nobody noticed because `.travis.yml` doesn't include JRuby.
If it is not a principal position of the team, please merge this to fix both problems :)

### DIFF
--- a/lib/rubocop/cop/rspec/cop.rb
+++ b/lib/rubocop/cop/rspec/cop.rb
@@ -5,10 +5,11 @@ module RuboCop
     # Clone of the the normal RuboCop::Cop::Cop class so we can rewrite
     # the inherited method without breaking functionality
     class WorkaroundCop
-      # Remove the cop inherited method to be a noop. Our RSpec::Cop
+      # Remove the Cop.inherited method to be a noop. Our RSpec::Cop
       # class will invoke the inherited hook instead
       class << self
-        remove_method :inherited
+        undef inherited
+        def inherited(*) end
       end
 
       # Special case `Module#<` so that the rspec support rubocop exports
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,6 +7,7 @@ rvm:
   - 2.2.5
   - 2.3.1
   - ruby-head
+  - jruby-9.1.13.0
 
 matrix:
   allow_failures:


lib/rubocop/cop/rspec/cop.rb
```diff
@@ -5,11 +5,9 @@ module Cop # rubocop:disable Style/Documentation
     # Clone of the the normal RuboCop::Cop::Cop class so we can rewrite
     # the inherited method without breaking functionality
     class WorkaroundCop
-      # Remove the cop inherited method to be a noop. Our RSpec::Cop
+      # Remove the Cop.inherited method to be a noop. Our RSpec::Cop
       # class will invoke the inherited hook instead
-      class << self
-        remove_method :inherited
-      end
+      def self.inherited(*) end
```
pocke: `WorkaroundCop` uses `remove_method` to avoid MRI warning.
See https://github.com/backus/rubocop-rspec/pull/436

Could you use `undef`? Like:

```ruby
class WorkaroundCop
  # Remove the Cop.inherited method to be a noop. Our RSpec::Cop
  # class will invoke the inherited hook instead
  class << self
    undef inherited
    def inherited(*) end
  end
```

It works well on JRuby, and MRI does not warn about it.

zverok:
Sure. Updated the PR.

zverok:
image 
:cry:

Can I haz the merge?.. (Rebased from latest master, it is green)

bquorning:
Since @backus removed JRuby support in #131 and #133 I’d like to hear his opinion on this PR.

I think JRuby support would be a good thing, but if it slows down all builds at a very small risk of failure… I’m not sure how to solve this dilemma.

backus:
Yeah its just about maintenance overhead and I personally don't have any investment in jruby. Fine with adding it back to the build for now but we also need to get it into buildkite if we want to be serious about not breaking the integration. This PR is a good step forward though so 👍

bquorning:
@backus I’ll cut a patch release including this change.

bquorning:
@zverok Thank you for your patience. v1.17.1 is now released to rubygems.org.

zverok:
@bquorning Thanks!

@backus Well, I am not into JRuby myself, just maintain several gems, and sad when they are suddenly failing on CI. In fact, it is not a big deal, as it is reasonable to run `rubocop` on only one config of the "build matrix" (offenses are same for all Ruby versions, including JRuby), but ...

PR #245, opened 2016-11-26
Implement ContainExactly and MatchArray cops with autocorrect

faucct:
Here are some examples:

```ruby
# bad
it { is_expected.to contain_exactly(*array1, *array2) }

# good
it { is_expected.to match_array(array1 + array2) }

# bad
it { is_expected.to match_array([content1, content2]) }

# good
it { is_expected.to contain_exactly(content1, content2) }

# good
it { is_expected.to contain_exactly(content, *array) }

# good
it { is_expected.to match_array([content] + array) }
```

PR #464, opened 2017-09-19
Implement ContainExactly and MatchArray cops with autocorrect

Darhazer:
I have rebased #245 on master, added changelog and cleaned up a bit the code (e.g. applied my own feedback from the original pull request). I'm happy with this version of the cops.

So if anyone else approves, we can release this.


lib/rubocop/cop/rspec/match_array.rb
```diff
@@ -0,0 +1,39 @@
...
+      #   it { is_expected.to contain_exactly(content1, content2) }
+      #
+      #   # good
+      #   it { is_expected.to match_array([content] + array) }
+      class MatchArray < Cop
+        MSG = 'Prefer `contain_exactly` when matching an array literal.'.freeze
+
+        def on_send(node)
+          return unless node.method_name == :match_array
+          return unless node.each_child_node.all?(&:array_type?)
```
bquorning: I’m a bit confused by the `all?` here. This should only find nodes with *one* child where that child is an `array_type?`, right?

pirj:
My only concern is that there should be a cop to detect offense in the following code:
```
eq([single])
```
and propose to replace with `contains_exactly(single)`.
I'm not sure what the name for it should be so that it does not clash with the one in this pull request.

What is stopping you from updating and merging this one?

Darhazer:
Rebased and updated

spec/rubocop/cop/rspec/contain_exactly_spec.rb
```diff
@@ -0,0 +1,26 @@
...
+  end
+
+  it 'does not flag `contain_exactly` with mixed arguments' do
+    expect_no_offenses(<<-RUBY)
+      it { is_expected.to contain_exactly(content, *array) }
+    RUBY
+  end
+
+  include_examples 'autocorrect',
+                   'it { is_expected.to contain_exactly(*array1, *array2) }',
```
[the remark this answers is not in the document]
Darhazer: @faucct what do you think?

mvz:
First, a question. Given the following code:

```ruby
is_expected_to match_array %w(tremble in fear foolish mortals)
```

Will the MatchArray cop suggest the following change? It seems worse to me :slightly_smiling_face:.
```ruby
is_expected_to contain_exactly('tremble', 'in', 'fear', 'foolish', 'mortals')
```

I'm also not sure about the split between when to use `match_array` and when to use `contain_exactly`. It seems a bit arbitrary and I would prefer to be able to enforce using only one or the other.

Finally, since the function of these two cops is so similar, why not merge them into one? This would allow easy configuration to enforce either one or the other matcher, or the currently proposed mixed style.

spec/rubocop/cop/rspec/contain_exactly_spec.rb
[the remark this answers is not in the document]
Darhazer: I think we can add this as an unsafe cop (can yield false positive by design) and mark it as unsafe as soon as Rubocop itself support this meta data (see rubocop-hq/rubocop#5980)

Darhazer:
Thank you for your feedback @mvz 
I agree that %w() should not be flagged and will update the test cases accordingly.

Regarding merging the cops into one, it makes sense to me. I don't have strong opinion on this though.

mvz:
@Darhazer any news on this?

pirj:
@Darhazer the cops seem to be one hair from the finish line. Are you up to finish it?

spec/rubocop/cop/rspec/contain_exactly_spec.rb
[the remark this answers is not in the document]
pirj: Is skipping autocorrection when any of the arguments is a splat an option?

ydah:
@Darhazer I'm eagerly awaiting the addition of this cop. Please let me know if there is anything I can do to help.

Darhazer:
@ydah do you want to update the cops and handle the %w() case?
I don't have much time lately, but I promise to review with high priority ;)

spec/rubocop/cop/rspec/contain_exactly_spec.rb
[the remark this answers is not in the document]
pirj: Answer to self: it's not because the whole focus of the cop is when all arguments are splats.

But realistically, why would anyone use `contain_exactly(*hash1, *hash2)`? 🤔

pirj:
@Darhazer Done

lib/rubocop/cop/rspec/match_array.rb
```diff
@@ -0,0 +1,41 @@
...
+      #
+      #   # good
+      #   it { is_expected.to match_array(%w(tremble in fear foolish mortals)) }
+      class MatchArray < Base
+        extend AutoCorrector
+
+        MSG = 'Prefer `contain_exactly` when matching an array literal.'
+
+        def on_send(node)
+          return unless node.method?(:match_array)
```
pirj, line 26: I'm uncertain if we can rely on `RESTRIcT_ON_SEND` to replace this check. But I'll happily add it for performance.

pirj: Simplecov complains about this line. But I disagree with the need to test it.

image

Adjusted the constraints.

pirj: I don't get why the Test coverage job fails, as for me locally it's exactly within limits:
```
All Files ( 99.89% covered at 219.07 hits/line )
144 files in total.
3548 relevant lines, 3544 lines covered and 4 lines missed. ( 99.89% )
747 total branches, 712 branches covered and 35 branches missed. ( 95.31% ) 
```
@ydah WDYT?

.simplecov
```diff
@@ -2,7 +2,7 @@
 
 SimpleCov.start do
   enable_coverage :branch
-  minimum_coverage line: 99.86, branch: 95.33
+  minimum_coverage line: 99.89, branch: 95.31
```
bquorning, line 5:
```suggestion
  minimum_coverage line: 99.88, branch: 95.31
```

lib/rubocop/cop/rspec/match_array.rb
ydah: Sorry for the late reply. Looks like it's resolved.

lib/rubocop/cop/rspec/contain_exactly.rb
```diff
@@ -0,0 +1,46 @@
...
+      #   # good
+      #   it { is_expected.to contain_exactly(content, *array) }
+      class ContainExactly < Base
+        extend AutoCorrector
+
+        MSG = 'Prefer `match_array` when matching array values.'
+        RESTRICT_ON_SEND = %i[contain_exactly].freeze
+
+        def on_send(node)
+          return unless node.method?(:contain_exactly)
```
Darhazer, line 24: I believe this is no needed when we have RESTRICT_ON_SEND

lib/rubocop/cop/rspec/match_array.rb
Darhazer: Yes, I think we do not need that line

spec/rubocop/cop/rspec/contain_exactly_spec.rb
```diff
@@ -0,0 +1,26 @@
+# frozen_string_literal: true
+
+RSpec.describe RuboCop::Cop::RSpec::ContainExactly do
+  it 'flags `contain_exactly` with only splat arguments' do
+    expect_offense(<<-RUBY)
+      it { is_expected.to contain_exactly(*array1, *array2) }
```
Darhazer, line 6: Can we add more tests please?
contain_exactly(*[1,2,3])
contain_exactly(*a.merge(b))
contain_exactly(*(a + b))
Guess the later two are not handled, and that's OK for the initial version, but lets have a test for the basic case of a literal array
how about also contain_exactly(*%w(a b))

spec/rubocop/cop/rspec/match_array_spec.rb
```diff
@@ -0,0 +1,33 @@
+# frozen_string_literal: true
+
+RSpec.describe RuboCop::Cop::RSpec::MatchArray do
+  it 'flags `match_array` with array literal arguments' do
+    expect_offense(<<-RUBY)
+      it { is_expected.to match_array([content1, content2]) }
```
Darhazer, line 6: again, how about match_array([*content, content2])
Whilw I do not insist of support, I just want to make sure it doesn't break

Darhazer (commented):
Thank you.
Lets look into more cases (as listed in the specs) and also I believe we can fix the branch coverage by relying purely on RESTRICT_ON_SEND

spec/rubocop/cop/rspec/contain_exactly_spec.rb
pirj: Do you mean `contain_exactly(*%w(a b))`? Added

pirj:
@Darhazer Thanks for the review. Please have another look.

spec/rubocop/cop/rspec/contain_exactly_spec.rb
```diff
@@ -0,0 +1,45 @@
...
+    expect_correction(<<-RUBY)
+      it { is_expected.to match_array(%w(a b)) }
+    RUBY
+  end
+
+  it 'ignores non-splat usages' do
+    expect_no_offenses(<<-RUBY)
+      it { is_expected.to contain_exactly([1,2,3]) }
+      it { is_expected.to contain_exactly(a.merge(b)) }
+      it { is_expected.to contain_exactly((a + b)) }
```
Darhazer, line 30: Does were meant to be splats

Darhazer, line 30: All those examples were meant to be splats

spec/rubocop/cop/rspec/contain_exactly_spec.rb
```diff
@@ -0,0 +1,45 @@
...
+
+  it 'does not flag `match_array`' do
+    expect_no_offenses(<<-RUBY)
+      it { is_expected.to match_array(array1 + array2) }
+    RUBY
+  end
+
+  it 'does not flag `contain_exactly` with mixed arguments' do
+    expect_no_offenses(<<-RUBY)
+      it { is_expected.to contain_exactly(content, *array) }
```
Darhazer, line 42: what is the splat is the first argument. I know the answer from the source (check that all are splats, but I guess for spec completeness we need both)

spec/rubocop/cop/rspec/contain_exactly_spec.rb
pirj:
```diff
-       it { is_expected.to contain_exactly(*[1,2,3]) }
-       it { is_expected.to contain_exactly(*a.merge(b)) }
-       it { is_expected.to contain_exactly(*(a + b)) }
+       it { is_expected.to match_array([1,2,3]) }
+       it { is_expected.to match_array(a.merge(b)) }
+       it { is_expected.to match_array((a + b)) }
```
Does this look fine?

Darhazer (commented):
Thank you @pirj 🙇 
I can't formally approve, as I'm the author of the PR 😅 but feel free to merge

Darhazer:
Oh I can't approve, but thanks to previous approvals I can merge 😄

pirj:
Thanks everyone involved 🎉

Darhazer:
Took more than 6 years since the original PR, but finally ... @faucct you can consider yourself contributor to rubocop-rspec 😅

ydah:
@pirj @Darhazer 
Thanks so much for bringing this forward!
I'm sorry I said I would help but I haven't done anything 😅.

stoivo:
Why is contain_exactly preferred over match_array?

PR #469, opened 2017-09-26
Fix false positives in Capybara/Feature cop. Fixes #468

### LINKED ISSUE #468: Capybara/FeatureMethods false positive - wants to rename all variables
We've got a rails app where customer's services can have additional features attached to them, so in our specs we have variables named `feature`. The Capybara/FeatureMethods cop is, I think incorrectly, suggesting they all be renamed to `describe`.

I'm using rubocop 0.50.0 with rubocop-rspec 1.17.1

Given this example file I've just built that minimally reproduces what I'm seeing:
```
# ./spec/models/rubocop_repro_spec.rb
require 'rails_helper'

RSpec.describe SomeServiceClass, type: :model do
  describe '#add_feature' do
    subject { service.add_feature(feature) }
    
    let(:service) { FactoryGirl.create :service }
    let(:feature) { FactoryGirl.create :feature_element }

    it { is_expected.to eq true }
  end
end
```

Running Rubocop gives me:
```
$ rubocop ./spec/models/rubocop_repro_spec.rb --only Capybara/FeatureMethods
Inspecting 1 file
C

Offenses:

spec/models/rubocop_repro_spec.rb:5:35: C: Capybara/FeatureMethods: Use describe instead of feature.
    subject { service.add_feature(feature) }
                                  ^^^^^^^

```

Darhazer:
@rspeicher @bquorning 

Generally we should add some common tests, as mentioned earlier by @backus, or at least some guidelines, in order to check for such errors (false positive when the searched token is mentioned inside the example, empty examples group or so on)

### DIFF
--- a/lib/rubocop/cop/rspec/capybara/feature_methods.rb
+++ b/lib/rubocop/cop/rspec/capybara/feature_methods.rb
@@ -45,13 +45,15 @@ module RuboCop
             feature:    :describe
           }.freeze
 
-          def_node_matcher :feature_method?, <<-PATTERN
-            (send {(const nil :RSpec) nil} ${:#{MAP.keys.join(' :')}} ...)
+          def_node_matcher :feature_method, <<-PATTERN
+            (block
+              $(send {(const nil :RSpec) nil} ${#{MAP.keys.map(&:inspect).join(' ')}} ...)
+            ...)
           PATTERN
 
-          def on_send(node)
-            feature_method?(node) do |match|
-              add_offense(node, :selector, format(MSG, MAP[match], match))
+          def on_block(node)
+            feature_method(node) do |send_node, match|
+              add_offense(send_node, :selector, format(MSG, MAP[match], match))
             end
           end
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Master (Unreleased)
 
+* Fix false positive in `Capybara/FeatureMethods`. ([@Darhazer][])
+
 ## 1.17.1 (2017-09-20)
 
 * Improved `RSpec/ReturnFromStub` to handle string interpolation, hashes and do..end blocks. ([@Darhazer][])
--- a/spec/rubocop/cop/rspec/capybara/feature_methods_spec.rb
+++ b/spec/rubocop/cop/rspec/capybara/feature_methods_spec.rb
@@ -43,6 +43,16 @@ RSpec.describe RuboCop::Cop::RSpec::Capybara::FeatureMethods do
     RUBY
   end
 
+  it 'ignores variables inside examples' do
+    expect_no_offenses(<<-RUBY)
+      it 'is valid code' do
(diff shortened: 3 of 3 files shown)


rspeicher:
@Darhazer Good call, sorry for the oversight. Thanks for fixing it!

lib/rubocop/cop/rspec/capybara/feature_methods.rb
```diff
@@ -45,13 +45,15 @@ class FeatureMethods < Cop
             feature:    :describe
           }.freeze
 
-          def_node_matcher :feature_method?, <<-PATTERN
-            (send {(const nil :RSpec) nil} ${:#{MAP.keys.join(' :')}} ...)
+          def_node_matcher :feature_method, <<-PATTERN
+            (block
+              $(send {(const nil :RSpec) nil} ${:#{MAP.keys.join(' :')}} ...)
```
bquorning: I’m tempted to replace `:#{MAP.keys.join(' :')}` with `MAP.keys.map(&:inspect).join(' ')`, just out of personal taste. Having the (symbol prefix) colon as part of the `join` argument just rubs me the wrong way.

lib/rubocop/cop/rspec/capybara/feature_methods.rb
```diff
@@ -45,13 +45,15 @@ class FeatureMethods < Cop
...
+            ...)
           PATTERN
 
-          def on_send(node)
-            feature_method?(node) do |match|
-              add_offense(node, :selector, format(MSG, MAP[match], match))
+          def on_block(node)
+            feature_method(node) do |send, match|
```
bquorning: Could we use `send_node` instead of `send` here?

lib/rubocop/cop/rspec/capybara/feature_methods.rb
rspeicher: @bquorning I think that's more elegant than my original implementation. 👍

Darhazer:
PR updated

lib/rubocop/cop/rspec/capybara/feature_methods.rb
```diff
@@ -45,13 +45,15 @@ class FeatureMethods < Cop
             feature:    :describe
           }.freeze
 
-          def_node_matcher :feature_method?, <<-PATTERN
-            (send {(const nil :RSpec) nil} ${:#{MAP.keys.join(' :')}} ...)
+          def_node_matcher :feature_method, <<-PATTERN
+            (block
+              $(send {(const nil :RSpec) nil} ${#{MAP.keys.map(&:inspect).join}} ...)
```
rspeicher: I'm surprised `.join`, rather than `.join(' ')`, is still passing.

bquorning: Probably the pattern is split on every `:` in a string. I would prefer if you keep the `(' ')` though :-)

PR #470, opened 2017-09-26
Add cop for Capybara expectations set on `current_path`

timrogers:
Setting expectations on `current_path` in our Capybara feature specs leads to flaky tests.  Switching to using [Capybara's `have_current_path` matcher](http://www.rubydoc.info/github/teamcapybara/capybara/master/Capybara/RSpecMatchers#have_current_path-instance_method) helps since it uses Capybara's [waiting functionality](https://github.com/teamcapybara/capybara/blob/master/README.md#asynchronous-javascript-ajax-and-friends) that ensures that preceding actions (like `click_link`) have completed. 

This cop tells you not to use `expect(current_path).to match(/lol/)`, encouraging you to use the `have_current_path` matcher instead.


Darhazer:
You have to:
* add entry to the `config/default.yml` (run bin/build_config)
* add entry to the changelog
* add bad and good examples as comments (see the other cops)
* add test that shows you can have `current_path` outside of the expectation block (like just assigning to a variable)
* I see often `expect(page.current_path).to`. Please cover this scenario

timrogers:
Will do! Thanks @Darhazer - really helpful list.

timrogers:
@Darhazer I've managed to do everything except no. 1 - I get the following:

```bash
$ bin/build_config
/Users/timrogers/Projects/rubocop-rspec/lib/rubocop/rspec/config_formatter.rb:23:in `fetch': key not found: "Capybara/CurrentPathExpectation" (KeyError)
	from /Users/timrogers/Projects/rubocop-rspec/lib/rubocop/rspec/config_formatter.rb:23:in `block in unified_config'
	from /Users/timrogers/Projects/rubocop-rspec/lib/rubocop/rspec/config_formatter.rb:22:in `each'
	from /Users/timrogers/Projects/rubocop-rspec/lib/rubocop/rspec/config_formatter.rb:22:in `each_with_object'
	from /Users/timrogers/Projects/rubocop-rspec/lib/rubocop/rspec/config_formatter.rb:22:in `unified_config'
	from /Users/timrogers/Projects/rubocop-rspec/lib/rubocop/rspec/config_formatter.rb:16:in `dump'
	from bin/build_config:20:in `<main>'
```

Any ideas? I'm sure I'm missing something obvious!

lib/rubocop/cop/rspec/capybara/current_path_expectation.rb
```diff
@@ -0,0 +1,40 @@
...
+        #   expect(page).to have_current_path("/callback")
+        #   expect(page).to have_current_path(/widgets/)
+        #
+        class CurrentPathExpectation < Cop
+          MSG = 'Do not set an RSpec expectation on `current_path` in ' \
+                'Capybara feature specs - instead, use the ' \
+                '`have_current_path` matcher on `page`'.freeze
+
+          def_node_matcher :expectation_set_on_current_path, <<-PATTERN
+            (send nil :expect {(send nil :current_path) (send (send nil :page) :current_path)})
```
Darhazer: I believe this could be written as `(send nil :expect (send {(send nil :page) nil} :current_path)})`

Darhazer:
@timrogers you need to add the key in `config/default.yml`. `build_config` will take care to sync description between the cop and the config

timrogers:
@Darhazer Aha - thanks! All done and I've pushed a finished, squashed version.

Darhazer:
Great job, @timrogers 

Would you be willing to add another one, to check for `not_to have_` calls, that should be avoided for the very same reason (see #378)?

timrogers:
Yes, sure!

Darhazer:
Note: generally only `eq` / `match` kind of matchers can be replaced with `have_current_path`. I didn't see other usages though; and even if someone uses current_path with some fancy matcher, refactoring to `have_current_path` is the right thing to do. So I'm going to merge this as is, unless there are any objections; the note should be respected however if someone adds an autocorrect to this.

@bquorning any thoughts?

config/default.yml
```diff
@@ -334,3 +334,8 @@ FactoryGirl/DynamicAttributeDefinedStatically:
   Description: Prefer declaring dynamic attribute values in a block.
   Enabled: true
   StyleGuide: http://www.rubydoc.info/gems/rubocop-rspec/RuboCop/Cop/RSpec/FactoryGirl/DynamicAttributeDefinedStatically
+
+Capybara/CurrentPathExpectation:
```
bquorning: This is a semi-alphabetical list (`RSpec/` is before `Capybara/`) – could we move this entry to line 328?

timrogers:
That makes sense to me. I was aware that people might be using other
matchers, but where they are they’re likely to see non-deterministic test
results such that they’d be well-advised to change their implementation.

I’ll look at the other cop you suggested tomorrow, and may look at
implementing an autocorrect too - but let’s merge this as-is for now if
that’s okay with @bquorning as this feels like a positive addition.

lib/rubocop/cop/rspec/capybara/current_path_expectation.rb
```diff
@@ -0,0 +1,42 @@
...
+          MSG = 'Do not set an RSpec expectation on `current_path` in ' \
+                'Capybara feature specs - instead, use the ' \
+                '`have_current_path` matcher on `page`'.freeze
+
+          def_node_matcher :expectation_set_on_current_path, <<-PATTERN
+            (send nil :expect (send {(send nil :page) nil} :current_path))
+          PATTERN
+
+          def on_send(node)
+            expectation_set_on_current_path(node) do |_match|
```
bquorning: There are no captures in the node matcher, so no block arguments are passed: You can remove ` |_match|`.

spec/rubocop/cop/rspec/capybara/current_path_expectation_spec.rb
```diff
@@ -0,0 +1,29 @@
...
+    RUBY
+  end
+
+  it 'doesn\'t flag a violation for other expectations' do
+    expect_no_offenses(<<-RUBY)
+      expect(current_user).to eq(user)
+    RUBY
+  end
+
+  it 'doesn\'t flag a violation for other references to `current_path`' do
```
bquorning: Nitpick: If you use double quotes around the string, you don’t need to escape the `'` in the middle. (also in line 18).

timrogers:
@bquorning Done!

PR #473, opened 2017-09-27
Don't check single-line example blocks. Fixes #456

### LINKED ISSUE #456: LetBeforeExamples when using a let block within it_behaves_like block
I'm seeing LetBeforeExamples offenses after upgrading to rubocop-rspec 1.16.0 (from 1.15.1)

```
RSpec.describe Model, type: :model do
  it_behaves_like 'shared_example' do
    let(:model_variable) { 'some variable specific to this model' }
  end
end
```
results in:
`RSpec/LetBeforeExamples: Move let before the examples in the group.`

Similar to https://github.com/backus/rubocop-rspec/issues/454

Darhazer:
When the example has single line, the `let` becomes sibling on the `include_examples`, hence the false positive while checking the `describe` block.
When the `let` is the only node in the block however, there is no way it can be before another example

### DIFF
--- a/lib/rubocop/cop/rspec/let_before_examples.rb
+++ b/lib/rubocop/cop/rspec/let_before_examples.rb
@@ -44,9 +44,13 @@ module RuboCop
         def on_block(node)
           return unless example_group_with_body?(node)
 
-          _describe, _args, body = *node
+          check_let_declarations(node.body) if multiline_block?(node.body)
+        end
 
-          check_let_declarations(body)
+        private
+
+        def multiline_block?(block)
+          block.begin_type?
         end
 
         def check_let_declarations(node)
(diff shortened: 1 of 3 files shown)


lib/rubocop/cop/rspec/let_before_examples.rb
```diff
@@ -44,9 +44,7 @@ class LetBeforeExamples < Cop
         def on_block(node)
           return unless example_group_with_body?(node)
 
-          _describe, _args, body = *node
-
-          check_let_declarations(body)
+          check_let_declarations(node.body) if node.body.begin_type?
```
bquorning: I’m not sure what this (`node.body.begin_type?`) means.

Darhazer: Single-line block looks like this:
```
(block
  (send nil :include_examples
    (str "special user"))
  (args)
  (block
    (send nil :let
      (sym :foo))
    (args)
    (send nil :bar)))
```

while multi-line have a (begin)
```
(block
  (send nil :include_examples
    (str "special user"))
  (args)
  (begin
    (block
      (send nil :let
        (sym :foo))
      (args)
      (send nil :bar))
    (block
      (send nil :let
        (sym :bar))
      (args)
      (send nil :baz))))
```

So I use this fact to determine if the body of the block have more than one statement. I can extract it in appropriately named method for better readability

PR #476, opened 2017-09-29
Add autocorrect to RSpec/LetBeforeExamples

Darhazer:


lib/rubocop/cop/rspec/let_before_examples.rb
```diff
@@ -64,6 +74,39 @@ def check_let_declarations(node)
             end
           end
         end
+
+        def find_first_example(node)
+          node.parent.children.find { |sibling| example_or_group?(sibling) }
+        end
+
+        def node_range(node)
```
bquorning: I generally prefer if methods always call "down" in the source file, if possible. Since this method is called by `#node_range_with_surrounding_space` and `#source`, it should be moved below them.

lib/rubocop/cop/rspec/let_before_examples.rb
```diff
@@ -64,6 +74,39 @@ def check_let_declarations(node)
...
+          range
+        end
+
+        def source(node)
+          node_range(node).source
+        end
+
+        def last_node_loc(node)
+          last_line = node.loc.end.line
+          heredoc_line(node) do |loc|
```
bquorning: Why are heredocs so hard? Isn’t there some shared code to be extracted that can help us deal with heredocs?

Darhazer: Rubocop definitely needs some helpers to work with heredocs. I managed to simplify a bit

Darhazer:
PR updated. Unless there are any objections, merge?

PR #477, opened 2017-10-02
Use `nil?` instead of `nil` to match nil value in AST in node pattern

pocke:
In https://github.com/bbatsov/rubocop/pull/4815 , NodePattern is changed.
This pull-request follows the change.

Sorry for the braking change. Please merge this pull-request after RuboCop v0.50.1 (or something) is released.


Darhazer:
RuboCop 0.51 is released, so I guess we need to merge this.
Also they deprecated the current add_offense API in favor of keyword args

pocke:
> Also they deprecated the current add_offense API in favor of keyword args


Yes. but sorry I have no time to update the add_offense now.

PR #479, opened 2017-10-06
Fix spelling in return_from_stub cop

srushe:
Two small spelling fixes in the documentation for the `ReturnFromStub` cop that I noticed when reading the docs.


bquorning:
Thank you

PR #482, opened 2017-10-10
Fix false positive in RSpec/Focus cop. Fixes #398

### LINKED ISSUE #398: Focus cop raises incorrectly on variables named `fit`.
Example:
```Ruby
let(:fit) { Tax.federal_income_tax }
let(:fit_attributes) { tax_id: fit.id }
```
raises
```
C: RSpec/Focus: Focused spec found.
       { tax_id: fit.id }
                 ^^^
```

Darhazer:
Since the cop was checking `send` and not `block` pattern, it was giving offences when a message to a variable called fit, fdescribe or similar was sent

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Master (Unreleased)
 
+* Fix false positive in `RSpec/Focus` cop. ([@Darhazer][])
+
 ## 1.18.0 (2017-09-29)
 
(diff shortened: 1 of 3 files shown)


dgollahon:
@Darhazer: I think this is not always a false positive. With my rspec configuration if I do

```ruby
fit 'I haven't actually implemented this yet'
```

This will still focus the specs and run that on example and just tell me that it isn't implemented yet.

PR #486, opened 2017-10-26
Cop that checks for pending specs

patrickomatic:
For use on teams where pending or skipped specs should not make their way into master.


backus:
Neat. Ideally this doesn't overlap with the Focus cop but if this covers new stuff then I'm all for it

bquorning:
Since `skip` and `pending` are part of the RSpec api (and in my opinion perfectly fine to use), I think cop should be disabled by default.

patrickomatic:
Thanks for the feedback @backus - I don't think this functionality overlaps with the focus cop at all? I noticed that one of my examples mentioned focus but I've removed that.  But other than that the two are fairly distinct features.

I agree @bquorning - most teams probably don't have this rule.  I've added some config for it and set to not enabled by default

backus:
@patrickomatic kk yeah sorry i only took a 20 second glance at the code so just wanted to point out the similarity

patrickomatic:
Hi, just wanted to check in on this and see if there's anything I can do to get it approved for merge?  I've signed up for buildkite to try to see why the build is failing but unfortunately can't seem to get the appropriate access

bquorning:
It’s the rubocop check (`bundle exec rubocop --require rubocop-rspec` or `bundle exec rake internal_investigation`) that’s failing.

I’ll investigate how Buildkite access works.

bquorning:
It’s really problematic that we’re using a CI that our contributors doesn’t have access to. If I can’t find a way to give open access to Buildkite I’ll take a look at e.g. Circle CI instead…

bquorning:
`bundle exec rake confirm_config` seems to be failing.

patrickomatic:
Thanks for the quick feedback @bquorning - I have the feeling this next build will pass

bquorning:
There are no spec examples for e.g.

```rb
it "works" do
  skip
end

it "works", skip: "doesn't work yet" do
end
```
I assume these (perhaps less common) skip/pending styles are not covered by this PR?

patrickomatic:
Ah good call.  In theory both of those should work because I just check for `Examples::PENDING + Examples::SKIPPED + ExampleGroups::SKIPPED`.  

I added spec coverage for the former which works as expected.  However the latter doesn't work, but seems like a bug further up in https://github.com/backus/rubocop-rspec/blob/master/lib/rubocop/rspec/language.rb#L81 where it doesn't take into account `skip` in the metadata.  I can take a deeper look into fixing that tomorrow

spec/rubocop/cop/rspec/pending_spec.rb
```diff
@@ -0,0 +1,51 @@
+RSpec.describe RuboCop::Cop::RSpec::Pending do
+  subject(:cop) { described_class.new }
+
+  # rubocop:disable RSpec/ExampleLength
+  it 'flags all pending specs' do
+    expect_offense(<<-RUBY)
+      it 'test' do; skip; end
+                    ^^^^ Pending spec found.
```
bquorning: I think this example would actually work better as separate examples (just as RSpec/ExampleLength suggests). If the code didn’t work for e.g. `xfeature`, I imagine the RSpec output would be pretty hard to read. Much simpler if there were a spec example like
```rb
it 'flags xfeature' do
  expect_offense <<-RUBY
    xfeature  'test' do
    ^^^^^^^^^^^^^^^^ Pending spec found.
    end
  RUBY
end
```

backus:
> It’s really problematic that we’re using a CI that our contributors doesn’t have access to. If I can’t find a way to give open access to Buildkite I’ll take a look at e.g. Circle CI instead…


Travis is supposed to mirror what is on buildkite

bquorning:
> Travis is supposed to mirror what is on buildkite

Ah, I see. Anyway, I have branch that successfully tests this repo on circle-ci. Perhaps I’ll open a PR and discuss moving away from Travis instead :-)

lib/rubocop/cop/rspec/pending.rb
```diff
@@ -0,0 +1,66 @@
...
+
+        SKIPPABLE_SELECTORS = skippable.node_pattern_union
+
+        SKIP_SYMBOL = s(:sym, :skip)
+        SKIP_TRUE   = s(:pair, SKIP_SYMBOL, s(:true))
+
+        def_node_matcher :metadata, <<-PATTERN
+          {(send nil? #{SKIPPABLE_SELECTORS} ... (hash $...))
+           (send nil? #{SKIPPABLE_SELECTORS} $...)}
+        PATTERN
```
bquorning: I’m having a hard time reading this `:metadata` node matcher, together with the `grep` methods in the last method. Would it be possible to make it more explicit how things work?

I believe all of these metadata examples should be matched:

- `:skip`
- `skip: true`
- `skip: "does not work on CI"`
- `:pending`
- `pending: true`
- `pending: "fix it later"`

Even dynamic values should work, e.g. `skip: !ENV['CI']`, right?

patrickomatic:
Hey @bquorning sorry for the delay and thanks for the continued feedback. 

I've pushed a couple more commits that simplify some of the logic that previously passed around blocks to just return booleans (though I'd like to point out, the prior code was very similar to what's already going on in the `Focus` cop) and fixed a bug where `skip: false` would have been incorrectly identified.

lib/rubocop/cop/rspec/pending.rb
patrickomatic: I refactored the code a bit to help with the `grep`s, but in regard to that list of examples, I think only the second and third are supported.  I can't find any rspec docs that say you can use `pending` as metadata, just `skip`.  If you find something that states otherwise let me know and I'll update accordingly. 

As far as the last point, I'm not sure... I can't think of how a static code analyzer could ever deal with that in an expected manner

bquorning: > I can't find any rspec docs that say you can use `pending` as metadata, just `skip`. If you find something that states otherwise let me know and I'll update accordingly.

https://relishapp.com/rspec/rspec-core/v/3-7/docs/pending-and-skipped-examples/pending-examples#`pending`-for-an-example-that-is-currently-passing-with-a-reason shows the example `example("something else getting finished", :pending => 'unimplemented') do`…

I actually can’t find any documentation for the `:skip` and `:pending` (symbol) metadata, only the hash variants.

.rubocop.yml
```diff
@@ -25,6 +25,7 @@ Lint/BooleanSymbol:
   Exclude:
     - lib/rubocop/cop/rspec/expect_actual.rb
     - lib/rubocop/cop/rspec/focus.rb
+    - lib/rubocop/cop/rspec/pending.rb
```
bquorning: Is this line still needed?

lib/rubocop/cop/rspec/pending.rb
phallguy: According [to the docs](https://relishapp.com/rspec/rspec-core/v/3-7/docs/pending-and-skipped-examples), the `skip` behavior is to prevent a spec from running for whatever reason, whereas `pending` is for specs that are _known_ to fail and are still run which I think are really distinct use cases especially with regard to a cop. Could we consider introducing separate cops for each use case?

patrickomatic: That's a good point @phallguy - I could see the argument for splitting them up, though I'll admit I don't feel very strongly either way.

phallguy: Yeah I could be convinced either way. There are a few real world scenarios I'm thinking about:

1. `skip` because something broke and I don't want to fix it.
1. `skip` because I just wrote an  spec while debugging to make sure I cover a new edge case I discovered but forgot to go back and actually write the spec.
1. `pending` because we're refactoring and this is expected to fail in the interim but it still represents logic that we want covered and if it suddenly starts working when we weren't expecting it to thats a problem.

I'd like to be able to block checkin or at least alert in CI based on 1 & 2 but allow 3 with some sort of TODO comment with ticket/kanban link.

I suppose I could achieve what I want with a `#robcop disable ....` comment on the `pending` specs so definitely not a blocker.

Also wanted to confirm if 'empty' specs will be caught as well. I believe the following is also considered a pending spec (notice no block):

```ruby
it "checks a new condition I just thought of"
```

and prefixing with an _x_ should also be considered pending.
```ruby
xit "used to work but whatever YOLO!" do
  #...
end
```

patrickomatic: Hmm I'm not sure if it will catch empty blocks and report them.  I can look into adding that functionality, but if there are any other features I'd prefer to explore planning them for future work and not part of this PR?  It's starting to become difficult to continue pouring time into this

phallguy: Yeah no worries. Thank you for the work you've already done!

patrickomatic:
I just updated with master, I'm not sure why the ruby-2.1 builds are failing.  I didn't make any gem changes and it's complaining about rack versions...

```
Gem::InstallError: rack requires Ruby version >= 2.2.2.
An error occurred while installing rack (2.0.4), and Bundler cannot continue.
Make sure that `gem install rack -v '2.0.4'` succeeds before bundling.
```

bquorning:
That is weird indeed. Usually Bundler knows to use rack v1.6.9 for Ruby 2.1.

bquorning:
I could not reproduce bundler’s behavior locally, nor inside the circleci/ruby:2.1 Docker image. I have no idea what happened.

I have re-run the two failing builds in Circle, and they passed.

bquorning:
And perhaps add a line to the changelog – also mentioning that this cop is disabled by default, thanks. And thank you for your patience on this PR.

patrickomatic:
Great, ok @bquorning this is squashed, but looks like for some reason github is showing some unrelated merged commits.  I don't think it will cause a problem when merging this in, but if you'd like I can create a fresh PR that doesn't have those.  

Cheers and thanks again - excited to get this merged in

bquorning:
It seems you have rebased on the master branch of your own fork (fda73ea813), and not the master branch of the main fork (a41231d392) (probably origin/master). As you can see, the two master branches do not point to the same SHA, and that is why this PR looks a bit funky.

I think you can fix it with `git rebase --onto a41231d392 fda73ea813 pending-spec-cop` (if your local branch is also named `pending-spec-cop`).

patrickomatic:
Ah good call.  Alrighty, it's nice and clean now!

bquorning:
Thank you for the contribution.

PR #487, opened 2017-11-02
Fix to avoid return_from_stub error on empty block

yevhene:
I have a bunch of legacy code and guys use empty blocks, like:
```
before { allow(MyModule).to receive(:perform_async) {} }
```
And rubocop-rspec fails with:
```
undefined method `recursive_literal?' for nil:NilClass
/home/yevhene/.rvm/gems/ruby-2.4.1/gems/rubocop-rspec-1.19.0/lib/rubocop/cop/rspec/return_from_stub.r
b:85:in `dynamic?'
...
```
When I try to `--auto-gen-config`.
I decided to treat empty block like a static value, but I think separate cop should check this behavior in future.

### DIFF
--- a/lib/rubocop/cop/rspec/return_from_stub.rb
+++ b/lib/rubocop/cop/rspec/return_from_stub.rb
@@ -72,7 +72,7 @@ module RuboCop
           return unless block
 
           _receiver, _args, body = *block
-          unless dynamic?(body) # rubocop:disable Style/GuardClause
+          unless body && dynamic?(body) # rubocop:disable Style/GuardClause
             add_offense(
               node,
               location: :expression,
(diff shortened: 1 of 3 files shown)


lib/rubocop/cop/rspec/return_from_stub.rb
```diff
@@ -82,7 +82,7 @@ def check_block_body(node)
         end
 
         def dynamic?(node)
-          !node.recursive_literal?
+          node && !node.recursive_literal?
```
Darhazer: IMO this should be pushed to the `check_block_body` - return if body is empty

Darhazer:
Thank you for the bug report and the PR for the fix. Can you please add a changelog entry as well?

yevhene:
@Darhazer done

PR #489, opened 2017-11-03
Rspec::NestedGroups cop support --auto-gen-config

walf443:


spec/rubocop/cop/rspec/nested_groups_spec.rb
```diff
@@ -20,6 +20,8 @@
         end
       end
     RUBY
+
+    expect(cop.config_to_allow_offenses['Max']).to eq(4)
```
bquorning: I’d prefer if you added a separate spec example.

walf443:
@bquorning  fixed

spec/rubocop/cop/rspec/nested_groups_spec.rb
```diff
@@ -22,6 +22,22 @@
     RUBY
   end
 
+  it 'support --auto-gen-config' do
+    expect_offense(<<-RUBY)
```
bquorning: This expectation exists already in line 7. You can just use `inspect_source` here instead of `expect_offense`.

bquorning:
@walf443 Do you want to fix & rebase this PR, then I’ll merge it in. Or, if you prefer, I can do the changes.

PR #491, opened 2017-11-05
Fix `RSpec/SubjectStub` cop matches receive message inside all matcher

### LINKED ISSUE #488: RSpec/SubjectStub matches all matcher
Here is a patch to reproduce problem.

```
diff --git spec/rubocop/cop/rspec/subject_stub_spec.rb spec/rubocop/cop/rspec/subject_stub_spec.rb
index fc661f2..4b0aa88 100644
--- spec/rubocop/cop/rspec/subject_stub_spec.rb
+++ spec/rubocop/cop/rspec/subject_stub_spec.rb
@@ -43,6 +43,18 @@ RSpec.describe RuboCop::Cop::RSpec::SubjectStub do
     RUBY
   end

+  it 'ignores stub when inside all matcher' do
+      describe Foo do
+        subject(:foo) { [Object.new] }
+
+        it 'tries to trick rubocop-rspec' do
+            expect(foo).to all(receive(:baz))
+        end
+      end
+    RUBY
+  end
+
   it 'ignores stub within context where subject name changed' do
     expect_no_offenses(<<-RUBY)
       describe Foo do
```

In above case, subject is array, and it just stub item of subject. So it should not add offence, I think.

walf443:
fix #488


walf443:
@Darhazer rebased

bquorning:
I’m sorry for messing with your branch, @walf443. The Github UI won’t let me fix the merge conflict in an orderly fashion. Would you please do another rebase – then I’ll merge. Thanks.

Issue #417, opened 2017-05-27
Proposal: Detect use of `context` for method names instead of `describe`


andyw8:
I often see things like:

`context "#foo_bar"`

rather than

`describe "#foo_bar"`

The linter could check character of the string for a `.` or `#`.

pirj:
Addressed in #492.
Or should a companion cop be created?

andyw8:
Oh interesting. This would catch the issue, but it would be nice if it recognised `.` and `#` as special cases and recommended using `describe`.

PR #492, opened 2017-11-05
Add new `RSpec/ContextWording` cop

pirj:
### Rationale

https://github.com/reachlocal/rspec-style-guide#context-descriptions
http://www.betterspecs.org/#contexts

### Synopsis

`context` block descriptions should always start with `when` or `with`. Some may want to also use`without` as an allowed prefix.

#### Bad Example

```ruby
  # ...
end
```

#### Good Examples

```ruby
context 'when the display name is not present' do
  # ...
end

context 'with display name' do
  # ...
end

context 'without display name' do
  # ...
end
```

This pull request also addresses #417

## Checks

* [x] Added tests.


config/default.yml
```diff
@@ -43,6 +43,11 @@ RSpec/BeforeAfterAll:
   - spec/support/**/*.rb
   StyleGuide: http://www.rubydoc.info/gems/rubocop-rspec/RuboCop/Cop/RSpec/BeforeAfterAll
 
+RSpec/ContextDescription:
+  Description: "`context` block descriptions should start with 'when'."
+  Enabled: false
```
pirj: false by default because of:
```
Offenses:

spec/rubocop/cop/rspec/example_length_spec.rb:59:11: C: RSpec/ContextDescription: Start context description with when.
  context 'with CountComments enabled' do
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
spec/rubocop/cop/rspec/example_wording_spec.rb:4:11: C: RSpec/ContextDescription: Start context description with when.
  context 'with configuration' do
          ^^^^^^^^^^^^^^^^^^^^
spec/rubocop/cop/rspec/multiple_expectations_spec.rb:6:11: C: RSpec/ContextDescription: Start context description with when.
  context 'without configuration' do
          ^^^^^^^^^^^^^^^^^^^^^^^
spec/rubocop/cop/rspec/multiple_expectations_spec.rb:99:11: C: RSpec/ContextDescription: Start context description with when.
  context 'with meta data' do
          ^^^^^^^^^^^^^^^^
spec/rubocop/cop/rspec/multiple_expectations_spec.rb:135:11: C: RSpec/ContextDescription: Start context description with when.
  context 'with configuration' do
          ^^^^^^^^^^^^^^^^^^^^
spec/rubocop/cop/rspec/return_from_stub_spec.rb:8:11: C: RSpec/ContextDescription: Start context description with when.
  context 'with EnforcedStyle `and_return`' do
spec/rubocop/cop/rspec/return_from_stub_spec.rb:120:11: C: RSpec/ContextDescription: Start context description with when.
  context 'with EnforcedStyle `block`' do
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
spec/rubocop/cop/rspec/shared_context_spec.rb:4:11: C: RSpec/ContextDescription: Start context description with when.
  context 'shared_context' do
          ^^^^^^^^^^^^^^^^
spec/rubocop/cop/rspec/shared_context_spec.rb:58:11: C: RSpec/ContextDescription: Start context description with when.
  context 'shared_examples' do
          ^^^^^^^^^^^^^^^^^
spec/rubocop/cop/rspec/single_argument_message_chain_spec.rb:51:13: C: RSpec/ContextDescription: Start context description with when.
    context 'with single-key hash argument' do
spec/rubocop/cop/rspec/single_argument_message_chain_spec.rb:80:13: C: RSpec/ContextDescription: Start context description with when.
    context 'with multiple keys hash argument' do
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
spec/rubocop/rspec/example_spec.rb:39:11: C: RSpec/ContextDescription: Start context description with when.
  context 'value object semantics' do
          ^^^^^^^^^^^^^^^^^^^^^^^^
spec/rubocop/rspec/language/selector_set_spec.rb:14:11: C: RSpec/ContextDescription: Start context description with when.
  context '#include?' do
          ^^^^^^^^^^^
spec/rubocop/rspec/language/selector_set_spec.rb:24:11: C: RSpec/ContextDescription: Start context description with when.
  context '#node_pattern' do
          ^^^^^^^^^^^^^^^
spec/rubocop/rspec/language/selector_set_spec.rb:30:11: C: RSpec/ContextDescription: Start context description with when.
  context '#node_pattern_union' do
          ^^^^^^^^^^^^^^^^^^^^^
spec/rubocop/rspec/language/selector_set_spec.rb:36:11: C: RSpec/ContextDescription: Start context description with when.
  context '#send_pattern' do
          ^^^^^^^^^^^^^^^
spec/rubocop/rspec/language/selector_set_spec.rb:42:11: C: RSpec/ContextDescription: Start context description with when.
  context '#block_pattern' do
          ^^^^^^^^^^^^^^^^

149 files inspected, 17 offenses detected
```

lib/rubocop/cop/rspec/context_description.rb
```diff
@@ -0,0 +1,40 @@
...
+      #   context 'the display name not present' do
+      #     ...
+      #   end
+      #
+      #   # good
+      #   context 'when the display name is not present' do
+      #     ...
+      #   end
+      class ContextDescription < Cop
+        MSG = 'Start context description with `when`.'
```
pirj: I believe this is a false offense considering a magic `frozen_string_literal` comment above.

```
lib/rubocop/cop/rspec/context_description.rb:19:15: C: Style/MutableConstant: Freeze mutable objects assigned to constants.
        MSG = 'Start context description with `when`.'
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

config/default.yml
pirj: Or should this cop be enabled by default, but turned on in `.rubocop.yml` instead?

Drenmi:
This is crazy. I started working on this at the exact same time. 🙂 

Some comments: 

- [betterspecs.org](http://www.betterspecs.org/#contexts) also allows "with" as a prefix, as that will also produce a correct sentence when tests fail.

- There's a cop named `ExampleWording`. I think it might make sense to name this one `ContextWording`, part for uniformity, part for the possible confusion between "description" and `describe`.

The part of my implementation that supports multiple prefixes, if you want to ✂️  it:

```ruby
def on_block(node)
  context_description(node) do |description_node, message|
    return if allowed_prefix?(message)

    add_wording_offense(description_node)
  end
end

private

def allowed_prefix?(message)
  ALLOWED_PREFIXES.include?(message.split.first)
end
```

This could easily be extended to support a whitelist as well. Just merge it with the constant.

pirj:
This is double crazy considering this https://github.com/backus/rubocop-rspec/compare/master...toptal:add-context-description-cop

Will borrow best parts from your PR and @telmofcosta's one.

Drenmi:
Stars are aligning, etc. 😅

bquorning:
+1 for configurable allowed prefixes.

telmofcosta:
Nice to see so much interest in getting a cop for this 😀.

telmofcosta:
+1 to name this one `ContextWording`.

Darhazer:
I wonder whether this should check shared_context descriptions as well

pirj:
Renamed to `ContextWording`, added @telmofcosta's configurable options, added `shared_context` detection, updated from upstream master.

lib/rubocop/cop/rspec/context_wording.rb
```diff
@@ -0,0 +1,61 @@
...
+      #     ...
+      #   end
+      #
+      #   # good
+      #   context 'when the display name is not present' do
+      #     ...
+      #   end
+      class ContextWording < Cop
+        MSG = 'Start context description with %s.'.freeze
+        DEFAULT_PREFIXES = %w[when with].freeze
```
Darhazer: I think this should also include `without` as it's the opposite context of `with`

lib/rubocop/cop/rspec/context_wording.rb
```diff
@@ -0,0 +1,61 @@
...
+      # @example
+      #   # bad
+      #   context 'the display name not present' do
+      #     ...
+      #   end
+      #
+      #   # good
+      #   context 'when the display name is not present' do
+      #     ...
+      #   end
```
jonatas: Maybe it can also contain an extra example showing configurable prefixes?

```yaml
Prefix:
 - with
 - without
 - when
```

```ruby
# good
context 'without display name' do
  ...
end
```

lib/rubocop/cop/rspec/context_wording.rb
bquorning: I think there is no need for default prefixes. Shouldn’t it work by defining prefixes in the configuration file?

pirj:
```
───────────────────────────────────────────────────
modified: config/default.yml
───────────────────────────────────────────────────
@ default.yml:49 @ RSpec/BeforeAfterAll:
RSpec/ContextWording:
  Description: "`context` block descriptions should start with 'when', or 'with'."
  Enabled: true
  Prefixes:
  - when
  - with
  StyleGuide: http://www.rubydoc.info/gems/rubocop-rspec/RuboCop/Cop/RSpec/ContextWording

RSpec/DescribeClass:
───────────────────────────────────────────────────
modified: lib/rubocop/cop/rspec/context_wording.rb
───────────────────────────────────────────────────
@ context_wording.rb:29 @ module RuboCop
      #   end
      class ContextWording < Cop
        MSG = 'Start context description with %s.'.freeze
        DEFAULT_PREFIXES = %w[when with].freeze

        def_node_matcher :context_wording, <<-PATTERN
          (block (send _ { :context :shared_context } $(str $_)) ...)
@ context_wording.rb:48 @ module RuboCop
        end

        def prefixes
          cop_config['Prefixes'] || DEFAULT_PREFIXES
          cop_config['Prefixes']
        end
```

ends up with:
```
     Failure/Error: prefixes.include?(description.split(' ', 2).first)

     NoMethodError:
       undefined method `include?' for nil:NilClass
```

@bquorning Is that correct to add those default options to `config/default.yml`?

If I stub default optionns in spec, how do we make sure those options are not removed from `config/default.yml` and break everything while specs still pass?

lib/rubocop/cop/rspec/context_wording.rb
pirj: Added an example!

lib/rubocop/cop/rspec/context_wording.rb
```diff
@@ -0,0 +1,65 @@
+# frozen_string_literal: true
+
+module RuboCop
+  module Cop
+    module RSpec
+      # `context` block descriptions should start with 'when', or 'with'.
```
telmofcosta: Maybe changing this to something like `block descriptions should start with a whitelisted prefix.`?

lib/rubocop/cop/rspec/context_wording.rb
```diff
@@ -0,0 +1,65 @@
...
+      #
+      # @see https://github.com/reachlocal/rspec-style-guide#context-descriptions
+      # @see http://www.betterspecs.org/#contexts
+      #
+      # @example `Prefixes` configuration option, defaults: 'when', and 'with'
```
telmofcosta: Missing `without`.

lib/rubocop/cop/rspec/context_wording.rb
```diff
@@ -0,0 +1,65 @@
...
+      #     ...
+      #   end
+      #
+      #   # good
+      #   context 'when the display name is not present' do
+      #     ...
+      #   end
+      class ContextWording < Cop
+        MSG = 'Start context description with %s.'.freeze
+        DEFAULT_PREFIXES = %w[when with].freeze
```
pirj: @Darhazer Regarding "without"

betterspecs state it as:
> start its description with "when" or "with"

and rspec-style-guide as:
> should always start with 'when'

So I don't see a reason to add it on our side.

lib/rubocop/cop/rspec/context_wording.rb
pirj: Tried a more detailed description, it does not fit into 80 columns, and adding a newline does not meet another constraint of not having newlines in cops descriptions. ¯\_(ツ)_/¯

telmofcosta: 😄  we came across that problem also.

lib/rubocop/cop/rspec/context_wording.rb
```diff
@@ -0,0 +1,65 @@
...
+
+        def_node_matcher :context_wording, <<-PATTERN
+          (block (send _ { :context :shared_context } $(str $_)) ...)
```
bquorning: You can change the node pattern to include what `next if good_prefix?(description)` does in line 37:
```
(block (send _ { :context :shared_context } $(str #bad_prefix?)) ...)
```
The function needs to return true in the case where you want an offense, so you’ll want a `#bad_prefix?` method instead of `#good_prefix?`.

pirj: Thanks, that simplified things significantly!
Unrelated question: any idea why
`(block (send _ _ _) ...)` matches, but
`(block (send nil _ _) ...)` does not even though matched sexp is:
```
(block
  (send nil :context
    (str "something"))
  (args) nil)
```
?

`(block (send nil? _ _) ...)` works, but it's kind of a hack I think.

pirj:
@bquorning In response to [this](https://github.com/backus/rubocop-rspec/pull/492#discussion_r149199312) (can't reply in the thread for some reason):

Moved the default configuration from code to `default.yml`.
Decided not to `fetch`, but rather `||` for several reasons:
1. If it happens that a user decides to configure the cop as `Prefixes: []`, there is no difference if we `fetch` or `||`. For `Prefixes: true` there is no difference either.
2. `.fetch(..., something)` evaluates `something`, while `... || something` does not, thus saving on an array object.
3. For `Prefixes: ~`, fetch returns ... `nil`, which is quite unexpected for the code, and ends up with `An error occurred while RSpec/ContextWording cop was inspecting`.

`fetch('Prefixes') { [] }` might be an option, but same culprit with `Prefixes: ~`.

pirj:
Still don't get how to spec what is in the default config. I change it and nothing breaks, not even rubocop's checks in case default config is overridden in `.rubocop.yml` as it is with `ContextWording`.

I'd expect specs to load defaults file and populate `cop_config` in spec root context for specs with `:config` describe option.

That would remove some redundant code, would make `defaults` the only source of truth, would give more visibility how cops behave when default configuration is used (not some configuration quite similar to default as with `ExampleWording`, or no configuration given as opposed to a defined configuration in defaults).
WDYT? Does this sound like an idea for a clean-up pull request?

lib/rubocop/cop/rspec/context_wording.rb
bquorning: Yeah, RuboCop recently changed that: https://github.com/bbatsov/rubocop/pull/4815.

If I remember right, `nil?` matches an implicit nil, while `nil` matches the actual nil literal.

bquorning:
I’m fine with either way of defaulting the config to `[]`. But I don’t think performance should be the argument :-)

Loading the default config in every spec sounds like an interesting follow-up PR. I’d definiteky be interested to see that. Good idea.

pirj:
Thanks for the reviews guys!
Is this cop good to merge in its current implementation?

lib/rubocop/cop/rspec/context_wording.rb
pirj: Ha! Thanks for pointing this out.
I clearly see a `rubocop-patterns` gem and its first cop `prefer 'nil?' to 'nil' after 'send'` XD

spec/rubocop/cop/rspec/context_wording_spec.rb
```diff
@@ -0,0 +1,62 @@
...
+
+  it 'finds context without separate `when` at the beginning' do
+    expect_offense(<<-RUBY)
+      context 'whenever you do' do
+              ^^^^^^^^^^^^^^^^^ Start context description with 'when', or 'with'.
+      end
+    RUBY
+  end
+
+  context 'when configured' do
```
bquorning: This context applies to all of the specs. But perhaps have a "with default configuration" context, a "with changed configuration" context, and a "with no configuration".

pirj: I've structured the spec file in a way that root context represents "default configuration" examples, and a "configured cop" context with the same examples as in root context, but different expectations showing how differently configured cop behaves compared to default config.

I can't figure out who would anyone use the cop with empty configuration, e.g. deliberately putting an empty prefix list in config file, they will be getting weird error messages, and that unhappy path is not something I feel we need to add examples for.

bquorning: > I can't figure out who would anyone use the cop with empty configuration … that unhappy path is not something I feel we need to add examples for.

You are absolutely right.

pirj:
Squashed everything into a single commit.

lib/rubocop/cop/rspec/context_wording.rb
```diff
@@ -0,0 +1,63 @@
...
+      #   context 'the display name not present' do
+      #     ...
+      #   end
+      #
+      #   # good
+      #   context 'when the display name is not present' do
+      #     ...
+      #   end
+      class ContextWording < Cop
+        MSG = 'Start context description with %s.'.freeze
```
Drenmi: In RuboCop core, we've adopted the convention of always using tagged format strings. In this case:

```ruby
MSG = 'Start context description with %<prefixes>s.'
```

lib/rubocop/cop/rspec/context_wording.rb
```diff
@@ -0,0 +1,63 @@
...
+        def on_block(node)
+          context_wording(node) do |context|
+            add_offense(context, message: message)
+          end
+        end
+
+        private
+
+        def bad_prefix?(description)
+          !prefixes.include?(description.partition(' ').first)
```
Drenmi: You probably want to use `#split` over `#partition`, as the latter will add the match to the result. `String#split` also defaults to splitting on spaces. 🙂

lib/rubocop/cop/rspec/context_wording.rb
```diff
@@ -0,0 +1,63 @@
...
+          cop_config['Prefixes'] || []
+        end
+
+        def message
+          format(MSG, sentence)
+        end
+
+        def sentence
+          quoted = prefixes.map { |prefix| "'#{prefix}'" }
+          return quoted if quoted.length == 1
```
Drenmi: Would be neat if this used `#size`, or `#one?` since the method operates on both collections and strings. I found myself thinking: "whenever would we get a string of length 1?" 🙂

lib/rubocop/cop/rspec/context_wording.rb
```diff
@@ -0,0 +1,63 @@
...
+
+        def prefixes
+          cop_config['Prefixes'] || []
+        end
+
+        def message
+          format(MSG, sentence)
+        end
+
+        def sentence
```
Drenmi: There is probably a subtle bug here, that's missing a test case. The problem is this method can return *either* a string, or an array.

`Kernel#format` will simply call `#to_s` on its argument, so in the case where you have a single prefix configured:

> ```
> "Start context description with [\"when\"]."
> ```

lib/rubocop/cop/rspec/context_wording.rb
Drenmi:
```ruby
" foo bar".split
#=> ["foo", "bar"]

" foo bar".partition(" ")
#=> ["", " ", "foo bar"]
```

pirj: It's so tempting to switch it back to `partition` now when you mention this.
If it's worth registering an offense for descriptions starting from whitespace as well.

pirj:
Good to merge?

bquorning:
@Drenmi All good?

Drenmi:
@bquorning: LGTM!

bquorning:
Thank you @pirj

lib/rubocop/cop/rspec/context_wording.rb
```diff
@@ -0,0 +1,63 @@
...
+      #
+      #   # good
+      #   context 'when the display name is not present' do
+      #     ...
+      #   end
+      class ContextWording < Cop
+        MSG = 'Start context description with %<prefixes>s.'.freeze
+
+        def_node_matcher :context_wording, <<-PATTERN
+          (block (send _ { :context :shared_context } $(str #bad_prefix?)) ...)
```
pirj: This doesn't seem to catch contexts with tags:
```
context 'lol missed!', haha: 'lol' do
...
```

halilim:
Sorry, a bit late to the party. I am trying to understand why `shared_context` is included in this cop. Is it always supposed to be used with examples in a block like this?

```ruby
include_context 'when foo' do
  it { is_expected.to eq('bar') }
end
```

Because I often use shared contexts to do some setup across different tests in models, controllers and other stuff like

```ruby
shared_context 'foo setup' do
  let(:bar) { 1 }

  before do
    allow(baz).to receive(:message).and_return(2)
  end
end

describe '#something' do
  include_context 'foo setup'

  it 'is an example' do
    is_expected.to eq(3)
  end
end
```
, and it does not make sense to use "when" et. al. in this case.

pirj:
@halilim It's hard to reason about this with a synthetic example.
I would rather expect something:
```
include_context 'with several other existing books'
```
or
```
include_context 'when user is logged in'
```

Speaking of an `Include_context` does include context to the section it is called from, so using the syntax with a nested example is weird, e.g.:
```
describe do
  include_context 'when foo' do
    it { is_expected.to eq('bar') }
  end

  include_context 'when bar' do
    it { is_expected.to eq('bar') }
  end

  # what would it be here, a `bar` or a `foo`?
end
```

This creates an ambiguity and I believe is equal to:
```
describe do
  include_context 'when foo'
  it { is_expected.to eq('bar') }

  include_context 'when bar'
  it { is_expected.to eq('bar') }
end
```

But I find this kind of scattered setup confusing.

halilim:
Just to be clear, I was not advocating the use of nested examples, it was just one of the ways I could imagine this cop supports.

So let me try to come up with a closer to real world example. Imagine an external service is called in a certain method, and the method is called in a controller. Ideally we'd want to test both the method and the controller.

So we'd need to stub the external service in both places. I use shared contexts for this, and "with the external service" is not very descriptive. Because it's not actual context, e.g. the external service is always there. It just needs to be stubbed out.

pirj:
@halilim I clearly understand the case.

What does this service do? Say, it's a `ContinousIntegrationService`, and you check if the build has succeeded.
What do you think of naming a shared context where you stub the result of the service request out so that it returns `true`, a "when build is successful", or a "with a successful build"?

PR #493, opened 2017-11-05
[Fix #432] Add new FactoryGirl/DynamicAssociationDeclaration cop

### LINKED ISSUE #432: Cop idea: Replace FactoryGirl association shorthand with call to `#association`
I have started preferring `association :foo, ...` over the shorthand `foo` in factories.

It would be nice to have a cop for this. 🙂

Drenmi:
This adds a new FactoryBot cop as outlined in #432. It checks for dynamic association declarations in factories, and suggests replacing them with calls to `#association`, which is more explicit, and supports configuration options.


config/default.yml
```diff
@@ -335,6 +335,11 @@ Capybara/FeatureMethods:
   Enabled: true
   StyleGuide: http://www.rubydoc.info/gems/rubocop-rspec/RuboCop/Cop/RSpec/Capybara/FeatureMethods
 
+FactoryGirl/DynamicAssociationDeclaration:
+  Description: Prefer declaring associations using `#association`.
```
Darhazer: you have to sync this with the first comment in the cop itself. Run `bin/build_config` - it will do the job

lib/rubocop/cop/rspec/factory_girl/dynamic_association_declaration.rb
```diff
@@ -0,0 +1,41 @@
...
+        #   end
+        #
+        #   # good
+        #   factory :post do
+        #     association :comments
+        #   end
+        class DynamicAssociationDeclaration < Cop
+          MSG = 'Use `association :%<name>s` to declare association.'.freeze
+
+          def_node_search :dynamic_attribute_declarations, <<-PATTERN
```
Darhazer: maybe :dynamic_association_declarations will be a better name

koic:
Since the namespace was changed from `FactoryGirl` to `FactoryBot` at #490, it seems good to make a change accordingly 😃

lib/rubocop/cop/rspec/factory_girl/dynamic_association_declaration.rb
Drenmi: Whoops. Good catch @Darhazer! 🙂

Drenmi:
> Since the namespace was changed from FactoryGirl to FactoryBot at #490, it seems good to make a change accordingly 😃

I used FactoryGirl in all places except the namespaces. I think this needs to be a separate PR with deprecation warnings.

Drenmi:
> I tried on a local codebase and it seems it doesn't register offence if there is anything in the factory but the offending code. Also it triggers false positives on legitimate code, like calling skip_create in the factory or referencing a trait

Thanks for the feedback, @Darhazer. Will add some test cases, and exclusions for false positives. Guess I'll need to go dig deeper into FactoryBot documentation. 😅

walf443:
> I used FactoryGirl in all places except the namespaces. I think this needs to be a separate PR with deprecation warnings.

I think this cop also seems to work under FactoryBot namespace.
So just renaming cop name to FactoryBot, is not compatibility problem, I think.

CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 
 * Rename namespace `FactoryGirl` to `FactoryBot` following original library update. ([@walf443][])
 * Fix exception in `RSpec/ReturnFromStub` on empty block. ([@yevhene][])
+* Add `FactoryGirl/DynamicAssociationDeclaration` cop. ([@drenmi][])
```
Darhazer: FactoryBot

Drenmi:
@Darhazer: Sorry. Have been caught up with work. What remains is to remove false positives for traits referenced from inside other traits.

My strategy is to skip up to the factory root, then scan through all the immediate children and collect any defined trait names in a list. Does this seem sound? Or can you think of any cases where this would not work?

Darhazer:
> My strategy is to skip up to the factory root, then scan through all the immediate children and collect any defined trait names in a whitelist.

I would to exactly the same.

Please check also the case when there is association followed by other properties, as in my initial tests those were not flaged:
```
FactoryGirl.define do
  factory :user do
    country
    sequence(:email) { |n| "user#{n}@example.com" }
```

Drenmi:
> I would to exactly the same.

I will proceed with this approach, then. 🙂 

> Please check also the case when there is association followed by other properties, as in my initial tests those were not flaged

I fixed this already. 😊

Darhazer:
btw with something like `def_node_search :registered_trait?, ' (send nil :trait (sym %))'` and then use it like `registered_trait?(node, name)`

lib/rubocop/cop/rspec/factory_bot/dynamic_association_declaration.rb
```diff
@@ -0,0 +1,55 @@
...
+        #   factory :post do
+        #     association :comments
+        #   end
+        class DynamicAssociationDeclaration < Cop
+          MSG = 'Use `association :%<name>s` to declare association.'.freeze
+
+          FACTORY_BOT_METHODS = %i[skip_create].freeze
+
+          def_node_matcher :dynamic_association_declaration, <<-PATTERN
+            (send nil? [!#whitelisted_method? $_])
```
pirj: Will `$!#whitelisted_method?` work?

spec/rubocop/cop/rspec/factory_bot/dynamic_association_declaration_spec.rb
```diff
@@ -0,0 +1,80 @@
...
+
+# rubocop:disable Metrics/LineLength
+RSpec.describe RuboCop::Cop::RSpec::FactoryBot::DynamicAssociationDeclaration do
+  subject(:cop) { described_class.new(config) }
+
+  let(:config) { RuboCop::Config.new }
+
+  it 'registers an offense for offending code' do
+    expect_offense(<<-RUBY)
+      FactoryGirl.define do
```
pirj: This is part of all examples, but it's not being considered in the cop itself.

pirj:
The development has stalled. Nobody picked it up.

Also, the implicit association syntax is similar to inline global attribute definition and inclusion of traits, which make it ambiguous, and risks false positives.

Besides, the implicit style is a pretty valid syntax. I'm not aware of any deprecations.
Just as everything about FactoryGirl, it needs a deep thought when used, and the fact that it comes with a caveat doesn't mean it should be always avoided.

See https://github.com/rubocop-hq/rubocop-rspec/issues/432#issuecomment-667682937

PR #495, opened 2017-11-07
Implement autocorrect for ReturnFromStub

bquorning:
~So far just for the `and_return` config.~ Works for both `block` and `and_return` configurations. It does however not work for HEREDOCs.

@Darhazer I remember you were working on this in #422. I hacked this one autocorrector together today and it seems to work. Do you have some test cases lying around? You’re welcome to push to this branch if you want to contribute.

-----------------

* [x] Added tests.

### DIFF
--- a/lib/rubocop/cop/rspec/return_from_stub.rb
+++ b/lib/rubocop/cop/rspec/return_from_stub.rb
@@ -53,6 +53,14 @@ module RuboCop
           end
         end
 
+        def autocorrect(node)
+          if style == :block
+            AndReturnCallCorrector.new(node)
+          else
+            BlockBodyCorrector.new(node)
+          end
+        end
+
         private
 
         def check_and_return_call(node)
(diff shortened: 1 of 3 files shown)


Darhazer:
See #426
I’ll review in details a bit later today

Darhazer:
@bquorning regarding test cases, empty block turned out to be an edge case you should cover. You should return `nil` there. See #487
Also I would test the auto-correct on a do/end style of a block, like in this spec:
```
allow(Foo).to receive(:bar) do
  "You called" \
  "me"
end
```

HEREDOCS are usually the trickest part, you may try them, at least in order to ignore them if it generates invalid syntax

bquorning:
@Darhazer I’ve updated the code to skip heredocs and (hopefully) autocorrect everything else.

bquorning:
Hmm, seems I need to add extra braces around the hash when changing from `and_return(foo: 42)` to ` { foo: 42 }`.

lib/rubocop/cop/rspec/return_from_stub.rb
```diff
@@ -84,6 +92,99 @@ def check_block_body(node)
...
+              receiver.source_range.end_pos,
+              node.source_range.end_pos
+            )
+          end
+
+          def multiline_replacement
+            buffer = StringIO.new
+            buffer << " do\n"
+            args.source.lines.each do |line|
+              buffer << "#{indentation}#{line.strip}\n"
```
bquorning: How much indentation logic does this cop need to know? I assume that the Layout cops will be run again after each autocorrect anyway, and fix any (most of the) wrong indentation.

Darhazer: IMO we should not go there. Even if we replace with `{ }` type of block, it's another cop that will say (and auto-correct) multi-line blocks to do/end.

I'll test on a real code-base tomorrow and let you know how it works.

lib/rubocop/cop/rspec/return_from_stub.rb
```diff
@@ -84,6 +92,83 @@ def check_block_body(node)
         def dynamic?(node)
           !node.recursive_literal?
         end
+
+        # :nodoc:
+        class AndReturnCallCorrector
```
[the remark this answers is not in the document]
bquorning: > fix the description_extractor to ignore classes that are not cops

This is definitely the way to go. I may need to call on @backus to help me…

bquorning:
@Darhazer It appears I have a green build here 🎉 Would you please give the final review before merging?

PR #499, opened 2017-11-09
Let DescriptionExtractor ignore helper classes

bquorning:
In some cases, it is desirable to group some logic together in a "helper" class *inside* a cop. But the `DescriptionExtractor` would mis-identify these classes as RSpec cops and fail while trying to read their configuration from default.yml.

We can avoid this by only including classes that inherit from `RuboCop::Cop` or `RuboCop::Cop::RSpec::Cop`. The cops in this project (currently) always inherits from `RuboCop::Cop::RSpec::Cop`, but YARD superclass resolution is a bit flaky: All classes loaded before `RuboCop::Cop::WorkaroundCop` are shown as having `RuboCop::Cop` as superclass, while all the following classes are listed as having `RuboCop::Cop::RSpec::Cop` as their superclass.

Examples of where helper classes would be useful: #476, #495.

-----------------

* [x] Added tests.
* [ ] ~Added an entry to the [Changelog](../blob/master/CHANGELOG.md) if the new code introduces user-observable changes.~

### DIFF
--- a/lib/rubocop/rspec/description_extractor.rb
+++ b/lib/rubocop/rspec/description_extractor.rb
@@ -19,6 +19,7 @@ module RuboCop
 
       # Decorator of a YARD code object for working with documented rspec cops
       class CodeObject
+        COP_CLASS_NAMES = %w[RuboCop::Cop RuboCop::Cop::RSpec::Cop].freeze
         RSPEC_NAMESPACE = 'RuboCop::Cop::RSpec'.freeze
 
         def initialize(yardoc)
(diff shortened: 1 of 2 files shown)


lib/rubocop/rspec/description_extractor.rb
```diff
@@ -61,6 +65,14 @@ def documented_constant
           yardoc.to_s
         end
 
+        def inherits_from_cop?
+          # YARD superclass resolution is a bit flaky: All classes loaded before
+          # RuboCop::Cop::WorkaroundCop are shown as having RuboCop::Cop as
+          # superclass, while all the following classes are listed as having
+          # RuboCop::Cop::RSpec::Cop as their superclass.
+          COP_CLASS_NAMES.include?(yardoc.superclass.path)
```
Darhazer: Alternatively, `yardoc.superclass.path.include?('RuboCop::Cop')` will work for anything under the RuboCop::Cop namespace

bquorning: It would, but by coincidence. I prefer this more explicit working around the Yard bug.

PR #509, opened 2017-11-27
Fix violations of Lint/MissingCopEnableDirective

bquorning:
Rubocop master branch have a few new cops. This PR makes us comply with Lint/MissingCopEnableDirective.

-----------------

* [ ] Added tests.


spec/rubocop/cop/rspec/factory_bot/dynamic_attribute_defined_statically_spec.rb
```diff
@@ -89,3 +89,4 @@
     end
   end
 end
+# rubocop:enable Metrics/LineLength
```
Darhazer: you may re-enable right after the 4th line, which is the only one offending

bquorning: Fixed now @Darhazer

Issue #451, opened 2017-09-05
Detect aggregate_failures set via metadata


AlexVPopov:
Hello,

I have enabled `aggregate_failures` [globally](https://relishapp.com/rspec/rspec-core/docs/expectation-framework-integration/aggregating-failures#enable-failure-aggregation-globally-using-`define-derived-metadata`):
```ruby
config.define_derived_metadata { |meta| meta[:aggregate_failures] = true }
```
I still get violations from `RSpec/MultipleExpectations`. Is this the expected behavior? What is the solution/workaround - should I just turn off the cop?

Darhazer:
As this is static analysis tool, I don't think we can catch such global configuration. In this case you can safely disable the cop, as you will always have just one aggregate_failures block per example (unless you override with `aggregate_failures: false` in some example.

AlexVPopov:
Thank you :)

PR #511, opened 2017-12-11
Add AggregateFailuresByDefault configuration for RSpec/MultipleExpectations cop

onk:
Support cases that `aggregate_failures` metadata is set in spec_helper.
In that case, examples without metadata will not detect offence.

> If you want to enable this feature everywhere,
> you can use define_derived_metadata:
>
> ```
> RSpec.configure do |c|
>   c.define_derived_metadata do |meta|
>     meta[:aggregate_failures] = true unless meta.key?(:aggregate_failures)
>   end
> end
> ```

see: http://rspec.info/blog/2015/06/rspec-3-3-has-been-released/

Related Issues: #451 
With this option, It can detect offense when the example has `aggregate_failures: false` metadata.

-----------------

* [x] Added tests.

### DIFF
--- a/spec/rubocop/cop/rspec/multiple_expectations_spec.rb
+++ b/spec/rubocop/cop/rspec/multiple_expectations_spec.rb
@@ -132,7 +132,7 @@ RSpec.describe RuboCop::Cop::RSpec::MultipleExpectations, :config do
     end
   end
 
-  context 'with configuration' do
+  context 'with Max configuration' do
     let(:cop_config) do
       { 'Max' => '2' }
     end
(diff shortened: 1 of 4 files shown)


spec/rubocop/cop/rspec/multiple_expectations_spec.rb
```diff
@@ -3,7 +3,7 @@
 RSpec.describe RuboCop::Cop::RSpec::MultipleExpectations, :config do
   subject(:cop) { described_class.new(config) }
 
-  context 'without configuration' do
+  context 'when without configuration' do
```
Darhazer: `without` is valid prefix for contexts

onk: Fixed! @Darhazer

PR #512, opened 2017-12-13
Fix RuboCop v0.52.0 compatibility

bquorning:
RuboCop v0.52.0 has been released, and we were not completely compatible.

-----------------

* [ ] Added tests.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -36,3 +36,11 @@ RSpec/ExampleLength:
 RSpec/DescribeClass:
   Exclude:
     - spec/project/**/*.rb
+
+# FIXME
+Style/FormatStringToken:
+  Enabled: false
+
+# Awaiting bugfix: https://github.com/bbatsov/rubocop/issues/5224
+Layout/EmptyLinesAroundArguments:
+  Enabled: false
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -37,10 +37,6 @@ RSpec/DescribeClass:
   Exclude:
     - spec/project/**/*.rb
 
-# FIXME
-Style/FormatStringToken:
-  Enabled: false
-
 # Awaiting bugfix: https://github.com/bbatsov/rubocop/issues/5224
 Layout/EmptyLinesAroundArguments:
   Enabled: false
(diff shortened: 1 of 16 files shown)


spec/rubocop/rspec/example_spec.rb
```diff
@@ -9,12 +9,12 @@ def example(source)
...
   it 'extracts doc string' do
     expect(example("it('does x') { foo }").doc_string)
-      .to eql(s(:str, 'does x'))
+      .to eql(RuboCop::AST::StrNode.new(:str, ['does x']))
   end
 
   it 'extracts doc string for unimplemented examples' do
     expect(example("it('does x')").doc_string)
-      .to eql(s(:str, 'does x'))
+      .to eql(RuboCop::AST::StrNode.new(:str, ['does x']))
```
bquorning: @pocke Does this look right? It is a bit confusing that I can’t tell the difference between `Node#inspect` and `StrNode#inspect`:
```
    Failure/Error:
       expect(example("it('does x')").doc_string)
         .to eql(s(:str, 'does x'))

       expected: s(:str, "does x")
            got: s(:str, "does x")
```

.rubocop.yml
```diff
@@ -36,3 +36,11 @@ RSpec/ExampleLength:
 RSpec/DescribeClass:
   Exclude:
     - spec/project/**/*.rb
+
+# FIXME
+Style/FormatStringToken:
```
Darhazer: I can fix our codebase regarding this ;)

spec/rubocop/rspec/example_spec.rb
pocke: `AST::Node#eql?` checks classes. https://github.com/whitequark/ast/blob/241cac1bea04fd7dfd9a83959713d7fa12b4e2ec/lib/ast/node.rb#L85-L89

Maybe the test cases does not expect the class equivalent. So I think we can replace `.to eql` with `.to eq`, and use `s` instead of `RuboCop::AST::StrNode.new`.

.rubocop.yml
bquorning: That would be great ❤️

spec/rubocop/rspec/example_spec.rb
bquorning: Thank you – I’ll switch to using `eq` instead.

.rubocop.yml
Darhazer: @bquorning Done, please review

PR #520, opened 2017-12-30
Add RSpec/ExpectChange cop. Fixes #515

### LINKED ISSUE #515: Cop idea: force consistent usage of change matcher
> You can specify what will change using either of two forms:
> expect { do_something }.to change(object, :attribute)
> expect { do_something }.to change { object.attribute }
> 

While migrating a code base to comply with Rubocop 0.52, I noticed that a lot of `change { }` blocks could be used in the non-block form. It would be nice to have a cop that recommends this usage

Darhazer:
> You can specify what will change using either of two forms:
> expect { do_something }.to change(object, :attribute)
> expect { do_something }.to change { object.attribute }

This cops allows for forcing consistent usage of either one of the two forms.

-----------------

* [x] Added tests.


lib/rubocop/cop/rspec/expect_change.rb
```diff
@@ -0,0 +1,102 @@
...
+          expect_change_with_arguments(node) do |receiver, message|
+            add_offense(
+              node,
+              message: format(MSG_CALL, obj: receiver, attr: message)
+            )
+          end
+        end
+
+        def on_block(node)
+          return unless style == :method_call
```
bquorning: Add a blank line after this one (or remove blank line after `return` in `#on_send`). I wonder if it would be possible to have a cop check where blank lines should be.

lib/rubocop/cop/rspec/expect_change.rb
```diff
@@ -0,0 +1,102 @@
...
+
+        def_node_matcher :expect_change_with_block, <<-PATTERN
+          (block
+            (send nil? :change)
+            (args)
+            (send ({const send} nil? $_) $_)
+          )
+        PATTERN
+
+        def_node_matcher :expect_change_with_arguments, <<-PATTERN
```
bquorning: Pardon my OCD, but could you define this method before defining `expect_change_with_block`? That is the order in which the methods are used throughout the class.

Darhazer: I value your OCD.

lib/rubocop/cop/rspec/expect_change.rb
Darhazer: That would be a nice cop to add to RuboCop - I prefer having a blank line after early return

Darhazer:
PR updated. btw I couldn't find out why the 2.1 build doesn't work on Travis

PR #522, opened 2018-01-02
Use range_by_whole_lines helper method

Darhazer:
A nice helper I found while working on #476

-----------------

* [ ] Added tests.


bquorning:
Nice one!

PR #524, opened 2018-01-03
Add RSpec/ExampleWithoutDescription cop. Fixes #516

### LINKED ISSUE #516: Cop idea: it without description
RSpec allows for auto-generated descriptions, based on the expectation used. I personally don't like this style. But even if used, I've seen people put an empty string as description, instead of omitting it completely. So I propose cop that detect this and removes the empty string, as well as having an option to completely disable usage of it without a description

Darhazer:
Forbids usage of empty string as a description and also allows users to configure when it's allowed to have an example without any description 

-----------------

* [x] Added tests.


lib/rubocop/cop/rspec/example_without_description.rb
```diff
@@ -0,0 +1,88 @@
...
+      #
+      # @example `EnforcedStyle: always_allow`
+      #   # bad
+      #   it('') { is_expected.to be_good }
+      #   it '' do
+      #     result = service.call
+      #     expect(result).to be(true)
+      #   end
+      #
+      #   it { is_expected.to be_good }
```
bquorning: These are `# good`, right? (missing header)

lib/rubocop/cop/rspec/example_without_description.rb
```diff
@@ -0,0 +1,88 @@
...
+      #
+      # @example `EnforcedStyle: single_line_only`
+      #   # bad
+      #   it('') { is_expected.to be_good }
+      #   it do
+      #     result = service.call
+      #     expect(result).to be(true)
+      #   end
+      #
+      #   good
```
bquorning: `#` missing.

lib/rubocop/cop/rspec/example_without_description.rb
```diff
@@ -0,0 +1,88 @@
...
+        def_node_matcher :example?, Examples::ALL.send_pattern
+        def_node_matcher :example_description, '(send nil? _ $(str $_))'
+
+        def on_send(node)
+          return unless example?(node)
+
+          check_example_without_description(node)
+
+          example_description(node) do |message_node, message|
+            return unless message.to_s.empty?
```
bquorning: Just curious – what might `message` be besides a string?

lib/rubocop/cop/rspec/example_without_description.rb
```diff
@@ -0,0 +1,88 @@
...
+        private
+
+        def check_example_without_description(node)
+          return unless dissalow_empty_description?(node)
+
+          _send, _method, arg = *node
+          add_offense(node, message: MSG_ADD_DESCRIPTION) if arg.nil?
+        end
+
+        def dissalow_empty_description?(node)
```
bquorning: Typo: dissalow -> disallow.

lib/rubocop/cop/rspec/example_without_description.rb
```diff
@@ -0,0 +1,88 @@
...
+          return true if style == :single_line_only && node.parent.multiline?
+
+          style == :disallow
+        end
```
bquorning: I’m always trying to come up with alternatives to `return <boolean logic>\n<other boolean logic>`. How about this:
```rb
def disallow_empty_description?(node)
  style == :disallow ||
    (style == :single_line_only && node.parent.multiline?)
end
```

lib/rubocop/cop/rspec/example_without_description.rb
```diff
@@ -0,0 +1,88 @@
...
+            return unless message.to_s.empty?
+
+            add_offense(message_node, message: MSG_DEFAULT_ARGUMENT)
+          end
+        end
+
+        private
+
+        def check_example_without_description(node)
+          return unless dissalow_empty_description?(node)
```
bquorning: I’d prefer the `arg.nil?` check earlier, before the `style` check here.

lib/rubocop/cop/rspec/example_without_description.rb
Darhazer: Nothing else, I just have the habit of converting to the expected type

lib/rubocop/cop/rspec/example_without_description.rb
Darhazer: I'm not fan either, but I also dislike mixing || and &&. Anyway, I'll use your version, as it looks cleaner indeed

bquorning:
Please rebase on master to get a green build. A PR description would be nice, too :-)

PR #526, opened 2018-01-08
add ruby 2.5 to the travis build matrix

yob:
Ruby 2.5 was released ~2 weeks ago, so I thought it'd be interesting to see if the build passes


bquorning:
Good call.

Travis seems to be confused about which version of bundler to use (1.16.1 vs 1.15.4) – and I wonder what’s up with `bundler/dep_proxy`…

bquorning:
@yob Thank you for the PR, but Travis didn’t like it. In fact, I got so tired of Travis that rubocop-rspec now is using Circle CI instead, and is testing against Ruby 2.5 there.

PR #529, opened 2018-01-09
Improve documentation

Darhazer:
* fix invalid Ruby syntax in some examples, preventing syntax-colored output
* made the indentation of good and bad examples consistent
* tagged see more as `@see`

-----------------

* [ ] Added tests.


bquorning:
Where do you see that the documentation is wrong? How do we generate documentation?

dgollahon:
As an aside, it would be kind of neat if the test suite could automatically parse out the inline examples and make sure the cop works on them.

Darhazer:
@bquorning I saw non-colored examples at http://www.rubydoc.info/gems/rubocop-rspec/RuboCop/Cop/RSpec and got curious; then I used `yard doc` to generated locally

PR #533, opened 2018-01-12
Fix with method in RSpec/ReturnFromStub

flyerhzm:
`RSpec/ReturnFromStub` cop ignores `with` method right now, e.g. it replaces

```
allow(Foo).to receive(:bar).with(1) {}
```

with

```
allow(Foo).to receive(:bar).and_return(nil)
```

-----------------

* [x] Added tests.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Master (Unreleased)
 
+* Fix `RSpec/ReturnFromStub` cop for `with` method. ([@flyerhzm][])
+
 ## 1.22.0 (2018-01-10)
 
 * Updates `describe_class` to account for RSpecs `:system` wrapper of rails system tests. ([@EliseFitz15][])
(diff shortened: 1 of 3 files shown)


bquorning:
Good find. We apparently completely overlooked this use case 😳 

Could you also please add spec examples similar to the existing “it 'finds static values returned from block'” examples, so we are sure the offense message is displayed correct for both configurations?

Would it be possible (and reasonable) to make a single node matcher instead of having 2? I don’t like how `#check_and_return_call` and `#check_and_return_with_call` are almost identical methods.

Darhazer:
This definitely isn't the way to go, as we will miss `receive(:msg).once {`
IMO we should move the offence to the block to avoid such false positives
Like:
```
           unless body && dynamic?(body) # rubocop:disable Style/GuardClause
             add_offense(
-              node,
-              location: :expression,
+              block,
+              location: :begin,
               message: MSG_AND_RETURN
             )
           end
```

Darhazer:
I have pushed a WIP branch [fix-return-from-stub-chained-calls](https://github.com/backus/rubocop-rspec/compare/fix-return-from-stub-chained-calls?expand=1)

Fixed blocks correct to method call, should fix the opposite.

flyerhzm:
@Darhazer okay, I'll leave it to you.

PR #535, opened 2018-01-13
Refined Circle CI configuration

bquorning:
With this PR we run a number of small, fast jobs:

- a `confirm_config` job running on Ruby 2.5.
- an `rspec` job and an `rubocop` job for each of the Ruby versions 2.1, 2.2, 2.3, 2.4 and 2.5.
- a combined `rspec` and `rubocop` job running on JRuby, with caching around the `bundle install` step to make it a bit faster.

We can run 4 jobs in parallel on Circle CI. Since the JRuby job is significantly slower than any of the other, I wanted to ensure that it one of the first 4 being started. I accomplished this by making all the regular `rspec` and `rubocop` jobs being “dependent” on the very fast `confirm_config` job:

circle-ci-workflow

Example build time before: [1:39](https://circleci.com/workflow-run/8c8c6d0d-c779-48d0-a2f3-e5ea9cf784c0)
Example build time after: [1:13](https://circleci.com/workflow-run/e28bed8c-9a46-4238-98ce-d70d7b75b8ee)

Actual build times will of course vary a bit (~15 seconds).

-----------------

* [ ] Added tests.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -77,7 +77,16 @@ jobs:
       - image: circleci/jruby:9.1
     steps:
       - checkout
-      - run: bundle install
+      - run: bundle lock
+      - restore_cache:
+          keys:
+            - bundle-v1-{{ checksum "Gemfile.lock" }}
+            - bundle-v1-
+      - run: bundle install --path vendor/bundle
+      - save_cache:
+          key: bundle-v1-{{ checksum "Gemfile.lock" }}
+          paths:
+            - vendor/bundle
       - run: rake internal_investigation spec
 
 workflows:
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -1,64 +1,136 @@
...
+            - bundle-v1-{{ checksum "Gemfile.lock" }}
+            - bundle-v1-
+      - run: bundle install --path vendor/bundle
+      - save_cache:
+          key: bundle-v1-{{ checksum "Gemfile.lock" }}
+          paths:
+            - vendor/bundle
+      - restore_cache:
+          keys:
+            - rubocop-v1-{{ .Environment.CIRCLE_JOB }}
```
bquorning: Hmm, on second thought, this isn’t a good cache key. The Circle caches are immutable once written, and this cache key would only work if the cache could be appended to.

PR #546, opened 2018-02-06
[#544] ADD Rails/HttpStatus cop

### LINKED ISSUE #544: [Cop idea] ADD cop that check have_http_status value
There is two ways to check the http status of a request with rspec: with the numeric or with symbol value.

```ruby
it { is_expected.to have_http_status :ok }
it { is_expected.to have_http_status 200 }
```

That could be a good cop idea to force the way to check status and set to numeric or symbol value ! What do you think about this ?

anthony-robin:
This cop enforces consistent use of `symbol` or `numeric`
value to describe HTTP status.

Example `EnforcedStyle: symbol` (default)
```ruby
  # bad
  it { is_expected.to have_http_status 200 }
  it { is_expected.to have_http_status 404 }

  # good
  it { is_expected.to have_http_status :ok }
  it { is_expected.to have_http_status :not_found }
  it { is_expected.to have_http_status :success }
  it { is_expected.to have_http_status :error }
```

Example `EnforcedStyle: numeric`
```ruby
  # bad
  it { is_expected.to have_http_status :ok }
  it { is_expected.to have_http_status :not_found }

  # good
  it { is_expected.to have_http_status 200 }
  it { is_expected.to have_http_status 404 }
  it { is_expected.to have_http_status :success }
  it { is_expected.to have_http_status :error }
```

Closes #544

-----------------

* [x] Added tests.

### DIFF
--- a/config/default.yml
+++ b/config/default.yml
@@ -368,3 +368,12 @@ FactoryBot/DynamicAttributeDefinedStatically:
   Description: Prefer declaring dynamic attribute values in a block.
   Enabled: true
   StyleGuide: http://www.rubydoc.info/gems/rubocop-rspec/RuboCop/Cop/RSpec/FactoryBot/DynamicAttributeDefinedStatically
+
+Rails/HttpStatus:
+  Description: Enforces use of symbolic or numeric value to describe HTTP status.
+  Enabled: true
+  EnforcedStyle: symbol
+  SupportedStyles:
+  - numeric
+  - symbol
+  StyleGuide: http://www.rubydoc.info/gems/rubocop-rspec/RuboCop/Cop/RSpec/Rails/HttpStatus
--- a/lib/rubocop/cop/rspec/rails/http_status.rb
+++ b/lib/rubocop/cop/rspec/rails/http_status.rb
@@ -1,0 +1,126 @@
+# frozen_string_literal: true
+
+module RuboCop
+  module Cop
+    module RSpec
+      module Rails
+        # Enforces use of symbolic or numeric value to describe HTTP status.
+        #
+        # @example `EnforcedStyle: symbol` (default)
+        #   # bad
+        #   it { is_expected.to have_http_status 200 }
+        #   it { is_expected.to have_http_status 404 }
+        #
+        #   # good
+        #   it { is_expected.to have_http_status :ok }
+        #   it { is_expected.to have_http_status :not_found }
+        #   it { is_expected.to have_http_status :success }
+        #   it { is_expected.to have_http_status :error }
+        #
+        # @example `EnforcedStyle: numeric`
+        #   # bad
+        #   it { is_expected.to have_http_status :ok }
+        #   it { is_expected.to have_http_status :not_found }
+        #
+        #   # good
+        #   it { is_expected.to have_http_status 200 }
+        #   it { is_expected.to have_http_status 404 }
+        #   it { is_expected.to have_http_status :success }
+        #   it { is_expected.to have_http_status :error }
+        #
+        class HttpStatus < Cop
+          begin
+            require 'rack/utils'
+            AUTOCORRECTS = true
+          rescue LoadError
+            AUTOCORRECTS = false
+          end
+
+          include ConfigurableEnforcedStyle
+
+          MSG = 'Prefer `%<prefer>s` over `%<current>s` '\
+                'to describe HTTP status code.'.freeze
+          WHITELIST_STATUS = %i[error success missing redirect].freeze
+
+          def_node_matcher :http_status, <<-PATTERN
+            (send nil? :have_http_status ${int sym})
+          PATTERN
+
+          def on_send(node) # rubocop:disable Metrics/CyclomaticComplexity
+            http_status(node) do |ast_node|
+              if style == :numeric
+                return if ast_node.int_type?
+                return if ast_node.sym_type? &&
+                    WHITELIST_STATUS.include?(ast_node.value)
+              end
+
+              return if style == :symbol && ast_node.sym_type?
+
+              add_offense(ast_node)
+            end
+          end
+
+          def message(node)
+            prefer, current = message_without_autocorrect unless AUTOCORRECTS
+            prefer, current = message_for_autocorrect(node) if AUTOCORRECTS
+
+            format(MSG, prefer: prefer, current: current)
+          end
+
+          if AUTOCORRECTS
+            def autocorrect(node)
+              replacement = new_value(node)
+              return if replacement.nil?
+
+              lambda do |corrector|
+                corrector.replace(node.loc.expression, replacement.to_s)
+              end
+            end
+          end
+
+          private
+
+          def new_value(node)
+            case style
+            when :numeric
+              numeric, = symbolic_to_numeric_value(node)
+              numeric
+            when :symbol
+              symbol, = numeric_to_symbolic_value(node)
+              symbol
+            end
+          end
+
+          def numeric_to_symbolic_value(node)
+            numeric = node.source.to_i
+            symbol = ":#{::Rack::Utils::SYMBOL_TO_STATUS_CODE.key(numeric)}"
+
+            [symbol, numeric]
+          end
+
+          def symbolic_to_numeric_value(node)
+            symbol = node.value
+            numeric = ::Rack::Utils::SYMBOL_TO_STATUS_CODE[symbol]
+
+            [numeric, symbol]
+          end
+
+          def message_for_autocorrect(node)
+            if style == :numeric
+              prefer, current = symbolic_to_numeric_value(node)
+              return [prefer, ":#{current}"]
+            end
+
+            numeric_to_symbolic_value(node)
+          end
+
+          def message_without_autocorrect
+            return %i[numeric symbolic] if style == :numeric
+
+            %i[symbolic numeric]
+          end
+        end
+      end
+    end
+  end
+end
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Master (Unreleased)
 
+* Add `RSpec/Rails/HttpStatus` cop to enforce consistent usage of the status format (numeric or symbol). ([@anthony-robin][], [@jojos003][])
+
 ## 1.22.2 (2018-02-01)
 
 * Fix error in `RSpec/DescribedClass` when working on an empty `describe` block. ([@bquorning][])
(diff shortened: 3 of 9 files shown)


lib/rubocop/cop/rspec/http_status.rb
```diff
@@ -0,0 +1,82 @@
...
+      # @example `EnforcedStyle: numeric`
+      #   # bad
+      #   it { is_expected.to have_http_status :ok }
+      #   it { is_expected.to have_http_status :not_found }
+      #
+      #   # good
+      #   it { is_expected.to have_http_status 200 }
+      #   it { is_expected.to have_http_status 404 }
+      #
+      class HttpStatus < Cop
```
Darhazer: we should add `Rails` namespace, similar to Capybara and FactoryBot

lib/rubocop/cop/rspec/http_status.rb
```diff
@@ -0,0 +1,82 @@
+# frozen_string_literal: true
+
+require 'symbol_to_status_code'
+
+module RuboCop
+  module Cop
+    module RSpec
+      # This cop enforces consistent use of `symbol` or `numeric`
```
Darhazer: You need a single line first-sentense description. You may add more on the next line

lib/rubocop/cop/rspec/http_status.rb
```diff
@@ -0,0 +1,82 @@
...
+          http_status(node) do |value_type|
+            return if style == :numeric && value_type.int_type?
+            return if style == :symbol && value_type.sym_type?
+
+            add_offense(node)
+          end
+        end
+
+        def message(node)
+          if node.each_descendant.find(&:sym_type?)
```
Darhazer: this can be written as `if node.each_child_node(:sym_type).any?`

lib/rubocop/cop/rspec/http_status.rb
```diff
@@ -0,0 +1,82 @@
...
+          end
+        end
+
+        def autocorrect(node)
+          replacement = new_value(node.children[2])
+          return if replacement.nil?
+
+          lambda do |corrector|
+            replacement = "#{node.children[1]} #{replacement}"
+            corrector.replace(node.loc.expression, replacement)
```
Darhazer: if you use node.loc.selector, I guess you don't need to build the replacement with the node.children[1], but just use the new_value

lib/rubocop/cop/rspec/http_status.rb
```diff
@@ -0,0 +1,82 @@
...
+              'to describe http status code.'.freeze
+
+        def_node_matcher :http_status, <<-PATTERN
+          (send _ :have_http_status ${int sym})
+        PATTERN
+
+        def on_send(node)
+          http_status(node) do |value_type|
+            return if style == :numeric && value_type.int_type?
+            return if style == :symbol && value_type.sym_type?
```
Darhazer: what if the symbol is of a more generic type, not mapping to single numeric value?

lib/rubocop/cop/rspec/http_status.rb
```diff
@@ -0,0 +1,82 @@
...
+
+            add_offense(node)
+          end
+        end
+
+        def message(node)
+          if node.each_descendant.find(&:sym_type?)
+            format(MSG, prefer: 'numeric', current: 'symbol')
```
Darhazer: Maybe we can add the specific symbol and numeric used? Like "Prefer :ok over 200 to describe http status code"

spec/rubocop/cop/rspec/http_status_spec.rb
```diff
@@ -0,0 +1,55 @@
...
+    end
+
+    it 'does not register an offense when using numeric value' do
+      expect_no_offenses(<<-RUBY)
+        it { is_expected.to have_http_status 200 }
+      RUBY
+    end
+
+    include_examples 'autocorrect',
+                     'it { is_expected.to have_http_status :ok }',
```
Darhazer: Please check if this will work as expected if parenthesis are used
`it { is_expected.to have_http_status(:ok) }'`

lib/rubocop/cop/rspec/http_status.rb
```diff
@@ -0,0 +1,82 @@
...
+      #   it { is_expected.to have_http_status 404 }
+      #
+      class HttpStatus < Cop
+        include ConfigurableEnforcedStyle
+
+        MSG = 'Prefer `%<prefer>s` over `%<current>s` '\
+              'to describe http status code.'.freeze
+
+        def_node_matcher :http_status, <<-PATTERN
+          (send _ :have_http_status ${int sym})
```
Darhazer: I think we are interested only in `(send nil? :have_http_status ${int sym})` status here, e.g. only on implicit receiver

lib/symbol_to_status_code.rb
```diff
@@ -0,0 +1,62 @@
+# Inspired from
+# http://www.rubydoc.info/github/rack/rack/master/Rack/Utils#HTTP_STATUS_CODES-constant
+module SymbolToStatusCode
```
bquorning: I’m not too crazy about having this module as part of rubocop-rspec – it doesn’t really belong here…

Would it be possible to hook into Rails or Rack to get this mapping, or would that be too messy? I’m thinking something like
```rb
begin
  require 'rack/utils'
  AUTOCORRECTS = true
rescue LoadError
  AUTOCORRECTS = false
end
```

spec/rubocop/cop/rspec/http_status_spec.rb
```diff
@@ -0,0 +1,55 @@
...
+RSpec.describe RuboCop::Cop::RSpec::HttpStatus, :config do
+  subject(:cop) { described_class.new(config) }
+
+  context 'when EnforcedStyle is `symbol`' do
+    let(:cop_config) { { 'EnforcedStyle' => 'symbol' } }
+
+    it 'registers an offense when using numeric value' do
+      expect_offense(<<-RUBY)
+        it { is_expected.to have_http_status 200 }
+                            ^^^^^^^^^^^^^^^^^^^^ Prefer `symbol` over `numeric` to describe http status code.
```
bquorning: It would be nice if we could have the `^^^`’s just under the `200` value. I haven’t entirely figured out how to control this highlighting myself, so it’s not a blocker for me :-)

config/default.yml
```diff
@@ -77,6 +77,15 @@ RSpec/DescribeSymbol:
   Enabled: true
   StyleGuide: http://www.rubydoc.info/gems/rubocop-rspec/RuboCop/Cop/RSpec/DescribeSymbol
 
+RSpec/HttpStatus:
+  Description: This cop enforces consistent use of `symbol` or `numeric` value to describe http status.
```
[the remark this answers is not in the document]
bquorning: Also, please check https://circleci.com/gh/backus/rubocop-rspec/296 (or `rake confirm_config`) to see why the build is failing.

lib/rubocop/cop/rspec/http_status.rb
bquorning: That was mentioned in #544 too, and I agree. I think that the “general response codes” `:error`, `:success`, `:missing`, and `:redirect` should still be allowed, even if the preferred style is `:numeric`.

bquorning:
Thank you for your first PR on rubocop-rails, @anthony-robin.

Don’t be intimidated by the *four failing builds* – they’re all failing for the same reason: a too long line in the YAML config. You can see the error locally by running `rake confirm_config`.

spec/rubocop/cop/rspec/http_status_spec.rb
Darhazer: Adding the offense to the `:selector` should do the trick

anthony-robin:
Thanks for your review !

I want to apologies for my naive approach on this cop. As I didn’t want to add `rack` as dependency of the gem, I recreated the hash without thinking I could require it from the rails dependency :(

I will update my code according to your recommendation this week !
Thanks 👍

bquorning:
> I want to apologies for my naive approach on this cop.

No need to apologise.

> As I didn’t want to add `rack` as dependency of the gem, I recreated the status hash without thinking I could require it from the rails dependency

Just to be clear, this gem doesn’t have a dependency on either of `rack` or `rails`. So I think enabling/disabling autocorrect by trying `require 'rack' ; rescue LoadError` would be a reasonable approach.

anthony-robin:
I made some changes according to your comments :)

However, I have some hard time figuring out how works the pattern in the `def_node_matcher` method:
- How can I set the selector to be the status code instead of the method name ? (I had a look at several other cops but the syntax doesn’t make sense to me)
- How can I handle the parenthesis in the pattern ?

I added more tests (that are still failing for now) until I understand how works the AST pattern.
Also, is it fine to add `rack` as a *development* dependency only ? I don’t know how I can make specs to pass without specifying it in the gemspec …

Thanks !

lib/rubocop/cop/rspec/rails/http_status.rb
```diff
@@ -0,0 +1,121 @@
+# frozen_string_literal: true
+
+begin
+  require 'rack/utils'
+  AUTOCORRECTS = true
```
bquorning: Ah, there a gotcha here. We want the `AUTOCORRECTS` constant to be defined inside the `RuboCop::Cop::RSpec::Rails::HttpStatus` module and not in global namespace. Not that it’s a problem right now, but if another cop should do the same trick, they’d overwrite each other’s constants.

anthony-robin:
The `confirm_config` is failing. I'm not sure why the blank line is being removed at the top of my configuration cop [here](https://github.com/backus/rubocop-rspec/pull/546/files#diff-e93280b3b31a6438c533a5f3232340d8R371)

lib/rubocop/cop/rspec/rails/http_status.rb
```diff
@@ -0,0 +1,126 @@
...
+
+          MSG = 'Prefer `%<prefer>s` over `%<current>s` '\
+                'to describe HTTP status code.'.freeze
+          WHITELIST_STATUS = %i[error success missing redirect].freeze
+
+          def_node_matcher :http_status, <<-PATTERN
+            (send nil? :have_http_status ${int sym})
+          PATTERN
+
+          def on_send(node) # rubocop:disable Metrics/CyclomaticComplexity
```
anthony-robin: I haven't been able to avoid the `CyclomaticComplexity` offense without disabling the cop. Do you have any idea how I could improve my code ?

anthony-robin:
Feel free to refactor this cop as your wish after merging it :)

lib/rubocop/cop/rspec/rails/http_status.rb
bquorning: I’m not sure how to fix it. Will take a closer look later.

PR #547, opened 2018-02-09
Simplify namespace regex and glob patterns

bquorning:
We currently have the namespaces `Capybara` (capybara), `FactoryBot` (factory_bot), and `RSpec` (with no sub-folder).

With a simpler regex and more consistent glob patterns, I hope it will be simpler to figure out which code to change when adding a new namespace. It doesn’t happen often enough that I want to write actual documentation.

-----------------

* [ ] Added tests.

### DIFF
--- a/bin/build_config
+++ b/bin/build_config
@@ -8,8 +8,8 @@ require 'rubocop-rspec'
 require 'rubocop/rspec/description_extractor'
 require 'rubocop/rspec/config_formatter'
 
-cops = '{*.rb,capybara/*.rb,factory_bot/*.rb}'
-glob = File.join(__dir__, '..', 'lib', 'rubocop', 'cop', 'rspec', cops)
+glob = File.join(__dir__, '..', 'lib', 'rubocop', 'cop', 'rspec',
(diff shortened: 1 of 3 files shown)


bin/build_config
```diff
@@ -8,8 +8,8 @@ require 'rubocop-rspec'
 require 'rubocop/rspec/description_extractor'
 require 'rubocop/rspec/config_formatter'
 
-cops = '{*.rb,capybara/*.rb,factory_bot/*.rb}'
-glob = File.join(__dir__, '..', 'lib', 'rubocop', 'cop', 'rspec', cops)
+glob = File.join(__dir__, '..', 'lib', 'rubocop', 'cop', 'rspec',
```
dgollahon: I'm not sure where this would go, but it might be nice to lift this into a constant that can be shared by the spec file as well.

Nice changes either way though 👍

bquorning: I thought the exact same thing: it would nice to share, but where would it go?

bquorning:
Buildkite is failing for unrelated reasons.

PR #549, opened 2018-02-16
Add DescribeWithoutExamples cop

pocke:
The cop checks `describe` without examples.

```ruby
# bad - the `describe` tests nothing.
describe do
  let(:foo) { 'bar' }
end

# good - the `describe` has an example.
describe do
  let(:foo) { 'bar' }
  it do
    expect(foo).to eq 'bar'
  end
end
```

I have two concerns for the cop.

Firstly, naming. I named the cop "DescribeWithoutExamples", but the cop also checks `context` and `feature`. So probably `ExampleGroupWithoutExamples` is better. What do you think?

Secondly, the cop ignore `include_context`. For example:

```ruby
# This cop adds an offense for the code.
describe do
  incldue_context :foo
end
```

I believe it is a "bad" code. However shared context can include examples. So it may be a false positive.

```ruby
# Shared context can include an example!
shared_context :foo do
  it do
    expect(:foo).to eq(bar)
  end
end

# This cop adds an offense for the code, but it is a false positive in this case!
describe do
  incldue_context :foo
end
```

I think shared context should not have examples (in this case we should use shared examples instead). So the cop should ignore shared context.
But sometimes shared context has examples. For instance: https://github.com/gitlabhq/gitlabhq/blob/dfe9d49534a26e86c773d81a4f5b81165efdd506/spec/lib/banzai/filter/milestone_reference_filter_spec.rb#L297-L309
So, I guess the cop should have `AllowSharedContext` option, and disable it by default. I'll add the option in another pull-request.

BTW, I think rubocop-rspec should have a cop that checks `shared_context` with examples. For example:

```ruby
# bad - shared context should not have an example.
shared_context :foo do
  it do
    expect(foo).to eq bar
  end
end
```

-----------------

* [x] Added tests.


dgollahon:
> BTW, I think rubocop-rspec should have a cop that checks shared_context with examples.

Are you aware of the SharedContext cop (https://github.com/backus/rubocop-rspec/blob/master/lib/rubocop/cop/rspec/shared_context.rb)?

It looks like it's implemented to be "conservative" in the sense that it only flags you for using shared contexts if there is no context added, only examples. Mixed cases are allowed to pass. Might be interesting to have a more aggressive style as an opt-in. See: #293

dgollahon:
Also, regarding

> So probably ExampleGroupWithoutExamples is better.  

I'm inclined to agree.

pocke:
> Are you aware of the SharedContext cop (https://github.com/backus/rubocop-rspec/blob/master/lib/rubocop/cop/rspec/shared_context.rb)?

Oops, I didn't know the cop. Thank you!

> Might be interesting to have a more aggressive style as an opt-in.

Good idea :+1:

---

And I'll rename the cop.

Darhazer:
`RSpec/EmptyExampleGroup` already handles empty describes

dgollahon:
> RSpec/EmptyExampleGroup already handles empty describes

I was thinking we already had something like this, but I couldn't remember it... It looks incredibly similar to this one.

pocke:
> RSpec/EmptyExampleGroup already handles empty describes

Oops, I didn't know the cop, Thank you!
I'll close this pull-request.

PR #550, opened 2018-02-16
More descriptive `RSpec/VerifiedDouble` example

pocke:
I think the current examples of `RSpec/VerifiedDouble` are not sufficiently descriptive.
It is not easy to understand the `"Widget"` is a class name. So this pull-request changes the class names, add a stubbed method, and add an unnamed double.

-----------------

* [ ] Added tests.

### DIFF
--- a/lib/rubocop/cop/rspec/verified_doubles.rb
+++ b/lib/rubocop/cop/rspec/verified_doubles.rb
@@ -9,13 +9,18 @@ module RuboCop
       #
       # @example
       #   # bad
-      #   it '...' do
-      #     widget = double("Widget")
+      #   let(:foo) do
+      #     double(method_name: 'returned value')
+      #   end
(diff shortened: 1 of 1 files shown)


lib/rubocop/cop/rspec/verified_doubles.rb
```diff
@@ -9,13 +9,18 @@ module RSpec
       #
       # @example
       #   # bad
-      #   it '...' do
-      #     widget = double("Widget")
+      #   let(:foo) do
+      #     widget = double(method_name: 'returned value')
```
bquorning: The assignment (`widget =`) can be removed.

pocke: Oops, I forget it. I'll remove them soon!

PR #553, opened 2018-02-16
Refactoring of Rails/HttpStatus 

bquorning:
As mentioned in #546, I was never completely happy with the `Rails/HttpStatus` cop, so I finally got around to refactoring it.

By extracting all style specific logic into two classes (`SymbolicStyleChecker` and `NumericStyleChecker`), the configured `style` is only checked once, instead of every method having identical `if style == :foo ... else ...` statements.

I hope you can help me come up with better names for these “checker” classes. And if you have suggestions on how to avoid having `def_node_matcher` called *on the singleton class*, please let me know.

By the way don’t try to read the diff. Better to look separately at the cop [before](https://github.com/backus/rubocop-rspec/blob/219ef75583f82bcca636727afa1774ac1681e11f/lib/rubocop/cop/rspec/rails/http_status.rb) and [after](https://github.com/backus/rubocop-rspec/blob/rails-http-status-refactoring/lib/rubocop/cop/rspec/rails/http_status.rb).

cc @anthony-robin 

-----------------

* [ ] Added tests.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## Master (Unreleased)
 
-* Add `RSpec/Rails/HttpStatus` cop to enforce consistent usage of the status format (numeric or symbol). ([@anthony-robin][], [@jojos003][])
+* Add `RSpec/Rails/HttpStatus` cop to enforce consistent usage of the status format (numeric or symbolic). ([@anthony-robin][], [@jojos003][])
 
 ## 1.22.2 (2018-02-01)
 
--- a/config/default.yml
+++ b/config/default.yml
@@ -372,8 +372,8 @@ FactoryBot/DynamicAttributeDefinedStatically:
 Rails/HttpStatus:
   Description: Enforces use of symbolic or numeric value to describe HTTP status.
   Enabled: true
-  EnforcedStyle: symbol
+  EnforcedStyle: symbolic
   SupportedStyles:
   - numeric
-  - symbol
+  - symbolic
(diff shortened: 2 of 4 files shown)


anthony-robin:
That's so much cleaner with your refactoring ! Wish I had figured out something like that by myself :)

Darhazer:
> And if you have suggestions on how to avoid having def_node_matcher called on the singleton class, please let me know.

Match the send node in the cop as it was in the original, and extract the value used in the checker with `one(send_node.arguments)` - then check if the type of the argument is the offensive one.

One think that I don't like with those helper classes is the amount of duplication. On the other hand I find this approach very good for cops supporting different styles - we may refactor the rest and extract some common patterns in base checker classes.

bquorning:
> Match the send node in the cop as it was in the original, and extract the value used in the checker with `one(send_node.arguments)` - then check if the type of the argument is the offensive one.

Good suggestions. I had to change the `#autocorrect` method to initiate the checker on `node.parent`, but that’s ok I guess.

I ended up pushing another 3 commits, which I haven’t squashed yet – should make it easier for you to review.

bquorning:
Thank you. Squashing and merging.

anthony-robin:
@bquorning I would like to submit this same cop in the `rubocop` repository that check `render` and `redirect_to` methods. Would you allow me to use your refactored code for it ?

bquorning:
Yes absolutely, go ahead.

The license even says something about “without limitation the rights to use, copy, modify, …” :-)

PR #554, opened 2018-02-17
[NEW] Add `FactoryBot/StaticAttributeDefinedDynamically` cop

abrom:
Adds new FactoryBot cop to check that static values are defined as such. I'm no ruby AST expert so it'd be great to get some feedback on this.

@jonatas, I've based this on your original dynamic FactoryBot check so you've probably got thoughts about this?

-----------------

* [x] Added tests.

### DIFF
--- a/lib/rubocop/cop/rspec/factory_bot/static_attribute_defined_dynamically.rb
+++ b/lib/rubocop/cop/rspec/factory_bot/static_attribute_defined_dynamically.rb
@@ -1,0 +1,75 @@
+# frozen_string_literal: true
+
+module RuboCop
+  module Cop
+    module RSpec
+      module FactoryBot
+        # Prefer declaring static attribute values without a block.
+        #
+        # @see DynamicAttributeDefinedStatically
+        #
+        # @example
+        #   # bad
+        #   kind { :static }
+        #
+        #   # good
+        #   kind :static
+        #
+        #   # bad
+        #   comments_count { 0 }
+        #
+        #   # good
+        #   comments_count 0
+        #
+        #   # bad
+        #   type { User::MAGIC }
+        #
+        #   # good
+        #   type User::MAGIC
+        class StaticAttributeDefinedDynamically < Cop
+          MSG = 'Do not use a block to set a static value ' \
+                'to an attribute.'.freeze
+
+          def_node_matcher :block_value_matcher, <<-PATTERN
+            (block (send nil? _) _ $...)
+          PATTERN
+
+          def_node_search :factory_attributes, <<-PATTERN
+            (block (send nil? { :factory :trait } ...) _ { (begin $...) $(send ...) $(block ...) } )
+          PATTERN
+
+          def on_block(node)
+            factory_attributes(node).to_a.flatten.each do |attribute|
+              values = block_value_matcher(attribute)
+              next if values.to_a.none? { |v| static?(v) }
+              add_offense(attribute, location: :expression)
+            end
+          end
+
+          def autocorrect(node)
+            lambda do |corrector|
+              corrector.replace(
+                node.loc.expression,
+                autocorrected_source(node)
+              )
+            end
+          end
+
+          private
+
+          def static?(node)
+            node.recursive_literal? || node.const_type?
+          end
+
+          def autocorrected_source(node)
+            if node.body.hash_type?
+              "#{node.send_node.source}(#{node.body.source})"
+            else
+              "#{node.send_node.source} #{node.body.source}"
+            end
+          end
+        end
+      end
+    end
+  end
+end
--- a/spec/rubocop/cop/rspec/factory_bot/static_attribute_defined_dynamically_spec.rb
+++ b/spec/rubocop/cop/rspec/factory_bot/static_attribute_defined_dynamically_spec.rb
@@ -1,0 +1,102 @@
+# frozen_string_literal: true
+
+# rubocop:disable Metrics/LineLength
+RSpec.describe RuboCop::Cop::RSpec::FactoryBot::StaticAttributeDefinedDynamically do
+  # rubocop:enable Metrics/LineLength
+
+  subject(:cop) { described_class.new(config) }
+
+  let(:config) { RuboCop::Config.new }
+
+  %w[FactoryBot FactoryGirl].each do |factory_bot|
+    context "when using #{factory_bot}" do
+      it 'registers an offense for offending code' do
+        expect_offense(<<-RUBY)
+          #{factory_bot}.define do
+            factory :post do
+              kind { :static }
+              ^^^^^^^^^^^^^^^^ Do not use a block to set a static value to an attribute.
+              comments_count { 0 }
+              ^^^^^^^^^^^^^^^^^^^^ Do not use a block to set a static value to an attribute.
+              type { User::MAGIC }
+              ^^^^^^^^^^^^^^^^^^^^ Do not use a block to set a static value to an attribute.
+              description { nil }
+              ^^^^^^^^^^^^^^^^^^^ Do not use a block to set a static value to an attribute.
+              recent_statuses { [:published, :draft] }
+              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ Do not use a block to set a static value to an attribute.
+              meta_tags { { foo: 1 } }
+              ^^^^^^^^^^^^^^^^^^^^^^^^ Do not use a block to set a static value to an attribute.
+            end
+          end
+        RUBY
+      end
+
+      it 'registers an offense in a trait' do
+        expect_offense(<<-RUBY)
+          #{factory_bot}.define do
+            factory :post do
+              title "Something"
+              trait :something_else do
+                title { "Something Else" }
+                ^^^^^^^^^^^^^^^^^^^^^^^^^^ Do not use a block to set a static value to an attribute.
+              end
+            end
+          end
+        RUBY
+      end
+
+      it 'accepts valid factory definitions' do
+        expect_no_offenses(<<-RUBY)
+          #{factory_bot}.define do
+            factory :post do
+              trait :something_else do
+                title "Something Else"
+              end
+              title "Something"
+              comments_count 0
+              description { FFaker::Lorem.paragraph(10) }
+              tag Tag::MAGIC
+              recent_updates { [Time.current] }
+              meta_tags { { first_like: Time.current } }
+            end
+          end
+        RUBY
+      end
+
+      it 'does not add offense if out of factory girl block' do
+        expect_no_offenses(<<-RUBY)
+          kind { :static }
+          comments_count { 0 }
+          type { User::MAGIC }
+          description { nil }
+        RUBY
+      end
+
+      bad = <<-RUBY
+        #{factory_bot}.define do
+          factory :post do
+            comments_count { 0 }
+            type { User::MAGIC }
+            description { nil }
+            recent_statuses { [:published, :draft] }
+            meta_tags { { foo: 1 } }
+          end
+        end
+      RUBY
+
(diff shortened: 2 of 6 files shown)


abrom:
If you're happy with this change, I would also suggest removing the parts of the dynamic FactoryBot attribute cop documentation that refer to static attributes (as they aren't really applicable there. 

Let me know and I'll push the change

lib/rubocop/cop/rspec/factory_bot/static_attribute_defined_dynamically.rb
```diff
@@ -0,0 +1,60 @@
...
+        #   comments_count { 0 }
+        #
+        #   # good
+        #   comments_count 0
+        #
+        #   # bad
+        #   type { User::MAGIC }
+        #
+        #   # good
+        #   type User::MAGIC
```
jonatas: I'd add a cross-link between them in the documentation:
```
# @see DynamicAttributeDefinedStatically
```

Because probably who enable this will enable the other one 😉

lib/rubocop/cop/rspec/factory_bot/static_attribute_defined_dynamically.rb
```diff
@@ -0,0 +1,60 @@
...
+        class StaticAttributeDefinedDynamically < Cop
+          MSG = 'Dont use a block to set a static value to an attribute.'.freeze
+
+          def_node_matcher :block_ast_method, <<-PATTERN
+          (block (send nil? _) _ ($_ ...))
+          PATTERN
+
+          def_node_search :factory_attributes, <<-PATTERN
+          (block (send nil? { :factory :trait } ...) _ { (begin $...) $(send ...) $(block ...) } )
```
jonatas: How this `:trait` is being used? Because it's being skipped in line 39, right?

lib/rubocop/cop/rspec/factory_bot/static_attribute_defined_dynamically.rb
```diff
@@ -0,0 +1,60 @@
...
+        #   # bad
+        #   type { User::MAGIC }
+        #
+        #   # good
+        #   type User::MAGIC
+        class StaticAttributeDefinedDynamically < Cop
+          MSG = 'Dont use a block to set a static value to an attribute.'.freeze
+
+          def_node_matcher :block_ast_method, <<-PATTERN
+          (block (send nil? _) _ ($_ ...))
```
jonatas: Instead of `$_` you can be explicit that you don't want a `send` using negation `$!send`:

```
(block (send nil? _) _ ($!send ...))
```

I didn't test it locally but it should work 😄

lib/rubocop/cop/rspec/factory_bot/static_attribute_defined_dynamically.rb
```diff
@@ -0,0 +1,60 @@
...
+          PATTERN
+
+          def_node_search :factory_attributes, <<-PATTERN
+          (block (send nil? { :factory :trait } ...) _ { (begin $...) $(send ...) $(block ...) } )
+          PATTERN
+
+          def on_block(node)
+            return if node.method_name == :trait
+            factory_attributes(node).to_a.flatten.each do |attribute|
+              next if [nil, :send].include? block_ast_method(attribute)
```
jonatas: If you follow the approach above, you can start yielding the attributes:

```ruby
block_ast_method(attribute) do |attribute_value|
   add_offense(attribute_value, location: :expression)
end
```

lib/rubocop/cop/rspec/factory_bot/static_attribute_defined_dynamically.rb
abrom: Good question ;)

I'm new to working with the source parsing. This bulk of this was copied verbatim from the `FactoryBot/DynamicAttributeDefinedStatically` cop

Having said that, it would appear that L39 is the one that should be removed (similarly for the same line in the dynamic attribute cop!)

lib/rubocop/cop/rspec/factory_bot/static_attribute_defined_dynamically.rb
abrom: Yes, works great! I removed the matcher as we shouldn't need it anymore 👍

lib/rubocop/cop/rspec/factory_bot/static_attribute_defined_dynamically.rb
abrom: I changed it to just a `next` operator instead. Reads a little better IMO:

69ca2dac1e1c65fa6dc2290c7de52167951e7815

abrom:
Dang.. I've just realised that #555 also applies to this change.

It currently detects that `foo { [Time.current] }` should *not* use a block. Hmm, thoughts?

jonatas:
Maybe we can start checking the array content as well?

```
(block
  (send nil :foo)
  (args)
  (array
    (send
      (const nil :Time) :current)))
```

We can enhance the expression to include it:

```ruby
def_node_matcher :block_ast_method?, <<-PATTERN
  (block (send nil? _) _ {(!send ...) (array (!send ...))})
PATTERN
```

abrom:
Ok, I believe the above discussed issues have been addressed. Specs added to test a block value with array/hash containing dynamic code _isn't_ identified as an offence.

Thanks for the suggestion @Darhazer !

lib/rubocop/cop/rspec/factory_bot/static_attribute_defined_dynamically.rb
```diff
@@ -0,0 +1,65 @@
...
+        #   # bad
+        #   type { User::MAGIC }
+        #
+        #   # good
+        #   type User::MAGIC
+        class StaticAttributeDefinedDynamically < Cop
+          MSG = 'Dont use a block to set a static value to an attribute.'.freeze
+
+          def_node_matcher :block_value_matcher, <<-PATTERN
+          (block (send nil? _) _ $...)
```
Darhazer: 2 spaces indentation

lib/rubocop/cop/rspec/factory_bot/static_attribute_defined_dynamically.rb
```diff
@@ -0,0 +1,65 @@
...
+          PATTERN
+
+          def_node_search :factory_attributes, <<-PATTERN
+          (block (send nil? { :factory :trait } ...) _ { (begin $...) $(send ...) $(block ...) } )
```
Darhazer: 2 spaces indentation

lib/rubocop/cop/rspec/factory_bot/static_attribute_defined_dynamically.rb
```diff
@@ -0,0 +1,65 @@
...
+          PATTERN
+
+          def on_block(node)
+            factory_attributes(node).to_a.flatten.each do |attribute|
+              values = block_value_matcher(attribute)
+              next if values.nil? ||
```
Darhazer: encapsulate this in a method. Are you sure you should check for `none?` here

lib/rubocop/cop/rspec/factory_bot/static_attribute_defined_dynamically.rb
```diff
@@ -0,0 +1,65 @@
...
+                    v.recursive_literal? || v.const_type?
+                  end
+              add_offense(attribute, location: :expression)
+            end
+          end
+
+          def autocorrect(node)
+            lambda do |corrector|
+              attribute = node.descendants.first
+              block_contents = node.descendants[2]
```
Darhazer: isn't this just `node.body` ?

spec/rubocop/cop/rspec/factory_bot/static_attribute_defined_dynamically_spec.rb
```diff
@@ -0,0 +1,94 @@
...
+  let(:config) { RuboCop::Config.new }
+
+  %w[FactoryBot FactoryGirl].each do |factory_bot|
+    context "when using #{factory_bot}" do
+      it 'registers an offense for offending code' do
+        expect_offense(<<-RUBY)
+          #{factory_bot}.define do
+            factory :post do
+              kind { :static }
+              ^^^^^^^^^^^^^^^^ Dont use a block to set a static value to an attribute.
```
Darhazer: the message should be `Don't` or `Do not`

lib/rubocop/cop/rspec/factory_bot/static_attribute_defined_dynamically.rb
abrom: Yes, we want to add an offence if there are any 'static' nodes matched (ie next if none).

Will change to `next if values.to_a.none? { |v| static?(v) }` and encapsulate the static checks into a method

lib/rubocop/cop/rspec/factory_bot/static_attribute_defined_dynamically.rb
```diff
@@ -0,0 +1,67 @@
...
+              next if values.to_a.none? { |v| static?(v) }
+              add_offense(attribute, location: :expression)
+            end
+          end
+
+          def autocorrect(node)
+            lambda do |corrector|
+              corrector.replace(
+                node.loc.expression,
+                "#{node.descendants.first.source} #{node.body.source}"
```
Darhazer: I guess `node.descendants.first` is actually `node.send_node`

lib/rubocop/cop/rspec/factory_bot/static_attribute_defined_dynamically.rb
```diff
@@ -0,0 +1,75 @@
...
+          end
+
+          private
+
+          def static?(node)
+            node.recursive_literal? || node.const_type?
+          end
+
+          def autocorrected_source(node)
+            if node.body.hash_type?
```
abrom: Note, I realised I hadn't added spec tests for auto-correcting a static hash in a block. In doing so I found a bug where the resulting value wasn't encapsulated with brackets leading to a ruby interpreter complaint. This change handles that case

spec/rubocop/cop/rspec/factory_bot/static_attribute_defined_dynamically_spec.rb
```diff
@@ -0,0 +1,102 @@
...
+      RUBY
+
+      corrected = <<-RUBY
+        #{factory_bot}.define do
+          factory :post do
+            comments_count 0
+            type User::MAGIC
+            description nil
+            recent_statuses [:published, :draft]
+            meta_tags({ foo: 1 })
```
abrom: And this tests it

lib/rubocop/cop/rspec/factory_bot/static_attribute_defined_dynamically.rb
abrom: Sounds good 👍

Darhazer:
On a side note, it seems that StaticAttributeDefinedDynamically and DynamicAttributeDefinedStatically are both parts on the same cop - it checks for the correct usage of blocks in the factory attributes. @jonatas what do you think?

abrom:
They're similar (inverses really), but there is very little overlap in common code. All of the matchers are different, as is the auto-correction.  Happy to have a go at it, but it feels like it'd end up being pretty messy. Would you think named something like `FactoryBot/AttributeDefinition` ?

Darhazer:
I'm not sure whether they should be part of the same cop, so I'm just bringing up the question.
I'm :ok: with merging as it is, and moving the discussion whether they should be combined in a single cop to an issue

jonatas:
I was also thinking about it. They are related to the same topic. The proper usage of attribute values.

But, with the current implementation, we can disable or enable them independently. I'd keep as it's. It's easier to maintain, they're already in the same department and we have a cross-link between the cops.

PR #556, opened 2018-02-19
[FIX #555] FactoryBot/DynamicAttributeDefinedStatically handle wrapped dynamic values

### LINKED ISSUE #555: FactoryBot/DynamicAttributeDefinedStatically does handle dynamic values in arrays/hashes
Neither of the following cases are detected as being dynamic factory girl values
```
foo [Time.current]
foo(bar: Time.current)
```

I tried having a go at fixing it by iterating over the attribute values but it ends up marking the trait tests as an offence (if a child has an offence):

```
def_node_search :dynamic_calls, <<-PATTERN
(send ...)
PATTERN

def_node_search :attribute_values, <<-PATTERN
(send nil? _ $...)
PATTERN

def_node_search :factory_attributes, <<-PATTERN
(block (send nil? {:factory :trait} ...) _ { (begin $...) $(send ...) } )
PATTERN

def on_block(node)
  return if node.method_name == :trait
  factory_attributes(node).to_a.flatten.each do |attribute|
    attribute_values(attribute).to_a.flatten.each do |value|
      next if dynamic_calls(value).to_a.none?
      add_offense(attribute, location: :expression)
    end
  end
end
```

fails the trait spec with:

```
        factory :post do
          title "Something"
          trait :published do
+         ^^^^^^^^^^^^^^^^^^^ Use a block to set a dynamic value to an attribute.
            published_at 1.day.from_now
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^ Use a block to set a dynamic value to an attribute.
          end
```

I suspect the fix might be to check if the `attribute` root node is a trait, and if so, get the child and add the offence to that? Thoughts?

abrom:
There are some edge cases relating to dynamic attributes not being detected when inside an array or hash, as per below:

```
# bad
recent_status_updates [1.hour.ago]
meta_tags(first_payment: 5.minutes.ago)

# good
recent_status_updates { [1.hour.ago] }
meta_tags { { first_payment: 5.minutes.ago } }
```

This change detects these edge cases and handles them appropriately.

Fixes #555 

-----------------

* [x] Added tests.

### DIFF
--- a/lib/rubocop/cop/rspec/factory_bot/dynamic_attribute_defined_statically.rb
+++ b/lib/rubocop/cop/rspec/factory_bot/dynamic_attribute_defined_statically.rb
@@ -30,42 +30,56 @@ module RuboCop
         class DynamicAttributeDefinedStatically < Cop
           MSG = 'Use a block to set a dynamic value to an attribute.'.freeze
 
-          def_node_matcher :dynamic_defined_statically?, <<-PATTERN
-          (send nil? _ (send ... ))
+          def_node_matcher :value_matcher, <<-PATTERN
+            (send nil? _ $...)
           PATTERN
 
           def_node_search :factory_attributes, <<-PATTERN
-          (block (send nil? {:factory :trait} ...) _ { (begin $...) $(send ...) } )
+            (block (send nil? {:factory :trait} ...) _ { (begin $...) $(send ...) } )
           PATTERN
 
           def on_block(node)
-            return if node.method_name == :trait
             factory_attributes(node).to_a.flatten.each do |attribute|
-              if dynamic_defined_statically?(attribute)
-                add_offense(attribute, location: :expression)
-              end
+              next if value_matcher(attribute).to_a.all? { |v| static?(v) }
+              add_offense(attribute, location: :expression)
             end
           end
 
           def autocorrect(node)
-            if method_uses_parens?(node.location)
-              autocorrect_replacing_parens(node)
-            else
+            if !method_uses_parens?(node.location)
               autocorrect_without_parens(node)
+            elsif value_hash_without_braces?(node.descendants.first)
+              autocorrect_hash_without_braces(node)
+            else
+              autocorrect_replacing_parens(node)
             end
           end
 
           private
+
+          def static?(node)
+            node.recursive_literal? || node.const_type?
+          end
+
+          def value_hash_without_braces?(node)
+            node.hash_type? && !node.braces?
+          end
 
           def method_uses_parens?(location)
             return false unless location.begin && location.end
             location.begin.source == '(' && location.end.source == ')'
           end
 
-          def autocorrect_replacing_parens(node)
+          def autocorrect_hash_without_braces(node)
+            autocorrect_replacing_parens(node, ' { { ', ' } }')
+          end
+
+          def autocorrect_replacing_parens(node,
+                                           start_token = ' { ',
+                                           end_token = ' }')
             lambda do |corrector|
-              corrector.replace(node.location.begin, ' { ')
-              corrector.replace(node.location.end, ' }')
+              corrector.replace(node.location.begin, start_token)
+              corrector.replace(node.location.end, end_token)
             end
           end
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 * Add `RSpec/Rails/HttpStatus` cop to enforce consistent usage of the status format (numeric or symbolic). ([@anthony-robin][], [@jojos003][])
 * Fix false negative in `RSpec/ReturnFromStub` when a constant is being returned by the stub. ([@Darhazer][])
+* Fix `FactoryBot/DynamicAttributeDefinedStatically` to handle dynamic attributes inside arrays/hashes. ([@abrom][])
 
 ## 1.22.2 (2018-02-01)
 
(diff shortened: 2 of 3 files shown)


lib/rubocop/cop/rspec/factory_bot/dynamic_attribute_defined_statically.rb
```diff
@@ -41,15 +41,20 @@ class DynamicAttributeDefinedStatically < Cop
...
+              next if values.nil?
+              next if values.all? { |v| v.recursive_literal? || v.const_type? }
+              add_offense(attribute, location: :expression)
             end
           end
 
           def autocorrect(node)
             if method_uses_parens?(node.location)
-              autocorrect_replacing_parens(node)
+              if value_is_inferred_hash?(node.descendants.first)
```
abrom: Check for `attribute(foo: bar)` => `attribute { { foo: bar } }`

abrom:
Following on from your suggestion in #554 @jonatas, this change should handle the various edge cases. Your suggestion would work, although I believe it would get caught up if there were additional levels of arrays? `foo [[Time.now]]` ??

lib/rubocop/cop/rspec/factory_bot/dynamic_attribute_defined_statically.rb
```diff
@@ -41,15 +41,20 @@ class DynamicAttributeDefinedStatically < Cop
           def on_block(node)
             return if node.method_name == :trait
             factory_attributes(node).to_a.flatten.each do |attribute|
-              if dynamic_defined_statically?(attribute)
-                add_offense(attribute, location: :expression)
-              end
+              values = value_matcher attribute
```
Darhazer: for consistency, `value_matcher(attribute)`

lib/rubocop/cop/rspec/factory_bot/dynamic_attribute_defined_statically.rb
```diff
@@ -41,15 +41,20 @@ class DynamicAttributeDefinedStatically < Cop
...
+              next if values.nil?
```
Darhazer: when the value is either nil or array, I prefer just to cast it to `to_a`
`next if values.to_a.all?`
also I would add a helper method `static?` which encapsulates recursive_literal? || const_type? check

lib/rubocop/cop/rspec/factory_bot/dynamic_attribute_defined_statically.rb
```diff
@@ -62,10 +67,20 @@ def method_uses_parens?(location)
             location.begin.source == '(' && location.end.source == ')'
           end
 
-          def autocorrect_replacing_parens(node)
+          def value_is_inferred_hash?(value)
+            value.type == :hash && value.location.begin.nil?
```
Darhazer: `value.hash_type? && value.location.begin.nil?` but what is an inferred hash anyway?
maybe you want to check for `value.hash_type? && value.braces?`

lib/rubocop/cop/rspec/factory_bot/dynamic_attribute_defined_statically.rb
```diff
@@ -62,10 +67,20 @@ def method_uses_parens?(location)
...
+          end
+
+          def autocorrect_inferred_hash(node)
+            autocorrect_replacing_parens node, ' { { ', ' } }'
```
Darhazer: `autocorrect_replacing_parens(node, ' { { ', ' } }')`

lib/rubocop/cop/rspec/factory_bot/dynamic_attribute_defined_statically.rb
```diff
@@ -30,8 +30,8 @@ module FactoryBot
         class DynamicAttributeDefinedStatically < Cop
           MSG = 'Use a block to set a dynamic value to an attribute.'.freeze
 
-          def_node_matcher :dynamic_defined_statically?, <<-PATTERN
-          (send nil? _ (send ... ))
+          def_node_matcher :value_matcher, <<-PATTERN
+          (send nil? _ $...)
```
Darhazer: this should be indented with 2 spaces

lib/rubocop/cop/rspec/factory_bot/dynamic_attribute_defined_statically.rb
abrom: Great, thanks! Still learning the node helpers 👍 Although I'm thinking you mean `value.hash_type? && !value.braces?` ?

By inferred hash I was attempting to describe a hash defined without braces (ie as a function parameter).

`foo(bar: 'baz')` vs `foo({ bar: 'baz' })`

The issue being, if the hash is defined without braces, we need to give it some when correcting the result and putting it into a block

abrom:
Regarding the extra tests, surely can. Although the accepts constants check is already there 

https://github.com/backus/rubocop-rspec/pull/556/files#diff-6fcbfafcbb24f7fa5e8cbbcfa6941499R58

`tag Tag::MAGIC`

lib/rubocop/cop/rspec/factory_bot/dynamic_attribute_defined_statically.rb
abrom: Changing the method name to `value_hash_without_braces?` to avoid confusion

Darhazer:
Can you please rebase so I can merge this?

abrom:
Done 👍

jonatas:
@abrom I don't think people will use such syntax.

We can create a [recursive build in search](https://rubocop.readthedocs.io/en/latest/node_pattern/#composing-complex-expressions-with-multiple-matchers) to check it and also cover hashes. But it's like a very rare case.

Darhazer:
Sorry @abrom can I ask you to rebase one more time, as there was a problem with the build and I was unable to merge

abrom:
No probs, done 👍

PR #558, opened 2018-02-22
Add readthedocs.io style documentation

Darhazer:
I'm trying to build a read-the-docs version similar to Rubocop, using Rubocop's own rake tasks for this.
Here are the files for the documentation. I've already setup a readthedocs project, but there's no documentation for it to fetch.
Few notes:
* ~~I have some issues with Yard not wanting to pick up the docs for all departments when I filter the departments~~
* ~~We have conflict with the Rails departments and I didn't figure out yet how to generate documentation only for RSpec/Rails~~
* ~~Once the issues above are resolved, the rake tasks can be pushed as wel and publishing documentation - automatedl~~
* The PR is published at http://rubocop-rspec.readthedocs.io/en/read-the-docs/

-----------------

* [ ] Added tests.
* [ ] The build (`bundle exec rake`) is passing.


bquorning:
Could you please split into a commit with your code, and a commit with the generated files?

Darhazer:
Done. The code of the rake tasks have the minimum required modifications over the Rubocop's one so it prints only the RSpec cops

mkdocs.yml
```diff
@@ -0,0 +1,13 @@
+site_name: "RSpec-specific analysis for your projects, as an extension to RuboCop."
+repo_url: https://github.com/backus/rubocop-rspec
+docs_dir: manual
+pages:
+- Home: index.md
+- Installation: installation.md
+- Usage: usage.md
+- Cops: cops.md
+- Cops Documentation:
```
Darhazer: oops this should be updated as well, and perhaps extracted into it's own commit

Darhazer:
Rebased and updated

bquorning:
I am not 100% sure about how readthedocs.io works. Do they fetch the `manual/` folder from our repository when we notify them (webhook), or will they fetch all source and run the rake task themselves?

If we need to keep the manual up to date, perhaps the rake task should be part of our release process which I imagine would need to be formalized :-)

Darhazer:
It works with a web hook. We can add the task to the default rake build and we can keep the readthedocs synced with master (I think RuboCop does exactly that) and not with the latest release

bquorning:
Yeah, add to the default rake task and perhaps update the Readme and the pull request template. Is anything else needed?

Darhazer:
@bquorning added to the default task and to the contributing noets.
We have to add a hook to https://readthedocs.org/api/v2/webhook/rubocop-rspec/28346/ ([set up documentation](https://docs.readthedocs.io/en/latest/webhooks.html))

bquorning:
I have set up the webhook – please let me know if it works or not.

.circleci/config.yml
```diff
@@ -13,13 +13,13 @@ rubocop: &rubocop
     - run: rake internal_investigation
 
 jobs:
-  confirm_config:
+  confirm_config_and_documentation:
```
bquorning: ❤️

Darhazer:
webhook works :beers:

.circleci/config.yml
```diff
@@ -111,28 +111,28 @@ workflows:
   version: 2
   build:
     jobs:
-      - confirm_config
+      - confirm_config_and_documentation
 
-      # Use `requires: [confirm_config]` to trick Circle CI into starting the slow jruby job early.
+      # Use `requires: [confirm_config_and_documentation]` to trick Circle CI into starting the slow jruby job early.
```
bquorning: I noticed how this build can be incredibly slow now, e.g. https://circleci.com/gh/rubocop-rspec/rubocop-rspec/1982

Perhaps we should use `required: [code-climate]` instead for this “trick”?

bquorning: Ah, I found the issue. See comment on the rake task.

tasks/cops_documentation.rake
```diff
@@ -0,0 +1,298 @@
...
+      .join("\n")
+  end
+
+  def assert_manual_synchronized
+    # Do not print diff and yield whether exit code was zero
+    sh('git diff --quiet manual') do |outcome, _|
+      return if outcome
+
+      # Output diff before raising error
+      sh('git diff manual')
```
bquorning: The default pager for `git diff` is `less`. And `less` doesn’t quit, which then causes the CI to hang.

E.g. `sh('PAGER=cat git diff manual')` should fix it.

Darhazer: I think we don't need this assert, since we have such check in `confirm_documentation` as well

bquorning: True, that is exactly what `rake confirm_documentation` does.

.circleci/config.yml
bquorning: > Perhaps we should use required: [code-climate] instead for this “trick”?

Just to be clear: Please disregard that suggestion now. :-)

bquorning:
I have removed `ci/circleci: confirm_config` as a required status on the repository. After merging, I will enable `ci/circleci: confirm_config_and_documentation` as a new required status.

PR #561, opened 2018-02-23
Bugfix for custom HTTP status codes

### LINKED ISSUE #560: Rails/HttpStatus custom HTTP status
rubocop : 0.52.1
rubocop-rspec : 1.23.0

Hello the new Cop `Rails/HttpStatus` is relying on `::Rack::Utils::SYMBOL_TO_STATUS_CODE` to check if we're using a symbol.
`::Rack::Utils::SYMBOL_TO_STATUS_CODE` only has symbols for the following http status :
```
[100, 101, 102, 200, 201, 202, 203, 204, 205, 206, 207, 208, 226, 300, 301, 302, 303, 304, 305, 307, 308, 400, 401, 402, 403, 404, 405, 406, 407, 408, 409, 410, 411, 412, 413, 414, 415, 416, 417, 421, 422, 423, 424, 426, 428, 429, 431, 451, 500, 501, 502, 503, 504, 505, 506, 507, 508, 510, 511]
```
If we're using custom HTTP status (like `550`), then we get an offense.

bquorning:
Fixes #560.

When Rails/HttpStatus is configured to use `symbolic` style, a custom HTTP status code (e.g. 550) that couldn't be found in Rack's `SYMBOL_TO_STATUS_CODE` hash would result in an error:

    ^^^ Prefer `nil` over `550` to describe HTTP status code.

This is solved by allowing integer values that are *not* present in `SYMBOL_TO_STATUS_CODE` when using the `symbolic` stile.

For this to work, I found that the `Rails/HttpStatus` cop couldn’t continue to be flexible in its dependency on Rack. But we cannot allow rubocop-rspec to have a hard dependency on Rack either. So I’m suggesting that we `rescue LoadError` either in the cop itself, or in the rspec_cops.rb file, right after `require_relative 'rspec/rails/http_status'`.

-----------------

* [x] Added tests.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Master (Unreleased)
 
+* The `Rails/HttpStatus` cop is unavailable if the `rack` gem cannot be loaded. ([@bquorning][])
+
 ## 1.23.0 (2018-02-23)
 
 * Add `RSpec/Rails/HttpStatus` cop to enforce consistent usage of the status format (numeric or symbolic). ([@anthony-robin][], [@jojos003][])
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## Master (Unreleased)
 
 * The `Rails/HttpStatus` cop is unavailable if the `rack` gem cannot be loaded. ([@bquorning][])
+* Fix `Rails/HttpStatus` not working with custom HTTP status codes. ([@bquorning][])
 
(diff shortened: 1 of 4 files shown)


Darhazer:
I was thinking the same - this false positive can be fixed only if rack is available (and generally whoever uses this cop will have it available). Not loading the cop at all, when there is no rack available, seems reasonable.

bquorning:
cc @anthony-robin

PR #562, opened 2018-02-24
[FIX] False positive in `FactoryBot/DynamicAttributeDefinedStatically`

abrom:
When defining a before/after callback with a symbol proc argument, the dynamic attribute cop will incorrectly identify it as an offense:

```
before(:create, &:initialize_something)
after(:create, &:rebuild_cache)
```

-----------------

* [x] Added tests.


abrom:
The opposite situation (callback with a block) doesn't match in the static attribute cop so the only change should be in the dynamic cop. Happy to add in something to the static attribute cop specs to prove the point if you think it prudent?

PR #566, opened 2018-03-05
ADD RSpec/SharedExamples cop

anthony-robin:
Enforces use of string to titleize shared examples.

@example
```ruby
  # bad
  it_behaves_like :foo_bar_baz
  it_should_behave_like :foo_bar_baz
  shared_examples :foo_bar_baz
  shared_examples_for :foo_bar_baz
  include_examples :foo_bar_baz

  # good
  it_behaves_like 'foo bar baz'
  it_should_behave_like 'foo bar baz'
  shared_examples 'foo bar baz'
  shared_examples_for 'foo bar baz'
  include_examples 'foo bar baz'
```

-----------------

* [x] Added tests.


bquorning:
Why do you prefer `string` over `symbolic` as the default config?

lib/rubocop/cop/rspec/shared_examples_style.rb
```diff
@@ -0,0 +1,136 @@
...
+          def offensive?
+            !node.sym_type?
+          end
+
+          def message
+            format(MSG, prefer: preferred_style, current: string)
+          end
+
+          def preferred_style
+            ":#{string.tr(' ', '_').tr('"', '')}"
```
bquorning: The substitution of `”` could be moved into `#string`?

lib/rubocop/cop/rspec/shared_examples_style.rb
```diff
@@ -0,0 +1,136 @@
...
+          def offensive?
+            !node.str_type?
+          end
+
+          def message
+            format(MSG, prefer: preferred_style, current: symbol.inspect)
+          end
+
+          def preferred_style
+            "'#{symbol.to_s.tr('_', ' ')}'"
```
bquorning: Perhaps extract a method that wraps a given string with single quotes?

bquorning:
If you’re wondering about the red builds… #567 should fix them.

bquorning:
Please rebase on master.

anthony-robin:
> Why do you prefer string over symbolic as the default config?

I chose string as I think it makes shared examples more readable than symbols but I can change it. 
Anyway, both configuration add offenses for already existing shared examples. Should I fix them in this PR or in a separate one according to the default style ?

bquorning:
I’m looking through [the documentation](https://relishapp.com/rspec/rspec-core/v/3-6/docs/example-groups/shared-examples), and [the source code](https://github.com/rspec/rspec-core), and in both places, only string arguments are used.

We should discuss the option of only allowing string arguments to these methods.

lib/rubocop/cop/rspec/shared_examples_style.rb
```diff
@@ -0,0 +1,136 @@
...
+      #   include_examples :foo_bar_baz
+      #
+      class SharedExamplesStyle < Cop
+        include ConfigurableEnforcedStyle
+
+        def_node_matcher :shared_examples_style, <<-PATTERN
+          (send nil? {
+              :it_behaves_like :it_should_behave_like
+              :shared_examples :shared_examples_for
+              :include_examples
```
bquorning: I think we may have some useful constants available in the `RuboCop::RSpec::Language` namespace.

bquorning:
> […] offenses for already existing shared examples. Should I fix them in this PR or in a separate one according to the default style ?

Yes please, fix them (in a separate commit) in this PR.

anthony-robin:
> We should discuss the option of only allowing string arguments to these methods.

I agree, only allowing strings would be better.

anthony-robin:
I just discovered that the pattern does not detect examples with several arguments such as:

```
include_examples :explicit_autocorrect, 'be_truthy', 'be_falsey'
```

What did I miss ?

lib/rubocop/cop/rspec/shared_examples_style.rb
```diff
@@ -0,0 +1,80 @@
...
+
+        def autocorrect(node)
+          lambda do |corrector|
+            checker = StringStyleChecker.new(node)
+            corrector.replace(node.loc.expression, checker.preferred_style)
+          end
+        end
+
+        # :nodoc:
+        class StringStyleChecker
```
anthony-robin: I kept the class as I think it makes things easier to understand but if you think it is not necessary anymore since there is only one style allowed I can change it.

bquorning: Perhaps the class name is too verbose now that there is just 1 style. How do you like `StyleChecker` or just `Checker`?

bquorning: I kind of like this logic encapsulated in a class, even if there is only one style.

lib/rubocop/cop/rspec/shared_examples_style.rb
```diff
@@ -0,0 +1,80 @@
...
+      #   include_examples :foo_bar_baz
+      #
+      #   # good
+      #   it_behaves_like 'foo bar baz'
+      #   it_should_behave_like 'foo bar baz'
+      #   shared_examples 'foo bar baz'
+      #   shared_examples_for 'foo bar baz'
+      #   include_examples 'foo bar baz'
+      #
+      class SharedExamplesStyle < Cop
```
bquorning: Perhaps the “Style” suffix is unnecessary?

lib/rubocop/cop/rspec/shared_examples_style.rb
```diff
@@ -0,0 +1,80 @@
...
+        def_node_matcher :shared_examples_style, <<-PATTERN
+          (send nil? {#{(SharedGroups::ALL + Includes::ALL).node_pattern}} $sym)
```
bquorning: We want to capture everything that isn’t a string, right? So I think it should be `$!str` instead of `$sym`.

lib/rubocop/cop/rspec/shared_examples_style.rb
```diff
@@ -0,0 +1,80 @@
...
+          MSG = 'Prefer %<prefer>s over `%<current>s` ' \
+                'to titleize shared examples.'.freeze
+
+          attr_reader :node
+          def initialize(node)
+            @node = node
+          end
+
+          def offensive?
+            !node.str_type?
```
bquorning: This method will always return true, since `node` is (currently) always a string node. You can remove the `#offensive?` method altogether.

bquorning:
> What did I miss ?

You missed a failing test case 😉

You can add `...` to the end of the node pattern to allow the extra arguments.

anthony-robin:
I refactored code according to your review. I let the `offensive?` method as I think it is necessary now.

Thanks !

lib/rubocop/cop/rspec/shared_examples.rb
```diff
@@ -0,0 +1,80 @@
...
+      #
+      class SharedExamples < Cop
+        def_node_matcher :shared_examples, <<-PATTERN
+          (send nil? {#{(SharedGroups::ALL + Includes::ALL).node_pattern}} $!str ...)
+        PATTERN
+
+        def on_send(node)
+          shared_examples(node) do |ast_node|
+            checker = Checker.new(ast_node)
+            return unless checker.offensive?
```
bquorning: This block is only executed when an `ast_node` is found that matches the criteria `!str`. This means `ast_node.str_type?` will always be false, and `checker.offensive?` will always be true.

anthony-robin: Right, I got it now !

bquorning:
So, I just checked this, and… the valid argument types to `shared_examples` etc. are `String`, `Symbol` and `Module`. 

https://github.com/rspec/rspec-core/blob/6fdcbfec431a918afedeabcae17532fb0b1b2c2d/lib/rspec/core/shared_example_group.rb#L206-L211

I managed to dig 7 years back to this pr: https://github.com/rspec/rspec-core/pull/408

anthony-robin:
> the valid argument types to `shared_examples` etc. are `String`, `Symbol` and `Module`.

Do you want me to handle all of this types as before my refactoring and allow users to configure the style they want or we still enforce usage of string ?

bquorning:
I believe `Module`/`Class` arguments cannot be easily substituted for a string/symbol equivalent. And vice versa. The best we can do it probably to *discourage* using symbols instead of insisting on *always* using strings.

So you were right earlier, to have `$sym` in your node matcher instead of the `$!str` that I suggested.

spec/rubocop/cop/rspec/shared_examples_spec.rb
```diff
@@ -0,0 +1,52 @@
...
+RSpec.describe RuboCop::Cop::RSpec::SharedExamples do
+  subject(:cop) { described_class.new }
+
+  it 'registers an offense when using symbolic title' do
+    expect_offense(<<-RUBY)
+      it_behaves_like :foo_bar_baz
+                      ^^^^^^^^^^^^ Prefer 'foo bar baz' over `:foo_bar_baz` to titleize shared examples.
+      it_should_behave_like :foo_bar_baz
+                            ^^^^^^^^^^^^ Prefer 'foo bar baz' over `:foo_bar_baz` to titleize shared examples.
+      shared_examples :foo_bar_baz
```
Darhazer: can you test those with blocks (shared_examples :foo_bar_baz do |param|)?

anthony-robin:
I refactored my code to only detect offenses for symbols titles.
I also added tests for shared examples with blocks and shared examples with modules/class name as title to ensure they don't throw offenses.

Thanks !

bquorning:
@Darhazer I would to head your opinion on this cop, too.

Darhazer:
I've never seen usage of symbol for shared_example title, but the fact that our own code base had such examples shows that it might be useful cop

PR #576, opened 2018-03-07
Increase code coverage

bquorning:
Looked at the code coverage report and found 2 methods that could be removed (1 in the specs), one spec example that needed to be added, and one spec that I have marked as `pending` instead of `skip` so that its code is exercised.

-----------------

* [x] Added tests.


bquorning:
\o/

PR #580, opened 2018-03-12
Make SharedExamples work also with explicit receiver

bquorning:
Shared examples may be defined using e.g. `RSpec.shared_examples`, which wasn’t handled by the cop.

RSpec/SharedExamples hasn’t been released yet, so I haven’t updated the changelog.

cc @anthony-robin 

-----------------

* [x] Added tests.

### DIFF
--- a/lib/rubocop/cop/rspec/shared_examples.rb
+++ b/lib/rubocop/cop/rspec/shared_examples.rb
@@ -22,7 +22,9 @@ module RuboCop
       #
       class SharedExamples < Cop
         def_node_matcher :shared_examples, <<-PATTERN
-          (send nil? {#{(SharedGroups::ALL + Includes::ALL).node_pattern}} $sym ...)
+          (send
+            {(const nil? :RSpec) nil?}
+            {#{(SharedGroups::ALL + Includes::ALL).node_pattern}} $sym ...)
         PATTERN
 
         def on_send(node)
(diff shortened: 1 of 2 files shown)


lib/rubocop/cop/rspec/shared_examples.rb
```diff
@@ -22,7 +22,9 @@ module RSpec
       #
       class SharedExamples < Cop
         def_node_matcher :shared_examples, <<-PATTERN
-          (send nil? {#{(SharedGroups::ALL + Includes::ALL).node_pattern}} $sym ...)
+          (send
+            {(const nil? :RSpec) nil?}
```
bquorning: `{(const nil? :RSpec) nil?}` is being used in a few places now. Is this something we should add to the language.rb file?

Darhazer: Definitely. And we generally have to check for both explicit and implicit receiver wherever appilcable

Darhazer: I think perhaps the `send_pattern` method in language may generate both explicit/implicit receiver and also provide capture for the arguments?

bquorning: It looks like `SelectorSet#send_pattern` will match something starting with `(send _` – basically setting no restriction on the receiver.

Darhazer: Unfortunately we can not use it here because we need the capture, but at least we may also use `_` for consistency and shorter code

bquorning: Ah, I get it now – you were considering using `(SharedGroups::ALL + Includes::ALL).send_pattern` directly as the node pattern. Yeah, that would have been neat.

Ok to merge this PR as is?

PR #581, opened 2018-03-12
Fix false positive when const is being used in a hash.

### LINKED ISSUE #577: FactoryBot/DynamicAttributeDefinedStatically false positive for "class"
Hi,

I believe there are false positive, with FactoryBot and STI you can override class, but you can't really expand it due to `class` being reserved word.

Here is an example:

`factory(:admin, class: Admin)`

Darhazer:
const is not a literal so recursive_literal? returns false when there is a const inside.
implemented own version of recursive_literal? that accounts const as a literal.

This fixes #577 

The `recursive_literal?` is now duplicated and we will need to extract it in a helper at some point.

-----------------

* [x] Added tests.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 * Add `RSpec/SharedExamples` cop to enforce consistent usage of string to titleize shared examples. ([@anthony-robin][])
 * Add `RSpec/Be` cop to enforce passing argument to the generic `be` matcher. ([@Darhazer][])
+* Fix false positives in `StaticAttributeDefinedDynamically` and `ReturnFromStub` when a const is used in an array or hash. ([@Darhazer][])
 
 ## 1.24.0 (2018-03-06)
 
--- a/lib/rubocop-rspec.rb
+++ b/lib/rubocop-rspec.rb
@@ -6,6 +6,7 @@ require 'rubocop'
 require_relative 'rubocop/rspec'
 require_relative 'rubocop/rspec/version'
 require_relative 'rubocop/rspec/inject'
+require_relative 'rubocop/rspec/node'
 require_relative 'rubocop/rspec/top_level_describe'
 require_relative 'rubocop/rspec/wording'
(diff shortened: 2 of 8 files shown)


bquorning:
Is this a bug in Rubocop? Should `:const` be added to `RuboCop::AST::Node::TRUTHY_LITERALS`?

It seems wrong to implement our own version of `recursive_literal?` that includes the `const_type?` if const isn’t really a literal… Perhaps I just don’t understand properly what we are trying to fix here. Can you help me?

bquorning:
> Maybe we have to add a recursive_literal_or_const? method to make the difference clear

Sounds like a good idea 👍

Darhazer:
Refactored in a separate commit

bquorning:
The three places where we use the new `#recursive_literal_or_const?` method have slight differences that annoy me a little bit:

`StaticAttributeDefinedDynamically#static?` has a nil check, while the single caller of `ReturnFromStub#dynamic?` has a nil check.

In `DynamicAttributeDefinedStatically` we have `.to_a.all?(&:recursive_literal_or_const?)`, while `StaticAttributeDefinedDynamically` has `to_a.none? { |v| static?(v) }`.

I wonder if these similarities are another refactoring waiting to happen?

Darhazer:
Updated

bquorning:
Good job on the cleanup of `static?` and `dynamic?` methods @Darhazer 👍 


I’m a little concerned about monkey patching `RuboCop::AST::Node`, and the maintenance problems that may come down the road.

But (as I see it) the alternative is to have our own concept of a node – i.e. wrapping each `RuboCop::AST::Node` instance in a `RuboCop::RSpec::Node` instance. Let’s not go there just yet.

We cannot push the `Node#recursive_literal_or_const?` method into RuboCop proper, right? It’s an rspec specific thing.

Darhazer:
It's a decision we made regarding specific cops. I don't see how it would fit in the RuboCop core.

PR #582, opened 2018-03-12
Enforce using  matcher with argument. Fixes #531

### LINKED ISSUE #531: Cop idea: don't allow be without arguments
I came across this 3 years old discussion about using `be` without arguments: 
https://github.com/rspec/rspec-expectations/issues/761

It will be nice to have a cop to detect such usage and it will actually help with the deprecation ;)

Darhazer:
See #531 for description.

-----------------

* [x] Added tests.


bquorning:
Good job. I think we might add `not_to` and `to_not` to the node pattern as well.

Darhazer:
Updated

PR #590, opened 2018-04-01
Don't check for feature methods outside spec. Fixes #474

### LINKED ISSUE #474: Capybara/FeatureMethods: False positive in FactoryGirls factory.
This file creates a false positive when running Rubocop. 
The autocorrect also will change the word `feature` for `describe`.

```ruby
# spec/factories/things.rb

FactoryGirl.define do
  factory :thing do
    name 'my_name'
    feature 'my_feature'
  end
end
```

Rubocop version: 0.5.0
Rubocop-rspec version: 1.17.1

Darhazer:
Since we check not only specs, but factories as well, some cops may give false positives for factories.
The FactoryBot cops check if they are inside a factory, while RSpec and Capybara cops assume to be in a spec. For the specific cop that checks for general usage of some terms that might legally exist in a factory as well, added explicit check for being inside a spec 

cc: @rspeicher 

-----------------

* [x] Added tests.

### DIFF
--- a/lib/rubocop/cop/rspec/capybara/feature_methods.rb
+++ b/lib/rubocop/cop/rspec/capybara/feature_methods.rb
@@ -45,6 +45,12 @@ module RuboCop
             feature:    :describe
           }.freeze
 
+          def_node_matcher :spec?, <<-PATTERN
+            (block
+              (send {(const nil? :RSpec) nil?} {:describe :feature} ...)
+            ...)
+          PATTERN
+
           def_node_matcher :feature_method, <<-PATTERN
             (block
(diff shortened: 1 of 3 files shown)


lib/rubocop/cop/rspec/capybara/feature_methods.rb
```diff
@@ -45,13 +45,21 @@ class FeatureMethods < Cop
             feature:    :describe
           }.freeze
 
+          def_node_matcher :spec?, <<-PATTERN
+            (block
+              $(send {(const nil? :RSpec) nil?} {:describe :feature} ...)
```
bquorning: No need for `$`.

Issue #538, opened 2018-01-19
Capybara/FeatureMethods conflicts with RSpec/DescribeClass in feature specs


mvz:
rubocop-rspec  should support rspec-rails' feature specs, however:

The following gives a `RSpec/DescribeClass` offense:
```
RSpec.describe 'user does foo', type: :feature do
  it 'succesfully foos the bar' do
    ...
  end
end
```

The following gives a Capybara/FeatureMethods offense:
```
RSpec.feature 'user does foo', type: :feature do
  scenario 'succesfully fooing the bar' do
    ...
  end
end
```

Darhazer:
@mvz would #593 work for you?

mvz:
@Darhazer I'm not sure. I was going to write a response based on reasoning, but I concluded I need to give this a try and see how it works out.

mvz:
@Darhazer I'm afread 75baab383d3d60d0972b3d8e30bc88891bb1c2b0 broke this cop completely for me, so it now never complains!

Darhazer:
Well, I guess you have to add the features spec folder to `Exclude` for the DescribeClass cop

mvz:
@Darhazer Yes, that's what I did. An alternative for users who want the FeatureMethods cop enabled is to disable DescribeClass for the features spec folder.

bquorning:
I am closing this issue, since the RSpec/Capybara/FeatureMethods cop was removed in #1876 / v3.0.0.

PR #593, opened 2018-04-06
Allow configuring enabled methods in .

Darhazer:
Some teams may prefer using some of the capybara methods (like `feature` and `scenario`) while enforce using `before` and `let` for the setup.

This will also fix cases like #538 by allowing the usage of `feature`.
(We should also allow ignoring DescribeClass bases on the type meta-data)

cc: @rspeicher 

-----------------

* [x] Added tests.


bquorning:
According to https://status.circleci.com/ workflows aren’t working at the moment. Apparently there’s no such thing as the perfect CI.

PR #596, opened 2018-04-09
Fix false positive in  when a sequence has a symbol proc argument.

tdeo:
A simple follow-up on https://github.com/rubocop-rspec/rubocop-rspec/pull/562 to handle the same case for sequences.

-----------------

* [x] Added tests.
* [x] Updated documentation (run `rake generate_cops_documentation`).

### DIFF
--- a/lib/rubocop/cop/rspec/factory_bot/dynamic_attribute_defined_statically.rb
+++ b/lib/rubocop/cop/rspec/factory_bot/dynamic_attribute_defined_statically.rb
@@ -31,13 +31,13 @@ module RuboCop
             (block (send nil? {:factory :trait} ...) _ { (begin $...) $(send ...) } )
           PATTERN
 
-          def_node_matcher :callback_with_symbol_proc?, <<-PATTERN
-            (send nil? {:before :after} sym (block_pass sym))
+          def_node_matcher :callback_or_sequence_with_symbol_proc?, <<-PATTERN
+            (send nil? {:before :after :sequence} sym (block_pass sym))
           PATTERN
 
           def on_block(node)
             factory_attributes(node).to_a.flatten.each do |attribute|
-              next if callback_with_symbol_proc?(attribute) ||
+              next if callback_or_sequence_with_symbol_proc?(attribute) ||
                   static?(attribute)
               add_offense(attribute, location: :expression)
             end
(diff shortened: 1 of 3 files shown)


lib/rubocop/cop/rspec/factory_bot/dynamic_attribute_defined_statically.rb
```diff
@@ -31,13 +31,13 @@ class DynamicAttributeDefinedStatically < Cop
             (block (send nil? {:factory :trait} ...) _ { (begin $...) $(send ...) } )
           PATTERN
 
-          def_node_matcher :callback_with_symbol_proc?, <<-PATTERN
-            (send nil? {:before :after} sym (block_pass sym))
+          def_node_matcher :callback_or_sequence_with_symbol_proc?, <<-PATTERN
+            (send nil? {:before :after :sequence} sym (block_pass sym))
```
Darhazer: maybe we need to skip block_pass on any arguments, not only the ones listed here?

lib/rubocop/cop/rspec/factory_bot/dynamic_attribute_defined_statically.rb
```diff
@@ -31,13 +31,13 @@ class DynamicAttributeDefinedStatically < Cop
             (block (send nil? {:factory :trait} ...) _ { (begin $...) $(send ...) } )
           PATTERN
 
-          def_node_matcher :callback_with_symbol_proc?, <<-PATTERN
-            (send nil? {:before :after} sym (block_pass sym))
+          def_node_matcher :callback_or_sequence_with_symbol_proc?, <<-PATTERN
```
bquorning: I am thinking this should perhaps be two different methods? I.e. adding a new `sequence_with_symbol_proc` node matcher. What do you think?

lib/rubocop/cop/rspec/factory_bot/dynamic_attribute_defined_statically.rb
Darhazer: That is exactly my point. Not sure if we should handle proc in a recursive manner, or can still use the recursive_literal_or_const helper

tdeo: I think we could handle proc only on the top-level only for this purpose (as we only need to handle arguments to the function), which would let us re-use `recursive_literal_or_const`

Maybe @bquorning has an opinion on that?

bquorning: So something like `proc?(foo) || recursive_literal_or_const?(foo)`? Sounds good.

Darhazer:
Thank you for your contribution, @tdeo

PR #597, opened 2018-04-09
Fix false positive in  cop.

### LINKED ISSUE #595: False positive in RSpec/Pending
We have a scope defined on an ActiveRecord model called 'pending'.

The RSpec/Pending cop is falsely flagging this as an issue on a line as below:

`subject { Vacancy.pending }`

Darhazer:
The `send_pattern` in language does not specify receiver. Usually the receiver must be either `nil?` or `RSpec`. When not-specified, a method named like some of the language methods would trigger false positive.

Fixes #595 

-----------------

* [x] Added tests.
* [ ] ~Updated documentation (run `rake generate_cops_documentation`).~

### DIFF
--- a/lib/rubocop/cop/rspec/pending.rb
+++ b/lib/rubocop/cop/rspec/pending.rb
@@ -43,8 +43,12 @@ module RuboCop
 
         def_node_matcher :pending_block?, PENDING_EXAMPLES.send_pattern
 
+        def_node_matcher :rspec?, '(send {(const nil? :RSpec) nil?} ...)'
+
         def on_send(node)
           return unless pending_block?(node) || skipped_from_metadata?(node)
+          return unless rspec?(node)
+
           add_offense(node, location: :expression)
         end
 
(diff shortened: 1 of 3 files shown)


lib/rubocop/cop/rspec/pending.rb
```diff
@@ -65,6 +67,10 @@ def skip_node?(node)
         def skip_symbol?(symbol_node)
           symbol_node == SKIP_SYMBOL || symbol_node == PENDING_SYMBOL
         end
+
+        def rspec?(node)
+          node.nil? || (node.const_type? && node.const_name == 'RSpec')
```
bquorning: Would it be possible to instead use our usual `{(const nil? :RSpec) nil?}` node matcher?

Darhazer: Possible, if match is performed on the send node, and not on the receiver (as `receiver` may return nil instead of ast::node). Updated.

Issue #599, opened 2018-04-18
How to “Report the versions of `rubocop-rspec`”


bquorning:
Our [contributing guidelines](https://github.com/rubocop-rspec/rubocop-rspec/blob/master/.github/CONTRIBUTING.md#issue-reporting) tell our users to

> Report the versions of `rubocop-rspec`, as well as the output of `rubocop -V`

But since there is no `rubocop-rspec -V` tool, how *do* people easily find the version of `rubocop-rspec`? Something like
```
bundle show rubocop-rspec | awk -F'/' '{ print $NF }'
```
or
```
bundle exec ruby -e 'require "rubocop-rspec"; puts RuboCop::RSpec::Version::STRING'
```
should work – should we add such a line to the guidelines? Or add a file in our `bin/` folder, for this one purpose?

dgollahon:
This is a really interesting point. I'm not sure what the right thing to do here is, but of your two proposed solutions I slightly prefer the latter since it happens within ruby / is an api we control.

If we ever get a proper plugin api for `rubocop`, maybe `rubocop -V` could list the versions of its plugins.

bquorning:
> If we ever get a proper plugin api for `rubocop`, maybe `rubocop -V` could list the versions of its plugins.

Yeah, that would be the preferred option.

pirj:
```
$ rubocop -V
  - rubocop-performance 1.9.2
  - rubocop-rspec 2.2.0
```

The issue has been resolved in RuboCop.

PR #600, opened 2018-04-18
Improve contribution guidelines

dgollahon:
I noticed our changelog link was not working in our CONTRIBUTING.md today, and from there I went on to fix a number of other minor issues.

- Fixed broken changelog link in CONTRIBUTING.md. When you are viewing CONTRIBUTING.md, you are already in /blob/ so the extra ../blob/master breaks the link. I decided to just use absolute urls for CONTRIBUTING.md and PULL_REQUEST_TEMPLATE.md to reduce confusion.
- Removed unnecessary line breaks (these were sometimes present after 80 characters, sometimes not--now they are consistently absent), whitespace, etc.
- Fixed a number of grammatical errors
- Improved misc. wording
- Added a reference to `bin/build_config` in PULL_REQUEST_TEMPLATE.md

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -1,41 +1,34 @@
 # Contributing
 
-If you discover issues, have ideas for improvements or new features,
-please report them to the [issue tracker][1] of the repository or
-submit a pull request. Please, try to follow these guidelines when you
-do so.
+If you encounter problems or have ideas for improvements or new features, please report them to the [issue tracker](https://github.com/rubocop-rspec/rubocop-rspec/issues) or submit a pull request. Please, try to follow these guidelines when you do so.
 
 ## Issue reporting
 
 * Check that the issue has not already been reported.
-* Check that the issue has not already been fixed in the latest code
-  (a.k.a. `master`).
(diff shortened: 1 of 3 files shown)


dgollahon:
I also wonder if we ought to have a default task that builds the config and documentation and simplifies our template like happened recently in `rubocop` proper: https://github.com/bbatsov/rubocop/pull/5176

bquorning:
> On a related note: I wonder if we should add an issue template as well

I was thinking the same thing.

bquorning:
Slightly related: #599.

Darhazer:
Execuring `rake` will generate documentation and build the config, you just need to add them to the commit

dgollahon:
> Execuring rake will generate documentation and build the config, you just need to add them to the commit

Ah! I should just reword a couple of these bullet points then.

dgollahon:
One more thought:

> Add an entry to `config/default.yml`. It's an ordered list, so be sure to insert at the appropriate place.

Feels burdensome and necessary. Any reason we can't just sort this automatically?

bquorning:
> Feels burdensome and necessary. Any reason we can't just sort this automatically?

Yeah, that should be possible.

PR #605, opened 2018-04-30
Add equality autocorrection option to PredicateMatcher

tiagotex:
I think it would be a good idea to use equality matcher when autocorrecting a PreficateMatcher to be explicit. 

In our codebase we don't want to allow `be` matchers in order to keep things explicit and consistent, so we would like to use this cop to automatically correct any use of `be_something` or `be true` to `eq true`.

I was not sure if we should drop the previous version (`be boolean`), so I added this as an config option.

---

* [x] Added tests.


bquorning:
Thank you for the PR @tiagotex, but I’m afraid I disagree with the proposed change.

[The documentation](https://relishapp.com/rspec/rspec-expectations/v/3-6/docs/built-in-matchers/equality-matchers) says

> [rspec-expectations] also ships with two matchers that have more of a DSL feel to them:
> ```rb
> expect(a).to be(b) # passes if a.equal?(b)
> expect(a).to eq(b) # passes if a == b
> ```

I would recommend always using the `be` matcher when comparing booleans (and `nil`s, and `Numeric`s)

PR #608, opened 2018-05-10
Allow comment to be placed after latest let

### LINKED ISSUE #603: RSpec/EmptyLineAfterFinalLet should ignore neighboring comments
I've had a few frustrating examples in my codebase where a comment line following a let (usually to re-enable a cop) has resulted in a violation I would not expect.

Example:
```ruby
RSpec.describe Foo do
  # rubocop:disable Metrics/LineLength
  let(:bar) do
    'hey this is a loooooooooooooooooooooooooooooooooooooooooooooooooooooooooong string'
  end
  # rubocop:enable Metrics/LineLength

  it 'has the expected bar' do
    expect(described_class.bar).to eql(bar)
  end
end
```

Result:
image

Solution:
My suggested approach would be to ignore comment lines immediately following the let as long as the next non-comment line is empty.

Darhazer:
Fixes #603

---

* [x] Added tests.

### DIFF
--- a/lib/rubocop/cop/rspec/empty_line_after_final_let.rb
+++ b/lib/rubocop/cop/rspec/empty_line_after_final_let.rb
@@ -17,9 +17,11 @@ module RuboCop
       #
       #   it { does_something }
       class EmptyLineAfterFinalLet < Cop
+        include RangeHelp
+
         MSG = 'Add an empty line after the last `let` block.'.freeze
 
-        def_node_matcher :let?, '(block $(send nil? {:let :let!} ...) args ...)'
+        def_node_matcher :let?, Helpers::ALL.block_pattern.freeze
 
         def on_block(node)
           return unless example_group_with_body?(node)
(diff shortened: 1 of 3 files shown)


spec/rubocop/cop/rspec/empty_line_after_final_let_spec.rb
```diff
@@ -136,6 +160,29 @@
...
   RUBY
 
+  include_examples 'autocorrect',
+                   bad_example,
+                   good_example
+
+  bad_example = <<-RUBY
+  RSpec.describe User do
+    let(:params) { foo }
+    # end of setup
```
bquorning: Since the code implements checks for multi-line comments, shouldn’t we let one of the specs have multi-line comments?

lib/rubocop/cop/rspec/empty_line_after_final_let.rb
```diff
@@ -29,24 +31,32 @@ def on_block(node)
...
         private
 
         def no_new_line_after(node)
           loc = last_node_loc(node)
+          line = loc.line
+          has_comment = false
+          while comment_line?(processed_source[line])
+            line += 1
+            has_comment = true
+          end
```
bquorning: I wonder if this cannot somehow be simplified. We want two things here, a boolean `has_comment` that is true if `loc.line` is a comment; and `line` which I guess is the last line of a multi-line comment. Could we have some code like this:
```rb
has_comment = comment_line?(processed_source[line])
last_commented_line = line + commented_lines_from(line) - 1
```
Also, bonus points if we can avoid `.offending_loc` taking a flag argument :-)

Darhazer:
Updated

dgollahon:
🎉 yay!

PR #609, opened 2018-05-13
Fix Capybara/FeatureMethods not working when there is require before the spec

Darhazer:
As reported in 75baab383d3d60d0972b3d8e30bc88891bb1c2b0 the cop doesn't work when describe/feature is not the first statement in the file

---

* [x] Added tests.


Darhazer:
Updated

PR #610, opened 2018-05-14
Improve language usage

Darhazer:
As a side effect, some cops will start detect `subject!` as well ;)

---

* [x] Added tests.
* [ ] ~Added an entry to the [changelog](https://github.com/rubocop-rspec/rubocop-rspec/blob/master/CHANGELOG.md) if the new code introduces user-observable changes.~

### DIFF
--- a/lib/rubocop/cop/rspec/let_before_examples.rb
+++ b/lib/rubocop/cop/rspec/let_before_examples.rb
@@ -35,7 +35,8 @@ module RuboCop
 
         MSG = 'Move `let` before the examples in the group.'.freeze
 
-        def_node_matcher :let?, '(block (send nil? {:let :let!} ...) ...)'
+        def_node_matcher :let?, Helpers::ALL.block_pattern
+
         def_node_matcher :example_or_group?, <<-PATTERN
           {
             #{(Examples::ALL + ExampleGroups::ALL).block_pattern}
--- a/lib/rubocop/rspec/example_group.rb
+++ b/lib/rubocop/rspec/example_group.rb
@@ -14,14 +14,12 @@ module RuboCop
         ExampleGroups::ALL + SharedGroups::ALL + Includes::ALL
       ).block_pattern
 
-      # @!method hook(node)
+      # @!method hook?(node)
       #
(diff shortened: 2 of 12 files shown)


lib/rubocop/cop/rspec/let_before_examples.rb
```diff
@@ -35,7 +35,8 @@ class LetBeforeExamples < Cop
 
         MSG = 'Move `let` before the examples in the group.'.freeze
 
-        def_node_matcher :let?, '(block (send nil? {:let :let!} ...) ...)'
+        def_node_matcher :let?, Helpers::ALL.block_pattern.freeze
```
bquorning: You probably shouldn’t `freeze` here.

lib/rubocop/rspec/example_group.rb
```diff
@@ -17,9 +17,7 @@ class ExampleGroup < Concept
       # @!method hook(node)
       #
       #   Detect if node is `before`, `after`, `around`
-      def_node_matcher :hook, <<-PATTERN
-        (block {$(send nil? #{Hooks::ALL.node_pattern_union} ...)} ...)
-      PATTERN
+      def_node_matcher :hook, Hooks::ALL.block_pattern
```
bquorning: Please also rename the matcher to `hook?`.

Darhazer:
Updated

lib/rubocop/rspec/example_group.rb
```diff
@@ -14,14 +14,12 @@ class ExampleGroup < Concept
...
+      # @!method hook?(node)
       #
       #   Detect if node is `before`, `after`, `around`
-      def_node_matcher :hook, <<-PATTERN
-        (block {$(send nil? #{Hooks::ALL.node_pattern_union} ...)} ...)
-      PATTERN
+      def_node_matcher :hook?, Hooks::ALL.block_pattern
 
-      def_node_matcher :subject, Subject::ALL.block_pattern
+      def_node_matcher :subject?, Subject::ALL.block_pattern
```
bquorning: ❤️

PR #617, opened 2018-05-23
Add RSpec/ReceiveCounts cop.

### LINKED ISSUE #615: Cop idea: detect 1.times / 2.times and suggest using once / twice
`exactly(1).times` -> `.once`
`exactly(2).times` -> `.twice`
`atLeast(1).times` -> `atLeast(:once)`
`atLeast(2).times` -> `atLeast(:twice)`
`atMost(1).times` -> `atMost(:once)`
`atMost(2).times` -> `atMost(:twice)`

zero is a special case and might fit in another cop:
`receive().exactly(0).times` -> `not_to receive()` (related to #607)

Darhazer:
Fixes #615 

---

* [x] Added tests.


lib/rubocop/cop/rspec/receive_counts.rb
```diff
@@ -0,0 +1,82 @@
...
+              replacement.sub('.', '') # remove the leading dot
+            )
+          end
+        end
+
+        private
+
+        def message_for(method, count)
+          alternative = matcher_for(method, count)
+          original = ".#{method}(#{count}).times"
```
bquorning: Shouldn’t this be available from the node source?

lib/rubocop/cop/rspec/receive_counts.rb
```diff
@@ -0,0 +1,82 @@
...
+          if method == :exactly
+            ".#{matcher}"
+          else
+            ".#{method}(:#{matcher})"
+          end
+        end
+
+        def range(node, offending_node)
+          range_between(
+            offending_node.loc.selector.begin_pos,
```
bquorning: If you use `offending_node.loc.selector.begin_pos - 1` here, you can avoid the `sub` call above, and have the offense messages include the dot when marking up with `^^^^^`.

lib/rubocop/cop/rspec/receive_counts.rb
Darhazer: The receiver of `exactly` is the `.receive` node (which may have other calls chained, like `with`). I'm trying to match `times` and only the selector part of `exactly`, which is a bit tricky, but makes autocorrection easy.
Having said that, it's possible to get substring from the original source, from the point of `times` call until the end

bquorning: Nah, this is probably fine.

Darhazer: Although it would have to operate on source of node, while method is from offending_node. And generally using slice won't be much prettier

Darhazer:
Updated

lib/rubocop/cop/rspec/receive_counts.rb
Darhazer: But I can get the source from the range, so after all it's not a bad solution

PR #618, opened 2018-05-25
Fix OverwritingSetup raising an error for unnamed subject

Darhazer:
57820a1c introduced an error for which there was no test case - handling unnamed subjects

---

* [x] Added tests.
* [ ] ~Added an entry to the [changelog](https://github.com/rubocop-rspec/rubocop-rspec/blob/master/CHANGELOG.md) if the new code introduces user-observable changes.~

### DIFF
--- a/lib/rubocop/cop/rspec/overwriting_setup.rb
+++ b/lib/rubocop/cop/rspec/overwriting_setup.rb
@@ -45,10 +45,15 @@ module RuboCop
         def find_duplicates(node)
           setup_expressions = Set.new
           node.each_child_node do |child|
-            setup?(child) do
-              name = one(child.send_node.arguments).value
-              yield child, name unless setup_expressions.add?(name)
-            end
+            next unless setup?(child)
+
+            name = if child.send_node.arguments?
+                     child.send_node.first_argument.value
+                   else
+                     :subject
+                   end
+
+            yield child, name unless setup_expressions.add?(name)
           end
         end
       end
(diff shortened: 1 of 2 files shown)


lib/rubocop/cop/rspec/overwriting_setup.rb
```diff
@@ -45,10 +45,11 @@ def on_block(node)
...
           setup_expressions = Set.new
           node.each_child_node do |child|
-            setup?(child) do
-              name = one(child.send_node.arguments).value
-              yield child, name unless setup_expressions.add?(name)
-            end
+            next unless setup?(child)
+            next unless child.send_node.arguments.any?
+
+            name = child.send_node.first_argument.value
```
dgollahon: Something else you could consider here is inferring that the name is `:subject` when you just have `subject { }` and then you could handle cases like:

```ruby
RSpec.describe MyClass do
  subject { foo }
  # ...
  let(:subject) { bar }
  # ...
end
```

Darhazer: There should be another cop disallowing `let(:subject)`

bquorning: Is `let(:let)` allowed?

dgollahon: @Darhazer:
> There should be another cop disallowing let(:subject)

Yeah, that seems like a good idea. I think you could reasonably do both though--I'm pretty sure `subject {}` is implemented with `let(:subject) {}` so this cop could rightfully record it as duplicate setup.

@bquorning:
I thought your remark was hilarious... and then i tried it. You _can_ do `let(:let)` and `let` continues to work correctly _and_ you can reference `let` in your tests and it will be the value you let with `let(:let) { value }` o_O

Darhazer:
Updated

Issue #272, opened 2017-01-02
Empty lines around examples, hooks, example groups


backus:
I'm often surprised when I see specs written like so

```ruby
describe 'foo' do
  subject { thing }
  let(:foo) { bar }
  let(:bar) { baz }
  before { some_setup }
  after { some_cleanup }
  it 'does x' do
    blah
  end
  it 'does y' do
    blah
  end
end

```

I think it is reasonable to say that this should **at least** be written like so:

```ruby
describe 'foo' do
  subject { thing }

  let(:foo) { bar }
  let(:bar) { baz }

  before { some_setup }
  after { some_cleanup }

  it 'does x' do
    blah
  end

  it 'does y' do
    blah
  end
end
```

This isn't without precedent too. If I wrote code like this:

```ruby
module Foo
  def bar
  end
  def baz
  end
  private
  def qux
  end
end
```

Rubocop would flag it

```
ex.rb:4:3: C: Style/EmptyLineBetweenDefs: Use empty lines between method definitions.
  def baz
  ^^^
ex.rb:6:3: C: Style/EmptyLinesAroundAccessModifier: Keep a blank line before and after private.
  private
  ^^^^^^^
```

backus:
@andyw8 @bquorning what do you think?

andyw8:
Should RuboCop's `EmptyLinesAroundBlockBody` already handle this? (can't check at the moment)

backus:
@andyw8 I don't think so because the code base that I most recently saw this in was in rubocop itself

andyw8:
Hmm, looks like the default is `no_empty_lines`?

https://github.com/bbatsov/rubocop/blob/master/config/default.yml#L438

backus:
@andyw8 interesting. I set that config and ran it on my example and it produced this

```ruby
describe 'foo' do

  subject { thing }
  let(:foo) { bar }
  let(:bar) { baz }
  before { some_setup }
  after { some_cleanup }
  it 'does x' do

    blah

  end
  it 'does y' do

    blah

  end

end
```

heh

andyw8:
Ah, looking at [the examples](http://www.rubydoc.info/gems/rubocop/RuboCop/Cop/Style/EmptyLinesAroundBlockBody) I misunderstood this.

I actually think a general "whitespace around multiline blocks" cop for RuboCop would be useful, but it might be easier to implement an RSpec-only cop here to begin with.

backus:
[...]

PR #619, opened 2018-05-25
Make LeadingSubject enforce subject to be above any examples or hooks

### LINKED ISSUE #616: RSpec/LeadingSubject should account for hooks
In the original issue for for adding the LeadingSubject cop (#52) the described goal was putting the subject first in a context/describe (which I think makes sense). I noticed when adding `rubocop-rspec` to a project today that it only sorts `subject`s ahead of `let`s. Personally, I'd like to enforce an entire ordering like `subject` -> `let` -> `before` -> `after`, but I think it'd at least be appropriate for RSpec/LeadingSubject to put subjects before `before` and `after`.

Darhazer:
Fixes #616 

---

* [x] Added tests.

### DIFF
--- a/config/default.yml
+++ b/config/default.yml
@@ -215,7 +215,7 @@ RSpec/ItBehavesLike:
   StyleGuide: http://www.rubydoc.info/gems/rubocop-rspec/RuboCop/Cop/RSpec/ItBehavesLike
 
 RSpec/LeadingSubject:
-  Description: Checks for `subject` definitions that come after `let` definitions.
+  Description: Enforce that subject is the first definition in the test.
   Enabled: true
   StyleGuide: http://www.rubydoc.info/gems/rubocop-rspec/RuboCop/Cop/RSpec/LeadingSubject
 
--- a/lib/rubocop/cop/rspec/leading_subject.rb
+++ b/lib/rubocop/cop/rspec/leading_subject.rb
@@ -3,64 +3,83 @@
 module RuboCop
   module Cop
     module RSpec
-      # Checks for `subject` definitions that come after `let` definitions.
+      # Enforce that subject is the first definition in the test.
       #
       # @example
       #   # bad
-      #   RSpec.describe User do
       #     let(:params) { blah }
       #     subject { described_class.new(params) }
       #
-      #     it 'is valid' do
-      #       expect(subject.valid?).to be(true)
-      #     end
-      #   end
+      #     before { do_something }
+      #     subject { described_class.new(params) }
+      #
+      #     it { expect_something }
+      #     subject { described_class.new(params) }
+      #     it { expect_something_else }
+      #
       #
       #   # good
-      #   RSpec.describe User do
       #     subject { described_class.new(params) }
-      #
       #     let(:params) { blah }
       #
-      #     it 'is valid' do
-      #       expect(subject.valid?).to be(true)
-      #     end
-      #   end
+      #   # good
+      #     subject { described_class.new(params) }
+      #     before { do_something }
+      #
+      #   # good
+      #     subject { described_class.new(params) }
+      #     it { expect_something }
+      #     it { expect_something_else }
+      #
       class LeadingSubject < Cop
         include RangeHelp
 
-        MSG = 'Declare `subject` above any other `let` declarations.'.freeze
+        MSG = 'Declare `subject` above any other `%<offending>s` ' \
+          'declarations.'.freeze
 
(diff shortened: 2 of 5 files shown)


config/default.yml
```diff
@@ -215,7 +215,7 @@ RSpec/ItBehavesLike:
   StyleGuide: http://www.rubydoc.info/gems/rubocop-rspec/RuboCop/Cop/RSpec/ItBehavesLike
 
 RSpec/LeadingSubject:
-  Description: Checks for `subject` definitions that come after `let` definitions.
+  Description: Enforce `subject` being the first definition in test.
```
dgollahon: I think "Enforce that `subject` is the first definition in the test." would read a bit better.

bquorning:
I was kind of hoping we could enforce the entire order of code, e.g.

1. subject
2. lets
3. hooks (before/after/around blocks)
4. examples
5. methods

Would that be possible, or is it out of scope for this PR / this cop?

lib/rubocop/cop/rspec/leading_subject.rb
```diff
@@ -3,64 +3,89 @@
...
         include RangeHelp
 
-        MSG = 'Declare `subject` above any other `let` declarations.'.freeze
+        MSG = 'Declare `subject` above any other `%<offending>s` ' \
+          'declarations.'.freeze
 
         def_node_matcher :subject?, Subject::ALL.block_pattern
+        def_node_matcher :let?, Helpers::ALL.block_pattern
+        def_node_matcher :hook?, Hooks::ALL.block_pattern
+        def_node_matcher :include?, <<-PATTERN
```
dgollahon: Hm. I don't know if this is just me or inconsistent with the principles here, but I frequently put `include_context` before my subject, something roughly like:

```ruby
RSpec.describe FooController do
  include_context 'controller setup'

  subject(:controller) { ... }

  # ...
end
```

I think this feels natural to me because it's analogous to the way I typically write classes

```ruby
class Foo
  include WhateverImInheriting
  
  def initialize
   # ...
  end

  # ...

end
```

where I always want to have my constructor at the "top" but inheritance comes before that.

I'm not sure if it's exactly the same, but it feels different than the other forms of setup like  lets and hooks. WDYT?

cc @bquorning

Darhazer: I'm fine with allowing include_context to be on the top. What about include_examples?

dgollahon: On the one hand, I think `include_examples` generally belongs below the subject. On the other hand `include_examples` is just an alias for `include_context` so it feels sort of weird to have one but not the other.

Darhazer: The semantics are different; that's why we enforce shared_context or shared_examples depending whether examples are defined in the group ;)

dgollahon:
@bquorning 

> I was kind of hoping we could enforce the entire order of code ...

I'm in favor of being able to do this, but it should either be multiple cops _or_ this cop should be renamed, as "LeadingSubject" doesn't really make sense when you mean something more like "organize all the parts". I suspect some people may have different preferences or prefer not to enforce some parts, so it may be reasonable to have it as multiple cops. Or maybe we can have a reasonable set of configuration options or we just give people all or nothing--I'm not totally convinced one way or another yet.

lib/rubocop/cop/rspec/leading_subject.rb
dgollahon: The expressed intent is different, yeah, but not the behavior. While I do like that we try to enforce the naming, we can't enforce the choice of `include_context` v.s. `include_examples` within a test because we can't tell if it is referring to examples or setup context or some fuzzy mix of both that maybe deserves a different term entirely. Since the proper choice of `include_*` is not enforced, it feels weird to base an additional rule off of that. I expect that this would cause some amount of frustration to people who don't use one or the other consistently, but I'm not positive.

My gut feeling is leaving it to helpers/hooks/examples is possibly a better starting place. Feel free to disagree though, I'm not firm on this.

Darhazer:
PR updated

Darhazer:
@bquorning 

> I was kind of hoping we could enforce the entire order of code ...

We already have LeadingSubject, LetBeforeExamples,ScatteredSetup,EmptyLineAfterSubject and EmptyLineAfterFinalLet. We need to add cops to check the position of hooks, empty lines around examples (#272), include_examples/context usage

PR #627, opened 2018-06-07
JRuby build image needs `make` installed

bquorning:
The `jaro_winkler` gem cannot be installed on jruby unless the command `make` is available. So I ~made a docker image~ added a `run` step that installs `make`, and for some reason I had to also bump the CircleCI bundle cache.

---

* [ ] Added tests.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -77,14 +77,15 @@ jobs:
       - image: circleci/jruby:9.1
     steps:
       - checkout
+      - run: sudo apt-get install -y make
       - run: bundle lock
       - restore_cache:
           keys:
-            - bundle-v1-{{ checksum "Gemfile.lock" }}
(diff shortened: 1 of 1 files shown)


Darhazer:
Can we add `-run: sudo apt install -y make` as a step in the jRuby build?

.circleci/config.yml
```diff
@@ -77,6 +77,7 @@ jobs:
       - image: circleci/jruby:9.1
     steps:
       - checkout
+      - run: apt-get install -y make
```
Darhazer: you need `sudo`

bquorning:
Thank you @Darhazer, yours was a much simpler solution than mine :-)

bquorning:
I’m making `jruby` a required build step again.

PR #631, opened 2018-06-11
Fix "new cop generator" rake task

bquorning:
The `RuboCop::Cop::Generator.initialize` API was changed in https://github.com/rubocop-hq/rubocop/commit/488c6dee600f4b23326fa862a66e8e16783 which was released with RuboCop v0.56.0.

---

* [ ] Added tests.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 * Add `RSpec/EmptyLineAfterHook` cop. ([@bquorning][])
 * Add `RSpec/EmptyLineAfterExampleGroup` cop to check that there is an empty line after example group blocks. ([@bquorning][])
 * Fix `RSpec/DescribeClass` crashing on `RSpec.describe` without arguments. ([@Darhazer][])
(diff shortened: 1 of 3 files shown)


bquorning:
Oh, I just realised we’d need to bump the dependency on Rubocop also…

Darhazer:
> Oh, I just realised we’d need to bump the dependency on Rubocop also…

Only for development :)

bquorning:
> Only for development :)

That is true. But one version requirement in the gemspec and another in the Gemfile, won’t that be confusing at some point?

bbatsov:
I think it will be. As we plan to make RuboCop upgrades very painless soon maybe it will make sense to just require whatever the latest RuboCop was around the time of the release?

dgollahon:
> But one version requirement in the gemspec and another in the Gemfile, won’t that be confusing at some point?

Yeah, I'd think so. We should definitely be careful not to release this way and I don't think there's an automated process to catch that class of mistake yet.

bquorning:
Done: I’m bumping the RuboCop requirement to `>= 0.56.0` in the gemspec.

PR #632, opened 2018-06-11
New cop: RSpec/EmptyLineAfterHook

bquorning:
This cop will verify that there is an empty line after hook blocks (`before`, `after`, `around`, etc.)

Ref: https://github.com/rubocop-hq/rubocop-rspec/issues/272

This cop is very much inspired by `EmptyLineAfterSubject` and perhaps we can extract some common parts to e.g. a module. We are also going to need an `EmptyLineAfterContext` cop soon, and perhaps others too?

---

* [x] Added tests.


lib/rubocop/cop/rspec/empty_line_after_hook.rb
```diff
@@ -0,0 +1,67 @@
...
+        def on_block(node)
+          return unless hook?(node)
+          return if node.equal?(node.parent.children.last)
+
+          return if next_line(node).blank?
+
+          add_offense(
+            node,
+            location: :expression,
+            message: format(MSG, hook: node.send_node.source)
```
Darhazer: you can use `node.method_name`

bquorning: Thank you, I didn’t know that.

bquorning:
Hmm, just thought of a missing feature: Two one-line hooks shouldn’t need to have a newline between them.

Darhazer:
> Hmm, just thought of a missing feature: Two one-line hooks shouldn’t need to have a newline between them.

That's my request for a cop that forces empty lines between examples - option for single-line examples to stay together ;)

bquorning:
The `EmptyLineAfterFinalLet` has that feature, perhaps it can be extracted from there?

dgollahon:
I believe this also has the problem https://github.com/rubocop-hq/rubocop-rspec/pull/608 fixed for the other cop.

> Hmm, just thought of a missing feature: Two one-line hooks shouldn’t need to have a newline between them.

 Interesting thought. In the cases where we have two of the same kind of hook, there's another cop that should complain about this (unless the user disables it). In the case where there are different kinds of setup, if they are one-liners i don't have a strong opinion and could live with both, though i think my gut reaction might be to put a line between `before` and `after`.

But how does this behave for something like this?

```ruby
before do
  do_setup!
end
after do
  do_teardown!
end

it 'works' do
end
```

Just skimming the cop, i think this won't raise an offense, right? I definitely think there should be space between different kinds of hooks though.

I think what I might want is something that just enforces spaces between every different category of things: subject(!), let(!), before, around, after, contexts, describes, its (and similar) in a unified cop (maybe with configuration for things like grouping before and after? or do we just need like a dozen of these cops?)

Darhazer:
Perhaps the extraction of common code for the EmptyLineAfter* cops would help to tread comments consistently

bquorning:
> (maybe with configuration for things like grouping before and after? or do we just need like a dozen of these cops?)

I guess it all depends on how granularly people want to be able to configure them. I am all for a simple setup, but having several, smaller cops allows for enabling/disabling them independently from each other.

bquorning:
```rb
before do
  do_setup!
end
after do
  do_teardown!
end

it 'works' do
end
```

will indeed be reported as an offense:

```
before do
  do_setup!
end
^^^ Add an empty line after `before` block.
after do
  do_teardown!
end

it 'works' do
end
```

dgollahon:
@bquorning Ahh, I misread this. 👍 So it unconditionally requires an empty line after every hook, right? I was imagining it as being after the last hook like our let cop works. I'm considerably happier with this now that I understand it. :)

I do really think we should fix the comment issue though. I think @Darhazer is right about extracting the logic if we can:

> Perhaps the extraction of common code for the EmptyLineAfter* cops would help to tread comments consistently

bquorning:
> Hmm, just thought of a missing feature: Two one-line hooks shouldn’t need to have a newline between them.

I just noticed that this issue still isn’t fixed… Any ideans?

Darhazer:
Should be configuration option IMO. If we decide on this, I'll be happy to make the cop for empty lines around examples as well

bquorning:
I don’t think is has been considered yet. Feel free to open a dedicated issue for it.

PR #634, opened 2018-06-12
New cop: RSpec/EmptyLineAfterExampleGroup

bquorning:
This cop will verify that there is an empty line after hook blocks (before, after, around, etc.)

Ref: #272

This cop is very much inspired by `EmptyLineAfterSubject` and `EmptyLineAfterHook` (#632) and perhaps we can extract some common parts to e.g. a module.

---

* [x] Added tests.


lib/rubocop/cop/rspec/empty_line_after_example_group.rb
```diff
@@ -0,0 +1,53 @@
...
+        def on_block(node)
+          return unless example_group(node)
+          return if node.parent && node.equal?(node.parent.children.last)
+
+          return if next_line(node).blank?
+
+          add_offense(
+            node,
+            location: node.loc.end,
+            message: format(MSG, example_group: node.send_node.method_name)
```
Darhazer: method_name works for block_nodes as well, so no need for `send_node`

lib/rubocop/cop/rspec/empty_line_after_example_group.rb
```diff
@@ -0,0 +1,53 @@
...
+module RuboCop
+  module Cop
+    module RSpec
+      # Checks if there is an empty line after example group blocks.
+      #
+      # @example
+      #   # bad
+      #   RSpec.describe Foo do
+      #   end
+      #   RSpec.describe Bar do
```
Darhazer: those seems as offenses for the multipledescribes cop, better add examples for few different describes in a single top-level describe ;)

Darhazer:
> This cop is very much inspired by EmptyLineAfterSubject and EmptyLineAfterHook (#632) and perhaps we can extract some common parts to e.g. a module.

Sure we can. I would extract the check if the next line is blank, as well whether the node is the one one in a block and perhaps the autocorrect
```
# common checks
return if node.parent && node.equal?(node.parent.children.last)

return if next_line(node).blank?
```

lib/rubocop/cop/rspec/empty_line_after_example_group.rb
bquorning: That is a very good point, actually. I’ll change here, and all the specs.

bquorning:
Fixed the issues mentioned in a separate commit for easier review. I’ll squash before merging :-)

Then, I’ll take a stab a extracting shared logic from the “empty line after …” cops.

PR #636, opened 2018-06-12
Use location#dot instead of fixed offset

Darhazer:
> Just a thought: How does this work if the dot is the last character on the preceding line?

Nice question. Turns out Rubocop have a method to give you the location of a dot.

---

* [ ] ~Added tests.~
* [ ] ~Added an entry to the [changelog](https://github.com/rubocop-hq/rubocop-rspec/blob/master/CHANGELOG.md) if the new code introduces user-observable changes.~

### DIFF
--- a/lib/rubocop/cop/rspec/receive_counts.rb
+++ b/lib/rubocop/cop/rspec/receive_counts.rb
@@ -76,7 +76,7 @@ module RuboCop
 
         def range(node, offending_node)
           range_between(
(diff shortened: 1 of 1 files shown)


bbatsov:
> Nice question. Turns out Rubocop have a method to give you the location of a dot.

Actually that's coming straight from the Parser gem. :-)

Darhazer:
> Is this the only location we missed this trick?

Didn't find any others

PR #641, opened 2018-06-13
Implement autocorrect for Capybara/CurrentPathExpectation

ypresto:
Capybara/CurrentPathExpectation did not have autocorrect so I implemented it!

Note that `match(variable)` is not supported due to ambiguous replacement:

- If `variable` returns string, it should be replaced with `have_current_path(/#{variable}/)`.
- If `variable` returns regexp, it should be replaced with `have_current_path(variable)`.

ref: #470

---

* [x] Added tests.

### DIFF
--- a/lib/rubocop/cop/rspec/capybara/current_path_expectation.rb
+++ b/lib/rubocop/cop/rspec/capybara/current_path_expectation.rb
@@ -30,10 +30,58 @@ module RuboCop
             (send nil? :expect (send {(send nil? :page) nil?} :current_path))
           PATTERN
 
+          # Supported matchers: eq(...) / match(/regexp/) / match('regexp')
+          def_node_matcher :as_is_matcher, <<-PATTERN
+            (send
+              #expectation_set_on_current_path ${:to :not_to :to_not}
+              ${(send nil? :eq ...) (send nil? :match (regexp ...))})
+          PATTERN
+
+          def_node_matcher :regexp_str_matcher, <<-PATTERN
+            (send
+              #expectation_set_on_current_path ${:to :not_to :to_not}
+              $(send nil? :match (str $_)))
+          PATTERN
+
           def on_send(node)
             expectation_set_on_current_path(node) do
               add_offense(node, location: :selector)
             end
+          end
+
+          def autocorrect(node)
+            lambda do |corrector|
+              return unless node.chained?
+
+              as_is_matcher(node.parent) do |to_sym, matcher_node|
+                rewrite_expectation(corrector, node, to_sym, matcher_node)
+              end
+
+              regexp_str_matcher(node.parent) do |to_sym, matcher_node, regexp|
+                rewrite_expectation(corrector, node, to_sym, matcher_node)
+                convert_regexp_str_to_literal(corrector, matcher_node, regexp)
+              end
+            end
+          end
+
+          private
+
+          def rewrite_expectation(corrector, node, to_symbol, matcher_node)
+            current_path_node = node.first_argument
+            corrector.replace(current_path_node.loc.expression, 'page')
+            corrector.replace(node.parent.loc.selector, 'to')
+            matcher_method = if to_symbol == :to
+                               'have_current_path'
+                             else
+                               'have_no_current_path'
+                             end
+            corrector.replace(matcher_node.loc.selector, matcher_method)
+          end
+
+          def convert_regexp_str_to_literal(corrector, matcher_node, regexp_str)
+            str_node = matcher_node.first_argument
+            regexp_expr = Regexp.new(regexp_str).inspect
+            corrector.replace(str_node.loc.expression, regexp_expr)
           end
         end
       end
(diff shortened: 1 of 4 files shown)


ypresto:
Possible bug: if current_path is overridden in current context (`let` or local variable), it can cause false-positive.

ypresto:
> What are the downsides of splitting the 1st matcher in two

Firstly I'm putting all the code into `def autocorrect` so much of duplication with two (large) matchers pattern. But it can be simpler now:

```rb
def autocorrect
  asis_matcher(node.parent) do |to_symbol, matcher_node|
    rewrite_expectation(corrector, node, to_symbol, matcher_node)
  end
  regexp_str_matcher(node.parent) do |to_symbol, matcher_node|
    rewrite_expectation(corrector, node, to_symbol, matcher_node)
    convert_regexp_str_to_literal(corrector, matcher_node)
end
```

ypresto:
Thank you for reviewing, finally fixed them! :)

lib/rubocop/cop/rspec/capybara/current_path_expectation.rb
```diff
@@ -30,11 +30,54 @@ class CurrentPathExpectation < Cop
...
+
           def on_send(node)
             expectation_set_on_current_path(node) do
               add_offense(node, location: :selector)
             end
           end
+
+          def autocorrect(node)
+            lambda do |corrector|
+              return unless node.chained?
```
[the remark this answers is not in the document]
ypresto: added test!

lib/rubocop/cop/rspec/capybara/current_path_expectation.rb
```diff
@@ -30,11 +30,57 @@ class CurrentPathExpectation < Cop
             (send nil? :expect (send {(send nil? :page) nil?} :current_path))
           PATTERN
 
+          # Supported matchers: eq(...) / match(/regexp/) / match('regexp')
+          def_node_matcher :as_is_matcher, <<-PATTERN
+            (send
+              (send nil? :expect (send {(send nil? :page) nil?} :current_path)) ${:to :not_to :to_not}
```
Darhazer: At least reuse the `expectation_set_on_current_path` - you can write 
```
#expectation_set_on_current_path ${:to :not_to :to_not}
```

lib/rubocop/cop/rspec/capybara/current_path_expectation.rb
```diff
@@ -30,11 +30,57 @@ class CurrentPathExpectation < Cop
...
+              ${(send nil? :eq ...) (send nil? :match (regexp ...))})
+          PATTERN
+
+          def_node_matcher :regexp_str_matcher, <<-PATTERN
+            (send
+              (send nil? :expect (send {(send nil? :page) nil?} :current_path)) ${:to :not_to :to_not}
```
Darhazer: reuse `expectation_set_on_current_path`

lib/rubocop/cop/rspec/capybara/current_path_expectation.rb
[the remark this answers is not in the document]
Darhazer: Add empty line after the return

lib/rubocop/cop/rspec/capybara/current_path_expectation.rb
```diff
@@ -30,11 +30,57 @@ class CurrentPathExpectation < Cop
...
             end
           end
+
+          def autocorrect(node)
+            lambda do |corrector|
+              return unless node.chained?
+              as_is_matcher(node.parent) do |to_sym, matcher_node|
+                rewrite_expectation(corrector, node, to_sym, matcher_node)
+              end
+              regexp_str_matcher(node.parent) do |to_sym, matcher_node, regexp|
```
Darhazer: maybe you can combine the 2 matchers in one and just check if the method_name of the matcher_node is `match` and argument is of regexp_type

Darhazer:
@ypresto would you have time to push this forward or do you prefer if I apply my own feedback on the code? It's a really nice feature and it would be great to include it in the next release

lib/rubocop/cop/rspec/capybara/current_path_expectation.rb
ypresto: Firstly they was combined, but it was split according to below comment:
https://github.com/rubocop-hq/rubocop-rspec/pull/641#pullrequestreview-129310147

ypresto:
updated PR!

Darhazer:
@bquorning maybe take another look?

lib/rubocop/cop/rspec/capybara/current_path_expectation.rb
bquorning: I guess my comment “that would reduce duplication in the matchers, but perhaps introduce extra duplication further down in the class…” turned out to be true :-)

Darhazer:
@ypresto please rebase :package:

ypresto:
Thank you all for reviewing, just rebased!

ypresto:
Thank you all for reviewing!
@Darhazer just rebased!

PR #640, opened 2018-06-13
Bump version to 1.27.0

bquorning:
Before submitting the PR make sure the following are checked:

* [x] Feature branch is up-to-date with `master` (if not - rebase it).
* [x] Squashed related commits together.
* [ ] Added tests.
* [x] Added an entry to the [changelog](https://github.com/rubocop-hq/rubocop-rspec/blob/master/CHANGELOG.md) if the new code introduces user-observable changes.
* [x] The build (`bundle exec rake`) passes (be sure to run this locally, since it may produce updated documentation that you will need to commit).

PR #642, opened 2018-06-14
Unify empty line cop logic

dgollahon:
Resolves https://github.com/rubocop-hq/rubocop-rspec/pull/632#issuecomment-396821220 and unblocks #640 by:

- Adding the FinalEndLocation mixin
  * This mixin DRYs up the logic for determining the true end location of a node, including when it contains a heredoc.
- Unifying the empty line cops
    * DRYs things up a bit.
    * Ensures each cop correctly handles heredocs.
    * Unifies highlighting format.
      - Note: It was very hard to find some sort of unified way to highlight
        the different possible node endings, so I just went with
        highlighting the lowest offending line starting right of any
        whitespace.
    * Ensures each cop allows comments immediately following the offending
      line, notably to provide support for rubocop:disable directives
      (especially disable/enable pairs that might surround the block).
---

* [x] Added tests.
* [ ] Added an entry to the [changelog](https://github.com/rubocop-hq/rubocop-rspec/blob/master/CHANGELOG.md) if the new code introduces user-observable changes. (note: changes are probably too subtle to merit documentation)

### DIFF
--- a/lib/rubocop/rspec/final_end_location.rb
+++ b/lib/rubocop/rspec/final_end_location.rb
@@ -1,0 +1,15 @@
+module RuboCop
+  module RSpec
+    # Helps find the true end location of nodes which might contain heredocs.
+    module FinalEndLocation
+      def final_end_location(start_node)
+        heredoc_endings =
+          start_node.each_node(:str, :dstr, :xstr)
+            .select(&:heredoc?)
+            .map { |node| node.loc.heredoc_end }
+
+        [start_node.loc.end, *heredoc_endings].max_by(&:line)
+      end
+    end
+  end
+end
--- a/spec/rubocop/cop/rspec/empty_line_after_final_let_spec.rb
+++ b/spec/rubocop/cop/rspec/empty_line_after_final_let_spec.rb
@@ -19,9 +19,9 @@ RSpec.describe RuboCop::Cop::RSpec::EmptyLineAfterFinalLet do
       RSpec.describe User do
         let(:a) { a }
         let!(:b) do
-        ^^^^^^^^^^^ Add an empty line after the last `let` block.
           b
         end
+        ^^^ Add an empty line after the last `let` block.
         it { expect(a).to eq(b) }
       end
     RUBY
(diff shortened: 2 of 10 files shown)


dgollahon:
As promised, @bquorning :) Turned out to be a little more involved than I thought, but I think this moves us in the right direction.

lib/rubocop/rspec/final_end_location.rb
```diff
@@ -0,0 +1,15 @@
+module RuboCop
+  module RSpec
+    # Helps find the true end location of nodes which might contain heredocs.
+    module FinalEndLocation
+      def final_end_location(start_node)
+        heredoc_endings =
+          start_node.each_node
```
Darhazer: You can use 
```
start_node.each_node(:str, :dstr, :xstr)
          .select(&:heredoc?)
          .map { |node| node.loc.heredoc_end }
```

spec/rubocop/cop/rspec/empty_line_after_final_let_spec.rb
```diff
@@ -19,9 +19,9 @@
       RSpec.describe User do
         let(:a) { a }
         let!(:b) do
-        ^^^^^^^^^^^ Add an empty line after the last `let` block.
           b
         end
+        ^^^ Add an empty line after the last `let` block.
```
bquorning: ❤️

spec/rubocop/cop/rspec/empty_line_after_final_let_spec.rb
```diff
@@ -185,6 +200,31 @@
...
   RUBY
 
+  include_examples 'autocorrect',
+                   bad_example,
+                   good_example
+
+  bad_example = <<-RUBY
+  RSpec.describe User do
+    let(:params) { <<-DOC }
+     I'm super annoying!
```
bquorning: The truth about heredocs.

bquorning: Actually, perhaps `parser` and `ast` gems could be changed to provide better support for heredocs and "line continuation string"…

dgollahon: > "line continuation string"

WDYM by that?

lib/rubocop/rspec/final_end_location.rb
dgollahon: Neat, good call!

spec/rubocop/cop/rspec/empty_line_after_final_let_spec.rb
dgollahon: Oh, do things like

```ruby
'hi i'm a string'\
'and still continuing'
```

produce weird locations too?

bquorning: Yup. I will open an issue on Rubocop one of these days.

PR #643, opened 2018-06-14
Fix alphabetic require spec

dgollahon:
- This project structure spec was effectively a no-op because the
  require statements have (presumably) moved since this was added. I
  have fixed the assertion and allowed for varying cop categories as
  well as added a quick sanity check. Having this might be overkill, but
  it least it works now.

---

* [ ] Added tests.


spec/project/project_requires_spec.rb
```diff
@@ -1,8 +1,19 @@
 RSpec.describe 'Project requires' do
+  # rubocop:disable RSpec/MultipleExpectations
```
bquorning: I think we can just use `:aggregate_failures` on the `it` block?

spec/project/project_requires_spec.rb
```diff
@@ -1,8 +1,19 @@
...
   it 'alphabetizes cop requires' do
-    source   = SpecHelper::ROOT.join('lib', 'rubocop-rspec.rb').read
-    requires = source.split("\n").grep(%r{rubocop/cop/rspec/[^(?:cop)]})
+    source   = SpecHelper::ROOT.join('lib', 'rubocop', 'cop', 'rspec_cops.rb')
+    captures = source.read.scan(%r{^(require_relative 'rspec/(.*?/)?(.*?)')$})
 
-    expect(requires.join("\n")).to eql(requires.sort.join("\n"))
+    # Sanity check in case these move again.
```
bquorning: I think this comment will provide less and less information with time. Perhaps remove “again” and describe “these” in words?

spec/project/project_requires_spec.rb
```diff
@@ -1,8 +1,19 @@
 RSpec.describe 'Project requires' do
+  # rubocop:disable RSpec/MultipleExpectations
   it 'alphabetizes cop requires' do
-    source   = SpecHelper::ROOT.join('lib', 'rubocop-rspec.rb').read
-    requires = source.split("\n").grep(%r{rubocop/cop/rspec/[^(?:cop)]})
+    source   = SpecHelper::ROOT.join('lib', 'rubocop', 'cop', 'rspec_cops.rb')
+    captures = source.read.scan(%r{^(require_relative 'rspec/(.*?/)?(.*?)')$})
```
bquorning: Perhaps using named captures will make the `sort_by` block a bit clearer?

spec/project/project_requires_spec.rb
```diff
@@ -1,8 +1,19 @@
...
 
-    expect(requires.join("\n")).to eql(requires.sort.join("\n"))
+    # Sanity check in case these move again.
+    expect(captures).not_to be_empty
+
+    require_statements = captures.map(&:first).join("\n")
+    sorted_require_statements =
+      captures.sort_by do |_require_statment, cop_category, name|
+        [cop_category || 'rspec', name]
+      end.map(&:first).join("\n")
```
bquorning: Do we need the `.join("\n")`s? Can’t we just compare the arrays?

spec/project/project_requires_spec.rb
dgollahon: Sure.

spec/project/project_requires_spec.rb
dgollahon: Hm. `scan` and `grep` both ignore named captures. I think doing a manual map and match for the named captures doesn't feel necessary to me.

spec/project/project_requires_spec.rb
dgollahon: I was wondering that at first too, but it produces a much nicer diff:
image

v.s.

image

PR #644, opened 2018-06-14
Remove duplicate `simplecov` gem specification

dgollahon:
- Removes warning bundler emits when installing. Since this is in the gemspec already, I can't see any reason to keep this around.

---

* [ ] Added tests.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -2,10 +2,6 @@ source 'https://rubygems.org'
 
 gemspec
 
-group :test do
-  gem 'simplecov', require: false
-end
-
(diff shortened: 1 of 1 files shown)


bquorning:
Actually, I would prefer moving all development dependencies from the gemspec into the Gemfile. In this particular case, the `require: false` is nice, since `simplecov` is used *only* on CI. But even that can be questioned, because what if someone wants to get a coverage report locally (I’ve done that a few times)…

Also related: `Bundler.setup(:default, :development)` in the Rakefile seems a bit weird.

PR #645, opened 2018-06-15
Add smoke tests

### LINKED ISSUE #388: "Common edge cases" shared examples
We should add some shared examples to our specs which include tests like 

```ruby
it 'does not encounter an error when handling an empty describe' do
  expect { inspect_source(cop, 'RSpec.describe(User) do end', 'a_spec.rb') }
    .not_to raise_error
end
```

and then in our spec helper we can include these shared examples for all specs in the `spec/rubocop/cop/rspec` directory.

This should help us avoid easy to overlook regressions like #385

dgollahon:
- Closes #388
- Adds a handful of rspec files that will stress our cops. These will be
  run with our normal tests (though the assertions are pointless) to
  audit whether the rspec is valid / close to valid. In the future I
  want to develop a corpus test from real ruby projects, but this is a
  good sanity check and includes a number of cases that have previously
  caused bugs.
- Shared example is automatically added to all cop specs based on our
  existing directory structure.
- Removed a few examples which just asserted no error (now covered by
  the smoke test).
- I jumped back 25 commits, cherry-picked the changes, and it caught at
  least two separate bugs that we've already fixed. I think this will
  definitely help us catch some things early.
- ~Also adds default file names to the various source inspection helpers.
  I was tearing my hair out for about an hour trying to figure out why
  in the heck my smoke tests were producing no smoke... only to very
  gradually remember that we filter based on filename and not realizing
  the default was `nil`.~
- Note: This will falsely inflate our coverage metrics a bit because it
  is more likely to arbitrarily exercise the various branches of new
  cops but assert nothing more than it didn't crash. We could account
  for this in a few ways (mutation testing, for instance, is unaffected
  by these changes) if we really wanted to, but I mainly just want
  for us maintainers to be aware that it can happen.
- Note: This is just a starting place for random RSpec usage, largely
  stream of consciousness, though i did mine our issues for cases that
  previously caused bugs. The file can grow with time when anyone
  thinks of something they could imagine raising an error.

---

* [x] Added tests.

### DIFF
--- a/spec/smoke_tests/weird_rspec_spec.rb
+++ b/spec/smoke_tests/weird_rspec_spec.rb
@@ -1,0 +1,233 @@
+RSpec.describe 'Weirdness' do
+  subject! { nil }
+  subject { nil }
+  subject(:foo) { nil }
+  subject!(:foo) { nil }
+
+  subject! (:foo) { |something| nil }
+  subject  :foo  do end
+
+  let(:foo) { |something| something }
+  let (:foo) { 1 }
+  let!   (:bar){}
+
+  let :a do end
+
+  let(:bar) { <<-HEREDOC }
+    What a pain.
+  HEREDOC
+
+  let(:bar) { <<-'HEREDOC' }
+    Even odder.
+  HEREDOC
+
+  let(:baz) do
+    <<-INSIDE
+      Hi. I'm in your lets.
+    INSIDE
+  end
+
+  let(:hi)      {}
+  let(:bye) do
+  end
+
+  let(:doop) { foo; 1 }
+
+  it {}
+  specify {}
+
+  it 'works', metadata: true do
+  end
+
+  describe {}
+  context {}
+
+  describe '#nothing' do
+  end
+
+  it 'is empty' do
+  end
+
+  it '' do end
+  describe do end
+  context do end
+  shared_examples 'a' do end
+
+  describe 'things' do
+    context 'with context' do
+    end
+  end
+
+  shared_examples 'weird rspec' do
+  end
+
+  shared_examples :something do
+  end
+
+  context 'test' do
+    include_examples 'weird rspec'
+    include_examples('weird rspec', serious: true) do
+      it_behaves_like :something
+    end
+  end
+
+  it_behaves_like :something
+  it_should_behave_like :something
+
+  it_behaves_like :something do
+    let(:foo) { 'bar' }
+  end
+
+  it_behaves_like(:something) do |arg, *args, &block|
+  end
+
+  before {}
+  context 'never run' do
+    around {}
+  end
+  after {}
+
+  before { <<-DOC }
+   Eh, what's up?
+  DOC
+
+  around { |test| test.run; <<-DOC }
+   Eh, what's up?
+  DOC
+
+  after { <<-DOC }
+   Eh, what's up?
+  DOC
+
+  around do |test|
+    test.run
+  end
+
+  it 'is expecting you' do
+    expect('you').to eql('you')
+  end
+
+  it 'is expecting you not to raise an error' do
+    expect { 'you' }.not_to raise_error
+  end
+
+  it 'has chained expectations' do
+    expect('you').to eql('you').and(match(/y/))
+  end
+
+  %w[who likes dynamic examples].each do |word|
+    let(word) { word }
+
+    describe "#{word}" do
+      context "#{word}" do
+        it "lets the word '#{word}' be '#{word}'" do
+          expect(send(word)).to eql(word)
+        end
+      end
+    end
+  end
+
+  it { foo; 1 && 2}
+  it('has a description too') { foo; 1 && 2}
+
+  it %{quotes a string weird} do
+  end
+
+  it((' '.strip! ; 1 && 'isnt a simple string')) do
+    expect(nil).to be(nil)
+  end
+
+  it((' '.strip! ; 1 && 'isnt a simple string')) do
+    double = double(:foo)
+
+    allow(double).to receive(:oogabooga).with(nil).and_return(nil)
+
+    expect(double.oogabooga(nil)).to be(nil)
+
+    expect(double).to have_received(:oogabooga).once
+  end
+
+  it 'uses a matcher' do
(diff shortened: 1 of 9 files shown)


dgollahon:
Hm. Looks like something in the smoke tests causes jruby to stack overflow...?

Errors are raised for

```ruby
RuboCop::Cop::RSpec::AlignLeftLetBrace
RuboCop::Cop::RSpec::AlignRightLetBrace
RuboCop::Cop::RSpec::AnyInstance
RuboCop::Cop::RSpec::AroundBlock
RuboCop::Cop::RSpec::Be
RuboCop::Cop::RSpec::BeEql
RuboCop::Cop::RSpec::BeforeAfterAll
RuboCop::Cop::RSpec::Capybara::CurrentPathExpectation
RuboCop::Cop::RSpec::Capybara::FeatureMethods
RuboCop::Cop::RSpec::ContextWording
RuboCop::Cop::RSpec::Cop
RuboCop::Cop::RSpec::DescribeClass
RuboCop::Cop::RSpec::DescribedClass
RuboCop::Cop::RSpec::DescribeMethod
RuboCop::Cop::RSpec::DescribeSymbol
RuboCop::Cop::RSpec::EmptyExampleGroup
RuboCop::Cop::RSpec::EmptyLineAfterExampleGroup
RuboCop::Cop::RSpec::EmptyLineAfterFinalLet
RuboCop::Cop::RSpec::EmptyLineAfterHook
RuboCop::Cop::RSpec::EmptyLineAfterSubject
RuboCop::Cop::RSpec::ExampleLength
RuboCop::Cop::RSpec::ExampleWithoutDescription
RuboCop::Cop::RSpec::ExampleWording
RuboCop::Cop::RSpec::ExpectActual
RuboCop::Cop::RSpec::ExpectChange
RuboCop::Cop::RSpec::ExpectInHook
RuboCop::Cop::RSpec::ExpectOutput
RuboCop::Cop::RSpec::FactoryBot::CreateList
RuboCop::Cop::RSpec::FactoryBot::DynamicAttributeDefinedStatically
RuboCop::Cop::RSpec::FactoryBot::StaticAttributeDefinedDynamically
RuboCop::Cop::RSpec::FilePath
RuboCop::Cop::RSpec::Focus
RuboCop::Cop::RSpec::HookArgument
RuboCop::Cop::RSpec::ImplicitExpect
RuboCop::Cop::RSpec::InstanceSpy
RuboCop::Cop::RSpec::InstanceVariable
RuboCop::Cop::RSpec::InvalidPredicateMatcher
RuboCop::Cop::RSpec::ItBehavesLike
RuboCop::Cop::RSpec::IteratedExpectation
RuboCop::Cop::RSpec::LeadingSubject
RuboCop::Cop::RSpec::LetBeforeExamples
RuboCop::Cop::RSpec::LetSetup
RuboCop::Cop::RSpec::MessageChain
RuboCop::Cop::RSpec::MessageExpectation
RuboCop::Cop::RSpec::MessageSpies
RuboCop::Cop::RSpec::MultipleDescribes
RuboCop::Cop::RSpec::MultipleExpectations
RuboCop::Cop::RSpec::MultipleSubjects
RuboCop::Cop::RSpec::NamedSubject
RuboCop::Cop::RSpec::NestedGroups
RuboCop::Cop::RSpec::NotToNot
RuboCop::Cop::RSpec::OverwritingSetup
RuboCop::Cop::RSpec::Pending
RuboCop::Cop::RSpec::PredicateMatcher
RuboCop::Cop::RSpec::Rails::HttpStatus
RuboCop::Cop::RSpec::ReceiveCounts
RuboCop::Cop::RSpec::RepeatedDescription
RuboCop::Cop::RSpec::RepeatedExample
RuboCop::Cop::RSpec::ReturnFromStub
RuboCop::Cop::RSpec::ScatteredLet
RuboCop::Cop::RSpec::ScatteredSetup
RuboCop::Cop::RSpec::SharedContext
RuboCop::Cop::RSpec::SharedExamples
RuboCop::Cop::RSpec::SingleArgumentMessageChain
RuboCop::Cop::RSpec::SubjectStub
RuboCop::Cop::RSpec::VerifiedDoubles
RuboCop::Cop::RSpec::VoidExpect
```

which I think is all or almost all our cops.

Not sure how to resolve this atm (I guess we could skip smoke tests for jruby?) since I have never interacted with jruby and don't feel like setting up an environment rigth now.

bquorning:
Perhaps we should change these RuboCop modules that we depend on, so the default filename can be configured.

spec/smoke_tests/weird_rspec_spec.rb
```diff
@@ -0,0 +1,233 @@
...
+  let(:uh_oh) { <<-HERE.strip + ", #{<<-THERE.strip}" }
+      Seriously
+  HERE
+      who designed these things?
+  THERE
+
+  it 'is insane' do
+    expect(uh_oh).to eql('Seriously, who designed these things?')
+  end
+end
```
bquorning: I think you had a lot of fun writing this file 😆

dgollahon: I categorically deny fun as a construct and emphatically disavow that it was experienced by me.

dgollahon:
> Docker makes it pretty easy:

Thanks for the tip, I'll try that out next time.

> Turns out the three methods you added to ExpectOffense are not actually defined in RuboCop::RSpec::ExpectOffense, so the super calls fail.

Yeah, I was actually aware of that when I added them (I didn't know where else felt right to put them), but I reasoned (correctly in MRI) that it didn't actually matter at all because the module gets included into the rspec context instance _after_ the original `rubocop` version so the `super` method should be in the ancestor chain. I think that's a JRuby bug, since logically it should be something _vaguely_ like:

```ruby
 # Where Class.new is anonymous rspec context
 Class.new.include(CopHelper).include(ExpectOffense)
```

just with more rspec magic under the hood.

I think your prepend trick might work here, but every approach feels a bit hacky, I'm just going to leave it off for now and open an issue.

> Perhaps we should change these RuboCop modules that we depend on, so the default filename can be configured.

Yeah, maybe. I'm not sure what a reasonable API is here though, given the way it gets included into rspec.

Issue #635, opened 2018-06-12
Option to require the first-argument of describe presence


Darhazer:
Follow up on #626, there should be an option that registers an offense if the describe have no parameters

PR #652, opened 2018-06-20
Add `FirstArgumentRequired` option to `RSpec/DescribeClass`

geniou:
See #635 

---

* [x] Added tests.

### DIFF
--- a/lib/rubocop/cop/rspec/describe_class.rb
+++ b/lib/rubocop/cop/rspec/describe_class.rb
@@ -22,11 +22,11 @@ module RuboCop
         MSG = 'The first argument to describe should be '\
               'the class or module being tested.'.freeze
 
-        def_node_matcher :valid_describe?, <<-PATTERN
-          {
-            (send {(const nil? :RSpec) nil?} :describe const ...)
-            (send {(const nil? :RSpec) nil?} :describe)
-          }
+        FIRST_ARGUMENT_REQUIRED_MSG =
+          'The first argument to describe is required.'.freeze
+
+        def_node_matcher :describe_with_const?, <<-PATTERN
+          (send {(const nil? :RSpec) nil?} :describe const ...)
         PATTERN
 
         def_node_matcher :describe_with_metadata, <<-PATTERN
@@ -48,11 +48,28 @@ module RuboCop
           return if shared_group?(root_node)
           return if valid_describe?(node)
 
+          unless node.arguments?
+            return add_offense(node, location: :expression,
+                                     message: FIRST_ARGUMENT_REQUIRED_MSG)
+          end
+
           describe_with_metadata(node) do |pairs|
             return if pairs.any?(&method(:rails_metadata?))
           end
 
           add_offense(args.first, location: :expression)
+        end
+
+        private
+
+        def valid_describe?(node)
+          return true unless node.arguments? || first_argument_required?
+
+          describe_with_const?(node)
+        end
+
+        def first_argument_required?
+          cop_config.fetch('FirstArgumentRequired', false)
         end
       end
     end
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## Master (Unreleased)
 
 * Add `RSpec/ReceiveNever` cop enforcing usage of `not_to receive` instead of `never` matcher. ([@Darhazer][])
+* Add `FirstArgumentRequired` option to `RSpec/DescribeClass`. ([@geniou][])
 
 ## 1.27.0 (2018-06-14)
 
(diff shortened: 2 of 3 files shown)


geniou:
Should `FirstArgumentRequired` be `true` by default?

bquorning:
Welcome back @geniou 👋 

Build fixed in #653.

lib/rubocop/cop/rspec/describe_class.rb
```diff
@@ -22,11 +22,15 @@ class DescribeClass < Cop
...
-            (send {(const nil? :RSpec) nil?} :describe)
-          }
+        FIRST_ARGUMENT_REQUIRED_MSG =
+          'The first argument to describe is required.'.freeze
+
+        def_node_matcher :describe_with_const?, <<-PATTERN
+          (send {(const nil? :RSpec) nil?} :describe const ...)
+        PATTERN
+
+        def_node_matcher :first_argument_empty?, <<-PATTERN
```
Darhazer: is it really necessary to use a node_matcher for this? Can you just check `node.arguments?` instead

lib/rubocop/cop/rspec/describe_class.rb
```diff
@@ -48,12 +52,31 @@ def on_top_level_describe(node, args)
...
+
+        private
+
+        def valid_describe?(node)
+          if first_argument_empty?(node) && !first_argument_required?
+            return true
+          end
+
+          describe_with_const?(node)
+        end
```
bquorning: This method reads a bit odd to me… Would something like the following work?
```
return false if first_argument_required? && node.arguments.none?

describe_with_const?
```

lib/rubocop/cop/rspec/describe_class.rb
geniou: Good point. I changed it.

lib/rubocop/cop/rspec/describe_class.rb
geniou: Had the same feeling, but did not know how to solve it. Now its better.

geniou:
@bquorning thanks.

lib/rubocop/cop/rspec/describe_class.rb
```diff
@@ -48,12 +48,29 @@ def on_top_level_describe(node, args)
...
             return if pairs.any?(&method(:rails_metadata?))
           end
 
           add_offense(args.first, location: :expression)
         end
+
+        private
+
+        def valid_describe?(node)
+          return true unless node.arguments? || first_argument_required?
```
Darhazer: This doesn't sound right. If first argument isn't required it doesn't check if it's describe with a const

geniou: The more readable way would be
`return true if !node.arguments? && !first_argument_required?`
and this should be the same as
`return true if !(node.arguments? || first_argument_required?)`
which is the same as
`return true unless node.arguments? || first_argument_required?`

Or am I missing something?

geniou: `return false if first_argument_required? && node.first_argument.nil?` is not working: when `first_argument_required` is `false` it will go to `describe_with_const?(node)` and this will return false.

The current code is working - event if I add more examples to the case where `FirstArgumentRequired` is set to `true`.

I see that its not that easy to read, but I have a hard time to find a more readable code, that is doing the same.

Darhazer: Yup, the fact that the node has arguments enters the check if the argument is `const`, so the code is correct, although counter-intuitive.

It seems that you can just check if node.first_argument.const_type? and not use the node matcher at all

lib/rubocop/cop/rspec/describe_class.rb
```diff
@@ -48,12 +48,29 @@ def on_top_level_describe(node, args)
           return if shared_group?(root_node)
           return if valid_describe?(node)
 
+          unless node.arguments?
+            return add_offense(node, location: :expression,
+                                     message: FIRST_ARGUMENT_REQUIRED_MSG)
+          end
```
bquorning: Could we merge this with the other `add_offense` call? Something like
```rb
if args.empty?
  add_offense(node, location: :expression, message: FIRST_ARGUMENT_REQUIRED_MSG)
else
  add_offense(args.first, location: :expression)
end
```

geniou:
@bquorning @Darhazer I realized a different problem: the `DescribeClass` is only checking in the top level `describe`, but I think this check should also apply to nested describes, right? What do you think? If so it would no longer fit into this cop.

Darhazer:
`The first argument to describe should be the class or module being tested` is aimed exactly for the top level describe. Nested describes are usually strings, describing what's being described ;)

geniou:
@Darhazer yes - but shouldn't we also check, that non-top-level describes are not empty?

bquorning:
> Nested describes are usually strings

Or nested constants.

> shouldn't we also check, that non-top-level describes are not empty?

I think that would be a useful, generic cop – checking that `describe` (and `context`) aren’t called without arguments.

geniou:
So I would go for a new cop that checks for empty `describe`, and `context` - feels also better then to squeeze it in.

geniou:
Close in favor of #657

PR #654, opened 2018-06-20
Fix false positive when the node is the only one inside of if/else branch

### LINKED ISSUE #648: RSpec/EmptyLineAfterExampleGroup invalid behavior
When I have a conditional spec based on some external arguments EmptyLine linter requires a new line after the spec but this does not seem right. I use flags like this sometimes for extensive integration testing of some internal components that I want to run only on special occasions. For such cases, this linter shouldn't require a new line as it doesn't make sense.

```ruby
describe Example do
  if ENV['FLAG']
    describe 'something funny' do
      let (:name) { 45 }

      it 'whatever' do
      end
    end
  end
end
```

Please use this code for reproduction:

```
git clone https://github.com/mensfeld/broken-rubocop-rspec4.git
cd broke-rubocop-rspec4
./run.sh
```

Result:

```
Inspecting 2 files
.C

Offenses:

spec/example_spec.rb:8:5: C: RSpec/EmptyLineAfterExampleGroup: Add an empty line after describe.
    end
    ^^^
```

Darhazer:
Fixes #648

---

* [x] Added tests.

### DIFF
--- a/lib/rubocop/rspec/blank_line_separation.rb
+++ b/lib/rubocop/rspec/blank_line_separation.rb
@@ -25,6 +25,12 @@ module RuboCop
         source_range(processed_source.buffer, last_line, start, content_length)
       end
 
+      def last_child?(node)
+        return true unless node.parent && node.parent.begin_type?
+
+        node.equal?(node.parent.children.last)
+      end
+
       def autocorrect(node)
         lambda do |corrector|
           missing_separating_line(node) do |location|
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## Master (Unreleased)
 
 * Add `RSpec/ReceiveNever` cop enforcing usage of `not_to receive` instead of `never` matcher. ([@Darhazer][])
+* Fix false positive in `RSpec/EmptyLineAfterExampleGroup` cop when example is inside `if`. ([@Darhazer][])
 
 ## 1.27.0 (2018-06-14)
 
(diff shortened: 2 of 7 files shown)


lib/rubocop/rspec/blank_line_separation.rb
```diff
@@ -25,6 +25,16 @@ def offending_loc(last_line)
         source_range(processed_source.buffer, last_line, start, content_length)
       end
 
+      def single_node?(node)
+        return if node.parent.nil?
+
+        if node.parent.if_type?
```
dgollahon: Is there a reason you've chosen to solve for only the if type explicitly? I imagine there are other constructs that would have this same problem (my first thought is `case`, but i possibly also `while` or `for` loops or something i'm not thinking of)

Would @bquorning's thought about checking if it's in a block/begin node not work https://github.com/rubocop-hq/rubocop-rspec/issues/648#issuecomment-397804501 ?

Darhazer: You are right, we have to check `begin` nodes. I even don't see why a block node have to be checked - if there is more then one child of a block, it have to be wrapped in a begin.

Also, it's not single_node, but last_child

PR updated

Darhazer:
@bquorning it seems that you've reviewed the old version of the diff. It's already called `last_child?` and doesn't return any `nil` :)

lib/rubocop/rspec/blank_line_separation.rb
```diff
@@ -25,6 +25,16 @@ def offending_loc(last_line)
         source_range(processed_source.buffer, last_line, start, content_length)
       end
 
+      def single_node?(node)
+        return if node.parent.nil?
```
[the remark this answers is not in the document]
Darhazer: It would be a good cop

PR #657, opened 2018-06-25
Add MissingExampleGroupArgument cop

geniou:
While doing #652 we realized we should not only check the top level `describe`. This closes the issue in #635

---

* [x] Added tests.


lib/rubocop/cop/rspec/missing_example_group_argument.rb
```diff
@@ -0,0 +1,44 @@
...
+      #   describe TestedClass do
+      #   end
+      #
+      #   describe "A feature example" do
+      #   end
+      class MissingExampleGroupArgument < Cop
+        MSG = 'The first argument to `%<method>s` should not be empty.'.freeze
+
+        def on_send(node)
+          return unless method_names.include?(node.method_name)
```
Darhazer: This would generate false positives on methods named describe,conext,feature,scenario, like
`current_context = Fixture.context`
Also I would prefer using the Language module here, and providing exclude configuration instead

lib/rubocop/cop/rspec/missing_example_group_argument.rb
```diff
@@ -0,0 +1,44 @@
...
+          return unless node.first_argument.nil?
```
Darhazer: You can use just `return if node.arguments?`

lib/rubocop/cop/rspec/missing_example_group_argument.rb
geniou: I hope I got your point. Can you please have another look if its fine like this now?

lib/rubocop/cop/rspec/missing_example_group_argument.rb
geniou: Good point. Thanks.

lib/rubocop/cop/rspec/empty_line_after_example_group.rb
```diff
@@ -25,11 +25,10 @@ module RSpec
       #
       class EmptyLineAfterExampleGroup < Cop
         include RuboCop::RSpec::BlankLineSeparation
+        include RuboCop::RSpec::Language::NodePattern
```
Darhazer: `NodePattern` is already included in cop class

lib/rubocop/cop/rspec/missing_example_group_argument.rb
```diff
@@ -0,0 +1,37 @@
...
+      #   # good
+      #   describe TestedClass do
+      #   end
+      #
+      #   describe "A feature example" do
+      #   end
+      class MissingExampleGroupArgument < Cop
+        MSG = 'The first argument to `%<method>s` should not be empty.'.freeze
+
+        def_node_matcher :example_group_send?, ExampleGroups::ALL.send_pattern
```
Darhazer: We don't specify if it's send or block or whatever in matcher names. `example_group?` reads better.

Darhazer: Also if you prefer to reuse the existing definition of `example_group?` you can use it, change the cop to listen to `on_block` and then return on `node.send_node.arguments?`

spec/rubocop/cop/rspec/missing_example_group_argument_spec.rb
```diff
@@ -0,0 +1,47 @@
+RSpec.describe RuboCop::Cop::RSpec::MissingExampleGroupArgument do
+  subject(:cop) { described_class.new }
+
+  it 'checks is ok if first argument to describe is given' do
```
Darhazer: this needs some rewording. `Accepts describe with an argument` for example

lib/rubocop/cop/rspec/missing_example_group_argument.rb
geniou: Good idea. Thanks.

spec/rubocop/cop/rspec/missing_example_group_argument_spec.rb
```diff
@@ -0,0 +1,59 @@
...
+
+      RSpec.describe FooClass do
+      end
+    RUBY
+  end
+
+  it 'accepts methods with a name like an example block' do
+    expect_no_offenses(<<-RUBY)
+      describe FooClass do
+        def describe
```
Darhazer: This is not a good false-positive test, as the node is of `def` type.
Use something like Scenario.context { "static" }, e.g. you need a method call

geniou: Sorry - got it wrong. Hopefully its now ok like this.

PR #658, opened 2018-06-26
Add new RSpec/HooksBeforeExamples cop

### LINKED ISSUE #376: Cop idea: no subject/let/before after an example
There will be good to have some cops that force the structure of an example group.
Having definition of let after the first example seems like a smell; and having the subject defined after an example is even worse. ScatteredLet only checks if the let are grouped together, but not if they are in the beginning of the example group.

Darhazer:
Closes #376 

---

* [x] Added tests.


lib/rubocop/cop/rspec/hooks_before_examples.rb
```diff
@@ -0,0 +1,101 @@
...
+      #     expect(foo).to be
+      #   end
+      #
+      class HooksBeforeExamples < Cop
+        include RangeHelp
+        include RuboCop::RSpec::FinalEndLocation
+
+        MSG = 'Move `%<hook>s` before the examples in the group.'.freeze
+
+        def_node_matcher :hook?, Hooks::ALL.block_pattern
```
geniou: Is this necessary? We have it defined here https://github.com/rubocop-hq/rubocop-rspec/blob/master/lib/rubocop/rspec/example_group.rb#L20

Darhazer: Well, this will require instantiating ExampleGroup. But since this matcher is common, we can move it to RSpec::Language::NodePattern module

Darhazer: I've extracted them in #659 and will update this if 659 is merged.

10x @geniou for mentioning this unnecessary duplication

lib/rubocop/cop/rspec/hooks_before_examples.rb
```diff
@@ -0,0 +1,99 @@
...
+      #
+      #   it 'checks what foo does' do
+      #     expect(foo).to be
+      #   end
+      #
+      class HooksBeforeExamples < Cop
+        include RangeHelp
+        include RuboCop::RSpec::FinalEndLocation
+
+        MSG = 'Move `%<hook>s` before the examples in the group.'.freeze
```
bquorning: I’m not sure if “before” is the right word here. Above? Atop? @dgollahon can you help us?

Darhazer: I borrowed the wording from `LetBeforeExamples` but for the `before` hook it sounds weird indeed. LedingSubject use `above` so I'll use it

PR #659, opened 2018-06-26
Extract common RSpec matchers into NodePattern

Darhazer:
hook?, let? and subject? are found in multiple cops and we already have a place to define common matchers.

Also some of the cops used over-complicated matchers that I've simplified

### DIFF
--- a/lib/rubocop/cop/rspec/multiple_subjects.rb
+++ b/lib/rubocop/cop/rspec/multiple_subjects.rb
@@ -36,10 +36,6 @@ module RuboCop
       class MultipleSubjects < Cop
         MSG = 'Do not set more than one subject per example group'.freeze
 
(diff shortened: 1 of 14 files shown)


bquorning:
Just FYI: I think CodeClimate has discovered zero of these duplications. 🙄

lib/rubocop/cop/rspec/multiple_subjects.rb
```diff
@@ -53,7 +49,7 @@ def on_block(node)
         def autocorrect(node)
           return unless node.method_name.equal?(:subject) # Ignore `subject!`
 
-          if named_subject?(node)
+          if node.send_node.arguments?
```
bquorning: I find it a bit unclear exactly what `node.send_node.arguments?` does, so perhaps still wrap this line as a private `#named_subject?` method?

PR #660, opened 2018-06-27
Use build in node matchers

Darhazer:
the content of a str_node can be retrieved with `str_content`.
This was the last place in code the `one()` util is actually used, so I think we should remove this method.

Regarding block nodes, send_node, arguments and body are retrieved by block's methods that access the children[0], children[1] and children[2] respectively, and it's more readable to use those named methods instead.

---

### DIFF
--- a/lib/rubocop/cop/rspec/shared_context.rb
+++ b/lib/rubocop/cop/rspec/shared_context.rb
@@ -68,11 +68,11 @@ module RuboCop
 
         def on_block(node)
           context_with_only_examples(node) do
-            add_shared_item_offense(node, MSG_EXAMPLES)
+            add_shared_item_offense(node.send_node, MSG_EXAMPLES)
           end
 
           examples_with_only_context(node) do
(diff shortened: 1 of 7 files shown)


lib/rubocop/cop/rspec/shared_context.rb
```diff
@@ -100,7 +100,7 @@ def examples_with_only_context(node)
 
         def add_shared_item_offense(node, message)
           add_offense(
-            node.children.first,
+            node,
```
dgollahon: Why not put the `send_node` call here instead of twice above?

dgollahon:
Oh, and one more minor note: I always try to push commits one at a time (I have a `git push-each` command setup for this) so that they each trigger a build and that way we can guarantee every commit passes CI, which I think is generally helpful for keeping the git history sensible and making it easy to do things like `git bisect` on.

Not a requirement, just a suggestion^

PR #666, opened 2018-07-27
Add FactoryBot/AttributeDefinedStatically Cop

composerinteralia:
We are planning to deprecate static attributes in
[`factory_bot#1135`][pull], and it would be nice to have a simple way to
autocorrect all static attributes into dynamic attributes.

This new cop is not compatible with the existing
`FactoryBot/DynamicAttributeDefinedStatically` and
`FactoryBot/StaticAttributeDefinedDynamically` cops, so we disabled
those in `config/default.yml`.

It probably makes sense to deprecate those in a future pull request, or
keep them and have them implemented as no-ops.

Co-authored-by: Sean Doyle <sean.p.doyle24@gmail.com>

[pull]: https://github.com/thoughtbot/factory_bot/pull/1135

---

* [x] Added tests.


composerinteralia:
Oops, just ran this cop against FactoryBot itself and realized there are a couple things we need to fix:

`user factory: :user_with_name, age: 20` is being changed to `user { factory: :user_with_name, age: 20 }`, which is not valid Ruby. We don't want to change that one at all, since it is an association, not a static attribute.

composerinteralia:
Looks like this is a problem with the existing cop as well https://github.com/rubocop-hq/rubocop-rspec/issues/655

composerinteralia:
@seanpdoyle I pushed a couple of commits to fix what I mentioned in the previous comments. If you have time next Friday, it would be great if you could take a look.

Darhazer:
Thank you! 

I think this cop could replace `DynamicAttributeDefinedStatically`, it just needs to provide configuration whether to add offence just for dynamic attributes or for all (which now should be the default).

bquorning:
@Darhazer Is there any reason to make this cop configurable – and to keep `StaticAttributeDefinedDynamically` – when the syntax is deprecated in FactoryBot?

We need alsoto discuss *when* to merge and release this pr. Right after FactoryBot has been released with the deprecation message, I think.

composerinteralia:
@bquorning I was actually hoping to refer to this Cop in the upgrade guide, or maybe even right in the deprecation warning. I want to make this as painless as possible for people. I was thinking something like:

> To automatically update from static attributes to dynamic attributes,
> you can add rubocop-rspec to your Gemfile, then run
> `rubocop --require rubocop-rspec --only FactoryBot/AttributeDefinedStatically`

 If we would rather not merge this until the new version of FactoryBot is out, I may just link to this PR in our upgrade guide.

@Darhazer I will follow whatever you think is best regarding making this cop configurable. I do imaging we would want to remove the old cops at some point.

Darhazer:
@bquorning @composerinteralia  the only reason to make this cop configurable is to be easier to support both cops, as users may not upgrade to the new FactoryBot (that's why we still support FactoryGirl namespace). But I'm fine with just removing the other 2 cops, as we ought to push users towards the current best practives.

As for the merging, I think we should merge it as soon as FactoryBot commit to the deprecation (e.g. when they merge in master), so we help users prepare for the deprecation - the same we did for the `be` matcher without arguments.

composerinteralia:
@bquorning OK, it's official. I merged https://github.com/thoughtbot/factory_bot/pull/1135

I can update this PR to remove the other 2 cops later this week.

lib/rubocop/cop/rspec/factory_bot/attribute_defined_statically.rb
```diff
@@ -111,25 +109,31 @@ def method_uses_parens?(location)
...
+          def autocorrect_replacing_parens(node)
+            left_braces, right_braces = braces(node)
 
-          def autocorrect_replacing_parens(node,
-                                           start_token = ' { ',
-                                           end_token = ' }')
             lambda do |corrector|
-              corrector.replace(node.location.begin, start_token)
-              corrector.replace(node.location.end, end_token)
+              corrector.replace(node.location.begin, ' ' + left_braces)
```
seanpdoyle: What makes the `' ' +` padding necessary?

spec/rubocop/cop/rspec/factory_bot/attribute_defined_statically_spec.rb
```diff
@@ -86,6 +86,16 @@
     RUBY
   end
 
+  it 'accepts valid association definitions' do
+    expect_no_offenses(<<-RUBY)
+      FactoryBot.define do
+        factory :post do
+          author age: 42, factory: :user
```
seanpdoyle: Do we need a test case for `association :author, age: 42, factory: :user`, or is that covered by our other code?

lib/rubocop/cop/rspec/factory_bot/attribute_defined_statically.rb
composerinteralia: Without it `attribute(value)` would turn into `attribute{ value }` instead of `attribute { value }` We don't need the extra space when there are no parens because there is already a space  
(`attribute value` becomes `attribute { value }`.

spec/rubocop/cop/rspec/factory_bot/attribute_defined_statically_spec.rb
composerinteralia: I think that is in the same category as the sequence test below, so probably already covered well enough.

composerinteralia:
I removed those two cops. This should be ready to go.

bquorning:
> Is there any reason […] to keep `StaticAttributeDefinedDynamically` […] when the syntax is deprecated in FactoryBot?

> But I'm fine with just removing the other 2 cops, as we ought to push users towards the current best practices.

We are not following semver *yet*, are we? Because this is a breaking change, and I’m reluctant to make a v2.0 release just because of this new cop.

I seem to recall there was some discussion about breaking changes and semver in the RuboCop project recently. @bbatsov do you recall the outcome of that discussion?

Darhazer:
I think `semver` would be followed after the release of RuboCop 1.0. Our users will forgive us removing two cops (actually upgrading one and removing the other). We have to mention the removal and the reasoning in the changelog though.

bquorning:
@composerinteralia I think the first three commits can be squashed – do you agree? And could you please add a (separate) changelog entry for the last commit (removal of two existing cops).

composerinteralia:
@bquorning I squashed those three commits and added the changelog entry for the other commit. 👍

composerinteralia:
@Darhazer I updated with your suggestions. Good call on using `#first_argument` and `#parenthesized?`.

bbatsov:
The PR of the Beast! 😆

mvz:
> I think semver would be followed after the release of RuboCop 1.0.

Since rubocop-rspec is using post-1.x versioning itself, I was kind of assuming it followed SemVer.

However, since each added cop could be a considered a breaking change, in the sense that it could break the build of projects that use rubocop-rspec, using SemVer in the strictest sense may not be practical.

In any case, I am tightening any dependencies on rubocop-rspec to `~> 1.28.0`.

samjewell:
Can I ask why this cop was added to `rubocop-rspec` rather than just added to `rubocop` directly?

I'm trying to run this cop locally on a repo that uses minitest, but it's not making any autocorrections (although it does say 107 files inspected, no offenses detected). Does this cop have some dependencies on RSpec that aren't obvious from this PR?

```
➜  bridgeu git:(master) ✗ rubocop test/factories \
  --require rubocop-rspec \
  --only FactoryBot/AttributeDefinedStatically \
  --auto-correct
Inspecting 107 files
...........................................................................................................

107 files inspected, no offenses detected
```

composerinteralia:
> Can I ask why this cop was added to rubocop-rspec rather than just added to rubocop directly?

There were already `DynamicAttributeDefinedStatically` and `StaticAttributeDefinedDynamically` cops in this library, and `AttributeDefinedStatically` was a modification of those.

I think a factory_bot-specific cop would be outside the scope of the main rubocop library, but one idea someone proposed was to release a separate gem with just this one cop. That might make the upgrade less awkward for non-rspec uses of factory_bot.

> Does this cop have some dependencies on rspec that aren't obvious from this PR?

I think you are running into https://github.com/thoughtbot/factory_bot/issues/1173. This is a problem on the factory_bot and not the rubocop-rspec side. rubocop-rspec is only meant to work for rspec (it's in the name! 😄), but we tried to use it for an additional purpose. We tend to use rspec at thoughtbot, so the awkwardness for non-rspec use cases didn't come up right away.

bquorning:
> one idea someone proposed was to release a separate gem with just this one cop.

Having an entire gem for just two cops (I count also `FactoryBot/CreateList`) is probably overhead, at least if you’re thinking something like rubocop-rails, rubocop-performance and rubocop-rails (rubocop-factorybot). But you *could* release the cop files separately, and just ask users to add the file in `lib` and then `require` it like described at the top of https://docs.rubocop.org/en/stable/extensions/.

PR #668, opened 2018-08-14
Bump version to 1.28.0

bquorning:
Two months have passed. Let’s release.

Main question is whether it’s ok to release the new cop introduced in #666 (`FactoryBot/AttributeDefinedStatically`) without FactoryBot having released a new version? The deprecation is merged into master, and the cop’s recommendation is backwards compatible, so I think it’s ok to release now.

---

* [ ] Added tests.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Master (Unreleased)
 
+## 1.28.0 (2018-08-14)
+
 * Add `RSpec/ReceiveNever` cop enforcing usage of `not_to receive` instead of `never` matcher. ([@Darhazer][])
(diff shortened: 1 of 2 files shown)


bquorning:
I found one potential bug that I’ll just investigate before releasing.

bquorning:
```rb
FactoryBot.define do
  factory :foo do
    subject { 'subject' }
    bar { 'bar' }
  end
end
```
gives me
```
spec/factories.rb:3:5: C: RSpec/EmptyLineAfterSubject: Add empty line after subject.
    subject { 'subject' }
    ^^^^^^^^^^^^^^^^^^^^^
```
but it’s not specifically related to the changes in this release.

Merging!

PR #670, opened 2018-08-17
[Fix #669] Add ImplicitSubject cop

### LINKED ISSUE #669: Implicit subject?
I recently saw code similar to this in a spec file:

```rb
RSpec.describe Foo do
  subject { described_class.new }

  it 'does its thing' do
    some { setup }
    some.other.setup
    is_expected.to be_truthy
  end
end
```

That expectation – `is_expected.to be_truthy` – struck me as weird. I understand the readability of the one-liner version, e.g. `it { is_expected.to be_truthy }`, but when there is more than one line in the block, that expectation syntax is strange.

Would it be possible to check for – and disallow – this syntax?

Darhazer:
---

* [x] Added tests.


lib/rubocop/cop/rspec/implicit_subject.rb
```diff
@@ -0,0 +1,57 @@
...
+      #
+      #   # good
+      #   it { expect(subject).to be_truthy }
+      #
+      class ImplicitSubject < Cop
+        include ConfigurableEnforcedStyle
+
+        MSG = "Don't use implicit subject.".freeze
+
+        def_node_matcher :implicit_subject?, '(send nil? :is_expected)'
```
bquorning: Should we support `:should` and `:should_not` too? Or do we have a policy of not supporting the pre-rspec 3 syntax?

I noticed that `ImplicitExpect` support the older syntax, but I don’t know if any other cop does.

Darhazer: Since `should` is available in RSpec 3 (even with the `expect` expectation syntax) and we even have a cop that can be configured to prefer `should` over `is_expected`, it's better to cover them both here.

I'll update later today

Darhazer:
Updated

Darhazer:
I've found a crash in a real code base and fixed it.

Not sure if we have to add offense in this specific case though, or it may cause false positives

easybills-admin:
Hi @Darhazer.

Why do consider `it { is_expected.to be_truthy }` bad?
I'm just trying to understand your motivation :)

Cheers.

Darhazer:
I don't

The default configuration (single_line_only) allows `is_expected` in single line examples

easybills-admin:
@Darhazer I see, thanks.

But why multiple lines are considered bad? Ie.: what's wrong with:

```ruby
it 'is foobar', versioning: true do
  is_expected.to be_foobar
end
```

Cheers.

Darhazer:
Implicits don't read so well when they span multiple lines.

Personally, I'm against `it do` blocks (multiline block without description) and as soon as you add a description, `is_expected` is not read naturally. 

But it seems that the community opinion is that `is_expected` should be allowed for multi-line, single-statement examples (see #682) so we will relax the cop

PR #671, opened 2018-08-18
InstanceVariable: Also recommend local variable

jaredbeck:
Often, a local variable is a better choice than a `let`. For example,
an object that will only be used in a single example.

When we use too many `let`s, examples become hard to understand, because
it is not clear which are applicable to a given example.

So, we should recommend both local variables and `let`s.

* [x] Added tests.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Master (Unreleased)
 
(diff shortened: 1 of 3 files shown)


jaredbeck:
> Please link to your Github profile page.

done

PR #673, opened 2018-08-23
Only check method_name on nodes that define it

Darhazer:
https://github.com/rubocop-hq/rubocop/pull/6198 removed the generic #method_name matcher from Node.
Now it's defined only for block and method dispatch nodes

This change makes rubocop-rspec compatible with both 0.58.2 and master

---

* [x] ~Added tests.~

### DIFF
--- a/lib/rubocop/cop/rspec/described_class.rb
+++ b/lib/rubocop/cop/rspec/described_class.rb
@@ -115,7 +115,7 @@ module RuboCop
           if style == :described_class
             node.eql?(@described_class)
           else
-            node.method_name == :described_class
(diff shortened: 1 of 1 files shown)


Drenmi:
Sorry. Should have notified you about the change. I'm working on the node extensions, and that involves clearing up methods on `Node` that can go in more specific node types. 🙇

Darhazer:
No problems. [Testing against master](https://github.com/rubocop-hq/rubocop-rspec/issues/283) is in our goals since a long time. And I just (try to) keep an eye on rubocop anyway ;)

Darhazer:
We're still compatible :pizza:

PR #677, opened 2018-08-28
[Fix #676] Check attributes defined using `self`

### LINKED ISSUE #676: Some deprecated FactoryBot static attributes are not detected or auto-corrected
If a model has an attribute that is a Ruby keyword, you need to use `self.attribute` to set the value in a factory. For example:

```ruby
factory :course do
    start Date.new(2015, 1, 1)
    self.end Date.new(2015, 6, 1)
end
```

Only the `start` attribute is detected as a rubocop violation and fixed.

After running rubocop-rspec (per the FactoryBot deprecation warning), it looks like this, with the second deprecated instance still there:

```ruby
factory :course do
    start { Date.new(2015, 1, 1) }
    self.end Date.new(2015, 6, 1)
end
```

Darhazer:
---

* [x] Added tests.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Master (Unreleased)
 
+* Fix false negative in `FactoryBot/AttributeDefinedStatically` when attribute is defined on `self`. ([@Darhazer][])
+
 ## 1.29.0 (2018-08-25)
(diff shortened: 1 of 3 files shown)


Darhazer:
cc:  @composerinteralia

PR #689, opened 2018-09-23
Enforce dynamic attributes with explicit receiver

### LINKED ISSUE #687: Autocorrect skipped for FactoryBot/AttributeDefinedStatically in some cases
When an attribute is passed in the `do/end` block and used to define the value, autocorrection of the `FactoryBot/AttributeDefinedStatically` cop does not work.

Given this factory,

```ruby
factory :user do |f| 
  f.first_name 'John' # not being autocorrected
  last_name 'Doe'     # autocorrected properly
end 
```

it doesn't take into consideration the `first_name` attribute because of the `f.` before:
```ruby
# After running autocorrection

factory :user do |f| 
  f.first_name 'John'
  last_name { 'Doe' }
end 
```

Useful informations:
- Rubocop: `0.59.1 (using Parser 2.5.1.2, running on ruby 2.3.3 x86_64-darwin17)`
- Rubocop RSpec: `1.29.1`
- FactoryBot: `4.11.1`

composerinteralia:
Fixes #687

factory_bot allows you to define attributes on an explicit receiver by
passing in the definition proxy as the first argument to the block.

This is not quite ready to be merged, but I wanted to get some feedback before going any further. Does this approach make sense to you?

The class is too big, so I will have to move some things around to get the build to pass. Any advice on the best way to break this up?

---

* [X] Added tests.

### DIFF
--- a/lib/rubocop/cop/rspec/factory_bot/attribute_defined_statically.rb
+++ b/lib/rubocop/cop/rspec/factory_bot/attribute_defined_statically.rb
@@ -27,51 +27,17 @@ module RuboCop
         class AttributeDefinedStatically < Cop
           MSG = 'Use a block to declare attribute values.'.freeze
 
-          ATTRIBUTE_DEFINING_METHODS = %i[factory trait transient ignore].freeze
-
-          UNPROXIED_METHODS = %i[
-            __send__
-            __id__
-            nil?
-            send
-            object_id
-            extend
-            instance_eval
-            initialize
-            block_given?
-            raise
-            caller
-            method
-          ].freeze
-
-          DEFINITION_PROXY_METHODS = %i[
-            add_attribute
-            after
-            association
-            before
-            callback
-            ignore
-            initialize_with
-            sequence
-            skip_create
-            to_create
-          ].freeze
-
-          RESERVED_METHODS =
-            DEFINITION_PROXY_METHODS +
-            UNPROXIED_METHODS +
-            ATTRIBUTE_DEFINING_METHODS
-
           def_node_matcher :value_matcher, <<-PATTERN
-            (send {self nil?} !#reserved_method? $...)
+            (send _ !#reserved_method? $...)
           PATTERN
 
           def_node_search :factory_attributes, <<-PATTERN
-            (block (send nil? #attribute_defining_method? ...) _ { (begin $...) $(send ...) } )
+            (block (send _ #attribute_defining_method? ...) _ { (begin $...) $(send ...) } )
           PATTERN
 
           def on_block(node)
             factory_attributes(node).to_a.flatten.each do |attribute|
+              next unless offensive_receiver?(attribute.receiver, node)
               next if proc?(attribute) || association?(attribute)
 
               add_offense(attribute, location: :expression)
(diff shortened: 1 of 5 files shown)


spec/rubocop/cop/rspec/factory_bot/attribute_defined_statically_spec.rb
```diff
@@ -98,6 +113,16 @@
     RUBY
   end
 
+  it 'does not add offense it method called on another object' do
```
composerinteralia: oops. "it" -> "if"

spec/rubocop/cop/rspec/factory_bot/attribute_defined_statically_spec.rb
```diff
@@ -66,6 +66,21 @@
     RUBY
   end
 
+  it 'registers an offense for attributes defined on explicit receiver' do
+    expect_offense(<<-RUBY)
+      FactoryBot.define do
+        factory :post do |post_definition|
+          post_definition.end Date.tomorrow
+          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ Use a block to declare attribute values.
+          post_defintion.trait :published do |published_definition|
```
jonatas: typo: `post_defintion` -> `post_defintion`

lib/rubocop/cop/rspec/factory_bot/attribute_defined_statically.rb
```diff
@@ -88,6 +89,20 @@ def autocorrect(node)
 
           private
 
+          def offensive_receiver?(receiver, node)
```
jonatas: I guess we can manage to solve this problem by mixing custom methods with node pattern. Example:

```ruby
          def capture_arg(sym)
            @block_argument = sym
          end

          def from_capture?(node)
            node.node_parts == [@block_argument]
          end
```

and then we need to use the capture arg and the function to compare the nodes:

```ruby
          def_node_matcher :value_matcher, <<-PATTERN
            (send { self nil? (lvar #capture_arg) } !#reserved_method? $...)
          PATTERN

          def_node_search :factory_attributes, <<-PATTERN
            (block (send {nil? #from_capture?} #attribute_defining_method? ...) _ { (begin $...) $(send ...) } )
          PATTERN
```

composerinteralia:
Sorry it took so long for me to come back to this.

@jonatas thanks for the review - I took your suggestions.

I ended up moving the constants out of this specific cop and onto `RuboCop::RSpec::FactoryBot`, but then on rebasing realized that the module was removed in https://github.com/rubocop-hq/rubocop-rspec/commit/a5df6e2588401982b80ce8036614231d78aa4745#diff-3b3465ed7d5d185d4bfa9acc4d26546f. I am open to other suggestions on how to organize things.

bquorning:
The module was removed only because it was empty. You’re welcome to add the module back with some contents :-)

lib/rubocop/cop/rspec/factory_bot/attribute_defined_statically.rb
```diff
@@ -27,47 +27,12 @@ module FactoryBot
...
-          ].freeze
-
-          RESERVED_METHODS =
-            DEFINITION_PROXY_METHODS +
-            UNPROXIED_METHODS +
-            ATTRIBUTE_DEFINING_METHODS
-
           def_node_matcher :value_matcher, <<-PATTERN
-            (send {self nil?} !#reserved_method? $...)
+            (send { self nil? (lvar #capture_arg) } !#reserved_method? $...)
```
Darhazer: Generally, you can use `%` in the pattern to pass variables, and avoid using the instance variable this way

composerinteralia: Interesting. I am unsure how to do use that. Is there documentation somewhere? Or an example?

I realized that when I took the suggestion to use instance variables I also introduced a regression that hadn't been tested, so I added a test case for it and went back to the version I had before for now.

PR #691, opened 2018-10-02
Add an option to allow enforcement of verified doubles in nameless doubles

BrentWheeldon:
Currently if you have a double/spy with no name (or default stubs) it will never cause a violation according to the `RSpec/VerifiedDoubles` cop. This adds an optional configuration to allow violations to be added for code that looks like:

```
foo = double
allow(foo).to receive(:bar) do |arg1|
  “#{arg1} based result”
end
```

I left the default configuration not enforcing this cop on nameless doubles so as to make this a non-breaking change.

### DIFF
--- a/spec/rubocop/cop/rspec/verified_doubles_spec.rb
+++ b/spec/rubocop/cop/rspec/verified_doubles_spec.rb
@@ -53,12 +53,25 @@ RSpec.describe RuboCop::Cop::RSpec::VerifiedDoubles, :config do
     end
   end
 
-  it 'ignores doubles without a name' do
-    expect_no_offenses(<<-RUBY)
+  it 'doubles that have no name specified' do
+    expect_offense(<<-RUBY)
       it do
         foo = double
+              ^^^^^^ Prefer using verifying doubles over normal doubles.
       end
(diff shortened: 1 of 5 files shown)


BrentWheeldon:
@Darhazer Thanks for the quick review, and sorry for the spec sloppiness. Those are fixed up now.

spec/rubocop/cop/rspec/verified_doubles_spec.rb
```diff
@@ -53,14 +53,27 @@
     end
   end
 
-  it 'ignores doubles without a name' do
-    expect_no_offenses(<<-RUBY)
+  it 'finds a double when the no name is specified' do
```
Darhazer: this one needs improving. Perpahs `finds a double without a name` is enough

Darhazer: `doubles that have no name specified` also works, as it is consistent with the other test

BrentWheeldon: 👍 fixed that up.

PR #692, opened 2018-10-02
Add a cop to enforce expected exceptions to be specified

daveworth:
In order to avoid false-positives in the context of exception checking it is valuable to be able to enforce that each `to raise` pair to include specifying information about the type of exception.

We have this vendored internally at Strava and have been happy with the improvement and would love an opt-in solution upstream.

---

* [x] Added tests.


bquorning:
I think this is a brilliant idea. I’m open to having this cop enabled by default – why not?

I have a few comments regarding the implementation, but I’ll wait with those until I’ve heard other opinions on the rule itself.

/cc @Darhazer

Darhazer:
I agree, the rule is good and let's have it enabled by default

lib/rubocop/cop/rspec/unspecified_exception.rb
```diff
@@ -0,0 +1,83 @@
...
+          (block
+            #{EMPTY_RAISE_ERROR_PATTERN}
+            (args (arg ...))
+            ...
+          )
+        PATTERN
+
+        def on_send(node)
+          return unless empty_raise_error_call(node) &&
+            !raise_error_with_block(node.parent)
```
bquorning: What does the `raise_error_with_block` check on the parent node do?

lib/rubocop/cop/rspec/unspecified_exception.rb
```diff
@@ -0,0 +1,83 @@
...
+      #
+      #     expect {
+      #       raise StandardError.new('error')
+      #     }.to raise_error(StandardError) { |err| expect(err.data).to eq(42) }
+      #
+      #     expect {
+      #       raise StandardError.new('error')
+      #     }.to raise_error(StandardError) do |err|
+      #       expect(err.data).to eq(42)
+      #     end
```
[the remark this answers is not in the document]
daveworth: the difference is the style of block (curly braces vs. `do`/`end` and while I was parsing the AST one passed while the other did not.  It's been about three months since I did that work so I don't remember the exact details of the issue but I suspect it was because of the different bindings of the two types of blocks.  Is that reasonable?

lib/rubocop/cop/rspec/unspecified_exception.rb
daveworth: oh!  I think this is exactly what I was referring to before, that is the matcher that catches `do`/`end` defined blocks but without it curly-braced blocks pass!  To validate I commented out the matcher and the second half of the conditional and the `allows classes with blocks with do/end` example failed.

daveworth:
Alright, I've addressed all of the feedback I believe - I've made the cop a default, used `MSG` correctly, and even remembered some tricky semantics.  Please take another look.

bquorning:
Any comments on the implementation @Darhazer ?

lib/rubocop/cop/rspec/unspecified_exception.rb
```diff
@@ -0,0 +1,82 @@
...
+          (send
+            (block
+                (send nil? :expect) ...)
+            :to
+            (send nil? :raise_error)
+          )
+        PATTERN
+        def_node_matcher :empty_raise_error_call, EMPTY_RAISE_ERROR_PATTERN
+
+        def_node_matcher :raise_error_with_block, <<-PATTERN.freeze
```
Darhazer: IMO this matcher makes code harder to understand. I would replace it with a simple method that checks whether the parent node is a block with an argument, like this one:

```ruby
def on_send(node)
  return unless empty_raise_error_call(node) &&
    !block_with_args?(node.parent)
    # ... rest of the code
end

def block_with_args?(node)
  return unless node && node.block_type?
  node.arguments?
end
```

lib/rubocop/cop/rspec/unspecified_exception.rb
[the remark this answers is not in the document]
Darhazer: I'm not sure we need so much good examples in the documentation, as most of them include `to raise_error(StandardError)`. It's good that you enumerated all of them for the tests though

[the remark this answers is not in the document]
daveworth: That's fair, any feelings on which particular ones I should remove?

lib/rubocop/cop/rspec/unspecified_exception.rb
daveworth: That's great, much simpler - I appreciate it!

daveworth:
I tried to do something clever to DRY out the specs to cover both `raise_error` and `raise_exception` but failed to do so in a way that didn't make the specs worse (IMO).  I've duplicated the tests in two contexts but am definitely willing to make them better if you have good ideas on how.

daveworth:
Alright, I've addressed the second round of feedback - how are we feeling about this now @Darhazer ?

lib/rubocop/cop/rspec/unspecified_exception.rb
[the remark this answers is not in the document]
Darhazer: I would leave the first one (that contains only `raise_error(StandardError)`) as it's shortest and cover the fact that you have to pass an error class

lib/rubocop/cop/rspec/unspecified_exception.rb
```diff
@@ -0,0 +1,79 @@
...
+          (send
+            (block
+                (send nil? :expect) ...)
+            :to
+            (send nil? {:raise_error :raise_exception})
+          )
+        PATTERN
+
+        def on_send(node)
+          return unless empty_raise_error_or_exception_call(node) &&
```
Darhazer: I wonder if this can be named shorter, so it fits one line

lib/rubocop/cop/rspec/unspecified_exception.rb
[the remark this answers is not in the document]
daveworth: cool.  Done!

lib/rubocop/cop/rspec/unspecified_exception.rb
daveworth: it can!

lib/rubocop/cop/rspec/unspecified_exception.rb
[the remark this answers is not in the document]
Darhazer: You could leave the ones that use the message though :)

CHANGELOG.md
```diff
@@ -6,6 +6,7 @@
 * Fix `FactoryBot/AttributeDefinedStatically` not working when there is a non-symbol key. ([@vzvu3k6k][])
 * Fix false positive in `RSpec/ImplicitSubject` when `is_expected` is used inside `its()` block. ([@Darhazer][])
 * Add `single_statement_only` style to  `RSpec/ImplicitSubject` as a more relaxed alternative to `single_line_only`. ([@Darhazer][])
+* Add `RSpec/UnspecifiedException` as a non-default cop to encourage more-specific `expect{}.to raise_error(ExceptionType)`, or `raise_exception` style handling of exceptions. ([@daveworth][])
```
Darhazer: Hm, we actually made it a default cop :smirk:

PR #695, opened 2018-10-04
Sort config alphabetically

Darhazer:
Follow up to #694 

---

* [x] ~Added tests.~

### DIFF
--- a/spec/project/default_config_spec.rb
+++ b/spec/project/default_config_spec.rb
@@ -3,13 +3,16 @@ RSpec.describe 'config/default.yml' do
     RuboCop::ConfigLoader.load_file('config/default.yml')
   end
 
-  let(:cop_names) do
-    namespaces = {
+  let(:namespaces) do
+    {
       'rspec' => 'RSpec',
       'capybara' => 'Capybara',
       'factory_bot' => 'FactoryBot',
       'rails' => 'Rails'
     }
+  end
+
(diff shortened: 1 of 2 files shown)


spec/project/default_config_spec.rb
```diff
@@ -38,7 +41,17 @@ def cop_configuration(config_key)
...
-    expect(default_config.keys.sort).to eq(config_keys.sort)
+    expect(default_config.keys).to match_array(config_keys)
+  end
+
+  it 'sorts configuration keys alphabetically' do
+    namespaces.each do |_path, prefix|
+      expected = config_keys.select { |key| key.start_with?(prefix) }.sort
+      actual = default_config.keys.select { |key| key.start_with?(prefix) }
+      actual.each_with_index do |key, idx|
+        expect(key).to eq expected[idx]
```
bquorning: I tried to come up with a better way to set an expectation (`expect(actual).to match(expected)`) but I must admit that your matcher will give the most useful message in case of mismatch.

bquorning:
I am removing the “Require branches to be up to date before merging” criteria in order to merge this. Let me know if that was a bad idea.

spec/project/default_config_spec.rb
Darhazer: I borrowed it from Rubocop :blush:

PR #700, opened 2018-10-05
Remove unused modules

bquorning:
Removes the modules `RuboCop::RSpec::FactoryBot` and `RuboCop::RSpec::Capybara`. We still have `RuboCop::Cop::RSpec::FactoryBot` and `RuboCop::Cop::RSpec::Capybara`

The modules were introduced in #409 and #439 – I think by mistake. /cc @rspeicher and @jonatas 

---

* [ ] Added tests.


rspeicher:
Thanks @bquorning. I can't remember if/why they were actually required at the time, but looks like they are indeed extraneous. 👍

PR #707, opened 2018-10-13
Fix #702: don't flag n.times block with an argument

### LINKED ISSUE #702: CreateList with index
```yaml
FactoryBot/CreateList:
  EnforcedStyle: create_list
```
```
test 'create list test' do
  3.times { |n| create :user, created_at: n.months.ago }
end
```
```bash
spec/test_spec.rb:4:3: C: FactoryBot/CreateList: Prefer create_list.
  3.times { |n| create :user, created_at: n.months.ago }
  ^^^^^^^
```

Based on the [docs](https://www.rubydoc.info/gems/rubocop-rspec/RuboCop/Cop/RSpec/FactoryBot/CreateList) this should not be failing:
```
# bad
3.times { create :user }

# good
create_list :user, 3

# good
3.times { |n| create :user, created_at: n.months.ago }
```

Versions: 
  - rubocop (0.59.2)
  - rubocop-rspec (1.29.1)

Darhazer:

* [x] Added tests.

### DIFF
--- a/lib/rubocop/cop/rspec/factory_bot/create_list.rb
+++ b/lib/rubocop/cop/rspec/factory_bot/create_list.rb
@@ -30,9 +30,10 @@ module RuboCop
           MSG_CREATE_LIST = 'Prefer create_list.'.freeze
           MSG_N_TIMES = 'Prefer %<number>s.times.'.freeze
 
-          def_node_matcher :n_times_block?, <<-PATTERN
+          def_node_matcher :n_times_block_without_arg?, <<-PATTERN
             (block
               (send (int _) :times)
+              (args)
               ...
             )
           PATTERN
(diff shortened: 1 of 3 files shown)


lib/rubocop/cop/rspec/factory_bot/create_list.rb
```diff
@@ -30,9 +30,10 @@ class CreateList < Cop
           MSG_CREATE_LIST = 'Prefer create_list.'.freeze
           MSG_N_TIMES = 'Prefer %<number>s.times.'.freeze
 
-          def_node_matcher :n_times_block?, <<-PATTERN
+          def_node_matcher :n_times_block_without_arg?, <<-PATTERN
             (block
               (send (int _) :times)
+              (args)
```
bquorning: What does `(args)` mean here? Without arg?

Darhazer: empty arguments list, e.g. no `| |` in the block

bquorning: Ah, of course. It would be e.g. `(args _)` if there were any arguments.

PR #708, opened 2018-10-25
[Fix #453] Allow implicit subject in shared examples

### LINKED ISSUE #453: NamedSubject in shared examples
I think explicit calls to `subject` within `shared_examples` should not trigger a violation, as the subject could have different names in different contexts. Of, course, one could name the subject with some common name, like `subject(:named_subject)` and use this in the shared examples, but I think this defeats the purpose of the cop.

RST-J:
Having explicit subjects is a good idea, except for shared_examples which are meant to be used in different contexts that likely/possibly have different subjects. This PR introduces the `AllowInSharedExamples` option for `RSpec/NamedSubject` to let users decide whether to allow implicit subjects in shared examples or not.

I set the default for this option to `true` because I think, intuitively it is what one would expect.

Fixes #453

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,10 +2,11 @@
 
 ## Master (Unreleased)
 
+* Add `IgnoreSharedExamples` option for `RSpec/NamedSubject`. ([@RST-J][])
+
 ## 1.30.1 (2018-11-01)
 
 * `FactoryBot/CreateList` now ignores `times` blocks with an argument. ([@Darhazer][])
-* Add `AllowInSharedExamples` option for `RSpec/NamedSubject`. ([@RST-J][])
 
 ## 1.30.0 (2018-10-08)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## Master (Unreleased)
 
 * `FactoryBot/CreateList` now ignores `times` blocks with an argument. ([@Darhazer][])
+* Add `AllowInSharedExamples` option for `RSpec/NamedSubject`. ([@RST-J][])
 
 ## 1.30.0 (2018-10-08)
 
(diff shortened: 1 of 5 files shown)


bquorning:
I am just curious: when would you want to disable the `AllowInSharedExamples` option? Shouldn’t we just make it the default behavior, without it being a configuration option?

RST-J:
I have no problem with removing the flag, I personally also wouldn't want to use it with that option disabled. But maybe there are people who want to?

dgollahon:
Seems reasonable / consistent to be able to enforce it anywhere, but probably YAGNI so I'm kind of neutral here.

As a note on naming if the option is kept, rather than `AllowInSharedExamples`, what about `IgnoreSharedExamples`? It sounds like it's allowing/disallowing named subjects in examples (since it's the `NamedSubject` cop) but i think you want to allow/disallow unnamed subjects in shared examples.

RST-J:
Agreed, `IgnoreSharedExamples` better conveys the purpose - if we keep it at all.

Darhazer:
Well, I think it's weird to have arbitrary named subject in a shared example and forcing whoever uses it to name the subject after it. It may lead to code like:
```
describe '#jobs' do
  subject(:jobs) { board.jobs }
  it { ..... }

  it_behaves_like 'filterable collection', %i[title category] do
    subject(:collection) { jobs }
  end
```

We have perfectly named subject, we have options we can pass to the shared example, and then we need to alias the subject in the include_examples/it_behaves_like call. Even worse if it happen name to colide with another `let`. 

So shared example is meant to be used in different specs. It should not force specific subject name for all of them. And I doubt anyone would change the above option.

On the other hand, configuration doesnt hurt, it's just pitty there's no way to track if it's used or not, but since it's already configurable, why dropping it, just to add it again in case someone complains?

PR #710, opened 2018-10-27
Compatibility with RuboCop v0.60.0

bquorning:
There are a few interesting changes between RuboCop v0.59.2 and v0.60.0:

1. We need to add `[:exclude_limit]` when checking `cop.config_to_allow_offenses`.
2. Style/MutableConstant reported a few new issues.
3. Layout/AlignHash has been changed so a combination of `key` and `table` styles is no longer allowed. I disagree with this decision, and have voiced my opinion on the issue rubocop-hq/rubocop#6410
4. There seems to be some namespace collision between the cops `Rails/HttpStatus` and `RSpec/Rails/HttpStatus` … again. @Darhazer would you take a look at my last commit “tmp”?

---

* [ ] Added tests.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -9,6 +9,10 @@ AllCops:
     - 'tmp/**/*'
     - 'spec/smoke_tests/**/*.rb'
 
+# See https://github.com/rubocop-hq/rubocop/issues/6410
+Layout/AlignHash:
+  Enabled: false
+
 Metrics/BlockLength:
   Exclude:
     - rubocop-rspec.gemspec
--- a/Gemfile
+++ b/Gemfile
@@ -2,8 +2,6 @@ source 'https://rubygems.org'
 
 gemspec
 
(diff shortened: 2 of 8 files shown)


manual/cops_rails.md
```diff
@@ -41,6 +41,7 @@ it { is_expected.to have_http_status :error }
 
 Name | Default value | Configurable values
 --- | --- | ---
+VersionAdded | `0.54` | Float
```
bquorning: This is wrong. There must be some compartment/namespace bug hiding somewhere.

/cc @Darhazer @bbatsov

anthony-robin:
The point number `4.` can be related to this [issue](https://github.com/rubocop-hq/rubocop-rspec/issues/611) as well.

bquorning:
Probably also related to rubocop-hq/rubocop#5251

bquorning:
@Darhazer Fixed and rebased.

PR #711, opened 2018-11-01
Load only RuboCop-RSpec config when generating cops documentation

Darhazer:
This resolves the case where there is a name conflict, as it doesn't load the rest of RuboCop configuration (which is not needed, we are trying to skip the rest of the cops anyway)

---

* [x] ~Added tests.~

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -2,6 +2,8 @@ source 'https://rubygems.org'
 
 gemspec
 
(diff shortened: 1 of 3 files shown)


bquorning:
You can add a restriction to `'rubocop', '< 0.60.0'` in ./Gemfile. That will only affect the version used in development/test.

PR #713, opened 2018-11-01
[Fix #712] check that the method is called on `receive`

### LINKED ISSUE #712: False positive for RSpec/ReceiveCounts auto-correction
Whilst running `rubocop --safe-auto-correct` on a large code base, I encountered the following error:

```
C: RSpec/ReceiveCounts: Use .twice instead of .exactly(2).times.
      expect { described_class.perform_now }.to have_published_event.exactly(2).times
                                                                    ^^^^^^^^^^^^^^^^^
```

The issue is that `have_published_event` is a custom class, which did not define a `#twice` method.

I have now gone ahead and defined this method. But the important point I wish to raise is that this autocorrection should be flagged as "unsafe" (and therefore not be run by `--safe-auto-correct`), since it cannot know its receiver.

Darhazer:
The cop didn't check that the receiver is actually a stub, leading to false positives if other objects implement similar API

---

* [x] Added tests.

### DIFF
--- a/spec/rubocop/cop/rspec/receive_counts_spec.rb
+++ b/spec/rubocop/cop/rspec/receive_counts_spec.rb
@@ -78,6 +78,12 @@ RSpec.describe RuboCop::Cop::RSpec::ReceiveCounts do
     RUBY
   end
 
+  it 'allows exactly(1).times when not called on `receive`' do
+    expect_no_offenses(<<-RUBY)
+      expect(action).to have_published_event.exactly(1).times
+    RUBY
+  end
+
   include_examples 'autocorrect',
                    'expect(foo).to receive(:bar).exactly(1).times { true }',
                    'expect(foo).to receive(:bar).once { true }'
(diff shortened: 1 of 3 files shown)


Darhazer:
I switched to just skipping the auto-correct, but still flagging, as if you are going to mimic RSpec API, better mimic it completely :)

I'll add Safe/SafeAutocorrect config to cops, following the 0.60 release, in a separate PR.

spec/rubocop/cop/rspec/receive_counts_spec.rb
```diff
@@ -78,11 +78,23 @@
     RUBY
   end
 
+  it 'flags exactly(1).times when not called on `receive`' do
```
bquorning: 1 or 2? :-)

bquorning:
It seems a bit weird to have a cop suggesting alternative code that won’t work… I guess the `Safe`/`SafeAutocorrect` PR will fix that?

Darhazer:
> It seems a bit weird to have a cop suggesting alternative code that won’t work… I guess the Safe/SafeAutocorrect PR will fix that?

I'm a bit on a fence on this as well, but it will suggest a better API. The cop will not run under `--safe` in this case. Which I don't like much, maybe it's better to be it's own option `Safe`, so it reports offenses, ignoring the false positives....

In the implementation, is just deciding where to add the check for stub. We just have to make a decision about whether to suggest changing the API (which might not be in the control of the user anyway) or we will skip on those cases.

spec/rubocop/cop/rspec/receive_counts_spec.rb
Darhazer: Fixed

Darhazer:
Rebased and changed the cop not to report the error

Darhazer:
@bquorning what about this one and doing a bug-fix release after?

PR #717, opened 2018-12-05
Add support for built-in exist matcher

### LINKED ISSUE #716: Prefer the built-in exist matcher
I've inherited a codebase that is full of `expect(foo.exists?).to be_truthy`. The RSpec/PredicateMatcher already checks for most built-ins, but doesn't know about [the `exist` matcher](https://relishapp.com/rspec/rspec-expectations/v/3-8/docs/built-in-matchers/exist-matcher). As a result, it suggests I replace my code with `expect(foo).to be_exists`.

I have failing tests in _predicate_matcher_spec.rb_ to add in the built-in exists matcher and I am working on making them pass.

mkenyon:
I've inherited a codebase that is full of `expect(foo.exists?).to be_truthy`. The RSpec/PredicateMatcher already checks for most built-ins, but doesn't know about the [built-in `exist` matcher][built-in]. As a result, it suggests I replace my code with`expect(foo).to be_exists`.

[built-in]: https://relishapp.com/rspec/rspec-expectations/v/3-8/docs/built-in-matchers/exist-matcher

I am not fully happy with [the change I made in the explicit module][the-change]. I'm happy to adjust that if y'all have a better idea.

[the-change]: https://github.com/mkenyon/rubocop-rspec/blob/feature/prefer-exist-matcher-716/lib/rubocop/cop/rspec/predicate_matcher.rb#L179

[Fixes #716]

--

* [x] Added tests.


bquorning:
Thank you for the PR @mkenyon.

I saw your bug report #716, but I am not sure the logic belongs in `PredicateMatcher`. Strictly speaking, that cop should only be concerned with `be_<foo>` and `has/have_<foo>` matchers. Well, at least that’s my opinion – based on how rspec-expectations has separate logic for these matchers, as well as a separate documentation topic: https://relishapp.com/rspec/rspec-expectations/v/3-8/docs/built-in-matchers/predicate-matchers

That said, the cop already contains logic for *other* built-in matchers (is_a?, instance_of?, include?, respond_to?)… Perhaps we should have separate cops for each built-in matcher? That would make it easier for our users to enable/disable them, but possibly lead to a lot of almost duplicated cops…

Darhazer:
The cop actually needs to know about other predicate matchers in order not to suggest turning them to `be_` methods. Having said that, I'm OK with keeping the logic in this cop for the time being.

lib/rubocop/cop/rspec/predicate_matcher.rb
```diff
@@ -174,7 +176,7 @@ def check_explicit(node) # rubocop:disable Metrics/MethodLength
 
         def predicate_matcher_name?(name)
           name = name.to_s
-          name.start_with?('be_', 'have_') &&
+          (name.start_with?('be_', 'have_') || name == 'exist') &&
```
Darhazer: maybe just add `exist` in the BUILD_IN_MATCHERS ?

Darhazer: how about:
```ruby
def predicate_matcher_name?(name)
  name = name.to_s
  return false if BUILT_IN_MATCHERS.include?(name)
  name.start_with?('be_', 'have_') && !name.end_with?('?')
end
```

mkenyon: Yep, that'll work.

Darhazer: Oh, I've been mistaken. It will not work, as it have to return `true` for exist. Seems that we'll need another array for the build-in matchers that should not be ignored.

I see you just removed it, so with `explicit`  it will not require replacing `to exist` with `.exist?).to be_truthy`. I'm fine with leaving this out of scope and just fix the `inflect` style to use `exist` and not `be_exist`. We will have to rework the cop at some point in the future.

@bquorning WDYT?

Darhazer: On second thought, `to exist` can't be auto-corrected, as it could not know if the object responds to `exist?` or `exists?`, so it's even better now that `exist` is accepted.

PR #720, opened 2018-12-22
[Fix #719] Allow using arrays, variables and method in receive_message_chain

### LINKED ISSUE #719: False positive RSpec/SingleArgumentMessageChain
When run agains the following code:
```ruby
RSpec.describe 'the Test' do
  let(:foo) { double('Foo') }
  let(:methods) { %i[some method] }

  before do
    allow(foo).to receive_message_chain(methods).and_return(true)
  end

  it 'returns true' do
    expect(foo.some.method).to eq(true)
  end
end
```
Rubocop suggests changing `receive_message_chain` into `receive`. Running `rubocop -a` makes the change and breaks existing code. The same behaviour can also be seen with `methods` set to `'some.method'`

The cop works alright when passing array or string with many methods to `receive_message_chain`. The following code will not cause `RSpec/SingleArgumentMessageChain`:
```ruby
allow(foo).to receive_message_chain(%i[some method]).and_return(true)
allow(foo).to receive_message_chain('some.method').and_return(true)
```

This can be seen with rubocop-rspec 1.30.1 and rubocop 0.61.1

Darhazer:
Array with multiple elements is a valid way to specify a message chain.
Also, in case a variable (or send node) is used, we cannot know if it doesn't evaluate to an array with multiple elements, so we should not report those.

---

* [x] Added tests.

### DIFF
--- a/lib/rubocop/cop/rspec/single_argument_message_chain.rb
+++ b/lib/rubocop/cop/rspec/single_argument_message_chain.rb
@@ -26,9 +26,7 @@ module RuboCop
 
         def on_send(node)
           message_chain(node) do |arg|
-            return if arg.to_s.include?('.')
-
-            return if arg.hash_type? && !single_key_hash?(arg)
+            return if valid_usage?(arg)
 
             add_offense(node, location: :selector)
           end
@@ -39,11 +37,26 @@ module RuboCop
             corrector.replace(node.loc.selector, replacement(node.method_name))
             message_chain(node) do |arg|
               autocorrect_hash_arg(corrector, arg) if single_key_hash?(arg)
+              autocorrect_array_arg(corrector, arg) if arg.array_type?
             end
           end
         end
 
         private
+
+        def valid_usage?(node)
+          return true unless node.literal? || node.array_type?
+
+          case node.type
+          when :hash then !single_key_hash?(node)
+          when :array then !single_element_array?(node)
+          else node.to_s.include?('.')
+          end
+        end
+
+        def single_element_array?(node)
+          node.child_nodes.one?
+        end
 
         def autocorrect_hash_arg(corrector, arg)
           key, value = *arg.children.first
@@ -51,6 +64,12 @@ module RuboCop
           corrector.replace(arg.loc.expression, key_to_arg(key))
           corrector.insert_after(arg.parent.loc.end,
                                  ".and_return(#{value.source})")
+        end
+
+        def autocorrect_array_arg(corrector, arg)
+          value = arg.children.first
+
+          corrector.replace(arg.loc.expression, value.source)
         end
 
         def key_to_arg(node)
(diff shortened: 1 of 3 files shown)


spec/rubocop/cop/rspec/single_argument_message_chain_spec.rb
```diff
@@ -48,6 +48,50 @@
...
+
+      include_examples(
+        'autocorrect',
+        'before { allow(foo).to receive_message_chain([:one]) { :two } }',
+        'before { allow(foo).to receive(:one) { :two } }'
+      )
+    end
+
+    context 'with multiple-element array argument' do
+      it 'reports an offence' do
```
bquorning: Does it? :-)

bquorning:
I don’t see the documentation mention anything about passing an array to `receive_messge_chain`: https://relishapp.com/rspec/rspec-mocks/docs/working-with-legacy-code/message-chains

Darhazer:
I didn't find documentation as well, so I've tested manually. `receive_messge_chain([:one, :two)` works exactly the same as `receive_message_chain(:one, :two)`.

Fixed the example wording

lib/rubocop/cop/rspec/single_argument_message_chain.rb
```diff
@@ -39,12 +37,27 @@ def autocorrect(node)
...
               autocorrect_hash_arg(corrector, arg) if single_key_hash?(arg)
+              autocorrect_array_arg(corrector, arg) if arg.array_type?
             end
           end
         end
 
         private
 
+        def valid_usage?(node)
+          return true unless node.literal? || node.array_type?
```
bquorning: Isn’t Array also a literal?

Darhazer: Nope. I guess it's because it may contain non-literals. I'm checking Rubocop's Node source every time.

bquorning: Hmm, `RuboCop::AST::Node::TRUTHY_LITERALS` includes `:array`, and with https://github.com/rubocop-hq/rubocop/blob/master/lib/rubocop/ast/node.rb#L28-L32 and https://github.com/rubocop-hq/rubocop/blob/40d81efc59d28038a3f64b02f7ff6f324d8667ce/lib/rubocop/ast/node.rb#L374-L376 it looks like `#literal?` would always return true for Array nodes.

Darhazer: True story. Didn't expected that it's both coposite_literal and a basic one at the same time :)

bquorning: There are literals, basic literals, and composite literals :-)
`BASIC_LITERALS = (LITERALS - COMPOSITE_LITERALS).freeze`

Darhazer: Yup, got it. Overlooked it and I'm pretty sure I'm overlooking this for array for at least second time. Might clean it up at some point, but it's not a big deal :)

PR #721, opened 2018-12-27
Add cop metadata

Darhazer:
This is the foundation to support the planned RuboCop behaviors (currently only Safe and SafeAutocorrect affect the actual behavior).

As far as I know we don't have any cops that are unsafe by design, so I've just added the versions. Now we can start also mark cops as Safe: false or SafeAutocorrect: false

---

* [x] ~Added tests.~

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -8,7 +8,7 @@ Before submitting the PR make sure the following are checked:
 * [ ] Squashed related commits together.
 * [ ] Added tests.
 * [ ] Updated documentation.
-* [ ] Added an entry to the [changelog](https://github.com/rubocop-hq/rubocop-rspec/blob/master/CHANGELOG.md) if the new code introduces user-observable changes.
+* [ ] Added an entry to the `CHANGELOG.md` if the new code introduces user-observable changes.
 * [ ] The build (`bundle exec rake`) passes (be sure to run this locally, since it may produce updated documentation that you will need to commit).
 
 If you have created a new cop:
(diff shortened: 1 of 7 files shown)


bquorning:
I’m not sure how to review this PR.

Darhazer:
Basically this just follows on what was done in Rubocop, so I guess someone else from @rubocop-hq/rubocop-core can take a look.

Btw, we can use Safe and SafeAutocorrect without importing the versions, they are independent ;)

bbatsov:
Looks good to me! Great work!

bquorning:
Ah, could you rebase please, to get the 2.6 builds running?

bbatsov:
@Darhazer I think forgot to merge this. :-)

dgollahon:
> As far as I know we don't have any cops that are unsafe by design

I thought we had a few but I'd have to re-examine the cops to jog my memory.

dgollahon:
When this lands, we should add `RSpec/DescribedClass` to the list of unsafe autocorrects: https://github.com/rubocop-hq/rubocop-rspec/issues/795.

Also, i think it might be safe if we skipped blocks by default (which , btw, maybe we should consider making the default), but it's interesting because I wonder how many cops might be safe with certain options but not others. I wonder if there should ever be a system for safe under certain conditions (though I suspect that would add too much complexity).

dgollahon:
We should add [RSpec/PredicateMatcher](https://github.com/rubocop-hq/rubocop-rspec/issues/806) to the list of unsafe autocorrects as well.

pirj:
@Darhazer do you need some help on that one?

I've [heard on the grapevine](https://metaredux.com/posts/2020/06/08/the-rubocop-name-drama-redux.html) that RuboCop 1.0 may be released this week! 🎉

bbatsov:
@pirj I'm pretty sure he won't mind help. :-)

pirj:
image 1.0-ready ✔️

Darhazer:
Thank you @pirj 

Now I can't approve it as I'm still the author :joy: 
From now on we should introduce the new cops as pending, maybe even those that are currently not released?

bbatsov:
I think that's a good idea.

pirj:
I'll merge while @bquorning is not looking :D

pirj:
> From now on we should introduce the new cops as pending, maybe even those that are currently not released?

👍 

Or we can do a 2.0 release, bump the dependency to `rubocop` 1.0, and start following the same versioning policy. Probably with a reduced release cycle, say, 6 months.

PR #722, opened 2018-12-28
Test on CI also with Ruby 2.6

bquorning:
Ruby 2.6.0 is released: https://www.ruby-lang.org/en/news/2018/12/25/ruby-2-6-0-released/

---

* [ ] Added tests.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -15,7 +15,7 @@ rubocop: &rubocop
 jobs:
   confirm_config_and_documentation:
     docker:
-      - image: circleci/ruby:2.5
+      - image: circleci/ruby
     steps:
       - checkout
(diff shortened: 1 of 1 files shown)


bquorning:
I’ll mark the `ruby-2.6-rspec` and `ruby-2.6-rubocop` jobs as requiered right before merging.

Darhazer:
Meanwhile, 2.2 reached EOL. If we drop it, we can start using `&.` :)

bquorning:
> If we drop it

I think we’ll keep supporting 2.2 until RuboCop itself drops support for it.

> we can start using `&.` :)

Isn’t there a cop that will prevent us from using the lonely operator? :trollface:

PR #725, opened 2019-01-04
[Fix #661] Add Yield cop

### LINKED ISSUE #661: Suggest using .and_yield
Today I noticed several instances of this pattern in a spec suite:
```rb
allow(foo).to receive(:bar) { |&block| block.call }
```
First I changed it to this:
```rb
allow(foo).to receive(:bar).and_yield
```

Second, I wondered why we don’t have a cop checking these.

Darhazer:
A bit late Christmas present for @bquorning ;)

---

* [x] Added tests.


lib/rubocop/cop/rspec/yield.rb
```diff
@@ -0,0 +1,77 @@
...
+            node.children.map { |child| convert_block_to_yield(child) }.join
+          else
+            convert_block_to_yield(node)
+          end
+        end
+
+        def convert_block_to_yield(node)
+          args = node.arguments
+          replacement = '.and_yield'
+          replacement += "(#{args.map(&:source).join(',')})" if args.any?
```
bquorning: Let’s join with `', '`, shouldn’t we?

Darhazer: Sure. Done,

PR #726, opened 2019-01-09
Add a cop to enforce aggregating examples

pirj:
The "one expectation per example" rule has been relaxed and allows for several expectations to be set in the same example.
https://github.com/rubocop-hq/rspec-style-guide#expectations-per-example

In cases the examples don't have any setup, metadata, or even a docstring, and may be aggregated into one thus saving on sometimes expensive context setup.

The cop still does report the cases when the examples might be aggregated.

Block expectation syntax is deliberately not supported due to:
 - `subject { -> { ... } }` syntax being hard to detect
 - aggregation should use composition with `.and`
 - aggregation of the `not_to` is barely possible when a matcher doesn't
 provide a negated variant
 - aggregation of block syntax with non-block syntax should be in a
 specific order

Known caveats:
The usages if `its` that are testing private methods/readers will result in spec failure. Test only public interface!

Matchers with side effects might affects following expectations when aggregated.

The original idea and initial implementation by @palkan 
https://github.com/palkan/test-prof/blob/master/lib/test_prof/cops/rspec/aggregate_failures.rb

---

* [x] Added tests.


lib/rubocop/cop/rspec/aggregate_examples.rb
```diff
@@ -0,0 +1,294 @@
...
+      #
+      class AggregateExamples < Cop # rubocop:disable Metrics/ClassLength
+        include RangeHelp
+
+        MSG = 'Aggregate with the example above.'.freeze
+        MSG_FOR_EXPECTATIONS_WITH_SIDE_EFFECTS =
+          "#{MSG} IMPORTANT! Pay attention to the expectation order, some " \
+          'of the matchers have side effects.'.freeze
+
+        MATCHERS_WITH_SIDE_EFFECTS = %w[
```
Darhazer: Better move this to config, so users can add their own

lib/rubocop/cop/rspec/aggregate_examples.rb
```diff
@@ -0,0 +1,294 @@
...
+            message = message_for(examples)
+            add_offense(examples[1], location: :expression, message: message)
+          end
+        end
+
+        def autocorrect(example_node)
+          node = example_node.parent
+          examples = aggregateable_examples(node.each_child_node(:block))
+            .select { |child| examples_for_autocorrect?(child) }
+          return if examples.count < 2
```
Darhazer: It's a bit weird that there could be an offense to a node that doesn't have examples for auto-correction.
I'm ok with having this guard though.

lib/rubocop/cop/rspec/aggregate_examples.rb
```diff
@@ -0,0 +1,294 @@
...
+            corrector.replace(range, replacement)
+            examples[1..-1].each { |example| drop_example(corrector, example) }
+          end
+        end
+
+        private
+
+        def_node_matcher :example_group_with_several_examples, <<-PATTERN
+          (block
+            (send { nil? (const nil? :RSpec) } { :describe :context :feature :example_group } ...)
```
Darhazer: You have to use the language module here

lib/rubocop/cop/rspec/aggregate_examples.rb
```diff
@@ -0,0 +1,294 @@
...
+          examples
+            .reject { |example| has_metadata?(example) }
+            .select { |example| example_with_expectations_solely?(example) }
+        end
+
+        def range_for_replace(examples)
+          range = range_by_whole_lines(examples.first.source_range,
+                                       include_final_newline: true)
+          next_range = range_by_whole_lines(examples[1].source_range)
+          range = range.resize(range.length + 1) if adjacent?(range, next_range)
```
Darhazer: Didn't know this method :heart:

lib/rubocop/cop/rspec/aggregate_examples.rb
```diff
@@ -0,0 +1,294 @@
...
+          corrector.remove(aggregated_range)
+        end
+
+        # Extracts and transforms the body.
+        # `its(:something) { is_expected.to ... }` is a special case, since
+        # it's impossible to aggregate its body as is,
+        # it needs to be converted to `expect(subject.something).to ...`
+        # Additionally indents the example code properly.
+        def transform_body(node, base_indent)
+          method, _args, body = *node
```
Darhazer: you can use `node.method_name` and `node.body` and avoid manual deconstruction - block nodes delegate method_name to the send node

lib/rubocop/cop/rspec/aggregate_examples.rb
```diff
@@ -0,0 +1,294 @@
...
+        PATTERN
+
+        def_node_matcher :its_with_array_argument?, <<-PATTERN
+          (block (send nil? :its array) ...)
+        PATTERN
+
+        # Searches for HEREDOC in examples. It can be tricky to aggregate,
+        # especially when interleaved with parenthesis or curly braces.
+        def contains_heredoc?(node)
+          return unless node.is_a? AST::Node
```
Darhazer: you can replace with `node.respond_to(:heredoc?) && node.heredoc?`

Darhazer: Also if you use `each_child_node` instead of `children` you will always have an AST node

lib/rubocop/cop/rspec/aggregate_examples.rb
```diff
@@ -0,0 +1,294 @@
...
+          return unless node && node.begin_type?
+
+          node.children.all? { |statement| single_expectation?(statement) }
+        end
+
+        def_node_matcher :single_expectation?, <<-PATTERN
+          (send #expectation? { :to :to_not :not_to } _)
+        PATTERN
+
+        # Matches example block
```
Darhazer: Use the language module

lib/rubocop/cop/rspec/aggregate_examples.rb
```diff
@@ -0,0 +1,294 @@
...
+      #   # 3. Aggregation of block syntax with non-block syntax should be in a
+      #   # specific order.
+      #
+      class AggregateExamples < Cop # rubocop:disable Metrics/ClassLength
+        include RangeHelp
+
+        MSG = 'Aggregate with the example above.'.freeze
+        MSG_FOR_EXPECTATIONS_WITH_SIDE_EFFECTS =
+          "#{MSG} IMPORTANT! Pay attention to the expectation order, some " \
+          'of the matchers have side effects.'.freeze
```
pirj: Double freeze.

lib/rubocop/cop/rspec/aggregate_examples.rb
pirj: Agree. Also, this setting may eventually be also used by another cop that detects existing examples that have expectations with side effects.

lib/rubocop/cop/rspec/aggregate_examples.rb
pirj: An example of this is:
```
it('be this') { is_expected.to be_this }
it('be that') { is_expected.to be_that }
```
it's not automatically aggregateable, since the docstrings would be lost.

pirj:
> This can easily lead to breaking the max example length or max expectations count. Should the cop be aware of those?
Not sure what is considered "too many", and if there are more expectations than the limitation setting, how to divide them across the examples.
The [referenced betterspecs](http://betterspecs.org/#single) says:
> it's fine to specify more than one isolated behavior.
and does not dictate any limitation on the number of the expectations.
I would let this up to `Metrics/BlockLength`.

> Maybe the cop should wrap the example in aggregate_failures
Agree, it completely slipped my mind that it's not the default.

palkan:
@pirj 

>  if we would add metadata to the example

Btw, why did you decide not to add it (as in the original implementation)?

lib/rubocop/cop/rspec/aggregate_examples.rb
pirj: Actually the cop complains anyway.

palkan:
> On the project where this cop is mostly used, aggregate_failures is on globally with a few exceptions.

What do you think about adding a configuration parameter to add the tag automatically? 

I'm not sure that enabling `aggregate_failures` globally fits everyone; sometimes it's better to fail fast)

So, there could be two strategies: fail-fast (default RSpec behaviour) + explicit `aggregate_failures` and _global_ aggregation + in-place disabling. And for the first one it would be great to have an ability to add the tag automatically.

pirj:
Added aggregation of the examples with identical metadata.

@palkan Loving the configuration idea, will implement shortly.

pirj:
@Darhazer @palkan 
Added configuration option to add `:aggregate_failures` to metadata of the aggregated example.
Please take another look. Appreciate if you run it against the codebase of your choice.

pirj:
@Darhazer @palkan ping.

lib/rubocop/cop/rspec/aggregate_examples.rb
```diff
@@ -0,0 +1,377 @@
...
+      #       expect(number).to be_odd
+      #     end
+      #
+      #     it { is_expected.to be_prime }
+      #   end
+      #
+      #   # good
+      #   describe do
+      #     specify do
+      #       expect(number).to be_positive
```
Darhazer: even better
```
expect(number).to be_positive
  .and be_odd
  .and be_prime
```

config/default.yml
```diff
@@ -10,6 +10,23 @@ AllCops:
...
 
+RSpec/AggregateExamples:
+  Description: Checks if example groups contain more than one aggregateable example.
+  Enabled: false
+  StyleGuide: http://www.rubydoc.info/gems/rubocop-rspec/RuboCop/Cop/RSpec/AggregateExamples
+  EnforcedStyle: skip_aggregate_failures_metadata
+  SupportedStyles:
+  - skip_aggregate_failures_metadata
+  - add_aggregate_failures_metadata
+  MatchersWithSideEffects:
```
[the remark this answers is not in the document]
pirj: Indeed!
Actually, I was thinking of fixing those matchers in shoulda-matchers instead.

lib/rubocop/cop/rspec/aggregate_examples.rb
```diff
@@ -0,0 +1,377 @@
...
+          metadata = example_parameters(example) || []
+
+          symbols = metadata_symbols_without_aggregate_failures(metadata)
+          pairs = metadata_pairs_without_aggegate_failures(metadata)
+          symbols << Object.new if aggregate_failures_disabled(pairs)
+
+          [*symbols, pairs].flatten.compact
+        end
+
+        def_node_matcher :example_parameters, <<-PATTERN
```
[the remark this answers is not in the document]
pirj: Thanks, worked for me, except that there's a special case now: `its(:one) { ... }` is detected as an example with metadata, while it's actually not. 👍

lib/rubocop/cop/rspec/aggregate_examples.rb
```diff
@@ -0,0 +1,377 @@
...
+        PATTERN
+
+        def_node_matcher :its_with_array_argument?, <<-PATTERN
+          (block (send nil? :its array) ...)
+        PATTERN
+
+        # Searches for HEREDOC in examples. It can be tricky to aggregate,
+        # especially when interleaved with parenthesis or curly braces.
+        def contains_heredoc?(node)
+          return true if node.respond_to?(:heredoc) && node.heredoc?
```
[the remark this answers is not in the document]
pirj: Yes, it's weird. Replaced the whole statement with `return false` and it turns out that the whole `'detects, but does not autocorrect'` example group was not doing what I expected it to due to the fact that I did not pass a fake filename to `autocorrect_source` 😱

lib/rubocop/cop/rspec/aggregate_examples.rb
```diff
@@ -0,0 +1,377 @@
...
+
+          node.each_child_node.any? { |child| contains_heredoc?(child) }
```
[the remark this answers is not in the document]
pirj: TIL `node.each_descendant` that also walks down the tree.

lib/rubocop/cop/rspec/aggregate_examples.rb
```diff
@@ -0,0 +1,377 @@
...
+
+        # Matchers examples with:
+        # - expectation statements exclusively
+        # - no metadata (e.g. `freeze: :today`)
+        # - no title (e.g. `it('jumps over the lazy dog')`)
+        # and skips `its` with an array argument due to ambiguous conversion
+        #   e.g. the SUT can be an object (`expect(object.property)`)
+        #   or a hash/array (`expect(hash['property'])`)
+        # and also skips matchers with known side-effects
+        def_node_matcher :example_for_autocorrect?, <<-PATTERN
```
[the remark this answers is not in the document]
pirj: Tried both changing `example_for_autocorrect?` to a method:
```
        def example_for_autocorrect?(node)
          return if example_has_title?(node) ||
            its_with_array_argument?(node) ||
            contains_heredoc?(node) ||
            example_with_side_effects?(node)
          example_with_expectations_only?(node)
        end
```
and extracting to a select/reject
```
          examples_in_group = examples_for_autocorrect(
            example_node.parent.each_child_node(:block)
          )
...
        def examples_for_autocorrect(examples)
          examples
            .select(&method(:example_with_expectations_only?))
            .reject(&method(:example_has_title?))
            .reject(&method(:its_with_array_argument?))
            .reject(&method(:contains_heredoc?))
            .reject(&method(:example_with_side_effects?))
        end
```

but I think both options are less concise.

lib/rubocop/cop/rspec/aggregate_examples.rb
pirj: Definitely!
I didn't go for this since some of the matchers can't be compounded to a single expectation, e.g.
```
is_expected.to validate_length_of(:about).is_at_most(5000)
  .and ...
...
NoMethodError: undefined method `and' for #<Shoulda::Matchers::ActiveModel::ValidateLengthOfMatcher:0x007fd71d699078>
```

I'm not sure why it is so, probably they were never supposed to be used like that due to side effects.
Pretty sure there are others like those `validate_...`, I don't have a list.

Probably it would be a good idea to add a `CompoundExpectations` cop eventually.

pirj:
@Darhazer I believe I've addressed your concerns, appreciate another round of code review.
cc @bquorning @palkan

spec/rubocop/cop/rspec/aggregate_examples_spec.rb
```diff
@@ -0,0 +1,790 @@
...
+      expect_offense(offensive_source)
+    end
+
+    it 'does not autorcorrect source code' do
+      bad_code_without_message =
+        offensive_source.lines.delete_if { |line| line =~ /\A\s*\^/ }.join
+      expect(autocorrect_source(bad_code_without_message, 'spec/foo_spec.rb'))
+        .to eq(bad_code_without_message)
+    end
+  end
```
pirj: Should I extract those two `shared_example`'s somewhere? Will they be useful elsewhere?

spec/rubocop/cop/rspec/aggregate_examples_spec.rb
```diff
@@ -0,0 +1,790 @@
...
+        it { is_expected.to be_cool }
+        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ Aggregate with the example above.
+        its(:another) { is_expected.to be(nil) }
+      end
+    RUBY
+
+    good_source = <<-RUBY
+      describe 'aggregations' do
+        specify do
+          expect(subject.one).to be(true)
```
pirj: This doesn't seem to respect named subject. Is it ok to rely on the other cops for that?

spec/rubocop/cop/rspec/aggregate_examples_spec.rb
```diff
@@ -0,0 +1,790 @@
...
+      expect_no_offenses(good_source)
+    end
+
+    it 'detects an offense in offensive source code' do
+      expect_offense(offensive_source)
+    end
+
+    bad_code_without_message =
+      offensive_source.lines.delete_if { |line| line =~ /\A\s*\^/ }.join
+    include_examples 'autocorrect', bad_code_without_message, good_source
```
pirj: The output of this included example is excessive:
image

Is that ok?
I've checked, a similar output comes from e.g. `rspec spec/rubocop/cop/rspec/predicate_matcher_spec.rb:342`.

pirj:
Bump.

spec/rubocop/cop/rspec/aggregate_examples_spec.rb
bquorning: I think it’s fine. Anyway it’s an unrelated issue that should be fixed in a separate PR.

pirj:
@bquorning Thanks for the review.

Actually, explicit subject [is supported as well]
(https://github.com/rubocop-hq/rubocop-rspec/pull/726/files#diff-65c1681945efcb2fb8eb9ebe1afd5bd6R325) ([spec](https://github.com/rubocop-hq/rubocop-rspec/pull/726/files#diff-a3540cf8ed264164ae01ab2cf7789997R549)):
```
      describe do
        it { expect(division.result).to eq(5) }
        it { expect(division.modulo).to eq(3) }
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ Aggregate with the example above.
      end
```

I used implicit subject in examples to avoid additional `subject(...) { ... }` line.

I completely get your point, unit tests are already fast enough usually to aggregate them, however sometimes aggregation might be used to reduce the size of the spec file.

The usual suspect of this cop are integration tests, if the setup is shared across several examples, and it's only the expectations differ, those can actually be aggregated. In the codebase I'm mostly working on it's was a rather frequent case.

This started off relaxing the "one expectation per example" rule, as it wasn't that strict in `betterspecs` compared to `rspec-style-guide`, then active development of this cop, and I believe is helping to reduce the effort to change the code, since if some part of the functionality is extracted, fewer examples should be changed, and smaller the chances to leave some legacy examples in the code.

I feel responsible for it, and Maxim is familiar with the code as well, as he was doing a number of internal code reviews.

I really hope that the size of the cop will be reduced eventually as its node matchers are extracted when the other cops find those parts useful.

Let's pair and I will walk you through the code if you like.

I'm not sure what is the best approach to highlight the disabled cops, just mention them in the docs? Inject into project's `.rubocop-todo.yml` somehow? The enabled/disabled distinction looks way too strict, I would rather go for recommended/optional, and would ask users to explicitly disable even those that are optional.

pirj:
@bquorning I think I understand what you mean concerning the disabling by default.
There was a convention regarding cop versioning and enabling previously introduced disabled cops in next minor versions.

There are different styles of tests, and for some, it might not be a good fit, as it pivots the way the tests are written. However, while applying it to a big codebase I didn't see any really big issues with it (apart from numerous that I have addressed already, e.g. deliberately skipped block syntax and matchers with side effects).

Amongst codebases I've seen, there were always a good amount of disabled cops.

On the other hand, `AggregateExamples` is mutually exclusive with `MultipleExpectations`. However, it might be perceived as a single cop with two styles (like `MessageExpectation` that has two opposite style options), but split. I don't consider joining the two together as a good option though, as the former has grown huge already.

That being said, do you think this cop should remain disabled forever, or until the next minor `rubocop-rspec` release?

pirj:
Updated to use the `Language::RUNNERS` from `master`.

spec/rubocop/cop/rspec/aggregate_examples_spec.rb
bquorning: These shared examples could be rewritten to use `expect_correction`, e.g.
```rb
shared_examples 'detects and autocorrects' do |offensive_source, good_source|
  it 'does not detect an offense in good_source code' do
    expect_no_offenses(good_source)
  end

  it 'detects an offense in offensive source code' do
    expect_offense(offensive_source)
    expect_correction(good_source)
  end
end
```
But it’s probably better to do in a separate follow-up pull request.

spec/rubocop/cop/rspec/aggregate_examples_spec.rb
bquorning: `expect_no_corrections` will help with this:
```rb
  shared_examples 'detects, but does not autocorrect' do |offensive_source|
    it 'detects an offense in offensive source code code' do
      expect_offense(offensive_source)
      expect_no_corrections
    end
  end
```

spec/rubocop/cop/rspec/aggregate_examples_spec.rb
```diff
@@ -0,0 +1,790 @@
...
+  it_behaves_like 'detects and autocorrects in example group',
+                  :example_group
+
+  context 'with `its`' do
+    offensive_source = <<-RUBY
+      describe 'aggregations' do
+        its(:one) { is_expected.to be(true) }
+        it { is_expected.to be_cool }
+        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ Aggregate with the example above.
+        its(:another) { is_expected.to be(nil) }
```
bquorning: I notice that during autocorrect, this line is also treated as a cop offense. So, shouldn’t it be marked with an offense message here (`^^^^ Aggregate`…)?

spec/rubocop/cop/rspec/aggregate_examples_spec.rb
```diff
@@ -0,0 +1,790 @@
+RSpec.describe RuboCop::Cop::RSpec::AggregateExamples, :config do
+  subject(:cop) { described_class.new(config) }
+
+  let(:cop_config) { {} }
+
+  shared_examples 'detects and autocorrects' do |offensive_source, good_source|
+    it 'does not detect an offense in good_source code' do
+      expect_no_offenses(good_source)
```
bquorning: I see how this test has been useful when developing autocorrection, but I think it is not necessary after the cop is completed.

spec/rubocop/cop/rspec/aggregate_examples_spec.rb
```diff
@@ -0,0 +1,790 @@
...
+          foo
+          is_expected.to be_foo
+        end
+        it { is_expected.to be_second }
+        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ Aggregate with the example above.
+        specify do
+          bar
+          is_expected.to be_bar
+        end
+        it { is_expected.to be_third }
```
bquorning: This line is missing an offense message.

spec/rubocop/cop/rspec/aggregate_examples_spec.rb
```diff
@@ -0,0 +1,790 @@
...
+  context 'with scattered aggregateable examples' do
+    offensive_source = <<-RUBY
+      describe 'aggregations' do
+        it { is_expected.to be_first }
+        specify do
+          foo
+          is_expected.to be_foo
+        end
+        it { is_expected.to be_second }
+        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ Aggregate with the example above.
```
bquorning: The offense message is misleading here. The example above is the `specify do` block with cannot be aggregated.

spec/rubocop/cop/rspec/aggregate_examples_spec.rb
```diff
@@ -0,0 +1,790 @@
...
+    it_behaves_like 'detects, but does not autocorrect', offensive_source
+  end
+
+  context 'with hash metadata' do
+    offensive_source = <<-RUBY
+      describe 'aggregations' do
+        it { is_expected.to be_mild }
+        it(freeze: -30) { is_expected.to be_cold }
+        it(aggregate_failures: true) { is_expected.to be_warm }
+        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ Aggregate with the example above.
```
bquorning: Here too, the offense messages are a bit misleading.

spec/rubocop/cop/rspec/aggregate_examples_spec.rb
```diff
@@ -0,0 +1,790 @@
...
+      end
+    RUBY
+
+    it_behaves_like 'detects and autocorrects', offensive_source, good_source
+  end
+
+  context 'with validation actions with side effects' do
+    context 'without configuration' do
+      let(:cop_config) do
+        { 'MatchersWithSideEffects' => [] }
```
bquorning: How is this different from the cop_config in the outer example group (`let(:cop_config) { {} }`)? Testing “validation actions with side effects” isn’t a special case in my opinion. It’s when we configure the cop with a list of “matchers with side effects” that we need a new example group.

spec/rubocop/cop/rspec/aggregate_examples_spec.rb
```diff
@@ -0,0 +1,790 @@
...
+          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ Aggregate with the example above.
+          context 'with side effects' do
+            it { is_expected.to validate_presence_of(:comment) }
+          end
+        end
+      end
+    RUBY
+
+    it 'detects an offense in offensive_source code' do
+      expect_offense(offensive_source)
```
bquorning: I think the autocorrection result would be interesting in this example. I’ve seen above how only the first offensive line would be marked with `^^^`, so I wonder if the `validate_presence_of` line is actually considered offensive or not.

lib/rubocop/cop/rspec/aggregate_examples.rb
```diff
@@ -0,0 +1,374 @@
...
+      #   end
+      #
+      # @example AddAggregateFailuresMetadata: true
+      #
+      #   specify(:aggregate_failures) do
+      #     expect(number).to be_positive
+      #     expect(number).to be_odd
+      #   end
+      #
+      class AggregateExamples < Cop # rubocop:disable Metrics/ClassLength
```
bquorning: It would be nice to (perhaps in a later PR) enable `Metrics/ClassLength` on this class.

bquorning: Actually (as mentioned earlier) I’m having a really hard time reviewing this class. Which suggests that it will be hard to maintain…

Would it be possible to extract some logic into separate classes (just keep the in the same time, namespaced within `AggregateExamples` would be fine. I’d suggest grouping some of the autocorrect code (`range_for_replace`, `aggregated_example`, `drop_example`) in a class, or the “metadata” methods (`metadata_without_aggregate_failures`, `metadata_symbols_without_aggregate_failures`, `metadata_pairs_without_aggegate_failures`, `aggregate_failures_disabled`)?

bquorning:
I’ve read *most* of the code now, but as mentioned above I’m still struggling with the actual cop implementation. It may well be that *I* just need to pull myself together and focus :-)

spec/rubocop/cop/rspec/aggregate_examples_spec.rb
pirj: Awesome, thank you! 👍

config/default.yml
```diff
@@ -10,6 +10,20 @@ AllCops:
...
+  StyleGuide: http://www.rubydoc.info/gems/rubocop-rspec/RuboCop/Cop/RSpec/AggregateExamples
+  AddAggregateFailuresMetadata: false
+  MatchersWithSideEffects:
+  - allow_value
+  - allow_values
+  - validate_presence_of
+  - validate_absence_of
+  - validate_length_of
+  - validate_inclusion_of
+  - validates_exclusion_of
```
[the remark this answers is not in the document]
pirj: Agree, but honestly, I'm not sure how to make this safe and not tiresome to configure and by default at the same time.

A couple of arguments for keeping this in the default config:
 - this is a generic cop, but it also supports Rails. It could probably inject those shoulda-matchers to config, but that might not be desirable, say if side effects are eventually fixed in some version of it. Don't see a good solution to this
 - shoulda-matchers also [support non-Rails](https://github.com/thoughtbot/shoulda-matchers#non-rails-apps) that use ActiveModel or ActiveRecord
 - we already have `Rails/HttpStatus` in our default config, which is for one of the few `rspec-rails` matchers. I don't think there's a critical mass yet to extract those to a separate gem

lib/rubocop/cop/rspec/aggregate_examples.rb
pirj: Now it's been almost a month since I last touched this cop, and I feel your pain when I look at the code.
Will structure to it, and will add clarifications to the code that is not obvious.
Thanks for the suggestion.

bquorning: >  (just keep the in the same time, namespaced within `AggregateExamples` would be fine.

That’s the kind of language that comes from doing code review on my phone… 😆 I think I meant to say “the new classes can just be defined inside the `AggregateExamples` namespace.”

spec/rubocop/cop/rspec/aggregate_examples_spec.rb
pirj: Good point, I mistakenly thought that only the first (of what?) indicated, now I see that typically all are indicated as offences:
```
          should be_good
          ^^^^^^^^^^^^^^ Don't use implicit subject.
          should_not be_bad
          ^^^^^^^^^^^^^^^^^ Don't use implicit subject.
```
👍

spec/rubocop/cop/rspec/aggregate_examples_spec.rb
pirj: Good catch. Will add a line number instead of "above".

spec/rubocop/cop/rspec/aggregate_examples_spec.rb
```diff
@@ -0,0 +1,790 @@
...
+            is_expected.to validate_absence_of(:comment)
+            is_expected.to validate_presence_of(:description)
+          end
+        end
+      RUBY
+
+      it_behaves_like 'detects and autocorrects', offensive_source, good_source
+    end
+
+    context 'with `is_expected`' do
```
pirj: This context is misleading, it should be "with the default configuration".

spec/rubocop/cop/rspec/aggregate_examples_spec.rb
pirj: This context meant to show the behaviour when the default `MatchersWithSideEffects` option is empty.
See below how a wrong context name disconnected the balancing context.

spec/rubocop/cop/rspec/aggregate_examples_spec.rb
```diff
@@ -0,0 +1,790 @@
...
+          it { is_expected.to validate_absence_of(:comment) }
+          it { is_expected.to validate_presence_of(:description) }
+          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ Aggregate with the example above. IMPORTANT! Pay attention to the expectation order, some of the matchers have side effects.
+        end
+      RUBY
+
+      it_behaves_like 'detects, but does not autocorrect', offensive_source
+    end
+
+    context 'with `expect(something)`' do
```
pirj: Those three contexts with different syntaxes should be grouped into one.

spec/rubocop/cop/rspec/aggregate_examples_spec.rb
pirj: Learned about those helpers from your comment. Got rid of those shared examples. The change made the spec slightly less compact (+55 lines), but I think more readable.

pirj:
Soon after its anniversary 50th commit, the cop is ready for another review @Darhazer, @bquorning.

Please find the reasoning behind some design decisions that I've made below.

## What's new

- significantly improved `its` support;
- fixed `--auto-correct` showing all detected offenses as corrected;
- fixed offense detection of matchers with side effects with qualifiers;
- removed redundant spec coverage;
- divided the code into logical parts (more on that below), that allowed removing disabling of `Metrics/ClassLength` on the cop;
- extracted specs for `its` and matchers with side effects to dedicated spec files;
- ... and much more!

## Why not object-oriented approach to divide the cop?

I tried to cleanly separate parts of the cop into self-contained objects and miserably failed.

### Extraction to self-contained functional modules

The first approach I made was to extract logically related node matchers into modules. However, it turns out it's impossible to use other class's methods in node pattern, e.g.:

    module Example
      module_function

      def_node_matcher :with_expectations_only, '...'
    end

    def_node_matcher :example_for_autocorrect?, <<-PATTERN
      [
        #Example.with_expectations_only?
        ...
      ]
    PATTERN

So it comes with a burden of defining delegators, which made things look unreadable.

### Extraction to self-contained classes

Another approach, which worked, was to extract all aggregation related logic to `Aggregator` class, that an `autocorrect` would pass up and RuboCop would call `call` upon.
Oddly, `range_by_whole_lines` that I make use of [expects `@processed_source`](https://github.com/rubocop-hq/rubocop/blob/c40e68381b5d11758e093ac9c636cdca585275dd/lib/rubocop/cop/mixin/range_help.rb#L65) to be accessible, so I had to pass it as an argument to `Aggregator.new` as well.
Furthermore, it needed `cop_config`, so I had to pass it as well. I ended up with:

        def autocorrect(example_node)
          clusters = example_clusters_for_autocorrect(example_node)
          return if clusters.empty?

          Aggregator.new(clusters, @processed_source, cop_config)
        end

        class Aggregator
          prepend Its
          include Metadata
          include RangeHelp

          attr_reader :cop_config, :clusters

          def initialize(clusters, processed_source, cop_config)
            @clusters = clusters
            @processed_source = processed_source
            @cop_config = cop_config
          end

          def call(corrector)
            clusters.each do |metadata, examples|
              range = range_for_replace(examples)
              replacement = aggregated_example(examples, metadata)
              corrector.replace(range, replacement)
              examples[1..-1].map { |example| drop_example(corrector, example) }
            end
          end

          # Plus a bunch of private methods
        end

It worked, but I wouldn't say it's less tangled.

### Includable modules

I ended up with a design that I initially considered an interim one, modules all included into the cop.
Using prepending I achieved functional inheritance, that helped to hide the complexity behind some methods, specifically `new_body` and `message_for` being extended in `MatchersWithSideEffects` and `Its`.

Another sudden benefit was that documentation laid out quite well across those files.

## How the documentation for the cop's is split across several files?

The intention was to spread the documentation over source files where it would correspond to the source code, e.g. core usage, usage of `its`, and usage with matchers with side effects.
Initially I was going to put it all on the class level, but it hit me by surprise that `yard` overrides the docs when is defined in several files for the same class ([see this issue](https://github.com/lsegal/yard/issues/1173)), it's impossible to spread the docs to several files so that they are compiled into one document (manual).
So I added support for gathering cop docs from modules in cop's namespace.


Please let me know what you think.

pirj:
Updated with the latest master.
Please take a look.
I'm all for addressing your concerns, including making required systematic changes e.g. to how we recommend new cops.

bolshakov:
@pirj could you explain how it's expected to aggregate these two examples:

```ruby
subject { -> { do_side_effect } }

it 'has effect on this' do
  is_expected.to change {
    check_effect_this_way
  }.from(42).to(43)
end

it 'has effect on that' do
  is_expected.to change {
    check_effect_that_way
  }.from('foo').to('bar')
end
```

bolshakov:
Consider I managed to aggregate two examples into one. Now I want to test the second one in isolation (because it has a bug). How it's expected to be done?

For me it's the same ani-pattern as iterationg overt tests like this 

```ruby
['foo', 'bar', 'baz'].each do |subject|
  test_it(subject)
end
```

because you can't run single example in isolation

pirj:
@bolshakov Thanks for asking. Yes, that one would be easy:

```ruby
def has_effect_on_this
  change {
    check_effect_this_way
  }.from(42).to(43)
end

def has_effect_on_that
  change {
    check_effect_that_way
  }.from('foo').to('bar')
end

it 'does this and that' do
  expect { do_side_effect }
    .to has_effect_on_this
    .and has_effect_on_that
end
```

You are testing side effects of calling `do_side_effects`. What do you mean by testing it in isolation, what would be the reason for that isolation? I'm not sure I correctly understand your question.
With `aggregate_failures` turned on, you will see all the expectation failures, even if there are several of them in the same example.

I'm not going to reason about to SRP, since the tests are usually just indicators of the production code flaws, and if the code does this and that doesn't necessarily mean that testing those two side effects is a benefit.

You might be interested in an open `rspec-style-guide` ticket to [recommend against implicit block syntax](https://github.com/rubocop-hq/rspec-style-guide/issues/76).

The reasoning behind recommending against using iterators [is different](https://github.com/rubocop-hq/rspec-style-guide#it-in-iterators) than lack of ability to test the expectations separately, it's:

> When another developer adds a feature to one of the items in the iteration, they must then break it out into a separate test - they are forced to edit code that has nothing to do with their pull request.

lib/rubocop/cop/rspec/support/aggregate_examples/matchers_with_side_effects.rb
```diff
@@ -0,0 +1,80 @@
...
+        #   # bad, but isn't automatically correctable
+        #   describe do
+        #     it { is_expected.to validate_presence_of(:comment) }
+        #     it { is_expected.to be_valid }
+        #   end
+        #
+        # @internal
+        #   Support for taking special care of the matchers that have side
+        #   effects, i.e. leave the subject in a modified stte.
+        module MatchersWithSideEffects
```
pirj: I'm trying to reduce the scope of this pull request, and one of the things I plan to extract is a `MatcherWithSideEffects` cop that would prevent from using the matchers that leave the dirty subject after them together with the other matchers.

However, if I would remove this part from `AggregateExamples`, and it will suggest combining several examples into one, but `MatcherWithSideEffects` will suggest separating them.
I don't see that those two cops are in any way mutually exclusive in general, but this how would `AggregateExamples` know which examples to ignore?

@dgollahon You had a perspective on cop responsibilities separation, what do you think about this case?

pirj:
> The main reason someone may use such a cop is to improve spec performance. This is good to be noted in the docs.


> This can easily lead to breaking the max example length or max expectations count. Should the cop be aware of those? Maybe the cop should wrap the example in aggregate_failures
I believe those two are mutually exclusive cops.

pirj:
Just in case if someone would like to use this, the cop has found a place in [`test-prof`](https://github.com/palkan/test-prof/pull/170).

PR #729, opened 2019-01-24
[Fix #728] Fix false negatives in Focus cop

### LINKED ISSUE #728: RSpec/Focus not working when I use RSpec.describe
RSpec/Focus not working against `RSpec.describe` 
Is it correct?

```ruby
# RSpec/Focus working
describe SomeClass, focus: true do
end

# RSpec/Focus not working
RSpec.describe SomeClass, focus: true do
end
```

`rubocop-0.62.0`
`rubocop-rspec-1.31.0`

Darhazer:
A while ago we made sure that our matchers include both explicit `RSpec` receiver and implicit (`nil?`) one. There were a couple of cops however that use just `node_pattern_union`, skipping the language module, and using only implicit receiver.

---

* [ ] Added tests.

### DIFF
--- a/lib/rubocop/cop/rspec/focus.rb
+++ b/lib/rubocop/cop/rspec/focus.rb
@@ -36,8 +36,8 @@ module RuboCop
         FOCUS_TRUE   = s(:pair, FOCUS_SYMBOL, s(:true))
 
         def_node_matcher :metadata, <<-PATTERN
-          {(send nil? #{FOCUSABLE_SELECTORS} ... (hash $...))
-           (send nil? #{FOCUSABLE_SELECTORS} $...)}
+          {(send {(const nil? :RSpec) nil?} #{FOCUSABLE_SELECTORS} ... (hash $...))
+           (send {(const nil? :RSpec) nil?} #{FOCUSABLE_SELECTORS} $...)}
         PATTERN
 
         def_node_matcher :focused_block?, focused.send_pattern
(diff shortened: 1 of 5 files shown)


lib/rubocop/cop/rspec/focus.rb
```diff
@@ -36,8 +36,8 @@ class Focus < Cop
         FOCUS_TRUE   = s(:pair, FOCUS_SYMBOL, s(:true))
 
         def_node_matcher :metadata, <<-PATTERN
-          {(send nil? #{FOCUSABLE_SELECTORS} ... (hash $...))
-           (send nil? #{FOCUSABLE_SELECTORS} $...)}
+          {(send {(const nil? :RSpec) nil?} #{FOCUSABLE_SELECTORS} ... (hash $...))
+           (send {(const nil? :RSpec) nil?} #{FOCUSABLE_SELECTORS} $...)}
```
bquorning: Should we extract `{(const nil? :RSpec) nil?}` to the `Language` module?

Darhazer:
> I think one spec is failing after rubocop-hq/rubocop#6170 was merged into RuboCop 0.63.1.

Actually, a failing auto-correct for which we don't have a spec. Seems that now each `expect_offence` also run auto-correction which helps catch at least when auto-correct would crash. So now I have 2 more tasks: fix auto-correction and improve the `{(const nil? :RSpec) nil?}` usage :joy:

bquorning:
> So now I have 2 more tasks

Or, you can delegate :-)

Darhazer:
Build passes. Extraction of ` {(const nil? :RSpec) nil?}` will be a follow-up

Issue #727, opened 2019-01-23
Add cop which suggests using be_an_instance_of


dgollahon:
I recently wrote something like this in a spec:

```ruby
expect(foo.class).to be(Foo)
```

and then it occurred to me that I could use

```ruby
expect(foo).to be_an_instance_of(Foo)
```

If it makes sense to others, I'd be happy to create a BeAnInstanceOf cop to enforce this.

I'm not sure if it fits in this cop idea or should be split into another one, but i also think it would be nice to have something that enforces a preferred alias for `be_an_instance_of` v.s. `be_instance_of` and, similarly, `be_a/be_an` v.s. `be_kind_of` or `be_a_kind_of`.

PR #732, opened 2019-01-27
Add RSpec/Dialect

gsamokovarov:
In @receipt-bank we prefer to only use describe blocks. This is a cop
that can help us enforce such a style automatically. It's an uncommon
style chose that trips most of our new colleagues.


lib/rubocop/cop/rspec/no_context.rb
```diff
@@ -0,0 +1,39 @@
...
+      #
+      #   # good
+      #   describe 'display name presence' do
+      #     # ...
+      #   end
+      class NoContext < Cop
+        MSG = "Don't use context blocks.".freeze
+
+        def_node_matcher :context_block, <<-PATTERN
+          (block (send _ { :context } _) ...)
```
Darhazer: you don't need the curly braces around the method name, as they act as an OR but you have only one option

Darhazer: It's guaranteed to produce false positives for methods that are named context and accept a block. That's why we check for either `nil?` or `(const nil? RSpec)` receiver

gsamokovarov:
@Darhazer it invites nesting and we prefer to have more shallow suites. We don't even prefer multiple `describe`s unless we have a good reason to do it.

@pocke you have a point, I'll take another stab for the name of the cop.

Darhazer:
@gsamokovarov then we have the `NestedGroups` cop.

gsamokovarov:
`NestedGroups` may help us enforce clear boundry on exactly how much nesting we want. :+1:  Still, we'd need to avoid the `context`, because we also like to describe class responsibilities rather than methods, for example, and `describe` fits better with our vocabulary.

gsamokovarov:
Hey, sorry for the ping. I have rebased and used `RSPEC` constant in the matcher definition.

Darhazer:
I still don't see how this cop helps. 
`ContextWording` requires `when`, `with` or `without` for context description. It should make it easy to decide whether to use context or not.

`describe` also can be nested. I agree that it's more natural to nest multiple conditions, and having conditions invites using context, but won't this cop just force the person to change context to describe, leaving the nesting? I don't think context invites the nesting, but checking combinations of conditions do.

```
describe 'display name presence' do
  describe 'when it\'s present' do
    describe 'and it\'s valid' do
     # ...
    end
  end

  describe 'when it\'s not' do
     # ...
  end
end
```


On the other hand, `NestedGroups` configured with a low value can help. A cop for maximum number of example groups can help you too.

> and `describe` fits better with our vocabulary.

In this case a better solution wll be to have a cop with `PreferredMethods` configuration, like Rubocop's `CollectionMethods`. It will be configurable and flexible for the user and will not introduce a cop per every language construct. Can't think of a good name though

gsamokovarov:
To your suggestion, I have turned the cop around and made it like Rubocop's `Style/CollectionMethods`. Named it `ExampleGroupMethods` as it deals only with example groups defining methods for now, but it can be more generic actually, don't know if it's needed, though.

config/default.yml
```diff
@@ -109,6 +109,22 @@ RSpec/EmptyLineAfterSubject:
   Enabled: true
   StyleGuide: http://www.rubydoc.info/gems/rubocop-rspec/RuboCop/Cop/RSpec/EmptyLineAfterSubject
 
+RSpec/ExampleGroupMethods:
+  Description: This cop enforces custom example group definitions methods.
+  Enabled: false
+  PreferredMethods:
+    describe: describe
```
Darhazer: I guess there is no need to set a preference for ones that remain the same.
You can set some default, like using `describe` instead of `example_group` 
Such a cop would be useful for examples as well (using `it` instead of an `example`, for example.).

config/default.yml
```diff
@@ -109,6 +109,22 @@ RSpec/EmptyLineAfterSubject:
   Enabled: true
   StyleGuide: http://www.rubydoc.info/gems/rubocop-rspec/RuboCop/Cop/RSpec/EmptyLineAfterSubject
 
+RSpec/ExampleGroupMethods:
+  Description: This cop enforces custom example group definitions methods.
```
Darhazer: Well, maybe not custom, but just consistent.

config/default.yml
gsamokovarov: I can extend it to catch all of the RSpec methods. Call it `RSpec/RSpecPreferredMethods` (still may need a better name) and use the `Language::ALL.block_pattern` that will allow all the group/examples/hooks/subject/etc definition methods. Then I can set the proposed defaults.

Darhazer: @bquorning WDYT? One generic cop, separate cops per group?

dgollahon: We already have several cops that perform a similar function of enforcing a particular alias. (i.e. [RSpec/ItBehavesLike](https://github.com/rubocop-hq/rubocop-rspec/blob/master/lib/rubocop/cop/rspec/it_behaves_like.rb)). If we want to have a PreferedMethods / PreferedAliases / WhateverWeCallThisThing cop I wonder if we should (eventually?) consolidate all the existing cops that perform the same kind of function. This could also include more things like [RSpec/NotToNot](https://github.com/rubocop-hq/rubocop-rspec/blob/master/lib/rubocop/cop/rspec/not_to_not.rb) if we wanted it to.

dgollahon: There are a large number of RSpec aliases we haven't covered yet with existing cops and it's not always clear if they merit their own cops. For example, #727 mentions several.

gsamokovarov:
Sorry, it took me a while to get back to this. I have generalized this a bit more to support all of the RSpec constructs and called it `RSpec/Dialect` while at it. I actually like the name, @Darhazer, please tell me what you think.

config/default.yml
gsamokovarov: Decided to generalize it and extend it to all them RSpec methods. Called it `RSpec/Dialect` as you can choose your own dialect, in a way. Tell me what you folks think.

lib/rubocop/cop/rspec/dialect.rb
```diff
@@ -0,0 +1,60 @@
+# frozen_string_literal: true
+
+module RuboCop
+  module Cop
+    module RSpec
+      # This cop enforces custom RSpec dialects.
+      #
+      # By default all of the RSpec methods and aliases are allowed allowed. By
```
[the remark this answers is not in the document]
gsamokovarov: Thanks!

gsamokovarov:
Thanks, @bquorning for the review!

I have listed the allowed methods in the documentation for the cop and send all the aliases through the `send` node. This will produce offenses for variables named like `context`, `describe`, `it` and the likes, but I hope those cases will be rare and I can handle them as they arrive, as long as the cop is used. 😅

Can you take another look?

spec/rubocop/cop/rspec/dialect_spec.rb
```diff
@@ -0,0 +1,77 @@
...
+      end
+    RUBY
+  end
+
+  it 'registers an offense for RSpec.context blocks' do
+    expect_offense(<<-RUBY)
+      RSpec.context 'context' do
+      ^^^^^^^^^^^^^^^^^^^^^^^ Prefer `describe` over `context`.
+        it 'tests common context invocations' do
+          expect(request.context).to be_empty?
```
Darhazer: I don't get why this isn't flagged with the cop config 'expect' => 'expect'

Darhazer:
> But it’s much simpler this way, and if people want to abuse it I guess they’ll have to live with the pain themselves :-)

With great power comes great responsibility.

I'd like to have the options restricted to the actual aliases, but that adds much complexity to the code, makes it harder to support and just adds some safety, that might not be needed at all. I guess the core Rubocop cops that allow enforcing preferred methods can be abused pretty much the same way as this one.

It gives great power to the users, so let's ship it

spec/rubocop/cop/rspec/dialect_spec.rb
bquorning: Good point. 

@gsamokovarov ?

gsamokovarov: Let me check, I think I had this when I experimented with both `on_block` and `on_send` and testing that the auto-correcting code can handle well both node types.

config/default.yml
dgollahon: I don't think Dialect is a bad name, necessarily, but my current suggestion would be `PreferredAliases` i think because i expect that would be more self-evident, though "dialect" is probably clear enough once you understand what it does. But without knowing what the cop does, i think `PreferredAliases` might be a hair more descriptive.

Not a blocker, just my $0.02.

spec/rubocop/cop/rspec/dialect_spec.rb
gsamokovarov: [Rubocop::Cop::MethdoPreferrence#preferred_methods](https://github.com/rubocop-hq/rubocop/blob/master/lib/rubocop/cop/mixin/method_preference.rb#L13-L24) filters entries like `{'foo' => 'foo'}`. If I change:

```diff
- 'expect' => 'expect'
+ 'expect' => 'should'
```

I do get an offense (even though, it does not auto-correct properly, but this is expected).

gsamokovarov:
Folks, I'm leaving this as is for the moment. Ping me if you require more changes.

Darhazer:
I think It's time to merge this.
@bquorning @dgollahon any final objections?

gsamokovarov:
Thank you, folks! 🎉

PR #736, opened 2019-02-05
Add support for metadata to ContextWording

pirj:
RSpec/ContextWording cop previously missed the `context`s with metadata.
This change adds support for metadata detection.

---

* [x] Added tests.

### DIFF
--- a/lib/rubocop/cop/rspec/context_wording.rb
+++ b/lib/rubocop/cop/rspec/context_wording.rb
@@ -30,7 +30,7 @@ module RuboCop
         MSG = 'Start context description with %<prefixes>s.'.freeze
 
         def_node_matcher :context_wording, <<-PATTERN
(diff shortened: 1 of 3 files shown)


lib/rubocop/cop/rspec/context_wording.rb
```diff
@@ -30,7 +30,7 @@ class ContextWording < Cop
         MSG = 'Start context description with %<prefixes>s.'.freeze
 
         def_node_matcher :context_wording, <<-PATTERN
-          (block (send _ { :context :shared_context } $(str #bad_prefix?)) ...)
```
Darhazer: perhaps also the receiver should be `RSPEC`

pirj:
Good to merge?

pirj:
Thanks guys!

PR #738, opened 2019-02-09
Use expect_correction on more specs

bquorning:
---

* [x] Added tests.

### DIFF
--- a/spec/rubocop/cop/rspec/hooks_before_examples_spec.rb
+++ b/spec/rubocop/cop/rspec/hooks_before_examples_spec.rb
@@ -9,6 +9,13 @@ RSpec.describe RuboCop::Cop::RSpec::HooksBeforeExamples do
         ^^^^^^^^^^^^^^^^ Move `before` above the examples in the group.
       end
     RUBY
+
+    expect_correction(<<-RUBY)
+      RSpec.describe User do
+        before { setup }
+        it { is_expected.to be_after_let }
+      end
+    RUBY
   end
 
   it 'flags `before` after `context`' do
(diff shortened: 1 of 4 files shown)


spec/rubocop/cop/rspec/hooks_before_examples_spec.rb
```diff
@@ -33,6 +50,14 @@
         ^^^^^^^^^^^^^^^^ Move `before` above the examples in the group.
       end
     RUBY
+
+    expect_correction(<<-RUBY)
+      RSpec.describe User do
+        before { setup }
+        include_examples('should be after let')
```
Darhazer: guess we should reword this example to use `before` and not `let`. Now I see how you find that those two cops are nearly identical :)

spec/rubocop/cop/rspec/multiple_subjects_spec.rb
```diff
@@ -19,6 +19,20 @@
         end
       end
     RUBY
+
+    expect_correction [
```
Darhazer: any reason a heredoc couldn't be used here?

bquorning: Yeah, the code currently in line 27 (`      '  ',`) would contain trailing spaces if formatted as a heredoc.

spec/rubocop/cop/rspec/hooks_before_examples_spec.rb
bquorning: 😆 

Fixed

bquorning:
Another review @Darhazer ? (please merge on +1)

spec/rubocop/cop/rspec/multiple_subjects_spec.rb
Darhazer: Guess we can improve the autocorrect by using `range_with_surrounding_space` but that can happen in another PR.
I wonder how it was working before with the `include_examples` though

bquorning: > I wonder how it was working before with the `include_examples` though

In the previous examples there are two `include_examples('autocorrect', …`. In the first example, two `subject`s are changed to `let`, and heredoc is used. In the second example, two unnamed `subject`s are *removed*, and the indentation is left behind as trailing spaces – and heredoc isn’t used in that example.

And yes, we might be able to improve it in another PR. But I would assume people will have cops guarding against trailing spaces, empty lines etc.

PR #751, opened 2019-04-03
Add ContextWorking cop. Resolves #417

### LINKED ISSUE #417: Proposal: Detect use of `context` for method names instead of `describe`

geniou:
As discussed in #417 is would be good to have a cop to check for the combination of `context` and a method name.

---

* [x] Added tests.


lib/rubocop/cop/rspec/context_method.rb
```diff
@@ -0,0 +1,54 @@
...
+      #
+      #   describe '.foo_bar' do
+      #     # ...
+      #   end
+      class ContextMethod < Cop
+        MSG = 'Use describe for testing methods.'.freeze
+        METHOD_INDICATORS = %w[. #].freeze
+
+        def_node_matcher :context_method, <<-PATTERN
+          (block (send #{RSPEC} { :context } $(str #method_name?) ...) ...)
```
pirj: I believe curly braces are redundant here.

lib/rubocop/cop/rspec/context_method.rb
```diff
@@ -0,0 +1,54 @@
...
+        def autocorrect(node)
+          lambda do |corrector|
+            corrector.replace(node.parent.loc.selector, 'describe')
+          end
+        end
+
+        private
+
+        def method_name?(description)
+          METHOD_INDICATORS.include?(description[0])
```
pirj: This is a bit confusing, and not obvious that `description` is a string. My mistake was that I thought it might be some kind of a node.

WDYT of:
```
description.start_with?(*METHOD_INDICATORS)
```

lib/rubocop/cop/rspec/context_method.rb
```diff
@@ -0,0 +1,54 @@
...
+      #   # good
+      #   describe '#foo_bar' do
+      #     # ...
+      #   end
+      #
+      #   describe '.foo_bar' do
+      #     # ...
+      #   end
+      class ContextMethod < Cop
+        MSG = 'Use describe for testing methods.'.freeze
```
pirj: It seems to be typical to quote code, e.g.:
> 'Declare `subject` above any other
> 'Move `%<hook>s` above the examples in the group
> 'Use `instance_spy` when you check your double

dgollahon:
Don't forget to update the commit + PR title with the current cop name :)

PR #756, opened 2019-04-24
Let `RSpec/DescribedClass` pass `Struct` instantiation closures

schmijos:
The `RSpec/DescribedClass` cop complained about the following commonly used pattern in RSpec tests for concerns:

```ruby
RSpec.describe Uploadable do
  …
  let(:uploadable) do
    Struct
      .new(:mounted_as, :model) { include Uploadable }
      .new(mounted_as, model)
  end
  …
end
```

This PR adds `Struct` to the whitelist of blocks where `described_class` isn't mandatory

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Master (Unreleased)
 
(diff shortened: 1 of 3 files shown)


bquorning:
Thanks for the PR.

PR #760, opened 2019-05-09
Remove AggregateFailuresByDefault config option for MultipleExpectations.

### LINKED ISSUE #697: Pointless AggregateFailuresByDefault config option of RSpec/MultipleExpectations
Having configuration like:

``` yaml
RSpec/MultipleExpectations:
  AggregateFailuresByDefault: true
```

Effectively disables this cop, thus option itself seems to me quiet confusing. ~This option would make sense only if there will be `MaxAggregate` count as well IMO.~

pirj:
Remove AggregateFailuresByDefault config option for MultipleExpectations.

Fixes #697

---

* [x] *Removed* tests.


pirj:
@bquorning Please take a look. I have [another improvement](https://github.com/rubocop-hq/rubocop-rspec/issues/761) in progress for the same cop.

PR #762, opened 2019-05-09
Fix doc issue for ContextWording cop

### LINKED ISSUE #758: Small docs issue for ContextWording
https://github.com/rubocop-hq/rubocop-rspec/blob/749520d187410dc440345bded66150237a6a8662/manual/cops_rspec.md#prefixes-configuration-option-defaults-when-with-and

The end of the sentence is cutoff:

> Prefixes configuration option, defaults: 'when', 'with', and

pirj:
fixes #758

The end of the sentence was cut off:
> Prefixes configuration option, defaults: 'when', 'with', and

Also, the description was not updated when `without` was legalized.

---

* [-] Added tests.

### DIFF
--- a/config/default.yml
+++ b/config/default.yml
@@ -50,7 +50,7 @@ RSpec/BeforeAfterAll:
   StyleGuide: http://www.rubydoc.info/gems/rubocop-rspec/RuboCop/Cop/RSpec/BeforeAfterAll
 
 RSpec/ContextWording:
-  Description: "`context` block descriptions should start with 'when', or 'with'."
+  Description: Checks that `context` docstring starts with an allowed prefix.
   Enabled: true
   Prefixes:
   - when
--- a/lib/rubocop/cop/rspec/context_wording.rb
+++ b/lib/rubocop/cop/rspec/context_wording.rb
@@ -3,18 +3,22 @@
 module RuboCop
   module Cop
     module RSpec
-      # `context` block descriptions should start with 'when', or 'with'.
+      # Checks that `context` docstring starts with an allowed prefix.
       #
       # @see https://github.com/reachlocal/rspec-style-guide#context-descriptions
       # @see http://www.betterspecs.org/#contexts
       #
-      # @example `Prefixes` configuration option, defaults: 'when', 'with', and
-      # 'without'
-      #   Prefixes:
-      #     - when
-      #     - with
-      #     - without
-      #     - if
+      # @example `Prefixes` configuration
+      #
+      #   # .rubocop.yml
+      #   # RSpec/ContextWording:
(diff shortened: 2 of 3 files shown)


config/default.yml
```diff
@@ -50,7 +50,7 @@ RSpec/BeforeAfterAll:
   StyleGuide: http://www.rubydoc.info/gems/rubocop-rspec/RuboCop/Cop/RSpec/BeforeAfterAll
 
 RSpec/ContextWording:
-  Description: "`context` block descriptions should start with 'when', or 'with'."
+  Description: "`context` block descriptions start with 'when', 'with', or 'without'."
```
Darhazer: `Check that` ?

pirj: The primary reason is:
```
lib/rubocop/cop/rspec/context_wording.rb:6:81: C: Metrics/LineLength: Line is too long. [88/80]
      # Check that `context` block descriptions start with 'when', 'with', or 'without'.
```

If I wrap it, there's another failure:
```

  1) config/default.yml does not have newlines in cop descriptions
     Failure/Error: expect(value).not_to include("\n")
       expected "Check that `context` block descriptions start with 'when', 'with',\nor 'without'." not to include "\n"
     # ./spec/project/default_config_spec.rb:63:in `block (3 levels) in <top (required)>'
```

Anyway, there is a number of cops with descriptions that don't conform to most commonly used 'Check that':

> This cop checks for calling a block within a stub.
> This cop checks void `expect()`.

> Prefer using verifying doubles over normal doubles.

> Checks for stubbed test subjects.

> Enforces use of string to titleize shared examples.

Darhazer: Yup this one was also not following it, but at least included `should`, not it just sounds incomplete to me

pirj:
Managed to fit into hard constraints. Please take another look.

config/default.yml
```diff
@@ -50,7 +50,7 @@ RSpec/BeforeAfterAll:
   StyleGuide: http://www.rubydoc.info/gems/rubocop-rspec/RuboCop/Cop/RSpec/BeforeAfterAll
 
 RSpec/ContextWording:
-  Description: "`context` block descriptions should start with 'when', or 'with'."
+  Description: Checks that `context` docstring start with 'when', 'with', or 'without'.
```
pirj: Isn't proper English, should be either `docstrings` either `starts`. However, it's 80 chars long already.
Do you think it makes sense to replace "Checks" with "Check"?

Engineering tradeoffs. I imagined them differently.

config/default.yml
pirj: Please take another look, I think I've nailed it.

Issue #750, opened 2019-03-24
RSpec/InstanceVariable should ignore static (literal) classes


AlexWayfer:
We have some specs where we're defining classes inside specs like this:

```ruby
describe 'Foo' do
  class Bar
    def initialize(name)
      @name = name
    end

    def execute
      p @name
    end
  end
end
```

`RSpec/InstanceVariable` should not fire in this case.

dgollahon:
@AlexWayfer I think you're right that `RSpec/InstanceVariable` is probably not the right mechanism to flag this, but it should actually still get flagged by _something_ as written.

When you define `Bar` using `class` inside of a describe, it is not scoped to a describe. You are permanently introducing `Bar` into the global scope and it will be available across all specs.

See #197 for more details.

I'm assuming this is still broken for dynamic class declarations that do not pollute the global scope where accessing instance variables would be legitimate (using `Class.new { ... }`) so we should at least confirm and fix that.

AlexWayfer:
> When you define Bar using class inside of a describe, it is not scoped to a describe. You are permanently introducing Bar into the global scope and it will be available across all specs.

Yep, I know. And I don't know how to define scoped constants normally, beside ugly (and sometimes slow) `stub_const`. There is approaches like `self::Bar = Class.new`, but it doesn't work for any case (constant nested usage).

dgollahon:
I commonly do something like

```ruby
let(:bar) do
	Class.new(WhateverParentClass) do
		...
	end
end

it 'works' do
	expect(bar.new).to be_whatever
end
```

or within an example just using a local variable

```ruby
it 'works' do
	bar =
		Class.new(WhateverParentClass) do
			...
		end
	
  expect(bar.new).to be_whatever
end
```

If you need a generated constant within a constant for some reason you can use `const_set` on your dynamic class, though it can be a little hairy if you're doing something complex.

```ruby
let(:bar) do
	Class.new(WhateverParentClass) do
		const_set(:BAZ, 5)
	end
end

it 'works' do
	expect(bar::BAZ).to eql(5)
end
```

And you can still inherit from your dynamic class using:

```ruby
baz = Class.new(bar)
```

etc.

Usually I try to think carefully about what I'm doing if any of this is necessary and make sure I haven't created a difficult to test design when I didn't need to.

[...]

PR #765, opened 2019-05-10
Add a cop that detects constant, class, and module declarations in DSL

### LINKED ISSUE #197: Suggestion: Check for constants being defined in specs

pirj:
Those declarations are defined in the global namespace and leak between examples.

Fixes #197 

Related to #750

All credit goes to @jonatas, the original creator. I've changed the original cop created for in-house use to comply with the strict coding style of this repository.

---

* [x] Added tests.


pirj:
@dgollahon Appreciate if you take a look. I'm reading your comments to #381 ATM.

pirj:
```
before do
  stub_const('Foo', Class.new)
  class Foo
    # define the class now
  end
end
```

that you mention can be written differently, by defining the method inside a block passed to `Class.new`, e.g.
```
before do
  fake_class = Class.new do
    # define the class now
  end
  stub_const('Foo', fake_class)
end
```

Do you think that it's a frequent situation when *some* methods are defined in a block, and at the same time examples/contexts need to add/redefine methods on that class?

cc @backus

lib/rubocop/cop/rspec/leaky_constant_declaration.rb
```diff
@@ -0,0 +1,105 @@
...
+          add_offense(node, location: :expression)
+        end
+
+        private
+
+        def inside_describe_block?(node)
+          node.each_ancestor(:block).any?(&method(:in_example_or_shared_group?))
+        end
+
+        def_node_search :in_example_or_shared_group?, <<-PATTERN
```
Darhazer: why is this a `search` and not a matcher - you are basically going up the tree, but every time perform search down ...
Also, why you are not using a block_pattern in the first place?

lib/rubocop/cop/rspec/leaky_constant_declaration.rb
```diff
@@ -0,0 +1,105 @@
...
+
+        def on_module(node)
+          return unless inside_describe_block?(node)
+
+          add_offense(node, location: :expression)
+        end
+
+        private
+
+        def inside_describe_block?(node)
```
Darhazer: this doesn't seem to be tested.

lib/rubocop/cop/rspec/leaky_constant_declaration.rb
```diff
@@ -0,0 +1,105 @@
...
+      #     before do
+      #       fake_class = Class.new(described_class) do
+      #         def do_something
+      #         end
+      #       end
+      #       stub_const('SomeModule::SomeClass', fake_class)
+      #     end
+      #   end
+      class LeakyConstantDeclaration < Cop
+        MSG = 'Stub class or constant instead of declaring explicitly.'.freeze
```
Darhazer: guess we can provide more specific message for class/constant

lib/rubocop/cop/rspec/leaky_constant_declaration.rb
pirj: Oh, this is awesome!

lib/rubocop/cop/rspec/leaky_constant_declaration.rb
pirj: We do here:
```
      bad_code = <<-RUBY
        describe MyClass do
          specify do
            CONSTANT = 'is not allowed'.freeze
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ Stub class or constant instead of declaring explicitly.
```

Do you think this is doesn't cover it enough?

Darhazer: doesn't test if it's outside of a describe block ;)

pirj: Added an example 👍

pirj:
@bquorning please take a look.

spec/rubocop/cop/rspec/leaky_constant_declaration_spec.rb
```diff
@@ -0,0 +1,99 @@
...
+          end
+        end
+      RUBY
+      expect_offense(bad_code)
+    end
+
+    it 'ignores outside of example/shared group' do
+      fair_code = <<-RUBY
+        factory :some_class do
+          CONSTANT = 'is not allowed'.freeze
```
bquorning:
```suggestion
          CONSTANT = 'is allowed'.freeze
```

spec/rubocop/cop/rspec/leaky_constant_declaration_spec.rb
```diff
@@ -0,0 +1,99 @@
...
+        describe SomeClass do
+          class DummyClass < described_class
+          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ Stub class constant instead of declaring explicitly.
+          end
+        end
+      RUBY
+      expect_offense(bad_code)
+    end
+
+    it 'ignores anonymous classes' do
```
bquorning: The anonymous class will also leak into global scope, won’t it?

spec/rubocop/cop/rspec/leaky_constant_declaration_spec.rb
```diff
@@ -0,0 +1,99 @@
...
+
+  describe 'constant assignment' do
+    it 'flags inside an example group' do
+      bad_code = <<-RUBY
+        describe SomeClass do
+          CONSTANT = 'is not allowed'.freeze
+          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ Stub constant instead of declaring explicitly.
+        end
+      RUBY
+      expect_offense(bad_code)
```
bquorning: Most of our specs are written as `expect_offense(<<-RUBY)`, i.e. not defining the code string as a local variable. I know we don’t exactly have a cop for it, but could we keep the consistency anyway? 😉

spec/rubocop/cop/rspec/leaky_constant_declaration_spec.rb
pirj: I'll change this to something more self-explanatory, e.g. `'I will be accessible as ::CONSTANT'`, or neutral, e.g. "I'm a constant", what do you think?

spec/rubocop/cop/rspec/leaky_constant_declaration_spec.rb
pirj: It will define a class, but that class won't clash by name with any other class.
The main problem this cop is addressing is that e.g. several examples may define a `DummyClass`, and instead of being a blank slate class as it will be in the first example, subsequent examples will be reopening it and modifying its behaviour in unpredictable ways. Another case when a class that exists in the codebase is redefined that way.
Do you think it makes sense to add this to cop's documentation?

pirj:
> What to do when I want to instantiate the class inside an example? That is not possible with stub_const, right? (e.g. RuboCop::RSpec::FakeCop)

You can, e.g.:
```
  before do
    fake_class = Class.new(described_class) do
                   def do_something
                   end
                 end
    stub_const('OtherClass', fake_class)
  end
  let(:instance) { fake_class.new }
```

The reason this is not possible with `FakeCop` is different, I've described the reason in [the message for "Disable LeakyConstantDeclaration" commit](https://github.com/rubocop-hq/rubocop-rspec/pull/765/commits/3985e1567cbb7e9efe835fd262b3c23e3076d7ae).

spec/rubocop/cop/rspec/leaky_constant_declaration_spec.rb
bquorning: Yes, that’s probably better. 👍

lib/rubocop/cop/rspec/leaky_constant_declaration.rb
```diff
@@ -0,0 +1,104 @@
+# frozen_string_literal: true
+
+module RuboCop
+  module Cop
+    module RSpec
+      # Checks that no class, module, or a constant is declared.
```
bquorning:
```suggestion
      # Checks that no class, module, or constant is declared.
```

spec/rubocop/cop/rspec/leaky_constant_declaration_spec.rb
bquorning: Yes, I think we should add a node that the cop allows anonymous classes. Should there be a bad/good example for it too?

spec/rubocop/cop/rspec/leaky_constant_declaration_spec.rb
```diff
@@ -0,0 +1,93 @@
...
+            CONSTANT = 'is not allowed'.freeze
+            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ Stub constant instead of declaring explicitly.
+          end
+        end
+      RUBY
+    end
+
+    it 'ignores outside of example/shared group' do
+      fair_code = <<-RUBY
+        factory :some_class do
```
pirj: Does it make sense to submit this cop to core RuboCop? Are there some legit cases for constant declaration in a block?

```
c = Class.new do
  A = 1
  def a
    A
  end
end

# Pretty much as expected
c.new.a # => 1

# Totally unexpected
A # => 1
```

spec/rubocop/cop/rspec/leaky_constant_declaration_spec.rb
pirj: Added a note.

Existing good examples are using anonymous classes, so probably no additional example is needed.

spec/rubocop/cop/rspec/leaky_constant_declaration_spec.rb
bquorning: Good question. You could try it out on some open source projects, see if you find any violations there.

lib/rubocop/cop/rspec/leaky_constant_declaration.rb
```diff
@@ -17,6 +17,8 @@ module RSpec
       # Anonymous classes are fine, since they don't result in global
       # namespace name clashes.
       #
+      # @see https://relishapp.com/rspec/rspec-mocks/docs/mutating-constants
```
bquorning: While I like this change (extract duplication), it seems the link doesn’t make it to the manual now.

spec/rubocop/cop/rspec/leaky_constant_declaration_spec.rb
pirj: It turns outs the vast majority of the offences are in specs.

lib/rubocop/cop/rspec/leaky_constant_declaration.rb
pirj: It seems to be quite common:
```
lib/rubocop/cop/rspec/subject_stub.rb|8| 10:      # @see https://robots.thoughtbot.com/don-t-stub-the-system-under-test
lib/rubocop/cop/rspec/leaky_constant_declaration.rb|20| 10:      # @see https://relishapp.com/rspec/rspec-mocks/docs/mutating-constants
lib/rubocop/cop/rspec/context_wording.rb|8| 10:      # @see https://github.com/reachlocal/rspec-style-guide#context-descriptions
lib/rubocop/cop/rspec/context_wording.rb|9| 10:      # @see http://www.betterspecs.org/#contexts
lib/rubocop/cop/rspec/example_wording.rb|10| 10:      # @see http://betterspecs.org/#should
lib/rubocop/cop/rspec/verified_doubles.rb|8| 10:      # @see https://relishapp.com/rspec/rspec-mocks/docs/verifying-doubles
lib/rubocop/cop/rspec/describe_symbol.rb|19| 10:      # @see https://github.com/rspec/rspec-core/issues/1610
lib/rubocop/cop/rspec/multiple_expectations.rb|8| 10:      # @see http://betterspecs.org/#single Single expectation test
```

Do you mind if I file a ticket and we could either fix code documentation to use plain "See: ", or fix documentation generator to include `see` to the manual?

bquorning: Good idea.

pirj: Filed #772.

pirj:
@bquorning @Darhazer Do you have something in mind that needs to be made here or is it ready to be merged?

bquorning:
Nope, this is merge-able. Thank you for this cop ❤️

dgollahon:
@pirj Sorry I didn't get a chance to review this. I haven't been doing much OSS lately and the ping got lost in my notification queue. Glad you pushed through and implemented the cop, I think it's a valuable one :)

I just skimmed my old comments so I don't know if I clearly explained this before or if it has a place in the good/bad examples in the docs (or maybe in the style guide?) is that cases like:

```ruby
before do
  stub_const('Foo', Class.new)
  class Foo
    # define the class now
  end
end
```

that could be written to pass the cop like this:

```ruby
before do
  fake_class = Class.new do
    # define the class now
  end
  stub_const('Foo', fake_class)
end
```

should, imo, usually be written like this:

```ruby
let(:foo_class) do
  Class.new do
    # define the class now
  end
end
```

This way _nothing_ mutates the constant space and you can't possibly run into the normal issues you can with `stub_const`. Usually when you need a faux class you don't need it to be referenced as a particular constant name. In rare cases where that is needed, I find it's usually overriding some existing constant (occasionally necessary in weird situations), or dynamically setting a constant on a newly generated dynamic class (and in this case i would use my `let` version on the outermost class, still not polluting the global constant space). My main point is that it's sometimes worth reminding people that classes are just objects and don't have to be assigned to a constant, so this strategy might have a place in our guide or documentation somewhere, but as with most techniques, it should be used judiciously.

As an aside / trivia, if you use this it helps to be aware that anonymous classes have unusual `#name` behavior.

```ruby
Class.new.name # => nil
```
but if you write it this way

```ruby
Foo = Class.new
Foo.name # => 'Foo'
```

the constant assignment causes the anonymous class to receive a name (i'm guessing this is some sort of const assignment hook that mutates the anonymous class, but I don't know exactly how it works, mechanically). In practice this makes it seem the same as if you had written your class the standard way

```ruby
class Foo
end
```

Anyway, nice work.

dgollahon:
Ah, @bquorning also pointed to a good alternate explanation of what I was saying here: https://github.com/rubocop-hq/rubocop-rspec/issues/197#issuecomment-261495168

pirj:
@dgollahon, love the reasoning for sacrificial classes.

I had a problem similar to what you mention just today:

```ruby
    let(:dummy_job) do
      Class.new(ActiveJob::Base) do
        def self.name
          'DummyJob'
        end
 
        def perform
          # ...
        end
      end
    end
```

I ended up adding:
```
    before do
      stub_const('DummyJob', dummy_job_class)
    end
```

simply because ActiveJob is expecting `::DummyJob` to be present in integration tests.

So, YMMV.
Also, [this case](https://github.com/rubocop-hq/rubocop-rspec/pull/765/commits/1599b1864efd00e66cf2044f1ded887a4f09ed6a) won't work, since parent class's `def self.inherited` expects to have a class name, the situation you mentioned.

Leaking constants problem is undeservedly missed from the guide.

Are you up to add sacrificial classes guideline to the style guide and update `LeakyConstantDeclaration` cop's good examples to reflect this?

dgollahon:
> because ActiveJob is expecting ::DummyJob to be present in integration tests

Really? Wow, that's odd. I haven't interacted with ActiveJob in a long time but that is a weird API.

> Also, this case won't work, since parent class's def self.inherited expects to have a class name, the situation you mentioned.

Ah, yeah, interesting case. Any time you have inheritance hooks things get weird.

> Are you up to add sacrificial classes guideline to the style guide and update LeakyConstantDeclaration cop's good examples to reflect this?

Maybe, but likely not. I'm a little less keen on maintaining the style guide v.s. `rubocop-rspec`, but I'm happy to weigh in on and review any changes if someone else would like to update the information. I might change my mind, but there are other things I'd probably rather tackle first.

pirj:
@dgollahon In response to this https://github.com/rubocop-hq/rubocop-rspec/issues/750#issuecomment-515582100

It's actually a very good point that re-opening the class doesn't by itself lead to constant leakage.

I'm concerned about the possibility of the leakage when `stub_const` is being removed. Some cases of that are impossible to detect statically, e.g. when `stub_const` is defined in a shared example residing in another file.
This makes it a really hard task to detect if it's re-opening of a stubbed class or a new class definition and with all the downsides of the latter and existing workaround (`Class.new do`) for the former I'm not sure if it is worth it to allow re-opening case.

At the same time, I can think of a case when constants need to be defined in that stubbed class, e.g.
```ruby
let(:klass) do
  Class.new(ClassUnderTest) do
    SOME_CONSTANT = ...
  end
end
```

and without re-opening, this example may become quite clumsy.

Let's give it another thought if we notice or someone complains that the cop doesn't simplify our lives as it is, what do you think?

PR #768, opened 2019-05-14
Improve MultipleExpectations metadata detection

### LINKED ISSUE #761: Improve RSpec/MultipleExpectations `aggregate_failures` detection
There are cases when two `expect`s are actually a single expectation, specifically when testing matchers:
```ruby
  it 'fails if the method is called with an unexpected argument', :reset do
    expect(receiver).to receive(:something).with(any_of(1, 2)).twice
    expect { receiver.something(3) }
      .to fail_with('unexpected arguments')
  end
```
In this case, `MultipleExpectations` complains that there are more than two expectations, even though the first expectation is an expectation under test.

This can be reworked as a nested expectation, neither test result or cop result change:
```ruby
  it 'fails if the method is called with an unexpected argument', :reset do
    expect do
      expect(receiver).to receive(:something).with(any_of(1, 2)).twice
      receiver.something(3)
    end
      .to fail_with('unexpected arguments')
  end
```

I understand that by wrapping the whole example into an `aggregate_failures do` helps to pacify the cop, as does `:aggregate_failures` *example* metadata.

I'd agree to add `:aggregate_failures` to example *group* containing those examples. Would you accept a pull request that walks the ancestors to detects if it's set for those examples?

pirj:
Previously, `aggregate_failures` were only detected on the example itself, even though this metadata can be set on any of ancestor example groups.
E.g.:

    it '...', :aggregate_failures do
      expect(foo).to eq(bar)
      expect(baz).to eq(bar)
    end

was ignored, but:

    describe '...', :aggregate_failures do
      it '...' do
        expect(foo).to eq(bar)
        expect(baz).to eq(bar)
      end
    end

was not.

Fixes #761 

---

* [x] Added tests.

### DIFF
--- a/lib/rubocop/cop/rspec/multiple_expectations.rb
+++ b/lib/rubocop/cop/rspec/multiple_expectations.rb
@@ -50,20 +50,20 @@ module RuboCop
 
         MSG = 'Example has too many expectations [%<total>d/%<max>d].'
 
-        def_node_search :with_aggregated_failures?, '(sym :aggregate_failures)'
-        def_node_search :disabled_aggregated_failures?, <<-PATTERN
+        def_node_search :with_aggregate_failures?, '(sym :aggregate_failures)'
+        def_node_search :disabled_aggregate_failures?, <<-PATTERN
           (pair (sym :aggregate_failures) (false))
         PATTERN
 
         def_node_matcher :expect?, Expectations::ALL.send_pattern
-        def_node_matcher :aggregate_failures?, <<-PATTERN
+        def_node_matcher :aggregate_failures_block?, <<-PATTERN
           (block (send _ :aggregate_failures ...) ...)
         PATTERN
 
         def on_block(node)
           return unless example?(node)
 
-          return if example_with_aggregated_failures?(node)
+          return if example_with_aggregate_failures?(node)
 
           expectations_count = to_enum(:find_expectation, node).count
 
@@ -76,18 +76,40 @@ module RuboCop
 
         private
 
-        def example_with_aggregated_failures?(node)
-          example = node.send_node
+        def example_with_aggregate_failures?(example_node)
+          node_with_aggregate_failures = find_aggregate_failures(example_node)
+          return false unless node_with_aggregate_failures
 
-          with_aggregated_failures?(example) &&
-            !disabled_aggregated_failures?(example)
+          aggregate_failures?(node_with_aggregate_failures)
+        end
+
+        def find_aggregate_failures(example_node)
+          example_node.send_node.each_ancestor(:block)
+            .find { |block_node| aggregate_failures_present?(block_node) }
+        end
+
+        def aggregate_failures_present?(node)
+          metadata(node)&.any?(&method(:with_aggregate_failures?))
+        end
+
+        def aggregate_failures?(example_or_group_node)
+          metadata(example_or_group_node)&.any? do |metadata|
+            with_aggregate_failures?(metadata) &&
+              !disabled_aggregate_failures?(metadata)
+          end
+        end
+
+        def metadata(example_or_group_node)
+          RuboCop::RSpec::Example
+            .new(example_or_group_node)
+            .metadata
         end
 
         def find_expectation(node, &block)
-          yield if expect?(node) || aggregate_failures?(node)
+          yield if expect?(node) || aggregate_failures_block?(node)
 
           # do not search inside of aggregate_failures block
-          return if aggregate_failures?(node)
+          return if aggregate_failures_block?(node)
 
           node.each_child_node do |child|
             find_expectation(child, &block)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 * Remove `AggregateFailuresByDefault` config option of `RSpec/MultipleExpectations`. ([@pirj][])
 * Add `RSpec/LeakyConstantDeclaration` cop. ([@jonatas][], [@pirj][])
+* Improve `aggregate_failures` metadata detection of `RSpec/MultipleExpectations`. ([@pirj][])
 
 ## 1.33.0 (2019-05-13)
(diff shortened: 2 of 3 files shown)


lib/rubocop/cop/rspec/multiple_expectations.rb
```diff
@@ -77,10 +77,22 @@ def on_block(node)
...
+          node
+            .each_ancestor(:block)
+            .find { |ancestor| aggregated_failures?(ancestor) }
+        end
 
-          with_aggregated_failures?(example) &&
-            !disabled_aggregated_failures?(example)
+        def aggregated_failures?(node)
+          example = RuboCop::RSpec::Example.new(node)
+          example.metadata&.any? do |metadata|
```
Darhazer: it would be so easier when we update to Rubocop 0.68 and start using [children matching](https://github.com/rubocop-hq/rubocop/pull/6965)

pirj: Awesome, loving [it in action](https://github.com/rubocop-hq/rubocop/pull/6999)!

pirj:
Thanks for the thorough review @Darhazer, you discovered an edge case I've missed in specs and found a bug in the code when example group `aggregate_failures` metadata was actually taking precedence over example metadata.
Fixed.

lib/rubocop/cop/rspec/multiple_expectations.rb
```diff
@@ -77,10 +77,27 @@ def on_block(node)
         private
 
         def example_with_aggregated_failures?(node)
-          example = node.send_node
+          aggregate_failures = aggregated_failures?(node)
 
-          with_aggregated_failures?(example) &&
-            !disabled_aggregated_failures?(example)
+          aggregate_failures ||
```
Darhazer: this is a weird pattern (making distinctions between false and nil). Perhaps aggregated_failures? should perform the search up in the three, in case it's not set in the current node, and just return true/false

pirj: Fair enough. Reworked the search to a recursive method.

pirj:
@Darhazer Appreciate if you take another look.

lib/rubocop/cop/rspec/multiple_expectations.rb
pirj: Sent #773 to update RuboCop to 0.70.0+.
Going to update this pull request to use the new children matching syntax.

pirj: For the following code:
```
code = <<-CODE
describe SomeClass, :inline, :aggregate_failures, :bulk do
  a
end
CODE
```

```
def match(pattern, source)
  ruby_version = RUBY_VERSION.to_f
  node_pattern = RuboCop::NodePattern.new(pattern)
  ast = RuboCop::ProcessedSource.new(source, ruby_version).ast
  node_pattern.match(ast)
end

match('(block (send $<(sym :aggregate_failures)>) ...)', code) # => nil
match('(block (send $<(sym :aggregate_failures)> ...) ...)', code) # => nil
match('(block (send $...) ...)', code)
# => [nil, :describe, s(:const, nil, :SomeClass), s(:sym, :inline), s(:sym, :aggregate_failures), s(:sym, :bulk)] 
```

Not sure how children matching can help me in this case. Can you please give an advise, maybe I'm missing something obvious?

Also, metadata can be in different positions, e.g. in `specify` without a docstring, only symbolic metadata, or only hash metadata, this is only getting more complicated.


Just a clarification why I'm trying to avoid using `def_node_search`:
```
describe SomeClass, :inline, aggregate_failures: false, :bulk do
  describe '#something' do
    # current example being examined
    it 'does something' do
      # ...
    end

    # sibling example
    it 'does something else', aggregate_failures: true do # I don't want it to match this
      # ...
    end
  end
end
```

When I do tree walking through parents, I only want to match parent's metadata, so doing a `def_node_search` might mistakenly find metadata from a sibling example. Or at least it's confusing.

pirj:
Flipped recursive walking condition, so `example_with_aggregate_failures?` looks slightly better.

lib/rubocop/cop/rspec/multiple_expectations.rb
```diff
@@ -76,18 +76,39 @@ def on_block(node)
...
         private
 
-        def example_with_aggregated_failures?(node)
-          example = node.send_node
+        def example_with_aggregate_failures?(node)
+          return aggregate_failures?(node) if aggregate_failures_present?(node)
 
-          with_aggregated_failures?(example) &&
-            !disabled_aggregated_failures?(example)
+          parent = parent_block(node)
```
Darhazer: Not a big deal, but I guess you can get `node.each_ancestor(:block).find { |parent| aggregate_failures_present?(parent) }` (I feel better naming would be `with_aggregate_failures?`, but then we need to change the matcher name) and then just return `aggregate_failures?(parent) if parent`
Maybe you can move to a method that just returns node (if it defines aggregate failures) or first parent that define, so this just calls aggregate_failures? on the returned node or returns false

Anyway, I'm OK with this code as well, but since this is a `?` method, it should return false, and not nil, if there is no parent defining the metadata

Darhazer:
Generally looks good. I've added a comment - really a nitpick, but we always ought to return booleans from `?` methods

lib/rubocop/cop/rspec/multiple_expectations.rb
pirj: Fair enough. The code you suggested is more elegant, loving it.

lib/rubocop/cop/rspec/multiple_expectations.rb
pirj: Sure, [added one](https://github.com/rubocop-hq/rubocop-rspec/pull/768/files#diff-7ce36e28810f5439e2331ed3b50431aaR180).

PR #770, opened 2019-05-15
Improve RSpec/SubjectStub cop

### LINKED ISSUE #705: Inconsistent behavior of SubjectStub
``` ruby
# fails
expect(subject).to receive(:my_method)
expect(subject).to receive(:my_method).and_call_original
expect(subject).to receive(:my_method).and_wrap_original { |m, *| m.call(42) }

# passes
expect(subject).not_to receive(:my_method)
```

It should either fail on `not_to` and `to_no`; or it should pass when matcher has `and_call_original` or `and_wrap_original` modifiers.

pirj:
**1.** Use consistent wording in docstrings

**2.** Improve stub detection

Previously, the following code would still be detected as an offence:

    allow(subject)

even though there are no message expectations.

**3.** Add detection of negated runners (`not_to`, `to_not`) - fixes #705

**4.** Add support for chain and multiple stub detection (`receive_messages`, `receive_message_chain`)

**5.** Add more reasoning references

**6.** Improve offence message

**7.** Remove special case `all` matcher (see #488)

This special case treatment was added to address one use case, of an object that acts as a container for and a delegator of method calls to containing items by inheriting from an `Array`.

The original code that suffered from a false detection was:

    expect(formatter_set).to all(receive(:started).with(files))
    formatter_set.started(files)

However, it has been [reworked to](https://github.com/rubocop-hq/rubocop/blob/70ae6ca7978770ea699a2809f57ca2a6bdeec6be/spec/rubocop/formatter/formatter_set_spec.rb#L17):

      formatter_set.each do |formatter|
        allow(formatter).to receive(:started)
      end
      formatter_set.started(files)
      expect(formatter_set).to all(have_received(:started).with(files))

and can be further simplified to:

    formatter_set.each do |formatter|
      expect(formatter).to receive(:started).with(files)
    end
    formatter_set.started(files)

which in case of this specific object design is more indicative of what happens.

**8.** Added support for `have_received` detection, to detect the bad code above.

**9.** Fixed detection of the unnamed subject stubbing when the subject is named.

Fixes #705

---

* [x] Added tests.

### DIFF
--- a/lib/rubocop/cop/rspec/subject_stub.rb
+++ b/lib/rubocop/cop/rspec/subject_stub.rb
@@ -60,10 +60,14 @@ module RuboCop
         #     expect(foo).to all(receive(:bar))
         #
         def_node_matcher :message_expectation?, <<-PATTERN
-          {
-            (send nil? :allow (send nil? %))
-            (send (send nil? :expect (send nil? %)) :to #expectation?)
-          }
+          (send
+            {
+              (send nil? { :expect :allow } (send nil? %))
+              (send nil? :is_expected)
+            }
+            :to
+            #expectation?
+          )
         PATTERN
 
         def_node_matcher :all_matcher?, '(send nil? :all ...)'
--- a/lib/rubocop/cop/rspec/subject_stub.rb
+++ b/lib/rubocop/cop/rspec/subject_stub.rb
@@ -6,6 +6,8 @@ module RuboCop
       # Checks for stubbed test subjects.
       #
       # @see https://robots.thoughtbot.com/don-t-stub-the-system-under-test
+      # @see https://samphippen.com/introducing-rspec-smells-and-where-to-find-them#smell-1-stubject
+      # @see https://github.com/rubocop-hq/rspec-style-guide#dont-stub-subject
       #
       # @example
       #   # bad
--- a/lib/rubocop/cop/rspec/subject_stub.rb
+++ b/lib/rubocop/cop/rspec/subject_stub.rb
@@ -70,7 +70,9 @@ module RuboCop
         PATTERN
 
         def_node_search :message_expectation_matcher?, <<-PATTERN
-          (send nil? { :receive :receive_messages :receive_message_chain } ...)
+          (send nil? {
+            :receive :receive_messages :receive_message_chain :have_received
+            } ...)
         PATTERN
 
         def on_block(node)
--- a/lib/rubocop/cop/rspec/subject_stub.rb
+++ b/lib/rubocop/cop/rspec/subject_stub.rb
@@ -61,7 +61,7 @@ module RuboCop
         def_node_matcher :message_expectation?, <<-PATTERN
           (send
             {
-              (send nil? { :expect :allow } (send nil? %))
+              (send nil? { :expect :allow } (send nil? {% :subject}))
               (send nil? :is_expected)
             }
             #{Runners::ALL.node_pattern_union}
--- a/lib/rubocop/cop/rspec/subject_stub.rb
+++ b/lib/rubocop/cop/rspec/subject_stub.rb
@@ -65,7 +65,7 @@ module RuboCop
               (send nil? { :expect :allow } (send nil? %))
               (send nil? :is_expected)
             }
-            :to
+            #{Runners::ALL.node_pattern_union}
             #expectation?
           )
         PATTERN
--- a/lib/rubocop/cop/rspec/subject_stub.rb
+++ b/lib/rubocop/cop/rspec/subject_stub.rb
@@ -72,7 +72,9 @@ module RuboCop
 
         def_node_matcher :all_matcher?, '(send nil? :all ...)'
 
-        def_node_search :receive_message?, '(send nil? :receive ...)'
+        def_node_search :receive_message?, <<-PATTERN
+          (send nil? { :receive :receive_messages :receive_message_chain } ...)
+        PATTERN
 
         def expectation?(node)
           return if all_matcher?(node)
--- a/lib/rubocop/cop/rspec/subject_stub.rb
+++ b/lib/rubocop/cop/rspec/subject_stub.rb
@@ -20,7 +20,7 @@ module RuboCop
       class SubjectStub < Cop
         include RuboCop::RSpec::TopLevelDescribe
 
-        MSG = 'Do not stub your test subject.'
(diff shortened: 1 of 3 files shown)


lib/rubocop/cop/rspec/subject_stub.rb
```diff
@@ -56,25 +58,19 @@ class SubjectStub < Cop
...
-        #   @example source that not matches
-        #     expect(foo).to all(receive(:bar))
-        #
         def_node_matcher :message_expectation?, <<-PATTERN
-          {
-            (send nil? :allow (send nil? %))
-            (send (send nil? :expect (send nil? %)) :to #expectation?)
-          }
+          (send
+            (send nil? { :expect :allow } (send nil? %))
```
Darhazer: Wonder whether we should extend this to all expectations, using the Language module. Then we may need to add the stubs there as well

lib/rubocop/cop/rspec/subject_stub.rb
```diff
@@ -56,25 +58,19 @@ class SubjectStub < Cop
...
-        def_node_search :receive_message?, '(send nil? :receive ...)'
-
-        def expectation?(node)
-          return if all_matcher?(node)
-
-          receive_message?(node)
-        end
+        def_node_search :message_expectation_matcher?, <<-PATTERN
+          (send nil? {
+            :receive :receive_messages :receive_message_chain :have_received
```
Darhazer: this also might play nice in `Language`

lib/rubocop/cop/rspec/subject_stub.rb
pirj: For this specific case when we check if `subject` is stubbed, `expect/allow_any_instance_of` won't come into play. Is there anything else?

lib/rubocop/cop/rspec/subject_stub.rb
pirj: I was thinking about this, but couldn't find another usage that would benefit from that extraction.
Just noticed that `stub_chain` (an alias of `receive_message_chain`?) is supported in a number of cops, even though it's not in RSpec 3 anymore. Do you think it makes sense to add it to this cop?

lib/rubocop/cop/rspec/subject_stub.rb
Darhazer: Didn't see it in the wild, but `is_expected.to receive(:message)`

lib/rubocop/cop/rspec/subject_stub.rb
Darhazer: Since it requires monkey patching, and we defined as a no-goal to enforce expect over should, I think we should not take care of any of the old syntax methods and clean up them.

lib/rubocop/cop/rspec/subject_stub.rb
pirj: Fair enough, pretty possible use case.

bquorning:
I’m not sure I understand the case for removing the `all` matcher. Yes, it was implemented after just a single bug report, but that doesn’t mean there are not more examples out there. I assume – since RSpec allows it – that they are valid use cases.

Or is the code removed unnecessarily complex?

I just need a bit more clarification on _why_ we should remove this feature.

pirj:
I'm sorry, but I disagree that if RSpec allows something, it's a valid use case. `rubocop-rspec` wouldn't have existed. Did I understand your statement correctly?

As an experiment, I've run [a modified version](https://gist.github.com/pirj/810a918e74e74ff9224c471e834ab34a) of this cop that only detects `expect(subject).to all receive(...)` (along with `not_to`, `have_received` etc) against [`real-world-ruby-apps`](https://github.com/jeromedalbert/real-world-ruby-apps) (11k files) and [`real-world-rails`](https://github.com/eliotsykes/real-world-rails) (79k files).
The result is that there was no single offence for the cop.

I don't have a strong opinion if this special `all` case should be removed or kept.

What do you think?

pirj:
Tried simplifying that spec on RuboCop side, and it turns out another cop, `RSpec/MessageSpies`, starts complaining that `receive` is used instead of `have_received`. What confuses me is that according to [the chat from 10 Oct 2016](https://gitter.im/rubocop-rspec/Lobby):

> If a spec example has multiple stubs (allow) and you want a focused spec with just one expectation per example – and RuboCop is now telling you that you must change all those stubs into expectations. Not good.

However, [in `rubocop`](https://github.com/rubocop-hq/rubocop/blob/master/.rubocop_todo.yml#L67) this is not a problem, as multiple expectations per example are allowed:
```
RSpec/MultipleExpectations:
Max: 25
```

I'll try my chances with changing the enforced style for `MessageSpies` to `receive`, and fixing that lonely example.

Would you accept this pull request with the removal of `all` special case if those two are accepted?

pirj:
One last known use case for `expect(...).to all receive` [has been fixed](https://github.com/rubocop-hq/rubocop/pull/7067/files#diff-be297a2e1f10d4ed00f8faaf78f0f06cR18).

I'll be more than happy to bring `all` exclusion back in case someone complains about it, please feel free to assign the ticket to me when this comes up.

Is there anything else left to be addressed in this pull request?

pirj:
Is there anything else how I can improve this?

pirj:
@bquorning @Darhazer I've sent [a pull request](https://github.com/pirj/rubocop-rspec/pull/1) that would quickly restore the `all` special case.

In case of emergency, it can be immediately merged.

Do you have some other doubts regarding the changes made in this pull request?

pirj:
@bquorning @Darhazer Is there anything else that needs to be addressed in this pull request?

bquorning:
Hi @pirj, apologies for the slow response rate around here. Your PR is in my review queue, but I just haven’t had time to get a good look at it yet.

pirj:
No worries, please take your time.

pirj:
@bquorning I was considering this, but haven't found any other cop that would benefit from it immediately. Probably there are some missed cases due to this in the other cops, but I can't tell that off the top of my head.

pirj:
Please don't merge this just now, @Darhazer has found a case this cop is missing, flagging of an unnamed subject (`expect(subject).to receive`). Will address this quickly.

pirj:
Fixed.

bquorning:
Thank you for the contributions, persistence, and patience 😄

pirj:
Thanks for accepting @bquorning, @Darhazer!

PR #773, opened 2019-05-22
Update RuboCop version to 0.70.0+

pirj:
Since only a minimal version to use is defined, update that version to the latest available at the moment.

This allows for use of [children matching](https://github.com/rubocop-hq/rubocop/pull/6965), that was introduced in 0.68.

---

* [-] Added tests.

### DIFF
--- a/rubocop-rspec.gemspec
+++ b/rubocop-rspec.gemspec
@@ -39,7 +39,7 @@ Gem::Specification.new do |spec|
     'documentation_uri' => 'https://rubocop-rspec.readthedocs.io/'
   }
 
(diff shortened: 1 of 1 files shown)


Darhazer:
This will require our users to update RuboCop in order to update RubocopRSpec and we are postponing such changes for as long as possible. Usually, we update when rubocop-rspec becomes incompatible with newer rubocop versions.

pirj:
Thanks for the heads up. This makes total sense to me 👍

bquorning:
> we are postponing such changes for as long as possible.

That said, I would like to see a PR where children matching is used to improve one of our cops (and perhaps mention other similar possible improvements). That would provide a good context for discussing the upgrade.

PR #774, opened 2019-05-22
Use RuboCop internal affairs department

pirj:
This InternalAffairs/NodeDestructuring cop is [globally disabled in RuboCop itself](https://github.com/rubocop-hq/rubocop/blob/e002ac2/.rubocop_todo.yml#L11), because there's no direct replacement for node deconstruction's var extraction, only `asgn_rhs`.

---

* [-] Added tests.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -1,4 +1,6 @@
-require: rubocop-rspec
+require: 
+ - rubocop-rspec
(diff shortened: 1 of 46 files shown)


dgollahon:
Awesome, nice little cleanup. I love when we improve our own linting.

Issue #775, opened 2019-06-03
Focus Cop fails to detect focus instances when preceding a hash


f1sherman:
Example blocks that contain `:focus` preceding a hash `e.g. js: true` are not flagged.

Reproduction spec:
```ruby
  it 'flags rspec example blocks that include `:focus` preceding a hash' do
    expect_offense(<<-RUBY)
      describe 'test', :focus, js: true do; end
                       ^^^^^^ Focused spec found.
    RUBY
  end
```

```
Failures:

  1) RuboCop::Cop::RSpec::Focus flags rspec example blocks that include `:focus` preceding a hash
     Failure/Error: super

       expected: "      describe 'test', :focus, js: true do; end\n                       ^^^^^^ Focused spec found.\n"
            got: "      describe 'test', :focus, js: true do; end\n"

       (compared using ==)

       Diff:
       @@ -1,3 +1,2 @@
              describe 'test', :focus, js: true do; end
       -                       ^^^^^^ Focused spec found.
```

PR #777, opened 2019-06-18
Improve Focus cop

dgollahon:
Improves the Focus cop in a couple of ways.

- Fixes the incorrect metadata detection from #775
- Adds `pending` to the examples that it can detect (since these can be focused).
- WIP/TODO: Trying to address #398 as well but it's actually much trickier than it might first appear.

---

* [x] Added tests.

### DIFF
--- a/lib/rubocop/cop/rspec/focus.rb
+++ b/lib/rubocop/cop/rspec/focus.rb
@@ -32,12 +32,9 @@ module RuboCop
 
         FOCUSABLE_SELECTORS = focusable.node_pattern_union
 
-        FOCUS_SYMBOL = s(:sym, :focus)
-        FOCUS_TRUE   = s(:pair, FOCUS_SYMBOL, s(:true))
-
         def_node_matcher :metadata, <<-PATTERN
-          {(send #{RSPEC} #{FOCUSABLE_SELECTORS} ... (hash $...))
-           (send #{RSPEC} #{FOCUSABLE_SELECTORS} $...)}
+          {(send #{RSPEC} #{FOCUSABLE_SELECTORS} <$(sym :focus) ...>)
+           (send #{RSPEC} #{FOCUSABLE_SELECTORS} ... (hash <$(pair (sym :focus) true) ...>))}
         PATTERN
 
         def_node_matcher :focused_block?, focused.send_pattern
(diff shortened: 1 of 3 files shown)


lib/rubocop/cop/rspec/focus.rb
```diff
@@ -26,18 +26,16 @@ class Focus < Cop
...
 
         FOCUSABLE_SELECTORS = focusable.node_pattern_union
 
-        FOCUS_SYMBOL = s(:sym, :focus)
-        FOCUS_TRUE   = s(:pair, FOCUS_SYMBOL, s(:true))
-
         def_node_matcher :metadata, <<-PATTERN
-          {(send #{RSPEC} #{FOCUSABLE_SELECTORS} ... (hash $...))
-           (send #{RSPEC} #{FOCUSABLE_SELECTORS} $...)}
+          {(send #{RSPEC} #{FOCUSABLE_SELECTORS} <$(sym :focus) ...>)
```
Darhazer: I believe we need to update the minimum required rubocop version to use this. But nearly every cop that deals with a hash could take advantage of this, so maybe we just need to open a PR with the related updates?

dgollahon: Ah, ok. I didn't realize how new it was. It's possible write without using this, but this was definitely the cleanest option I came up with.

Definitely worth reviewing other hash matchers for similar edge cases and simplifications.

I can open a PR to update our constraints if you think that's reasonable.

bquorning: I’d love to see a PR that updates our Rubocop version requirement, and uses the new node matcher syntax where appropriate.

pirj: This is awesome! 💯 
I thought for some reason this new syntax only works for hashes.

bquorning: #804 does it 🎉

pirj:
@bquorning @Darhazer Is there anything apart from the missing changelong entry that needs to be addressed here?

I believe this may be released together with #804 that bumps RuboCop to the version that provides children matching.

bquorning:
Added 1 Changelog commit. Rebased on master (and force-pushed) so we can merge without conflicts.

bquorning:
@Darhazer @pirj @dgollahon Could one of you please double check that this PR is ready to merge? I’ll rebase #824 afterwards.

dgollahon:
Thanks @bquorning.

Issue #735, opened 2019-01-31
`describe`/`context` inside module


bquorning:
Today I discovered code like this in a spec file:

```rb
module Foo
  RSpec.describe Bar do
  end
end
```

which I found a bit strange. I would rewrite that as

```rb
RSpec.describe Foo::Bar do
end
```

Is this a common enough problem that we need a cop it?

pirj:
Haven't' seen this so far.

pirj:
Actually, I have. https://github.com/rspec/rspec-expectations/blob/master/spec/rspec/expectations/configuration_spec.rb#L3

This allows to write:
```
allow(Expectations::Syntax.default_should_host).to
```
instead of a more verbose:
```
allow(RSpec::Expectations::Syntax.default_should_host).to
```

While fixing RuboCop offences in that repository, I tried fixing this as well (due to `Metrics/ModuleLength`), and it turned out not to bring any improvements except for reduced indentation.

dgollahon:
I've actually seen this in multiple codebases I've worked on. See my comment here: https://github.com/rubocop-hq/rubocop-rspec/pull/381#discussion_r107826572

pirj:
Did you have `DescribedClass`'s pending spec in mind @bquorning ?
I've fixed the cop, going to submit it soon.

bquorning:
No, I hadn’t noticed [that spec](https://github.com/rubocop-hq/rubocop-rspec/blob/77fabdc7f3f2a4876ba6bd44a40d2e83fded8296/spec/rubocop/cop/rspec/described_class_spec.rb#L195-L206). But yes, that is an example of the issue I mentioned.

pirj:
As `DescribedClass` has been merged and the nasty pending spec is green, and the approach of namespacing specs with modules is quite usable with no visible downsides, do you think there's still something to address in the context of this issue, @bquorning?.

dgollahon:
Still seems worthwhile to me for consistency's sake. I would have liked something to autocorrect one project that used a mix of this form and the standard form.

[...]

PR #784, opened 2019-07-10
Improve `RSpec/DescribedClass` cop

### LINKED ISSUE #604: RSpec/DescribedClass does not handle nested describes
Example:
```ruby
RSpec.describe Foo do
  it 'has a baz' do
    expect(described_class.baz).to eql('baz')
  end

  describe Bar do
    it 'has a matching baz implementation' do
      expect(Foo.baz).to eql(Bar.baz)
    end
  end
end
```

Outcome:
image

The cop wrongly flags `Foo` and suggests a semantically invalid change. Meanwhile `Bar` is not highlighted. `RSpec/DescribedClass` should account for nested class describes.

pirj:
Previously, `RSpec/DescribedClass` proposed to replace `A` in the following code to `described_class`:

```ruby
RSpec.describe A do
  describe B do
    subject { B }
    let(:a) { A }
  end
end
```

It is incorrect since RSpec 3.0, which has introduced a breaking change, and the innermost example group with a class argument populates `described_class`, not the outermost.
[Related RSpec CHANGELOG](https://github.com/rspec/rspec-core/blob/master/Changelog.md#300rc1--2014-05-18).
Unfortunately, [RSpec documentation was outdated](https://github.com/rspec/rspec-core/issues/2627), but now it's [fixed](https://github.com/rspec/rspec-core/pull/2629).

Fixes #604, #785
Relates to (fixes?) #735

## NOT DONE (deliberately)

 - examples, detection for `explicit` style
 - detection of the other example group (`example_group`, `feature`, `context`, ...)

---

* [x] Added tests.

### DIFF
--- a/lib/rubocop/cop/rspec/described_class.rb
+++ b/lib/rubocop/cop/rspec/described_class.rb
@@ -48,7 +48,7 @@ module RuboCop
         def_node_matcher :scope_changing_syntax?, '{def class module}'
 
         def_node_matcher :described_constant, <<-PATTERN
-          (block (send _ :describe $(const ...)) (args) $_)
+          (block (send _ :describe $(const ...) ...) (args) $_)
         PATTERN
 
         def on_block(node)
--- a/lib/rubocop/cop/rspec/described_class.rb
+++ b/lib/rubocop/cop/rspec/described_class.rb
@@ -109,7 +109,7 @@ module RuboCop
         end
 
         def skip_blocks?
-          cop_config['SkipBlocks'].equal?(true)
+          cop_config['SkipBlocks']
         end
 
         def offensive?(node)
--- a/lib/rubocop/cop/rspec/described_class.rb
+++ b/lib/rubocop/cop/rspec/described_class.rb
@@ -33,7 +33,6 @@ module RuboCop
       #   end
       #
       class DescribedClass < Cop
-        include RuboCop::RSpec::TopLevelDescribe
         include ConfigurableEnforcedStyle
 
         DESCRIBED_CLASS = 'described_class'
@@ -48,13 +47,16 @@ module RuboCop
 
         def_node_matcher :scope_changing_syntax?, '{def class module}'
 
+        def_node_matcher :described_constant, <<-PATTERN
+          (block $(send _ :describe $(const ...)) (args) $_)
+        PATTERN
+
         def on_block(node)
-          # In case the explicit style is used, we needs to remember what's
-          # being described. Thus, we use an ivar for @described_class.
+          # In case the explicit style is used, we need to remember what's
+          # being described.
           describe, @described_class, body = described_constant(node)
 
           return if body.nil?
-          return unless top_level_describe?(describe)
 
           find_usage(body) do |match|
             add_offense(
@@ -111,8 +113,13 @@ module RuboCop
         end
 
         def offensive?(node)
+          _, nearest_described_class, _ = node.each_ancestor(:block)
+            .map { |ancestor| described_constant(ancestor) }
+            .find(&:itself)
+
           if style == :described_class
-            node.eql?(@described_class)
+            node == nearest_described_class &&
+              node.parent != nearest_described_class.parent
           else
             node.send_type? && node.method_name == :described_class
           end
--- a/lib/rubocop/cop/rspec/described_class.rb
+++ b/lib/rubocop/cop/rspec/described_class.rb
@@ -48,21 +48,18 @@ module RuboCop
         def_node_matcher :scope_changing_syntax?, '{def class module}'
 
         def_node_matcher :described_constant, <<-PATTERN
-          (block $(send _ :describe $(const ...)) (args) $_)
+          (block (send _ :describe $(const ...)) (args) $_)
         PATTERN
 
         def on_block(node)
           # In case the explicit style is used, we need to remember what's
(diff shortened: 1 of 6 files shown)


lib/rubocop/cop/rspec/described_class.rb
```diff
@@ -107,16 +106,84 @@ def skippable_block?(node)
...
+          [*namespace[0, intersection], *const]
+        end
+
+        # @param node [RuboCop::AST::Node]
+        # @return [Array<Symbol>]
+        # @example
+        #   const_name(s(:const, nil, :C))                # => [:C]
+        #   const_name(s(:const, s(:const, nil, :M), :C)) # => [:M, :C]
+        #   const_name(s(:const, s(:cbase), :C))          # => [nil, :C]
+        def const_name(node)
```
pirj: There is a `const_name` method in `Node`, but it ignores the fact that a node is defined with `cbase` completely.

lib/rubocop/cop/rspec/described_class.rb
```diff
@@ -48,19 +47,19 @@ class DescribedClass < Cop
 
         def_node_matcher :scope_changing_syntax?, '{def class module}'
 
+        def_node_matcher :described_constant, <<-PATTERN
```
pirj: Moved from `TopLevelDescribe`, it was exclusively used in this class.

lib/rubocop/cop/rspec/described_class.rb
```diff
@@ -107,16 +106,84 @@ def skippable_block?(node)
...
+          else
+            [*const_name(namespace), name]
+          end
+        end
+
+        # @param node [RuboCop::AST::Node]
+        # @return [Array<Symbol>]
+        # @example
+        #   namespace(node) # => [:A, :B, :C]
+        def namespace(node)
```
pirj: There is a similar `parent_module_name` in `Node`, but it can't be used as a substitute.

lib/rubocop/cop/rspec/described_class.rb
```diff
@@ -107,16 +106,84 @@ def skippable_block?(node)
...
+          node
+            .each_ancestor(:class, :module)
+            .reverse_each
+            .flat_map { |ancestor| ancestor.defined_module_name.split('::') }
+            .map(&:to_sym)
+        end
```
pirj: It looks as over-engineering to me as well.
I'd rather convert both example group's constant and suspect constant to their fully namespaced variants, and would find and collapse intersections between those strings. However, it doesn't seem to be much simpler than those several methods above.

spec/rubocop/cop/rspec/described_class_spec.rb
```diff
@@ -192,43 +198,60 @@ class Foo
       RUBY
     end
 
-    it 'checks for the use of described class with module' do
-      pending
```
pirj: Yay, last pending spec!

pirj:
@bquorning @Darhazer @dgollahon Is there anything that needs to be adjusted here?

spec/rubocop/cop/rspec/described_class_spec.rb
```diff
@@ -37,7 +37,7 @@
                   ^^^^^^^ Use `described_class` instead of `MyClass`.
           end
 
-          before(:each) do
+          before do
```
bquorning: Don’t we have a cop checking for unnecessary arguments to hooks?

spec/rubocop/cop/rspec/described_class_spec.rb
```diff
@@ -179,16 +199,45 @@ class Foo
...
+      expect_correction(<<-RUBY)
+        module MyNamespace
+          describe MyClass do
+            subject { described_class }
+          end
+        end
+      RUBY
+    end
+
+    it 'flags the use of described class with nested namespace' do
```
bquorning: 😮 This is awesome!

spec/rubocop/cop/rspec/described_class_spec.rb
pirj: ¯\_(ツ)_/¯

pirj:
Thank you!

spec/rubocop/cop/rspec/described_class_spec.rb
bquorning: Ahh, this is heredoc’ed Ruby code, so it’s only checked by the `DescribedClass` cop.

pirj:
Some food for the thought https://github.com/rspec/rspec-core/issues/1610#issuecomment-515507697
>  in RSpec 4 we should just remove described_class, we kind of consider it an anti-pattern (because using it means the class must be loaded when the file is evaluated rather than when it's run) and we thus don't recommend it, and it causes this kind of ambiguity.

https://github.com/rspec/rspec-core/issues/1610#issuecomment-46236668
> describe [1, 2, 3] in RSpec 2.x would cause described_class to return the array), the method is misnamed. It's more like described_thing -- which is a class if you pass a class or whatever object you pass

bquorning:
interesting indeed. Should we open an issue on the rspec style guide for this discussion?

pirj:
Good point, I'll take care of this.

pirj:
@bquorning @Darhazer https://github.com/rubocop-hq/rspec-style-guide/issues/104

PR #789, opened 2019-07-23
Add ImplicitBlockExpectation cop

pirj:
Based on (soon to be merged) https://rspec.rubystyle.guide/#implicit-block-expectations

The implicit syntax is discouraged to use by RSpec Core team and the majority of voters.

There were no good arguments for using the syntax except for brevity and avoiding repetition, but there are better options to achieve the same goal, e.g. by extracting the lengthy block to methods (instead of putting it inside the lambda).

References:
https://blog.rubystyle.guide/rspec/2019/07/17/rspec-implicit-block-syntax.html
https://www.reddit.com/r/ruby/comments/cejl3q/call_for_discussion_rspec_implicit_block/
https://lobste.rs/s/e8yxmd/call_for_discussion_rspec_implicit_block
https://github.com/rubocop-hq/rspec-style-guide/issues/76
https://github.com/rubocop-hq/rspec-style-guide/pull/103

---

* [x] Added tests.

Tested on a 10k+ files repo, no false offences raised.


lib/rubocop/cop/rspec/implicit_block_expectation.rb
```diff
@@ -0,0 +1,45 @@
...
+      #   it { is_expected.to change(something).to(new_value) }
+      #
+      #   # good
+      #   it 'changes something to a new value' do
+      #     expect { do_something }.to change(something).to(new_value)
+      #   end
+      class ImplicitBlockExpectation < Cop
+        MSG = 'Avoid implicit block expectations.'
+
+        def_node_matcher :implicit_block_expectation, <<-PATTERN
```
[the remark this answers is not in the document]
pirj: Nice! Can't stop learning from you.

There's one minor thing I've noticed about `block_pattern` though, it's defined as `"(block (send {(const {nil? cbase} :RSpec) nil?} {:subject :subject!} ...) ...)"`, however `RSpec.subject` is not a valid syntax.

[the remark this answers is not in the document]
Darhazer: Yup we need to improve the language module so sub-modules define whether they work on RSpec or only on nil receiver

pirj:
It could be written with the use of `def_node_search`:
```ruby
        def_node_matcher :lambda?, <<-PATTERN
          {
            (send (const nil? :Proc) :new)
            (send nil? :proc)
            (send nil? :lambda)
          }
        PATTERN
        def_node_matcher :lambda_block, '$(block #lambda? ...)'
        def_node_search :subject_block, Subject::ALL.block_pattern

        def on_block(node)
          subject_block(node) do |block_node|
            lambda_block(block_node.body) { |lambda| add_offense(lambda) }
          end
        end
```

I'm wary with `def_node_search` knowing its recursive nature, but couldn't find a case when it fails.

Which do you prefer, @Darhazer, the current or the one above?

Darhazer:
def_node_search will perform the recursive search on each block. If you go in this direction, you should limit the search only in the top level describe, to avoid wasting CPU cycles. 
I'm ok with the previous version though

lib/rubocop/cop/rspec/implicit_block_expectation.rb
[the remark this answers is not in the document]
Darhazer: I imagined just using node methods, once block is confirmed to be a subject; and avoid the patterns at all, but this version is nice and short

pirj:
Thanks, pushed the former solution without `def_node_search` and squashed the commits.

pirj:
> Let fist merge the style guide though
Agree. 👍 
Looking forward to your approval of the style guide pull request.

Darhazer:
Just noticed that the cop flags `subject { -> ... }` while the guide only suggests avoiding `is_expected` with such a subject ;)

pirj:
Fair enough.
I couldn't find any good reason to put a lambda to `subject` other than to use `is_expected`, but it's better to be explicit on what is detected, and the offence message might be confusing when attached to `subject` definition as opposed to `is_expected`.

I'll take care of that, will add detection of `is_expected` usages in the scope where `subject` is defined with a lambda.

spec/rubocop/cop/rspec/implicit_block_expectation_spec.rb
```diff
@@ -0,0 +1,127 @@
...
+      end
+    RUBY
+  end
+
+  it 'flags with `should` as implicit subject' do
+    expect_offense(<<-RUBY)
+      describe do
+        subject { -> { boom } }
+        it { should change { something }.to(new_value) }
+             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ Avoid implicit block expectations.
```
pirj: The range is not ideal, but I decided to avoid complicating the cop for the sake of highlighting `should` itself.
The difference is that `is_expected` is a self-contained `send`, while for `should` it's the whole `should(change { something }.to(new_value))`.

lib/rubocop/cop/rspec/implicit_block_expectation.rb
```diff
@@ -0,0 +1,66 @@
...
+          (block _ _args $(block #lambda? ...))
+        PATTERN
+
+        def_node_matcher :implicit_expect, <<-PATTERN
+          $(send nil? {:is_expected :should :should_not} ...)
+        PATTERN
+
+        def on_send(node)
+          implicit_expect(node) do |implicit_expect|
+            subject = nearest_subject(implicit_expect)
```
pirj: We're finding the nearest subject, and only then checking if it's a lambda subject.
Check [this example](https://github.com/rubocop-hq/rubocop-rspec/pull/789/files#diff-3df60334fb5a381a9d344ade6b46ad63R92).

lib/rubocop/cop/rspec/implicit_block_expectation.rb
```diff
@@ -0,0 +1,66 @@
...
+            add_offense(implicit_expect) if lambda_subject?(subject)
+          end
+        end
+
+        private
+
+        def nearest_subject(node)
+          node
+            .each_ancestor(:block)
+            .lazy
```
pirj: `lazy` is here to avoid mapping/selecting through all of the ancestors just to find the nearest subject.

lib/rubocop/cop/rspec/implicit_block_expectation.rb
```diff
@@ -0,0 +1,66 @@
...
+            .select { |block_node| multi_statement_example_group?(block_node) }
+            .map { |block_node| find_subject(block_node) }
+            .find(&:itself)
+        end
+
+        def multi_statement_example_group?(node)
+          example_group_with_body?(node) && node.body.begin_type?
```
pirj: Non-`begin_type` blocks (example groups) are no interest for us, they don't have other blocks than those we came from, and we're searching for `subject`. If there are at least two statements in the block, it's a `begin_type`.

lib/rubocop/cop/rspec/implicit_block_expectation.rb
```diff
@@ -0,0 +1,66 @@
...
+            (send nil? :proc)
+            (send nil? :lambda)
+          }
+        PATTERN
+
+        def_node_matcher :lambda_subject?, <<-PATTERN
+          (block _ _args $(block #lambda? ...))
+        PATTERN
+
+        def_node_matcher :implicit_expect, <<-PATTERN
```
pirj: Shamelessly borrowed from `ImplicitSubject`. Not sure if it makes sense to extract this with just two usages.

lib/rubocop/cop/rspec/implicit_block_expectation.rb
```diff
@@ -0,0 +1,66 @@
...
+            .map { |block_node| find_subject(block_node) }
+            .find(&:itself)
+        end
+
+        def multi_statement_example_group?(node)
+          example_group_with_body?(node) && node.body.begin_type?
+        end
+
+        def find_subject(block_node)
+          block_node.body.children.find { |send_node| subject?(send_node) }
```
Darhazer: generally `child_nodes` is preferred over `children`, as it ignores non-Node objects

lib/rubocop/cop/rspec/implicit_block_expectation.rb
```diff
@@ -0,0 +1,66 @@
...
+
+        def_node_matcher :lambda?, <<-PATTERN
+          {
+            (send (const nil? :Proc) :new)
+            (send nil? :proc)
+            (send nil? :lambda)
+          }
+        PATTERN
+
+        def_node_matcher :lambda_subject?, <<-PATTERN
```
Darhazer: You are interested in checking the block body. You can just check whether (block #lambda? ...) matches the subject.body and remove the outer part of the matcher

Darhazer: Also you are not using the captured node

pirj: Nice catch, thank you!

lib/rubocop/cop/rspec/implicit_block_expectation.rb
pirj: Fine last touch, thanks 👍

lib/rubocop/cop/rspec/implicit_block_expectation.rb
```diff
@@ -0,0 +1,64 @@
...
+
+        private
+
+        def nearest_subject(node)
+          node
+            .each_ancestor(:block)
+            .lazy
+            .select { |block_node| multi_statement_example_group?(block_node) }
+            .map { |block_node| find_subject(block_node) }
+            .find(&:itself)
```
bquorning: This line is equivalent to `.reject(&:nil?).first`, right? Just trying to understand.

pirj: Yes, I think those are equivalent.

pirj:
Thank you!

bquorning:
And thank you for the new cop :-)

PR #797, opened 2019-08-03
Fix autocorrect of multi-line docstrings in RSpec/ExampleWording

### LINKED ISSUE #792: RSpec/ExampleWording autocorrect breaking code on multiline example strings
When I run `rubocop --require rubocop-rspec --only RSpec/ExampleWording --auto-correct` against the code below:

```
RSpec.describe 'Testing' do
  it 'should be flagged '\
     'but correctly fixed' do
  end
end
```

I get:
```
RSpec.describe 'Testing' do
  it 'is flagged '\ 'but correctly fixed' do
  end
end
```

Which is invalid syntax:
```
bundle exec rspec test_spec.rb                                                                                                                                    13:30:12

An error occurred while loading ./test_spec.rb.
Failure/Error: __send__(method, file)

SyntaxError:
  /Users/rgibim/workspace/ruby-scripts/test-rubocop-rspec/test_spec.rb:2: syntax error, unexpected $undefined, expecting keyword_end
    it 'is flagged '\ 'but correctly fixed' do
                    ^
  /Users/rgibim/workspace/ruby-scripts/test-rubocop-rspec/test_spec.rb:2: syntax error, unexpected keyword_do_block, expecting end-of-input
  ...ged '\ 'but correctly fixed' do
  ...                             ^~
```

pirj:
\\-separated multi-line docstrings were incorrectly combined during auto-correction, e.g.:

```ruby
it 'should be flagged '\
   'but correctly fixed' do
end
```

was corrected to:

```ruby
it 'is flagged '\ 'but correctly fixed' do
end
```

fixes #792

### Not Done

To be honest, there's still one example left that won't be properly auto-corrected, specifically (notice both single and double braces):

```ruby
it 'should do something '\
    "with #{object}" do
end
```

---

* [x] Added tests.

### DIFF
--- a/lib/rubocop/cop/rspec/example_wording.rb
+++ b/lib/rubocop/cop/rspec/example_wording.rb
@@ -53,27 +53,30 @@ module RuboCop
           end
         end
 
-        def autocorrect(range)
-          ->(corrector) { corrector.replace(range, replacement_text(range)) }
+        def autocorrect(node)
+          lambda do |corrector|
+            corrector.replace(docstring(node), replacement_text(node))
+          end
         end
 
         private
 
         def add_wording_offense(node, message)
-          expr = node.loc.expression
-
-          docstring =
-            Parser::Source::Range.new(
-              expr.source_buffer,
-              expr.begin_pos + 1,
-              expr.end_pos - 1
-            )
-
-          add_offense(docstring, location: docstring, message: message)
+          add_offense(node, location: docstring(node), message: message)
         end
 
-        def replacement_text(range)
-          text = range.source
+        def docstring(node)
+          expr = node.loc.expression
+
+          Parser::Source::Range.new(
+            expr.source_buffer,
+            expr.begin_pos + 1,
+            expr.end_pos - 1
+          )
+        end
+
+        def replacement_text(node)
+          text = text(node)
 
           if text =~ SHOULD_PREFIX
             RuboCop::RSpec::Wording.new(
@@ -83,6 +86,19 @@ module RuboCop
             ).rewrite
           else
             text.sub(IT_PREFIX, '')
+          end
+        end
+
+        # Recursive processing is required to process nested dstr nodes
+        # that is the case for \-separated multiline strings with interpolation.
+        def text(node)
+          case node.type
+          when :dstr
+            node.node_parts.map { |child_node| text(child_node) }.join
+          when :str
+            node.value
+          when :begin
+            node.source
           end
         end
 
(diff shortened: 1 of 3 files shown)


lib/rubocop/cop/rspec/example_wording.rb
```diff
@@ -53,27 +53,30 @@ def on_block(node)
...
+          lambda do |corrector|
+            corrector.replace(docstring(node), replacement_text(node))
+          end
         end
 
         private
 
         def add_wording_offense(node, message)
-          expr = node.loc.expression
+          add_offense(node, location: docstring(node), message: message)
```
pirj: It's a bummer that `docstring(node)` is called twice on offending nodes, once when we're adding an offence, and for the second time in auto-correction, but this doesn't seem to be a big deal.

lib/rubocop/cop/rspec/example_wording.rb
```diff
@@ -53,27 +53,30 @@ def on_block(node)
           end
         end
 
-        def autocorrect(range)
-          ->(corrector) { corrector.replace(range, replacement_text(range)) }
+        def autocorrect(node)
+          lambda do |corrector|
```
bquorning: It’s a bit unfortunate that when switching from a one-line to multi-line block, we have to also change how the `lambda` is defined. That’s not diff friendly…

lib/rubocop/cop/rspec/example_wording.rb
bquorning: If anyone has a code base with enough offenses that this will cause a performance problem, I’m sure they will disable the cop 😉

pirj: That should be a codebase of an extraordinary size! ;)
```ruby
> t0=Time.now
> 10_000_000.times { Parser::Source::Range.new(expr.source_buffer, expr.begin_pos + 1, expr.end_pos - 1) }
> puts Time.now - t0
5.982299
```
And we're calling it for the second time only when auto-correcting.

lib/rubocop/cop/rspec/example_wording.rb
pirj: Indeed it would be very nice if git knew language semantics. Just checked all four diff algorithms, none showed anything that would hint that `range` became `node` in the signature, `docstring(node)` in method call. /shrug

pirj:
@Darhazer Rebased the code, please take a look.

lib/rubocop/cop/rspec/example_wording.rb
```diff
@@ -86,6 +89,27 @@ def replacement_text(range)
           end
         end
 
+        def text(node)
+          if node.dstr_type?
```
Darhazer: generally, if you move the `join` to the result of `flatten_node`, so it always returns a string, you can remove this if/else part

pirj: Great tip!
It's so much better with one method less 👍

PR #800, opened 2019-08-06
Ignore `described_class` being used as part of a constant

### LINKED ISSUE #799: error .. RSpec/DescribedClass cop was inspecting .. describe(CONSTANT)
Upgrading PaperTrail from 1.33.0 to 1.35.0, got this error:

```
An error occurred while RSpec/DescribedClass cop was 
inspecting .../spec/paper_trail/version_number_spec.rb:6:2.
```

In [`version_number_spec.rb`](https://github.com/paper-trail-gem/paper_trail/blob/da9c743831ff7b948680ee0ef847ad8516f1a206/spec/paper_trail/version_number_spec.rb) our `described_class` is a module named `VERSION`. 🤔 I admit, it's unusual (I didn't name it) but I think we can agree that rubocop should not crash :smile:

```
module PaperTrail
  ::RSpec.describe VERSION do
    describe "STRING" do
      it "joins the numbers into a period separated string" do
        expect(described_class::STRING).to eq(
          ...
```

Please see also [paper_trail/version_number.rb](https://github.com/paper-trail-gem/paper_trail/blob/da9c743831ff7b948680ee0ef847ad8516f1a206/lib/paper_trail/version_number.rb) if necessary.

I'm happy to work on this issue if someone can point me in the right direction. :heart:

pirj:
Sometimes, `described_class` is used as part of the constant, e.g. a namespace in the constant being analyzed, e.g.:

    expect(described_class::STRING).to eq('1.1.1')

In this case the constant was still detected as offensive, while it should not be.

Fixes #799

---

* [x] Added tests.


pirj:
@bquorning Please take a look.

lib/rubocop/cop/rspec/described_class.rb
```diff
@@ -143,8 +145,7 @@ def full_const_name(node)
         #   collapse_namespace([:A, :B], [nil, :C)     # => [nil, :C]
         #   collapse_namespace([:A, :B], [nil, :B, :C) # => [nil, :B, :C]
         def collapse_namespace(namespace, const)
-          return const if namespace.empty?
-          return const if const.first.nil?
+          return const if namespace.empty? || const.first.nil?
```
bquorning: I suspect this change is made _only_ to “please the cop”. I would rather increase the max file length than make unnecessary changes.

lib/rubocop/cop/rspec/described_class.rb
```diff
@@ -102,11 +106,7 @@ def scope_change?(node)
         end
 
         def skippable_block?(node)
-          node.block_type? && !rspec_block?(node) && skip_blocks?
-        end
-
-        def skip_blocks?
-          cop_config['SkipBlocks']
+          node.block_type? && !rspec_block?(node) && cop_config['SkipBlocks']
```
bquorning: This change also looks like a ClassLength “fix”, but it ends up making the code _less_ clear, in my opinion.

pirj:
@bquorning Please take another look. I've introduced `.rubocop_todo.yml` with 104 lines for `Metrics/ClassLength`.

.rubocop_todo.yml
```diff
@@ -0,0 +1,12 @@
...
+# on 2019-08-06 21:29:02 +0300 using RuboCop version 0.70.0.
+# The point is for the user to remove these configuration records
+# one by one as the offenses are removed from the code base.
+# Note that changes in the inspected code, or installation of new
+# versions of RuboCop, may require this file to be generated again.
+
+# Offense count: 1
+# Configuration parameters: CountComments.
+Metrics/ClassLength:
+  Max: 104
```
bquorning: @Darhazer Is this ok with you?

PR #804, opened 2019-08-11
Update RuboCop and use children matching

pirj:
[Children matching](https://github.com/rubocop-hq/rubocop/issues/6965) was introduced in RuboCop 0.68.0.
This is a conservative update of RuboCop dependency to 0.68.1 (that has [some minor bug fixes compared to 0.68.0](https://github.com/rubocop-hq/rubocop/blob/master/CHANGELOG.md#0681-2019-04-30)).

A couple of our cops were updated to use children matching.
It allowed to convert iterations with that new pattern syntax, and reduce the amount of code.

Focus cop is intentionally left unchanged, there are improvements for it in #777 that are making use of the new syntax as well.

As per https://github.com/rubocop-hq/rubocop-rspec/pull/777#discussion_r311897354 and https://github.com/rubocop-hq/rubocop-rspec/pull/773#issuecomment-495966090.

It seems I just had a stroke of image syndrome.

---

* [-] Added tests.

### DIFF
--- a/lib/rubocop/cop/rspec/factory_bot/attribute_defined_statically.rb
+++ b/lib/rubocop/cop/rspec/factory_bot/attribute_defined_statically.rb
@@ -38,7 +38,7 @@ module RuboCop
           def on_block(node)
             factory_attributes(node).to_a.flatten.each do |attribute|
               next unless offensive_receiver?(attribute.receiver, node)
-              next if proc?(attribute) || association?(attribute)
+              next if proc?(attribute) || association?(attribute.first_argument)
 
               add_offense(attribute)
             end
@@ -72,14 +72,7 @@ module RuboCop
             value_matcher(attribute).to_a.all?(&:block_pass_type?)
           end
 
-          def association?(attribute)
-            argument = attribute.first_argument
-            argument.hash_type? && factory_key?(argument)
-          end
-
-          def factory_key?(hash_node)
-            hash_node.keys.any? { |key| key.sym_type? && key.value == :factory }
-          end
+          def_node_matcher :association?, '(hash <(pair (sym :factory) _) ...>)'
 
           def autocorrect_replacing_parens(node)
             left_braces, right_braces = braces(node)
--- a/lib/rubocop/cop/rspec/multiple_expectations.rb
+++ b/lib/rubocop/cop/rspec/multiple_expectations.rb
@@ -50,14 +50,23 @@ module RuboCop
 
         MSG = 'Example has too many expectations [%<total>d/%<max>d].'
 
-        def_node_search :with_aggregate_failures?, '(sym :aggregate_failures)'
-        def_node_search :disabled_aggregate_failures?, <<-PATTERN
-          (pair (sym :aggregate_failures) (false))
+        def_node_matcher :aggregate_failures?, <<-PATTERN
+          (block {
+              (send _ _ <(sym :aggregate_failures) ...>)
+              (send _ _ ... (hash <(pair (sym :aggregate_failures) true) ...>))
+            } ...)
+        PATTERN
+
+        def_node_matcher :aggregate_failures_present?, <<-PATTERN
+          (block {
+              (send _ _ <(sym :aggregate_failures) ...>)
+              (send _ _ ... (hash <(pair (sym :aggregate_failures) _) ...>))
+            } ...)
         PATTERN
 
         def_node_matcher :expect?, Expectations::ALL.send_pattern
         def_node_matcher :aggregate_failures_block?, <<-PATTERN
-          (block (send _ :aggregate_failures ...) ...)
+          (block (send nil? :aggregate_failures ...) ...)
         PATTERN
 
         def on_block(node)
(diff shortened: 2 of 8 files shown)


lib/rubocop/cop/rspec/multiple_expectations.rb
```diff
@@ -50,14 +50,23 @@ class MultipleExpectations < Cop
...
+              (send _ _ <(sym :aggregate_failures) ...>)
+              (send _ _ ... (hash <(pair (sym :aggregate_failures) true) ...>))
+            } ...)
+        PATTERN
+
+        def_node_matcher :aggregate_failures_present?, <<-PATTERN
+          (block {
+              (send _ _ <(sym :aggregate_failures) ...>)
+              (send _ _ ... (hash <(pair (sym :aggregate_failures) _) ...>))
+            } ...)
```
pirj: Appreciate a hint how to collapse the two above node patterns into one with [`%` syntax](https://github.com/rubocop-hq/rubocop/blob/master/lib/rubocop/node_pattern.rb#L69).

Tried replacing them with:
```ruby
        def_node_matcher :aggregate_failures?, <<-PATTERN
          (block {
              (send _ _ <(sym :aggregate_failures) ...>)
              (send _ _ ... (hash <(pair (sym :aggregate_failures) %) ...>))
            } ...)
        PATTERN
```

and using as `aggregate_failures?(node, true)`, `aggregate_failures?(node, :true)`, and `aggregate_failures?(node, 'true')` - none of those variants worked.

Darhazer: How about just capturing the argument, then in the block that checks for aggregate_failures you just need to check whether the node matched and if there is a captured value - it is true?

pirj: It turns out to become quite hacky for two reasons:
 - I have to also capture from the other `{}` branch
 - in the latter branch it's expected to be a boolean, for the former it will be a symbol

so I have to check it like that:
```ruby
          aggregate_failures = aggregate_failures?(node_with_aggregate_failures)
          aggregate_failures.true_type? || aggregate_failures.sym_type?
```

it works, but it's hard to understand.
Tried reading RuboCop's own usages of `%`-syntax, still can't understand how it works, and if it's possible to pass it an `_`.

I'd rather keep this as is, it's easier to spot the redundancy like that and address in the future. WDYT?


On the other note, I added a line to the changelog, but there were no user-facing changes, so just mentioning that RuboCop version dependency has been updated.

pirj:
@bquorning @Darhazer @dgollahon Is there anything that needs to be addressed here?

Speaking of `aggregate_failures?`/`aggregate_failures_present?` - I would rather keep two very similar patterns in hope to collapse them into one in the moment of enlightenment rather than to build support code around an improperly collapsed one to work around the difference in its application.

pirj:
Fixed CHANGELOG conflict.

bquorning:
@dgollahon Would you have time to review this PR?

pirj:
@dgollahon Appreciate if you could take a look. Those changes are very similar to what you did in #777.

lib/rubocop/cop/rspec/factory_bot/attribute_defined_statically.rb
```diff
@@ -72,14 +72,7 @@ def proc?(attribute)
...
 
-          def association?(attribute)
-            argument = attribute.first_argument
-            argument.hash_type? && factory_key?(argument)
-          end
-
-          def factory_key?(hash_node)
-            hash_node.keys.any? { |key| key.sym_type? && key.value == :factory }
-          end
+          def_node_matcher :association?, '(hash <(pair (sym :factory) _) ...>)'
```
bquorning: ❤️ It’s so concise! I just hope people (we) will find it readable in the future.

lib/rubocop/cop/rspec/multiple_expectations.rb
pirj: @marcandre Do you happen to know if it's possible to pass a boolean or an `any` as a node matcher argument (to substitute `%`)? I think I tried with booleans and failed.

marcandre: Super interesting question. Turns out you won't be able to because the check is that `node == your_arg`. We should revert that order for more flexibility, so it is possible to define an any matcher (`ANY = Object.new.tap {|o| def o.==(_); true; end}`)

pirj: Got it, thanks!
Please keep in mind it's a pretty narrow case with just one potential usage I know if. Other cops here in `rubocop-rspec` using `%` provide symbols, and it's quite sufficient for them.

And this cop works as well, I just keep it somewhere in my mind's dirty corner that it might be improved.

marcandre: It's a simple change, but so much more powerful `https://github.com/rubocop-hq/rubocop-ast/pull/31`. You could have `ANY = ->(_) { true }` and pass that.

pirj: Two node patterns easily collapsed into one with this new generic parameters https://github.com/rubocop-hq/rubocop-rspec/pull/930

PR #805, opened 2019-08-11
Add ContextMethod cop

pirj:
As discussed in #417 is would be good to have a cop to check for the combination of context and a method name.

Based on #751, addressed code review notes.

---

* [x] Added tests.


pirj:
@bquorning @Darhazer @dgollahon Is there anything that needs to be addressed here?

spec/rubocop/cop/rspec/context_method_spec.rb
```diff
@@ -0,0 +1,41 @@
...
+      end
+    RUBY
+
+    expect_correction(<<-RUBY)
+      describe '.foo_bar' do
+      end
+    RUBY
+  end
+
+  it 'finds context with `#` at the beginning' do
```
bquorning:
```suggestion
  it 'flags context with `#` at the beginning' do
```

pirj:
@bquorning Done.

PR #810, opened 2019-08-18
Remove redundant files from gem distribution

pirj:
It's hard to tell why spec files were shipped with a gem, but that doesn't seem to be necessary.

Shipping of Gemfile (that is intended to be used for local development), and especially gemspec file is unnecessary as well.

Rakefile is not useful except for local RuboCop-RSpec development.

There doesn't seem to be any mention of `test_files` option in [gem specification reference](https://guides.rubygems.org/specification-reference/#extra_rdoc_files),
most probably it's a legacy option.

---

* [-] Added tests.

### DIFF
--- a/rubocop-rspec.gemspec
+++ b/rubocop-rspec.gemspec
@@ -25,13 +25,10 @@ Gem::Specification.new do |spec|
 
   spec.require_paths = ['lib']
   spec.files = Dir[
-    '{config,lib,spec}/**/*',
-    '*.md',
-    '*.gemspec',
-    'Gemfile',
-    'Rakefile'
+    'lib/**/*',
+    'config/default.yml',
+    '*.md'
   ]
-  spec.test_files = spec.files.grep(%r{^spec/})
   spec.extra_rdoc_files = ['MIT-LICENSE.md', 'README.md']
 
   spec.metadata = {


rubocop-rspec.gemspec
```diff
@@ -25,13 +25,10 @@ Gem::Specification.new do |spec|
...
-    '{config,lib,spec}/**/*',
-    '*.md',
-    '*.gemspec',
-    'Gemfile',
-    'Rakefile'
+    'lib/**/*',
+    'config/default.yml',
+    '*.md'
   ]
-  spec.test_files = spec.files.grep(%r{^spec/})
```
bquorning: I found this: https://github.com/rubygems/rubygems/blob/40804426dbb5781817b90f37fd6cce3b9e603558/lib/rubygems/specification.rb#L687-L697

I’m not sure what it should be used for 🤷‍♂

pirj: Found this https://github.com/rubygems/rubygems/blob/77942c17384ed4a3f194f50d6e8dc64ffedca90c/History.txt#L1466

And also this back from 2014 https://github.com/rubygems/guides/issues/90#issuecomment-49213972

In relation to why the setting of this option still works is:
> Starting in RubyGems 2.0, a Specification can hold arbitrary metadata.

PR #819, opened 2019-09-04
Improve autocorrect for Capybara/CurrentPathExpectation

### LINKED ISSUE #798: Auto-correct of offence Capybara/CurrentPathExpectation is not safe
After running the `rubocop -a`, this offence was autocorrected but caused tests to fail, where, imo, they should not.

Diff made by --autocorrect:

```
- expect(page.current_path).to eq checkout_path(checkout_id)
+ expect(page).to have_current_path checkout_path(checkout_id)
```

RSpec fails with:

```
expected "/checkouts/1fd28388-0a1e-403a-915b-ac6d5aa7e07d?access_token=51e8564762150170aa81994eca6c9792" to equal "/checkouts/1fd28388-0a1e-403a-915b-ac6d5aa7e07d"
```

Seems like `have_current_path` is using some method that includes the querystring on the path, while `page.current_path` does not.

Onumis:
`expect(page.current_path).to eq(EXPECTED_VALUE)` is not directly equivalent to `expect(page).to have_current_path(EXPECTED_VALUE)`
In particular, `have_current_path` with no options will include the querystring on the path, while `page.current_path` does not.

This fix ensures the option `ignore_query: true` is set on `have_current_path` expect when `EXPECTED_VALUE` is a string or regexp.
In those cases, it should be safe to not use this param.

Fixes #798

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 * Add `RSpec/ContextMethod` cop, to detect method names in `context`. ([@geniou][])
 * Update RuboCop dependency to 0.68.1 with support for children matching node pattern syntax. ([@pirj][])
 * Add `RSpec/EmptyLineAfterExample` cop to check that there is an empty line after example blocks. ([@pirj][])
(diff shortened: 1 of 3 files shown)


Onumis:
@Darhazer done! :D

PR #823, opened 2019-09-23
Add RSpec/EmptyLineAfterExample cop

### LINKED ISSUE #822: EmptyLine rules for `it` and `context` blocks
There are `EmptyLine` rules for several test constructs like `describe` and `subject`, but not for `it` blocks and `context` blocks. It would be nice to be able to enforce consistency around these as well. For example:

```rb
# bad
it 'does a thing' do
end
it 'does another thing' do
end

# good
it 'does a thing' do
end

it 'does another thing' do
end
```

Similarly for `context` blocks:

```rb
# bad
context 'when it is true' do
  it 'does a thing' do
  end
end
context 'when it is false' do
  it 'does another thing' do
  end
end

# good
context 'when it is true' do
  it 'does a thing' do
  end
end

context 'when it is false' do
  it 'does another thing' do
  end
end
```

pirj:
Fixes #822
Related to (fixes?) #272

---

* [x] Added tests.


pirj:
@Darhazer I understand [what you mean](https://github.com/rubocop-hq/rubocop-rspec/issues/272#issuecomment-400326666) by making an empty line between examples configurable.

I still have some doubts on how should the examples below should be treated:
```ruby
it { ... }
it "does something" do
  # ...
end
```

```ruby
it "does something" do
  # ...
end
it { ... }
```

Should the option allow several consecutive one-liners to be written without separating empty lines, but always enforce an empty line between mixed style examples?
Is `AllowConsecutiveOneLiners` a good name, and should it be the default style?

Darhazer:
> Should the option allow several consecutive one-liners to be written without separating empty lines, but always enforce an empty line between mixed style examples?

Yes

> Is AllowConsecutiveOneLiners a good name, and should it be the default style?

Yes, for both

pirj:
@Darhazer Fixed, thanks for the review.

PR #825, opened 2019-09-26
Use rubocop-dev gem to untangle dev and runtime dependencies

pirj:
For testing RuboCop RSpec with the earliest RuboCop version supported (for cases when users avoid updating RuboCop dependency to evade new offences), and still be able to use the new additions to spec support code, separate spec support code from RuboCop runtime code.

Related: #815
Based on the idea from https://github.com/rubocop-hq/rubocop-rspec/pull/815#issuecomment-529005581
Extracted: https://github.com/pirj/rubocop-dev

---

* [-] Added tests.
* [ ] Remove temporary commit, use the gem when it's published to RubyGems and owned by RuboCop HQ org

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -2,6 +2,9 @@
 
 source 'https://rubygems.org'
 
+gem 'rubocop', github: 'pirj/rubocop', branch: 'use-rubocop-dev'
+gem 'rubocop-dev', github: 'pirj/rubocop-dev'
+
 gemspec
 
 local_gemfile = 'Gemfile.local'
--- a/rubocop-rspec.gemspec
+++ b/rubocop-rspec.gemspec
@@ -41,6 +41,7 @@ Gem::Specification.new do |spec|
   spec.add_development_dependency 'rack'
   spec.add_development_dependency 'rake'
   spec.add_development_dependency 'rspec', '>= 3.4'
+  spec.add_development_dependency 'rubocop-dev'
   spec.add_development_dependency 'simplecov'
   spec.add_development_dependency 'yard'
 end
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -2,7 +2,7 @@
 
 require 'rubocop'
 
-require 'rubocop/rspec/support'
+require 'rubocop/dev/rspec/support'
 
 if ENV['COVERAGE'] == 'true'
   require 'simplecov'


pirj:
@bquorning Please take a quick look.

pirj:
The reason of "confirm_config_and_documentation" failure is due to [this change](https://github.com/rubocop-hq/rubocop/pull/7295/files#diff-0fc75b42cc2076b1a2e69a39fa34ec6eR47):
```ruby
-      def initialize(config, cop_config, options)
+      def initialize(config, cop_name, cop_config, options)
```

`rubocop-rspec` is not anymore compatible *in development* with edge RuboCop.
[`tasks/cops_documentation.rake`](https://github.com/rubocop-hq/rubocop/tree/master/tasks/cops_documentation.rake) seems to be an idean next candidate for extraction to `rubocop-dev`.

The other builds were not triggered due to that failure.

pirj:
@rubocop-hq/rubocop-core Can you please take a look? More context in https://github.com/rubocop-hq/rubocop-rspec/pull/815#issuecomment-529005581

koic:
Thank you for the suggestion. However, Gemify is doubtful to me.

I have an alternative approach.

RuboCop Rails, RuboCop Performance, and RuboCop Minitest are following the edge by specifying GitHub of RuboCop core in Gemfile.

```diff
 source 'https://rubygems.org'

+git_source(:github) { |repo| "https://github.com/#{repo}.git" }

 gemspec

 gem 'rake'
 gem 'rspec'
+gem 'rubocop', github: 'rubocop-hq/rubocop'
 gem 'rubocop-rspec', '~> 1.29.0'
 gem 'yard', '~> 0.9'
```

https://github.com/rubocop-hq/rubocop-rails/pull/50

RuboCop plugins will eventually need to run on the edge of RuboCop core, this way of specifying GitHub seems reasonable.

This is the technique used by Active Record 3rd party adapter, and this approach works very well.
https://github.com/rsim/oracle-enhanced/blob/942d459b2e05dd4f17840979c8df2a12b530b667/Gemfile#L15

I apologize for the delay in sharing this approach with RuboCop RSpec team.

pirj:
I tend to agree that it would be easier to keep up to date with the most recent RuboCop.
@bquorning appreciate your feedback.

pirj:
@bquorning As far as I understand, the plan for RuboCop starting with 1.0 is not to introduce new enabled cops in minor versions, so if a spec helper/shared example/rake task was introduced in 1.1, it's safe for RuboCop-RSpec to update `rubocop` dependency version and use it without worrying about introducing offences that come from `rubocop` update.

In this case, there's no big difference between minor versions in runtime, and it's possible to use helpers introduced in minor versions, and it's possible to update minor version dependencies to improve specs/daily development tasks.

bbatsov:
Yeah, that's the plan indeed. Unfortunately the related PR got stuck and it's pretty much the only thing left for a 1.0 release.

bquorning:
It would be really nice to somehow re-use the code in tasks/cops_documentation.rake though. Each gem having a _slightly_ different version really rubs my OCD the wrong way :-)

pirj:
We talked with @Darhazer, I will take this task over.

PR #826, opened 2019-09-29
Fix implicit subjects mocked/stubbed not detected

### LINKED ISSUE #818: Rspec/SubjectStub: false negative with implicit subject
Rubocop-rspec does not detect the SubjectStub offenses when using the implicit subject

For example: no offense is reported with the following code, which it's supposed to be
```ruby
describe Foo do

  before do
    allow(subject).to receive(:bar).and_return(baz)
  end

  it 'uses expect twice' do
    expect(subject.bar).to eq(baz)
  end
end
```

On the other hand, the offense can be detected while the subject is explicitly defined.

```ruby
describe Foo do
  subject { Foo.new }

  before do
    allow(subject).to receive(:bar).and_return(baz)
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ Do not stub methods of the object under test.
  end

  it 'uses expect twice' do
    expect(subject.bar).to eq(baz)
  end
end
```

## Rubocop version

```
0.74.0 (using Parser 2.6.4.0, running on ruby 2.4.0 x86_64-darwin17)
```

QQism:
This PR fixes #818.

[Implicit Subject](https://relishapp.com/rspec/rspec-core/v/3-8/docs/subject/implicitly-defined-subject) had not been detected by the `Rspec/SubjectStub`, which reported false negative cases.

As suggested by @pirj, there are solutions to recursively traverse through AST (either upward or downward). However, to the best of my knowledge, they might require a fair amount of changes how the cop is currently working and impact the runtime performance.

I've chosen the somewhat safe fix that reports the defaut `:subject` if there is no named nor unnamed subject declared.

Since it's my first PR, any suggestions and feedback are more than welcome.

---

* [x] Added tests.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## Master (Unreleased)
 
 * Fix documentation rake task to support Rubocop 0.75. ([@nickcampbell18][])
+* Fix `RSpec/SubjectStub` to detect implicit subjects stubbed. ([@QQism][])
 
 ## 1.36.0 (2019-09-27)
 
@@ -455,3 +456,4 @@ Compatibility release so users can upgrade RuboCop to 0.51.0. No new features.
 [@nc-holodakg]: https://github.com/nc-holodakg
 [@onumis]: https://github.com/onumis
 [@nickcampbell18]: https://github.com/nickcampbell18
+[@QQism]: https://github.com/QQism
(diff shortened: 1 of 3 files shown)


QQism:
New push to resolve conflicts 😁

Darhazer:
I feel this way we do unnecessary iterations. The only issue is that when there is no explicit `subject` definition, we never call `find_subject_expectation`, while the `message_expectation?` matcher search not only for whatever name we give, but also for `subject`. 

This is an optimization we can tackle separately though

QQism:
> I feel this way we do unnecessary iterations. The only issue is that when there is no explicit `subject` definition, we never call `find_subject_expectation`, while the `message_expectation?` matcher search not only for whatever name we give, but also for `subject`.
> 
> This is an optimization we can tackle separately though

From my understanding, it happens not only to implicit subjects but also to the unnamed subjects which are out of the scope of this PR.

PR #827, opened 2019-10-03
Rubydoc.info links should use HTTPS

nickcampbell18:
It looks like rubydoc.info does upgrade non-HTTPS connections anyway, but clients should not be sent to the non-secure version to start with.

### DIFF
--- a/config/default.yml
+++ b/config/default.yml
@@ -13,32 +13,32 @@ AllCops:
 RSpec/AlignLeftLetBrace:
   Description: Checks that left braces for adjacent single line lets are aligned.
   Enabled: false
-  StyleGuide: http://www.rubydoc.info/gems/rubocop-rspec/RuboCop/Cop/RSpec/AlignLeftLetBrace
+  StyleGuide: https://www.rubydoc.info/gems/rubocop-rspec/RuboCop/Cop/RSpec/AlignLeftLetBrace
 
 RSpec/AlignRightLetBrace:
   Description: Checks that right braces for adjacent single line lets are aligned.
   Enabled: false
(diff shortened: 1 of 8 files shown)


nickcampbell18:
Ah, [this line](https://github.com/rubocop-hq/rubocop-rspec/blob/master/tasks/cops_documentation.rake#L143) is breaking because of [this change](https://github.com/rubocop-hq/rubocop/commit/95b86f8a704da6e989302f37b057e130a2418c01) in Rubocop 0.75. Let me know how you'd like me to proceed please?

pirj:
Thanks for fixing the links!

Regarding the failure - it seems we'll have to update `rubocop` dependency to 0.7.5 and adjust the code.

nickcampbell18:
> Regarding the failure - it seems we'll have to update rubocop dependency to 0.7.5 and adjust the code.

Would you like me to open a PR for that?

pirj:
I would highly appreciate that.
Let's confirm with @bquorning and @Darhazer to be sure.

Darhazer:
Regarding the failure, I'm a bit on the fence to update - we recently bumped the requirement for 0.68.1. It makes sense to fix the CI to use the minimum required version, or we may check the `arity` of the method in the rake task (or even the Rubocop version) and to have conditional instantiation

Darhazer:
@nickcampbell18 I merged the fix for the documentation, would you please rebase. The build should pass now

nickcampbell18:
We're all good now, thanks!

pirj:
@nickcampbell18 Thanks for persistence with this pull request and taking care of the compatibility/build issue as well. 💯

PR #829, opened 2019-10-03
Fix documentation rake tasks to support Rubocop 0.75

nickcampbell18:
Rubocop 0.75 changed the arity of the [`MessageAnnotator`](https://github.com/rubocop-hq/rubocop/commit/95b86f8a704da6e989302f37b057e130a2418c01), so this commit changes the documentation task to support both the old and new versions.

### DIFF
--- a/spec/rubocop/cop/rspec/cop_spec.rb
+++ b/spec/rubocop/cop/rspec/cop_spec.rb
@@ -21,13 +21,15 @@ RSpec.describe RuboCop::Cop::RSpec::Cop do
 
   let(:fake_cop) do
     stub_const('RuboCop::RSpec', Module.new)
-    # rubocop:disable ClassAndModuleChildren, RSpec/LeakyConstantDeclaration
+    # rubocop:disable Style/ClassAndModuleChildren
+    # rubocop:disable RSpec/LeakyConstantDeclaration
     class RuboCop::RSpec::FakeCop < described_class
       def on_send(node)
         add_offense(node, message: 'I flag everything')
       end
     end
-    # rubocop:enable ClassAndModuleChildren, RSpec/LeakyConstantDeclaration
+    # rubocop:enable Style/ClassAndModuleChildren
+    # rubocop:enable RSpec/LeakyConstantDeclaration
     RuboCop::RSpec::FakeCop
   end
 
--- a/tasks/cops_documentation.rake
+++ b/tasks/cops_documentation.rake
@@ -138,9 +138,32 @@ task generate_cops_documentation: :yard_for_generate_documentation do
   end
   # rubocop:enable Metrics/MethodLength
 
+  def cop_urls(config, cop)
+    rubocop_version = Gem::Version.new(RuboCop::Version::STRING)
+
+    # Since Rubocop v0.75.0 and above, MessageAnnotator#new changed from:
+    #   def initialize(config, cop_config, options)
+    # to:
+    #   def initialize(config, cop_name, cop_config, options)
+    #
+    # Since this library has a loose Rubocop dependency, we select the
+    # right arguments based on the installed version.
+    #
+    # TODO: When Rubocop < 0.75 is no longer supported, remove the second half
+    # of this condition.
+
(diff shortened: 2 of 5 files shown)


nickcampbell18:
Just pushed another commit which checks directly the Rubocop version. I started writing the comment saying "if Rubocop 0.75 it does this...", but I think that makes it weirder that we're checking arity (e.g. if the signature changes again in an upcoming version, this won't find it). Explicitly looking at the version hopefully makes it clearer and more/fewer conditions could be supported in future if needed.

tasks/cops_documentation.rake
```diff
@@ -138,16 +138,38 @@ task generate_cops_documentation: :yard_for_generate_documentation do
...
+    # right arguments based on the installed version.
+    rubocop_version = Gem::Version.new(RuboCop::Version::STRING)
+    urls =
+      if rubocop_version >= Gem::Version.new('0.75.0')
+        RuboCop::Cop::MessageAnnotator.new(
+          config, cop.name, cop_config, {}
+        ).urls
+      else
+        RuboCop::Cop::MessageAnnotator.new(config, cop_config, {}).urls
+      end
```
pirj: WDYT of extracting this to a method, e.g.:
```ruby
    def references(config, cop)
      cop_config = config.for_cop(cop)
      urls = message_annotator.urls
      # ...
    end

    # Since RuboCop v0.75.0 ...
    def method_annotator
      rubocop_version = Gem::Version.new(RuboCop::Version::STRING)
      if rubocop_version >= Gem::Version.new('0.75.0')
        RuboCop::Cop::MessageAnnotator.new(config, cop.name, cop_config, {})
      else
        RuboCop::Cop::MessageAnnotator.new(config, cop_config, {})
      end
    end
```

I suspect this way both offences will be fixed.

spec/rubocop/cop/rspec/cop_spec.rb
```diff
@@ -21,13 +21,14 @@
 
   let(:fake_cop) do
     stub_const('RuboCop::RSpec', Module.new)
-    # rubocop:disable ClassAndModuleChildren, RSpec/LeakyConstantDeclaration
```
pirj: I think `Leaky` is still detected here.

tasks/cops_documentation.rake
nickcampbell18: Can do, but `method_annotator` needs to take `config` and `cop_config` as arguments so it gets a bit verbose. Your call!

spec/rubocop/cop/rspec/cop_spec.rb
nickcampbell18: I have changed this twice and my editor keeps autocorrecting it away (incorrectly) 😆

tasks/cops_documentation.rake
bquorning: Could we make a local version of `MessageAnnotator` instead? I’m thinking something like this:
```rb
# Since RuboCop v0.75.0 ...
rubocop_version = Gem::Version.new(RuboCop::Version::STRING)
if rubocop_version >= Gem::Version.new('0.75.0')
  MessageAnnotator = RuboCop::Cop::MessageAnnotator
else
  MessageAnnotator = Class.new(RuboCop::Cop::MessageAnnotator) do
    def initialize(config, cop_name, cop_config, options)
      super(config, cop_config, options)
    end
  end
end
```

nickcampbell18: Started to do that, I think it looked neater as originally suggested by just extracting to a method 😄

spec/rubocop/cop/rspec/cop_spec.rb
```diff
@@ -21,13 +21,15 @@
...
+    # rubocop:disable Style/ClassAndModuleChildren
```
pirj: Do you think it's necessary to put `disable` statements on separate lines?

nickcampbell18: Yes it was tripping up on line length 😬

PR #831, opened 2019-10-09
Implement RSpec/DescribedClassModuleWrapping cop

kellysutton:
This cop implements the behavior discussed in https://github.com/rubocop-hq/rubocop-rspec/issues/735.

**Here are some questions I would like to get answered before merging:**

* [x] This is set to default to not enabled. I think this is good for now. Do the maintainers agree?
* [x] This is only relevant to Rails code bases. Is this gem the right place to include this? Should this have a `Rails/` prefix?
* [x] The name of the cop is a mouthful. I'd be happy to change that.

---

Here's the situation that this can help discourage within Rails code bases:

I can tell a small story as to why this should be a cop to consider. Specifically, re-opening modules in tests can confuse the [Rails autoloading mechanism](https://guides.rubyonrails.org/autoloading_and_reloading_constants.html). (Note: This is all pre-zeitwork on Rails 5.2 and earlier. I'm not sure if Rails 6+ with zeitwerk will change this.)

If you write:

```ruby
module Foo
  module Bar
    RSpec.describe Baz do 
      # ...
    end
  end
end
```

And then you have code that refers to a constant defined on `Foo::Bar`, e.g. `Foo::Bar::WIDGETS`, this test code will fail if eagerloading is disabled. Why? Rails autoloading thinks it's already loaded the Foo::Bar module because it's defined.

Even though `app/models/foo/bar.rb` might define the `WIDGETS` constant, our test will not load the file because our test file has already defined the module. To confuse the matter more, turning on eager loading will cause the test to pass, because all code in `app/` will be loaded before the test executes.

Not reopening the module, e.g.

```ruby
RSpec.describe Foo::Bar::Baz do 
  # ...
end 
```
will cause the test to pass in both an eagerloaded and not eagerloaded configuration.

I realize that this is a Rails-specific issue, so I'm not sure if this gem is the best place to put such a cop or not. Let me know if I can further clarify the issue above or if my explanation was confusing. I may open a PR to this effect. Thanks!

---

* [x] Added tests.


lib/rubocop/cop/rspec/described_class_module_wrapping.rb
```diff
@@ -0,0 +1,38 @@
...
+      #   # good
+      #   RSpec.describe MyModule::MyClass do
+      #     # ...
+      #   end
+      #
+      # @see https://github.com/rubocop-hq/rubocop-rspec/issues/735
+      class DescribedClassModuleWrapping < Cop
+        MSG = 'Avoid opening modules and defining specs within them.'
+
+        def_node_matcher :rspec_block?,
```
pirj: Can be `def_node_search`, and in this case `node.each_descendant` is unnecessary.

lib/rubocop/cop/rspec/described_class_module_wrapping.rb
```diff
@@ -0,0 +1,38 @@
...
+                         RuboCop::RSpec::Language::ExampleGroups::ALL.block_pattern # rubocop:disable Metrics/LineLength
```
pirj: `RuboCop::Cop::RSpec::Cop` includes `RuboCop::RSpec::Language` so this might be just `ExampleGroups::ALL.block_pattern`.

lib/rubocop/cop/rspec/described_class_module_wrapping.rb
kellysutton: Noted. I'll give that a shot!

lib/rubocop/cop/rspec/described_class_module_wrapping.rb
kellysutton: Done in 44baef49aba7c814d30aa1c3800da032e4b527dc. Thanks for the quick review!

kellysutton:
@pirj Let me know if you need anything else to merge!

pirj:
I'm volunteering as an intern here :)
@bquorning @Darhazer WDYT?

spec/rubocop/cop/rspec/described_class_module_wrapping_spec.rb
```diff
@@ -0,0 +1,57 @@
+# frozen_string_literal: true
+
+RSpec.describe RuboCop::Cop::RSpec::DescribedClassModuleWrapping do
+  subject(:cop) { described_class.new }
+
+  context 'when an outermost describe block' do
+    it 'has no errors' do
```
bquorning: Personally, I don’t like using `context` blocks that do not set up a state (with `let` or `before`/`around`) and contain only one `it` block. I think it reads better to just have the `it` block in the parent scope. (This being rubocop-rspec, perhaps we should have a cop for checking this style 😄)

We tend to use wording like e.g. “allows”/“registers an offense” instead of “The “has no errors”/“has an error” in the spec examples. I went ahead and prepared four texts that you may want to use instead. I hope you don’t mind.

```
it 'allows a describe block in the outermost scope'
it 'registers an offense when RSpec.describe is nested within a module'
it 'registers an offense when RSpec.describe is nested within two modules'
it 'allows a module that does not contain RSpec.describe'
```

kellysutton: Sweet. I'll get these integrated. Thanks for the review!

kellysutton:
@bquorning Ready for another look at your earliest convenience. Thanks!

lib/rubocop/cop/rspec/described_class_module_wrapping.rb
```diff
@@ -0,0 +1,36 @@
...
+      #   RSpec.describe MyModule::MyClass do
+      #     # ...
+      #   end
+      #
+      # @see https://github.com/rubocop-hq/rubocop-rspec/issues/735
+      class DescribedClassModuleWrapping < Cop
+        MSG = 'Avoid opening modules and defining specs within them.'
+
+        def_node_search :find_rspec_blocks,
+                        ExampleGroups::ALL.block_pattern
```
pirj: I tend to think that this might detect an offence in the following code:
```ruby
module M
  describe 'something' do
    # ...
  end
end
```
Notice the example group methods with no explicit receiver.
This is quite minor considering the fact that RSpec cops are limited to a specific set of folders, and I've never seen any of those methods being used in the modules.

However, since it's just Ruby, it's pretty valid to extract some common code with example groups into a module. It's uncommon due to the existence of shared examples, but not something that I would never expect to see in the wild. E.g.:

```
module SomeCommonExamples
  describe 'some common things' do
    it 'does this' do
      # ...
    end
  end
end

RSpec.describe SomeClass do
  include SomeCommonExamples
end
```

bquorning: The cop is disabled by default, so I think it’s not going to be a problem. One could exclude the cop for all files in a folder containing such re-usable examples.

bquorning:
@kellysutton Could you squash the commits, please?

lib/rubocop/rspec/version.rb
```diff
@@ -4,7 +4,7 @@ module RuboCop
   module RSpec
     # Version information for the RSpec RuboCop plugin.
     module Version
-      STRING = '1.36.0'
+      STRING = '1.36.1'
```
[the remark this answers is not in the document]
kellysutton: Whoops. Sorry. Bad rebase. Will remove that commit from this branch.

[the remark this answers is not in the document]
bquorning: Agree. The version changes only when we do a new release :-)

PR #838, opened 2019-11-13
Add AllowedExplicitMatchers config option for RSpec/PredicateMatcher

mkrawc:
Add AllowedExplicitMatchers config option for RSpec/PredicateMatcher as suggested in https://github.com/rubocop-hq/rubocop-rspec/issues/806

---

* [x] Added tests.

### DIFF
--- a/spec/rubocop/cop/rspec/predicate_matcher_spec.rb
+++ b/spec/rubocop/cop/rspec/predicate_matcher_spec.rb
@@ -5,8 +5,10 @@ RSpec.describe RuboCop::Cop::RSpec::PredicateMatcher, :config do
 
   let(:cop_config) do
     { 'EnforcedStyle' => enforced_style,
-      'Strict' => strict }
+      'Strict' => strict,
+      'AllowedExplicitMatchers' => allowed_explicit_matchers }
   end
+  let(:allowed_explicit_matchers) { [] }
 
   context 'when enforced style is `inflected`' do
     let(:enforced_style) { 'inflected' }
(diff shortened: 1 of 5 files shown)


spec/rubocop/cop/rspec/predicate_matcher_spec.rb
```diff
@@ -270,6 +272,16 @@
         RUBY
       end
 
+      context 'when custom matchers are allowed' do
+        let(:allowed_explicit_matchers) { ['be_a_custom_matcher'] }
```
pirj: Maybe give a real example explaining the reasoning behind this, e.g. `have_http_status`?

mkrawc: @pirj `have_http_status` is one of build in matchers that potentially could be allowed with this feature. I can change `be_a_custom_matcher` to it if you prefer, that said the usage is much broader, like there are many rspec libraries that implement matchers, but also each project may have it's own custom matchers. Let me know your thoughts.

pirj: Sure, I understand your point that `have_http_status` may be allowed by default in the future.
However, chances are it will be allowed using the default value for the `AllowedExplicitMatchers` configuration option that you've introduced.
With this in mind, the spec will not only make sure that the option works, but at the same time will also indicate the reasoning behind this option, since it's a widely known matcher.

pirj:
👏 🎉

PR #839, opened 2019-11-14
Introduce FactoryGirl/FactoryClassName cop

jfragoulis:
The cop ensures that when creating factories and passing the class
name explicitly, we do not pass the class itself.

Although FactoryBot allows passing the class, we find that by doing so,
application models are loaded before the application itself gets loaded
(during factory definition).

This affects libraries like simplecov in that they fail to capture those files.

Example:

```ruby
# bad
factory :foo, class: Foo do
end

# good
factory :foo, class: 'Foo' do
end
```

### Checklist

* [x] Added tests.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 * Fix `RSpec/SubjectStub` to detect implicit subjects stubbed. ([@QQism][])
 * Fix `RSpec/Pending` not flagging `skip` with string values. ([@pirj][])
 * Add `AllowedExplicitMatchers` config option for `RSpec/PredicateMatcher`. ([@mkrawc][])
+* Add `FactoryBot/FactoryClassName` cop. ([@jfragoulis][])
 
 ## 1.36.0 (2019-09-27)
 
@@ -462,3 +463,4 @@ Compatibility release so users can upgrade RuboCop to 0.51.0. No new features.
 [@QQism]: https://github.com/QQism
 [@kellysutton]: https://github.com/kellysutton
 [@mkrawc]: https://github.com/mkrawc
+[@jfragoulis]: https://github.com/jfragoulis
--- a/lib/rubocop/cop/rspec/factory_bot/factory_class_name.rb
+++ b/lib/rubocop/cop/rspec/factory_bot/factory_class_name.rb
@@ -1,0 +1,39 @@
+# frozen_string_literal: true
+
+module RuboCop
+  module Cop
+    module RSpec
+      module FactoryBot
+        # Use string value when setting the class attribute explicitly.
+        #
+        # @example
+        #   # bad
+        #   factory :foo, class: Foo do
+        #   end
+        #
+        #   # good
+        #   factory :foo, class: 'Foo' do
+        #   end
+        class FactoryClassName < Cop
+          MSG = "Pass '%<class_name>s' instead of %<class_name>s."
+
+          def_node_matcher :class_name, <<~PATTERN
+            (send _ :factory _ (hash <(pair (sym :class) $(const ...)) ...>))
+          PATTERN
+
+          def on_send(node)
+            class_name(node) do |cn|
+              add_offense(cn, message: format(MSG, class_name: cn.const_name))
+            end
+          end
+
+          def autocorrect(node)
+            lambda do |corrector|
+              corrector.replace(node.loc.expression, "'#{node.source}'")
+            end
+          end
+        end
+      end
+    end
+  end
+end
--- a/spec/rubocop/cop/rspec/factory_bot/factory_class_name_spec.rb
+++ b/spec/rubocop/cop/rspec/factory_bot/factory_class_name_spec.rb
@@ -1,0 +1,72 @@
+# frozen_string_literal: true
+
+RSpec.describe RuboCop::Cop::RSpec::FactoryBot::FactoryClassName do
+  subject(:cop) { described_class.new }
+
+  context 'when passing block' do
+    it 'flags passing a class' do
+      expect_offense(<<~RUBY)
+        factory :foo, class: Foo do
+                             ^^^ Pass 'Foo' instead of Foo.
+        end
+      RUBY
+
+      expect_correction(<<~RUBY)
+        factory :foo, class: 'Foo' do
+        end
+      RUBY
+    end
+
+    it 'flags passing a class from global namespace' do
+      expect_offense(<<~RUBY)
+        factory :foo, class: ::Foo do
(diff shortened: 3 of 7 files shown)


lib/rubocop/cop/rspec/factory_bot/factory_class_name.rb
```diff
@@ -0,0 +1,55 @@
...
+        #   end
+        #
+        #   # good
+        #   factory :foo, class: 'Foo' do
+        #   end
+        class FactoryClassName < Cop
+          MSG = "Pass '%<class_name>s' instead of %<class_name>s."
+
+          def_node_matcher :factory_call?, <<~PATTERN
+            (block (send _ :factory ...) (...) _)
```
pirj: I suggest using [children matching](https://github.com/rubocop-hq/rubocop/issues/6965) to find `class` key, you can find some examples [here](https://github.com/rubocop-hq/rubocop-rspec/pull/804).

pirj: As far as I remember, a syntax with no block is also possible, e.g.:
```ruby
factory :author, class: User
```
This node pattern will miss it.

pirj:
I actually see an additional benefit in this style. Since factories usually cover all models, all models will be auto-loaded when `class` is defined as a constant, which is an additional overhead if a test only needs a few of the models, and it's possible to lazily load them.

jfragoulis:
> What is the reasoning for this cop?

The reason behind this as mentioned in the description is `application models are loaded before the application itself`. I realise this is supported but I do not see any benefit in loading all application models that have a factory before they are actually needed. 

The loading overhead argument you mentioned is part of my argument as well, not explicitly mentioned though.

spec/rubocop/cop/rspec/factory_bot/factory_class_name_spec.rb
```diff
@@ -0,0 +1,46 @@
+# frozen_string_literal: true
+
+RSpec.describe RuboCop::Cop::RSpec::FactoryBot::FactoryClassName do
+  subject(:cop) { described_class.new }
+
+  context 'when passing block' do
+    it 'registers an offense when using passing class' do
+      expect_offense(<<~RUBY)
+        factory :foo, class: Foo do
+                             ^^^ Pass 'Foo' instead of Foo.
```
pirj: Just in case, do you mind adding a check, or just make sure `::Foo` and `Foo::Bar` are also flagged?

lib/rubocop/cop/rspec/factory_bot/factory_class_name.rb
```diff
@@ -0,0 +1,41 @@
...
+        #   end
+        #
+        #   # good
+        #   factory :foo, class: 'Foo' do
+        #   end
+        class FactoryClassName < Cop
+          MSG = "Pass '%<class_name>s' instead of %<class_name>s."
+
+          def_node_matcher :class_name, <<~PATTERN
+            (send _ :factory _ (hash <(pair (sym :class) $_) ...>))
```
pirj: You may rewrite `$_` as `$(const)` and get rid of a guard statement.
It's up to you to decide if this node pattern is becoming overly complex though. I have no objections on keeping it this way.

jfragoulis:
Thanks for the comments. I believe this is ready.

spec/rubocop/cop/rspec/factory_bot/factory_class_name_spec.rb
```diff
@@ -0,0 +1,78 @@
+# frozen_string_literal: true
+
+RSpec.describe RuboCop::Cop::RSpec::FactoryBot::FactoryClassName do
+  subject(:cop) { described_class.new }
+
+  context 'when passing block' do
+    context 'when a class' do
+      it 'registers an offense' do
```
pirj: There's no guideline and even no common practice, but "ignores"/"flags" seems to be a little bit more widespread.
If you don't feel it fits the current spec structure that good, feel free to keep as it, it's good enough already.

lib/rubocop/cop/rspec/factory_bot/factory_class_name.rb
```diff
@@ -0,0 +1,41 @@
...
+          MSG = "Pass '%<class_name>s' instead of %<class_name>s."
+
+          def_node_matcher :class_name, <<~PATTERN
+            (send _ :factory _ (hash <(pair (sym :class) $(const ...)) ...>))
+          PATTERN
+
+          def on_send(node)
+            cn = class_name(node)
+
+            return unless cn
```
bquorning: The defined node matcher `class_name` will actually yield the found values to a block, so you could instead do this:
```suggestion
            class_name(node) do |cn|
              add_offense(cn, message: format(MSG, class_name: cn.const_name))
            end
```

jfragoulis: Ahh great! Thanks.

spec/rubocop/cop/rspec/factory_bot/factory_class_name_spec.rb
jfragoulis: I guess you are referring to the use of the working `registers` with the context above it.

Most of what I did I saw in other cops (factory bot ones). However, I can see there are 2 or 3 ways of doing things. It is normal for standards and common practices to change over time. It would be nice if the documentation was updated to reflect that, the best practices etc. 

But anyway, it was a nice experience altogether. It will help me shape new cops for lots of things.

PS: I used the flags/ignores paradigm, I hope it is ok.

bquorning:
@Darhazer, any comments?

CHANGELOG.md
```diff
@@ -403,6 +403,7 @@ Compatibility release so users can upgrade RuboCop to 0.51.0. No new features.
 * Add `RSpecMultipleDescribes` cop to check for multiple top level describes. ([@geniou][])
 * Add `RSpecDescribedClass` to promote the use of `described_class`. ([@geniou][])
 * Add `RSpecInstanceVariable` cop to check for the usage of instance variables. ([@geniou][])
+* Add `FactoryBot/FactoryClassName` cop. ([@jfragoulis][])
```
bquorning: Ah, just one more thing. Could you move this line up towards the top of the file, as the last bullet point under `## Master (Unreleased)` please.

bquorning: …you‘ll need to rebase on latest master too, to avoid merge conflict.

jfragoulis: Wow, what an oversight! :) :) Done and thanks!

pirj:
Thanks and looking forward to your future contributions! 👍

PR #845, opened 2019-11-27
Improve message and description of `FactoryBot/FactoryClassName`

ybiquitous:
When I saw the cop's message first, I did not know how to fix my code.
This modification aims to increase the message's readability by explicitly describing "string" and "constant".

Next, I could not understand also why the cop warned when I saw its document.
So, this also adds the cop's rationale to its document. This rationale is based on the PR #839 description.

---

* [x] Added tests.

### DIFF
--- a/lib/rubocop/cop/rspec/factory_bot/factory_class_name.rb
+++ b/lib/rubocop/cop/rspec/factory_bot/factory_class_name.rb
@@ -6,6 +6,11 @@ module RuboCop
       module FactoryBot
         # Use string value when setting the class attribute explicitly.
         #
+        # This cop would promote faster tests by lazy-loading of
+        # application files. Also, this could help you suppress potential bugs
+        # in combination with external libraries by avoiding a preload of
+        # application files from the factory files.
+        #
         # @example
         #   # bad
         #   factory :foo, class: Foo do
@@ -15,7 +20,8 @@ module RuboCop
         #   factory :foo, class: 'Foo' do
         #   end
         class FactoryClassName < Cop
-          MSG = "Pass '%<class_name>s' instead of %<class_name>s."
+          MSG = "Pass '%<class_name>s' string instead of `%<class_name>s` " \
+                'constant.'
 
           def_node_matcher :class_name, <<~PATTERN
             (send _ :factory _ (hash <(pair (sym :class) $(const ...)) ...>))
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Master (Unreleased)
 
+* Improve message and description of `FactoryBot/FactoryClassName`. ([@ybiquitous][])
+
 ## 1.37.0 (2019-11-25)
 
 * Implement `RSpec/DescribedClassModuleWrapping` to disallow RSpec statements within a module. ([@kellysutton][])
@@ -466,3 +468,4 @@ Compatibility release so users can upgrade RuboCop to 0.51.0. No new features.
 [@kellysutton]: https://github.com/kellysutton
 [@mkrawc]: https://github.com/mkrawc
 [@jfragoulis]: https://github.com/jfragoulis
+[@ybiquitous]: https://github.com/ybiquitous
--- a/manual/cops_factorybot.md
+++ b/manual/cops_factorybot.md
@@ -86,6 +86,11 @@ Enabled | Yes
 
 Use string value when setting the class attribute explicitly.
 
+This cop would promote faster tests by lazy-loading of
+application files. Also, this could help you suppress potential bugs
+in combination with external libraries by avoiding a preload of
+application files from the factory files.
+
 ### Examples
 
 ```ruby
--- a/spec/rubocop/cop/rspec/factory_bot/factory_class_name_spec.rb
+++ b/spec/rubocop/cop/rspec/factory_bot/factory_class_name_spec.rb
@@ -7,7 +7,7 @@ RSpec.describe RuboCop::Cop::RSpec::FactoryBot::FactoryClassName do
     it 'flags passing a class' do
       expect_offense(<<~RUBY)
         factory :foo, class: Foo do
-                             ^^^ Pass 'Foo' instead of Foo.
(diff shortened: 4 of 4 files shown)


lib/rubocop/cop/rspec/factory_bot/factory_class_name.rb
```diff
@@ -6,6 +6,11 @@ module RSpec
       module FactoryBot
         # Use string value when setting the class attribute explicitly.
         #
+        # Although FactoryBot allows passing a class, passed application models
+        # are loaded before the application itself gets loaded.
+        # This affects libraries like "simplecov" in that they fail to capture
```
pirj: Is there a difference if you require `simplecov` before or after `factory_bot`?

lib/rubocop/cop/rspec/factory_bot/factory_class_name.rb
```diff
@@ -6,6 +6,11 @@ module RSpec
       module FactoryBot
         # Use string value when setting the class attribute explicitly.
         #
+        # Although FactoryBot allows passing a class, passed application models
+        # are loaded before the application itself gets loaded.
```
pirj: The original motivation was to avoid loading those classes if they are not needed, and let autoload to do its job. I'm not 100% convinced this works as expected though.

lib/rubocop/cop/rspec/factory_bot/factory_class_name.rb
```diff
@@ -15,7 +20,9 @@ module FactoryBot
         #   factory :foo, class: 'Foo' do
         #   end
         class FactoryClassName < Cop
-          MSG = "Pass '%<class_name>s' instead of %<class_name>s."
+          # rubocop:disable Metrics/LineLength
+          MSG = "Pass '%<class_name>s' string instead of `%<class_name>s` constant."
```
pirj: Please wrap.

lib/rubocop/cop/rspec/factory_bot/factory_class_name.rb
ybiquitous: Honestly, I don't know because I never had this situation.

This description is originated from #839, but if it may be unclear, I can accept any better suggestions.

lib/rubocop/cop/rspec/factory_bot/factory_class_name.rb
ybiquitous: Do you mean that adding a description like "When not to use it" or "Justification" is good?

lib/rubocop/cop/rspec/factory_bot/factory_class_name.rb
pirj: I'd go with something like this https://github.com/rubocop-hq/rubocop-rspec/pull/839#issuecomment-554687768, and would keep simplecov note, however, those two hypotheses need to be confirmed first.

lib/rubocop/cop/rspec/factory_bot/factory_class_name.rb
ybiquitous: Sorry, what should I wrap?

pirj: The message. Please take a loot at [an example](https://github.com/rubocop-hq/rubocop-rspec/blob/master/lib/rubocop/cop/rspec/describe_method.rb#L23).

ybiquitous: I got it, thank you.

ybiquitous: Fixed via a8167f3b3262943d890a016afbac959a4bbdce41

lib/rubocop/cop/rspec/factory_bot/factory_class_name.rb
pirj: There's no common practice, some cops only explain what they do, [some](https://github.com/rubocop-hq/rubocop-rspec/blob/master/manual/cops_rspec.md#rspecexamplelength) why they do it. However, if there's a motivation, it needs to be solid.

lib/rubocop/cop/rspec/factory_bot/factory_class_name.rb
ybiquitous: You're right. But I don't have resources to confirm these hypotheses and the overhead can be ignored in small applications.

After all, does this cop warn a stylistic issue?

lib/rubocop/cop/rspec/factory_bot/factory_class_name.rb
jfragoulis: First of all, I am OK with being more specific regarding both the what and the why. Secondly, this is low profile cop that may not affect you ever.

The main why is speed, lazy-loading of course. 
However, my motivation behind this cop comes from personal experience on (let's say) bad code, legacy or not, in combination with external libraries, gems etc. There are cases where gems/engines evaluate application files before the applications get loaded. This is something potentially wrong, bad etc.
One could make use of this cop if needed to avoid pre-loading application files **at least** from factories.

ybiquitous: @jfragoulis Thanks for your detail feedback! Then, I try to change the description as follows. I'm not native in English, so its wording may feel strange. I would be grateful if you could point out some strange parts.

Here is my draft:

> This cop would promote faster tests by lazy-loading of application files. Also, this could help you suppress potential bugs in combination with external libraries by avoiding a preload of application files from the factory files.

jfragoulis: This is perfect.

ybiquitous: Thank you. I've pushed the change via d4e1e5e56d7ad40538d43a29a5576a2963e8ab11.

Could you please re-review?

ybiquitous:
> You can squash the commits

Done 🙆‍♂

ybiquitous:
Thanks all! 😊

PR #846, opened 2019-12-12
Fix `FactoryBot/FactoryClassName` to ignore `Hash` and `OpenStruct`

jfragoulis:
Hash and OpenStruct are used in factories widely.
Then factory should not force them into strings as they are built in language objects.

---

* [x] Added tests.

### DIFF
--- a/spec/rubocop/cop/rspec/factory_bot/factory_class_name_spec.rb
+++ b/spec/rubocop/cop/rspec/factory_bot/factory_class_name_spec.rb
@@ -49,6 +49,20 @@ RSpec.describe RuboCop::Cop::RSpec::FactoryBot::FactoryClassName do
         end
       RUBY
     end
+
+    it 'ignores passing Hash' do
+      expect_no_offenses(<<~RUBY)
+        factory :foo, class: Hash do
+        end
+      RUBY
+    end
+
+    it 'ignores passing OpenStruct' do
+      expect_no_offenses(<<~RUBY)
+        factory :foo, class: OpenStruct do
+        end
+      RUBY
+    end
   end
 
   context 'when not passing block' do
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## Master (Unreleased)
 
 * Improve message and description of `FactoryBot/FactoryClassName`. ([@ybiquitous][])
+* Fix `FactoryBot/FactoryClassName` to ignore `Hash` and `OpenStruct`. ([@jfragoulis][])
 
 ## 1.37.0 (2019-11-25)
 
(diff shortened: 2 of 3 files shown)


spec/rubocop/cop/rspec/factory_bot/factory_class_name_spec.rb
```diff
@@ -49,6 +49,20 @@
...
+        factory :foo, class: Hash do
+        end
+      RUBY
+    end
+
+    it 'ignores passing OpenStruct' do
+      expect_no_offenses(<<~RUBY)
+        factory :foo, class: OpenStruct do
+        end
+      RUBY
```
pirj: I haven't seen this in my practice. Can you share some code that is using this?

jfragoulis: Sure. 

Here I am mocking an object created from an (internal) library which itself was created from an api's json response:

```ruby
FactoryBot.define do
  factory :everypay_seller, class: OpenStruct do
    skip_create

    sequence(:id) { |n| "sel_#{n}" }
    association :company, factory: :food_company

    initialize_with do
      atrs = Food::Acquirer::CompanyToSeller.new(company: company).from_api
      new(atrs.merge(id: id))
    end
  end
end
```

It is a very convenient way of creating *custom* models.

Another example of a response object.

```ruby
FactoryBot.define do
  factory :downloader_callback_params, aliases: [:successful_callback], class: OpenStruct do
    success { true }
    error { '' }
    download_url { Faker::Internet.url }
    resource_url { Faker::Internet.url }
    job_id { SecureRandom.urlsafe_base64(14) }
    response_code { 200 }
    extra { { 'groupno' => 'trf_32387729' }.to_json }

    factory :unsuccessful_callback do
      success { false }

      trait :with_404 do
        error { 'Received Status code 404' }
        response_code { 404 }
      end

      trait :with_TLS do
        error { 'TLS Error Occured: dial: x509: certificate' }
        response_code { 0 }
      end

      trait :with_mime do
        error { 'Expected mime-type to be (image/jpeg), found (image/png)' }
        response_code { 200 }
      end
    end
  end
end
```

I am also using it to create api responses from github and rubocop json results.

Darhazer:
Thank you for your contribution.
Please rebase on current master so I can merge

jfragoulis:
> Thank you for your contribution.
> Please rebase on current master so I can merge

Done.

spec/rubocop/cop/rspec/factory_bot/factory_class_name_spec.rb
pirj: Got it, value object factory. Looks legit 👍

PR #847, opened 2019-12-12
Fix master

jfragoulis:
`bundle exec rake` is producing some errors. These commits address these errors.


.rubocop.yml
```diff
@@ -12,8 +12,7 @@ AllCops:
     - 'tmp/**/*'
     - 'spec/smoke_tests/**/*.rb'
 
-# See https://github.com/rubocop-hq/rubocop/issues/6410
```
pirj: Do you mind to keep this reference? I believe it's here to explain why this is disabled.

lib/rubocop/cop/rspec/described_class.rb
```diff
@@ -139,7 +139,7 @@ def offensive?(node)
           if style == :described_class
             offensive_described_class?(node)
           else
-            node.send_type? && node.method_name == :described_class
+            node.send_type? && node.method?(:described_class)
```
pirj: Will this work with the earliest RuboCop version that we support, 0.68.1?

pirj: With 0.68.1 installed `./spec/rubocop/cop/rspec/implicit_subject_spec.rb` fails several examples with:
```
      NoMethodError:
        undefined method `method?' for #<RuboCop::AST::BlockNode:0x00007ff05a4aaef0>
        Did you mean?  method
                       methods
```

Do you mind to keep those intact and disabling the cop that suggested this to `.rubocop_todo.yml`?

jfragoulis: Hmm, what more can if `bundle exec rubocop` passes?

bquorning:
Ah, issues between Rubocop v0.76.0 and v0.77.0. Thank you for fixing.

PR #848, opened 2019-12-12
Fix RSpec/MessageChain "better" docs.

schwern:
The original code results in an error because `receive` takes a single argument of the method name which it tries to symbolize.

```
     NoMethodError:
       undefined method `to_sym' for {:bar=>#<Thing:0x00007ff34792c178 @baz=42>}:Hash
       Did you mean?  to_s
                      to_set
```

I believe what it's trying to do is `allow(foo).to receive(:bar).and_return(thing)`.

---

* [x] N/A Added tests.
* [x] N/A Added an entry to the [changelog](https://github.com/rubocop-hq/rubocop-rspec/blob/master/CHANGELOG.md) if the new code introduces user-observable changes.

### DIFF
--- a/lib/rubocop/cop/rspec/message_chain.rb
+++ b/lib/rubocop/cop/rspec/message_chain.rb
@@ -11,7 +11,7 @@ module RuboCop
       #
       #   # better
       #   thing = Thing.new(baz: 42)
-      #   allow(foo).to receive(bar: thing)
+      #   allow(foo).to receive(:bar).and_return(thing)
       #
       class MessageChain < Cop
         MSG = 'Avoid stubbing using `%<method>s`.'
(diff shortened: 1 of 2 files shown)


schwern:
Here's some code showing the equivalency between the bad and better.

```Ruby
require 'rspec'

class Thing
  attr_accessor :baz
  
  def initialize(baz:)
    @baz = baz
  end
end

class Foo
  def bar
    23
  end
end

RSpec.describe 'receive' do
  let(:foo) { Foo.new }
  
  shared_examples 'baz is 42' do
    it 'foo.bar.baz is 42' do
      expect(foo.bar.baz).to eq 42
    end
  end
  
  context 'with receive_message_chain' do
    before do
      allow(foo).to receive_message_chain(:bar, :baz).and_return(42)
    end

    it_behaves_like 'baz is 42'
  end
  
  context 'with receive.and_return' do
    let(:thing) {
      Thing.new(baz: 42)
    }
    before do
      allow(foo).to receive(:bar).and_return(thing)
    end
    
    it_behaves_like 'baz is 42'
  end
end
```

PR #851, opened 2019-12-20
Add metadata detection to RSpec/ScatteredSetup cop

### LINKED ISSUE #850: RSpec/ScatteredSetup falsely triggers on blocks with different metadata
Version: 1.37.0

Code:
```rb
before do
  # something
end

before(:each, :special_cases) do
  # something else
end
```

What happens:
RSpec/ScatteredSetup cop is triggered.

What I expected to happen:
RSpec/ScatteredSetup cop to not trigger, since both blocks are distinct and can not be combined.

pirj:
Added metadata detection to hooks. Taught the cop to distinguish hooks also by that metadata.

Transformation of metadata is necessary though, since the following is the same in RSpec's terms:

```ruby
before(special: true) { }
before(:example, :special) { }
before(:example, special: true) { }
```

NOTE: RSpec allows hash metadata to be present when no explicit hook scope is given.
However, if hook scope is omitted, symbolic metadata is not allowed, and that causes RSpec to raise an error.
Since we normally don't cover the cases when RSpec emits a meaningful warning, some of the validations were removed.

Fixes #850

---

* [x] Added tests.

### DIFF
--- a/lib/rubocop/rspec/hook.rb
+++ b/lib/rubocop/rspec/hook.rb
@@ -4,21 +4,24 @@ module RuboCop
   module RSpec
     # Wrapper for RSpec hook
     class Hook < Concept
-      STANDARDIZED_SCOPES = %i[each context suite].freeze
-      private_constant(:STANDARDIZED_SCOPES)
+      def_node_matcher :extract_metadata, <<~PATTERN
+        (block
+          {
+            (send _ _ #valid_scope? $...)
+            (send _ _ $...)
+          }
+          ...
+        )
+      PATTERN
 
       def name
         node.method_name
       end
 
       def knowable_scope?
-        return true unless scope_argument
-
-        scope_argument.sym_type?
-      end
-
-      def valid_scope?
-        STANDARDIZED_SCOPES.include?(scope)
+        scope_argument.nil? ||
+          scope_argument.sym_type? ||
+          scope_argument.hash_type?
       end
 
       def example?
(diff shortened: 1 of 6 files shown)


pirj:
@bquorning @Darhazer please take a look.

lib/rubocop/rspec/hook.rb
```diff
@@ -4,39 +4,65 @@ module RuboCop
...
 
+      def valid_scope?(node)
+        node&.sym_type? && Hooks::Scopes::ALL.include?(node.value)
+      end
+
+      def transform_metadata(meta)
+        if meta.sym_type?
+          { meta => true }
+        else
+          meta.pairs.map { |pair| { pair.key => transform_true(pair.value) } }
```
pirj: This check is to be able to compare those two hooks:
```
before(:example, :special) { ... }
before(:example, special: true) { ... }
```
In the second case it's a node of a `true_type?`.

bquorning: I think it would be helpful to add that comment as an actual code comment :-)

spec/rubocop/rspec/hook_spec.rb
```diff
@@ -11,43 +11,89 @@ def hook(source)
...
+  describe '#metadata' do
+    def metadata(source)
+      hook(source).metadata.to_a
+    end
 
-  include_examples 'standardizes scope', 'before(:all) { }',     :context
-  include_examples 'standardizes scope', 'around(:context) { }', :context
+    it 'extracts symbol metadata' do
+      expect(metadata('before(:example, :special) { foo }'))
+        .to contain_exactly([have_attributes(value: :special), true])
```
bquorning: Reading the `have_attributes(value: :special)` matcher, I’m not sure what’s actually in the metadata array.

pirj: `metadata('before(:example, :special) { foo }')` (`metadata('before(:example, :special) { foo }').metadata.to_a`)is `[[s(:sym, :special), true]]`.

Do you think something like that in mind that would make it more indicative?
```ruby
    def metadata(source)
      hook(source).metadata
    end

    def a_sex_symbol(name)
      be_sym_type
        .and have_attributes(value: name)
    end

it do
  expect(metadata(...))
    .to eq(a_sex_symbol(:special) => true)
end
```

Unfortunately, I couldn't make it work with `eq`/`match`/`include`, they don't seem to match hash keys against argument matchers.

bquorning: Would it be too obscure to just call `#to_s` on the metadata?
```rb
def metadata(source)
  hook(source).metadata.to_s
end

it 'extracts symbol metadata' do
  expect(metadata('before(:example, :special) { foo }'))
    .to eq('{s(:sym, :special)=>true}')
end
```

pirj: Not too bad. We can think later how to improve it if it's not clear enough.

bquorning:
Thank you @pirj 👍

pirj:
You're always welcome!

bquorning:
I don’t know why I haven’t done this before – I’ve added you to the rubocop-rspec team so you don’t have to push to your own fork anymore. That also makes it for me to pull down your branches for review locally :-D

pirj:
Thanks, it's a honour for me!

Darhazer:
offtopic: @bquorning you can use [hub](https://github.com/github/hub) to easily checkout PRs

PR #853, opened 2019-12-29
Fix RSpec/InstanceVariable inside custom matcher

### LINKED ISSUE #852: RSpec/InstanceVariable should ignore variables inside matchers.
RSpec/InstanceVariable is flagging instance variables inside matchers embedded in spec files. AFAIK instance variables are necessary for anything but a trivial matcher and replacing them with the suggested local variable, let, or method call would be inappropriate.

An example from the RSpec docs: [chained method with argument](https://relishapp.com/rspec/rspec-expectations/v/3-9/docs/custom-matchers/define-matcher-with-fluent-interface#chained-method-with-argument)

```
RSpec::Matchers.define :be_bigger_than do |first|
  match do |actual|
    (actual > first) && (actual < @second)
  end

  chain :and_smaller_than do |second|
    @second = second
  end
end
```

An example of writing a matcher which is a wrapper around another matcher and passing through the failure message.

```
  matcher :look_like_a_thing do
    match do |thing|
      @matcher = have_attributes(...)

      @matcher.matches?(thing)
    end

    failure_message do
      @matcher.failure_message
    end
  end
```

I propose `IgnoreMatchers` defaulting to `true`.

pirj:
fixes #852

---

* [x] Added tests.

### DIFF
--- a/lib/rubocop/cop/rspec/instance_variable.rb
+++ b/lib/rubocop/cop/rspec/instance_variable.rb
@@ -58,6 +58,13 @@ module RuboCop
           (block (send (const nil? :Class) :new ...) ...)
         PATTERN
 
+        def_node_matcher :custom_matcher?, <<-PATTERN
+          (block {
+            (send nil? :matcher sym)
+            (send (const (const nil? :RSpec) :Matchers) :define sym)
+          } ...)
+        PATTERN
+
         def_node_search :ivar_usage, '$(ivar $_)'
 
         def_node_search :ivar_assigned?, '(ivasgn % ...)'
@@ -66,8 +73,8 @@ module RuboCop
           return unless spec_group?(node)
 
           ivar_usage(node) do |ivar, name|
-            return if inside_dynamic_class?(ivar)
-            return if assignment_only? && !ivar_assigned?(node, name)
+            next if valid_usage?(ivar)
+            next if assignment_only? && !ivar_assigned?(node, name)
 
             add_offense(ivar)
           end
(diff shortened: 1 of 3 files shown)


pirj:
Config build failure is due to https://github.com/lsegal/yard/pull/1296

lib/rubocop/cop/rspec/instance_variable.rb
```diff
@@ -67,6 +74,7 @@ def on_block(node)
 
           ivar_usage(node) do |ivar, name|
             return if inside_dynamic_class?(ivar)
+            return if inside_custom_matcher?(ivar)
```
Darhazer: I'd just rename the `inside_dunamic_class?` method and modify the block inside to check for both cases. It would be easier to extend (no more conditions added here) and will perform only one lookup up in the tree.

example name:
`return if valid_usage?(ivar)`

spec/rubocop/cop/rspec/instance_variable_spec.rb
```diff
@@ -67,6 +67,41 @@ def serialize
     RUBY
   end
 
+  context 'when used in a custom matcher' do
+    it 'ignores instance variables inside `mather`' do
+      expect_no_offenses(<<~RUBY)
```
Darhazer: for full coverage, maybe add an instance usage (@matcher) outside of the matcher definition?

pirj: Sometimes I see you as a clairvoyant, Maxim, that spec immediately helped to find a bug. 👍

pirj:
Green 👍 
Not sure if it makes sense to mention the second bug fixed in the changelist.

spec/rubocop/cop/rspec/instance_variable_spec.rb
```diff
@@ -31,6 +31,18 @@
     RUBY
   end
 
+  it 'flags sefveral instance variables inside a describe' do
```
bquorning:
```suggestion
  it 'flags several instance variables inside a describe' do
```

pirj:
@bquorning Thanks for the fix.
Squashed, green.

PR #855, opened 2020-01-06
Add autocorrect for RSpec/ExpectActual

dduugg:
Adds an autocorrect to `RSpec/ExpectActual`, swapping the arguments to `expect` and `eq` when possible.

---

* [x] Added tests.

### DIFF
--- a/lib/rubocop/cop/rspec/expect_actual.rb
+++ b/lib/rubocop/cop/rspec/expect_actual.rb
@@ -49,6 +49,17 @@ module RuboCop
           end
         end
 
+        def autocorrect(node)
+          lambda do |corrector|
+            expectation = node.parent.parent
+            rhs = expectation.children.last
+            return unless rhs.is_a?(RuboCop::AST::MethodDispatchNode)
+            return if rhs.method_name != :eq
+
+            swap_order(corrector, node, rhs.children.last)
+          end
+        end
+
         private
 
         # This is not implement using a NodePattern because it seems
@@ -64,6 +75,11 @@ module RuboCop
         def complex_literal?(node)
           COMPLEX_LITERALS.include?(node.type) &&
             node.each_child_node.all?(&method(:literal?))
+        end
+
+        def swap_order(corrector, lhs_arg, rhs_arg)
+          corrector.replace(lhs_arg.source_range, rhs_arg.source)
+          corrector.replace(rhs_arg.source_range, lhs_arg.source)
         end
       end
     end
--- a/manual/cops_rspec.md
+++ b/manual/cops_rspec.md
@@ -977,7 +977,7 @@ IgnoredWords | `[]` | Array
 
 Enabled by default | Supports autocorrection
 --- | ---
-Enabled | No
+Enabled | Yes
 
 Checks for `expect(...)` calls containing literal values.
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 * Fix `RSpec/InstanceVariable` detection inside custom matchers. ([@pirj][])
 * Fix `RSpec/ScatteredSetup` to distinguish hooks with different metadata. ([@pirj][])
+* Add autocorrect support for `RSpec/ExpectActual` cop. ([@dduugg][])
 
 ## 1.37.1 (2019-12-16)
 
@@ -475,3 +476,4 @@ Compatibility release so users can upgrade RuboCop to 0.51.0. No new features.
 [@mkrawc]: https://github.com/mkrawc
 [@jfragoulis]: https://github.com/jfragoulis
 [@ybiquitous]: https://github.com/ybiquitous
+[@dduugg]: https://github.com/dduugg
--- a/spec/rubocop/cop/rspec/expect_actual_spec.rb
+++ b/spec/rubocop/cop/rspec/expect_actual_spec.rb
@@ -18,6 +18,17 @@ RSpec.describe RuboCop::Cop::RSpec::ExpectActual, :config do
         end
       end
     RUBY
+
+    expect_correction(<<-RUBY)
+      describe Foo do
+        it 'uses expect incorrectly' do
+          expect(bar).to eq(123)
+          expect(bar).to eq(12.3)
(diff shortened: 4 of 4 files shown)


manual/cops_rspec.md
```diff
@@ -977,7 +977,7 @@ IgnoredWords | `[]` | Array
 
 Enabled by default | Supports autocorrection
 --- | ---
-Enabled | No
+Enabled | Yes
```
pirj: I'm not sure how exactly this is generated, shouldn't we add `AutoCorrect: true` to `config/default.yml`?

lib/rubocop/cop/rspec/expect_actual.rb
```diff
@@ -49,6 +49,17 @@ def on_send(node)
           end
         end
 
+        def autocorrect(node)
+          lambda do |corrector|
+            expectation = node.parent.parent
+            eq = expectation.children.last.children.last
+            return unless eq.is_a?(RuboCop::AST::Node)
```
pirj: What is this case when it doesn't auto-correct? Do you mind to add an example or extend an existing one with `expect_no_correction` to cover it?

manual/cops_rspec.md
Darhazer: Nope.
AutoCorrect: true / AutoCorrect: false in config/default.yml just tells whether this autocorrect is enabled or disabled by default. It is a leftover from the times before we had a `SafeAutocorrect` option. 

This just tells whether the cop implements `autocorrect` at all.

lib/rubocop/cop/rspec/expect_actual.rb
dduugg: Hi @pirj, great question. This guard clause is for an entry in `spec/smoke_tests/weird_rspec_spec.rb`:
```
expect([]).to be_empty
```
In this case, there wasn't an obvious way to autocorrect it (although I'm new to rspec), so it would still need to be manually refactored.

Darhazer: It seems that what you are doing here is basically checking whether the mather has arguments.

What will happen with `expect([1,2,3]).to include(a)`?

Perhaps this should not be flagged in the first place, but the correction also doesn't make sense. I guess you should check the actual matcher, and whether it has arguments or not.

pirj: Maybe add a similar non-auto-correctable example to ExpectActual's spec?

dduugg: Great catch, i've scoped the autocorrect to only apply when the right-hand side is an `eq`. I've added a test using @Darhazer's example.

Please take another look. Thanks!

lib/rubocop/cop/rspec/expect_actual.rb
```diff
@@ -49,6 +49,17 @@ def on_send(node)
...
+        def autocorrect(node)
+          lambda do |corrector|
+            expectation = node.parent.parent
+            rhs = expectation.children.last
+            return unless rhs.is_a?(RuboCop::AST::MethodDispatchNode)
+            return if rhs.method_name != :eq
+
+            swap_order(corrector, node, rhs.children.last)
+          end
+        end
```
pirj: Played around with it, and came up with this:
```ruby
        def_node_matcher :expect_literal, <<~PATTERN
          (send
            (send nil? :expect $#literal?)
            #{Runners::ALL.node_pattern_union}
            $(send nil? ...)
          )
        PATTERN

        def on_send(node)
          expect_literal(node) do |argument|
            add_offense(node, location: argument.source_range)
          end
        end

        def autocorrect(node)
          argument, matcher = expect_literal(node)
          lambda do |corrector|
            return if matcher.method_name != :eq

            swap_order(corrector, argument, matcher.arguments.first)
          end
        end
```

Node pattern is hard to grasp at first but helps to get rid of type checks and node deconstruction/walking.

Up to you to keep as as or cherry-pick parts of that code.

lib/rubocop/cop/rspec/expect_actual.rb
```diff
@@ -65,6 +76,11 @@ def complex_literal?(node)
           COMPLEX_LITERALS.include?(node.type) &&
             node.each_child_node.all?(&method(:literal?))
         end
+
+        def swap_order(corrector, lhs_arg, rhs_arg)
```
pirj: In RSpec terminology `lhs_arg` is "expectation target", and `rhs_arg` is "matcher argument", but I agree that at this level we operate nodes.

lib/rubocop/cop/rspec/expect_actual.rb
```diff
@@ -49,6 +49,17 @@ def on_send(node)
           end
         end
 
+        def autocorrect(node)
+          lambda do |corrector|
+            expectation = node.parent.parent
+            rhs = expectation.children.last
+            return unless rhs.is_a?(RuboCop::AST::MethodDispatchNode)
+            return if rhs.method_name != :eq
```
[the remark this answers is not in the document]
dduugg: Hi @pirj, thanks for the suggestions. Could I trouble you to provide examples using the first three methods? I interpreted your suggestion as applying like so:
```ruby
expect(123).to ==(bar)
expect(123).to eql?(bar)
expect(123).to equal?(bar)
```
…but that doesn't appear to be valid:
```
     Failure/Error: expect(123).to equal?(123)

     ArgumentError:
       The expect syntax does not support operator matchers, so you must pass a matcher to `#to`.
```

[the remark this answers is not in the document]
pirj: My bad. [Should of course be](https://relishapp.com/rspec/rspec-expectations/v/3-9/docs/built-in-matchers/comparison-matchers):
```ruby
expect(123).to be
expect(123).to be == 123
expect(123).to eql(123)
expect(123).to equal(123)
```

pirj:
Do you need a hand with this @dduugg ? Please let me know if you get stuck with something.

dduugg:
Hi @pirj I'll close this until I have more bandwidth to cover the various examples. It seemed much simpler when I was able to autocorrect the cop's spec and my code with ease 😸

PR #858, opened 2020-01-11
Adds an autocorrect to `RSpec/ExpectActual`, swapping the arguments to `expect` and matcher when possible.

pirj:
Based on #855 by @dduugg

---

* [x] Added tests.

### DIFF
--- a/lib/rubocop/cop/rspec/expect_actual.rb
+++ b/lib/rubocop/cop/rspec/expect_actual.rb
@@ -60,7 +60,7 @@ module RuboCop
           lambda do |corrector|
             return if matcher.method_name != :eq
 
-            swap_order(corrector, argument, matcher.arguments.first)
+            swap(corrector, argument, matcher.arguments.first)
           end
         end
 
@@ -81,9 +81,9 @@ module RuboCop
             node.each_child_node.all?(&method(:literal?))
         end
 
-        def swap_order(corrector, lhs_arg, rhs_arg)
-          corrector.replace(lhs_arg.source_range, rhs_arg.source)
-          corrector.replace(rhs_arg.source_range, lhs_arg.source)
+        def swap(corrector, actual, expected)
+          corrector.replace(actual.source_range, expected.source)
+          corrector.replace(expected.source_range, actual.source)
         end
       end
     end
--- a/lib/rubocop/cop/rspec/expect_actual.rb
+++ b/lib/rubocop/cop/rspec/expect_actual.rb
@@ -49,6 +49,17 @@ module RuboCop
           end
         end
 
+        def autocorrect(node)
+          lambda do |corrector|
+            expectation = node.parent.parent
+            rhs = expectation.children.last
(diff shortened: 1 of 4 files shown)


lib/rubocop/cop/rspec/expect_actual.rb
```diff
@@ -41,11 +41,33 @@ class ExpectActual < Cop
...
-            add_offense(argument)
+            add_offense(node, location: argument.source_range)
+          end
+        end
+
+        def autocorrect(node)
+          actual, matcher, expected = expect_literal(node)
+          lambda do |corrector|
+            return unless SUPPORTED_MATCHERS.include?(matcher)
+            return if expected.is_a?(Array) # `be` without args
```
Darhazer: The comment is not precise, as arguments will also result in having an array. The `:==` case does not result in an array because of the `$_` (while `$...` results in an array). So you are not correcting `be`, unless followed by `==`

pirj: Nice catch, the `be` with an argument case completely slipped my mind 👍 
I assumed I'm using `$...` as a placeholder just to balance the number of captures in patter union clause, and it will always be `[]` for `be`.

spec/rubocop/cop/rspec/expect_actual_spec.rb
```diff
@@ -135,6 +207,108 @@
     RUBY
   end
 
+  it 'flags `be` with no argument' do
+    expect_offense(<<~RUBY)
+      describe Foo do
+        it 'uses expect incorrectly' do
+          expect(1).to be
+                 ^ Provide the actual you are testing to `expect(...)`.
```
Darhazer: Well, we don't have an actual here, and supporting this case complicates the pattern and the autocorrect code

pirj: Makes sense.

pirj: Fixed.

lib/rubocop/cop/rspec/expect_actual.rb
```diff
@@ -41,11 +41,32 @@ class ExpectActual < Cop
...
-        def_node_matcher :expect_literal, '(send _ :expect $#literal?)'
+        SUPPORTED_MATCHERS = %i[eq eql equal be].freeze
+
+        def_node_matcher :expect_literal, <<~PATTERN
+          (send
+            (send nil? :expect $#literal?)
+            #{Runners::ALL.node_pattern_union}
+            {
+              (send (send nil? $:be) :== $_)
+              (send nil? $:be $_)
```
Darhazer: I think the matcher below handles this case as well, as be is in the supported matchers

pirj: Fair enough, removed.

pirj:
Thanks, guys!

PR #859, opened 2020-01-12
Do not require gem in Gemspec

### LINKED ISSUE #857: Is `rubocop-rspec` require `require: false` in gemfile entry
RuboCop Readme has entry, that RuboCop should be added in gemfile with `require: false` param
https://github.com/rubocop-hq/rubocop/blame/master/README.md#L47

Does `rubocop-rspec` should be default with same param? Or it inherit this value from `rubocop`?

ShockwaveNN:
Since rubocop and its extension used only in development in command line
Fix #857

---

* [x] Added tests.

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,7 +20,7 @@ gem install rubocop-rspec
 or if you use bundler put this in your `Gemfile`
 
 ```
-gem 'rubocop-rspec'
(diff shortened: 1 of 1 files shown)


bquorning:
Thank you 👍

PR #860, opened 2020-01-16
[Close #856] implement repeated example group description/body cops

### LINKED ISSUE #856: Cop idea: detect repeated describe/context description
In the company, we had a situation when someone pushed specs with duplicated context occasionally and only PR reviewers caught it.
```ruby
context "when cool thing happens" do
    # some examples
end

context "when cool thing happens" do
    # some examples
end
```

We tried to find some cops in `rubocop-rspec` which will detect this kind of occasion, but there were only `RSpec/RepeatedExample` and `RSpec/RepeatedDescription` and these cops do not detect repeated descriptions in `describe/context`, only examples.
So my question is, does it worth to implement different cops for `describe/context` like above ones or we can extend them to cover these cases?

lazycoder9:
There are 2 cops, that ensures that we don't have repeated specs.

`RSpec/RepeatedExampleGroupDescription`
Checks for repeated descriptions in example group
Example:

```ruby
# bad
describe 'cool feature' do
  # example group
end

describe 'cool feature' do
  # example group
end

# good
describe 'cool feature' do
  # example group
end

describe 'another cool feature' do
  # example group
end
```

`RSpec/RepeatedExampleGroupBody`
Checks for repeated body or implementation for example group

```ruby
# bad
context 'when case x' do
  it { cool_predicate }
end

context 'when case y' do
  it { cool_predicate }
end

# good
context 'when case x' do
  it { cool_predicate }
end

context 'when case y' do
  it { another_predicate }
end
```

Fixes #856 

---

* [x] Added tests.


lazycoder9:
Hey @pirj thanks for your comment on the issue, it allows me to make better merge and implement all that I wanted with 2 cops. Looking for your review

lib/rubocop/cop/rspec/repeated_block_description.rb
```diff
@@ -0,0 +1,74 @@
...
+      #
+      #    # good
+      #    context 'when case x' do
+      #      # example group
+      #    end
+      #
+      #    describe 'when another case' do
+      #      # example group
+      #    end
+      #
```
pirj: Can you please add an example for several similar example groups defined in the same context (most frequently seen in root context actually) with an additional docstring?
```ruby
RSpec.describe 'A', 'dog' do
  # ...
end
RSpec.describe 'A', 'cat' do
  # ...
end
```

lib/rubocop/cop/rspec/repeated_block_description.rb
```diff
@@ -0,0 +1,74 @@
...
+      class RepeatedBlockDescription < Cop
+        MSG = "Don't repeat block descriptions"
+
+        def_node_matcher :several_blocks?, <<-PATTERN
+          (begin (block (send nil? {:context :describe} ...) ...)+ )
```
pirj: Please use `ExampleGroup::ALL` to cover `feature`/`example_group`/`fdescribe` etc.

spec/rubocop/cop/rspec/repeated_block_body_spec.rb
```diff
@@ -0,0 +1,57 @@
...
+
+  it 'registers an offense for repeated describe body' do
+    expect_offense(<<-RUBY)
+      describe 'doing x' do
+      ^^^^^^^^^^^^^^^^^^^^^ Don't repeat block body
+        it { cool_predicate_method }
+      end
+
+      describe 'doing y' do
+      ^^^^^^^^^^^^^^^^^^^^^ Don't repeat block body
```
pirj: Sometimes it might be hard to figure out where it is repeated. Would it be possible to point at the duplication?

spec/rubocop/cop/rspec/repeated_block_body_spec.rb
```diff
@@ -0,0 +1,57 @@
...
+      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ Don't repeat block body
+        it { cool_predicate_method }
+      end
+
+      context 'when another awesome case' do
+      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ Don't repeat block body
+        it { cool_predicate_method }
+      end
+    RUBY
+  end
```
pirj: What if one of them is a `context` and another is `describe`?

spec/rubocop/cop/rspec/repeated_block_body_spec.rb
lazycoder9: @pirj I am thinking about pointing to the line of code, WDYT?

pirj: That would be ideal.

spec/rubocop/cop/rspec/repeated_block_body_spec.rb
lazycoder9: I have added that case and there should not be the offense. But I'm not sure if it is the right behavior

lib/rubocop/cop/rspec/repeated_example_group_description.rb
```diff
@@ -0,0 +1,98 @@
...
+      #    # bad
+      #    context 'when case x' do
+      #      # example group
+      #    end
+      #
+      #    describe 'when case x' do
+      #      # example group
+      #    end
+      #
+      #    # bad
```
pirj: It's actually quite legit.

spec/rubocop/cop/rspec/repeated_example_group_body_spec.rb
```diff
@@ -0,0 +1,69 @@
+# frozen_string_literal: true
+
+RSpec.describe RuboCop::Cop::RSpec::RepeatedExampleGroupBody do
+  subject(:cop) { described_class.new }
+
+  it 'registers an offense for repeated describe body' do
+    expect_offense(<<-RUBY)
+      describe 'doing x' do
+      ^^^^^^^^^^^^^^^^^^^^^ Repeated describe block body on line(s) [5]
```
pirj: Awesome! 👏

spec/rubocop/cop/rspec/repeated_example_group_body_spec.rb
```diff
@@ -0,0 +1,69 @@
...
+        it { cool_predicate_method }
+      end
+
+      context 'when another awesome case' do
+        it { another_predicate_method }
+      end
+    RUBY
+  end
+
+  it 'does not register offense if rspec tag magic is involved' do
```
pirj: I love the term :D
WDYT of "if metadata is different"?

spec/rubocop/cop/rspec/repeated_example_group_body_spec.rb
```diff
@@ -0,0 +1,69 @@
...
+        it { cool_predicate_method }
+      end
+
+      describe 'doing x', :request do
+        it { cool_predicate_method }
+      end
+    RUBY
+  end
+
+  it 'does not register offense for different groups' do
```
pirj: It should I believe.

spec/rubocop/cop/rspec/repeated_example_group_description_spec.rb
```diff
@@ -0,0 +1,84 @@
...
+      RSpec.describe 'Animal', 'dog' do
+      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ Repeated describe block description on line(s) [5]
+        # example group
+      end
+
+      RSpec.describe 'Animal', 'cat' do
+      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ Repeated describe block description on line(s) [1]
+        # example group
+      end
+    RUBY
```
pirj: Maybe add an example here for mixed example group types as well (e.g. `context` + `describe`).

spec/rubocop/cop/rspec/repeated_example_group_description_spec.rb
```diff
@@ -0,0 +1,84 @@
...
+  end
+end
```
pirj: And one last case that comes to mind - that it should not flag groups with no descriptions.

spec/rubocop/cop/rspec/repeated_example_group_body_spec.rb
```diff
@@ -0,0 +1,69 @@
...
+    expect_no_offenses(<<-RUBY)
+      describe 'doing x' do
+        it { cool_predicate_method }
+      end
+
+      context 'doing x' do
+        it { cool_predicate_method }
+      end
+    RUBY
+  end
```
pirj: Another thing that is nice to be certain about is that it only flags groups that are in the exact same group only, e.g.
```
describe 'A' do
  describe '.b' do
    context 'when this' do
    end
  end
  describe '.c' do
    context 'when this' do
    end
  end
  context 'when this' do
  end
end
```
Nothing here should be flagged.

CHANGELOG.md
```diff
@@ -5,6 +5,8 @@
 * Fix `RSpec/InstanceVariable` detection inside custom matchers. ([@pirj][])
 * Fix `RSpec/ScatteredSetup` to distinguish hooks with different metadata. ([@pirj][])
 * Add autocorrect support for `RSpec/ExpectActual` cop. ([@dduugg][], [@pirj][])
+* Add `RSpec/RepeatedBlockBody` copy. ([@lazycoder9][])
+* Add `RSpec/RepeatedBlockDescription` copy. ([@lazycoder9][])
```
pirj: Old names.

config/default.yml
```diff
@@ -394,6 +394,16 @@ RSpec/RepeatedExample:
   Description: Check for repeated examples within example groups.
   StyleGuide: https://www.rubydoc.info/gems/rubocop-rspec/RuboCop/Cop/RSpec/RepeatedExample
 
+RSpec/RepeatedExampleGroupBody:
+  Enabled: true
+  Description: Check for repeated describe and context block body.
```
pirj: "describe and context" => "example group", it's more generic.

lib/rubocop/cop/rspec/repeated_example_group_body.rb
```diff
@@ -0,0 +1,94 @@
...
+      #
+      #    context 'when case y' do
+      #      it { another_predicate }
+      #    end
+      #
+      class RepeatedExampleGroupBody < Cop
+        MSG = 'Repeated %<group>s block body on line(s) %<loc>s'
+
+        def_node_matcher :several_blocks?, <<-PATTERN
+          (begin (block (send _ {#{ExampleGroups::ALL.node_pattern}} ...) ...)+ )
```
pirj: Elegant!

pirj: It's a nice idea to use `#{RSPEC}` instead of the first `_`, it's making sure the receiver is either `nil?` or `:RSpec`, so only `describe do` and `RSpec.describe do` will trigger, not `helper.describe do`.

lib/rubocop/cop/rspec/repeated_example_group_description.rb
lazycoder9: @pirj so you say, that if there is repeated description, but have additional docstring or tags, it is ok and there should not be offenses, right?

lib/rubocop/cop/rspec/repeated_example_group_body.rb
```diff
@@ -0,0 +1,94 @@
...
+            group, repeats = repeated_group.values_at(:group, :repeats)
+            add_offense(group, message: message(group, repeats))
+          end
+        end
+
+        private
+
+        def repeated_group_bodies(node)
+          node
+            .children
```
pirj: This is quite scary. What if a method is defined, or some other statement. Won't it fail?
Not sure if this needs an example (there are some in smoke test), but is still worth checking.
If fails, please feel free to extend `spec/smoke_tests/weird_rspec_spec.rb`.

lib/rubocop/cop/rspec/repeated_example_group_description.rb
```diff
@@ -0,0 +1,98 @@
+# frozen_string_literal: true
```
pirj: Are there any worth freezing? :D

lib/rubocop/cop/rspec/repeated_example_group_description.rb
```diff
@@ -0,0 +1,98 @@
...
+      #    end
+      #
+      #    # good
+      #    context 'when case x' do
+      #      # example group
+      #    end
+      #
+      #    context 'when another case' do
+      #      # example group
+      #    end
```
pirj: Keeping in mind there's no big difference between `context`/`describe`/`RSpec.describe` I'm not sure if so many good examples are needed.

lib/rubocop/cop/rspec/repeated_example_group_description.rb
```diff
@@ -0,0 +1,98 @@
...
+      #
+      class RepeatedExampleGroupDescription < Cop
+        MSG = 'Repeated %<group>s block description on line(s) %<loc>s'
+
+        def_node_matcher :several_groups?, <<-PATTERN
+          (begin (block (send _ {#{ExampleGroups::ALL.node_pattern}} ...) ...)+ )
+        PATTERN
+
+        def_node_matcher :extract_doc_string, '(block (send _ _ $str ...) ...)'
```
pirj: Are you sure that `extract_` adds some value in the method name?

lib/rubocop/cop/rspec/repeated_example_group_description.rb
```diff
@@ -0,0 +1,98 @@
...
+
+        def on_begin(node)
+          return unless several_groups?(node)
+
+          repeated_group_descriptions(node).each do |repeated_group|
+            group, repeats = repeated_group.values_at(:group, :repeats)
```
pirj: Cosmetic. Maybe use an array of two elements instead of a hash?

lib/rubocop/cop/rspec/repeated_example_group_description.rb
lazycoder9: Also I've checked similar cop `RSpec/RepeatedDescription` for example and I see offenses for such kind of piece of code
```ruby
it 'do something', :flag do
  # ...
end

it 'do something', :flag do
  # ...
end
```

Maybe I'm missing something, didn't understand your point fully or we have different rules for example groups

pirj: I can't find a good real example, but I remember seeing some this in [`real-world-ruby-apps`](https://github.com/jeromedalbert/real-world-ruby-apps/) or [`real-world-rails`](https://github.com/eliotsykes/real-world-rails).

RSpec maintainers consider additional docstrings as a feature that should fade away https://github.com/rspec/rspec-core/pull/2681#discussion_r361811453

However, I'd stay strong for metadata, sometimes it's even used to construct examples/example groups. This metadata is so descriptive that it doesn't make sense to change the title, e.g.:
```ruby
it 'runs pretty quickly', load_factor: 20 do
end

it 'runs pretty quickly', load_factor: 100 do
end
```
I see those as two distinct examples and see no good reason to flag them as repeated.
The same applies to similar bodies. Even though they might be exactly the same, it doesn't mean there's repetition that should by DRY'd out.

For the example groups, specifically:
```ruby
context group: :guest do
  it ...
end
```
actually constructs a wrapping context with a generated docstring `when the group is "guest"`, and it makes it distinct from the other one.

Using metadata might also be seen similar to passing values to shared examples, you wouldn't consider two calls if `it_behaves_like` with different parameters repetition, would you?

`RepeatedDescription` does not take metadata into account, and I see it as a flaw, not as an intended design.

Anyway, I'm all for discussing what are your rules for example groups.

lazycoder9: @pirj Ok, I got it, thanks for the detailed answer, I will fix it then

spec/rubocop/cop/rspec/repeated_example_group_description_spec.rb
lazycoder9: @pirj According to your [comment](https://github.com/rubocop-hq/rubocop-rspec/pull/860#discussion_r367634790) it should be an offense for mixed example group with similar description

spec/rubocop/cop/rspec/repeated_example_group_body_spec.rb
lazycoder9: @pirj it should flag `describe .b and .c` because they have a similar body. Maybe the main point here that it should not flag `context` in and out of `describe` group`?

lib/rubocop/cop/rspec/repeated_example_group_description.rb
lazycoder9: @pirj just tried to make method names self explainable

lib/rubocop/cop/rspec/repeated_example_group_description.rb
lazycoder9: Configured my editor to auto fix offenses, so this line is added automatically. And I see this line everywhere in the source code. And yes, there is `MSG` string literal, so maybe it is still worth to freeze it :)

lib/rubocop/cop/rspec/repeated_example_group_body.rb
lazycoder9: @pirj Sorry, I did not understand your point of view, since I am not an expert in RSpec. Are you afraid of the `children` method or the implementation of `repeat_group_bodies`?

pirj: What I mean is that example groups can contain arbitrary statements, e.g:
```ruby
RSpec.describe 'A' do
  stub_all_http_calls()
  allowed_statuses = %i[open submitted approved].freeze
  before { create(:admin) }

  describe '#load' do
    ...
  end
  describe '#load' do
    ...
  end
end
```

Will offences be detected?
I see two possible culprits, one of them is `several_blocks?` that wouldn't trigger on such a spec, and another is that `.children` would return those other statements as well.

lib/rubocop/cop/rspec/repeated_example_group_body.rb
```diff
@@ -0,0 +1,89 @@
...
+      #    context 'when case y' do
+      #      it { another_predicate }
+      #    end
+      #
+      class RepeatedExampleGroupBody < Cop
+        MSG = 'Repeated %<group>s block body on line(s) %<loc>s'
+
+        def_node_matcher :several_blocks?, '(begin #example_group_with_body?+ )'
+
+        def_node_matcher :metadata, '(block (send _ _ _ $...) ...)'
```
pirj: Does it properly detect the case with two docstrings?
```ruby
describe 'A', '#method', :uses_js, type: :request do
end
```

lib/rubocop/cop/rspec/repeated_example_group_body.rb
```diff
@@ -0,0 +1,89 @@
...
+            .children
+            .group_by { |child| group_signature(child) }
+            .values
+            .reject(&:one?)
+            .map { |groups| add_repeated_lines(groups) }
+            .flatten(1)
+        end
+
+        def group_signature(group)
+          [metadata(group), body(group)]
```
pirj: Since those two are used together, probably combine `metadata` and `body` matchers, and use two captures in that common matcher?

spec/rubocop/cop/rspec/repeated_example_group_body_spec.rb
```diff
@@ -0,0 +1,104 @@
...
+
+      RSpec.describe 'doing x' do
+      ^^^^^^^^^^^^^^^^^^^^^^^^^^^ Repeated describe block body on line(s) [5]
+        it { cool_predicate_method }
+      end
+
+      context 'when a is true' do
+      ^^^^^^^^^^^^^^^^^^^^^^^^^^^ Repeated context block body on line(s) [1]
+        it { cool_predicate_method }
+      end
```
pirj: The reason this fails is that a group with `helpers.describe` is not detected as `several_blocks?`.
It's possible to use children matching in the pattern, please check some examples [here](https://github.com/rubocop-hq/rubocop-rspec/pull/804/files).
I'm not 100% certain, but it might not support `+`, and this would result in example groups often being the only ones. It's worth to double-check, maybe it does support `+`, I just didn't try.

spec/rubocop/cop/rspec/repeated_example_group_body_spec.rb
pirj: Sorry for the confusion, edited the comment. I tried to empasize on `context 'when this'`.

spec/rubocop/cop/rspec/repeated_example_group_description_spec.rb
pirj: 👍

lib/rubocop/cop/rspec/repeated_example_group_body.rb
lazycoder9: @pirj yeah, greate note, will check it

lazycoder9:
@pirj Hello, I've resolved all of your issues and made some enhances in node matchers. Hope it will be your last review 😄

CHANGELOG.md
```diff
@@ -5,6 +5,8 @@
 * Fix `RSpec/InstanceVariable` detection inside custom matchers. ([@pirj][])
 * Fix `RSpec/ScatteredSetup` to distinguish hooks with different metadata. ([@pirj][])
 * Add autocorrect support for `RSpec/ExpectActual` cop. ([@dduugg][], [@pirj][])
+* Add `RSpec/RepeatedExampleGroupBody` copy. ([@lazycoder9][])
```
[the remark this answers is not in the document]
lazycoder9: @pirj Vim bro? 😄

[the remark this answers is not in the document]
pirj: Yes :D

lazycoder9:
@pirj You have left some comments, but anyway approved PR, why?) And also, I've fixed your last 4 tensions and added test cases for them. Looking for your review

lazycoder9:
@pirj sorry, requested re-review out of habit 😄

spec/rubocop/cop/rspec/repeated_example_group_description_spec.rb
```diff
@@ -0,0 +1,202 @@
...
+        ^^^^^^^^^^^^^^^^^^^ Repeated describe block description on line(s) [6]
+          it { cool_predicate_method }
+        end
+      end
+    RUBY
+  end
+
+  it 'considers interpolated docstrings as different descriptions' do
+    expect_no_offenses(<<-RUBY)
+      context "when class is \#{RuboCop::RSpec}" do
```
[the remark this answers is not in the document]
lazycoder9: @pirj Yes, I also came across this when I tried to test the case, so I used a real constant. After, I found that I can use a slash to get the corresponding `dstr` in AST, but forgot to change the constants to fake. Fixed

lib/rubocop/cop/rspec/repeated_example_group_description.rb
```diff
@@ -0,0 +1,93 @@
...
+        def_node_matcher :several_example_groups?, <<-PATTERN
+          (begin <#example_group? #example_group? ...>)
+        PATTERN
+
+        def_node_matcher :doc_string_and_metadata, <<-PATTERN
+          (block (send _ _ $_ $...) ...)
+        PATTERN
+
+        def_node_matcher :skip_or_pending?, <<-PATTERN
+          (block <(send _ {:skip :pending}) ...>)
```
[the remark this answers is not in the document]
lazycoder9: Fixed

pirj:
@bquorning @Darhazer do you mind to take a look?

Darhazer:
Tests-wise, can we check that repeated examples are flagged if there is something else between them?
```
describe 'repeated' do
  it { is_expected.to be_truthy }
end

before { do_something }

describe 'this is repeated' do
  it { is_expected.to be_truthy }
end
```

Can we test that context are not flagged for having the same example on different data?
```
context 'when admin' do
  let(:user) { admin }
  it { is_expected.to be_truthy }
end

context 'when regular user' do
  let(:user) { staff }
  it { is_expected.to be_truthy }
end
```

lazycoder9:
@Darhazer Done

Darhazer:
Thank you.
I've tested on a real codebase, and it has a false positive when the example refers to a `let` variable, and the let value is different in the different contexts.

This might be hard to catch using static analysis, and I see value in the `RepeatedExampleGroupBody` cop. Perhaps we can just mark it as Unsafe cop. Other options include to skip example groups with different variables set; or actually check if those variables are used within the repeated example

Darhazer:
It also fails on examples/group without descriptions, that are actually different

```
context do
  let(:preferences) { default_preferences }

  it { is_expected.to eq false }
end

context do
  let(:preferences) { %w[a] }

  it { is_expected.to eq true }
end
```

pirj:
> false positive when the example refers to a `let` variable and the let value is different in the different contexts.

@Darhazer But how is this possible for sibling contexts? We account for metadata differences already.

lazycoder9:
> It also fails on examples/group without descriptions, that are actually different
> 
> ```
> context do
>   let(:preferences) { default_preferences }
> 
>   it { is_expected.to eq false }
> end
> 
> context do
>   let(:preferences) { %w[a] }
> 
>   it { is_expected.to eq true }
> end
> ```

@Darhazer how we should handle such cases? Just ignore them and another cop (maybe new) should handle it?

Darhazer:
@pirj it's not metadata, it's a `let` node in the example group, but the cop selects the example nodes only, e.g. ignores anything else inside the example group

@lazycoder9 there's nothing to handle there - it's a false positive - the examples are different and should not be flagged as repeated.

lazycoder9:
@Darhazer Fixed false positive with example groups without descriptions. I'm going to think about false positive case with `let` variables

pirj:
@Darhazer I couldn't come up with a better example:
```ruby
  it 'registers an offense for repeated context body' do
    expect_offense(<<-RUBY)
      context 'when awesome case' do
      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ Repeated context block body on line(s) [5]
        let(:a) { 1 }
        it { cool_predicate_method }
      end

      context 'when another awesome case' do
      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ Repeated context block body on line(s) [1]
        let(:a) { 2 }
        it { cool_predicate_method }
      end
    RUBY
  end
```

And it fails since cop considers those two examples different. Can you please provide a minimal reproduction case?

Darhazer:
`cool_predicate_method(a)` would be different in the two examples, although their AST would remain the same

lazycoder9:
@Darhazer I've tried your example with `cool_predicate_method(a)` as
```ruby
  it 'does not register offense' do
    expect_no_offenses(<<-RUBY)
      context 'when awesome case' do
        let(:a) { 1 }
        it { cool_predicate_method(a) }
      end

      context 'when another awesome case' do
        let(:a) { 2 }
        it { cool_predicate_method(a) }
      end
    RUBY
  end
```
And it works. Maybe another cop was triggered in this code? I think about `RepeatedExample`. @pirj thoughts?

pirj:
Tried on two more pretty large codebases, no signs of false positives. Only traces of copy-pasta, inspiring false confidence in code impeccancy.
@Darhazer your only option is to reveal that example.

lib/rubocop/cop/rspec/repeated_example_group_body.rb
```diff
@@ -75,6 +75,10 @@ def add_repeated_lines(groups)
           groups.map { |group| [group, repeated_lines - [group.first_line]] }
         end
 
+        def signature_keys(group)
+          [metadata(group), body(group)]
```
Darhazer: the `Example` node already has `metadata` and `implementation` methods, so you might be able to reuse them

Darhazer:
@pirj perhaps it was connected with the previous false positives - with the latest code, I don't get any.

I'll make a final review early next week.

lib/rubocop/cop/rspec/repeated_example_group_body.rb
lazycoder9: I thought about it firstly, but it is `ExampleGroup` in the current context, not `Example` and I gave the idea to reuse `Example` methods, it will confuse domain concepts. Another option is to implement methods in `ExampleGroup` wrapper and use them. What do you think @Darhazer?

lib/rubocop/cop/rspec/repeated_example_group_body.rb
```diff
@@ -0,0 +1,88 @@
...
+      #      it { another_predicate }
+      #    end
+      #
+      #    # good
+      #    context 'when case x', :tag do
+      #      it { cool_predicate }
+      #    end
+      #
+      #    context 'when case y' do
+      #      it { another_predicate }
```
Darhazer: It's not clear how this good example differs from the previous one. Perhaps here the body should be the same, but the tag makes the difference?

lib/rubocop/cop/rspec/repeated_example_group_body.rb
Darhazer: While it is great to extend the Language and the Node concepts in rubocop-rspec with such things, let keep it outside of the current scope. Feel free to open a PR with related refactorings later if you'd like

lib/rubocop/cop/rspec/repeated_example_group_description.rb
```diff
@@ -0,0 +1,93 @@
...
+        end
+
+        private
+
+        def repeated_group_descriptions(node)
+          node
+            .children
+            .select { |child| example_group?(child) }
+            .reject { |child| skip_or_pending?(child) }
+            .group_by { |group| doc_string_and_metadata(group) }
```
Darhazer: This also gives a repeated description offense when there is no description at all. While it's repeating non-description, I don't think it makes sense to flag those. Perhaps another cop can flag empty example group descriptions.
@pirj what do you think?

spec/rubocop/cop/rspec/repeated_example_group_body_spec.rb
```diff
@@ -0,0 +1,216 @@
+# frozen_string_literal: true
+
+RSpec.describe RuboCop::Cop::RSpec::RepeatedExampleGroupBody do
+  subject(:cop) { described_class.new }
+
+  it 'registers an offense for repeated describe body' do
+    expect_offense(<<-RUBY)
+      describe 'doing x' do
+      ^^^^^^^^^^^^^^^^^^^^^ Repeated describe block body on line(s) [5]
```
Darhazer: You are using `line(s)` in the message. There is no test of multiple repetitions though.

spec/rubocop/cop/rspec/repeated_example_group_body_spec.rb
```diff
@@ -0,0 +1,216 @@
...
+      end
+    RUBY
+  end
+
+  it 'registers offense only for example group' do
+    expect_offense(<<-RUBY)
+      RSpec.describe 'A' do
+        stub_all_http_calls()
+        allowed_statuses = %i[open submitted approved].freeze
+        before { create(:admin) }
```
Darhazer: Since those are not duplicated, it is not clear how this test proves its description

spec/rubocop/cop/rspec/repeated_example_group_body_spec.rb
```diff
@@ -0,0 +1,216 @@
...
+      context do
+        let(:preferences) { default_preferences }
+
+        it { is_expected.to eq false }
+      end
+
+      context do
+        let(:preferences) { %w[a] }
+
+        it { is_expected.to eq true }
```
Darhazer: What if the examples were the same? Would it register the offense?

spec/rubocop/cop/rspec/repeated_example_group_description_spec.rb
```diff
@@ -0,0 +1,232 @@
...
+        # ...
+      end
+
+      context "when class is \#{C::D}" do
+        # ...
+      end
+    RUBY
+  end
+
+  it 'registers offense correctly for interpolated docstrings' do
```
Darhazer: Should it? if there is a variable and the variable value changes between the two context calls?
Although this is such an edge case that I could live with the false positives. Actually it would be very bad code to start with

lib/rubocop/cop/rspec/repeated_example_group_description.rb
pirj: Absolutely. Brought it up earlier [here](https://github.com/rubocop-hq/rubocop-rspec/pull/860/files#r367636161).

spec/rubocop/cop/rspec/repeated_example_group_description_spec.rb
pirj: I hope there's no one in the world writing `describe "when time is #{Time.now.to_f}" do` and repeat that exact description :D It's better to flag this even though technically those two will be two different descriptions.

spec/rubocop/cop/rspec/repeated_example_group_body_spec.rb
lazycoder9: My bad, actually this test should be in `RepeatedExampleGroupDescription` spec, but it is a good case for `RepeatedExampleGroupBody` too

lazycoder9:
@pirj @Darhazer Hey! Tried to fix all of your latest questions\comments. Squashed commits. Thanks a lot guys, it was a really good experience for me

pirj:
Thanks for those amazing new cops!

pirj:
And you are always welcome to contribute more cops and come up with ideas for cops/improvements.

marcandre:
Great addition 🎉 ! There [was a case in `rubocop-ast` 😄 ](https://github.com/rubocop-hq/rubocop-ast/commit/04cb7f38af373424b68a3048167cc05bfd664ff9) too :-)

PR #863, opened 2020-01-20
Add MultipleMemoizedHelpers cop

### LINKED ISSUE #862: Cop Idea: NoLet
It would be nice to have a rule that disallows the use of `let`. I've been moving in the direction of avoiding them in my tests. Instead I'll err on the side of a little more duplication in my tests. In cases where there is a lot of setup, I'll instead define methods or factories to encapsulate the logic. More in [this thoughtbot article](https://thoughtbot.com/blog/lets-not).

mockdeep:
Fixes #862

* [X] Added tests.
* [X] Added an entry to the [changelog](https://github.com/rubocop-hq/rubocop-rspec/blob/master/CHANGELOG.md) if the new code introduces user-observable changes.

* [X] Added the new cop to `config/default.yml`.
* [X] The cop documents examples of good and bad code.
* [X] The tests assert both that bad code is reported and that good code is not reported.


CHANGELOG.md
```diff
@@ -5,6 +5,7 @@
 * Fix `RSpec/InstanceVariable` detection inside custom matchers. ([@pirj][])
 * Fix `RSpec/ScatteredSetup` to distinguish hooks with different metadata. ([@pirj][])
 * Add autocorrect support for `RSpec/ExpectActual` cop. ([@dduugg][], [@pirj][])
+* Add `RSpec/NoLet` cop. ([@mockdeep][])
```
pirj: You have to add a reference to yourself at the end of this file.

lib/rubocop/cop/rspec/no_let.rb
```diff
@@ -0,0 +1,85 @@
...
+          (send nil? :let ...)
+        PATTERN
+
+        def_node_matcher :subject?, <<~PATTERN
+          (send nil? :subject ...)
+        PATTERN
+
+        def on_send(node)
+          return unless let?(node) || subject?(node)
+          return if subject?(node) && allow_subject?
```
pirj: If you swap those two lines, the `|| subject?` becomes redundant.

spec/rubocop/cop/rspec/no_let_spec.rb
```diff
@@ -0,0 +1,64 @@
...
+    subject(:cop) { described_class.new(config) }
+
+    let(:cop_config) do
+      { 'AllowSubject' => true }
+    end
+
+    it 'registers an offense when using `#let`' do
+      expect_offense(<<~RUBY)
+        let(:foo) { Foo.new }
+        ^^^^^^^^^ Avoid using `let` and `subject` – use a method call or local variable instead.
```
pirj: This message doesn't fit well when `subject` is allowed.

spec/rubocop/cop/rspec/no_let_spec.rb
```diff
@@ -0,0 +1,64 @@
...
+      RUBY
+    end
+
+    it 'does not register an offense when using `#subject` without a name' do
```
pirj: Minor: most often 'flags'/'ignores' are used as prefixes, but this is actually quite diverse.

lib/rubocop/cop/rspec/no_let.rb
mockdeep: @pirj I'm not sure I understand this one. I tried this and it didn't flag `subject` calls:

```rb
return if subject?(node) && allow_subject?
return unless let?(node)
```

mockdeep:
@pirj updated per your comments. Thanks for the review!

lib/rubocop/cop/rspec/no_let.rb
pirj: Sorry, my bad. The way you've written it is better than `return unless let?(node) || (subject?(node) && allow_subject?)`.

mockdeep:
Woops, `TODO` removed.

lib/rubocop/cop/rspec/no_let.rb
```diff
@@ -0,0 +1,85 @@
...
+      #       expect(foo).to be_empty
+      #     end
+      #   end
+      #
+      class NoLet < Cop
+        MSG = 'Avoid using `%<method>s` ' \
+              '– use a method call or local variable instead.'
+
+        def_node_matcher :let?, <<~PATTERN
+          (send nil? :let ...)
```
Darhazer: This misses `let!`. Better utilize the language module - `Helpers::ALL.send_pattern`

lib/rubocop/cop/rspec/no_let.rb
```diff
@@ -0,0 +1,85 @@
...
+        PATTERN
+
+        def_node_matcher :subject?, <<~PATTERN
+          (send nil? :subject ...)
```
Darhazer: This missed `subject!`. Better use `Subject:ALL.send_patern`

mockdeep:
Discussing renaming the rule [back in the issue](https://github.com/rubocop-hq/rubocop-rspec/issues/862#issuecomment-576757628).

pirj:
Let's keep this as is, as this suits your needs. I'm pretty sure this will come handy for other people. cc @sshaw

Do you happen to use `let!`/`subject!` in your codebase, @mockdeep ? It's a trivial change as @Darhazer suggests to detect and flag them as well.

mockdeep:
@pirj I'm happy to implement the other approach. It'll help narrow down in codebases that already make heavy use of `let` and keep it to a reasonable level where people aren't as dogmatic about it as I am.

Sorry about the delay in response. I've been home with a sick kid.

pirj:
Didn't intend to push a commit, thought it will add a suggested edit Github #$%^$

So, the plan to match what was discussed in the ticket:
 - [ ] Rename to `MemoizedHelpersInExampleGroup`
 - [ ] add a `Max` setting
 - [ ] for each example group
```ruby
def on_send(node)
```
   find `let`s/`subject`s, put them to a `Set` by var's name (`nil` for nameless `subject`)
- [ ] walk up the ancestor chain with `each_ancestor`, add their `let`s/`subject`s to that set
- [ ] if the count in the set exceeds `Max`, raise an offence
- [ ] Profit!

There should be plenty of examples in the code, but if you get stuck please don't hesitate to ping.

mockdeep:
@pirj I renamed the file. It might take me a few days to get to the rest of it. With the COVID-19 drama, I've got my kid home, so that leaves me like 1 hour a day of concentrated work time while he's napping.

pirj:
No worries @mockdeep, take your time and take care.
I know how it is to have kids at home :D

pirj:
@mockdeep A friendly reminder 🔔

mockdeep:
@pirj here's a working implementation. I'm planning on doing some cleanup on the tests and docs, as well as maybe de-duplicating some of the logic in `ExampleGroup`, but I'd appreciate any feedback you have in the meantime.

lib/rubocop/cop/rspec/memoized_helpers_in_example_group.rb
```diff
@@ -0,0 +1,136 @@
...
+      #     end
+      #
+      #     context 'when other stuff' do
+      #       subject { {} }
+      #       subject(:wat) { {} }
+      #       subject!(:boo) { {} }
+      #     end
+      #   end
+      #
+      # @example with AllowSubject configuration
```
pirj, line 58: 👍
There's `MultipleSubjects` to take care of more than one subject.
However, unlike this cop, it has a deficiency of not counting inside nested example groups.

lib/rubocop/cop/rspec/memoized_helpers_in_example_group.rb
```diff
@@ -0,0 +1,136 @@
...
+      #     let(:foo) { [] }
+      #   end
+      #
+      #   # good
+      #   describe MyClass do
+      #     def foo; []; end
+      #   end
+      #
+      class MemoizedHelpersInExampleGroup < Cop
+        MSG = 'Example has too many memoized helpers [%<count>d/%<max>d]'
```
pirj, line 101: "Example group"?

pirj:
@mockdeep I lost track of what is the state of this.
Do you need some help, at least with rebasing on the latest master?

mockdeep:
@pirj I just rebased it. Sorry, I've got this on my list, but it's always 3rd or 4th. I think it should be mostly good to go, so feel free to look it over and see if there's anything important I'm missing on it.

pirj:
> got this on my list, but it's always 3rd or 4th

Story of my life 👬

pirj:
Note to self:
```
$ hub pr checkout 863
From github.com:rubocop-hq/rubocop-rspec
 ! [rejected]        refs/pull/863/head -> rf-no_let  (non-fast-forward)
```
can be fixed by:
```
git branch -D rf-no_let
```

config/default.yml
```diff
@@ -394,6 +394,13 @@ RSpec/MultipleExpectations:
   VersionChanged: '1.21'
   StyleGuide: https://www.rubydoc.info/gems/rubocop-rspec/RuboCop/Cop/RSpec/MultipleExpectations
 
+RSpec/MultipleMemoizedHelpers:
+  Description: Checks if example groups contain too many `let` and `subject` calls.
+  Enabled: false
```
pirj, line 399: I would personally be happy to enable it by default with `AllowSubject: true` and `Max: 5`.

pirj:
Sorry, I've put this comment in the wrong place.

Wondering if:
```ruby
describe Weather do
  let(:temp) { 20.celsius }
  context "when it's warm" do
    let(:temp) { 30.celsius } # <== this should it count as an additional `let`, or should it collapse with an existing one
```

probably we need to collapse counts of memoized helpers by names.

CHANGELOG.md
```diff
@@ -8,6 +8,7 @@
 * Relax `RSpec/VariableDefinition` cop so interpolated and multiline strings are accepted even when configured to enforce the `symbol` style. ([@bquorning][])
 * Fix `RSpec/EmptyExampleGroup` to flag example groups with examples in invalid scopes. ([@mlarraz][])
 * Fix `RSpec/EmptyExampleGroup` to ignore examples groups with examples defined inside iterators. ([@pirj][])
+* Add `RSpec/MemoizedHelpersInExampleGroup` cop. ([@mockdeep][])
```
Darhazer: Seems we have some inconsistency in the naming. Isn't it `MultipleMemoizedHelpers` ?

config/default.yml
Darhazer: Did you run stats on some real world projects to see how many offenses are generated at different `Max` levels?

Darhazer: Also we have a cop for multiple subjects, so I doubt the need to check subject calls at all

lib/rubocop/cop/rspec/multiple_memoized_helpers.rb
```diff
@@ -0,0 +1,139 @@
...
+      #   end
+      #
+      #   describe MyClass do
+      #     let(:foo) { [] }
+      #     let(:bar) { [] }
+      #
+      #     context 'when stuff' do
+      #       let!(:booger) { [] }
+      #       subject { {} }
+      #       subject(:wat) { {} }
```
Darhazer, line 30: multiple subjects are not a good example :joy:

lib/rubocop/cop/rspec/multiple_memoized_helpers.rb
```diff
@@ -0,0 +1,139 @@
...
+        MSG = 'Example group has too many memoized helpers [%<count>d/%<max>d]'
+
+        def on_block(node)
+          return unless spec_group?(node)
+
+          count = total_helpers(node)
+
+          return if count <= max
+
+          add_offense(node, message: format(MSG, count: count, max: max))
```
Darhazer, line 110: shouldn't we set `self.max` so it can set it during auto-gen config ?

lib/rubocop/cop/rspec/multiple_memoized_helpers.rb
```diff
@@ -0,0 +1,139 @@
...
+        end
+
+        private
+
+        def total_helpers(node)
+          helpers(node) +
+            node.each_ancestor(:block).map(&method(:helpers)).sum
```
Darhazer, line 117: This way we count overrides in nested context which is quite a valid usage for `let`. I'd rather count the helpers on a single level instead. Or you'll have to map them to names and count uniq names

CHANGELOG.md
pirj: Just spotted it as well 👍

lib/rubocop/cop/rspec/multiple_memoized_helpers.rb
pirj: Done!

config/default.yml
pirj: I'm thinking of gathering a Real World RSpec, since Real World Ruby/Real World Rails don't use RSpec much.

pirj: Here it goes https://github.com/pirj/real-world-rspec/

Had to remove all `.rubocop*.yml` files around those projects before running `rubocop` on them. RuboCop hiccups when there's an incompatible `TargetRubyVersion` or a renamed cop in the configuraion.

```
$ rubocop --only RSpec/MultipleMemoizedHelpers ~/source/real-world-rspec
Inspecting 47132 files
```
Max helpers | Offence count
--- | ---
`Max: 5` | 24640
`Max: 7` | 14418
`Max: 10` | 6093

Notable examples (@mockdeep, fasten your seatbelts):
10 lets https://github.com/forem/forem/blob/master/spec/support/api_analytics_shared_examples.rb#L2
20 lets https://github.com/chef/chef/blob/master/spec/unit/provider/systemd_unit_spec.rb#L23
30 lets https://github.com/chef/chef/blob/master/spec/unit/policy_builder/policyfile_spec.rb#L24
40 lets https://github.com/diaspora/diaspora/blob/develop/spec/integration/migration_service_spec.rb#L7
50 lets https://github.com/chef/chef/blob/master/spec/functional/resource/dsc_script_spec.rb#L71

I can prepare the report per-repo, but I guess those two, diaspora and chef, are the worst offenders.

pirj: > Also we have a cop for multiple subjects, so I doubt the need to check subject calls at all

I think `subject` counts towards memoized helpers, as its differences from `let` are subtle.

Even though with the default configuration we only allow one `subject`, some may want zero of them. I'm quite ok to make `AllowSubject: true` the default though for those who don't.

lib/rubocop/cop/rspec/multiple_memoized_helpers.rb
```diff
@@ -0,0 +1,137 @@
...
+
+        def helpers(node)
+          example_group = RuboCop::RSpec::ExampleGroup.new(node)
+          variables =
+            if allow_subject?
+              example_group.lets
+            else
+              example_group.lets + example_group.subjects
+            end
+          variables.map { |variable| variable_definition?(variable.send_node) }
```
Darhazer, line 124: It doesn't work when `let` is used with `block_pass`. The following causes a crash:

```
let(:task, &create_task)
```

pirj: Very well noticed! I'll add this construct to our `spec/smoke_tests/weird_rspec_spec.rb`.

lib/rubocop/rspec/variable.rb
```diff
@@ -8,7 +8,7 @@ module Variable
       extend RuboCop::NodePattern::Macros
 
       def_node_matcher :variable_definition?, <<~PATTERN
-        (send #rspec? #{(Helpers::ALL + Subject::ALL).node_pattern_union}
+        (send nil? #{(Helpers::ALL + Subject::ALL).node_pattern_union}
```
pirj, line 11: Because `RSpec.let` is not possible.

lib/rubocop/cop/rspec/multiple_memoized_helpers.rb
```diff
@@ -0,0 +1,141 @@
...
+            *helpers(node),
+            *node.each_ancestor(:block).flat_map(&method(:helpers))
+          ]
+        end
+
+        def helpers(node)
+          variable_nodes(node).map do |variable_node|
+            if variable_node.block_type?
+              variable_definition?(variable_node.send_node)
+            else # block-pass (`let(:foo, &bar)`)
```
pirj, line 116: Yet another candidate for the global block-pass support induced refactoring.

config/default.yml
pirj: The numbers should be slightly lower now when `subject` doesn't count towards the total number by default.

A [bug in RuboCop](https://github.com/rubocop-hq/rubocop/issues/8465) prevents from building a script that would run `rubocop` on each of 35 repositories and tell which ones offend the new cops most of all.

Here's an [example pull request](https://github.com/rubocop-hq/rubocop/pull/8447) showing that it's possible to reduce the use memoized helpers to a reasonable number.

config/default.yml
```diff
@@ -394,6 +394,13 @@ RSpec/MultipleExpectations:
   VersionChanged: '1.21'
   StyleGuide: https://www.rubydoc.info/gems/rubocop-rspec/RuboCop/Cop/RSpec/MultipleExpectations
 
+RSpec/MultipleMemoizedHelpers:
+  Description: Checks if example groups contain too many `let` and `subject` calls.
+  Enabled: true
```
pirj, line 399: ⚠️ 
I personally don't have many doubts that this cop should be enabled by default. Open for the discussion, though.

Just in case, we only have a handful of disabled cops, and we encourage (not documented) to tune the configuration to meet the project style, not bend all projects to our defaults.

Darhazer: I'm fine with having it enabled, with `AllowSubject` true, by default

pirj:
@bquorning Do you mind to take a look?

The reasoning for turning it on by default:
 - https://github.com/rubocop-hq/rubocop-rspec/pull/863#discussion_r466075338
 - https://github.com/rubocop-hq/rubocop-rspec/pull/863/commits/e3b1ec939ab73cf91756877058d6ee600c1ca411

lib/rubocop/cop/rspec/multiple_memoized_helpers.rb
```diff
@@ -0,0 +1,141 @@
...
+      #   #   AllowSubject: false
+      #
+      #   # bad - `subject` counts towards memoized helpers
+      #   describe MyClass do
+      #     let(:foo) { [] }
+      #     let(:bar) { [] }
+      #     let!(:baz) { [] }
+      #     let(:qux) { [] }
+      #     let(:quux) { [] }
+      #     subject { {} }
```
bquorning, line 71: Shouldn’t we move `subject` above the `let`s?

spec/rubocop/cop/rspec/multiple_memoized_helpers_spec.rb
```diff
@@ -0,0 +1,162 @@
...
+        end
+
+        context 'when blah' do
+          let(:bar) { Bar.new }
+        end
+      end
+    RUBY
+  end
+
+  context 'when using AllowSubject configuration', :config do
```
bquorning, line 102: Perhaps it would be clearer to write “when disabling AllowSubject configuration”?

lib/rubocop/cop/rspec/multiple_memoized_helpers.rb
```diff
@@ -0,0 +1,141 @@
...
+          self.max = count
+          add_offense(node, message: format(MSG, count: count, max: max))
+        end
+
+        private
+
+        def all_helpers(node)
+          [
+            *helpers(node),
+            *node.each_ancestor(:block).flat_map(&method(:helpers))
```
bquorning, line 108: In deeply nested code, or very long spec files, I imagine this line has the potential to be a performance hotspot, right? We _might_ want to do some memoization for each ancestor.

I’d need to do some measurement to figure out if it’s really a problem or not.

pirj: @bquorning Fair enough.
Ran it the following against `gitlabhq` and `discourse`:
```
$ bundle exec bin/rubocop-profile --only RSpec --force-default-config  --require rubocop-rspec ../real-world-rspec/gitlabhq ../real-world-rspec/discourse
$ stackprof tmp/stackprof.dump --method 'RuboCop::Cop::RSpec'
```
The results (cleaned up, no idea how to get a more compact output by default):
```
RuboCop::Cop::RSpec::Base#rspec_pattern (/Users/pirj/source/rubocop-rspec/lib/rubocop/cop/rspec/base.rb:47)
  samples:  25142 self (4.1%)  /   28918 total (4.8%)
RuboCop::Cop::RSpec::AnyInstance#disallowed_stub (/Users/pirj/source/rubocop-rspec/lib/rubocop/cop/rspec/any_instance.rb:28)
  samples:  1721 self (0.3%)  /   2268 total (0.4%)
RuboCop::Cop::RSpec::Base#relevant_rubocop_rspec_file? (/Users/pirj/source/rubocop-rspec/lib/rubocop/cop/rspec/base.rb:43)
  samples:  1555 self (0.3%)  /   30473 total (5.0%)
RuboCop::Cop::RSpec::ImplicitExpect#implicit_expect (/Users/pirj/source/rubocop-rspec/lib/rubocop/cop/rspec/implicit_expect.rb:33)
  samples:  1478 self (0.2%)  /   2480 total (0.4%)
RuboCop::Cop::RSpec::MessageSpies#message_expectation (/Users/pirj/source/rubocop-rspec/lib/rubocop/cop/rspec/message_spies.rb:38)
  samples:  1392 self (0.2%)  /   1758 total (0.3%)

[45 items skipped]

RuboCop::Cop::RSpec::MultipleMemoizedHelpers#on_block (/Users/pirj/source/rubocop-rspec/lib/rubocop/cop/rspec/multiple_memoized_helpers.rb:92)
  samples:   214 self (0.0%)  /   20335 total (3.4%)

[4 items skipped]

RuboCop::Cop::RSpec::MultipleMemoizedHelpers#all_helpers (/Users/pirj/source/rubocop-rspec/lib/rubocop/cop/rspec/multiple_memoized_helpers.rb:105)
  samples:   179 self (0.0%)  /   18647 total (3.1%)

[38 items skipped]

RuboCop::Cop::RSpec::MultipleMemoizedHelpers#allow_subject? (/Users/pirj/source/rubocop-rspec/lib/rubocop/cop/rspec/multiple_memoized_helpers.rb:135)
  samples:    51 self (0.0%)  /    161 total (0.0%)

[...]

  callees (18 total):
      18  (  100.0%)  RuboCop::Cop::Base.inherited
```

I'm not certain how to interpret this, is it the first or the second percentage that is important.

spec/smoke_tests/weird_rspec_spec.rb
```diff
@@ -11,6 +11,9 @@
   let (:foo) { 1 }
   let!   (:bar){}
 
+  bar = -> {}
+  let(:foo, &bar)
```
mockdeep, line 15: I love that this file exists. A really clever way to ensure that future rule changes all account for weird ruby syntax.

mockdeep:
@pirj thanks so much for taking the baton on this one. Looks great!

lib/rubocop/cop/rspec/multiple_memoized_helpers.rb
bquorning: First, we should determine if `MultipleMemoizedHelpers` is slower than the other cops. If its speed is average, there’s no need to dig further.

I ran the same command as yours, but on another spec suite. Next, I search the stackprof result for all `on_*` method calls on RSpec cops and order them by number of samples (20335 for on_block in your example above)

```
stackprof tmp/stackprof.dump --method 'RuboCop::Cop::RSpec.*#on_.*' |
  grep -B1 'samples.* total' |
  paste -d " "  - - - |
  sort -k8 -n |
  awk '{ print $8 " samples: " $1 }'
```
The final lines of my result is
```
485 samples: RuboCop::Cop::RSpec::RepeatedExample#on_block
517 samples: RuboCop::Cop::RSpec::RepeatedDescription#on_block
577 samples: RuboCop::Cop::RSpec::MultipleExpectations#on_block
844 samples: RuboCop::Cop::RSpec::SubjectStub#on_top_level_group
960 samples: RuboCop::Cop::RSpec::DescribedClass#on_block
1132 samples: RuboCop::Cop::RSpec::MultipleMemoizedHelpers#on_block
```
So it seems that `SubjectStub`, `DescribedClass` and `MultipleMemoizedHelpers` are significantly slower than our other cops. (Of course, configuration and the nature of the specs may play a role here)

pirj: Added a trivial memoization.
**Before**
713.80s user 42.25s system 80% cpu 15:42.32 total
**After**
663.91s user 41.03s system 89% cpu 13:10.78 total

**Before**
```
7407 samples: RuboCop::Cop::RSpec::MultipleSubjects#on_block
7969 samples: RuboCop::Cop::RSpec::LetSetup#on_block
7998 samples: RuboCop::Cop::RSpec::RepeatedExample#on_block
8527 samples: RuboCop::Cop::RSpec::RepeatedDescription#on_block
9783 samples: RuboCop::Cop::RSpec::MultipleExpectations#on_block
12220 samples: RuboCop::Cop::RSpec::SubjectStub#on_top_level_group
13026 samples: RuboCop::Cop::RSpec::DescribedClass#on_block
27289 samples: RuboCop::Cop::RSpec::MultipleMemoizedHelpers#on_block <===
```
**After**
```
6297 samples: RuboCop::Cop::RSpec::NamedSubject#on_block
6305 samples: RuboCop::Cop::RSpec::MultipleSubjects#on_block
6686 samples: RuboCop::Cop::RSpec::RepeatedExample#on_block
6846 samples: RuboCop::Cop::RSpec::RepeatedDescription#on_block
6939 samples: RuboCop::Cop::RSpec::LetSetup#on_block
7592 samples: RuboCop::Cop::RSpec::MultipleMemoizedHelpers#on_block <===
8599 samples: RuboCop::Cop::RSpec::MultipleExpectations#on_block
10258 samples: RuboCop::Cop::RSpec::SubjectStub#on_top_level_group
```

It doesn't look like those results are very consistent, as other cops are shuffled. But at least `MultipleMemoizedHelpers` isn't the by far the worst offender anymore.

Is this good enough, @bquorning ?

pirj:
Merging, since @Darhazer [informally approved earlier](https://github.com/rubocop-hq/rubocop-rspec/pull/863#pullrequestreview-460648182).

pirj:
@mockdeep Thank you!

lib/rubocop/cop/rspec/multiple_memoized_helpers.rb
bquorning: Yeah, sampling results aren’t always consistent. But moving from 27000 samples down to 7600 is a significant change for a relatively simple memoization. Thank you for taking the time to implement this.

bquorning:
👍 Thank you.

lib/rubocop/cop/rspec/multiple_memoized_helpers.rb
pirj: Happy to do so, especially when it's that trivial.

config/default.yml
AlexWayfer: What about `Enabled: pending`? This cop broke my builds.

pirj: @AlexWayfer We didn't formally agree to adhere to this policy yet, neither we're sure that it works for extensions.
We'll do our best to make it work and to introduce new cops between 2.0 and 3.0 as `pending`.
Meanwhile - please accept my apologies.

marcandre:
For the record, this single cop that is enabled by default generates 311 offenses on the main `rubocop` gem and 372 offenses on the `rubocop-ast` gem.

pirj:
@marcandre That reminds me of `RSpec/ContextWording` cop that generated ~11000 offences in a project I was working on. 😱 Those offences, plus hastily added "workarounds" were all fixed by a volunteered colleague of mine, @separatio.

For sure RuboCop's (and extensions') defaults don't fit all, and:

> Ruby is all about diversity and doing things the way that makes you happy. Whatever style preferences you have RuboCop is there for you. That’s our promises and our guarantee. Within the subjective limits of sanity that is.

`s/Ruby/RSpec/`

separatio:
Thanks for the summon @pirj. I haven't really fixed ALL of the offenses, but few were indeed left.

FYI, by default, Rubocop generates offenses in a new Rails 6 project.
Under 100 offenses, but there are still a few tens of offenses IIRC.

The convention over configuration mindset is hard to overwrite and actually CONFIGURE Rubocop, but keep in mind that Rubocop gives the convention choice with some guidelines, and not setting it.

pirj:
Off-topic: There's this https://github.com/toshimaru/rubocop-rails_config. Wondering if for newly-generated Rails app code with this config turned on there will be any offences.

PR #865, opened 2020-01-24
Fix `RSpec/EmptyExampleGroup` to flag example groups with examples in invalid scopes

mlarraz:
Fixes https://github.com/rubocop-hq/rubocop-rspec/issues/864.

Previously any example inside an example group would cause this Cop to pass, even if the example was inside an invalid scope.

Because RSpec will not execute any code inside an example group unless it has an example, there was no way to catch this error.

---

* [x] Added tests.

### DIFF
--- a/lib/rubocop/cop/rspec/empty_example_group.rb
+++ b/lib/rubocop/cop/rspec/empty_example_group.rb
@@ -69,6 +69,7 @@ module RuboCop
             #{Examples::ALL.block_pattern}
             #{ExampleGroups::ALL.block_pattern}
             #{Includes::ALL.send_pattern}
+            #{Includes::ALL.block_pattern}
             (send nil? #custom_include? ...)
           }
         PATTERN
--- a/lib/rubocop/cop/rspec/empty_example_group.rb
+++ b/lib/rubocop/cop/rspec/empty_example_group.rb
@@ -66,6 +66,8 @@ module RuboCop
             #{ExampleGroups::ALL.block_pattern}
             #{Includes::ALL.send_pattern}
             (send nil? #custom_include? ...)
+            (block (send _ :each) _ #contains_example?)
+            (begin <(block (send _ :each) _ #contains_example?) ...>)
           }
         PATTERN
 
--- a/lib/rubocop/cop/rspec/empty_example_group.rb
+++ b/lib/rubocop/cop/rspec/empty_example_group.rb
@@ -60,15 +60,26 @@ module RuboCop
       class EmptyExampleGroup < Base
         MSG = 'Empty example group detected.'
 
-        def_node_search :contains_example?, <<-PATTERN
+        def_node_matcher :contains_example?, <<-PATTERN
           {
-            #{(Examples::ALL + Includes::ALL).send_pattern}
-            (send _ #custom_include? ...)
+            #{Examples::ALL.block_pattern}
+            #{ExampleGroups::ALL.block_pattern}
+            #{Includes::ALL.send_pattern}
+            (send nil? #custom_include? ...)
           }
         PATTERN
 
         def on_block(node)
-          return unless example_group?(node) && !contains_example?(node)
+          return unless example_group?(node)
+
+          body =
+            if node.body&.begin_type?
+              node.body.each_child_node
+            else
+              [node.body].each
+            end
+
+          return if body.any?(&method(:contains_example?))
 
           add_offense(node.send_node)
         end
--- a/lib/rubocop/cop/rspec/empty_example_group.rb
+++ b/lib/rubocop/cop/rspec/empty_example_group.rb
@@ -60,30 +60,41 @@ module RuboCop
       class EmptyExampleGroup < Base
         MSG = 'Empty example group detected.'
 
-        def_node_matcher :contains_example?, <<-PATTERN
+        def_node_matcher :example_group_body, <<~PATTERN
+          (block #{ExampleGroups::ALL.send_pattern} args $_)
+        PATTERN
+
+        def_node_matcher :example_or_group_or_include?, <<~PATTERN
           {
             #{Examples::ALL.block_pattern}
             #{ExampleGroups::ALL.block_pattern}
             #{Includes::ALL.send_pattern}
             (send nil? #custom_include? ...)
-            (block (send _ :each) _ #contains_example?)
-            (begin <(block (send _ :each) _ #contains_example?) ...>)
+          }
+        PATTERN
+
+        def_node_matcher :examples_in_iterator?, <<~PATTERN
+          (block (send _ :each) _ #examples?)
+        PATTERN
+
+        def_node_matcher :examples_directly_or_in_iterator?, <<~PATTERN
+          {
+            #example_or_group_or_include?
+            #examples_in_iterator?
+          }
+        PATTERN
+
+        def_node_matcher :examples?, <<~PATTERN
+          {
+            #examples_directly_or_in_iterator?
+            (begin <#examples_directly_or_in_iterator? ...>)
           }
         PATTERN
 
         def on_block(node)
-          return unless example_group?(node)
-
-          body =
-            if node.body&.begin_type?
-              node.body.each_child_node
-            else
-              [node.body].each
-            end
-
-          return if body.any?(&method(:contains_example?))
-
-          add_offense(node.send_node)
+          example_group_body(node) do |body|
+            add_offense(node.send_node) unless examples?(body)
+          end
         end
 
         private
--- a/lib/rubocop/cop/rspec/empty_example_group.rb
+++ b/lib/rubocop/cop/rspec/empty_example_group.rb
@@ -74,30 +74,21 @@ module RuboCop
           }
         PATTERN
 
-        def_node_matcher :examples_in_iterator?, <<~PATTERN
-          (block
-            {
-              (send _ :each)
-              (send _ :each_with_object _)
-              (send _ :each_with_index)
-              (send (send _ :each) :with_object _)
-              (send (send _ :each) :with_index)
-            }
-            _ #examples?
-          )
+        def_node_matcher :examples_inside_block?, <<~PATTERN
+          (block !#{Hooks::ALL.send_pattern} _ #examples?)
         PATTERN
 
-        def_node_matcher :examples_directly_or_in_iterator?, <<~PATTERN
+        def_node_matcher :examples_directly_or_in_block?, <<~PATTERN
           {
             #example_or_group_or_include?
-            #examples_in_iterator?
+            #examples_inside_block?
           }
         PATTERN
 
         def_node_matcher :examples?, <<~PATTERN
           {
-            #examples_directly_or_in_iterator?
-            (begin <#examples_directly_or_in_iterator? ...>)
+            #examples_directly_or_in_block?
+            (begin <#examples_directly_or_in_block? ...>)
           }
         PATTERN
 
--- a/lib/rubocop/cop/rspec/empty_example_group.rb
+++ b/lib/rubocop/cop/rspec/empty_example_group.rb
@@ -74,7 +74,16 @@ module RuboCop
         PATTERN
 
         def_node_matcher :examples_in_iterator?, <<~PATTERN
-          (block (send _ :each) _ #examples?)
+          (block
+            {
+              (send _ :each)
+              (send _ :each_with_object _)
+              (send _ :each_with_index)
+              (send (send _ :each) :with_object _)
+              (send (send _ :each) :with_index)
+            }
+            _ #examples?
+          )
         PATTERN
 
         def_node_matcher :examples_directly_or_in_iterator?, <<~PATTERN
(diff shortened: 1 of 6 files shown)


lib/rubocop/cop/rspec/empty_example_group.rb
```diff
@@ -68,13 +68,20 @@ class EmptyExampleGroup < Cop
...
+            c && should_contain_example?(c) && contains_example?(c)
+          end
 
           add_offense(node.send_node)
         end
 
         private
 
+        def should_contain_example?(node)
+          !hook?(node) && !let?(node) && !subject?(node)
```
Darhazer: Instead of listing possible node types that may have examples in them, what about replacing the recursive search with check only on immediate children and passing on an example or example group (as @pirj suggested in the original issue)

pirj: Yep, that's the whole point, I've stepped into this so many times now - `def_node_search` is recursive, while `def_node_matcher` only matches direct children.

lib/rubocop/cop/rspec/empty_example_group.rb
```diff
@@ -68,13 +68,20 @@ class EmptyExampleGroup < Cop
         PATTERN
 
         def on_block(node)
-          return unless example_group?(node) && !contains_example?(node)
+          return unless example_group?(node)
+          return if node.children.any? do |c|
+            c && should_contain_example?(c) && contains_example?(c)
```
Darhazer: You can use `each_child_node` and you won't have non-parser nodes yield, so you can skip the `c &&` part

lib/rubocop/cop/rspec/empty_example_group.rb
mlarraz: This was the first thing I tried, but it caused a number of test failures. Pushing for reference

pirj:
This is quite interesting. Let me play around with it quickly.

pirj:
Side observation: `RuboCop::RSpec::ExampleGroup.new(node).examples` seems to count examples inside hooks. Hope there's no real need to use it here.

pirj:
I disagree with an example on [line 85](https://github.com/rubocop-hq/rubocop-rspec/pull/865/files#diff-9a93f375888ae8523835d1127abf853cR85), I don't think it's the business of this cop to flag example groups with nested example groups.

pirj:
Fixed a spec I don't agree with a bit above.

I've pushed the fix for the cop as well, but I find this kind of child selection is not good enough as the code does not reveal its intent.
Similar bits can be found in some cops e.g. `Yield`: `calling_block?`.
Maybe you'll find a pattern to extract or a way to make the code more expressive.

Leaving the 80-columns offence and CHANGELOG.md conflict up to you.

Looks good.

pirj:
You are facing an issue with this cop failing to detect examples defined in iterators in a spec that you didn't touch.
```ruby
expected_rewrites.each do |old, new|
    it %(rewrites "#{old}" as "#{new}") do
```

It's hard to disagree since there's this guideline https://rspec.rubystyle.guide/#it-in-iterators

On the one hand, it's not this cop's business, and on the other hand there's no other cop to enforce that.

Even though my hands are itching to rewrite that spec, I can imagine other users of RuboCop RSpec being hit by a wave of offences from this cop for defining examples in iterators, which is not really fair.

How do you feel about taking a stab at improving the cop so that it detects examples inside iterators? I can help out if you get stuck with this.

pirj:
@mlarraz Did you have a chance to look at it?

pirj:
@mlarraz Ping

pirj:
@mlarraz A gentle reminder.

mlarraz:
Apologies, I've been been swamped and haven't really had time to look into this more. Will try to take another look this weekend

pirj:
@mlarraz Poke

pirj:
Ping

pirj:
Github is super-slow when pushing changes to author's branch:
```
To github.com:mlarraz/rubocop-rspec.git
 + 3d0deae...e28300d validate_scope -> validate_scope (forced update)
```
Still not reflected [in the branch](https://github.com/mlarraz/rubocop-rspec/tree/validate_scope) or here in the pull request. :angel:/l/r/

I've added the detection of examples inside iterators, and such example groups are not flagged now.

Used a selective recursive node pattern. We need to make node pattern language Turing complete already.

lib/rubocop/cop/rspec/empty_example_group.rb
```diff
@@ -60,15 +60,28 @@ module RSpec
...
         def on_block(node)
-          return unless example_group?(node) && !contains_example?(node)
+          return unless example_group?(node)
+
+          body =
+            if node.body&.begin_type?
+              node.body.each_child_node
+            else
+              [node.body].each
+            end
```
pirj, line 82: I guess this can be simplified and implemented in node pattern, but it's way too late around here for me to crack it.
Neither `return if contains_example?(node)` nor `return if node.children.any?(&method(:contains_example?))` worked with the existing node pattern - some specs fail.

pirj:
Rebased. @bquorning @Darhazer please take a look.

lib/rubocop/cop/rspec/empty_example_group.rb
Darhazer: I guess the else could be just `[node.body]`
easier to understand, provides `any?`

lib/rubocop/cop/rspec/empty_example_group.rb
```diff
@@ -60,15 +60,28 @@ module RSpec
...
-        def_node_search :contains_example?, <<-PATTERN
+        def_node_matcher :contains_example?, <<-PATTERN
           {
-            #{(Examples::ALL + Includes::ALL).send_pattern}
-            (send _ #custom_include? ...)
+            #{Examples::ALL.block_pattern}
+            #{ExampleGroups::ALL.block_pattern}
+            #{Includes::ALL.send_pattern}
+            (send nil? #custom_include? ...)
+            (block (send _ :each) _ #contains_example?)
```
Darhazer, line 69: So this is a recursive matcher?

Darhazer: `:each` is not the only way to define examples in an iterator (although it might be the most common), so I'm not sure why we should explicitly list it?

pirj: > :each is not the only way to define examples in an iterator

However, if we will relax this, the following becomes possible:
```ruby
RSpec.describe Foo do
  before do
    it { is_expected.to be_valid }
  end
end
```

Do you have some other iterator methods that can be legitimately used instead of `each` to define several examples?
I can think of:
```ruby
[:a, :b].each.with_index do |o, i|
  it "#{o} #{i}" do
    # ...
  end
end
```
and similarly `each.with_object`.

I have little to no idea why would anyone use that though.

lib/rubocop/cop/rspec/empty_example_group.rb
```diff
@@ -60,15 +60,28 @@ module RSpec
       class EmptyExampleGroup < Cop
         MSG = 'Empty example group detected.'
 
-        def_node_search :contains_example?, <<-PATTERN
+        def_node_matcher :contains_example?, <<-PATTERN
```
pirj, line 63: @Darhazer let me quickly refactor this into smaller matchers.

lib/rubocop/cop/rspec/empty_example_group.rb
pirj: Added support for `each.with_object`/`each_with_object`/`each.with_index`/`each_with_index`.

pirj:
@Darhazer @bquorning Naming node matchers was the hardest part, did my best, but open to better suggestions.

lib/rubocop/cop/rspec/empty_example_group.rb
```diff
@@ -60,17 +60,50 @@ module RSpec
       class EmptyExampleGroup < Cop
         MSG = 'Empty example group detected.'
 
-        def_node_search :contains_example?, <<-PATTERN
+        def_node_matcher :example_group_body, <<~PATTERN
+          (block #{ExampleGroups::ALL.send_pattern} args $_)
+        PATTERN
```
pirj, line 65: Other cops' specs break if I change `example_group?` to yield its body.
I decided to copy this pattern over here rather than defining it as
```
$(block #{ExampleGroups::ALL.send_pattern} args $_)
```
for every cop that would ignore the second argument. But it might be possible, and I don't have a preference here.

lib/rubocop/rspec/language/node_pattern.rb
```diff
@@ -14,7 +14,7 @@ module NodePattern
         def_node_matcher :spec_group?, spec_groups.block_pattern
 
         def_node_matcher :example_group_with_body?, <<-PATTERN
-          (block #{ExampleGroups::ALL.send_pattern} args [!nil?])
+          (block #{ExampleGroups::ALL.send_pattern} args !nil?)
```
pirj, line 17: Just a random observation that logical and was redundant.

spec/rubocop/cop/rspec/empty_example_group_spec.rb
```diff
@@ -1,8 +1,8 @@
 # frozen_string_literal: true
 
 RSpec.describe RuboCop::Cop::RSpec::EmptyExampleGroup, :config do
-  it 'flags an empty context' do
-    expect_offense(<<-RUBY)
+  it 'flags an empty example group' do
+    expect_offense(<<~RUBY)
```
pirj, line 5: Unrelated change, I can roll this back and fix it for all in a separate pull request if you like.
The failure spec output looks slightly better with `<<~`.

spec/rubocop/cop/rspec/empty_example_group_spec.rb
```diff
@@ -22,15 +22,114 @@
...
+        [1, 2, 3].each do |page|
+          no_examples_here
+          and_no_ads_either
+        end
+        more_surroundings
+      end
+    RUBY
+  end
+
+  it 'ignores example group with examples defined in an obscure iterators' do
```
pirj, line 85: Decided not to add this to cop's documentation to avoid proliferating bad practices.

lib/rubocop/cop/rspec/empty_example_group.rb
pirj: Very good point indeed.

I don't have an idea what this list should include though.
Hooks are kind of obvious. With hook aliases configuration (e.g. `setup do`, `tearup do` etc) it should become easier to detect examples that are defined in improper blocks, and that shouldn't be counted towards the number of examples in an example group.
Is that it?

I remember on one of my previous projects we've used some custom block that was wrapping examples:
```ruby
RSpec.describe SomeController do
  with_permission(:read) do
    it { ... }
  end
end
```

With this change, the root example group would be detected as an empty example group.
Even though I don't personally agree it's the best approach to the problem, e.g.:
```ruby
RSpec.describe SomeController do
  context permission: :read do
    it { ... }
  end
end
```
it doesn't mean that it's an illegitimate usage.

pirj:
I've just realized the following:
```
          context "when a third thing is true" do
       +  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ Empty example group detected.
            it_behaves_like "some thingy" do
              let(:a) { 'foo' }
            end
```
(but it works ok with a `blockpass`).

lib/rubocop/cop/rspec/empty_example_group.rb
pirj: Let me quickly limit it to hooks, as I think it's the major issue.
We'll see if someone notices this is insufficient.

pirj: Done.

pirj:
Please take another look @bquorning @Darhazer

pirj:
Minor touches to specs: removed a duplicate spec, changed "does not flag" to "ignores".

pirj:
Ping.

pirj:
Alarming ping @bquorning @Darhazer

lib/rubocop/cop/rspec/empty_example_group.rb
```diff
@@ -60,17 +60,42 @@ module RSpec
...
+            #examples_inside_block?
+          }
+        PATTERN
+
+        def_node_matcher :examples?, <<~PATTERN
+          {
+            #examples_directly_or_in_block?
+            (begin <#examples_directly_or_in_block? ...>)
+          }
+        PATTERN
```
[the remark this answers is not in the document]
Darhazer: Yup, some documentation would be nice

lib/rubocop/cop/rspec/empty_example_group.rb
```diff
@@ -60,17 +60,101 @@ module RSpec
...
+        #
+        #   @example source that matches
+        #     it { is_expected.to fly }
+        #     describe('non-empty example groups too') { }
+        #     it_behaves_like 'an animal'
+        #     it_behaves_like('a cat') { let(:food) { 'milk' } }
+        #     it_has_root_access
+        #
+        #   @param node [RuboCop::AST::Node]
+        #   @return [Array<RuboCop::AST::Node>] matching nodes
```
pirj, line 88: @Darhazer docs added

pirj:
@bquorning ping

PR #869, opened 2020-02-06
Fix `FilePath` detection when absolute path includes test subject 

eitoball:
This PR should fix `FilePath` cop when test subject is accidentally included in absolute path of file that being examined. For example, if file contains following spec,

```ruby
RSpec.describe Foo do; end
```

and the absolute path name is like `/home/foo/src/spec/models/bar_spec.rb`.

I think this case should be warned, but currently it is passed because glob pattern `foo*_spec.rb` matches.

This PR changes to use relative path from project root so that it wouldn't match this case.

---

* [x] Added tests.

### DIFF
--- a/lib/rubocop/cop/rspec/file_path.rb
+++ b/lib/rubocop/cop/rspec/file_path.rb
@@ -103,7 +103,9 @@ module RuboCop
         end
 
         def filename_ends_with?(glob)
-          File.fnmatch?("*#{glob}", processed_source.buffer.name)
+          filename =
+            RuboCop::PathUtil.relative_path(processed_source.buffer.name)
+          File.fnmatch?("*#{glob}", filename)
         end
 
         def relevant_rubocop_rspec_file?(_file)
--- a/spec/rubocop/cop/rspec/file_path_spec.rb
+++ b/spec/rubocop/cop/rspec/file_path_spec.rb
@@ -172,6 +172,15 @@ RSpec.describe RuboCop::Cop::RSpec::FilePath, :config do
     RUBY
   end
 
+  it 'uses relative path' do
+    allow(RuboCop::PathUtil)
+      .to receive(:relative_path).and_return('spec/models/bar_spec.rb')
+    expect_offense(<<-RUBY, '/home/foo/spec/models/bar_spec.rb')
+      describe Foo do; end
+      ^^^^^^^^^^^^ Spec path should end with `foo*_spec.rb`.
+    RUBY
+  end
+
   context 'when configured with CustomTransform' do
     let(:cop_config) { { 'CustomTransform' => { 'FooFoo' => 'foofoo' } } }
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Master (Unreleased)
 
+* Fix `RSpec/FilePath` detection when absolute path includes test subject. ([@eitoball][])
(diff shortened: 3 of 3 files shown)


eitoball:
> Will you please add a spec for this change?

Added. But, it's quite dirty...

spec/rubocop/cop/rspec/file_path_spec.rb
```diff
@@ -172,6 +172,15 @@
     RUBY
   end
 
+  it 'uses relative path' do
+    allow(Dir).to receive(:pwd).and_return('/home/foo')
+    allow(File).to receive(:absolute_path).and_return('/')
```
pirj: It's quite ok, but what do you think of stubbing `RuboCop::PathUtil.relative_path` directly? It's implementation-specific, but why not?

lib/rubocop/cop/rspec/file_path.rb
```diff
@@ -103,7 +103,9 @@ def ignore_methods?
         end
 
         def filename_ends_with?(glob)
-          File.fnmatch?("*#{glob}", processed_source.buffer.name)
+          filename =
+            RuboCop::PathUtil.relative_path(processed_source.buffer.name)
```
pirj: It's possible to call `rubocop` from a subdirectory, e.g. `/home/foo/spec/models/`, in this case relative path will be `./bar_spec.rb`, this shouldn't cause any thouble, should it?

spec/rubocop/cop/rspec/file_path_spec.rb
eitoball: True. I changed to stub `RuboCop::PathUtil.relative_path` instead.

lib/rubocop/cop/rspec/file_path.rb
eitoball: I think that it depends. If subject is like 'Foo::Bar', it wouldn't detect. I think that most people run from project root

pirj:
@eitoball Can you please rebase?

@bquorning @Darhazer @lazycoder9 What are your thoughts on this?

eitoball:
@pirj Rebased.

lib/rubocop/cop/rspec/file_path.rb
Darhazer: Seems this way it doesn't work if I run rubocop and provide a path to check, e.g.
`rubocop --require rubocop-rspec ~/my_project/specs`

pirj: @eitoball Can you please check this edge case?

eitoball: Oh, I didn't notice. Will do over this weekend. Thank you.

pirj: A friendly reminder. This is ready to be merged apart from this edge case.

pirj:
Thanks a lot for your contribution!
If there's something left to be done in terms of running `rubocop` from a nested folder, please feel free to send a follow-up. No obligations, no rush.

lib/rubocop/cop/rspec/file_path.rb
rolfschmidt: Hi guys,
i think i had this edge case. I tried to create a pull request for it:

https://github.com/rubocop-hq/rubocop-rspec/pull/917

PR #872, opened 2020-02-08
Support (multiple) spec file extensions other than '_spec.rb' 

cwilbur:
Rubocop::Rspec tries to validate that spec file names are sensibly named.  

Unfortunately, the only extension that Rubocop::Rspec recognizes for spec files is `_spec.rb`.  I've been using `.spec.rb` for a while, and it irked me that there was no configuration option to allow alternate extensions.  So I created one.

I altered the cop `RSpec/FilePath`, adding a configuration field `SpecFileExtensions` that should contain an array of strings. These are treated as literal strings, with file-glob metacharacters escaped, and joined using the `{opt_one,opt_two,opt_three}` syntax for file globs.

I kept the default behavior as it was before; using `.spec.rb` is something I strongly prefer but not something I necessarily want to impose upon thousands of others.  The behavior will only change if a value is provided for `SpecFileExtensions`.

* [X] Added tests.
* [X] Added an entry to the [changelog](https://github.com/rubocop-hq/rubocop-rspec/blob/master/CHANGELOG.md) if the new code introduces user-observable changes.

### DIFF
--- a/lib/rubocop/cop/rspec/file_path.rb
+++ b/lib/rubocop/cop/rspec/file_path.rb
@@ -41,6 +41,51 @@ module RuboCop
       #   # good
       #   my_class_spec.rb         # describe MyClass, '#method'
       #
+      # With the configuration option `SpecFileExtensions`
+      # alternate extensions for spec files may be specified
+      # as a list of strings.  The default is the
+      # formerly-hardcoded `[ '_spec.rb' ]`, but any string
+      # can be used as an extension.
+      #
+      # @example when `SpecFileExtensions` contains `.spec.rb`
+      #
+      #   # bad
+      #   my_class_spec.rb         # describe MyClass
+      #
+      #   # good
+      #   my_class.spec.rb         # describe MyClass
+      #
+      #   # good
+      #   my_class_method.spec.rb  # describe MyClass, '#method'
+      #
+      #   # good
+      #   my_class/method.spec.rb  # describe MyClass, '#method'
+      #
+      # @example when `SpecFileExtensions` contains  `[ '.spec.rb', _spec.rb' ]`
+      #
+      #   # bad
+      #   my_class_test.rb         # describe MyClass
+      #
+      #   # bad
+      #   my_class.test.rb         # describe MyClass
+      #
+      #   # good
+      #   my_class_spec.rb         # describe MyClass
+      #
+      #   # good
+      #   my_class.spec.rb         # describe MyClass
+      #
+      #   # good
+      #   my_class_method_spec.rb  # describe MyClass, '#method'
+      #
+      #   # good
+      #   my_class_method.spec.rb  # describe MyClass, '#method'
+      #
+      #   # good
+      #   my_class/method_spec.rb  # describe MyClass, '#method'
+      #
+      #   # good
+      #   my_class/method.spec.rb  # describe MyClass, '#method'
       class FilePath < Cop
         include RuboCop::RSpec::TopLevelDescribe
 
@@ -70,7 +115,9 @@ module RuboCop
         end
 
         def glob_for((described_class, method_name))
-          "#{expected_path(described_class)}#{name_glob(method_name)}*_spec.rb"
+          "#{expected_path(described_class)}" \
+          "#{name_glob(method_name)}" \
+          "#{extension_glob}"
         end
 
         def name_glob(name)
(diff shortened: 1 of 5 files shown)


lib/rubocop/cop/rspec/file_path.rb
```diff
@@ -102,8 +149,30 @@ def ignore_methods?
...
+            '*' + escape_glob(file_extensions.first)
+          else
+            "*{#{file_extensions.map(&method(:escape_glob)).join(',')}}"
+          end
+        end
+
         def filename_ends_with?(glob)
-          File.fnmatch?("*#{glob}", processed_source.buffer.name)
+          flags = file_extensions.count > 1 ? [File::FNM_EXTGLOB] : []
+          File.fnmatch?("*#{glob}", processed_source.buffer.name, *flags)
```
pirj: There's [a recent pull request](https://github.com/rubocop-hq/rubocop-rspec/pull/869/files#diff-7b5894acb0fe1766bf72709ffae459a6R106), not sure which one gets merged first.
Do you think changes will be easy to merge?

cwilbur: TL, DR: I think both your suggestions are good.  If you like the overall idea and think it might be generally useful, I'll improve it in accordance with your suggestions and either re-awaken this pull request or submit another one.
---

The existing configuration doesn't quite solve the problem. Setting the pattern in the RSpec config tells RSpec which files contain specs. Setting AllCops/RSpec/Patterns tells Rubocop::RSpec which Ruby files are more properly RSpec files. But when you get to the RSpec/FilePath cop, the only acceptable extension is a hardcoded `_spec.rb`. In fact, until you've set the RSpec config pattern and the Rubocop::RSpec config pattern to something that doesn't match `_spec.rb`, there's nothing about the extension to upset RSpec/FilePath.

I originally considered this as a band-aid patch for something that irked me. The original logic was to take the class structure of the code under test and use it to construct a glob that should match the file path of the spec file. So I stayed consistent with that logic, but altered the construction of the glob to match multiple extensions.

The question RSpec/FilePath tries to answer is this: given the class structure of the code under test, are the pathname and filename of the spec file consistent with Ruby's and RSpec's conventions?  In this context, I think it's best separated into two questions: are the pathname and filename consistent with the classes tested in the spec file?  and is the extension on the spec file one that RSpec, Rubocop::RSpec, and programmers will recognize as a spec file?

Dividing this into two rubrics is probably a better solution than doing elaborate gymnastics with strings and escaping and glob patterns to create a more complex (unmaintainable mess of a) glob pattern (that looks like bad Perl I wrote in my younger days) to answer both questions at the same time.  Pulling the acceptable file extensions from the RSpec and Rubocop config files is just a bonus.

And with all that said: since the best approach is likely to revert to the old behavior, but removing extensions altogether from the glob pattern, and add a second test for just the extension, the two pull requests are not likely to conflict. They're not in conflict logically: my change says "if there are two or more acceptable filename extensions, we need to pass `File::FNM_EXTGLOB` as a flag to `File.fnmatch?`" while the other pull request says "we need to handle the case when the path might be weird because we invoked rubocop from somewhere other than the project root, so we need to munge it before we pass it to `File.fnmatch?`. This is the sort of thing that casues easy merge conflicts because they both operate in fiddly ways on the same line of code, but they're easy for a human to resolve.

So.  I will revise the patch to make it better, and we can consider the new improved version  in a few days.

pirj: It's hard to disagree with your point that this cop does the job of several cops. In addition to matching file name and making sure extension is proper, I can think of additional duties that it could do, but that would really make it a god cop:
 - make sure files containing root `shared_examples` and `shared_context` have the right path names
 - make sure they have a proper extension that **doesn't** match RSpec's pattern (also reflected in `AllCops/RSpec/Pattern` configuration item.

Any kind of improvement on this front is appreciated, if you could extract the part checking just the extension from `FilePath` without compromising its current responsibilities, that would be just great.

pirj:
@cwilbur Do you think it would be possible to reuse [this existing option](https://github.com/rubocop-hq/rubocop-rspec#inspecting-files-that-dont-end-with-_specrb) and avoid adding a new cop-specific one?

pirj:
@cwilbur Would you like to move this forward?

pirj:
@cwilbur A friendly reminder not to let this slip through the cracks.
Did you solve your issue in some other way?

pirj:
@cwilbur Please feel free to reopen (or let us know if you want to) if you intend to proceed on this change.

PR #874, opened 2020-02-09
Take metadata into account in `RSpec/RepeatedDescription` cop

### LINKED ISSUE #873: RSpec/RepeatedDescription take metadata into account
According to this [comment](https://github.com/rubocop-hq/rubocop-rspec/pull/860#discussion_r367659300) from @pirj, there is an issue with `RSpec/RepeatedDescription` cop. This issue does not take metadata of examples into account and registers offense for examples like
```ruby
it 'do something' do
  # ...
end

it 'do something', :flag do
  # ...
end
```

And examples above should not be flagged

lazycoder9:
Close #873 
`RSpec/RepeatedDescription` will take metadata into account and will not flag examples like below
```ruby
it 'is valid' 
    # ...
end

it 'is valid', :flag do
    # ...
end
```

---

* [x] Added tests.

### DIFF
--- a/lib/rubocop/cop/rspec/repeated_description.rb
+++ b/lib/rubocop/cop/rspec/repeated_description.rb
@@ -29,6 +29,17 @@ module RuboCop
       #       end
       #     end
       #
+      #     # good
+      #     RSpec.describe User do
+      #       it 'is valid' do
+      #         # ...
+      #       end
+      #
+      #       it 'is valid', :flag do
+      #         # ...
+      #       end
+      #     end
+      #
       class RepeatedDescription < Cop
         MSG = "Don't repeat descriptions within an example group."
 
@@ -47,13 +58,17 @@ module RuboCop
           grouped_examples =
             RuboCop::RSpec::ExampleGroup.new(node)
               .examples
-              .group_by(&:doc_string)
+              .group_by { |example| example_signature(example) }
 
           grouped_examples
-            .select { |description, group| description && group.size > 1 }
+            .select { |signatures, group| signatures.any? && group.size > 1 }
             .values
             .flatten
             .map(&:definition)
+        end
+
+        def example_signature(example)
+          [example.metadata, example.doc_string]
         end
       end
     end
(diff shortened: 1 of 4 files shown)


lib/rubocop/cop/rspec/repeated_description.rb
```diff
@@ -47,14 +58,18 @@ def repeated_descriptions(node)
           grouped_examples =
             RuboCop::RSpec::ExampleGroup.new(node)
               .examples
-              .group_by(&:doc_string)
+              .group_by { |example| example_signature(example) }
 
           grouped_examples
-            .select { |description, group| description && group.size > 1 }
+            .select { |signatures, group| !signatures.empty? && group.size > 1 }
```
pirj: Maybe `signatures.any?`?
In this case `compact` below becomes redundant.

pirj:
@bquorning @Darhazer we have plenty of changes to release!

spec/rubocop/cop/rspec/repeated_description_spec.rb
```diff
@@ -73,4 +73,18 @@
       end
     RUBY
   end
+
+  it 'does not flag examples if metadata is different' do
```
Darhazer: what if there is the same metatada? with same description? With different descriptions?
I would convert this to a context and add a couple of more examples

Darhazer:
> @bquorning @Darhazer we have plenty of changes to release!

Unfortunately, I don't have access to rubygems to push releases. But I agree, let's release a new version after this one is merged

bquorning:
I’ll release a new version after this PR has been merged. As mentioned by @Darhazer, we should probably have a few more specs first.

lib/rubocop/cop/rspec/repeated_description.rb
lazycoder9: Fixed

spec/rubocop/cop/rspec/repeated_description_spec.rb
lazycoder9: Added two more tests

PR #877, opened 2020-02-11
RepeatedDescription detects not only string descriptions

lazycoder9:
Added to `RSpec/RepeatedDescription` cop ability to detect descriptions with interpolation and methods

---

* [x] Added tests.

### DIFF
--- a/lib/rubocop/rspec/example.rb
+++ b/lib/rubocop/rspec/example.rb
@@ -4,7 +4,7 @@ module RuboCop
   module RSpec
     # Wrapper for RSpec examples
     class Example < Concept
-      def_node_matcher :extract_doc_string,     '(send _ _ $str ...)'
+      def_node_matcher :extract_doc_string,     '(send _ _ $_ ...)'
       def_node_matcher :extract_metadata,       '(send _ _ _ $...)'
       def_node_matcher :extract_implementation, '(block send args $_)'
 
--- a/spec/rubocop/cop/rspec/repeated_description_spec.rb
+++ b/spec/rubocop/cop/rspec/repeated_description_spec.rb
@@ -117,4 +117,48 @@ RSpec.describe RuboCop::Cop::RSpec::RepeatedDescription do
       end
     RUBY
   end
+
+  it 'does not flag descriptions with different interpolated variables' do
+    expect_no_offenses(<<-RUBY)
+      describe 'doing x' do
+        it "does \#{x}" do
+        end
+
+        it "does \#{y}" do
+        end
+      end
+    RUBY
+  end
+
+  it 'registers offense for repeated description in same iterator' do
+    expect_offense(<<-RUBY)
+      describe 'doing x' do
+        %i[foo bar].each do |type|
+          it "does a thing \#{type}" do
+          ^^^^^^^^^^^^^^^^^^^^^^^^^ Don't repeat descriptions within an example group.
+          end
+
+          it "does a thing \#{type}" do
+          ^^^^^^^^^^^^^^^^^^^^^^^^^ Don't repeat descriptions within an example group.
+          end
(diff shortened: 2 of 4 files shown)


lib/rubocop/rspec/example.rb
```diff
@@ -4,7 +4,7 @@ module RuboCop
   module RSpec
     # Wrapper for RSpec examples
     class Example < Concept
-      def_node_matcher :extract_doc_string,     '(send _ _ $str ...)'
+      def_node_matcher :extract_doc_string,     '(send _ _ $_ ...)'
```
bquorning: I wonder if we should add a couple of examples to example_spec.rb

lazycoder9: @bquorning agree, fixed

spec/rubocop/cop/rspec/repeated_description_spec.rb
```diff
@@ -117,4 +117,48 @@
...
+    expect_offense(<<-RUBY)
+      describe 'doing x' do
+        %i[foo bar].each do |type|
+          it "does a thing \#{type}" do
+          ^^^^^^^^^^^^^^^^^^^^^^^^^ Don't repeat descriptions within an example group.
+          end
+
+          it "does a thing \#{type}" do
+          ^^^^^^^^^^^^^^^^^^^^^^^^^ Don't repeat descriptions within an example group.
+          end
```
pirj: Do you mind to add an example of two separate iterators in the same example group? And maybe a balancing example showing that those having different docstrings are ignored. Not certain if it gets the reader an idea of how it works.

I already see people fooling the cop by naming variables differently, like `type1` and `type2` :D

spec/rubocop/cop/rspec/repeated_description_spec.rb
```diff
@@ -117,4 +117,48 @@
...
+        ^^^^^^^^^^^^^^^ Don't repeat descriptions within an example group.
+          # ...
+        end
+
+        it(description) do
+        ^^^^^^^^^^^^^^^ Don't repeat descriptions within an example group.
+          # ...
+        end
+      end
+    RUBY
```
pirj: Do you think that adding a balancing example with different variables being used as examples' docstrings is ignored?

spec/rubocop/rspec/example_spec.rb
```diff
@@ -17,6 +17,16 @@ def example(source)
       .to eq(s(:str, 'does x'))
   end
 
+  it 'extracts interpolated doc string' do
```
pirj: I've rarely seen it in the wild, but sometimes a Symbol is passed as the first argument, and RSpec considers it a docstring, not metadata, at least as far as I remember.
It feels that that node pattern change covers `sym` pretty much, not sure if an additional example would prevent turning it into something like `${str dstr}` "for optimization".

bquorning: You mean e.g. `it :focus { expect(foo).to eq(42) }`?

bquorning: But if method invocations are allowed (e.g. `it(description)`) we’d probably need `${str dstr send}`.

pirj: No, I mean `_` is good, but just in case it's good to have a spec that fails if this gets changed to `${str dstr send}` and symbol would still be considered a valid description if passed as a first argument.

Side note: I remember there's also an `xstr`, but can't remember what it stands for.

Darhazer: `xstr` is for when a string is enclosed in `

bquorning:
Merging now to get the bugfix released.

@pirj I’ll address your comments in a separate PR.

PR #886, opened 2020-03-24
Add Capybara/VisibilityMatcher cop

### LINKED ISSUE #885: Cop idea: Don't allow `visible: true` or `visible: false` for capybara matchers
When trying to match a selector, you can use the [`visible:` finder option](https://www.rubydoc.info/gems/capybara/Capybara%2FNode%2FFinders:all). However this might have unexpected results, for example `visible: false` does not look for invisible elements, but visible **and** invisible elements.

```Ruby
expect(page).to have_selector('.my_element')
do_something_that_should_hide_my_element
# This would succeed, even if `.my_element` is still visible.
expect(page).to have_selector('.my_element', visible: false)
```

```Ruby
expect(page).to have_selector('.my_element')
do_something_that_should_hide_my_element
# This would fail if `.my_element` is still visible.
expect(page).to have_selector('.my_element', visible: :hidden)
```

Happy to discuss :)

aried3r:
Fixes #885.

This is a work in progress still, since it's my first cop.

---

* [x] Added tests.

### DIFF
--- a/config/default.yml
+++ b/config/default.yml
@@ -475,6 +475,12 @@ Capybara/FeatureMethods:
   EnabledMethods: []
   StyleGuide: https://www.rubydoc.info/gems/rubocop-rspec/RuboCop/Cop/RSpec/Capybara/FeatureMethods
 
+Capybara/VisibilityMatcher:
+  Description: Checks for boolean visibility in capybara finders.
+  Enabled: true
+  VersionAdded: '1.39'
+  StyleGuide: https://www.rubydoc.info/gems/rubocop-rspec/RuboCop/Cop/RSpec/Capybara/VisibilityMatcher
+
 FactoryBot/AttributeDefinedStatically:
   Description: Always declare attribute values as blocks.
   Enabled: true
--- a/lib/rubocop/cop/rspec/capybara/visibility_matcher.rb
+++ b/lib/rubocop/cop/rspec/capybara/visibility_matcher.rb
@@ -1,0 +1,54 @@
+# frozen_string_literal: true
+
+module RuboCop
+  module Cop
+    module RSpec
+      module Capybara
+        # Checks for boolean visibility in capybara finders.
+        #
+        # Capybara lets you find elements that match a certain visibility using
+        # the `:visible` option. `:visible` accepts both boolean and symbols as
+        # values, however using booleans can have unwanted effects. `visible:
+        # false` does not find just invisible elements, but both visible and
+        # invisible elements. For expressiveness and clarity, use one of the
+        # symbol values, `:all`, `:hidden` or `:visible`.
+        # (https://www.rubydoc.info/gems/capybara/Capybara%2FNode%2FFinders:all)
+        #
+        # @example
+        #
+        #   # bad
+        #   expect(page).to have_selector('.foo', visible: false)
+        #
+        #   # bad
+        #   expect(page).to have_selector('.foo', visible: true)
+        #
+        #   # good
+        #   expect(page).to have_selector('.foo', visible: :all)
+        #
+        #   # good
+        #   expect(page).to have_selector('.foo', visible: :hidden)
+        #
+        #   # good
+        #   expect(page).to have_selector('.foo', visible: :visible)
+        #
+        class VisibilityMatcher < Cop
+          MSG_FALSE = 'Use `:all` or `:hidden` instead of `false`.'
+          MSG_TRUE = 'Use `:visible` instead of `true`.'
+
+          def_node_matcher :visible_true?, <<~PATTERN
+            (send nil? :have_selector ... (hash <$(pair (sym :visible) true) ...>))
+          PATTERN
+
+          def_node_matcher :visible_false?, <<~PATTERN
+            (send nil? :have_selector ... (hash <$(pair (sym :visible) false) ...>))
+          PATTERN
+
+          def on_send(node)
+            visible_false?(node) { |arg| add_offense(arg, message: MSG_FALSE) }
+            visible_true?(node) { |arg| add_offense(arg, message: MSG_TRUE) }
+          end
+        end
+      end
+    end
+  end
+end
--- a/spec/rubocop/cop/rspec/capybara/visibility_matcher_spec.rb
+++ b/spec/rubocop/cop/rspec/capybara/visibility_matcher_spec.rb
@@ -1,0 +1,69 @@
+# frozen_string_literal: true
+
+RSpec.describe RuboCop::Cop::RSpec::Capybara::VisibilityMatcher do
+  subject(:cop) { described_class.new }
+
+  it 'registers an offense when using `visible: true`' do
+    expect_offense(<<-RUBY)
+      expect(page).to have_selector('.my_element', visible: true)
+                                                   ^^^^^^^^^^^^^ Use `:visible` instead of `true`.
+    RUBY
+  end
+
+  it 'registers an offense when using `visible: false`' do
+    expect_offense(<<-RUBY)
+      expect(page).to have_selector('.my_element', visible: false)
+                                                   ^^^^^^^^^^^^^^ Use `:all` or `:hidden` instead of `false`.
+    RUBY
+  end
+
+  it 'registers an offense when using a selector`' do
+    expect_offense(<<-RUBY)
+      expect(page).to have_selector(:css, '.my_element', visible: false)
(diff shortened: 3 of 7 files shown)


config/default.yml
```diff
@@ -475,6 +475,12 @@ Capybara/FeatureMethods:
   EnabledMethods: []
   StyleGuide: https://www.rubydoc.info/gems/rubocop-rspec/RuboCop/Cop/RSpec/Capybara/FeatureMethods
 
+Capybara/VisibilityMatcher:
+  Description: Checks for boolean visibility in capybara finders.
+  Enabled: true
+  VersionAdded: '0.81'
```
aried3r: No other cop seems to have it, should this be removed or use the rubocop-rspec version rather than the rubocop version number?

lib/rubocop/cop/rspec/capybara/visibility_matcher.rb
```diff
@@ -0,0 +1,50 @@
...
+        #   expect(page).to have_selector('.foo', visible: :all)
+        #
+        #   # good
+        #   expect(page).to have_selector('.foo', visible: :hidden)
+        #
+        #   # good
+        #   expect(page).to have_selector('.foo', visible: :visible)
+        #
+        class VisibilityMatcher < Cop
+          MSG = 'Use `:all`, `:hidden`, `:visible` instead of `true`, `false`.'
```
aried3r: Ideally I'd want `visible: true` to have a message "Use `:visible` instead of `true`" and the `visible: false` to have a message `Use `:all` or `:hidden` instead of `false`", but I don't know how to.

lib/rubocop/cop/rspec/capybara/visibility_matcher.rb
```diff
@@ -0,0 +1,50 @@
...
+
+          def_node_matcher :bad_method?, <<~PATTERN
```
aried3r: Should `bad_method?` be renamed? If so, to what?

lib/rubocop/cop/rspec/capybara/visibility_matcher.rb
```diff
@@ -0,0 +1,50 @@
...
+            (send nil? :have_selector ... (hash <(pair (sym :visible) {true false}) ...>))
```
aried3r: `:have_selector ...` should probably be something else, since the first, optional, parameter seems can be a symbol.
https://www.rubydoc.info/github/jnicklas/capybara/Capybara%2FNode%2FMatchers:has_selector%3F

This, if I understood correctly will match any amount of parameters before the hash, correct?

config/default.yml
pirj: We have a ticket and a pull request for adding this for all the cops https://github.com/rubocop-hq/rubocop-rspec/issues/699 https://github.com/rubocop-hq/rubocop-rspec/pull/721
It's safe to say it's `1.39` here, next `rubocop-rspec` minor release.

lib/rubocop/cop/rspec/capybara/visibility_matcher.rb
```diff
@@ -0,0 +1,50 @@
+# frozen_string_literal: true
+
+module RuboCop
+  module Cop
+    module RSpec
+      module Capybara
+        # Checks for boolean visibility in capybara finders.
+        #
+        # Capybara lets you find elements that match a certain visibiity using
```
pirj: Typo in "visibiity"

lib/rubocop/cop/rspec/capybara/visibility_matcher.rb
pirj: `visible_true?`/`visible_false?` should be sufficiently good.

lib/rubocop/cop/rspec/capybara/visibility_matcher.rb
pirj: It can be done by using two separate matchers and messages, you can check https://github.com/rubocop-hq/rubocop-rspec/blob/master/lib/rubocop/cop/rspec/message_spies.rb#L30 for an example.

lib/rubocop/cop/rspec/capybara/visibility_matcher.rb
pirj: [That's right](https://github.com/rubocop-hq/rubocop/blob/master/manual/node_pattern.md#-for-several-subsequent-nodes).

spec/rubocop/cop/rspec/capybara/visibility_matcher_spec.rb
```diff
@@ -0,0 +1,65 @@
+# frozen_string_literal: true
+
+RSpec.describe RuboCop::Cop::RSpec::Capybara::VisibilityMatcher do
+  subject(:cop) { described_class.new(config) }
```
pirj: `subject(:cop) { described_class.new }` is enough, since the cop is not configurable.

lib/rubocop/cop/rspec/capybara/visibility_matcher.rb
aried3r: I saw that, but I'm having a hard time changing this from `...` to something like `sym ?` by which I mean, "might take an optional first argument that is a symbol". But I guess the current `...` is also fine, even if it matches signatures that aren't allowed. WDYT?

aried3r:
I believe I have this PR in an acceptable working state except for the node matchers that are not as strict as they could be. But I don't know how to properly write the possible signatures.

pirj:
Looks great, I have nothing to add!
To me, it's none of this cop's business to check if the first argument is a sym or not, so a flexible pattern is fine.

Can you please resolve the conflict?

@bquorning @Darhazer do you have any objections on onboarding this new cop?

aried3r:
> Can you please resolve the conflict?

Rebased, squashed and resolved.

spec/rubocop/cop/rspec/capybara/visibility_matcher_spec.rb
```diff
@@ -0,0 +1,63 @@
...
+    expect_offense(<<-RUBY)
+      expect(page).to have_selector(:css, '.my_element', visible: false)
+                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ Use `:all` or `:hidden` instead of `false`.
+    RUBY
+  end
+
+  it 'registers an offense when using a using multiple options`' do
+    expect_offense(<<-RUBY)
+      expect(page).to have_selector('.my_element', count: 1, visible: false, normalize_ws: true)
+                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ Use `:all` or `:hidden` instead of `false`.
```
bquorning: Instead of
```
expect(page).to have_selector('.my_element', count: 1, visible: false, normalize_ws: true)
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ Use `:all` or `:hidden` instead of `false`.
```
I would prefer
```
expect(page).to have_selector('.my_element', count: 1, visible: false, normalize_ws: true)
                                                       ^^^^^^^^^^^^^^ Use `:all` or `:hidden` instead of `false`.
```

That is, highlighting _only_ the offending part of the hash argument.

You can do this by having the node matchers return that part of the node by using `$`, e.g.
```
(send nil? :have_selector ... (hash <$(pair (sym :visible) true) ...>))
```
and then using the return value (which is also yielded, if you call the node matcher with a block) – e.g.
```rb
            visible_true?(node) do |arg|
              add_offense(arg, message: MSG_TRUE)
            end
```

aried3r: Good point, I didn't know how to do that, thanks for the pointers. I've added these changes and am now waiting for CI. I will squash if you approve the changes.

pirj: Go ahead, looks perfect.

aried3r: Done.

pirj:

bquorning:
🙏 Thank you @aried3r

aried3r:
Thank you for all the help!

PR #891, opened 2020-04-02
Ignore String constants by `RSpec/Describe`

### LINKED ISSUE #889: Suggestion: Disable `RSpec/DescribeClass` for versions
Hello!

I have (almost standard from `bundle gem`) gem version spec:

```ruby
# frozen_string_literal: true

describe 'FooBar::VERSION' do
  subject { FooBar::VERSION }

  it { is_expected.to match(/\d+\.\d+\.\d+/) }
end
```

And it causes an offense:

```
RSpec/DescribeClass: The first argument to describe should be the class or module being tested.	
```

Also, my DRY version:

```ruby
describe(const_name = 'FooBar::VERSION') do
  subject { Object.const_get(const_name) }
```

---

If I'd have `describe FooBar::VERSION do`, then output of specs (`--format=doc`) will contain `VERSION` value instead of constant name, and `described_class` will be `nil`.

AlexWayfer:
Resolve #889 

---

* [x] Added tests.
* [x] Updated documentation.

### DIFF
--- a/lib/rubocop/cop/rspec/describe_class.rb
+++ b/lib/rubocop/cop/rspec/describe_class.rb
@@ -12,6 +12,11 @@ module RuboCop
       #
       #   # good
       #   describe TestedClass do
+      #     subject { described_class }
+      #   end
+      #
+      #   describe 'TestedClass::VERSION' do
+      #     subject { Object.const_get(self.class.description) }
       #   end
       #
       #   describe "A feature example", type: :feature do
@@ -44,12 +49,20 @@ module RuboCop
 
         def_node_matcher :shared_group?, SharedGroups::ALL.block_pattern
 
-        def on_top_level_describe(node, args)
+        def on_top_level_describe(node, (described_value, _))
           return if shared_group?(root_node)
           return if valid_describe?(node)
           return if describe_with_rails_metadata?(node)
+          return if string_constant_describe?(described_value)
 
-          add_offense(args.first)
+          add_offense(described_value)
+        end
+
+        private
+
+        def string_constant_describe?(described_value)
+          described_value.str_type? &&
+            described_value.value =~ /^((::)?[A-Z]\w*)+$/
         end
       end
     end
--- a/spec/rubocop/cop/rspec/describe_class_spec.rb
+++ b/spec/rubocop/cop/rspec/describe_class_spec.rb
@@ -51,6 +51,75 @@ RSpec.describe RuboCop::Cop::RSpec::DescribeClass do
     RUBY
   end
 
+  context 'when argument is a String literal' do
+    it 'ignores class without namespace' do
+      expect_no_offenses(<<-RUBY)
+        describe 'Thing' do
+          subject { Object.const_get(self.class.description) }
+        end
+      RUBY
+    end
+
+    it 'ignores class with namespace' do
+      expect_no_offenses(<<-RUBY)
+        describe 'Some::Thing' do
+          subject { Object.const_get(self.class.description) }
+        end
+      RUBY
+    end
+
+    it 'ignores value constants' do
+      expect_no_offenses(<<-RUBY)
+        describe 'VERSION' do
+          subject { Object.const_get(self.class.description) }
+        end
+      RUBY
+    end
+
+    it 'ignores value constants with namespace' do
+      expect_no_offenses(<<-RUBY)
+        describe 'Some::VERSION' do
+          subject { Object.const_get(self.class.description) }
+        end
+      RUBY
+    end
+
+    it 'ignores top-level constants with `::` at start' do
+      expect_no_offenses(<<-RUBY)
+        describe '::Some::VERSION' do
+          subject { Object.const_get(self.class.description) }
+        end
+      RUBY
+    end
+
+    it 'checks `camelCase`' do
+      expect_offense(<<-RUBY)
+        describe 'activeRecord' do
+                 ^^^^^^^^^^^^^^ The first argument to describe should be the class or module being tested.
+          subject { Object.const_get(self.class.description) }
+        end
+      RUBY
+    end
+
+    it 'checks numbers at start' do
+      expect_offense(<<-RUBY)
+        describe '2Thing' do
+                 ^^^^^^^^ The first argument to describe should be the class or module being tested.
+          subject { Object.const_get(self.class.description) }
+        end
+      RUBY
+    end
+
+    it 'checks empty strings' do
+      expect_offense(<<-RUBY)
+        describe '' do
+                 ^^ The first argument to describe should be the class or module being tested.
+          subject { Object.const_get(self.class.description) }
+        end
+      RUBY
+    end
+  end
+
   it 'ignores request specs' do
     expect_no_offenses(<<-RUBY)
       describe 'my new feature', type: :request do
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -7,6 +7,7 @@ Before submitting the PR make sure the following are checked:
 * [ ] Feature branch is up-to-date with `master` (if not - rebase it).
 * [ ] Squashed related commits together.
 * [ ] Added tests.
+* [ ] Updated documentation.
 * [ ] Added an entry to the [changelog](https://github.com/rubocop-hq/rubocop-rspec/blob/master/CHANGELOG.md) if the new code introduces user-observable changes.
 * [ ] The build (`bundle exec rake`) passes (be sure to run this locally, since it may produce updated documentation that you will need to commit).
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 * Fix `RSpec/FilePath` detection when absolute path includes test subject. ([@eitoball][])
 * Add new `Capybara/VisibilityMatcher` cop. ([@aried3r][])
+* Ignore String constants by `RSpec/Describe`. ([@AlexWayfer][])
 
 ## 1.38.1 (2020-02-15)
 
@@ -494,3 +495,4 @@ Compatibility release so users can upgrade RuboCop to 0.51.0. No new features.
 [@elebow]: https://github.com/elebow
 [@eitoball]: https://github.com/eitoball
 [@aried3r]: https://github.com/aried3r
+[@AlexWayfer]: https://github.com/AlexWayfer
--- a/manual/cops_rspec.md
+++ b/manual/cops_rspec.md
@@ -333,6 +333,11 @@ end
 
 # good
 describe TestedClass do
+  subject { described_class }
+end
+
+describe 'TestedClass::VERSION' do
+  subject { Object.const_get(self.class.description) }
 end
 
 describe "A feature example", type: :feature do


AlexWayfer:
I saw `def_node_matcher`, but I don't know how to insert Regexp and other logic into this, so I've created a regular private method.

spec/rubocop/cop/rspec/describe_class_spec.rb
```diff
@@ -51,6 +51,21 @@
...
+    expect_no_offenses(<<-RUBY)
+      describe 'Some::VERSION' do
+      end
+    RUBY
+  end
+
+  it 'ignores assignment of non-class constants' do
+    expect_no_offenses(<<-RUBY)
+      describe(const_name = 'Some::VERSION') do
+        subject { Object.const_get(const_name) }
```
pirj, line 64: I've never seen this notation before you've mentioned it in the original ticket.
WDYT of:
```ruby
describe 'Some::VERSION' do
  subject { Object.const_get(self.class.description) }
```
We could get rid of the complex dance with `lvasgn`, and you'll get your case covered, and the default one for string description as well.

lib/rubocop/cop/rspec/describe_class.rb
```diff
@@ -44,12 +44,22 @@ class DescribeClass < Cop
 
         def_node_matcher :shared_group?, SharedGroups::ALL.block_pattern
 
-        def on_top_level_describe(node, args)
+        def on_top_level_describe(node, (first_arg, _))
```
pirj, line 47: What would be the best name for the first argument? Maybe `described_class`?

lib/rubocop/cop/rspec/describe_class.rb
```diff
@@ -44,12 +44,22 @@ class DescribeClass < Cop
...
+          add_offense(first_arg)
+        end
+
+        private
+
+        def string_constant_describe?(first_arg)
+          value =
+            first_arg.lvasgn_type? ? first_arg.child_nodes.first : first_arg
+
+          value.str_type? && value.value =~ /(?:[A-Z]\w*(?:::)?)*[A-Z]\w*/
```
pirj, line 62: This regexp is passing for `activeRecord` (unless it's too late to do code reviews).

spec/rubocop/cop/rspec/describe_class_spec.rb
AlexWayfer: I think it'a better approach. Thank you!

lib/rubocop/cop/rspec/describe_class.rb
AlexWayfer: I've took this approach from `lib/rubocop/cop/rspec/describe_method.rb` (`second_arg`), but I think `described_value` is better, yes.

lib/rubocop/cop/rspec/describe_class.rb
AlexWayfer: Thank you! `^` and `$` have been added.

AlexWayfer:
There are also other invalid cases with Regexp, wait a moment please.

AlexWayfer:
> There are also other invalid cases with Regexp, wait a moment please.

Done. Tests have been added.

lib/rubocop/cop/rspec/describe_class.rb
```diff
@@ -44,12 +44,20 @@ class DescribeClass < Cop
...
 
-          add_offense(args.first)
+          add_offense(described_value)
+        end
+
+        private
+
+        def string_constant_describe?(described_value)
+          described_value.str_type? &&
+            described_value.value =~ /^(?:(?:::)?[A-Z]\w*)+$/
```
pirj, line 60: Deep research [here](https://stackoverflow.com/questions/53549477/regex-to-test-criteria-for-class-name), check it out.

This regex matched `'Active_Record::Base'`, I first thought it shouldn't, but it seems to be a valid class name though.

Is `?:` necessary? You can use `.match?` instead of `=~` to get a boolean.

spec/rubocop/cop/rspec/describe_class_spec.rb
```diff
@@ -51,6 +51,33 @@
     RUBY
   end
 
+  context 'with non-class constants' do
```
pirj, line 54: WDYT of "with class or constant name as a string literal"?

spec/rubocop/cop/rspec/describe_class_spec.rb
```diff
@@ -51,6 +51,33 @@
...
+    it 'ignores them' do
+      expect_no_offenses(<<-RUBY)
+        describe 'Some::VERSION' do
+          subject { Object.const_get(self.class.description) }
+        end
+      RUBY
+    end
+
+    it 'ignores root-constants with `::` at start' do
```
pirj, line 63: "ignores top-level constants"

spec/rubocop/cop/rspec/describe_class_spec.rb
```diff
@@ -51,6 +51,33 @@
...
   end
 
+  context 'with non-class constants' do
+    it 'ignores them' do
+      expect_no_offenses(<<-RUBY)
+        describe 'Some::VERSION' do
+          subject { Object.const_get(self.class.description) }
+        end
+      RUBY
+    end
```
pirj, line 61: Maybe add a similar example with a regular class, e.g. `'Some::Thing'`? And also one without namespace `'Something'`.

lib/rubocop/cop/rspec/describe_class.rb
AlexWayfer: > This regex matched `'Active_Record::Base'`, I first thought it shouldn't, but it seems to be a valid class name though.

Yep, that's why I leave this.

---

> Is `?:` necessary?

No. I thought it's faster, but not:

```ruby
# frozen_string_literal: true

require 'bundler/setup'
Bundler.setup

require 'pry-byebug'

require 'benchmark'
require 'benchmark/ips'
require 'benchmark/memory'


puts '```ruby'
puts File.read(__FILE__).gsub("\t", '  ')
puts '```'
puts
puts '### Output'
puts
puts '```'


STRING = 'Active_Record::Base'


def with_non_captured_group
  STRING =~ /^(?:(?:::)?[A-Z]\w*)+$/
end

def with_captured_group
  STRING =~ /^(?:(?:::)?[A-Z]\w*)+$/
end


def test
  results = %i[
    with_non_captured_group
    with_captured_group
  ].each_with_object([]) do |method_name, array|
    array << (result = send method_name)
    puts "Result of #{method_name}:"
    pp result
    puts
  end
  uniq_size = results.uniq.size
  puts "Uniq results: #{uniq_size}"
  puts
  exit 1 if uniq_size > 1
end

test

Benchmark.ips do |x|
  x.report('with_non_captured_group') { with_non_captured_group }
  x.report('with_captured_group') { with_captured_group }

  x.compare!
end

Benchmark.memory do |x|
  x.report('with_non_captured_group') { 100.times { with_non_captured_group } }
  x.report('with_captured_group') { 100.times { with_captured_group } }

  x.compare!
end


puts '```'
```

### Output

```
Result of with_non_captured_group:
0

Result of with_captured_group:
0

Uniq results: 1

Warming up --------------------------------------
with_non_captured_group
                       185.982k i/100ms
 with_captured_group   183.201k i/100ms
Calculating -------------------------------------
with_non_captured_group
                          2.463M (± 3.0%) i/s -     12.461M in   5.064546s
 with_captured_group      2.430M (± 2.7%) i/s -     12.274M in   5.055925s

Comparison:
with_non_captured_group:  2462742.5 i/s
 with_captured_group:  2429607.0 i/s - same-ish: difference falls within error

Calculating -------------------------------------
with_non_captured_group
                        16.800k memsize (     0.000  retained)
                       100.000  objects (     0.000  retained)
                         0.000  strings (     0.000  retained)
 with_captured_group    16.800k memsize (     0.000  retained)
                       100.000  objects (     0.000  retained)
                         0.000  strings (     0.000  retained)

Comparison:
with_non_captured_group:      16800 allocated
 with_captured_group:      16800 allocated - same
```

Removed.

---



> You can use `.match?` instead of `=~` to get a boolean.

I wanted, but I saw `spec.required_ruby_version = '>= 2.3.0'`, and `String#match?` was introduced in Ruby 2.4.

spec/rubocop/cop/rspec/describe_class_spec.rb
AlexWayfer: > WDYT of "with class or constant name as a string literal"?

OK, but "class or constant" is confusing (class is constant, is not it?), so I've left "with constant name as a String literal".

AlexWayfer: It turned into "when argument is a String literal".

spec/rubocop/cop/rspec/describe_class_spec.rb
AlexWayfer: > "ignores top-level constants"

Replaced "root-constants" with "top-level constants", but they may be referenced without `::` at the start, so I left this.

spec/rubocop/cop/rspec/describe_class_spec.rb
AlexWayfer: OK, added. Also another tests for empty strings and strings starting with numbers.

lib/rubocop/cop/rspec/describe_class.rb
pirj: Fair enough 👍

pirj:
Good job @AlexWayfer! 👏 

@bquorning @Darhazer can you please glance over?

AlexWayfer:
> Good job @AlexWayfer! :clap: 

Thank you. Your help is appreciated.

AlexWayfer:
I have an idea: add a check like `Object.const_defined?`. What do you think? More strict.

Darhazer:
> I have an idea: add a check like Object.const_defined?. What do you think? More strict.

It requires the app to be loaded. Rubocop does static analysis, so it doesn't know what constants are really defined. And if it needed to be defined before the specs are run, it would defeat the purpose of having it in a string

lib/rubocop/cop/rspec/describe_class.rb
```diff
@@ -44,12 +44,20 @@ class DescribeClass < Cop
...
 
-          add_offense(args.first)
+          add_offense(described_value)
+        end
+
+        private
+
+        def string_constant_describe?(described_value)
+          described_value.str_type? &&
+            described_value.value =~ /^((::)?[A-Z]\w*)+$/
```
bquorning, line 60: Could you please update the cop documentation with another “good” example, e.g. `'Some::Thing'`?

AlexWayfer:
> It requires the app to be loaded. Rubocop does static analysis, so it doesn't know what constants are really defined. And if it needed to be defined before the specs are run, it would defeat the purpose of having it in a string

I didn't understand.

Expected behavior by this cop: use constants. Not stubbed constants. And test them. So, to use (and test) constant you have to load them, don't you?

lib/rubocop/cop/rspec/describe_class.rb
AlexWayfer: Thank you, I forgot about this, added.

Also with `subject { Object.const_get(self.class.description) }` — it can be helpful for others.

And I've added PR template checkbox about documentation in the same commit — I hope it's OK.

Darhazer:
> Expected behavior by this cop: use constants. Not stubbed constants. And test them. So, to use (and test) constant you have to load them, don't you?

The point is, rubocop have no way of knowing about any constant defined outside of the file it inspects, it makes a static analysis that is limited to one file at a time, and doesn't actually run your application code.So it's not the right tool to verify that you are describing a real constant

AlexWayfer:
> The point is, rubocop have no way of knowing about any constant defined outside of the file it inspects

And what? RuboCop lints `version_spec`… oh, I got it: although `version_spec` requires `lib/version`, but RuboCop doesn't execute this `require` while linting. OK, sorry.

Darhazer:
> And what? RuboCop lints version_spec… oh, I got it: although version_spec requires lib/version, but RuboCop doesn't execute this require while linting. OK, sorry.

No problem. It's actually a limitation of Robocop actually.

pirj:
Congratulations on your first contribution 🎉  @AlexWayfer

AlexWayfer:
> Congratulations on your first contribution 🎉 @AlexWayfer

Thank you all, I'm glad to help (myself in the first place, yeah 😅 ).

PR #892, opened 2020-04-03
Add IgnoreSharedGroups to RSpec/ExpectInHook cop

jonatas:
Added this small option for shared examples, since we don't have a proper syntax to cover shared hooks that have mocks and stubs.

They seem quite useful, I'm not sure if that is true for the entire community. Anyway, it would be great if we have some better syntax in the RSpec to allow set some previous expectations about method calls.

* [x] Added tests.

* (n/a) Added the new cop to `config/default.yml`.

### DIFF
--- a/lib/rubocop/cop/rspec/expect_in_hook.rb
+++ b/lib/rubocop/cop/rspec/expect_in_hook.rb
@@ -5,6 +5,9 @@ module RuboCop
     module RSpec
       # Do not use `expect` in hooks such as `before`.
       #
+      # With `IgnoreSharedGroups` the cop will ignore expectations in hooks
+      # directly in shared groups.
+      #
       # @example
       #   # bad
       #   before do
--- a/lib/rubocop/cop/rspec/expect_in_hook.rb
+++ b/lib/rubocop/cop/rspec/expect_in_hook.rb
@@ -36,8 +36,8 @@ module RuboCop
 
         def on_block(node)
           return unless hook?(node)
-          return if node.body.nil?
-          return if ignore_shared_groups? && nested_in_shared_group?(node)
+          return unless node.body
+          return if ignore_shared_groups? && directly_in_shared_group?(node)
 
           expectation(node.body) do |expect|
             add_offense(expect, location: :selector,
@@ -47,8 +47,8 @@ module RuboCop
 
         private
 
-        def nested_in_shared_group?(node)
-          node.each_ancestor(:block).any?(&method(:shared_group?))
+        def directly_in_shared_group?(node)
+          shared_group?(node.each_ancestor(:block).first)
         end
 
         def ignore_shared_groups?
--- a/lib/rubocop/cop/rspec/expect_in_hook.rb
+++ b/lib/rubocop/cop/rspec/expect_in_hook.rb
@@ -20,6 +20,15 @@ module RuboCop
       #   it do
       #     expect(something).to eq 'foo'
       #   end
+      #
+      # @example `IgnoreSharedGroups: true`
+      #
+      #   # good
+      #   shared_example 'some shared setup' do
+      #     after do
+      #       expect_any_instance_of(Something).to receive(:foo)
+      #     end
+      #   end
       class ExpectInHook < Cop
         MSG = 'Do not use `%<expect>s` in `%<hook>s` hook'
 
@@ -28,6 +37,7 @@ module RuboCop
         def on_block(node)
           return unless hook?(node)
           return if node.body.nil?
+          return if ignore_shared_groups? && nested_in_shared_group?(node)
 
           expectation(node.body) do |expect|
             add_offense(expect, location: :selector,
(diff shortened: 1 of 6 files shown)


jonatas:
I did some work with @pirj and I moved my personal fork for him because of some permissions and forget completely. I hope Fillip don't mind, I'll recreate my fork 🤞

lib/rubocop/cop/rspec/expect_in_hook.rb
```diff
@@ -37,6 +47,14 @@ def on_block(node)
 
         private
 
+        def nested_in_shared_group?(node)
+          node.each_ancestor(:block).any?(&method(:shared_group?))
```
pirj, line 51: I can think of the following, where expect in hook is not directly nested in a shared group:
```ruby
shared_example 'some shared examles' do
  describe '#some_method' do
    context 'in some case' do
      after do
        expect_any_instance_of(Something).to receive(:foo) # NASTY!
      end

      it 'does that' do
        something.perform
      end
    end
  end
end
```

pirj:
> I did some work with @pirj and I moved my personal fork for him because of some permissions and forget completely. I hope Fillip don't mind, I'll recreate my fork 

Sure, please go ahead. It completely slipped my mind we had something unfinished. Let's finish it and make it public 👍

lib/rubocop/cop/rspec/expect_in_hook.rb
jonatas: Good catch @pirj! I'll add an extra spec to cover both and make it explicit that it works in all the levels.


Also, it's also not being actually covered, looks like the implementation needs some fixes too :D

spec/rubocop/cop/rspec/expect_in_hook_spec.rb
```diff
@@ -76,4 +76,56 @@
...
+        end
+      end
+        expect(object).to receive(:message)
+        expect_any_instance_of(something).to receive(:foo)
+      RUBY
+    end
+  end
+
+  context 'with config IgnoreSharedGroups set to true' do
+    let(:cop_config) { { 'IgnoreSharedGroups' => true } }
```
pirj, line 100: Ok, well. We have those two examples below. What makes you think both of them should be ignored by the cop?

jonatas:
I agree @pirj. We need to approach it in a better way and the examples are not reasonable.

pirj:
Fixed the cop slightly.
Is that what you had in mind @jonatas ?

jonatas:
Thanks for the updates @pirj. I thought it was not going to be like a good practice at all but I'm happy that you moved it forward 🥇

pirj:
@bquorning @Darhazer WDYT?

pirj:
@jonatas @Darhazer what are your thoughts on this pull request guys?

Darhazer:
I'm fine with having such a setting

What's the idea of `Only ignore expectations in hooks directly under shared examples`. Seems a bit dangerous approach to me.

pirj:
As far as I understand (and what my personal approach to specs is able to tolerate) is to have expectations in hooks inside shared examples (but not regular example groups) is due to:
```
context 'with this and that' do
  it_behaves_like 'it does something' do
    let(:this) { ... }
    let(:that) { ... }

    it { perform_action }
  end
end

shared_examples_for 'it does something' do
  before do # shared message expectation
    expect(something).to receive(:blah)
  end

  after do # shared expectations
    expect(this).to be_this
    expect(that).to be_that
  end

  describe 'something else' do
    before do # I don't have a justification to tolerate this
       expect(this).to be_have_some_other_property
    end

    it "I'm an example unrelated to the parent example group, I just add yet another expectation" do
      # ...
    end
  end
end
```

Darhazer:
Yeah, the question was while you check only the immediate block, and not the first example_group block, and then check if it is a shared or not?

pirj:
I got it, sorry for being slow.

I've crafter this example really quick:
```ruby
  it 'adds an offense for `expect` in `before` hook' do
    expect_offense(<<-RUBY)
      RSpec.describe do
        describe do
          before do
            expect(something).to eq('foo')
            ^^^^^^ Do not use `expect` in `before` hook
            is_expected.to eq('foo')
            ^^^^^^^^^^^ Do not use `is_expected` in `before` hook
            expect_any_instance_of(Something).to receive(:foo)
            ^^^^^^^^^^^^^^^^^^^^^^ Do not use `expect_any_instance_of` in `before` hook
          end
        end
        context do
          before do
            expect(something).to eq('foo')
            ^^^^^^ Do not use `expect` in `before` hook
            is_expected.to eq('foo')
            ^^^^^^^^^^^ Do not use `is_expected` in `before` hook
            expect_any_instance_of(Something).to receive(:foo)
            ^^^^^^^^^^^^^^^^^^^^^^ Do not use `expect_any_instance_of` in `before` hook
          end
        end
      end
      before do
        expect(something).to eq('foo')
        ^^^^^^ Do not use `expect` in `before` hook
        is_expected.to eq('foo')
        ^^^^^^^^^^^ Do not use `is_expected` in `before` hook
        expect_any_instance_of(Something).to receive(:foo)
        ^^^^^^^^^^^^^^^^^^^^^^ Do not use `expect_any_instance_of` in `before` hook
      end
    RUBY
  end
```

and it passes. Does it address your concern?

Considering `def_node_search` and `each_ancestor` it makes a lot of sense to me to add such a spec.

pirj:
Closing as it seems this pull request doesn't seem to be too active.
Please feel free to reopen.

PR #894, opened 2020-04-08
[WIP] catch RSpec/ExpectInHook offenses in the setup block

lazycoder9:
Allow catch RSpec/ExpectInHook offenses in the setup block.
Separate hooks to `BUILTIIN` and `CUSTOM`. Custom hooks will be loaded from from `AllCops/RSpec/CustomHooks`

---

* [ ] Added tests.
* [ ] Updated documentation.

### DIFF
--- a/spec/rubocop/cop/rspec/expect_in_hook_spec.rb
+++ b/spec/rubocop/cop/rspec/expect_in_hook_spec.rb
@@ -76,4 +76,22 @@ RSpec.describe RuboCop::Cop::RSpec::ExpectInHook do
       end
     RUBY
   end
+
+  it 'adds an offense for `expect` with block in `setup` hook' do
+    expect_offense(<<-RUBY)
+      setup do
+        expect { something }.to eq('foo')
+        ^^^^^^ Do not use `expect` in `setup` hook
+      end
+    RUBY
+  end
+
+  it 'adds an offense for `expect` with block in custom `custom_hook` hook' do
+    expect_offense(<<-RUBY)
+      custom_hook do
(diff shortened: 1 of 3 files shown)


lazycoder9:
@pirj this is very raw implementation and I've stuck in testing custom hooks. The problem is to avoid changing `config.yml` file to pass `custom hook` spec.
I really don't know how to stub or change config file during only one spec.
The main problem is that configuration file is read only once before running specs, which leads definition of all constants (e.g. Hooks::ALL) and keeps them immutable during all specs. Attempts to stubbing constants did not work
I will appreciate if you can tell me possible solution

spec/rubocop/cop/rspec/expect_in_hook_spec.rb
```diff
@@ -76,4 +76,22 @@
       end
     RUBY
   end
+
```
pirj, line 79: [This](https://github.com/rubocop-hq/rubocop/blob/master/lib/rubocop/rspec/shared_contexts.rb#L42) might be used somehow to stub `AllCops`. I believe we'll have to patch it in the long run to accept arbitrary keypairs (that would also allow moving Rails-related code to `rspec-rails`).
It's ok to use a copy of it now.

This doesn't solve the issue of `ALL` being pre-calculated though.
What if we take a non-performant way and define `ALL` as a method that would return `BUILD` plus fetch from config?

I vote for postponing the decision on how to cache config and be flexible in the specs at the same time.

pirj:
Closing in favour of #956

PR #896, opened 2020-04-20
WIP: Add autocorrect to RSpec/NamedSubject cop

aishek:
I'm working on autocorrect for `RSpec/NamedSubject` cop

---

* [ ] Added tests.
* [ ] Updated documentation.

* (n/a) Added the new cop to `config/default.yml`.
* (n/a) The cop documents examples of good and bad code.
* (n/a) The tests assert both that bad code is reported and that good code is not reported.

### DIFF
--- a/lib/rubocop/cop/rspec/named_subject.rb
+++ b/lib/rubocop/cop/rspec/named_subject.rb
@@ -59,6 +59,13 @@ module RuboCop
         def_node_search :subject_usage, '$(send nil? :subject)'
 
         def on_block(node)
+          add_offense(node) if node.parent.nil? && use_implicit_subject?(node)
+
+          RuboCop::RSpec::ExampleGroup.new(node).subjects.each do |subject|
+            next if named_subject?(subject)
+            add_offense(subject)
+          end
+
           return if !rspec_block?(node) || ignored_shared_example?(node)
 
           subject_usage(node) do |subject_node|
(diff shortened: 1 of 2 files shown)


lib/rubocop/cop/rspec/named_subject.rb
```diff
@@ -59,6 +59,11 @@ class NamedSubject < Cop
...
 
         def on_block(node)
+          RuboCop::RSpec::ExampleGroup.new(node).subjects.each do |subject|
+            next if named_subject?(subject)
+            add_offense(subject)
+          end
+
           return if !rspec_block?(node) || ignored_shared_example?(node)
 
           subject_usage(node) do |subject_node|
```
pirj, line 69: Existing `subject_usage` is sub-optimal since it seems to recursively find `subject` declarations due to `def_node_search`. Should probably be `def_node_matcher` in the first place.

lib/rubocop/cop/rspec/named_subject.rb
```diff
@@ -59,6 +59,11 @@ class NamedSubject < Cop
         def_node_search :subject_usage, '$(send nil? :subject)'
```
pirj, line 59: Do you think [this](https://github.com/rubocop-hq/rubocop-rspec/blob/master/lib/rubocop/rspec/language.rb#L116) existing helper may come handy (to also detect `subject!`)?

lib/rubocop/cop/rspec/named_subject.rb
```diff
@@ -70,6 +75,79 @@ def ignored_shared_example?(node)
...
+
+        def underscore(camel_cased_word)
+          return camel_cased_word unless /[A-Z-]|::/.match?(camel_cased_word)
+          word = camel_cased_word.to_s.gsub("::", "/")
+          word.gsub!(/([A-Z\d]+)([A-Z][a-z])/, '\1_\2')
+          word.gsub!(/([a-z\d])([A-Z])/, '\1_\2')
+          word.tr!("-", "_")
+          word.downcase!
+          word
+        end
```
pirj, line 150: This cop seems to do two things:
1. Detects unnamed `subject` declarations
2. Detects unnamed `subject` usages where the subject has a name

Do you think it would be easier to split it into two cops before implementing auto-correction? Is there something in common between `correct_subject_definition` and `correct_subject_usage`?

aishek: It seems like a good idea.

I'm working or task where I need to fix several thousands cop offences. My strategy is meet new case and then create autocorrection for it. When I done, I will update the PR and refactor it.

Thanks for comments and valuable thougths! Hope get back soon 😀

pirj:
Ping. This is a very nice addition, would you like to wrap it up?

pirj:
Please feel free to reopen when you have time to wrap this up.

PR #897, opened 2020-04-20
Add new `RSpec/EmptyHooks` cop

### LINKED ISSUE #811: Detect empty hooks (before, after, …)
This morning I found an empty `before` hook in our codebase:
```ruby
# frozen_string_literal: true

RSpec.describe Foo do
  before do
  end

  describe '#foo' do
    it { works }
  end
end
```

I would be nice if there were a cop that detected hooks with empty blocks.

tejasbubane:
Closes #811 

---

* [x] Added tests.
* [x] Updated documentation.


lib/rubocop/cop/rspec/empty_hook.rb
```diff
@@ -0,0 +1,41 @@
...
+    module RSpec
+      #  Checks for empty before and after hooks.
+      #
+      # @example
+      #   # bad
+      #   before {}
+      #   after do; end
+      #   before(:all) do
+      #   end
+      #   all(:all) { nil }
```
pirj, line 14: Sorry, I'm not familiar with this construct. Where does it come from?

lib/rubocop/cop/rspec/empty_hook.rb
```diff
@@ -0,0 +1,41 @@
...
+      #
+      #   # good
+      #   before { create_users }
+      #   after do; cleanup_users; end
```
pirj, line 18: WDYT of making it a multiline block just like `before` below?

spec/rubocop/cop/rspec/empty_hook_spec.rb
```diff
@@ -0,0 +1,76 @@
+# frozen_string_literal: true
+
+RSpec.describe RuboCop::Cop::RSpec::EmptyHook do
+  subject(:cop) { described_class.new(config) }
+
+  let(:config) { RuboCop::Config.new }
```
pirj, line 6: There's no config for this cop, so those two declarations are redundant.

lib/rubocop/cop/rspec/empty_hook.rb
```diff
@@ -0,0 +1,41 @@
...
+      #     create_feed
+      #   end
+      #   all(:all) { cleanup_feed }
+      class EmptyHook < Cop
+        MSG = 'Empty hook detected.'
+
+        def_node_matcher :empty_hook?, <<~PATTERN
+          (block
+            $(send nil? {:before :after}
+              (sym {:each :all :suite})?)
```
pirj, line 29: There are a few aliases you're missing here. Luckily, they are conveniently defined [here](https://github.com/rubocop-hq/rubocop-rspec/blob/master/lib/rubocop/rspec/language.rb#L101).

lib/rubocop/cop/rspec/empty_hook.rb
```diff
@@ -0,0 +1,41 @@
...
+      #   before(:all) do
+      #     create_feed
+      #   end
+      #   all(:all) { cleanup_feed }
+      class EmptyHook < Cop
+        MSG = 'Empty hook detected.'
+
+        def_node_matcher :empty_hook?, <<~PATTERN
+          (block
+            $(send nil? {:before :after}
```
pirj, line 28: There are some aliases and rarely used hooks, all defined [here](https://github.com/rubocop-hq/rubocop-rspec/blob/master/lib/rubocop/rspec/language.rb#L87). I understand that `after` is quite different, but wouldn't it be bad to detect an empty around (without `yield`) along the way?

lib/rubocop/cop/rspec/empty_hook.rb
```diff
@@ -0,0 +1,41 @@
...
+              (sym {:each :all :suite})?)
+            _ {(:nil) nil?})
```
pirj, line 30: Is `before { nil }` coming from a real-life example? I can't figure out why would anyone do that.

spec/rubocop/cop/rspec/empty_hook_spec.rb
```diff
@@ -0,0 +1,76 @@
...
+      RUBY
+    end
+
+    it 'registers offense with empty multi-line do-end' do
+      expect_offense(<<~RUBY)
+        #{hook_type} do
+        #{'^' * hook_type.length} Empty hook detected.
+        end
+      RUBY
+    end
```
pirj, line 29: I understand that you are trying to cover all the syntactic cases here, but there's no difference is AST that RuboCop operates on between an empty, or single-line, or multi-line block, so it's not really necessary to have separate examples for those two.

spec/rubocop/cop/rspec/empty_hook_spec.rb
```diff
@@ -0,0 +1,76 @@
...
+      expect_no_offenses(<<~RUBY)
+        #{hook_type} { create_users }
+      RUBY
+    end
+
+    it 'allows multi-line non-empty blocks' do
+      expect_no_offenses(<<~RUBY)
+        #{hook_type} do
+          create_users
+          create_businesses
```
pirj, line 57: A block with several statements is actually different in AST 👍

spec/rubocop/cop/rspec/empty_hook_spec.rb
```diff
@@ -0,0 +1,76 @@
...
+    it_behaves_like 'empty hook', 'before(:suite)'
+    it_behaves_like 'empty hook', 'before(:all)'
+  end
+
+  context 'with after hooks' do
+    it_behaves_like 'empty hook', 'after'
+    it_behaves_like 'empty hook', 'after(:each)'
+    it_behaves_like 'empty hook', 'after(:suite)'
+    it_behaves_like 'empty hook', 'after(:all)'
+  end
```
pirj, line 75: Even though this suite is comprehensive and concise, it has redundancy and is quite hard to read in regards to using string interpolation.
WDYT of writing it instead as:
```ruby
it 'detects an offense for an empty `before`'
  ...
end

it 'detects the offense for an empty `before` with a scope' do
  ...
end

it 'detects ... `after`' do
it 'detects ... `prepend_after` do
it 'detects `:all` scope' do
it 'detects `:context' scope' do
```
etc?

pirj:
Ping

lib/rubocop/cop/rspec/empty_hook.rb
tejasbubane: I meant `after(:all)` :) will change

tejasbubane:
@pirj Fixed. Please review.

spec/rubocop/cop/rspec/empty_hook_spec.rb
```diff
@@ -0,0 +1,418 @@
...
+    it 'detects offense for empty `around` with :suite scope' do
+      expect_offense(<<~RUBY)
+        around(:suite) {}
+        ^^^^^^^^^^^^^^ Empty hook detected.
+      RUBY
+    end
+
+    it 'accepts non-empty `around` hook' do
+      expect_no_offenses(<<~RUBY)
+        around { yield }
```
pirj, line 169: 👍

spec/rubocop/cop/rspec/empty_hook_spec.rb
```diff
@@ -0,0 +1,418 @@
...
+      RUBY
+    end
+
+    it 'accepts multiline `around` hook' do
+      expect_no_offenses(<<~RUBY)
+        around(:suite) do
+          setup_users
+          yield
```
pirj, line 177: 👍

spec/rubocop/cop/rspec/empty_hook_spec.rb
```diff
@@ -0,0 +1,418 @@
...
+        ^^^^^ Empty hook detected.
+      RUBY
+    end
+
+    it 'detects offense for empty `after` with :each scope' do
+      expect_offense(<<~RUBY)
+        after(:each) {}
+        ^^^^^^^^^^^^ Empty hook detected.
+      RUBY
+    end
```
pirj, line 78: I love how you cover every possible case. However, with N hooks and M scopes, it's getting quite unwieldy.
Would you agree to keep only a few distinctive examples for each hook type to keep this spec short?

E.g. all scopes for `before`, `after {}`, `around {}`, `around { yield }`, `around { setup; yield }`, `prepend_before {}`, `append_before {}`, `prepend_after {}`, `append_after {}`, and maybe keep "accepts" examples for `before`/`prepend_before`/`append_before`/`prepend_after`/`append_after` if you see fit.

I can't remember a good example when the lack of covering the full matrix has missed a real problem given a simple combination of matchers in the node pattern.

In any case, `around(:suite)` is illegal as to my best knowledge, and `around(:context)`/`around(:all)` issues a warning that it actually behaves like `around(:example)`.

pirj: PS I really love how semantic current examples are, e.g. `before { setup_users }` vs `after { cleanup_users` 👍

lib/rubocop/cop/rspec/empty_hook.rb
```diff
@@ -0,0 +1,44 @@
...
+      class EmptyHook < Cop
+        MSG = 'Empty hook detected.'
+
+        def_node_matcher :empty_hook?, <<~PATTERN
+          (block
+            $(send nil? {:prepend_before :before :append_before :around
+                         :prepend_after :after :append_after}
+              (sym {:each :example :context :all :suite})?)
+            _ nil?)
+        PATTERN
```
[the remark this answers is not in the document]
bquorning: @pirj’s suggestion works, just replace `Scopes` with `Hooks::Scopes` :-)

spec/rubocop/cop/rspec/empty_hook_spec.rb
bquorning: I agree the specs are a bit long and maybe too explicit.

One way to shorten them might be to expect more than one type of offense at the same time, e.g.:

```rb
    it 'detects offenses for empty `after` hooks' do
      expect_offense(<<~RUBY)
        after {}
        ^^^^^ Empty hook detected.
        after(:each) {}
        ^^^^^^^^^^^^ Empty hook detected.
        after(:example) {}
        ^^^^^^^^^^^^^^^ Empty hook detected.
        after(:context) {}
        ^^^^^^^^^^^^^^^ Empty hook detected.
        after(:all) {}
        ^^^^^^^^^^^ Empty hook detected.
        after(:suite) {}
        ^^^^^^^^^^^^^ Empty hook detected.
      RUBY
    end
```

lib/rubocop/cop/rspec/empty_hook.rb
```diff
@@ -0,0 +1,45 @@
...
+      #     create_feed
+      #   end
+      #   after(:all) { cleanup_feed }
+      class EmptyHook < Cop
+        MSG = 'Empty hook detected.'
+
+        def_node_matcher :empty_hook?, <<~PATTERN
+          (block
+            $(send nil? #{Hooks::ALL.node_pattern_union}
+              (sym #{Hooks::Scopes::ALL.node_pattern_union})?
```
Darhazer, line 31: I guess we need to extend the `Language` module with such definitions.

Darhazer: But do we need to check the argument at all? Wouldn't `Hooks::ALL.block_pattern` actually do the job?

tejasbubane:
@pirj Changes done.

lib/rubocop/cop/rspec/empty_hook.rb
tejasbubane: Indeed we don't need to check the argument. `block_pattern` does not check for inner contents of the block so I am now using `Hooks::ALL.send_pattern` & empty block check with `nil?`

bquorning:
> You can add autocorrect that removes the offending node

Do you want to give it a shot, @tejasbubane? Otherwise let’s open a new issue so we don’t forget to implement it.

tejasbubane:
@bquorning Added autocorrect.

lib/rubocop/cop/rspec/empty_hook.rb
```diff
@@ -0,0 +1,51 @@
...
+        def on_block(node)
+          empty_hook?(node) do |hook|
+            add_offense(hook)
+          end
+        end
+
+        def autocorrect(node)
+          lambda do |corrector|
+            block = node.parent
+            range = range_by_whole_lines(block.loc.expression,
```
bquorning, line 43: In the unlikely event that anyone has `before {} ; after { clean_up(:foo) }` on the same line, the autocorrect would remove both. Maybe `#range_with_surrounding_space` is safer to use here?

I wonder how many coops are using `#range_by_whole_lines` and don’t work with multiple statements on the same line? 🤷‍♂️

tejasbubane: This is a little tricky. `range_with_surrounding_space` does not remove the semi-colon. We probably need to add method like `range_with_semicolon` to `RangeHelp`.

bquorning: Removing the semicolon is a job for the `Style/Semicolon` cop.

In our specs you’ll sometimes see an autocorrection introducing e.g. [a blank line at the end of a block](https://github.com/rubocop-hq/rubocop-rspec/blob/0f7721ed39c2276a8f8d7a796213b19ed6bf6b6a/spec/rubocop/cop/rspec/scattered_let_spec.rb#L71-L84) or similar “ugly” formatting. And that’s fine. When auto-correcting RuboCop will automatically run the resulting file though the cops once more to see if new offenses were introduced.

tejasbubane: I have this solution that passes the test - but is quite ugly (hence not committed):

```ruby
        def autocorrect(node)
          lambda do |corrector|
            corrector.remove(range_for_autocorrect(node))
          end
        end

        private

        def range_for_autocorrect(node)
          block = node.parent
          range = block.loc.expression
          buffer = @processed_source.buffer
          # Handle multiple statements on same line
          if buffer.source[range.end_pos] == ';'
            range = Parser::Source::Range.new(@processed_source.buffer,
                                              range.begin_pos,
                                              range.end_pos + 1)
          end
          range_with_surrounding_space(range: range)
        end
```

tejasbubane: That's neat 👍 I've changed autocorrect to use `range_with_surrounding_space`

bquorning:
I’d like a second review from @pirj or @Darhazer before merging.

tejasbubane:
@pirj Added spec for semicolon - it is a little ugly though - since it does not remove the semicolon.

pirj:
> Added spec for semicolon - it is a little ugly though - since it does not remove the semicolon.

No worries, it's less ugly than `before {}; after { delete_users }` 😄

PR #899, opened 2020-04-27
Replace `should`-example with `expect`-example

ngoral:
In the docs for `RSpec/NamedSubject` we use `should`-style example as a good one. Since `expect`-style is a modern and enforced one, I guess it would be more correct to use `expect`-style examples.

---

* [ ] Added tests.
* [x] Updated documentation.


pirj:
@ngoral Do you mind to update from `master` to make all build jobs green?

PR #902, opened 2020-04-27
Fix RuboCop builds

pirj:
Bump TargetRubyVersion to 2.4 as latest RuboCop version, 0.82, dropped support for 2.3 in TargetRubyVersion.

RuboCop has updated its lowest supported Ruby version syntax to 2.4 in version 0.82, and our builds pick the most recent release on CI, and this was failing all RuboCop checks.

    Gemspec/RequiredRubyVersion: required_ruby_version (2.3, declared in rubocop-rspec.gemspec) and TargetRubyVersion (2.4, which may be specified in .rubocop.yml) should be equal.

No, it shouldn't be equal. We're an extension that has support for earlier RuboCop versions and thus earlier Ruby versions.
---

* [-] Added tests.
* [-] Updated documentation.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -5,7 +5,7 @@ require:
 
 AllCops:
   DisplayCopNames: true
-  TargetRubyVersion: 2.3
+  TargetRubyVersion: 2.4
   Exclude:
     - 'vendor/**/*'
     - 'spec/fixtures/**/*'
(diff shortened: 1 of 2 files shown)


bquorning:
Great minds think alike: #901.

pirj:
I was triggered by a failure in https://github.com/rubocop-hq/rubocop-rspec/pull/899 XD and tried to ignore red `edge-rubocop` to the last moment.

pirj:
Closing in favour of #901

PR #903, opened 2020-04-28
Scattered let autocorrect

Darhazer:
I realized that we don't have an autocorrect support for scattered let, but it's quite easy to do, as we already move nodes around. 

I also refactored the autocorrect part, as I had to copy/paste 3 methods one more time, and we had enough copies already.

It seems that rubocop-rspec deals with the structure much more than rubocop itself - I found only one cop there that would benefit from the new helper, that's why I left it as part of rubocop-rspec instead.

---

* [x] Added tests.
* [x] Updated documentation.

### DIFF
--- a/lib/rubocop/cop/rspec/hooks_before_examples.rb
+++ b/lib/rubocop/cop/rspec/hooks_before_examples.rb
@@ -24,9 +24,6 @@ module RuboCop
       #   end
       #
       class HooksBeforeExamples < Cop
-        include RangeHelp
-        include RuboCop::RSpec::FinalEndLocation
-
         MSG = 'Move `%<hook>s` above the examples in the group.'
 
         def_node_matcher :example_or_group?, <<-PATTERN
@@ -45,11 +42,9 @@ module RuboCop
         def autocorrect(node)
           lambda do |corrector|
             first_example = find_first_example(node.parent)
-            first_example_pos = first_example.loc.expression
-            indent = "\n" + ' ' * first_example.loc.column
-
-            corrector.insert_before(first_example_pos, source(node) + indent)
-            corrector.remove(node_range_with_surrounding_space(node))
+            RuboCop::RSpec::Corrector::MoveNode.new(
+              node, corrector, processed_source
+            ).move_before(first_example)
           end
         end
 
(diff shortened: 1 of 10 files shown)


bquorning:
@Darhazer FYI the build failed.

Darhazer:
@bquorning fixed. Sorry, I wanted to share the code with a colleague and forgot to actually run the rake task

lib/rubocop/cop/rspec/hooks_before_examples.rb
```diff
@@ -45,11 +42,9 @@ def on_block(node)
...
           lambda do |corrector|
             first_example = find_first_example(node.parent)
-            first_example_pos = first_example.loc.expression
-            indent = "\n" + ' ' * first_example.loc.column
-
-            corrector.insert_before(first_example_pos, source(node) + indent)
-            corrector.remove(node_range_with_surrounding_space(node))
+            RuboCop::RSpec::Corrector::MoveNode.new(
+              node, corrector, processed_source
+            ).move_before(first_example)
```
pirj, line 47: 👏

spec/rubocop/cop/rspec/scattered_let_spec.rb
```diff
@@ -12,6 +12,76 @@
         ^^^^^^^^^^^^^ Group all let/let! blocks in the example group together.
       end
     RUBY
+
+    expect_correction(<<-RUBY)
+      RSpec.describe User do
+        let(:a) { a }
+        let(:b) { b }
+        subject { User }
```
pirj, line 20: I'm not sure why we use `subject` in examples here, it should be before `let`s. Maybe use a one-line example instead?

spec/rubocop/cop/rspec/scattered_let_spec.rb
```diff
@@ -12,6 +12,76 @@
         ^^^^^^^^^^^^^ Group all let/let! blocks in the example group together.
       end
     RUBY
+
+    expect_correction(<<-RUBY)
+      RSpec.describe User do
```
pirj, line 17: Do we need to stress on `RSpec`?

spec/rubocop/cop/rspec/scattered_let_spec.rb
```diff
@@ -12,6 +12,76 @@
...
+        describe '#property' do
+          let(:c) { c }
+          let(:d) { d }
+
+          it { expect(subject.property).to eq c }
+
+        end
+      end
+    RUBY
   end
```
pirj, line 85: Just in case, maybe add an example for `let!`?

bquorning:
Possible conflict with #904 – or at least there’s an extra test case to autocorrect.

Darhazer:
Yup I'll update with the test case from #904 and with the improvements that @pirj suggested

Darhazer:
Updated

PR #905, opened 2020-04-29
Use rubocop-rspec's version while generating new cop

tejasbubane:
Default rake task for creating new cop adds rubocop's version to
config file, use rubocop-rspec's version instead.

Waiting for https://github.com/rubocop-hq/rubocop/pull/7897 to be merged.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -81,7 +81,14 @@ task :new_cop, [:cop] do |_task, args|
   generator.write_source
   generator.write_spec
   generator.inject_require(root_file_path: 'lib/rubocop/cop/rspec_cops.rb')
-  generator.inject_config(config_file_path: 'config/default.yml')
+  generator.inject_config(config_file_path: 'config/default.yml',
+                          version_added: bump_minor_version)
 
   puts generator.todo
+end
+
+def bump_minor_version
+  major, minor, _patch = RuboCop::RSpec::Version::STRING.split('.')
+
+  "#{major}.#{minor.succ}.0"
 end


Darhazer:
As an alternative, we can subclass the generator

pirj:
> As an alternative, we can subclass the generator

Ideally, this rake task would remain in the RuboCop itself, and no extensions should add any custom code to have stub generators out of the box. So I'm in favour of this solution.

pirj:
By the way, even if this change gets in the next RuboCop release, we don't have to bump our minimum supported version of it, it would be sufficient to `bundle update rubocop` locally.

Rakefile
```diff
@@ -81,7 +81,14 @@ task :new_cop, [:cop] do |_task, args|
...
+  generator.inject_config(config_file_path: 'config/default.yml',
+                          version_added: bump_minor_version)
 
   puts generator.todo
 end
+
+def bump_minor_version
+  versions = RuboCop::RSpec::Version::STRING.split('.')
+
+  "#{versions[0]}.#{versions[1].succ}"
```
bquorning, line 93: This will return e.g. `1.39` where I would expect `1.39.0`.

bquorning: We can also be a bit more specific with the local variable names, e.g.
```suggestion
  major, minor, _patch = RuboCop::RSpec::Version::STRING.split('.')

  "#{major}.#{minor.succ}.0"
```

tejasbubane:
@bquorning Fixed.

PR #907, opened 2020-05-01
Add new `RSpec/VariableName` cop

### LINKED ISSUE #807: Style/VariableName doesn't enforce styles on variables defined via `let`
The `Style/VariableName` rule isn't catching bad variable names (e.g. camelCase when snake_case is configured) for `let` variable definitions.

It's common in specs to define the variable using `let` and then it gets passed to methods as an argument.  Because of this bug in Rubocop, method arguments were not being enforced by `Style/VariableName` rule either: https://github.com/rubocop-hq/rubocop/issues/7275

Tested with Rubocop 0.74.0 and rubocop-rspec 1.35.0

## Expected behavior

When configured to use `snake_case`, the `Style/VariableName` rule should catch usage such as:
```
let(:badVariable) { true }
```

## Actual behavior

No Rubocop violations are found.

```
$ rubocop test.rb
Inspecting 1 file
.

```

test.rb
```
# frozen_string_literal: true

let(:badVariable) { true }

def print_name(user_name)
  puts user_name
end

print_name(badVariable)
```
`rubocop test.rb` (probably need RSpec for `let`)

## RuboCop version

```
$ rubocop -V
0.74.0 (using Parser 2.6.3.0, running on ruby 2.3.7 universal.x86_64-darwin18)
```

tejasbubane:
Closes #807

---

* [x] Added tests.
* [x] Updated documentation.


lib/rubocop/rspec/language.rb
```diff
@@ -124,6 +124,10 @@ module Subject
         ALL = SelectorSet.new(%i[subject subject!])
       end
 
+      module Variables
```
Darhazer, line 127: So it's basically `Helpers::ALL` + `Subject::ALL`
Not sure if we need to introduce it as a separate concept, but even if do, we can reuse the above

lib/rubocop/cop/rspec/variable_name.rb
```diff
@@ -0,0 +1,50 @@
...
+      # @example EnforcedStyle: camelCase
+      #   # bad
+      #   let(:user_name) { 'Adam' }
+      #   subject(:user_name) { 'Adam' }
+      #
+      #   # good
+      #   let(:userName) { 'Adam' }
+      #   subject(:userName) { 'Adam' }
+      class VariableName < Cop
+        include RuboCop::RSpec::Language::NodePattern
```
Darhazer, line 26: This is included in every Cop, you shouldn't need to include it manually

lib/rubocop/cop/rspec/variable_name.rb
```diff
@@ -0,0 +1,50 @@
...
+        include ConfigurableNaming
+
+        MSG = 'Use %<style>s for variable names.'
+
+        def_node_matcher :variable_definition?, <<~PATTERN
+          (send #{RSPEC} #{Variables::ALL.node_pattern_union}
+            ({sym str} $_) ...)
```
Darhazer, line 33: perhaps we need a cop to ensure let/subject are being used with a symbol and not a string?

lib/rubocop/cop/rspec/variable_name.rb
```diff
@@ -0,0 +1,50 @@
...
+        PATTERN
+
+        def on_send(node)
+          variable_definition?(node) do |variable_name|
+            check_name(node, variable_name, node.loc.expression)
```
Darhazer, line 38: perhaps we can highlight only the actual variable name? For this you will have to capture the actual sym/str node, and not only the value (but then it have a `.value` method that you can use), otherwise with the current code you need `node.first_argument.loc.expression`

lib/rubocop/cop/rspec/variable_name.rb
tejasbubane: I created an issue for this: https://github.com/rubocop-hq/rubocop-rspec/issues/908

tejasbubane:
@Darhazer I have done the requested changes.

bquorning:
@Darhazer Would you please verify that your concerns have been addressed, and then merge the PR?

PR #909, opened 2020-05-03
Support multiple matchers for Capybara/VisibilityMatcher cop

twalpole:
Expand the Capybara/VisibilityMatcher cop to match all relevant Capybara matcher methods rather than just `have_selector`

* [x] Added tests.
* [x] Updated documentation.

### DIFF
--- a/lib/rubocop/cop/rspec/capybara/visibility_matcher.rb
+++ b/lib/rubocop/cop/rspec/capybara/visibility_matcher.rb
@@ -18,34 +18,49 @@ module RuboCop
         #
         #   # bad
         #   expect(page).to have_selector('.foo', visible: false)
-        #
-        #   # bad
-        #   expect(page).to have_selector('.foo', visible: true)
-        #
-        #   # good
-        #   expect(page).to have_selector('.foo', visible: :all)
-        #
-        #   # good
-        #   expect(page).to have_selector('.foo', visible: :hidden)
+        #   expect(page).to have_css('.foo', visible: true)
+        #   expect(page).to have_link('my link', visible: false)
         #
         #   # good
         #   expect(page).to have_selector('.foo', visible: :visible)
+        #   expect(page).to have_css('.foo', visible: :all)
+        #   expect(page).to have_link('my link', visible: :hidden)
         #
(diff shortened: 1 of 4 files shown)


lib/rubocop/cop/rspec/capybara/visibility_matcher.rb
```diff
@@ -20,33 +20,53 @@ module Capybara
         #   expect(page).to have_selector('.foo', visible: false)
         #
         #   # bad
-        #   expect(page).to have_selector('.foo', visible: true)
+        #   expect(page).to have_css('.foo', visible: true)
         #
         #   # good
-        #   expect(page).to have_selector('.foo', visible: :all)
+        #   expect(page).to have_xpath('.//foo', visible: :all)
```
Darhazer, line 26: It's better to leave the good examples to be exactly how the bad ones are supposed to be. Otherwise, someone might think that the cop proposes have_xpath instead of have_css

spec/rubocop/cop/rspec/capybara/visibility_matcher_spec.rb
```diff
@@ -17,6 +17,13 @@
     RUBY
   end
 
+  it 'recognizes multiple matchers' do
+    expect_offense(<<-RUBY)
+      expect(page).to have_css('.my_element', visible: false)
```
Darhazer, line 22: perhaps you can list the rest in the same example

pirj:
@Darhazer do we have to change the `VersionUpdated` of the metadata?

Darhazer:
> @Darhazer do we have to change the VersionUpdated of the metadata?

Only when the change affects the config, e.g. if this list was configurable. Autocorrects affect the config as they can be enabled or disabled. But general cop updates do not. 
The idea is to provide the user with feedback that he can now configure the new values.

btw I'll try to update the cop metadata for the project next week, so we can finally adopt it. We then can add the reminded of VersionAdded/VersionUpdated in the PR template

pirj:
@bquorning Do you mind to glance over?

pirj:
I swear I've pushed fixup commits yesterday. GH seems to be messing around this workflow.

PR #912, opened 2020-05-19
Add SpecSuffixOnly option to RSpec/FilePath cop

zdennis:
Add `SpecSuffixOnly` option to `RSpec/FilePath` cop which ensures spec files end with the minimal "_spec.rb" suffix only. This allows the cop to only find offenses on test files that do not have the "_spec.rb".

### Why this is useful

We want to use this cop to help ensure that all spec files have `_spec.rb` endings, but we don't want to conform to the opinion that all spec files need to be based on the class/module name being described. 

This helps prevent tests from getting missed which can happen with cross-functional teams where it can be easy for folks who touch multiple languages and tools to accidentally type in different filename suffixes common in other languages/tools. This change helps guard these files being named incorrectly and not being run by the test suite or CI.

---

* [x] Added tests.
* [x] Updated documentation.

### DIFF
--- a/lib/rubocop/cop/rspec/file_path.rb
+++ b/lib/rubocop/cop/rspec/file_path.rb
@@ -3,10 +3,11 @@
 module RuboCop
   module Cop
     module RSpec
-      # Checks that spec file paths are consistent with the test subject.
+      # Checks that spec file paths are consistent and well-formed.
       #
-      # Checks the path of the spec file and enforces that it reflects the
-      # described class/module and its optionally called out method.
+      # By default, this checks that spec file paths are consistent with the
+      # test subject and and enforces that it reflects the described
+      # class/module and its optionally called out method.
       #
       # With the configuration option `IgnoreMethods` the called out method will
       # be ignored when determining the enforced path.
@@ -14,6 +15,10 @@ module RuboCop
       # With the configuration option `CustomTransform` modules or classes can
       # be specified that should not as usual be transformed from CamelCase to
       # snake_case (e.g. 'RuboCop' => 'rubocop' ).
+      #
+      # With the configuration option `SpecSuffixOnly` test files will only
+      # be checked to ensure they end in '_spec.rb'. This option disables
+      # checking for consistency in the test subject or test methods.
       #
       # @example
       #   # bad
@@ -33,6 +38,16 @@ module RuboCop
       #
       # @example when configuration is `IgnoreMethods: true`
       #   # bad
+      #   whatever_spec.rb         # describe MyClass
+      #
+      #   # good
+      #   my_class_spec.rb         # describe MyClass
+      #
+      #   # good
+      #   my_class_spec.rb         # describe MyClass, '#method'
+      #
+      # @example when configuration is `SpecSuffixOnly: true`
+      #   # good
       #   whatever_spec.rb         # describe MyClass
       #
       #   # good
@@ -70,7 +85,13 @@ module RuboCop
         end
 
         def glob_for((described_class, method_name))
+          return glob_for_spec_suffix_only? if spec_suffix_only?
+
           "#{expected_path(described_class)}#{name_glob(method_name)}*_spec.rb"
+        end
+
+        def glob_for_spec_suffix_only?
+          '*_spec.rb'
         end
 
         def name_glob(name)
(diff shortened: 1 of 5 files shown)


lib/rubocop/cop/rspec/file_path.rb
```diff
@@ -70,7 +70,13 @@ def routing_spec?(args)
         end
```
pirj, line 71: Specs are good, but do you mind to add some documentation on how it works as comments for this class?

lib/rubocop/cop/rspec/file_path.rb
```diff
@@ -70,7 +70,13 @@ def routing_spec?(args)
...
 
         def glob_for((described_class, method_name))
-          "#{expected_path(described_class)}#{name_glob(method_name)}*_spec.rb"
+          glob = []
+          unless spec_suffix_only?
+            glob << expected_path(described_class)
+            glob << name_glob(method_name)
+          end
+          glob << '*_spec.rb'
+          glob.join
```
pirj, line 79: This looks pretty unreadable to me.
Do you mind to split it into two methods? E.g.
```ruby
def glob_for((described_class, method_name))
  return glob_for_spec_suffix_only?(described_class, method_name) if spec_suffix_only?

  "#{expected_path(described_class)}#{name_glob(method_name)}*_spec.rb"
end

def glob_for_spec_suffix_only?(described_class, method_name)
  "*_spec.rb"
end
```

zdennis: I don't mind at all. I've updated and re-pushed this branch up.

lib/rubocop/cop/rspec/file_path.rb
zdennis: Sure thing. I've updated the documentation and have re-pushed this branch.

zdennis:
Thanks for the review @pirj. I've incorporated your feedback and pushed up the changes. If any thing else comes to mind don't hesitate to let me know.

@pirj wrote:

> Do you mind to share how do you use RSpec in terms of naming spec files?

80% ~ 90% of the  time spec files that I write are consistent with the described class/module that the gem uses.  However, I don't think I've ever used the convention where a file represents a test against a singular method or where I've included a method name in the file name.

For that last 10% ~ 20% it often involves patching Rails or some other library and for that I name organize the code like this: `dual_boot/<library>-<version>/<patch_concern>.rb`

So I might end up with `dual_boot/rails-6.0/active_record/transaction_callbacks.rb` where I'm fixing a regression locally in Rails that has to do with transaction callbacks and that code may or may not touch multiple methods, classes/modules, etc in ActiveRecord to resolve the issue. This convention helps encourage me (and others) to separate patches that fix different things rather than muddle them all together in a giant active_record/base.rb patch file. And, I like to have focused spec against any patches (helps me ensure that the functionality continues to work and lets me know when I can remove an old patch).

I'm not sure this is the best way about it, but it has been working fairly well as a convention for the past few years. 

HTH

bquorning:
> However, I don't think I've ever used the convention where a file represents a test against a singular method or where I've included a method name in the file name.

I think I’ve only used one file tests one method when testing Rails controllers that do *too* much :-)

> 80% ~ 90% of the time spec files that I write are consistent with the described class/module that the gem uses. […]
> 
> For that last 10% ~ 20% it often involves patching Rails or some other library and for that I name organize the code like this: `dual_boot/<library>-<version>/<patch_concern>.rb`

If you want to have the FilePath cop enforce the right rule for 100% of your files, you would need to have a separate .rubocop.yml file in each folder where those 10-20% of your files live, right? If the `SpecSuffixOnly` options is enabled globally, you lose the stricter check for the 80-90%.

zdennis:
Thanks for your review, @pirj! 🤗 

@bquorning wrote:

> If you want to have the FilePath cop enforce the right rule for 100% of your files, you would need to have a separate .rubocop.yml file in each folder where those 10-20% of your files live, right?

🤔 I've never thought to use per-directory `.rubocop.yml` files. That definitely seems like a good migration strategy for projects that are moving to rubocop-rspec.

> If the SpecSuffixOnly options is enabled globally, you lose the stricter check for the 80-90%.

Very true.

zdennis:
@pirj / @bquorning , is there anything I can do to help continue this along? It looks like the edge-rubocop build is failing on CircleCI, but after inspecting it appears to be failing for reasons unrelated to this PR.

pirj:
I have no objections to merge this.
`edge-rubocop` failure is unrelated.

PR #914, opened 2020-05-21
Fix specs for code transparently fixed with RuboCop 0.84 update

pirj:
It seems that RuboCop comes with certain fixes that improve how RuboCop RSpec itself works!

---

* [x] Added tests.
* [-] Updated documentation.

### DIFF
--- a/spec/rubocop/cop/rspec/verified_doubles_spec.rb
+++ b/spec/rubocop/cop/rspec/verified_doubles_spec.rb
@@ -55,25 +55,25 @@ RSpec.describe RuboCop::Cop::RSpec::VerifiedDoubles, :config do
     end
   end
 
-  it 'doubles that have no name specified' do
-    expect_offense(<<-RUBY)
-      it do
-        foo = double
-              ^^^^^^ Prefer using verifying doubles over normal doubles.
-      end
-    RUBY
-  end
+  context 'when configured not to ignore nameless doubles' do
+    let(:cop_config) { { 'IgnoreNameless' => false } }
 
-  context 'when configured to ignore nameless doubles' do
-    let(:cop_config) { { 'IgnoreNameless' => true } }
-
-    it 'ignores doubles that have no name specified' do
-      expect_no_offenses(<<-RUBY)
+    it 'flags doubles that have no name specified' do
+      expect_offense(<<-RUBY)
         it do
           foo = double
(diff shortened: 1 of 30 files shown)


spec/rubocop/cop/rspec/verified_doubles_spec.rb
```diff
@@ -55,27 +55,27 @@
     end
   end
 
-  it 'doubles that have no name specified' do
-    expect_offense(<<-RUBY)
-      it do
-        foo = double
-              ^^^^^^ Prefer using verifying doubles over normal doubles.
```
pirj, line 62: It's quite strange how it worked like that, since `IgnoreNameless: true` by default.
Nice, it's fixed now.

This gives me an interesting thought, since on the one hand, it's a bug fix, and on the other, it's a change in behaviour. But we're unable to "protect" our users from that change.

Darhazer: Perhaps `cop_config` was not merged with the default values?
Anyway, it's better to have the values explicitly set in specs and not rely on defaults

bquorning: Exactly. Previously – before https://github.com/rubocop-hq/rubocop/pull/7970 – the default configuration was _not_ present when testing. And it is now, so we can simplify a lot of our test.

I was too working on fixing our specs (and applied the _exact_ same changes to spec/rubocop/cop/rspec/verified_doubles_spec.rb) and have taken the liberty of adding a few commits to your branch. I hope that is ok with you @pirj

pirj: This is such a wonderful idea. I recall it was discussed before, and so good it's now in place.

> is ok with you

Of course, nice cleanup!

bquorning: I mention in the commit message of 4e7272b7e0eda702006b1022950cc111ca707295 that the change in RuboCop actually introduces a “mystery guest”, which could be considered a code smell.

pirj: Everything works and looks nice. Do you have something else in mind tbd here, or is it good to merge?

bquorning: I think we‘re good to merge.

PR #915, opened 2020-05-24
Cleanup workaround

pirj:
It seems that the workaround was exclusively used in the DescriptionExtractor spec.

Am I missing something?

* [-] Added tests.
* [-] Updated documentation.
* [x] ADDRESS FIXME notes

### DIFF
--- a/bin/build_config
+++ b/bin/build_config
@@ -9,9 +9,15 @@ require 'rubocop-rspec'
 require 'rubocop/rspec/description_extractor'
 require 'rubocop/rspec/config_formatter'
 
-glob = File.join(__dir__, '..', 'lib', 'rubocop', 'cop', 'rspec',
+glob = File.join('lib', 'rubocop', 'cop', 'rspec',
                  '{,capybara,factory_bot,rails}', '*.rb')
-YARD.parse(Dir[glob], [])
+# Due to YARD's sensitivity to file require order (as of 0.9.25),
+# we have to prepend the list with our base cop, RuboCop::Cop::RSpec::Cop.
+# Otherwise, cop's parent class for cops loaded before our base cop class
+# are detected as RuboCop::Cop::Cop, and that complicates the detection
+# of their relation with RuboCop RSpec.
+rspec_cop_path = File.join('lib', 'rubocop', 'cop', 'rspec', 'cop.rb')
+YARD.parse(Dir[glob].prepend(rspec_cop_path), [])
 
 descriptions = RuboCop::RSpec::DescriptionExtractor.new(YARD::Registry.all).to_h
 current_config = YAML.load_file('config/default.yml')
--- a/lib/rubocop/rspec/description_extractor.rb
+++ b/lib/rubocop/rspec/description_extractor.rb
@@ -21,7 +21,7 @@ module RuboCop
 
       # Decorator of a YARD code object for working with documented rspec cops
       class CodeObject
-        COP_CLASS_NAMES = %w[RuboCop::Cop RuboCop::Cop::RSpec::Cop].freeze
+        COP_CLASS_NAME = 'RuboCop::Cop::RSpec::Cop'
         RSPEC_NAMESPACE = 'RuboCop::Cop::RSpec'
 
         def initialize(yardoc)
@@ -68,9 +68,7 @@ module RuboCop
         end
 
         def cop_subclass?
-          # FIXME: simplify
-
-          COP_CLASS_NAMES.include?(yardoc.superclass.path)
+          yardoc.superclass.path == COP_CLASS_NAME
         end
 
         def abstract?
--- a/lib/rubocop/rspec/description_extractor.rb
+++ b/lib/rubocop/rspec/description_extractor.rb
@@ -68,10 +68,8 @@ module RuboCop
         end
 
         def cop_subclass?
-          # YARD superclass resolution is a bit flaky: All classes loaded before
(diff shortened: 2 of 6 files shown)


bquorning:
> It seems that the workaround was exclusively used in the DescriptionExtractor spec.

The `WorkaroundCop` was introduced in #234, and the `DescriptionExtractor` changes were added later on.

The comments say it’s a “Clone of the the normal `RuboCop::Cop::Cop` class so we can rewrite the `inherited` method without breaking functionality”.

That said, I’m also not completely sure whether we need the workaround :-)

pirj:
There's a tiny change TBD here, `RuboCop::Cop::RSpec::Cop` gets documented despite being abstract. I'll clamp this, there's a FIXME flag where this should be done. 

> rewrite the inherited method without breaking functionality

I have no idea what *was* the exact problem there.


```ruby
@registry = Registry.new
class << self
  attr_reader :registry
end

def self.inherited(base)
  registry.enlist(base)
end
```

fails out of the box, because `@registry` is per-class and is `nil` in our `Cop`, so we have to relay `enlist` to the original.

Semi-related: https://github.com/rubocop-hq/rubocop/pull/8023

lib/rubocop/rspec/description_extractor.rb
```diff
@@ -68,11 +68,7 @@ def documented_constant
         end
 
         def cop_subclass?
-          # YARD superclass resolution is a bit flaky: All classes loaded before
-          # RuboCop::Cop::WorkaroundCop are shown as having RuboCop::Cop as
-          # superclass, while all the following classes are listed as having
-          # RuboCop::Cop::RSpec::Cop as their superclass.
```
pirj, line 74: Now it's more or less clear. There's a better way - pass `RuboCop::Cop::RSpec::Cop`'s path as the first argument to `YARD.parse` to avoid its confusion.

pirj:
@bquorning @Darhazer Green!

tasks/cops_documentation.rake
```diff
@@ -172,30 +168,29 @@ task generate_cops_documentation: :yard_for_generate_documentation do
...
 
-  # rubocop:disable Metrics/AbcSize
-  # rubocop:disable Metrics/MethodLength
-  def print_cops_of_department(cops, department, config)
-    selected_cops = cops_of_department(cops, department).select do |cop|
-      cop.to_s.start_with?('RuboCop::Cop::RSpec')
+  def selected_cops(cops, department)
+    cops_of_department(cops, department.to_sym).select do |cop|
+      cop.name.start_with?('RuboCop::Cop::RSpec') &&
+        cop.name != 'RuboCop::Cop::RSpec::Cop'
```
pirj, line 174: Skip the base abstract cop from being added to the docs.

lib/rubocop/rspec/description_extractor.rb
```diff
@@ -68,11 +68,7 @@ def documented_constant
...
-          COP_CLASS_NAMES.include?(yardoc.superclass.path)
+          yardoc.superclass.path == COP_CLASS_NAME
```
pirj, line 71: We're passing our base cop's path to YARD first, so it detects the superclass properly, and doesn't erroneously report some of them to be `RuboCop::Cop`.

pirj:
Maybe we should drop `RuboCop::Cop::RSpec::Cop` altogether? The other extensions don't seem to have their own base cops.

pirj:
Ping

bquorning:
FYI I found https://github.com/rubocop-hq/rubocop/pull/3837 which changed the `inherited` hook, which was mentioned in #234 as the reason for `WorkaroundCop`.

bin/build_config
```diff
@@ -9,9 +9,10 @@ require 'rubocop-rspec'
 require 'rubocop/rspec/description_extractor'
 require 'rubocop/rspec/config_formatter'
 
-glob = File.join(__dir__, '..', 'lib', 'rubocop', 'cop', 'rspec',
+glob = File.join('lib', 'rubocop', 'cop', 'rspec',
                  '{,capybara,factory_bot,rails}', '*.rb')
-YARD.parse(Dir[glob], [])
+rspec_cop_path = File.join('lib', 'rubocop', 'cop', 'rspec', 'cop.rb')
+YARD.parse([rspec_cop_path, *Dir[glob]], [])
```
bquorning, line 15: Since `'{,capybara,factory_bot,rails}', '*.rb')` already covers `cop.rb`, perhaps it should be made more explicit that we _want_ the `cop.rb` entry to be first. Maybe by using `[]#prepend`, or by adding a comment?

spec/rubocop/rspec/description_extractor_spec.rb
```diff
@@ -42,15 +42,24 @@ def bar
...
-    stub_const(
-      "RuboCop::Cop::RSpec::#{name}",
-      Class.new(RuboCop::Cop.const_get(:WorkaroundCop))
-    )
+    stub_const("RuboCop::Cop::RSpec::#{name}", Class.new(temp_class))
   end
 
   before do
     stub_cop_const('Foo')
+    stub_const('RuboCop::Cop::RSpec::Foo::HelperClassForFoo', Class.new)
```
bquorning, line 62: Why is this necessary now?

pirj: Nice catch! Something was breaking in the process, but doesn't now.
Removing.

bquorning:
Thank you @pirj . It truly does look like this `Cop` superclass was not necessary anymore.

PR #917, opened 2020-05-26
Fix `RSpec/FilePath` detection across sibling directories.

rolfschmidt:
Hi,
i saw your pull request https://github.com/rubocop-hq/rubocop-rspec/pull/869 and we use some of our ruby projects as a kind of plugins for our main repository where we use the rubocop config from. So the relative path for the external projects seems to be wrong currently.

Would you be kind enough to take a look at it and help me to get my projects tidied :-D

---

* [x] Added tests.
* [x] Updated documentation.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 * Expand `Capybara/VisibilityMatcher` to support more than just `have_selector`. ([@twalpole][])
 * Add new `SpecSuffixOnly` option to `RSpec/FilePath` cop. ([@zdennis][])
 * Allow `RSpec/RepeatedExampleGroupBody` to differ only by described_class. ([@robotdana][])
+* Fix `RSpec/FilePath` detection across sibling directories. ([@rolfschmidt][])
 
 ## 1.39.0 (2020-05-01)
 
@@ -512,3 +513,4 @@ Compatibility release so users can upgrade RuboCop to 0.51.0. No new features.
 [@twalpole]: https://github.com/twalpole
 [@zdennis]: https://github.com/zdennis
 [@robotdana]: https://github.com/robotdana
+[@rolfschmidt]: https://github.com/rolfschmidt
--- a/lib/rubocop/cop/rspec/file_path.rb
+++ b/lib/rubocop/cop/rspec/file_path.rb
@@ -126,6 +126,7 @@ module RuboCop
         def filename_ends_with?(glob)
           filename =
             RuboCop::PathUtil.relative_path(processed_source.buffer.name)
+              .gsub('../', '')
           File.fnmatch?("*#{glob}", filename)
         end
 
--- a/spec/rubocop/cop/rspec/file_path_spec.rb
+++ b/spec/rubocop/cop/rspec/file_path_spec.rb
@@ -179,6 +179,24 @@ RSpec.describe RuboCop::Cop::RSpec::FilePath, :config do
     RUBY
   end
 
+  it 'uses relative path for sibling directory project' do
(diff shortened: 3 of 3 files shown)


pirj:
Sorry, I don't completely understand the directory structure. I'd love to manually test this. Can you please help to reproduce the issue you're experiencing?

rolfschmidt:
Hi @pirj ,
maybe this downloadable example will help you :-)

1. Download this

[example.zip](https://github.com/rubocop-hq/rubocop-rspec/files/4686974/example.zip)

2. Extract to /home

3. switch to project1
`cd /home/project1`

4. Tidy project2
`rubocop -a -c .rubocop.yml -D /home/project2`

5. get error

```
/home/project2/spec/lib/blub_spec.rb:3:1: C: RSpec/FilePath: Spec path should end with blub*_spec.rb.
RSpec.describe ::Blub do
``` 

does this maybe help you?

CHANGELOG.md
```diff
@@ -5,6 +5,7 @@
 * Add new `RSpec/VariableName` cop. ([@tejasbubane][])
 * Add new `RSpec/VariableDefinition` cop. ([@tejasbubane][])
 * Expand `Capybara/VisibilityMatcher` to support more than just `have_selector`. ([@twalpole][])
+* Followup-Fix `RSpec/FilePath` detection when absolute path includes test subject. ([@rolfschmidt][])
```
bquorning, line 8: Can we rephrase this? Maybe
```suggestion
* Fix `RSpec/FilePath` detection across sibling directories. ([@rolfschmidt][])
```

lib/rubocop/cop/rspec/file_path.rb
```diff
@@ -105,6 +105,7 @@ def ignore_methods?
         def filename_ends_with?(glob)
           filename =
             RuboCop::PathUtil.relative_path(processed_source.buffer.name)
+              .gsub('../', '')
```
bquorning, line 108: I was hoping to find a built-in command for doing this. I am not sure if this works on Windows, for example.

CHANGELOG.md
rolfschmidt: check

lib/rubocop/cop/rspec/file_path.rb
rolfschmidt: Hi @bquorning ,

thank you for you feedback. I was not able to get my example files working with a specific gem path in the Gemfile in windows. I dont understand that, but i was able to have x.rb which shows up that the path is the same on windows:

```ruby
require 'rubocop'
p RuboCop::PathUtil.relative_path('c:\ruby\project2\spec\lib\blub_spec.rb')
```

grafik

On my research on stackoverflow someone wrote this:

`One thing to note. Ruby uses a "/" for file separator on all platforms, including Windows, so you don't actually need use different code for joining things together on different platforms. "C:/tmp/1.text" should work fine.` (https://stackoverflow.com/a/597500/3005815)

So maybe we don't need a different solution for windows? Alternatively we could think about to use this static variable: `File::SEPARATOR`. 

What do you think?

bquorning: It is _probably_ fine. I know there is both `File::SEPARATOR`, `File::ALT_SEPARATOR` and `File::Separator` to take into account.

Let’s just merge it, and if anyone files a bug we’ll take if from there.

Could you just rebase the branch, then I think it’s ready to merge.

rolfschmidt: Thank you !!! 👍 check

PR #920, opened 2020-05-29
Allow identical body with distinct described_class

robotdana:
const args are used for described_class and thus are part of the
'metadata' of an example.

```
describe Array do
  it_behaves_like 'list'
end

describe Set do
  it_behaves_like 'list'
end
```

should not be considered duplicate tests

i wasn't sure how to extract the constant with all its namespaces
as a list of symbols, so i just grab the whole constant node

* [x] Added tests.
* [x] Updated documentation.

### DIFF
--- a/lib/rubocop/cop/rspec/repeated_example_group_body.rb
+++ b/lib/rubocop/cop/rspec/repeated_example_group_body.rb
@@ -34,6 +34,15 @@ module RuboCop
       #      it { cool_predicate }
       #    end
       #
+      #    # good
+      #    context Array do
+      #      it { is_expected.to respond_to :each }
+      #    end
+      #
+      #    context Hash do
+      #      it { is_expected.to respond_to :each }
+      #    end
+      #
       class RepeatedExampleGroupBody < Cop
         MSG = 'Repeated %<group>s block body on line(s) %<loc>s'
 
@@ -43,6 +52,7 @@ module RuboCop
 
         def_node_matcher :metadata, '(block (send _ _ _ $...) ...)'
         def_node_matcher :body, '(block _ args $...)'
+        def_node_matcher :const_arg, '(block (send _ _ $const ...) ...)'
 
         def_node_matcher :skip_or_pending?, <<-PATTERN
           (block <(send nil? {:skip :pending}) ...>)
@@ -75,7 +85,7 @@ module RuboCop
         end
 
         def signature_keys(group)
-          [metadata(group), body(group)]
+          [metadata(group), body(group), const_arg(group)]
         end
 
         def message(group, repeats)
--- a/spec/rubocop/cop/rspec/repeated_example_group_body_spec.rb
+++ b/spec/rubocop/cop/rspec/repeated_example_group_body_spec.rb
@@ -74,6 +74,44 @@ RSpec.describe RuboCop::Cop::RSpec::RepeatedExampleGroupBody do
     RUBY
   end
 
+  it 'does not register offense if module arg is different' do
+    expect_no_offenses(<<-RUBY)
+      describe CSV::Row do
+        it { is_expected.to respond_to :headers }
+      end
+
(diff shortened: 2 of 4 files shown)


spec/rubocop/cop/rspec/repeated_example_group_body_spec.rb
```diff
@@ -74,6 +74,70 @@
     RUBY
   end
 
+  it 'does not register offense if module arg is different' do
+    expect_no_offenses(<<-RUBY)
+      describe Array do
+        it { described_class.cool_predicate_method }
```
pirj, line 80: Same here - `described_class.cool_predicate_method` doesn't look like a proper expectation.

spec/rubocop/cop/rspec/repeated_example_group_body_spec.rb
```diff
@@ -74,6 +74,70 @@
...
+      end
+
+      context Array do
+      ^^^^^^^^^^^^^^^^ Repeated context block body on line(s) [1]
+        it { described_class.cool_predicate_method }
+      end
+    RUBY
+  end
+
+  it 'does not register offense if namespaced module arg is different' do
```
pirj, line 103: It might sound reckless, but the following examples look excessive to me.

lib/rubocop/cop/rspec/repeated_example_group_body.rb
```diff
@@ -34,6 +34,15 @@ module RSpec
       #      it { cool_predicate }
       #    end
       #
+      #    # good
+      #    context Array do
+      #      it { described_class.cool_predicate }
```
pirj, line 39: Maybe `expect(described_class).to something`?
It's not a proper one-liner though, and I understand your intention to fit a `described_class` here to indicate what is the reason for those two identical group bodies to be considered different.
Can you think of a good example to fulfill both of those requirements?

lib/rubocop/cop/rspec/repeated_example_group_body.rb
```diff
@@ -43,6 +52,7 @@ class RepeatedExampleGroupBody < Cop
 
         def_node_matcher :metadata, '(block (send _ _ _ $...) ...)'
         def_node_matcher :body, '(block _ args $...)'
+        def_node_matcher :const_arg, '(block (send _ _ $(const _ _) ...) ...)'
```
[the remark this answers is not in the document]
robotdana: yeah, i don't entirely understand how node patterns work, this was mostly created with trying stuff until it got what i wanted

spec/rubocop/cop/rspec/repeated_example_group_body_spec.rb
robotdana: yeah, i needed this when tdd-ing, because my initial solution didn't cover it, but i can probably reduce the tests down

robotdana: now there are just 3 tests: allow different module, allow different namespace for module, complain when identical body and module. all with real looking expectations and classes from stdlib

lib/rubocop/cop/rspec/repeated_example_group_body.rb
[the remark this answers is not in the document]
robotdana: I tried, i don't think it's possible, i don't think even combining the metadata and body matchers is possible because of the possibility of groups with no arg.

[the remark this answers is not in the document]
pirj: > I don't entirely understand how node patterns work, this was mostly created with trying stuff until it got what i wanted

I struggle with it every time as well, tried to remember how named captures work and ended up looking up in the documentation yesterday when playing with your pull request :D

> don't think even combining the metadata and body matchers is possible because of the possibility of groups with no arg.

You're right, I haven't thought about this 👍

PR #924, opened 2020-06-06
Fix leaky constant declaration in specs

pirj:
https://github.com/rubocop-hq/rubocop/pull/8095 allowed to dynamically declare cops with `Class.new`

We don't have to bump minimally required RuboCop version since it only affects our specs, and we always use the latest RuboCop for specs.

---

* [-] Added tests.
* [-] Updated documentation.


pirj:
I guess this will have to wait for https://github.com/rubocop-hq/rubocop/pull/8095 to be released as part of RuboCop 0.86

pirj:
RuboCop 0.85.1 has been released and made this green.

PR #925, opened 2020-06-07
RSpec/SubjectStub. Refactor and decrease complexity

andrykonchin:
`RSpec/SubjectStub` takes about 35% of all the time spent in cops if run Rubocop on Rubocop's  own source code. 

PR addresses the performance problem mentioned in the https://github.com/rubocop-hq/rubocop/issues/8022 issue.

### Profiler reports

Profiler report (`stackprof`) before optimizaition:

```
stackprof tmp/stackprof-cpu-rubocop.master.with-config.dump --method 'RuboCop::Cop::Commissioner#trigger_responding_cops'

RuboCop::Cop::Commissioner#trigger_responding_cops (/Users/andrykonchin/projects/rubocop/lib/rubocop/cop/commissioner.rb:50)
  samples:   756 self (2.8%)  /   17379 total (65.3%)
  callers:
    17379  (  100.0%)  block (2 levels) in <class:Commissioner>
    17332  (   99.7%)  RuboCop::Cop::Commissioner#trigger_responding_cops
    16630  (   95.7%)  RuboCop::Cop::Commissioner#with_cop_error_handling
  callees (16623 total):
    17332  (  104.3%)  RuboCop::Cop::Commissioner#trigger_responding_cops
    16652  (  100.2%)  RuboCop::Cop::Commissioner#with_cop_error_handling
    5934  (   35.7%)  RuboCop::Cop::RSpec::SubjectStub#on_block
    1998  (   12.0%)  RuboCop::Cop::Layout::SpaceAroundMethodCallOperator#on_send
     374  (    2.2%)  RuboCop::Cop::Layout::SpaceInsideArrayLiteralBrackets#on_array
...
```

Profiler report (`stackprof`) afeter optimizaition:

```
stackprof tmp/stackprof-cpu-rubocop.master.with-config.rspec-optimized.dump --method 'RuboCop::Cop::Commissioner#trigger_responding_cops'

RuboCop::Cop::Commissioner#trigger_responding_cops (/Users/andrykonchin/projects/rubocop/lib/rubocop/cop/commissioner.rb:50)
  samples:   851 self (3.6%)  /   13455 total (56.9%)
  callers:
    13455  (  100.0%)  block (2 levels) in <class:Commissioner>
    13398  (   99.6%)  RuboCop::Cop::Commissioner#trigger_responding_cops
    12598  (   93.6%)  RuboCop::Cop::Commissioner#with_cop_error_handling
  callees (12604 total):
    13398  (  106.3%)  RuboCop::Cop::Commissioner#trigger_responding_cops
    12626  (  100.2%)  RuboCop::Cop::Commissioner#with_cop_error_handling
    2143  (   17.0%)  RuboCop::Cop::Layout::SpaceAroundMethodCallOperator#on_send
     395  (    3.1%)  RuboCop::Cop::Layout::SpaceInsideArrayLiteralBrackets#on_array
     305  (    2.4%)  RuboCop::Cop::Layout::SpaceBeforeFirstArg#on_send
     279  (    2.2%)  RuboCop::Cop::Style::RedundantSelf#on_block
     266  (    2.1%)  RuboCop::RSpec::TopLevelDescribe#on_send
     226  (    1.8%)  RuboCop::Cop::Layout::SpaceInsideReferenceBrackets#on_send
     200  (    1.6%)  RuboCop::Cop::Style::TrailingMethodEndStatement#on_def
     200  (    1.6%)  RuboCop::Cop::Layout::FirstArgumentIndentation#on_send
     191  (    1.5%)  RuboCop::Cop::RSpec::FactoryBot::AttributeDefinedStatically#on_block
     175  (    1.4%)  RuboCop::Cop::Layout::SpaceAroundMethodCallOperator#on_const
     164  (    1.3%)  RuboCop::Cop::Layout::IndentationConsistency#on_begin
     136  (    1.1%)  RuboCop::Cop::Layout::SpaceInsideHashLiteralBraces#on_hash
     130  (    1.0%)  RuboCop::Cop::Layout::DefEndAlignment#on_send
     113  (    0.9%)  RuboCop::Cop::Interpolation#on_dstr
     109  (    0.9%)  RuboCop::Cop::RSpec::SubjectStub#on_block
...
```

Samples number decreased from 65.3% to 56.9% and the cop samples (relative to `RuboCop::Cop::Commissioner#trigger_responding_cops` method execution) decreased from 35.7% to  0.9%.

### Total execution time

Before:
```
time bundle exec exe/rubocop --cache false --out rubocop.out .
bundle exec exe/rubocop --cache false --out rubocop.out .  83.46s user 0.40s system 99% cpu 1:24.40 total
```

After:
```
time bundle exec exe/rubocop --cache false --out rubocop.out .
bundle exec exe/rubocop --cache false --out rubocop.out .  62.60s user 0.35s system 99% cpu 1:03.21 total
```

Total execution time decreased from 1:24.40 to 1:03.21.

### Rubocop version

In performance tests was used the following Rubocop master commit:

```
ommit 7bafa099902871d9e6dc91800390a5c32e8d086b (HEAD -> master)
Author: Bozhidar Batsov <bozhidar@batsov.com>
Date:   Sat May 30 14:05:28 2020 +0300

    [Docs] Fix a couple of references to the docs folder
```

---

* [x] Added tests.
* [ ] Updated documentation.

### DIFF
--- a/lib/rubocop/cop/rspec/subject_stub.rb
+++ b/lib/rubocop/cop/rspec/subject_stub.rb
@@ -22,8 +22,6 @@ module RuboCop
       #   end
       #
       class SubjectStub < Cop
-        include RuboCop::RSpec::TopLevelDescribe
-
         MSG = 'Do not stub methods of the object under test.'
 
         # @!method subject(node)
@@ -77,15 +75,23 @@ module RuboCop
             } ...)
         PATTERN
 
-        def on_top_level_describe(node, _args)
-          @explicit_subjects = find_all_explicit_subjects(node.parent)
+        def on_block(node)
+          return unless example_group?(node)
+          return unless (processed_example_groups & node.ancestors).empty?
 
-          find_subject_expectations(node.parent) do |stub|
+          processed_example_groups << node
+          @explicit_subjects = find_all_explicit_subjects(node)
+
+          find_subject_expectations(node) do |stub|
             add_offense(stub)
           end
         end
 
         private
+
+        def processed_example_groups
+          @processed_example_groups ||= Set.new
+        end
 
         def find_all_explicit_subjects(node)
           node.each_descendant(:block).each_with_object({}) do |child, h|
--- a/lib/rubocop/cop/rspec/subject_stub.rb
+++ b/lib/rubocop/cop/rspec/subject_stub.rb
@@ -22,6 +22,8 @@ module RuboCop
       #   end
       #
       class SubjectStub < Cop
+        include RuboCop::RSpec::TopLevelDescribe
+
         MSG = 'Do not stub methods of the object under test.'
 
         # @!method subject(node)
@@ -75,23 +77,15 @@ module RuboCop
             } ...)
         PATTERN
 
-        def on_block(node)
-          return unless example_group?(node)
-          return unless (processed_example_groups & node.ancestors).empty?
+        def on_top_level_describe(node, _args)
+          @explicit_subjects = find_all_explicit_subjects(node.parent)
 
-          processed_example_groups << node
-          @explicit_subjects = find_all_explicit_subjects(node)
-
-          find_subject_expectations(node) do |stub|
+          find_subject_expectations(node.parent) do |stub|
             add_offense(stub)
           end
         end
 
         private
-
-        def processed_example_groups
-          @processed_example_groups ||= Set.new
-        end
 
         def find_all_explicit_subjects(node)
           node.each_descendant(:block).each_with_object({}) do |child, h|
--- a/lib/rubocop/cop/rspec/subject_stub.rb
+++ b/lib/rubocop/cop/rspec/subject_stub.rb
@@ -102,14 +102,10 @@ module RuboCop
         end
 
         def find_all_named_subjects(node)
-          named_subjects = {}
-
-          node.each_descendant(:block) do |child|
+          node.each_descendant(:block).each_with_object({}) do |child, h|
             name = subject(child)
-            named_subjects[child.parent.parent] = name if name
+            h[child.parent.parent] = name if name
           end
-
-          named_subjects
         end
 
         def find_subject_expectations(node, subject_name = nil, &block)
--- a/lib/rubocop/cop/rspec/subject_stub.rb
+++ b/lib/rubocop/cop/rspec/subject_stub.rb
@@ -96,17 +96,19 @@ module RuboCop
         def find_all_explicit_subjects(node)
           node.each_descendant(:block).each_with_object({}) do |child, h|
             name = subject(child)
-            if name
-              h[child.parent.parent] ||= []
-              h[child.parent.parent] << name
+            next unless name
+
+            outer_example_group = child.each_ancestor.find do |a|
+              example_group?(a)
             end
+
+            h[outer_example_group] ||= []
+            h[outer_example_group] << name
           end
         end
 
         def find_subject_expectations(node, subject_names = [], &block)
-          if example_group?(node) && @explicit_subjects[node]
-            subject_names = @explicit_subjects[node]
-          end
+          subject_names = @explicit_subjects[node] if @explicit_subjects[node]
 
           expectation_detected = (subject_names + [:subject]).any? do |name|
             message_expectation?(node, name)
--- a/lib/rubocop/cop/rspec/subject_stub.rb
+++ b/lib/rubocop/cop/rspec/subject_stub.rb
@@ -76,20 +76,12 @@ module RuboCop
         PATTERN
 
         def on_block(node)
-          # process only describe/context/... example groups
           return unless example_group?(node)
-
-          # skip already processed example group
-          # it's processed if is nested in one of the processed example groups
           return unless (processed_example_groups & node.ancestors).empty?
 
-          # add example group to already processed
           processed_example_groups << node
+          @explicit_subjects = find_all_explicit_subjects(node)
 
-          # find all custom subjects e.g. subject(:foo) { ... }
-          @named_subjects = find_all_named_subjects(node)
-
-          # look for method expectation matcher
           find_subject_expectations(node) do |stub|
             add_offense(stub)
           end
@@ -98,32 +90,31 @@ module RuboCop
         private
 
         def processed_example_groups
-          @processed_example_groups ||= Set[]
+          @processed_example_groups ||= Set.new
         end
 
-        def find_all_named_subjects(node)
+        def find_all_explicit_subjects(node)
           node.each_descendant(:block).each_with_object({}) do |child, h|
             name = subject(child)
-            h[child.parent.parent] = name if name
+            if name
+              h[child.parent.parent] ||= []
+              h[child.parent.parent] << name
+            end
           end
         end
 
-        def find_subject_expectations(node, subject_name = nil, &block)
-          # if it's a new example group - check whether new named subject is
-          # defined there
-          if example_group?(node)
-            subject_name = @named_subjects[node] || subject_name
+        def find_subject_expectations(node, subject_names = [], &block)
+          if example_group?(node) && @explicit_subjects[node]
+            subject_names = @explicit_subjects[node]
           end
 
-          # check default :subject and then named one (if it's present)
-          expectation_detected = message_expectation?(node, :subject) || \
-            (subject_name && message_expectation?(node, subject_name))
-
+          expectation_detected = (subject_names + [:subject]).any? do |name|
+            message_expectation?(node, name)
+          end
           return yield(node) if expectation_detected
 
-          # Recurse through node's children looking for a message expectation.
           node.each_child_node do |child|
-            find_subject_expectations(child, subject_name, &block)
+            find_subject_expectations(child, subject_names, &block)
           end
         end
       end
--- a/lib/rubocop/cop/rspec/subject_stub.rb
+++ b/lib/rubocop/cop/rspec/subject_stub.rb
@@ -1,5 +1,7 @@
 # frozen_string_literal: true
 
+require 'set'
+
 module RuboCop
   module Cop
     module RSpec
@@ -74,71 +76,58 @@ module RuboCop
         PATTERN
 
         def on_block(node)
+          # process only describe/context/... example groups
           return unless example_group?(node)
 
-          find_subject_stub(node) do |stub|
+          # skip already processed example group
+          # it's processed if is nested in one of the processed example groups
+          return unless (processed_example_groups & node.ancestors).empty?
+
+          # add example group to already processed
+          processed_example_groups << node
+
+          # find all custom subjects e.g. subject(:foo) { ... }
+          @named_subjects = find_all_named_subjects(node)
+
+          # look for method expectation matcher
+          find_subject_expectations(node) do |stub|
             add_offense(stub)
           end
         end
 
         private
 
-        # Find subjects within tree and then find (send) nodes for that subject
-        #
-        # @param node [RuboCop::Node] example group
-        #
-        # @yield [RuboCop::Node] message expectations for subject
-        def find_subject_stub(node, &block)
-          find_subject(node) do |subject_name, context|
-            find_subject_expectation(context, subject_name, &block)
-          end
+        def processed_example_groups
+          @processed_example_groups ||= Set[]
         end
 
-        # Find a subject message expectation
-        #
-        # @param node [RuboCop::Node]
-        # @param subject_name [Symbol] name of subject
-        #
-        # @yield [RuboCop::Node] message expectation
-        def find_subject_expectation(node, subject_name, &block)
-          # Do not search node if it is an example group with its own subject.
-          return if example_group?(node) && redefines_subject?(node)
+        def find_all_named_subjects(node)
+          named_subjects = {}
 
-          # Yield the current node if it is a message expectation.
-          yield(node) if message_expectation?(node, subject_name)
+          node.each_descendant(:block) do |child|
+            name = subject(child)
+            named_subjects[child.parent.parent] = name if name
+          end
+
+          named_subjects
+        end
+
+        def find_subject_expectations(node, subject_name = nil, &block)
+          # if it's a new example group - check whether new named subject is
+          # defined there
+          if example_group?(node)
+            subject_name = @named_subjects[node] || subject_name
+          end
+
+          # check default :subject and then named one (if it's present)
+          expectation_detected = message_expectation?(node, :subject) || \
+            (subject_name && message_expectation?(node, subject_name))
+
+          return yield(node) if expectation_detected
 
           # Recurse through node's children looking for a message expectation.
           node.each_child_node do |child|
-            find_subject_expectation(child, subject_name, &block)
-          end
-        end
-
-        # Check if node's children contain a subject definition
-        #
-        # @param node [RuboCop::Node]
-        #
-        # @return [Boolean]
-        def redefines_subject?(node)
-          node.each_child_node.any? do |child|
-            subject(child) || redefines_subject?(child)
-          end
-        end
-
-        # Find a subject definition
-        #
-        # @param node [RuboCop::Node]
-        # @param parent [RuboCop::Node,nil]
-        #
-        # @yieldparam subject_name [Symbol] name of subject being defined
-        # @yieldparam parent [RuboCop::Node] parent of subject definition
-        def find_subject(node, parent: nil, &block)
-          # An implicit subject is defined by RSpec when no subject is declared
-          subject_name = subject(node) || :subject
-
-          yield(subject_name, parent) if parent
-
-          node.each_child_node do |child|
-            find_subject(child, parent: node, &block)
+            find_subject_expectations(child, subject_name, &block)
           end
         end
       end
--- a/spec/rubocop/cop/rspec/subject_stub_spec.rb
+++ b/spec/rubocop/cop/rspec/subject_stub_spec.rb
@@ -21,7 +21,7 @@ RSpec.describe RuboCop::Cop::RSpec::SubjectStub do
   end
 
   it 'flags when subject is stubbed and there are several named subjects ' \
-     'in the same example group', :wip do
+     'in the same example group' do
     expect_offense(<<-RUBY)
       describe Foo do
         subject(:foo) { described_class.new }
--- a/spec/rubocop/cop/rspec/subject_stub_spec.rb
+++ b/spec/rubocop/cop/rspec/subject_stub_spec.rb
@@ -20,6 +20,26 @@ RSpec.describe RuboCop::Cop::RSpec::SubjectStub do
     RUBY
   end
 
+  it 'flags when subject is stubbed and there are several named subjects ' \
+     'in the same example group', :wip do
+    expect_offense(<<-RUBY)
+      describe Foo do
+        subject(:foo) { described_class.new }
+        subject(:bar) { described_class.new }
+        subject(:baz) { described_class.new }
+
+        before do
+          allow(bar).to receive(:bar).and_return(baz)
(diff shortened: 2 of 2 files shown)


lib/rubocop/cop/rspec/subject_stub.rb
```diff
@@ -74,71 +75,59 @@ class SubjectStub < Cop
         PATTERN
 
         def on_block(node)
+          # process only describe/context/... example groups
```
pirj, line 78: Usually, the code is pretty self-explanatory.
If you wish to clarify what's going on where, it's ok to put comments here for the reviewers.

lib/rubocop/cop/rspec/subject_stub.rb
```diff
@@ -74,71 +75,59 @@ class SubjectStub < Cop
...
-        #
-        # @param node [RuboCop::Node] example group
-        #
-        # @yield [RuboCop::Node] message expectations for subject
-        def find_subject_stub(node, &block)
-          find_subject(node) do |subject_name, context|
-            find_subject_expectation(context, subject_name, &block)
-          end
+        def processed_example_groups
+          @processed_example_groups ||= Set[]
```
pirj, line 100: Is there a difference with `Set.new`?

lib/rubocop/cop/rspec/subject_stub.rb
```diff
@@ -74,71 +75,59 @@ class SubjectStub < Cop
         PATTERN
 
         def on_block(node)
+          # process only describe/context/... example groups
           return unless example_group?(node)
 
-          find_subject_stub(node) do |stub|
+          # skip already processed example group
+          # it's processed if is nested in one of the processed example groups
+          return unless (processed_example_groups & node.ancestors).empty?
```
pirj, line 83: How big are your specs so this makes a significant difference?

lib/rubocop/cop/rspec/subject_stub.rb
```diff
@@ -74,71 +75,59 @@ class SubjectStub < Cop
...
-
-          # Yield the current node if it is a message expectation.
-          yield(node) if message_expectation?(node, subject_name)
+        def find_all_named_subjects(node)
+          named_subjects = {}
 
-          # Recurse through node's children looking for a message expectation.
-          node.each_child_node do |child|
-            find_subject_expectation(child, subject_name, &block)
+          node.each_descendant(:block) do |node|
```
pirj, line 106: Variable shadowing.

lib/rubocop/cop/rspec/subject_stub.rb
```diff
@@ -74,71 +75,59 @@ class SubjectStub < Cop
...
-        # Check if node's children contain a subject definition
-        #
-        # @param node [RuboCop::Node]
-        #
-        # @return [Boolean]
-        def redefines_subject?(node)
-          node.each_child_node.any? do |child|
-            subject(child) || redefines_subject?(child)
-          end
+          named_subjects
```
pirj, line 112: Is it possible to write this in a more functional style?

pirj:
I guess if you rebase your code, the build will pass. Fix for those has been merged https://github.com/rubocop-hq/rubocop-rspec/pull/914

lib/rubocop/cop/rspec/subject_stub.rb
andrykonchin: @pirj So would you recommend to remove all the added comments? (I updated the PR a little to pass Rubocop checks)

lib/rubocop/cop/rspec/subject_stub.rb
andrykonchin: I believe there is no any difference.This form is just shorter.

lib/rubocop/cop/rspec/subject_stub.rb
andrykonchin: I ran Rubocop on its own source code so the test suite isn't large. I believe the difference would be more significant on larger test code base.

lib/rubocop/cop/rspec/subject_stub.rb
```diff
@@ -74,71 +76,54 @@ class SubjectStub < Cop
         PATTERN
 
         def on_block(node)
+          # process only describe/context/... example groups
           return unless example_group?(node)
 
-          find_subject_stub(node) do |stub|
+          # skip already processed example group
+          # it's processed if is nested in one of the processed example groups
+          return unless (processed_example_groups & node.ancestors).empty?
```
pirj, line 84: Is this the same as `return if processed_example_groups.include?(node.ancestors)`?

lib/rubocop/cop/rspec/subject_stub.rb
```diff
@@ -74,71 +76,54 @@ class SubjectStub < Cop
...
 
-        # Check if node's children contain a subject definition
-        #
-        # @param node [RuboCop::Node]
-        #
-        # @return [Boolean]
-        def redefines_subject?(node)
-          node.each_child_node.any? do |child|
-            subject(child) || redefines_subject?(child)
+        def find_subject_expectations(node, subject_name = nil, &block)
```
pirj, line 111: I find this code hard to reason about.
Did you remove `redefines_subject?` for the sake of performance?
Can't recall if we have some benchmarks for cops here. I agree with you that cops, especially those using `def_node_search` should be benchmarked. I'm guilty of making changes to certain cops that improve their correctness by sacrificing performance, but I was doing this blindfolded without knowing how much the performance suffers exactly.

lib/rubocop/cop/rspec/subject_stub.rb
```diff
@@ -74,71 +76,54 @@ class SubjectStub < Cop
...
-        # @yieldparam parent [RuboCop::Node] parent of subject definition
-        def find_subject(node, parent: nil, &block)
-          # An implicit subject is defined by RSpec when no subject is declared
-          subject_name = subject(node) || :subject
+          # check default :subject and then named one (if it's present)
+          expectation_detected = message_expectation?(node, :subject) || \
+            (subject_name && message_expectation?(node, subject_name))
 
-          yield(subject_name, parent) if parent
+          return yield(node) if expectation_detected
```
pirj, line 122: There could be several subjects defined if I'm not mistaken:
```ruby
subject { ... } # unnamed
subject(:cop) { ... }
subject(:code) { ... }
subject(:performance) { ... }
```
in the same context. Are we handling this situation, do we detect attempts to stub all of them?

lib/rubocop/cop/rspec/subject_stub.rb
pirj: Maybe use the more commonly used (at least from my standpoint, I've never seen `Set[]` so far) to avoid confusion?

lib/rubocop/cop/rspec/subject_stub.rb
```diff
@@ -74,71 +76,54 @@ class SubjectStub < Cop
...
+          # process only describe/context/... example groups
           return unless example_group?(node)
 
-          find_subject_stub(node) do |stub|
+          # skip already processed example group
+          # it's processed if is nested in one of the processed example groups
+          return unless (processed_example_groups & node.ancestors).empty?
+
+          # add example group to already processed
+          processed_example_groups << node
```
pirj, line 87: Is there any evidence that `on_block` can be called more than once for the same example group?

lib/rubocop/cop/rspec/subject_stub.rb
pirj: Regarding the performance testing in the wild - you can pick some projects out of [this list](https://github.com/eliotsykes/real-world-rails), I find https://github.com/discourse/discourse/ specifically good.

pirj:
Oh, sorry, I've completely missed your update in the PR description.
> Total execution time decreased from 1:24.40 to 1:03.21.

This is truly impressive 👍 
And it seems to be an incredibly slow cop.

lib/rubocop/cop/rspec/subject_stub.rb
andrykonchin: Checked this optimization on Gitlab source code and got the following results:
- it speeds up the cope from 13%  to 5.8% of samples
- the whole execution time decreased by ~10s

lib/rubocop/cop/rspec/subject_stub.rb
andrykonchin: No, it isn't

lib/rubocop/cop/rspec/subject_stub.rb
andrykonchin: > Did you remove redefines_subject? for the sake of performance?

I just refactored the implementation. The case when subject is redefined is handled in another way.

lib/rubocop/cop/rspec/subject_stub.rb
andrykonchin: Good point. Didn't think about this case.

lib/rubocop/cop/rspec/subject_stub.rb
andrykonchin: I believe no. The goal of tracking nodes is to prevent handling of blocks nested into already processed outer blocks.

pirj: 👍

lib/rubocop/cop/rspec/subject_stub.rb
andrykonchin: The issue with comments is that the code seems obvious when you spends a lot of time thinking about it.

Okey, I've removed all the comments.

lib/rubocop/cop/rspec/subject_stub.rb
andrykonchin: It should be fixed now

lib/rubocop/cop/rspec/subject_stub.rb
andrykonchin: Fixed

andrykonchin:
@pirj Fixed all the issues. Please let me know to squash all the "code review" commits.

lib/rubocop/cop/rspec/subject_stub.rb
Darhazer: You can use `on_top_level describe` and then search for example groups, in order to avoid double lookups. E.g. once you are in an example group, you can break from going to nested example groups,

pirj: @Darhazer This is a truly amazing proposal.

Darhazer: But then it turns out that RuboCop::RSpec::TopLevelDescribe#on_send which triggers on_top_level_descrive, is one of [the slowest methods](https://github.com/rubocop-hq/rubocop/issues/8022)

pirj: Ha, fair enough. Never looked closely at `TopLevelDescribe`. Frankly, I'm having a hard time understanding why it's implemented that way. Should be: "call `on_top_level_describe` if method name is `:describe`, and there are no ancestors of the type `:block` with a method name `:describe`". It'd better keep a set of already detected TLDs to quickly return from `on_send` if `(node.ancestors & top_level_describes).any?` just like Andrew has implemented in this pull request.

Luckily, it doesn't fit our purposes in this pull request anyway, since it's only for `describe`:
```ruby
return false unless node.method_name == :describe
```
while we need something like `TopLevelExampleGroup`, probably implemented in a way I suggest above, but focusing on all example group methods, not only `:describe`.

Would you like to give this idea a shot in the scope of this pull request, or do you feel this improvement that you've done is good on its own already, @andrykonchin?

pirj:
I have never thought we'll have to optimize anytime soon, but it is happening :D
Probably it's for the best.

andrykonchin:
Addressed the issue with `#on_top_leve_describe` callback.

lib/rubocop/cop/rspec/subject_stub.rb
andrykonchin: I would love to optimize `TopLevelDescribe` but think it deserves a separate PR.

Currently I am working on optimizing cops from the main `rubocop` gem but will definitely look at `rubocop-rspec` after that. 

P.S.

> But then it turns out that RuboCop::RSpec::TopLevelDescribe#on_send which triggers on_top_level_descrive, is one of the slowest methods

Switching from manual caching to `#on_top_level_describe` callback decreased profiler samples for `SubjectStub` cop from 0.9% to 0% so it isn't as bad as you may think 😉.

pirj:
But it only detects subject stubbing inside a `describe`, does it?
On my current project, we use a top-level `context`.

lib/rubocop/cop/rspec/subject_stub.rb
```diff
@@ -73,72 +77,38 @@ class SubjectStub < Cop
             } ...)
         PATTERN
 
-        def on_block(node)
-          return unless example_group?(node)
+        def on_top_level_describe(node, _args)
+          @explicit_subjects = find_all_explicit_subjects(node.parent)
```
Darhazer, line 81: why you would start from node.parent? You may end up getting a subject from another top-level describe

lib/rubocop/cop/rspec/subject_stub.rb
```diff
@@ -73,72 +77,38 @@ class SubjectStub < Cop
...
-        #
-        # @yield [RuboCop::Node] message expectations for subject
-        def find_subject_stub(node, &block)
-          find_subject(node) do |subject_name, context|
-            find_subject_expectation(context, subject_name, &block)
+        def find_all_explicit_subjects(node)
+          node.each_descendant(:block).each_with_object({}) do |child, h|
+            name = subject(child)
+            if name
+              h[child.parent.parent] ||= []
```
Darhazer, line 94: what's the meaning of `parent.parent`? isn't it node.parent? while you are attaching to the parent?

lib/rubocop/cop/rspec/subject_stub.rb
```diff
@@ -73,72 +77,38 @@ class SubjectStub < Cop
...
-          return if example_group?(node) && redefines_subject?(node)
-
-          # Yield the current node if it is a message expectation.
-          yield(node) if message_expectation?(node, subject_name)
-
-          # Recurse through node's children looking for a message expectation.
-          node.each_child_node do |child|
-            find_subject_expectation(child, subject_name, &block)
+        def find_subject_expectations(node, subject_names = [], &block)
+          if example_group?(node) && @explicit_subjects[node]
```
Darhazer, line 101: well I guess you can just check if there are explicit_subjects for this node, as if it is not an example group, there won't be any anyway. That could simplify the code

Darhazer:
> But it only detects subject stubbing inside a describe, does it?
On my current project, we use a top-level context.

Maybe you got the definition of top-level describe wrong. Isn't your spec like:
```
describe Something do
  context '...' do
     ...
  end
end
```

As that is a top-level describe, and I never saw RSpec tests that begin directly with a context. If there are, other cops may not work as well

andrykonchin:
> pirj
> But it only detects subject stubbing inside a describe, does it?
> On my current project, we use a top-level context.

Agree with @pirj. There are may be top-level shared examples etc... So will revert the last commit

lib/rubocop/cop/rspec/subject_stub.rb
andrykonchin: I believe I will not. `node` here is `send` and parent is `block` in which this send is wrapped.

```ruby
s(:block,
  s(:send, nil, :describe,
    s(:const, nil, :Foo)),
  s(:args),
  s(:begin,
    s(:block,
      s(:send, nil, :subject,
        s(:sym, :foo)),
      s(:args),
      s(:send,
        s(:send, nil, :described_class), :new)),
...
```

lib/rubocop/cop/rspec/subject_stub.rb
andrykonchin: parent should be a `block` node for outer `describe`/`context`/...

lib/rubocop/cop/rspec/subject_stub.rb
Darhazer: I see. I forgot top_level_describe looks for a send node. It's weird to me

pirj:
> Maybe you got the definition of top-level describe wrong.

https://relishapp.com/rspec/rspec-core/docs/example-groups/basic-structure-describe-it surely hints that the basic structure is:
```ruby
RSpec.describe "something" do
  context "in one context" do
    it "does one thing" do
    end
  end
end
```

But how about:
```ruby
RSpec.shared_examples_for "it sabotages SubjectStub" do
  subject { haha }
  specify do
    expect(subject).to receive(:yes).and_return("no")
  end
end
```
We should detect this as well.
Is it a top-level example group? It doesn't really matter.

What about `scenario`, `fdescribe`, `xdescribe`? Why would we skip detection there?

> As that is a top-level describe, and I never saw RSpec tests that begin directly with a context.

I've never seen it before as well. But it is how it is.

pirj:
My point is not to limit `SubjectStub` to just `describe`, it's not how it worked before I believe.
A quick check shows that previously `SubjectStub` was detecting offences inside top-level `xdescribe`, `fdescribe` and `feature`.

Here's the default list. https://github.com/rspec/rspec-core/blob/2b913e1e36904227e3ecc8b4af0b7c622d3cb50d/lib/rspec/core/example_group.rb#L280
https://github.com/rspec/rspec-rails/blob/a9f90932d1939502d1024db12d99f5d413091bd4/lib/rspec/rails/example/feature_example_group.rb#L49 leaving us with:
 - example_group
 - describe
 - xdescribe
 - fdescribe
 - context
 - xcontext
 - fcontext
 - feature

For some of the cops, I don't see a good reason though to leave out those https://github.com/rspec/rspec-core/blob/ddbb43674be2a02c6cc839eba15f58265abcd8ca/lib/rspec/core/shared_example_group.rb#L100:
 - shared_examples
 - shared_examples_for
 - shared_context

pirj:
Please don't think that I'm defending a top-level `RSpec.context` :D I'm totally against it, as it violates https://rspec.rubystyle.guide/#context-cases at least.

pirj:
@andrykonchin sorry for dragging you through those inconsistencies. Please let us know how far you want to go with this.

lib/rubocop/cop/rspec/subject_stub.rb
pirj: Will it work properly for contexts defined in iterators? (yes, another funny concept from my current project)
```ruby
RSpec.describe do
  %i[one two].each do |count|
    context "for #{count}" do
      specify do
        expect(subject).to receive(count).and_return(0)
```
There's a parent block, but not an example group block.
I'm not sure this is the example to fail, hope you can figure out a better example to fail `.parent.parent`.

pirj:
> Is there any similar helper to detect any top-level example group? Not only describe?

Not that I'm aware of.

>  Does it make sense to add it?

I strongly believe so.

WDYT of:

>  or it's OK to keep manual caching

1. reverting the change involving `TopLevelDescribe`, and merging this as it was before

2. in another PR, adding a `TopLevelGroup` that would detect all, or several, depending on current usages of `TopLevelDescribe`. We'll have to keep `TopLevelDescribe` around since there can be custom cops using it. Maybe add a deprecation message?
2.1 implementing `TopLevelGroup` with performance in mind

andrykonchin:
@pirj Agree with adding `TopLevelGroup` in separate PR. Will update this PR today a little bit later.

andrykonchin:
Fixed all the mentioned issues. Please let me know to squash commits.

bquorning:
Running `rubocop --cache=false --only RSpec/SubjectStub -- spec` to check about 1400 spec files. It took 73.19s before your change, 16.24s after. 👏

pirj:
Thanks a lot @andrykonchin !

bquorning:
I think this change warrants a mention in the Changelog.

PR #930, opened 2020-06-11
Use node pattern generic parameters to simplify a cop

pirj:
https://github.com/rubocop-hq/rubocop-rspec/pull/804#discussion_r438747249

Support introduced in https://github.com/rubocop-hq/rubocop-ast/pull/31

Presumably generic parameters support will land in RuboCop 1.0 or 1.0.1 worst case.

---

* [-] Added tests.
* [-] Updated documentation.

### DIFF
--- a/lib/rubocop/cop/rspec/multiple_expectations.rb
+++ b/lib/rubocop/cop/rspec/multiple_expectations.rb
@@ -50,17 +50,13 @@ module RuboCop
 
         MSG = 'Example has too many expectations [%<total>d/%<max>d].'
 
+        ANYTHING = ->(_node) { true }
+        TRUE = ->(node) { node.true_type? }
+
         def_node_matcher :aggregate_failures?, <<-PATTERN
           (block {
               (send _ _ <(sym :aggregate_failures) ...>)
-              (send _ _ ... (hash <(pair (sym :aggregate_failures) true) ...>))
-            } ...)
-        PATTERN
-
-        def_node_matcher :aggregate_failures_present?, <<-PATTERN
-          (block {
-              (send _ _ <(sym :aggregate_failures) ...>)
-              (send _ _ ... (hash <(pair (sym :aggregate_failures) _) ...>))
+              (send _ _ ... (hash <(pair (sym :aggregate_failures) %1) ...>))
             } ...)
         PATTERN
 
@@ -89,12 +85,12 @@ module RuboCop
           node_with_aggregate_failures = find_aggregate_failures(example_node)
           return false unless node_with_aggregate_failures
 
-          aggregate_failures?(node_with_aggregate_failures)
+          aggregate_failures?(node_with_aggregate_failures, TRUE)
         end
 
         def find_aggregate_failures(example_node)
           example_node.send_node.each_ancestor(:block)
-            .find { |block_node| aggregate_failures_present?(block_node) }
+            .find { |block_node| aggregate_failures?(block_node, ANYTHING) }
         end
 
         def find_expectation(node, &block)


rubocop-rspec.gemspec
```diff
@@ -37,6 +37,7 @@ Gem::Specification.new do |spec|
   }
 
   spec.add_runtime_dependency 'rubocop', '>= 0.68.1'
+  spec.add_runtime_dependency 'rubocop-ast'
```
pirj, line 40: With `rubocop` 1.0, and explicit requirement of `rubocop-ast` becomes unnecessary.

Gemfile
```diff
@@ -4,6 +4,8 @@ source 'https://rubygems.org'
 
 gemspec
 
+gem 'rubocop-ast', github: 'marcandre/rubocop-ast', branch: 'nodepatternmatch'
```
pirj, line 7: This is merged actually https://github.com/rubocop-hq/rubocop-ast/pull/31

pirj:
RuboCop 0.8.7 [we're currently depend on](https://github.com/rubocop-hq/rubocop-rspec/blob/1be4c09d7c1963f88f360974a8b6bfcf735d7216/rubocop-rspec.gemspec#L39) [depends on `rubocop-ast` 0.1.0](https://github.com/rubocop-hq/rubocop/blob/v0.87.0/rubocop.gemspec#L41), and [`rubocop-ast` has this feature baked-in](https://github.com/rubocop-hq/rubocop-ast/blob/master/CHANGELOG.md#new-features-1). So this change is safe to merge without expecting breakages with earlier RuboCop versions.

lib/rubocop/cop/rspec/multiple_expectations.rb
```diff
@@ -89,12 +84,12 @@ def example_with_aggregate_failures?(example_node)
...
           return false unless node_with_aggregate_failures
 
-          aggregate_failures?(node_with_aggregate_failures)
+          aggregate_failures?(node_with_aggregate_failures, :true_type?.to_proc)
         end
 
         def find_aggregate_failures(example_node)
           example_node.send_node.each_ancestor(:block)
-            .find { |block_node| aggregate_failures_present?(block_node) }
+            .find { |block_node| aggregate_failures?(block_node, ANYTHING) }
```
pirj, line 92: A new addition to `rubocop-ast` made it possible to pass in `true` and `_`-alike's.
That was it became possible to collapse two node matchers into one, since they only had that distinction.

One is used to match:
```ruby
describe aggregate_examples: true do
# or
describe :aggregate_examples do
```

the other any of:
```ruby
describe aggregate_examples: false do
# or
describe aggregate_examples: true do
# or
describe :aggregate_examples do
```

pirj:
Green.

lib/rubocop/cop/rspec/multiple_expectations.rb
```diff
@@ -89,12 +84,12 @@ def example_with_aggregate_failures?(example_node)
           node_with_aggregate_failures = find_aggregate_failures(example_node)
           return false unless node_with_aggregate_failures
 
-          aggregate_failures?(node_with_aggregate_failures)
+          aggregate_failures?(node_with_aggregate_failures, :true_type?.to_proc)
```
Darhazer, line 87: the direct `to_proc` usage is very uncommon. I still wonder whether I prefer an explicit lambda

pirj: Agree, changed to an explicit lambda just like `ANYTHING`.

PR #932, opened 2020-06-12
RSpec/SubjectStub. Add top level group callback

andrykonchin:
It's a followup of https://github.com/rubocop-hq/rubocop-rspec/pull/925 and it addresses the following issue https://github.com/rubocop-hq/rubocop-rspec/pull/925#issuecomment-641151939

Changes:
- added module TopLevelGroup (similar to TopLevelDescribe)
- refactored SubjectStub to avoid manual caching and rely on TopLevelGroup

The goal of TopLevelGroup is to handle all the possible example and shared groups in a spec file. It calls `on_top_level_group` callback for every top-level `describe`, `context`, `feature`, `shared context`... group. 

---

* [x] Added tests.
* [ ] Updated documentation.


spec/rubocop/cop/rspec/subject_stub_spec.rb
```diff
@@ -321,4 +321,46 @@
       end
     RUBY
   end
+
+  it 'flags when there are several top level example groups' do
+    expect_offense(<<-RUBY)
+      describe Foo do
```
pirj, line 327: Maybe prefix with `RSpec.` to emphasize that those are top-level example groups?

lib/rubocop/rspec/top_level_group.rb
```diff
@@ -0,0 +1,43 @@
...
+      def top_level_group?(node)
+        top_level_groups.include?(node)
+      end
+
+      def top_level_groups
+        @top_level_groups ||= top_level_nodes.select { |n| example_or_shared_group?(n) }
+      end
+
+      def top_level_nodes
+        @top_level_nodes ||= \
```
pirj, line 34: `\` seems redundant.

lib/rubocop/rspec/top_level_group.rb
```diff
@@ -0,0 +1,43 @@
...
+      end
+
+      private
+
+      def top_level_group?(node)
+        top_level_groups.include?(node)
+      end
+
+      def top_level_groups
+        @top_level_groups ||= top_level_nodes.select { |n| example_or_shared_group?(n) }
```
pirj: Can't say off the top of my head if there is one single instance of a cop for all the files being checked, or a new cop instance is created for every file.
Wondering if this might change suddenly in the future as a flyweight optimization and cops using this module will only take action on the first file they analyze.
This might be checked by hand, but I'm not sure if this can be covered by our specs.

pirj: The only precautionary measure I can think of is a check that `processed_source` or its `ast` is still the same object.
I hope there are no plans to reuse those objects between files :D

spec/rubocop/cop/rspec/subject_stub_spec.rb
```diff
@@ -321,4 +321,46 @@
...
+      context xcontext fcontext
+      feature xfeature ffeature
+      example_group
+      shared_examples shared_examples_for shared_context
+    ].each do |method|
+      it "flags in top level #{method}" do
+        expect_offense(<<-RUBY)
+          #{method} '#{method}' do
+            it 'uses an implicit subject' do
+              expect(subject).to receive(:bar).and_return(baz)
```
pirj, line 358: This kind of goes against https://github.com/rubocop-hq/rubocop-rspec/pull/226/files#diff-bb66c4002cf82dad36a808e5e54c7475R11. Maybe remove the `and_return` part if it's not strictly necessary?

tags
```diff
@@ -0,0 +1,941 @@
+!_TAG_FILE_FORMAT	2	/extended format; --format=1 will not append ;" to lines/
```
pirj, line 1: Please feel free to add to `.gitignore`, or you may consider adding it globally to `~/.config/git/ignore`.

lib/rubocop/rspec/top_level_group.rb
```diff
@@ -0,0 +1,47 @@
...
+module RuboCop
+  module RSpec
+    # Helper methods for top level example group cops
+    module TopLevelGroup
+      extend NodePattern::Macros
+
+      def_node_matcher :example_or_shared_group?,
+                       (
+                         Language::ExampleGroups::ALL +
+                         Language::SharedGroups::ALL
```
pirj, line 12: I think `Language` is included (or you can include it if it's not), and you may skip this part of the namespace, and shrink it into less lines.

lib/rubocop/rspec/top_level_group.rb
andrykonchin: > Can't say off the top of my head if there is one single instance of a cop for all the files being checked, or a new cop instance is created for every file.

Right now a new cop instance is created for every file

pirj: Cool, please disregard then. 👍 
There's no need for us to be defensive.

andrykonchin: > Wondering if this might change suddenly in the future as a flyweight optimization and cops using this module will only take action on the first file they analyze.

It's definitely an issue.

I noticed at least one module (`SurroundingSpace`) which memoizes some values in cop's instance variables so it's an already existing issue :-)

https://github.com/rubocop-hq/rubocop/blob/master/lib/rubocop/cop/mixin/surrounding_space.rb#L47-L56

lib/rubocop/rspec/top_level_group.rb
```diff
@@ -0,0 +1,47 @@
...
+      end
+
+      def top_level_nodes
+        @top_level_nodes ||= \
+          if root_node.begin_type?
+            root_node.children
+          else
+            [root_node]
+          end
+      end
```
andrykonchin, line 40: @pirj 

TBH I am not sure whether this logic is correct. I just noticed there are two cases:
- when there is only one top-level `describe` in the file - the root node (`processed_source.ast`) is a `block` node
- when there are several top-level `describe` blocks - the root node is `block` and all the `describe`s are nested in it as `block` nodes.

Not sure whether there are no other cases.

I check there `block` node type. But I noticed there is a `kwblock` node type and I have no idea when it's used and should we take it into account.

andrykonchin:
@pirj Fixed all the issues

lib/rubocop/rspec/top_level_group.rb
andrykonchin: Looks like `kwbegin`  means just explicit `begin`/`end` block

lib/rubocop/rspec/top_level_group.rb
```diff
@@ -0,0 +1,45 @@
...
+        top_level_groups.include?(node)
+      end
+
+      def top_level_groups
+        @top_level_groups ||=
+          top_level_nodes.select { |n| example_or_shared_group?(n) }
+      end
+
+      def top_level_nodes
+        @top_level_nodes ||=
```
pirj, line 32: Minor: does it make sense to memoize it? It's only called once from a method that is also memoized.

lib/rubocop/rspec/top_level_group.rb
```diff
@@ -0,0 +1,44 @@
...
+      def top_level_group?(node)
+        top_level_groups.include?(node)
+      end
+
+      def top_level_groups
+        @top_level_groups ||=
+          top_level_nodes.select { |n| example_or_shared_group?(n) }
+      end
+
+      def top_level_nodes
```
Darhazer, line 31: At some point, we may want to DRY this between the different modules.
Also, top_level_describe seems to be a special case on top_level_group.

For now I'm fine with keeping it as it is

pirj:
Merging without consensus, since this change has been discussed in the past, and only affects one cop.

Thank you again, @andrykonchin !

lib/rubocop/rspec/top_level_group.rb
```diff
@@ -0,0 +1,44 @@
...
+    # Helper methods for top level example group cops
+    module TopLevelGroup
+      extend RuboCop::NodePattern::Macros
+      include RuboCop::RSpec::Language
+
+      def_node_matcher :example_or_shared_group?,
+                       (ExampleGroups::ALL + SharedGroups::ALL).block_pattern
+
+      def on_block(node)
+        return unless respond_to?(:on_top_level_group)
```
pirj, line 14: An afterthought.

I understand that this guard statement is on the very top to avoid a quite expensive check done in `top_level_group?`.

But is it really necessary? If a cop includes `TopLevelGroup`, that means it has a `on_top_level_group` method?
Or am I missing something?

andrykonchin: TBH I just copied this approach from `TopLevelDescribe` module. I don't see any benefit in having such condition here.

PR #934, opened 2020-06-14
Pull example group node matchers up

mockdeep:
This refactors to add `shared_group?` and `spec_group?` node matchers in
`NodePattern` for use across cops. I wanted to check shared contexts in
the [rule that I'm working on][1] and noticed that there were similar
implementations in a handful of other places. I thought it was probably
worth abstracting upwards.

[1]: https://github.com/rubocop-hq/rubocop-rspec/pull/863

---

* [n/a] Added tests.
* [n/a] Updated documentation.
* [n/a] Added an entry to the `CHANGELOG.md` if the new code introduces user-observable changes.

### DIFF
--- a/lib/rubocop/cop/rspec/describe_class.rb
+++ b/lib/rubocop/cop/rspec/describe_class.rb
@@ -51,8 +51,6 @@ module RuboCop
           )
         PATTERN
 
-        def_node_matcher :shared_group?, SharedGroups::ALL.block_pattern
-
         def on_top_level_describe(node, (described_value, _))
           return if shared_group?(root_node)
           return if valid_describe?(node)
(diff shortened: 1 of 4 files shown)


mockdeep:
Part of me wonders if there should be `on_example_group`, `on_spec_group`, `on_shared_group` macros, but that looks like it would be more work.

pirj:
> if there should be

Depends on how often they would be used. Can't tell that off the top of my head.

> it would be more work

Yes, but not significantly a lot more, there are references to copy-paste, e.g. https://github.com/rubocop-hq/rubocop-rspec/blob/master/lib/rubocop/rspec/top_level_group.rb

PR #935, opened 2020-06-14
Deduplicate logic in ExampleGroup

mockdeep:
This takes the various methods for different types of constructs in
`ExampleGroup` and combines them into a couple of generic methods. Not
sure, but I think it may also address an existing bug if `subject {}` or
other methods that match constructs are called within an example. The
`hooks` had a check for `example?`, but the others didn't. This prevents
recursing if the node is an example in all cases.

---

* [ ] ~~Added tests.~~
* [ ] ~~Updated documentation.~~
* [ ] ~~Added an entry to the `CHANGELOG.md` if the new code introduces user-observable changes.~~

### DIFF
--- a/lib/rubocop/rspec/example_group.rb
+++ b/lib/rubocop/rspec/example_group.rb
@@ -14,72 +14,44 @@ module RuboCop
         ExampleGroups::ALL + SharedGroups::ALL + Includes::ALL
       ).block_pattern
 
+      def lets
+        find_all_in_scope(node, :let?)
+      end
+
       def subjects
-        subjects_in_scope(node)
+        find_all_in_scope(node, :subject?)
       end
 
       def examples
-        examples_in_scope(node).map(&Example.public_method(:new))
+        find_all_in_scope(node, :example?).map(&Example.public_method(:new))
       end
 
       def hooks
-        hooks_in_scope(node).map(&Hook.public_method(:new))
+        find_all_in_scope(node, :hook?).map(&Hook.public_method(:new))
       end
 
       private
 
-      def subjects_in_scope(node)
-        node.each_child_node.flat_map do |child|
-          find_subjects(child)
-        end
-      end
-
-      def find_subjects(node)
-        return [] if scope_change?(node)
-
-        if subject?(node)
-          [node]
-        else
-          subjects_in_scope(node)
-        end
-      end
-
-      def hooks_in_scope(node)
-        node.each_child_node.flat_map do |child|
-          find_hooks(child)
-        end
-      end
-
-      def find_hooks(node)
-        return [] if scope_change?(node) || example?(node)
-
-        if hook?(node)
-          [node]
-        else
-          hooks_in_scope(node)
-        end
-      end
-
-      def examples_in_scope(node, &blk)
-        node.each_child_node.flat_map do |child|
-          find_examples(child, &blk)
-        end
-      end
-
-      # Recursively search for examples within the current scope
+      # Recursively search for predicate within the current scope
       #
-      # Searches node for examples and halts when a scope change is detected
+      # Searches node and halts when a scope change is detected
       #
-      # @param node [RuboCop::Node] node to recursively search for examples
+      # @param node [RuboCop::Node] node to recursively search
       #
-      # @return [Array<RuboCop::Node>] discovered example nodes
-      def find_examples(node)
-        return [] if scope_change?(node)
+      # @return [Array<RuboCop::Node>] discovered nodes
+      def find_all_in_scope(node, predicate)
+        node.each_child_node.flat_map do |child|
+          find_all(child, predicate)
+        end
+      end
 
-        if example?(node)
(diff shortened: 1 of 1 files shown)


lib/rubocop/rspec/example_group.rb
```diff
@@ -14,72 +14,44 @@ class ExampleGroup < Concept
...
       end
 
       def examples
-        examples_in_scope(node).map(&Example.public_method(:new))
+        find_all_in_scope(node, :example?).map(&Example.public_method(:new))
       end
 
       def hooks
-        hooks_in_scope(node).map(&Hook.public_method(:new))
+        find_all_in_scope(node, :hook?).map(&Hook.public_method(:new))
```
pirj, line 30:
```diff
- find_all_in_scope(node, :hook?).map(&Hook.public_method(:new))
+ find_all_in_scope(node, :hook?).map { |hook| Hook.new(hook) }
```
Looks even shorter when called with a block.
Are there any performance implications?

lib/rubocop/rspec/example_group.rb
```diff
@@ -14,72 +14,44 @@ class ExampleGroup < Concept
...
-        return [] if scope_change?(node)
+      # @return [Array<RuboCop::Node>] discovered nodes
+      def find_all_in_scope(node, predicate)
+        node.each_child_node.flat_map do |child|
+          find_all(child, predicate)
+        end
+      end
 
-        if example?(node)
+      def find_all(node, predicate)
```
pirj, line 48: It's so tempting to change this to `yield` somehow.

mockdeep: I poked around with it for a minute, but wasn't able to make that work. I think it would probably need more complex syntax above in order for it to work.

lib/rubocop/rspec/example_group.rb
mockdeep: The usual caveats about benchmarks apply, but it looks like the `&` syntax might actually be slower:

```
ampersand
  917906.0 i/s (±3.44%)
curlies
  1339367.1 i/s (±0.69%)

Comparison:
    curlies:   1339367.1 i/s
  ampersand:    917906.0 i/s - 1.46x slower
```

```rb
require 'bundler/inline'

gemfile do
  source 'https://rubygems.org'
  gem 'benchmark-inputs'
end

require "benchmark/inputs"

INPUTS = [
  [1, 2, 3],
  [1, nil, 2, nil, 3, nil],
  [nil, nil, nil],
]

class MyClass
  attr_accessor :foo
  def initialize(foo)
    self.foo = foo
  end
end

Benchmark.inputs(INPUTS) do |job|
  job.dup_inputs = true
  job.report("ampersand"){|array| array.map(&MyClass.method(:new)) }
  job.report("curlies"){|array| array.map { |element| MyClass.new(element) } }
  job.compare!
end
```

lib/rubocop/rspec/example_group.rb
pirj: Please disregard this. I took a stab at it, and some of **our** cops fail. Not to say that there can be more custom cops using `ExampleGroup` that would break too.

When it's merged, I'll file a ticket to add new methods (maybe `find_subjects`? etc) to be added, fix our cops to use the new methods, and add deprecations to the old methods returning arrays.

Just for the record:
```ruby
      # @return [Enumerator<RuboCop::Node>] discovered nodes
      def find_all_in_scope(node, predicate)
        Enumerator.new do |yielder|
          node.each_child_node.flat_map do |child|
            find_all(child, predicate) { |found| yielder.yield child }
          end
        end
      end

      def find_all(node, predicate, &block)
        yield node if public_send(predicate, node)
        return if scope_change?(node) || example?(node)

        find_all_in_scope(node, predicate, &block)
      end
```

Example fix:
```diff
subjects[0...-1].each do |subject|
subjects.reverse_each.drop(1).each do |subject|
 ```

This might work better in case we only need `.first` or we do `.find` over the results and don't want all of the results upfront.

lib/rubocop/rspec/example_group.rb
pirj: TIL `benchmark-inputs` 👍 

Oh, I didn't know `&method` is slower, and couldn't even imagine that it's that slower.

Thanks for using the scientific approach here!

lib/rubocop/rspec/example_group.rb
mockdeep: Ah, I see. You're trying to avoid doing extra work. Though, fwiw, without a benchmark, it's hard to know whether there will be a noticeable performance difference. There have been a handful of times where I've checked these sorts of optimizations and they've actually performed *worse*, surprisingly.

pirj: Absolutely.
There's some [RuboCop performance tuning going on](https://github.com/rubocop-hq/rubocop/issues/8022).

Just an option we may may try out to see. Don't think it's even a significant bottleneck.

Issue #919, opened 2020-05-27
Let's change the default for RSpec/PredicateMatcher


marcandre:
`RSpec/PredicateMatcher` tells me to prefer the first form to the second one:

```ruby
    context 'given some result' do
      let(:error) { 'Oops, a few things blew up and nothing worked' }

      it 'check with nil?' do
        expect(error.nil?).to be(true)
      end

      it 'check with be_nil' do
        expect(error).to be_nil
      end
    end
```

This is gravely mistaken. It's longer, it's imo less clear, but much more importantly a failure (like in this example) gives uninformative info:

```sh
  1) some example given some result check with nil?
     Failure/Error: expect(error.nil?).to be(true)
     
       expected true
            got false

  2) some example given some result check with be_nil
     Failure/Error: expect(error).to be_nil
     
       expected: nil
            got: "Oops, a few things blew up and nothing worked"
```

Notice how the second failure is much more informative, showing the unexpected error and all. That is the reason to *always* use precise matchers.

I don't mind if we keep the functionality for the (imo always bad) style, I don't care if we propose `be_nil` or `eq(nil)`, but the default should change.

Let's minimize the reasons to curse at RuboCop.

pirj:
We don't have a formal checklist for changing the defaults, but if I had to come up with it, it would be:

- [ ] make sure to only change defaults on major version updates
- [ ] check settings for that cop in
     - https://github.com/eliotsykes/real-world-rails
     - https://github.com/jeromedalbert/real-world-ruby-apps
- [x] check with the style guide https://rspec.rubystyle.guide/ (https://rspec.rubystyle.guide/#predicate-matchers)

It would be really nice to harmonize the defaults with the style guide and to harmonize the style guide with what is really commonly used in those open-source apps. It looks like a task on its own, but before we've done that, we can follow that "simple" checklist.

Would you like to grep through `.rubocop.yml` settings for `PredicateMatcher` of `real-world-rails`/`real-world-ruby-apps`? Part of them doesn't use RuboCop or RSpec, unfortunately, but some can be used to make a judgememnt on this topic.

pirj:
Found a mistake in the example, "also good" part:

```ruby
Strict: true, EnforcedStyle: inflected (default)

# bad
expect(foo.something?).to be_truthy

# good
expect(foo).to be_something

# also good - It checks "true" strictly.
expect(foo).to be(true)

# ^^^ should be
expect(foo.something?).to be(true)
```

[...]

PR #937, opened 2020-06-16
Fix mistake in RSpec/PredicateMatcher doc

pirj:
Found in #919

---

* [-] Added tests.
* [x] Updated documentation.

### DIFF
--- a/lib/rubocop/cop/rspec/predicate_matcher.rb
+++ b/lib/rubocop/cop/rspec/predicate_matcher.rb
@@ -277,12 +277,12 @@ module RuboCop
       #   expect(foo).to be_something
       #
       #   # also good - It checks "true" strictly.
-      #   expect(foo).to be(true)
+      #   expect(foo.something?).to be(true)
       #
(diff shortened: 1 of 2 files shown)


lib/rubocop/cop/rspec/predicate_matcher.rb
```diff
@@ -277,7 +277,7 @@ def replacement_matcher(node)
       #   expect(foo).to be_something
```
Darhazer, line 277: Actually this one should be updated as well

pirj: To what exactly?

Darhazer: Oh, sorry, misread the examples :facepalm: 
Seems not a good day for me to do reviews :)

Darhazer: What about line 285?

pirj: Good catch, missed that other one!

PR #942, opened 2020-06-19
Add mixins folder

mockdeep:
This moves several files to a new `mixins/` folder. I was going to add
some new macros, per [this comment][1], but thought it might clutter
things up a bit to put them in the root folder. I thought of adding a
`macros/` folder, but decided to emulate the [rubocop source][2]
instead.

[1]: https://github.com/rubocop-hq/rubocop-rspec/pull/934#issuecomment-643813427
[2]: https://github.com/rubocop-hq/rubocop/tree/master/lib/rubocop/cop/mixin

---

* [ ] ~~Added tests.~~
* [ ] ~~Updated documentation.~~
* [ ] ~~Added an entry to the `CHANGELOG.md` if the new code introduces user-observable changes.~~


lib/rubocop-rspec.rb
```diff
@@ -9,21 +9,23 @@
...
 require_relative 'rubocop/rspec/language'
 require_relative 'rubocop/rspec/language/node_pattern'
-require_relative 'rubocop/rspec/top_level_group'
+
+require_relative 'rubocop/rspec/mixin/align_let_brace'
+require_relative 'rubocop/rspec/mixin/top_level_describe'
+require_relative 'rubocop/rspec/mixin/top_level_group'
+require_relative 'rubocop/rspec/mixin/variable'
+require_relative 'rubocop/rspec/mixin/final_end_location'
+require_relative 'rubocop/rspec/mixin/blank_line_separation'
```
pirj, line 21: I guess maybe `rubocop/rspec/cop/mixin/*` if they are intended to be included in cops like [RuboCop does it](https://github.com/rubocop-hq/rubocop/blob/master/lib/rubocop.rb#L51)?

mockdeep: @pirj yeah, I thought about that briefly. The weird thing in our case is that `FinalEndLocation` gets included in two places, one being `lib/rubocop/rspec/corrector/move_node.rb`. Should I put that in the `cop/mixin/` directory anyway, or leave it in `rubocop/rspec/`? It also gets included in `blank_line_separation.rb` which is itself included in cops.

mockdeep: @pirj I went ahead and moved them over to the `cop/rspec/` folder. That one case is a little odd, but I think it's probably still better this way.

mockdeep:
Hmm, is this failure related to something on master?

lib/rubocop-rspec.rb
```diff
@@ -9,21 +9,23 @@
 require_relative 'rubocop/rspec/version'
 require_relative 'rubocop/rspec/inject'
 require_relative 'rubocop/rspec/node'
-require_relative 'rubocop/rspec/top_level_describe'
 require_relative 'rubocop/rspec/wording'
 require_relative 'rubocop/rspec/language'
 require_relative 'rubocop/rspec/language/node_pattern'
-require_relative 'rubocop/rspec/top_level_group'
+
+require_relative 'rubocop/cop/rspec/mixin/align_let_brace'
```
Darhazer, line 16: Technically, this is not a mixin. It's a helper class used by 2 cops.

pirj: Agree. Let's keep in `rubocop/rspec`.
Everything else is 👍

mockdeep: Okay, sorry for the delay on this. I've moved `align_let_brace` back to where it came from.

mockdeep:
Not sure why the build is failing on this. It works for me locally.

bquorning:
> Not sure why the build is failing on this. It works for me locally.

I think you’re blocked by #955.

pirj:
#955 is merged. Please rebase. Sorry for the trouble.

lib/rubocop-rspec.rb
```diff
@@ -9,21 +9,23 @@
 require_relative 'rubocop/rspec/version'
 require_relative 'rubocop/rspec/inject'
 require_relative 'rubocop/rspec/node'
-require_relative 'rubocop/rspec/top_level_describe'
 require_relative 'rubocop/rspec/wording'
 require_relative 'rubocop/rspec/language'
 require_relative 'rubocop/rspec/language/node_pattern'
-require_relative 'rubocop/rspec/top_level_group'
+
+require_relative 'rubocop/cop/rspec/mixin/top_level_describe'
```
bquorning, line 16: When we change the folder like this, shouldn’t we also change the namespace? E.g. `RuboCop::RSpec::TopLevelDescribe` -> `RuboCop::Cop::RSpec::TopLevelDescribe`.

pirj: Agree.

@mockdeep We've just renamed `RuboCop::Cop::RSpec::Cop` to `RuboCop::Cop::RSpec::Base` following RuboCop core's naming.
If we happen to move classes/modules around namespaces, and if it happens those are used by custom cops, we'll break them with a minor version update.
So for `Base`, we've decided to keep the old class, `Cop` as well until 2.0.
I don't see how it would be possible to release this change with namespace before we release 2.0 without similar breakage risks.

AFAI understand #863 depends on this pull request. But if this is going to be released in 2.0, that means that #863 will only be released as `enabled` in 3.0, until then it will be `pending`.

So I propose two options for #863:
 - untangle it from this pull request, and release separately
 - keep it dependent on this pull request and release post-2.0

Please let me know what you think.

pirj:
Adjusted module namespace and rebased.
Let's wait for Github to catch up.

pirj:
That's an option. Depends on how essential this is to move #863 forward.

mockdeep:
Unfortunately, as you may have noticed, I haven't been able to be very consistent on following up with these PRs. I don't think #863 depends on this, since, if I remember correctly it was built on #934 instead. I was picturing adding some more macros and using those instead, but we could stick with #863 as is if you want to get that out sooner.

pirj:
No rush, take your time and make any additions you see fit in #863.
We'll take care of this pull request later.
Even though RuboCop 1.0 and RuboCop RSpec 2.0 are around the corner, there's a fat list of leftovers there and something to work on here as well, so I guess it's not going to happen earlier than in 3-4 weeks.

pirj:
I'll take care of rebasing this.
This is optional, but a nice cleanup nevertheless.

pirj:
Superseded by #1056 
Thanks, @mockdeep, this will become part of the 2.0 release!

mockdeep:
@pirj thank you for rebasing!

PR #944, opened 2020-06-27
Allow `#traits_for_enum` to be used in a `FactoryBot` factory without a block

### LINKED ISSUE #943: Add `#traits_for_enum` to `RuboCop::RSpec::FactoryBot.reserved_methods`
Thank you for building and maintaining `rubocop-rspec`. My team uses this and a lot of the defaults are pretty sensible, and help us to write cleaner tests!

`FactoryBot` recently released version [`6.0.0`](https://github.com/thoughtbot/factory_bot/blob/master/NEWS.md#600-june-18-2020), which introduces a new method to use within a factory definition, `#traits_for_enum`. Using this causes Rubocop to flag it with `FactoryBot/AttributeDefinedStatically`. This issue stems from the fact that this new method is not included in `Rubocop::RSpec::FactoryBot.reserved_methods`.

https://github.com/rubocop-hq/rubocop-rspec/blob/fad338729759334cb7411802e54cdf84f0e685c4/lib/rubocop/rspec/factory_bot.rb#L37-L40

A simple fix would be to include this reserved method in the list. I don't mind creating a PR for this.

Curious to hear your thoughts!

harrylewis:
This PR fixes a bug that arose in relation to FactoryBot's recent [`6.0.0` version release](https://github.com/thoughtbot/factory_bot/blob/master/NEWS.md#600-june-18-2020). This release introduced a new method, `#traits_for_enum`, which can be used to dynamically define factory traits for an enumerated attribute.

The behaviour expected is that usage of this method (which does not take a block argument) does not produce a `FactoryBot/AttributeDefinedStatically` offense.

The actual behaviour is that it does produce this offense because it is assumed to be an attribute, which should be defined with a block. The fix is simply to add it to the list of pre-qualified methods.

The documentation surrounding this cop does not make any mention around what methods are allowed to be used without a block, so I didn't feel it was necessary to update the documentation.

I _believe_ this constitutes a "user-observable" change, but am not sure. If you feel it does, I will add a commit to add an entry to the `CHANGELOG.md`.

This fixes #943.

---

* [x] Added tests.
* [x] Updated documentation.

### DIFF
--- a/lib/rubocop/rspec/factory_bot.rb
+++ b/lib/rubocop/rspec/factory_bot.rb
@@ -4,7 +4,7 @@ module RuboCop
   module RSpec
     # RuboCop FactoryBot project namespace
     module FactoryBot
-      ATTRIBUTE_DEFINING_METHODS = %i[factory trait transient ignore].freeze
+      ATTRIBUTE_DEFINING_METHODS = %i[factory trait traits_for_enum transient ignore].freeze
 
       UNPROXIED_METHODS = %i[
         __send__
--- a/lib/rubocop/rspec/factory_bot.rb
+++ b/lib/rubocop/rspec/factory_bot.rb
@@ -4,7 +4,13 @@ module RuboCop
   module RSpec
     # RuboCop FactoryBot project namespace
     module FactoryBot
-      ATTRIBUTE_DEFINING_METHODS = %i[factory trait traits_for_enum transient ignore].freeze
+      ATTRIBUTE_DEFINING_METHODS = %i[
+        factory
+        ignore
+        trait
+        traits_for_enum
+        transient
+      ].freeze
 
       UNPROXIED_METHODS = %i[
         __send__
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## Master (Unreleased)
 
 * Extend the list of Rails spec types for `RSpec/DescribeClass`. ([@pirj][])
+* Fix `FactoryBot/AttributeDefinedStatically` to allow `#traits_for_enum` without a block. ([@harrylewis][])
 
 ## 1.40.0 (2020-06-11)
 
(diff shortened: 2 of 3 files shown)


lib/rubocop/rspec/factory_bot.rb
```diff
@@ -4,7 +4,7 @@ module RuboCop
   module RSpec
     # RuboCop FactoryBot project namespace
     module FactoryBot
-      ATTRIBUTE_DEFINING_METHODS = %i[factory trait transient ignore].freeze
+      ATTRIBUTE_DEFINING_METHODS = %i[factory trait traits_for_enum transient ignore].freeze
```
harrylewis, line 7: This list doesn't seem to be ordered alphabetically, so I decided to add the method near a similar entry, `:trait`.

lib/rubocop/rspec/factory_bot.rb
```diff
@@ -4,7 +4,13 @@ module RuboCop
...
     # RuboCop FactoryBot project namespace
     module FactoryBot
-      ATTRIBUTE_DEFINING_METHODS = %i[factory trait transient ignore].freeze
+      ATTRIBUTE_DEFINING_METHODS = %i[
+        factory
+        ignore
+        trait
+        traits_for_enum
+        transient
+      ].freeze
```
harrylewis, line 13: Adding `:traits_for_enum` caused a `Layout/LineLength` offense, so I turned the horizontal list into a vertical list. I also alphabetized the list to align with what was done with the other constants below.

harrylewis:
One of the non-required builds failed, `edge-rubocop`. It doesn't look like the failure has anything to do with my changes. Given it is testing using the "edge" of Rubocop, I'm assuming it is susceptible to failures like this. Let me know if there is anything I can do to resurrect the failing test.

harrylewis:
@Darhazer I added a `CHANGELOG` entry. I used the "Fix" verb because the `bug` label was applied to the original issue.

Also, regarding the checklist item "squashed related commits together", I'm not sure if it means to squash all commits in the branch into one, or if it is good enough to make sure the PR is merged into `master` with a squash commit. Happy to do either!

harrylewis:
I didn't realize at first what was required to fill out a `CHANGELOG` entry. I really appreciate the accompanying tests though. They helped me understand was was required!

PR #949, opened 2020-06-28
Performance. FactoryBot/AttributeDefinedStatically

andrykonchin:
Optimized performance of FactoryBot/AttributeDefinedStatically cop.

### Changes

- got rid of recursive `def_node_search` macros
- and avoid repeated processing the same block multiple times

Previous logic was
- take each block
- recursively find all the `factory`/`trait`/etc blocks in it
- and return all the defined attributes inside them

New logic
- take each block
- check whether it's `factory`/`trait`/etc
- return defined attributes

### Performance measurements

Timing for `RuboCop::Cop::RSpec::FactoryBot::AttributeDefinedStatically#on_block` is changed from 6.8% to 0.3%.

```
stackprof tmp/stackprof-cpu-gitlab.master.with-rubocop-rspec.AttributeDefinedStatically.dump --method 'RuboCop::Cop::RSpec::FactoryBot::AttributeDefinedStatically#on_block'
RuboCop::Cop::RSpec::FactoryBot::AttributeDefinedStatically#on_block (/Users/andrykonchin/projects/rubocop-rspec/lib/rubocop/cop/rspec/factory_bot/attribute_defined_statically.rb:38)
  samples:    21 self (0.1%)  /   1487 total (6.8%)
  callers:
    1487  (  100.0%)  RuboCop::Cop::Commissioner#trigger_responding_cops
       1  (    0.1%)  RuboCop::Cop::RSpec::FactoryBot::AttributeDefinedStatically#on_block
  callees (1466 total):
    1465  (   99.9%)  RuboCop::Cop::RSpec::FactoryBot::AttributeDefinedStatically#factory_attributes
       1  (    0.1%)  RuboCop::AST::Node#receiver
       1  (    0.1%)  RuboCop::Cop::RSpec::FactoryBot::AttributeDefinedStatically#on_block
  code:
                                  |    38  |
 1487    (6.8%) /    21   (0.1%)  |    39  |           def_node_matcher :factory_attributes_matcher, <<-PATTERN
    1    (0.0%)                   |    40  |             (block (send _ #attribute_defining_method? ...) _ { (begin $...) $(send ...) } )
                                  |    41  |           PATTERN
```
</details>

```
stackprof tmp/stackprof-cpu-gitlab.master.with-rubocop-rspec.AttributeDefinedStatically.2.dump --method 'RuboCop::Cop::RSpec::FactoryBot::AttributeDefinedStatically#on_block'
RuboCop::Cop::RSpec::FactoryBot::AttributeDefinedStatically#on_block (/Users/andrykonchin/projects/rubocop-rspec/lib/rubocop/cop/rspec/factory_bot/attribute_defined_statically.rb:43)
  samples:     6 self (0.0%)  /     57 total (0.3%)
  callers:
      57  (  100.0%)  RuboCop::Cop::Commissioner#trigger_responding_cops
  callees (51 total):
      51  (  100.0%)  RuboCop::Cop::RSpec::FactoryBot::AttributeDefinedStatically#factory_attributes_matcher
  code:
                                  |    43  |           def on_block(node)
   51    (0.2%)                   |    44  |             attributes = factory_attributes_matcher(node) || []
    5    (0.0%) /     5   (0.0%)  |    45  |             attributes = [attributes] if !attributes.is_a?(Array)
                                  |    46  |
    1    (0.0%) /     1   (0.0%)  |    47  |             attributes.each do |attribute|
                                  |    48  |               next unless offensive_receiver?(attribute.receiver, node)
```
</details>

### Measurements approach

```
bundle exec exe/rubocop --cache false --out gitlab-specs.out --force-default-config  --require rubocop-rspec --only FactoryBot/AttributeDefinedStatically ../rubocop-profiling-examples/gitlabhq/spec
```

---

* [ ] Added tests.
* [ ] Updated documentation.

### DIFF
--- a/lib/rubocop/cop/rspec/factory_bot/attribute_defined_statically.rb
+++ b/lib/rubocop/cop/rspec/factory_bot/attribute_defined_statically.rb
@@ -31,12 +31,15 @@ module RuboCop
             (send _ !#reserved_method? $...)
           PATTERN
 
-          def_node_search :factory_attributes, <<-PATTERN
+          def_node_matcher :factory_attributes, <<-PATTERN
             (block (send _ #attribute_defining_method? ...) _ { (begin $...) $(send ...) } )
           PATTERN
 
           def on_block(node)
-            factory_attributes(node).to_a.flatten.each do |attribute|
+            attributes = factory_attributes(node) || []
+            attributes = [attributes] unless attributes.is_a?(Array)
+
+            attributes.each do |attribute|
               next unless offensive_receiver?(attribute.receiver, node)
               next if proc?(attribute) || association?(attribute.first_argument)
 


lib/rubocop/cop/rspec/factory_bot/attribute_defined_statically.rb
```diff
@@ -31,12 +31,15 @@ class AttributeDefinedStatically < Cop
...
 
-          def_node_search :factory_attributes, <<-PATTERN
+          def_node_matcher :factory_attributes, <<-PATTERN
             (block (send _ #attribute_defining_method? ...) _ { (begin $...) $(send ...) } )
           PATTERN
 
           def on_block(node)
-            factory_attributes(node).to_a.flatten.each do |attribute|
+            attributes = factory_attributes(node) || []
+            attributes = [attributes] if !attributes.is_a?(Array)
```
andrykonchin, line 40: TBH I haven't found a better way to handle all the result types. So any advice is welcome.

`factory_attributes` may return:
- array of nodes (when there are several attribute declarations - it matches `(begin $...)`)
- one node (when there is only one attribute - it matches `$(send ...)`)
- nil (when a node doesn't match the pattern)

Method generated with `def_node_search` returned an array even if there was only one attribute

Darhazer:
```
Array(1)
=> [1]
irb(main):005:0> Array([1])
=> [1]
irb(main):006:0> Array([1,2])
=> [1, 2]
irb(main):007:0> Array(nil)
=> []
```
so you can just do
```
attributes = Array(factory_attributes(node))
```

pirj: Unfortunately, a `send` node deconstructs when wrapped with `Array()` and produces nonsense:
```ruby
> factory_attributes(node)
=> s(:send, nil, :author,
  s(:hash,
    s(:pair,
      s(:sym, :age),
      s(:int, 42)),
    s(:pair,
      s(:sym, :factory),
      s(:sym, :user))))

> Array(factory_attributes(node))
=> [nil, :author, s(:hash,
  s(:pair,
    s(:sym, :age),
    s(:int, 42)),
  s(:pair,
    s(:sym, :factory),
    s(:sym, :user)))]

> attributes
=> [s(:send, nil, :author,
  s(:hash,
    s(:pair,
      s(:sym, :age),
      s(:int, 42)),
    s(:pair,
      s(:sym, :factory),
      s(:sym, :user))))]
```

Darhazer: Then you can use [attribites].flatten.compact, but I would preserve your version actually, as it's obvious what is going there

lib/rubocop/cop/rspec/factory_bot/attribute_defined_statically.rb
```diff
@@ -31,12 +31,15 @@ class AttributeDefinedStatically < Cop
...
 
-          def_node_search :factory_attributes, <<-PATTERN
+          def_node_matcher :factory_attributes, <<-PATTERN
             (block (send _ #attribute_defining_method? ...) _ { (begin $...) $(send ...) } )
           PATTERN
 
           def on_block(node)
-            factory_attributes(node).to_a.flatten.each do |attribute|
+            attributes = factory_attributes(node) || []
+            attributes = [attributes] unless attributes.is_a?(Array)
```
bquorning, line 40: Nitpick: I _think_ these two lines could be written a bit simpler:
```rb
attributes = Array(factory_attributes(node))
```

Darhazer: I already suggested this, but unfortunately, node deconstructs itself in this case
https://github.com/rubocop-hq/rubocop-rspec/pull/949#discussion_r446888166

bquorning: Ah, thanks for clarifying.

PR #950, opened 2020-06-28
Performance. RSpec/NestedGroups

andrykonchin:
Optimized performance of RSpec/NestedGroups and #on_top_level_describe callback.

### Changes

- avoid excessive recursive iteration oven nested example groups
- fixed error with calculating nesting count

### Performance measurements

Timing for RSpec/NestedGroups is changed from 14.3% to 3.3% (for `RuboCop::RSpec::TopLevelDescribe#on_send` callback).

```
stackprof tmp/stackprof-cpu-gitlab.master.with-rubocop-rspec.NestedGroups.dump --method 'RuboCop::RSpec::TopLevelDescribe#on_send'
RuboCop::RSpec::TopLevelDescribe#on_send (/Users/andrykonchin/projects/rubocop-rspec/lib/rubocop/rspec/top_level_describe.rb:9)
  samples:   130 self (0.2%)  /   8732 total (16.6%)
  callers:
    8732  (  100.0%)  RuboCop::Cop::Commissioner#trigger_responding_cops
  callees (8602 total):
    8446  (   98.2%)  RuboCop::Cop::RSpec::NestedGroups#on_top_level_describe
     155  (    1.8%)  RuboCop::RSpec::TopLevelDescribe#top_level_describe?
       1  (    0.0%)  RuboCop::AST::MethodDispatchNode#arguments
  code:
                                  |     9  |       def on_send(node)
  127    (0.2%) /   127   (0.2%)  |    10  |         return unless respond_to?(:on_top_level_describe)
  158    (0.3%) /     3   (0.0%)  |    11  |         return unless top_level_describe?(node)
                                  |    12  |
 8447   (16.1%)                   |    13  |         on_top_level_describe(node, node.arguments)
                                  |    14  |       end

```
</details>

<details><summary>After</summary>

```
stackprof tmp/stackprof-cpu-gitlab.master.with-rubocop-rspec.NestedGroups.4.dump --method 'RuboCop::RSpec::TopLevelDescribe#on_send'
RuboCop::RSpec::TopLevelDescribe#on_send (/Users/andrykonchin/projects/rubocop-rspec/lib/rubocop/rspec/top_level_describe.rb:9)
  samples:    49 self (0.2%)  /    715 total (3.1%)
  callers:
     715  (  100.0%)  RuboCop::Cop::Commissioner#trigger_responding_cops
  callees (666 total):
     593  (   89.0%)  RuboCop::Cop::RSpec::NestedGroups#on_top_level_describe
      73  (   11.0%)  RuboCop::RSpec::TopLevelDescribe#top_level_describe?
  code:
                                  |     9  |       def on_send(node)
   49    (0.2%) /    49   (0.2%)  |    10  |         return unless respond_to?(:on_top_level_describe)
   73    (0.3%)                   |    11  |         return unless top_level_describe?(node)
                                  |    12  |
  593    (2.6%)                   |    13  |         on_top_level_describe(node, node.arguments)
                                  |    14  |       end

```
</details>

### Measurements approach

```
bundle exec exe/rubocop --cache false --out gitlab-specs.out --force-default-config  --require rubocop-rspec --only RSpec/NestedGroups ../rubocop-profiling-examples/gitlabhq/spec
```

---

* [x] Added tests.
* [ ] Updated documentation.

### DIFF
--- a/lib/rubocop/cop/rspec/nested_groups.rb
+++ b/lib/rubocop/cop/rspec/nested_groups.rb
@@ -97,13 +97,11 @@ module RuboCop
           "Configuration key `#{DEPRECATED_MAX_KEY}` for #{cop_name} is " \
           'deprecated in favor of `Max`. Please use that instead.'
 
-        def_node_search :find_contexts, ExampleGroups::ALL.block_pattern
-
         def on_top_level_describe(node, _args)
-          find_nested_contexts(node.parent) do |context, nesting|
+          find_nested_example_groups(node.parent) do |example_group, nesting|
             self.max = nesting
             add_offense(
-              context.send_node,
+              example_group.send_node,
               message: message(nesting)
             )
           end
@@ -111,13 +109,14 @@ module RuboCop
 
         private
 
-        def find_nested_contexts(node, nesting: 1, &block)
-          find_contexts(node) do |nested_context|
-            yield(nested_context, nesting) if nesting > max_nesting
+        def find_nested_example_groups(node, nesting: 1, &block)
+          example_group = example_group?(node)
+          yield node, nesting if example_group && nesting > max_nesting
 
-            nested_context.each_child_node do |child|
-              find_nested_contexts(child, nesting: nesting + 1, &block)
-            end
+          next_nesting = example_group ? nesting + 1 : nesting
+
+          node.each_child_node(:block, :begin) do |child|
+            find_nested_example_groups(child, nesting: next_nesting, &block)
           end
         end
 
--- a/spec/rubocop/cop/rspec/nested_groups_spec.rb
+++ b/spec/rubocop/cop/rspec/nested_groups_spec.rb
@@ -76,4 +76,20 @@ RSpec.describe RuboCop::Cop::RSpec::NestedGroups, :config do
         ).to_stderr
     end
   end
+
+  it 'counts nesting correctly when non-spec nesting' do
+    expect_offense(<<-RUBY)
+      describe MyClass do
+        context 'when foo' do
+          context 'when bar' do
+            [].each do |i|
+              context 'when baz' do
+              ^^^^^^^^^^^^^^^^^^ Maximum example group nesting exceeded [4/3].
+              end
+            end
+          end
+        end
+      end
+    RUBY
+  end
 end


lib/rubocop/cop/rspec/nested_groups.rb
```diff
@@ -111,16 +112,19 @@ def on_top_level_describe(node, _args)
...
 
-            nested_context.each_child_node do |child|
-              find_nested_contexts(child, nesting: nesting + 1, &block)
-            end
+            yield(nested_context, nesting) if nesting > max_nesting
           end
         end
 
+        def context_nesting_count(node)
+          parents = node.ancestors.select { |a| context?(a) }
```
pirj, line 124: Maybe count directly?

```diff
- parents = node.ancestors.select { |a| context?(a) }
+ parents = node.ancestors.count { |a| context?(a) }
```

pirj, line 124: It's a minor gain, but maybe only select `:block` ancestors?
```
node.each_ancestor(:block).count { |ancestor| context?(ancestor) }
```

lib/rubocop/cop/rspec/nested_groups.rb
```diff
@@ -98,6 +98,7 @@ class NestedGroups < Cop
           'deprecated in favor of `Max`. Please use that instead.'
 
         def_node_search :find_contexts, ExampleGroups::ALL.block_pattern
+        def_node_matcher :context?, ExampleGroups::ALL.block_pattern
```
pirj, line 101: `context` is slightly misleading, maybe `example_group?`?

lib/rubocop/cop/rspec/nested_groups.rb
```diff
@@ -97,30 +97,34 @@ class NestedGroups < Cop
...
         end
 
         private
 
-        def find_nested_contexts(node, nesting: 1, &block)
-          find_contexts(node) do |nested_context|
-            yield(nested_context, nesting) if nesting > max_nesting
+        def find_nested_example_groups(node)
+          find_example_groups(node) do |example_group|
+            nesting = nesting_count(example_group)
```
Darhazer, line 117: Do you need to count upwards or we can keep the count while going down the tree?
It's an open question - the code is fine and I guess there is no much performance difference, but still, seem like recursive search + count is the standard way to achieve this. Maybe just not feasible with the new implementation?

lib/rubocop/cop/rspec/nested_groups.rb
```diff
@@ -97,30 +97,34 @@ class NestedGroups < Cop
...
           'deprecated in favor of `Max`. Please use that instead.'
 
-        def_node_search :find_contexts, ExampleGroups::ALL.block_pattern
+        def_node_search :find_example_groups, ExampleGroups::ALL.block_pattern
+        def_node_matcher :example_group?, ExampleGroups::ALL.block_pattern
 
         def on_top_level_describe(node, _args)
-          find_nested_contexts(node.parent) do |context, nesting|
+          find_nested_example_groups(node.parent) do |example_group, nesting|
             self.max = nesting
```
Darhazer, line 105: I wonder if this works correctly, e.g. if there are two example groups on the first level, and the first one has a larger number of nested groups, would the max be set to that number, or to the nesting level of the last offending group?

andrykonchin: It should work correctly as far as we don't track nesting level and calculate it for every time:

```ruby
        def nesting_count(node)
          count = node.each_ancestor(:block).count { |n| example_group?(n) }
          count + 1
        end
```

lib/rubocop/cop/rspec/nested_groups.rb
andrykonchin: There is an issue with tracking counter manually with recursion. There may be user defined blocks which should be ignored:

```ruby
describe 'foo' do
  [].each do |i|
    context 'bar' do
    end
  end
end
```

Actually solution with tracking shouldn't be too tricky

lib/rubocop/cop/rspec/nested_groups.rb
andrykonchin: Maybe I don't understand you well. What in your opinion may work incorrectly?

andrykonchin: Hmm, not sure how this `self.max` value is used but looks like setting a value smaller that previous one is handled correctly:

```ruby
      def max=(value)
        cfg = config_to_allow_offenses
        cfg[:exclude_limit] ||= {}
        current_max = cfg[:exclude_limit][max_parameter_name]
        value = [current_max, value].max if current_max.                # <====
        cfg[:exclude_limit][max_parameter_name] = value
      end
```

https://github.com/rubocop-hq/rubocop/blob/master/lib/rubocop/cop/mixin/configurable_max.rb#L10-L16

Darhazer: Yes, it's used when you run `--auto-gen-config`
Thank you for checking it though. I was going to verify myself, as it's not in the scope of the PR but I just noticed while reviewing the code

lib/rubocop/cop/rspec/nested_groups.rb
Darhazer: If it makes code more complicated, let's keep it this way

lib/rubocop/cop/rspec/nested_groups.rb
```diff
@@ -97,30 +97,34 @@ class NestedGroups < Cop
...
-            nested_context.each_child_node do |child|
-              find_nested_contexts(child, nesting: nesting + 1, &block)
-            end
+            yield(example_group, nesting) if nesting > max_nesting
           end
         end
 
+        def nesting_count(node)
+          count = node.each_ancestor(:block).count { |n| example_group?(n) }
+          count + 1
```
pirj, line 125: It's really only a matter of personal preference, but maybe:
```ruby
node.each_ancestor(:block).count { |n| example_group?(n) } + 1
```
it's pretty explanatory what it does.

I also see Maxim proposing `count(&method(:example_group?))`, or insisting on a more explanatory variable name for `n`, e.g. `block` (which reflects what it is, but provokes confusion with `&block`).

lib/rubocop/cop/rspec/nested_groups.rb
andrykonchin: Checked this approach. Seems pretty simple.

lib/rubocop/cop/rspec/nested_groups.rb
```diff
@@ -97,27 +97,25 @@ class NestedGroups < Cop
...
 
         private
 
-        def find_nested_contexts(node, nesting: 1, &block)
-          find_contexts(node) do |nested_context|
-            yield(nested_context, nesting) if nesting > max_nesting
+        def find_nested_example_groups(node, nesting: 1, &block)
+          yield node, nesting if example_group?(node) && nesting > max_nesting
+
+          next_nesting = example_group?(node) ? nesting + 1 : nesting
```
pirj, line 115: We seem to call `example_group?(node)` twice. WDYT of:
```ruby
          if example_group?(node)
            yield node, nesting if nesting > max_nesting
            nesting = nesting + 1
          end

            find_nested_example_groups(child, nesting: nesting, &block)
          end
```

lib/rubocop/cop/rspec/nested_groups.rb
```diff
@@ -97,27 +97,25 @@ class NestedGroups < Cop
...
 
-            nested_context.each_child_node do |child|
-              find_nested_contexts(child, nesting: nesting + 1, &block)
-            end
+          node.each_child_node do |child|
```
pirj, line 117: It might be possible to squeeze more out of this with `node.each_descendant(:block)`.
Which one would perform better?

andrykonchin: Filtering by node type looks reasonable (but `each_descendant` isn't suitable here because it's recursive but we need here direct children).

There is again one issue. We should handle not only `block` but at least `begin` node as well - if there are several nested `context`s they are wrapped into `begin` node. TBH I am not sure whether there are another edge cases (e.g. `kwbegin`) or we could just filter `block` and `begin` nodes:

```ruby
node.each_child_node(:block, :begin) do |child|
```

andrykonchin: It has impressive impact - decreases the cop share from 3.3% to 1.8%.

There is no any difference between generated offenses with and without with filtering (checked on GitLab specs) but I still hesitate.

pirj: I'm a bit lost. What change reduced the share from 3.3% to 1.8%?

andrykonchin: I meant additional filtering of `block` and `begin` nodes (what you've proposed) decreased cop time from 3.3% to 1.8%.

andrykonchin: ping @Darhazer @bquorning What do you think?

bquorning: Yeah, I think using `node.each_child_node(:block, :begin)` is fine. If we are missing some edge cases, we can fix them as bugs. Provided of course that anyone will discover them.

lib/rubocop/cop/rspec/nested_groups.rb
```diff
@@ -97,27 +97,26 @@ class NestedGroups < Cop
...
         end
 
         private
 
-        def find_nested_contexts(node, nesting: 1, &block)
-          find_contexts(node) do |nested_context|
-            yield(nested_context, nesting) if nesting > max_nesting
+        def find_nested_example_groups(node, nesting: 1, &block)
+          example_group = example_group?(node)
+          yield node, nesting if example_group && nesting > max_nesting
```
bquorning, line 114: Couldn’t we optimise a bit by doing an early return if `nesting <= max_nesting`?

andrykonchin: Hmm. Not sure. Early return in this case means that none of child nodes will be processed and checked.

lib/rubocop/cop/rspec/nested_groups.rb
andrykonchin: Done

lib/rubocop/cop/rspec/nested_groups.rb
bquorning: Yeah, that’s probably not what we want to happen. :-)

andrykonchin:
Done

bquorning:
Hey @andrykonchin. I just wanted to let you know that I tried running
```
time bundle exec rubocop --cache false --only RSpec/DescribeClass,FactoryBot/AttributeDefinedStatically,RSpec/InstanceVariable,RSpec/LeakyConstantDeclaration,RSpec/LetSetup,RSpec/NestedGroups,RSpec/ReturnFromStub,RSpec/SubjectStub -- spec
```
on the ~1400 spec files in my main work repository (`--only` running the cops that have changed between v1.40 and master branch).

Time spent using rubocop-rspec v1.40.0: 1:20.71 total
Time spent using rubocop-rspec master branch (6e1d698a52bb4f151d46f5e8fe6bd2af4fa75122): 19.561 total

The time improvement is *absolutely* incredible, much better than I had hoped for. Thank you so much ❤️

bquorning:
Ahh, sorry. I was measuring the difference between 1.39.0 and master, not 1.40.0. But the speed improvement (probably in SubjectStub) is still your work.

The runtime using v1.40.0 is “26.552 total” on my 1400 files. Still a significant improvement to master branch though.

andrykonchin:
Yeah, `SubjectStub` cop optimization definitely was a big win. All the other recent optimizations should have fixed most of inefficient RSpec cops. So now every RSpec cop takes less than 1%, at least if run Rubocop on the GitLab test suit:

```
stackprof tmp/stackprof-cpu-gitlab.master.with-rspec.test.fix.dump  --method 'RuboCop::Cop::Commissioner#trigger_responding_cops' | grep RSpec
    1737  (    1.9%)  RuboCop::RSpec::TopLevelDescribe#on_send
    1120  (    1.2%)  RuboCop::RSpec::TopLevelGroup#on_block
     907  (    1.0%)  RuboCop::Cop::RSpec::DescribedClass#on_block
     805  (    0.9%)  RuboCop::Cop::RSpec::MultipleSubjects#on_block
     790  (    0.9%)  RuboCop::Cop::RSpec::RepeatedDescription#on_block
     721  (    0.8%)  RuboCop::Cop::RSpec::RepeatedExample#on_block
     699  (    0.8%)  RuboCop::Cop::RSpec::MultipleExpectations#on_block
     658  (    0.7%)  RuboCop::RSpec::TopLevelGroup#on_block
     592  (    0.6%)  RuboCop::Cop::RSpec::ScatteredSetup#on_block
...
```

pirj:
I might be reading it wrong, does it appear twice?
```
    1120  (    1.2%)  RuboCop::RSpec::TopLevelGroup#on_block
     658  (    0.7%)  RuboCop::RSpec::TopLevelGroup#on_block
```

andrykonchin:
It's the way how stackprof shows different callers - `TopLevelGroup` is used in two cops - InstanceVariable and SubjectStub:

```
$ stackprof tmp/stackprof-cpu-gitlab.master.with-rspec.test.fix.dump  --method 'RuboCop::RSpec::TopLevelGroup#on_block'

RuboCop::RSpec::TopLevelGroup#on_block (/Users/andrykonchin/projects/rubocop-rspec/lib/rubocop/rspec/top_level_group.rb:13)
  samples:    42 self (0.0%)  /    658 total (0.4%)
  callers:
     658  (  100.0%)  RuboCop::Cop::Commissioner#trigger_responding_cops
  callees (616 total):
     446  (   72.4%)  RuboCop::Cop::RSpec::InstanceVariable#on_top_level_group
     170  (   27.6%)  RuboCop::RSpec::TopLevelGroup#top_level_group?
  code:
                                  |    13  |       def on_block(node)
   40    (0.0%) /    40   (0.0%)  |    14  |         return unless respond_to?(:on_top_level_group)
  172    (0.1%) /     2   (0.0%)  |    15  |         return unless top_level_group?(node)
                                  |    16  |
  446    (0.3%)                   |    17  |         on_top_level_group(node)
                                  |    18  |       end

RuboCop::RSpec::TopLevelGroup#on_block (/Users/andrykonchin/projects/rubocop-rspec/lib/rubocop/rspec/top_level_group.rb:13)
  samples:    30 self (0.0%)  /   1120 total (0.7%)
  callers:
    1120  (  100.0%)  RuboCop::Cop::Commissioner#trigger_responding_cops
  callees (1090 total):
     968  (   88.8%)  RuboCop::Cop::RSpec::SubjectStub#on_top_level_group
     122  (   11.2%)  RuboCop::RSpec::TopLevelGroup#top_level_group?
  code:
                                  |    13  |       def on_block(node)
   27    (0.0%) /    27   (0.0%)  |    14  |         return unless respond_to?(:on_top_level_group)
  125    (0.1%) /     3   (0.0%)  |    15  |         return unless top_level_group?(node)
                                  |    16  |
  968    (0.6%)                   |    17  |         on_top_level_group(node)
                                  |    18  |       end
```

andrykonchin:
Hmm... So looks like these statistics aren't aggregated like I thought before :sweat: :

```
    1737  (    1.9%)  RuboCop::RSpec::TopLevelDescribe#on_send
    1120  (    1.2%)  RuboCop::RSpec::TopLevelGroup#on_block
```

PR #951, opened 2020-06-28
Performance. RSpec/LetSetup 

andrykonchin:
Optimize RSpec/LetSetup cop and #on_block callback.

### Changes

- avoid excessive iteration over **all** the `let!` nodes nested into current example group
- fixed handling shared example groups

The approach is to iterate over only **direct** child `let!` nodes of a current example group.

### Performance measurements

Timing for RuboCop::Cop::RSpec::LetSetup#on_block is changed from 10.2% to 1.3%.

```
stackprof tmp/stackprof-cpu-gitlab.master.with-rubocop-rspec.LetSetup.dump --method 'RuboCop::Cop::RSpec::LetSetup#on_block'
RuboCop::Cop::RSpec::LetSetup#on_block (/Users/andrykonchin/projects/rubocop-rspec/lib/rubocop/cop/rspec/let_setup.rb:37)
  samples:     1 self (0.0%)  /   4003 total (10.2%)
  callers:
    4003  (  100.0%)  RuboCop::Cop::Commissioner#trigger_responding_cops
     144  (    3.6%)  RuboCop::Cop::RSpec::LetSetup#unused_let_bang
  callees (4002 total):
    3895  (   97.3%)  RuboCop::Cop::RSpec::LetSetup#unused_let_bang
     144  (    3.6%)  RuboCop::Cop::Cop#add_offense
     107  (    2.7%)  RuboCop::RSpec::Language::NodePattern#example_group?
  code:
                                  |    37  |           (block $(send nil? :let! (sym $_)) args ...)
  108    (0.3%) /     1   (0.0%)  |    38  |         PATTERN
                                  |    39  |
                                  |    40  |         def_node_search :method_called?, '(send nil? %)'
 3895    (9.9%)                   |    41  |
  144    (0.4%)                   |    42  |         def on_block(node)
                                  |    43  |           return unless example_group?(node)
```
</details>

<details><summary>After</summary>

```
stackprof tmp/stackprof-cpu-gitlab.master.with-rubocop-rspec.LetSetup.2.dump --method 'RuboCop::Cop::RSpec::LetSetup#on_block'
RuboCop::Cop::RSpec::LetSetup#on_block (/Users/andrykonchin/projects/rubocop-rspec/lib/rubocop/cop/rspec/let_setup.rb:42)
  samples:     1 self (0.0%)  /    440 total (1.3%)
  callers:
     440  (  100.0%)  RuboCop::Cop::Commissioner#trigger_responding_cops
     113  (   25.7%)  RuboCop::Cop::RSpec::LetSetup#unused_let_bang
  callees (439 total):
     361  (   82.2%)  RuboCop::Cop::RSpec::LetSetup#unused_let_bang
     113  (   25.7%)  RuboCop::Cop::Cop#add_offense
      78  (   17.8%)  RuboCop::RSpec::Language::NodePattern#example_group?
  code:
                                  |    42  |         def on_block(node)
   79    (0.2%) /     1   (0.0%)  |    43  |           return unless example_group?(node)
                                  |    44  |
                                  |    45  |           #puts node
  361    (1.1%)                   |    46  |           unused_let_bang(node) do |let|
  113    (0.3%)                   |    47  |             add_offense(let)
                                  |    48  |           end
```
</details>

### Measurements approach

```
bundle exec exe/rubocop --cache false --out gitlab-specs.out --force-default-config  --require rubocop-rspec --only RSpec/LetSetup ../rubocop-profiling-examples/gitlabhq/spec
```

---

* [x] Added tests.
* [ ] Updated documentation.

### DIFF
--- a/lib/rubocop/cop/rspec/let_setup.rb
+++ b/lib/rubocop/cop/rspec/let_setup.rb
@@ -28,14 +28,20 @@ module RuboCop
       class LetSetup < Cop
         MSG = 'Do not use `let!` to setup objects not referenced in tests.'
 
-        def_node_search :let_bang, <<-PATTERN
+        def_node_matcher :example_or_shared_group_or_including?,
+                         (
+                           ExampleGroups::ALL + SharedGroups::ALL +
+                           Includes::ALL
+                         ).block_pattern
+
+        def_node_matcher :let_bang, <<-PATTERN
           (block $(send nil? :let! (sym $_)) args ...)
         PATTERN
 
         def_node_search :method_called?, '(send nil? %)'
 
         def on_block(node)
-          return unless example_group?(node)
+          return unless example_or_shared_group_or_including?(node)
 
           unused_let_bang(node) do |let|
             add_offense(let)
@@ -45,8 +51,14 @@ module RuboCop
         private
 
         def unused_let_bang(node)
-          let_bang(node) do |method_send, method_name|
+          child_let_bang(node) do |method_send, method_name|
             yield(method_send) unless method_called?(node, method_name)
+          end
+        end
+
+        def child_let_bang(node, &block)
+          RuboCop::RSpec::ExampleGroup.new(node).lets.each do |let|
+            let_bang(let, &block)
           end
         end
       end
--- a/spec/rubocop/cop/rspec/let_setup_spec.rb
+++ b/spec/rubocop/cop/rspec/let_setup_spec.rb
@@ -63,4 +63,53 @@ RSpec.describe RuboCop::Cop::RSpec::LetSetup do
       end
     RUBY
   end
+
+  it 'complains when let! is used and not referenced in shared example group' do
+    expect_offense(<<-RUBY)
+      shared_context 'foo' do
+        let!(:bar) { baz }
+        ^^^^^^^^^^ Do not use `let!` to setup objects not referenced in tests.
+
+        it 'does not use bar' do
+          expect(baz).to eq(qux)
+        end
+      end
+    RUBY
+  end
+
+  it 'complains when let! used in shared example including' do
+    expect_offense(<<-RUBY)
+      describe Foo do
+        it_behaves_like 'bar' do
+          let!(:baz) { foobar }
+          ^^^^^^^^^^ Do not use `let!` to setup objects not referenced in tests.
+          let(:a) { b }
+        end
+      end
+    RUBY
+  end
+
+  it 'complains when there is only one nested node into example group' do
+    expect_offense(<<-RUBY)
+      describe Foo do
+        let!(:bar) { baz }
+        ^^^^^^^^^^ Do not use `let!` to setup objects not referenced in tests.
+      end
+    RUBY
+  end
+
+  it 'complains when there is a custom nesting level' do
+    expect_offense(<<-RUBY)
+      describe Foo do
+        [].each do |i|
+          let!(:bar) { i }
+          ^^^^^^^^^^ Do not use `let!` to setup objects not referenced in tests.
+
+          it 'does not use bar' do
+            expect(baz).to eq(qux)
+          end
+        end
+      end
+    RUBY
+  end
 end


lib/rubocop/cop/rspec/let_setup.rb
```diff
@@ -45,10 +45,24 @@ def on_block(node)
...
             yield(method_send) unless method_called?(node, method_name)
           end
         end
+
+        def child_let_bang(block_node)
+          return unless block_node.body
+
+          body = block_node.body
+
+          if body.begin_type?
```
pirj, line 58: Agree with RuboCop here. Maybe
```ruby
return unless block_node.body&.begin_type?
```

lib/rubocop/cop/rspec/let_setup.rb
```diff
@@ -45,10 +45,19 @@ def on_block(node)
...
-          let_bang(node) do |method_send, method_name|
+          child_let_bang(node) do |method_send, method_name|
             yield(method_send) unless method_called?(node, method_name)
           end
         end
+
+        def child_let_bang(block_node)
+          return unless block_node.body&.begin_type?
+
+          block_node.body.each_child_node(:block) do |child|
```
Darhazer, line 56: It would be great if you could `yield` if the passed block yielded, e.g. if there is shorter or more FP way to write the code below

andrykonchin: Do you mean something like this?

```ruby
block_node.body.each_child_node(:block)
  .map { |child| let_bang(child) }
  .reject { |m| m.nil? }
  .each { |m| yield m }
```

Darhazer: yes, or
```
block_node.body.each_child_node(:block)
  .map(&method(:let_bang)
  .compact
  .each { |m| yield m }
```
But would prefer to have it even shorter, or at least without having to use explicit blocks
Ideally something like
`block_node.body.each_child_node(:block).yield_when(&method(:let_bang)`

Anyway, that was just a side note, as I hate to see things like
```
a = something
yield a if a
```

andrykonchin:
Compared generated offenses and found out some difference. Will investigate it so marked the PR as draft.

lib/rubocop/cop/rspec/let_setup.rb
```diff
@@ -45,10 +51,34 @@ def on_block(node)
...
-          let_bang(node) do |method_send, method_name|
+          child_let_bang(node) do |method_send, method_name|
             yield(method_send) unless method_called?(node, method_name)
           end
         end
+
+        def child_let_bang(node)
+          enum = child_block_until_group(node).map(&method(:let_bang))
+
+          enum.compact.each do |node_with_name|
```
Darhazer, line 62: Well, I told I'm missing a feature to make such things beautiful. Actually I think the code is now harder to read

andrykonchin: Indeed. Agree with you. Reverted these changes.

lib/rubocop/cop/rspec/let_setup.rb
```diff
@@ -45,10 +51,33 @@ def on_block(node)
...
+            next if example_or_shared_group_or_including?(child)
+
+            yield child
+            child_block_until_group(child, &blk)
+          end
+        end
+
+        def unwrap_nested_begin(node)
+          node.body.begin_type? ? node.body : node
+        end
```
andrykonchin, line 80: Just realized the whole implementation of `child_let_bang` could be replaced with `RuboCop::RSpec::ExampleGroup`:

```ruby
RuboCop::RSpec::ExampleGroup.new(node).lets.map { |let|  let_bang(let) }.compact
```

https://github.com/rubocop-hq/rubocop-rspec/blob/master/lib/rubocop/rspec/example_group.rb

I am wondering whether it makes sense to check this approach.

pirj: Absolutely. Those helpers keep being under the radar.

andrykonchin: Done. Looks pretty good.

lib/rubocop/cop/rspec/let_setup.rb
```diff
@@ -45,10 +51,19 @@ def on_block(node)
...
         end
+
+        def child_let_bang(node)
+          lets = RuboCop::RSpec::ExampleGroup.new(node).lets
+
+          lets.each do |let|
+            node_with_name = let_bang(let)
+            yield node_with_name if node_with_name
+          end
+        end
```
pirj, line 66: Just an option:
```ruby
        def child_let_bang(node, &block)
          RuboCop::RSpec::ExampleGroup.new(node)
            .lets
            .map { |let| let_bang(let) }
            .compact
            .each(&block)
        end
```
could be cleaner with `filter_map`, but it's Ruby 2.7 only AFAIR.

I think your current implementation is superior.

lib/rubocop/cop/rspec/let_setup.rb
pirj: Sorry, I've missed that this has been discussed already.

lib/rubocop/cop/rspec/let_setup.rb
```diff
@@ -45,10 +51,19 @@ def on_block(node)
...
           end
         end
+
+        def child_let_bang(node)
+          lets = RuboCop::RSpec::ExampleGroup.new(node).lets
+
+          lets.each do |let|
+            node_with_name = let_bang(let)
+            yield node_with_name if node_with_name
+          end
```
bquorning, line 65: Maybe this has been discussed already, but this could be rewritten to get rid of the `nil` check:
```rb
lets.each do |let|
  let_bang(let) do |send_node, method_name|
    yield send_node, method_name
  end
end
```

pirj: Well noticed 👍 
Maybe even
```ruby
lets.each do |let|
  let_bang(let, &block)
end
```

andrykonchin: I don't think so. `RuboCop::RSpec::ExampleGroup#lets` returns both `let` and `let!` but here we need only `let`s with bang. That's why there is the `nil` check.

Darhazer: yeah but `let_bang` would not yield if the node is not matched ;) So there is a difference between passing a block to a node_matcher or using the matcher's return value (which might be nil)

pirj: Right, but we filter this through `let_bang` that yields only when there is something to yield (bang-let found).
Checked both options, specs pass.

My final version of the method, leaving it up to you which option to pick:
```ruby
        def child_let_bang(node, &block)
          RuboCop::RSpec::ExampleGroup.new(node).lets
            .each { |let| let_bang(let, &block) }
        end
```

andrykonchin: Great. Will update the PR soon.

andrykonchin: Done

Darhazer:
Please squash the commits

andrykonchin:
Done

pirj:
Awesome perf-combo! 💯 Thanks a lot.

PR #954, opened 2020-07-07
New RuboCop autocorrect

bquorning:
The autocorrection API was changed in Rubocop v0.87.0 (pull request rubocop-hq/rubocop#7868). This PR makes rubocop-rspec compatible with those changes.

The superclass of `RuboCop::Cop::RSpec::Cop` is changed from `::RuboCop::Cop::Cop` to `::RuboCop::Cop::Base`.

This has a few consequences:

- Our `#message` methods get called with a different argument than before. It *can* be customized by defining a `#callback_argument` method, but I think it is clearer to avoid callbacks altogether.
- Our `#autocorrect` methods don't get called anymore. Instead, we extend `Autocorrector`, and the `corrector` is being yielded when calling `#add_offense`, so the code is mostly moved in there. For some cases, this means that some code can be removed, which is nice. For some cases, it means that the methods get too long, or the code complexity gets too high, and in those cases I chose to just call out to an `#autocorrect` method anyway, but of course passing the `corrector` and any usable local variables along.
- Our dependency of RuboCop is bumped from '>= 0.68.1' to '>= 0.87.0'.

See also https://github.com/rubocop-hq/rubocop/blob/master/docs/modules/ROOT/pages/v1_upgrade_notes.adoc for documentation of many of the changes.

---

* [ ] Added tests.
* [ ] Updated documentation.

### DIFF
--- a/.rubocop_todo.yml
+++ b/.rubocop_todo.yml
@@ -14,4 +14,4 @@ InternalAffairs/MethodNameEqual:
 # Offense count: 1
 # Configuration parameters: CountComments.
 Metrics/ClassLength:
-  Max: 104
+  Max: 106
--- a/lib/rubocop/cop/rspec/align_left_let_brace.rb
+++ b/lib/rubocop/cop/rspec/align_left_let_brace.rb
@@ -44,8 +44,7 @@ module RuboCop
         private
 
         def token_aligner
-          @token_aligner ||=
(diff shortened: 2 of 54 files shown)


bquorning:
We probably need some documentation changes too; I haven’t looked into that yet.

PR #955, opened 2020-07-07
New RuboCop autocorrect

bquorning:
The autocorrection API was changed in Rubocop v0.87.0 (pull request rubocop-hq/rubocop#7868). This PR makes rubocop-rspec compatible with those changes.

The superclass of `RuboCop::Cop::RSpec::Cop` is changed from `::RuboCop::Cop::Cop` to `::RuboCop::Cop::Base`.

This has a few consequences:

- Our `#message` methods get called with a different argument than before. It *can* be customized by defining a `#callback_argument` method, but I think it is clearer to avoid callbacks altogether.
- Our `#autocorrect` methods don't get called anymore. Instead, we extend `Autocorrector`, and the `corrector` is being yielded when calling `#add_offense`, so the code is mostly moved in there. For some cases, this means that some code can be removed, which is nice. For some cases, it means that the methods get too long, or the code complexity gets too high, and in those cases I chose to just call out to an `#autocorrect` method anyway, but of course passing the `corrector` and any usable local variables along.
- Our dependency of RuboCop is bumped from '>= 0.68.1' to '>= 0.87.0'.

See also https://github.com/rubocop-hq/rubocop/blob/master/docs/modules/ROOT/pages/v1_upgrade_notes.adoc for documentation of many of the changes.

We probably need some documentation changes too; I haven’t looked into that yet.

---

* [ ] Added tests.
* [ ] Updated documentation.

### DIFF
--- a/lib/rubocop/cop/rspec/align_left_let_brace.rb
+++ b/lib/rubocop/cop/rspec/align_left_let_brace.rb
@@ -18,34 +18,27 @@ module RuboCop
       #     let(:a)      { b }
       #
       class AlignLeftLetBrace < Cop
+        extend AutoCorrector
+
         MSG = 'Align left let brace'
 
         def self.autocorrect_incompatible_with
           [Layout::ExtraSpacing]
         end
 
-        def investigate(_processed_source)
+        def on_new_investigation
           return if processed_source.blank?
 
-          token_aligner.offending_tokens.each do |let|
-            add_offense(let, location: :begin)
-          end
-        end
-
-        def autocorrect(let)
-          lambda do |corrector|
-            corrector.insert_before(
-              let.loc.begin,
-              token_aligner.indent_for(let)
-            )
-          end
-        end
-
-        private
-
-        def token_aligner
-          @token_aligner ||=
+          token_aligner =
             RuboCop::RSpec::AlignLetBrace.new(processed_source.ast, :begin)
+
+          token_aligner.offending_tokens.each do |let|
+            add_offense(let.loc.begin) do |corrector|
+              corrector.insert_before(
+                let.loc.begin, token_aligner.indent_for(let)
+              )
+            end
+          end
         end
       end
     end
--- a/lib/rubocop/cop/rspec/capybara/current_path_expectation.rb
+++ b/lib/rubocop/cop/rspec/capybara/current_path_expectation.rb
@@ -24,6 +24,8 @@ module RuboCop
         #   expect(page).to have_current_path(/widgets/)
         #
         class CurrentPathExpectation < Cop
+          extend AutoCorrector
+
           MSG = 'Do not set an RSpec expectation on `current_path` in ' \
                 'Capybara feature specs - instead, use the ' \
                 '`have_current_path` matcher on `page`'
@@ -47,30 +49,30 @@ module RuboCop
 
           def on_send(node)
             expectation_set_on_current_path(node) do
-              add_offense(node, location: :selector)
-            end
-          end
+              add_offense(node.loc.selector) do |corrector|
+                next unless node.chained?
 
-          def autocorrect(node)
-            lambda do |corrector|
-              return unless node.chained?
-
-              as_is_matcher(node.parent) do |to_sym, matcher_node|
-                rewrite_expectation(corrector, node, to_sym, matcher_node)
-              end
-
-              regexp_str_matcher(node.parent) do |to_sym, matcher_node, regexp|
-                rewrite_expectation(corrector, node, to_sym, matcher_node)
-                convert_regexp_str_to_literal(corrector, matcher_node, regexp)
+                autocorrect(corrector, node)
               end
             end
           end
 
           private
 
+          def autocorrect(corrector, node)
+            as_is_matcher(node.parent) do |to_sym, matcher_node|
+              rewrite_expectation(corrector, node, to_sym, matcher_node)
+            end
+
+            regexp_str_matcher(node.parent) do |to_sym, matcher_node, regexp|
+              rewrite_expectation(corrector, node, to_sym, matcher_node)
+              convert_regexp_str_to_literal(corrector, matcher_node, regexp)
+            end
+          end
+
           def rewrite_expectation(corrector, node, to_symbol, matcher_node)
             current_path_node = node.first_argument
-            corrector.replace(current_path_node.loc.expression, 'page')
+            corrector.replace(current_path_node, 'page')
             corrector.replace(node.parent.loc.selector, 'to')
             matcher_method = if to_symbol == :to
                                'have_current_path'
@@ -84,7 +86,7 @@ module RuboCop
           def convert_regexp_str_to_literal(corrector, matcher_node, regexp_str)
             str_node = matcher_node.first_argument
             regexp_expr = Regexp.new(regexp_str).inspect
-            corrector.replace(str_node.loc.expression, regexp_expr)
+            corrector.replace(str_node, regexp_expr)
           end
 
           # `have_current_path` with no options will include the querystring
@@ -97,7 +99,7 @@ module RuboCop
             return if %i[regexp str].include?(expectation_last_child.type)
 
             corrector.insert_after(
-              expectation_last_child.loc.expression,
+              expectation_last_child,
               ', ignore_query: true'
             )
           end
--- a/lib/rubocop/cop/rspec/capybara/feature_methods.rb
+++ b/lib/rubocop/cop/rspec/capybara/feature_methods.rb
@@ -41,6 +41,8 @@ module RuboCop
         #     end
         #   end
         class FeatureMethods < Cop
+          extend AutoCorrector
+
           MSG = 'Use `%<replacement>s` instead of `%<method>s`.'
 
           # https://git.io/v7Kwr
@@ -71,18 +73,15 @@ module RuboCop
             feature_method(node) do |send_node, match|
               next if enabled?(match)
 
-              add_offense(
-                send_node,
-                location: :selector,
-                message: format(MSG, method: match, replacement: MAP[match])
-              )
+              add_offense(send_node.loc.selector) do |corrector|
+                corrector.replace(send_node.loc.selector, MAP[match].to_s)
+              end
             end
           end
 
-          def autocorrect(node)
-            lambda do |corrector|
-              corrector.replace(node.loc.selector, MAP[node.method_name].to_s)
-            end
+          def message(range)
+            name = range.source.to_sym
+            format(MSG, method: name, replacement: MAP[name])
           end
 
           private
--- a/lib/rubocop/cop/rspec/context_method.rb
+++ b/lib/rubocop/cop/rspec/context_method.rb
@@ -24,6 +24,8 @@ module RuboCop
       #     # ...
       #   end
       class ContextMethod < Cop
+        extend AutoCorrector
+
         MSG = 'Use `describe` for testing methods.'
 
         def_node_matcher :context_method, <<-PATTERN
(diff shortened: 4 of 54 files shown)


lib/rubocop/cop/rspec/empty_hook.rb
```diff
@@ -33,15 +34,11 @@ class EmptyHook < Cop
...
-          end
-        end
-
-        def autocorrect(node)
-          lambda do |corrector|
-            block = node.parent
-            range = range_with_surrounding_space(range: block.loc.expression)
-            corrector.remove(range)
+            add_offense(hook) do |corrector|
+              block = hook.parent
```
Darhazer, line 38: block is actually just the `node`, so we can remove this line and use node directly

lib/rubocop/cop/rspec/example_wording.rb
```diff
@@ -53,16 +55,12 @@ def on_block(node)
...
-          lambda do |corrector|
-            corrector.replace(docstring(node), replacement_text(node))
-          end
-        end
-
         private
 
         def add_wording_offense(node, message)
-          add_offense(node, location: docstring(node), message: message)
+          add_offense(docstring(node), message: message) do |corrector|
```
Darhazer, line 61: maybe save docstring(node) as a variable first and reuse?

lib/rubocop/cop/rspec/factory_bot/create_list.rb
```diff
@@ -51,26 +52,21 @@ def on_block(node)
...
 
           def on_send(node)
             return unless style == :n_times
 
             factory_list_call(node) do |_receiver, _factory, count, _|
               add_offense(
-                node,
-                location: :selector,
+                node.loc.selector,
                 message: format(MSG_N_TIMES, number: count)
```
Darhazer, line 66: lets use a variable for the message

lib/rubocop/cop/rspec/context_method.rb
```diff
@@ -32,13 +34,9 @@ class ContextMethod < Cop
...
           context_method(node) do |context|
-            add_offense(context)
-          end
-        end
-
-        def autocorrect(node)
-          lambda do |corrector|
-            corrector.replace(node.parent.loc.selector, 'describe')
+            add_offense(context) do |corrector|
+              corrector.replace(context.parent.loc.selector, 'describe')
```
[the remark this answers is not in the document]
bquorning: Using `node.send_node.loc.selector` instead of `context.parent.loc.selector` works.

lib/rubocop/cop/rspec/capybara/feature_methods.rb
```diff
@@ -72,16 +74,11 @@ def on_block(node)
               next if enabled?(match)
 
               add_offense(
-                send_node,
-                location: :selector,
+                send_node.loc.selector,
                 message: format(MSG, method: match, replacement: MAP[match])
```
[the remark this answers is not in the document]
bquorning: `#message` always receives a range now, but I can do this:
```rb
          def message(range)
            match = range.source.to_sym
            format(MSG, method: match, replacement: MAP[match])
          end
```

lib/rubocop/cop/rspec/capybara/current_path_expectation.rb
```diff
@@ -47,30 +49,30 @@ class CurrentPathExpectation < Cop
...
-          def autocorrect(node)
-            lambda do |corrector|
-              return unless node.chained?
+          private
 
-              as_is_matcher(node.parent) do |to_sym, matcher_node|
-                rewrite_expectation(corrector, node, to_sym, matcher_node)
-              end
+          def autocorrect(corrector, node)
+            return unless node.chained?
```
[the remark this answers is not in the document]
bquorning: This should be `next unless node.chained?` inside the `add_offense` block, right?

lib/rubocop/cop/rspec/align_left_let_brace.rb
```diff
@@ -18,34 +18,31 @@ module RSpec
...
+            end
           end
         end
 
         private
 
         def token_aligner
-          @token_aligner ||=
-            RuboCop::RSpec::AlignLetBrace.new(processed_source.ast, :begin)
+          RuboCop::RSpec::AlignLetBrace.new(processed_source.ast, :begin)
```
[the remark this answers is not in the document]
bquorning: Yeah, just removing the memoization probably isn’t the best approach. How about I just move the instantiation into the `#on_new_investigation` method?

bquorning:
CodeClimate complains about some added complexity, which should probably be addressed. But I think it’s important that we get a new release out pretty quickly, because right now we are incompatible with the most recent version of RuboCop.

marcandre:
There's a lot of activity here 👍
Let me know if I should review this now, or ping me when the dust settles 😄

lib/rubocop/cop/rspec/factory_bot/create_list.rb
```diff
@@ -51,26 +52,19 @@ def on_block(node)
...
-            add_offense(node.send_node, message: MSG_CREATE_LIST)
+            add_offense(node.send_node, message: MSG_CREATE_LIST) do |corrector|
+              CreateListCorrector.new(node.send_node).call(corrector)
+            end
           end
 
           def on_send(node)
             return unless style == :n_times
 
             factory_list_call(node) do |_receiver, _factory, count, _|
```
Darhazer, line 63: it seems that we never use the other captures, so we can just remove them from the pattern

lib/rubocop/cop/rspec/hook_argument.rb
```diff
@@ -100,10 +94,12 @@ def check_implicit(method_send)
           return if implicit_style?
 
           add_offense(
```
Darhazer, line 96: Can you make this call one line as well?

lib/rubocop/cop/rspec/implicit_expect.rb
```diff
@@ -55,19 +56,12 @@ def on_send(node) # rubocop:disable Metrics/MethodLength
             opposite_style_detected
 
             add_offense(
```
Darhazer, line 58: one line if possible :pray:

lib/rubocop/cop/rspec/leading_subject.rb
```diff
@@ -43,27 +45,25 @@ def on_block(node)
...
-            RuboCop::RSpec::Corrector::MoveNode.new(
-              node, corrector, processed_source
-            ).move_before(first_node)
-          end
-        end
-
         private
 
+        def autocorrect(corrector, node)
+          first_node = find_first_offending_node(node)
```
Darhazer, line 61: In a future refactoring, we can save the offending node in the check_previous_node method and just pass it to the autocorrect

lib/rubocop/cop/rspec/predicate_matcher.rb
```diff
@@ -301,6 +270,7 @@ def replacement_matcher(node)
       #   # good - the above code is rewritten to it by this cop
       #   expect(foo.something?).to be_truthy
       class PredicateMatcher < Cop
+        extend AutoCorrector
```
Darhazer, line 273: Why is this needed?

lib/rubocop/cop/rspec/return_from_stub.rb
```diff
@@ -59,24 +60,17 @@ def on_block(node)
...
-            BlockBodyCorrector.new(node)
-          end
-        end
-
         private
 
         def check_and_return_call(node)
           and_return_value(node) do |and_return, args|
             unless dynamic?(args)
               add_offense(
```
Darhazer, line 68: one line :pray:

lib/rubocop/cop/rspec/return_from_stub.rb
```diff
@@ -85,10 +79,11 @@ def check_block_body(block)
           body = block.body
           unless dynamic?(body) # rubocop:disable Style/GuardClause
             add_offense(
```
Darhazer, line 81: one line :pray:

lib/rubocop/cop/rspec/it_behaves_like.rb
```diff
@@ -28,14 +29,12 @@ class ItBehavesLike < Cop
 
         def on_send(node)
           example_inclusion_offense(node, alternative_style) do
-            add_offense(node)
+            add_offense(node) do |corrector|
```
mobilutz, line 32: Shouldn't there be a `message` parameter to `add_offense`?
```suggestion
            add_offense(node, message: format(MSG, replacement: style, original: alternative_style)) do |corrector|
```

lib/rubocop/cop/rspec/not_to_not.rb
```diff
@@ -24,14 +25,12 @@ class NotToNot < Cop
 
         def on_send(node)
           not_to_not_offense(node, alternative_style) do
-            add_offense(node, location: :selector)
+            add_offense(node.loc.selector) do |corrector|
```
mobilutz, line 28: The `add_offense` should have the `message` parameter:
```suggestion
            add_offense(node.loc.selector, message: format(MSG, replacement: style, original: alternative_style)) do |corrector|
```

bquorning:
@Darhazer I won’t have time to work on this today. If you have time could you please squash, merge and release? If not, I’ll look at it tomorrow.

lib/rubocop/cop/rspec/not_to_not.rb
bquorning: If no `message` parameter is passed, RuboCop will call the `#message` method which returns a string. I prefer the more explicit version, but the callback approach still works.

mobilutz: Ok, got it.
I was just checking the rest of the Changelog, and almost everywhere the code is now calling `add_offense` with the `message` parameter, except when there is no interpolation, and the `MSG` constant is being used automatically.

Maybe this is something, that can be aligned afterwards as well.
So that the code is in line with the expectation from [`add_offense`](https://github.com/rubocop-hq/rubocop/blob/v0.87.1/lib/rubocop/cop/internal_affairs/redundant_message_argument.rb#L12-L21)

lib/rubocop/cop/rspec/factory_bot/create_list.rb
bquorning: I’ll fix that in a separate PR.

lib/rubocop/cop/rspec/predicate_matcher.rb
bquorning: A cop must `extend AutoCorrector` to be able to autocorrect: https://github.com/rubocop-hq/rubocop/blob/v0.87.1/lib/rubocop/cop/base.rb#L304

marcandre:
Wow, impressive PR 💪!

bquorning:
Thanks. I hope I got everything right 😅

lib/rubocop/cop/rspec/predicate_matcher.rb
Darhazer: Oh, I didn't notice this is the actual cop, as it is like 270 lines below the start of the code, though it was a helper class

Darhazer: Still somehow confusing, as the offsense/correction is provided by the included modules. Do we need the `autocorrect` method in this class?

marcandre: Shouldn't it be `PredicateMatcher < Base`?

If is it reused, the `module` should probably do it itself:
```
def self.included(base)
  base.extend AutoCorrector
end
```

lib/rubocop/cop/rspec/factory_bot/create_list.rb
bquorning: See #959.

pirj:
Sleek! 👏 
Thanks a lot guys.

lib/rubocop/cop/rspec/leading_subject.rb
bquorning: #960

PR #956, opened 2020-07-07
RSpec aliases runtime matchers

sl4vr:
This is an attempt of implementation of issue https://github.com/rubocop-hq/rubocop-rspec/issues/333 and evolution of PR https://github.com/rubocop-hq/rubocop-rspec/pull/921. Allowing to configure 3rd partly DSLs as valid RSpec syntax.

Configurable language elements:
- [x] Hooks
- [x] Examples
- [x] Hook scopes
- [x] Includes
- [x] Example groups
- [x] Shared groups
- [x] Helpers
- [x] Subject
- [x] Expectations
- [x] Runners

---

* [x] Added tests.
* [ ] Updated documentation.


sl4vr:
Looks like smoke tests fail, need to rewrite `RuntimeMacros` in a same manner like RuboCop's `Macros` are made.

lib/rubocop/rspec/language/node_pattern.rb
```diff
@@ -19,11 +20,35 @@ module NodePattern
...
         def_node_matcher :subject?, Subject::ALL.block_pattern
+
+        def_runtime_node_matcher :hook? do
+          hook_selectors.block_pattern
+        end
+
+        def hook_selectors
+          custom_method_names = config
+            .for_all_cops
+            .fetch('RSpec', {})
```
Darhazer, line 34: This could be `.dig('RSpec', 'Language', 'Hooks').to_a`

We don't need to calculate this every time however, only once, when reading the config. So it still could be a constant

sl4vr: Yeah, I'll add some config access method once there will be more language elements implemented.

lib/rubocop/cop/rspec/named_subject.rb
```diff
@@ -45,12 +45,14 @@ class NamedSubject < Cop
...
-          {
-            #{Examples::ALL.block_pattern}
-            #{Hooks::ALL.block_pattern}
-          }
-        PATTERN
+        def_runtime_node_matcher :rspec_block? do
+          <<-PATTERN
+            {
+              #{Examples::ALL.block_pattern}
+              #{hook_selectors.block_pattern}
```
pirj, line 52: @marcandre summoning you to gather your opinion on a lazy node matcher definition.
We're experiencing a problem with hard-coded DSL definition (say, hooks are limited to `begin`/`after`/`around`), while RSpec allows for various aliases (examples, example groups) that we're unable to properly detect.

However, due to lazy config loading that happens after the cops are loaded and helper `Language` module is loaded, we resort to introduce a `def_runtime_node_matcher` that is passed a block to postpone node pattern evaluation.

Do you see any problem with that?

@andrykonchin summoning you as our de-facto performance expert. Do you think this change comes with a significant performance impact?

lib/rubocop/rspec/language.rb
```diff
@@ -53,6 +53,25 @@ def node_pattern
...
+          'Hooks' => %i[
+            prepend_before
+            before
+            append_before
+            around
+            prepend_after
+            after
+            append_after
+          ]
+        }.freeze
```
pirj, line 68: Do you think it's possible to move this to `config/default.yml`?

lib/rubocop/rspec/example_group.rb
```diff
@@ -4,6 +4,17 @@ module RuboCop
   module RSpec
     # Wrapper for RSpec example groups
     class ExampleGroup < Concept
+      # Cop object may be passed into overriden constructor as additional param,
+      # so we can call predicate on cop itself and will have access to #config
+      # method in order to use configured language aliases for predicates.
+      # If no predicate_receiver passed - predicate will be called on
+      # ExampleGroup as if was before.
+      def initialize(node, predicate_receiver = self)
```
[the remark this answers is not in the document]
sl4vr: Tbh I made it just to pass rubocop checks. There are some places where `ExampleGroup` instantiated and I wanted to keep them unchanged at least unless all RSpec language elements became dynamic.

lib/rubocop/cop/rspec/named_subject.rb
marcandre: `NodePattern` is more flexible than ever, especially with the named parameters that I introduced recently (which can be specified at definition time or when calling), so I doubt that it's needed. Performance-wise my guess is it would be somewhere between really slow and attrociously slow 🤣.
I can look at this PR in depth tomorrow if needed, but to be clear: don't use `def_runtime_node_matcher`...

marcandre: E.g. if the keywords `before/after/...` can only be known later...
```ruby
def_node_matcher :something, '(something (sym %keywords) ...', keywords: ->(keyword) { acceptable_keyword_set.include?(keyword) }
```

Just define `acceptable_keyword_set` to return a memoized `Set` of keywords...

lib/rubocop/rspec/language/runtime_macros.rb
```diff
@@ -0,0 +1,32 @@
...
+module RuboCop
+  module RSpec
+    module Language
+      # RuntimeMacros `.def_node_matcher` and `.def_node_search` accept block
+      # that will be evaluated within cop context to get pattern pattern.
+      # It can be used to pattern match values from config because are available
+      # only during runtime.
+      module RuntimeMacros
+        def def_runtime_node_matcher(method_name, &pattern_block)
+          define_method method_name do |node, *args, &block|
```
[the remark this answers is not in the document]
sl4vr: Yeah, even no one in their right mind is going to configure RSpec aliases per directory - I think we should probably keep such behaviour because it is expected behaviour.
I agree, memoizing becomes untrivial, maybe we could use `RSpec/Language` config hash as cache marker. Let's probably wait for some performance impact report?
Also I consider elimination of `SelectorSet` in favour of just arrays of symbols because they are instantiated a lot in runtime matchers implementation. But I'm not sure if it's a performance hit though.

lib/rubocop/cop/rspec/described_class.rb
```diff
@@ -131,11 +132,7 @@ def scope_change?(node)
         end
 
         def skippable_block?(node)
-          node.block_type? && !rspec_block?(node) && skip_blocks?
-        end
-
-        def skip_blocks?
-          cop_config['SkipBlocks']
+          node.block_type? && !rspec_block?(node) && cop_config['SkipBlocks']
```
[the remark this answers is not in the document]
sl4vr: The class is too large, so I've added this to pass rubocop. I'll add cop ignore comment better then.

lib/rubocop/rspec/language.rb
sl4vr: I struggle with testing it, it works if `config` context is included in spec. I obviously don't want to do that manually for every spec file, so I tried to include it in `spec_helper`, but it didn't work for some reason.
I'll try to look into that again.

lib/rubocop/rspec/language/runtime_macros.rb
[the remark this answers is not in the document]
pirj: > no one in their right mind is going to configure RSpec aliases per directory 

Certain specs, e.g. Pundit/CanCanCan/ActionPolicy specs will reside in their own directory.
Even though it's RSpec aliases are defined globally, this dynamic language resolution is not only about core RSpec DSL aliases.
I believe this feature will also allow to fine-tune detection of custom matchers, negated matchers that are used mostly local to certain directories.

lib/rubocop/rspec/language.rb
pirj: Please feel free to include it for cop specs only:
```ruby
config.define_derived_metadata(type: :cop) do |metadata|
  metadata[:config] = true
end
```

If `cop: true` metadata is not set for cops, it's possible to set it as well:
```ruby
config.define_derived_metadata(file_path: %r{/spec/rubocop/cop/rspec/}) do |metadata|
  metadata[:type] = :cop
end
```

lib/rubocop/cop/rspec/named_subject.rb
pirj: > between really slow and attrociously slow

😆 

Thanks a lot @marcandre! This is something I've totally overlooked, even with [my recent usage of this feature](https://github.com/rubocop-hq/rubocop-rspec/pull/930/files#diff-2aecb8a103588de3553f5448956d701fR59).

@sl4vr Would you like to give another approach a shot?

sl4vr: I'm ashamed to admit I was referring [outdated docs](https://www.rubydoc.info/github/bbatsov/RuboCop/RuboCop/NodePattern/Macros) while working on this implementation 🤷‍♂️. Actually I've noticed that only after publishing this PR when I tried to investigate on failing tests.

@pirj Of course! I'll try to rework this PR in order to adopt `NodePattern`'s named parameters.

lib/rubocop/cop/rspec/cop.rb
```diff
@@ -68,6 +69,10 @@ def rspec_pattern_config
             .fetch('RSpec', DEFAULT_CONFIGURATION)
             .fetch('Patterns')
         end
+
+        def rspec_language_config(*keys)
+          all_cops_config.dig('RSpec', 'Language', *keys)&.map(&:to_sym) || []
```
Darhazer, line 74: one way to avoid the nil safety and the || is to use `to_a` as nil converts to an empty array
```
all_cops_config.dig('RSpec', 'Language', *keys).to_a.map(&:to_sym)
```

Darhazer: Someday I should write a cop to avoid unnecessary nil safety, as I hate seeing &. chains where a conversion will do the job

pirj:
@sl4vr Do you need some help with this?
Highly appreciate if you could at least come up with a PoC, I can take it from there if you like, or we can share the work if you like.

lib/rubocop/cop/rspec/named_subject.rb
marcandre: Did you run into difficulties? I notice that `def_runtime_node_matcher` is still there

lib/rubocop/rspec/language.rb
pirj: @bquorning might have some ideas with stubbing `AllCops`.
I've recently discovered that in addition to `config` there's also a `all_cops_config` https://github.com/rubocop-hq/rubocop/blob/master/lib/rubocop/rspec/shared_contexts.rb#L76
Might come handy?

lib/rubocop/rspec/language.rb
```diff
@@ -5,6 +5,12 @@ module RSpec
     # RSpec public API methods that are commonly used in cops
     module Language
       RSPEC = '{(const {nil? cbase} :RSpec) nil?}'
+      BLOCK_PATTERN = <<-PATTERN
```
pirj, line 8: Are those two used only once, to define a `def_node_matcher`? Maybe inline them?

lib/rubocop/cop/rspec/dialect.rb
```diff
@@ -47,14 +47,14 @@ class Dialect < Cop
 
         MSG = 'Prefer `%<prefer>s` over `%<current>s`.'
 
-        def_node_matcher :rspec_method?, ALL.send_pattern
-
         def on_send(node)
-          return unless rspec_method?(node)
```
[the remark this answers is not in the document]
sl4vr: Hmm, yeah it's definitely possible and should be done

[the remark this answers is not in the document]
sl4vr: @pirj did you mean to define them under particular cops or in `NodePattern` module?

lib/rubocop/rspec/language.rb
sl4vr: I would leave it until converting other RSpec keywords

lib/rubocop/cop/rspec/dialect.rb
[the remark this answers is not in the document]
pirj: Keeping it in `NodePattern` is just fine.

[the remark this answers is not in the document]
sl4vr: Well, it looks like `rspec_method?` is used only in one place. I've implemented couple of simple methods for `Language` module for embedding keywords into block and send patterns (same as we have in `Selectors`), so used them for now.
And putted back `hook?` matcher because it's used in several places.

lib/rubocop/rspec/language.rb
sl4vr: Yeah, I actually tried to override `all_cops_config`, but the thing I was struggling with turned out to be that metadata assignment. So seems to work now.

sl4vr:
@pirj I've lost your comment regarding `shared_context` explicit vs implicit inclusion 🤷‍♂️
Anyway, I agree, changed that to including that context to all specs with `config` metadata. `all_cops_config` still can be overriden in spec, so we are good.

sl4vr:
`NodePattern` is bloated a bit with similar methods, I would wait for all RSpec language parts implementation and then do something like:
```
{
  regular_examples: ['Examples', 'Regular'],
  focused_examples: ['Examples', 'Focused'],
  skipped_examples: ['Examples', 'Skipped'],
  pending_examples: ['Examples', 'Pending'],
  hooks: ['Hooks'],
}.each do |name, config_keys|
  define_method "rspec_#{name}" do |keyword|
    send("#{name}_set").include?(keyword)
  end

  define_method "#{name}_set" do
    instance_variable_get("@#{name}_set") ||
      instance_variable_set(
        "@#{name}_set",
        Set.new(rspec_language_for(*config_keys))
      )
  end
end
```
Because it highly depends on config and may change yet.

marcandre:
I think much of what is in `Language` is now not/less needed, now that `NodePattern` now accepts named parameters, at runtime or at definition time; I believe this could be used to simplify greatly, one could pass `Set`s directly this way for example.

lib/rubocop/cop/rspec/align_left_let_brace.rb
```diff
@@ -40,6 +37,16 @@ def on_new_investigation
...
           end
         end
+
+        private
+
+        def token_aligner
+          RuboCop::RSpec::AlignLetBrace.new(
+            processed_source.ast,
+            :begin,
+            rspec_language_config
```
pirj, line 47: Maybe push `rspec_language_config` to `AlignLetBrace` not to pass it?

lib/rubocop/cop/rspec/base.rb
```diff
@@ -68,6 +68,12 @@ def rspec_pattern_config
             .fetch('RSpec', DEFAULT_CONFIGURATION)
             .fetch('Patterns')
         end
+
+        def rspec_language_config
+          all_cops_config
+            .fetch('RSpec', DEFAULT_CONFIGURATION)
+            .fetch('Language')
+        end
```
pirj, line 76: Maybe push this down to `Language`? This way `ExampleGroup`/`Concept` will also have access to it, and we won't have to pass `rspec_language_config` as an argument?

spec/rubocop/cop/rspec/describe_method_spec.rb
```diff
@@ -27,7 +25,7 @@
 
   it 'skips specs not having a string second argument' do
     expect_no_offenses(<<-RUBY)
-      describe Some::Class, :config do
```
pirj, line 30: Please keep this, it's a part of the example. Another symbol will do, e.g. `:foo`.

spec/rubocop/cop/rspec/dialect_spec.rb
```diff
@@ -9,7 +9,7 @@
     }
   end
 
-  it 'allows describe blocks' do
+  it 'allows describe blocks', x: true do
```
pirj, line 12: Debug leftover?

pirj:
How far are you from the point when you'll be happy enough about your changes to be merged, @sl4vr ?
Asking because there are a few other pull requests hanging, and I think it will be easier to adapt them to your changes than adapt this pull request to those other changes.

spec/rubocop/cop/rspec/describe_method_spec.rb
sl4vr: Oops, removed with semi-auto cleanup 😅

lib/rubocop/cop/rspec/base.rb
sl4vr: Would be cool, but how can I access `config` from outside of cop?

sl4vr:
@pirj, I was going to make Capybara and FactoryBot languages configurable as well, but now I see that it can be done in separate PRs.
So I'll just do something to repeating methods in `NodePattern` and I'm good. I'm already on it, so should be there pretty shortly.

lib/rubocop/cop/rspec/base.rb
pirj: Fair enough, there's no way.

lib/rubocop/cop/rspec/align_left_let_brace.rb
pirj: This seems impossible as well.
We may rework `AlignLetBrace` to be an abstract base class for those two, but it's completely outside of the scope.

sl4vr:
Well, PR seems ready for review now. Please let me know if I need to do anything besides any review suggestions. I mean squash commits, add any tests, update docs etc.

lib/rubocop/cop/rspec/empty_hook.rb
```diff
@@ -29,7 +29,7 @@ class EmptyHook < Base
         MSG = 'Empty hook detected.'
 
         def_node_matcher :empty_hook?, <<~PATTERN
-          (block $#{Hooks::ALL.send_pattern} _ nil?)
+          (block $#{send_pattern('#rspec_hooks')} _ nil?)
```
pirj, line 32: Trailing whitespace?

lib/rubocop/cop/rspec/focus.rb
```diff
@@ -22,19 +22,23 @@ module RSpec
...
 
-        focused = ExampleGroups::FOCUSED + Examples::FOCUSED
-
-        def_node_matcher :focusable_selector?,
-                         (ExampleGroups::GROUPS + ExampleGroups::SKIPPED +
-                          Examples::EXAMPLES + Examples::SKIPPED +
-                          Examples::PENDING).node_pattern_union
+        def_node_matcher :focusable_selector?, <<-PATTERN
+          {
+            {#rspec_regular_example_groups #rspec_skipped_example_groups}
```
pirj, line 27: Redundant `{ ... }`?

lib/rubocop/rspec/language/config.rb
```diff
@@ -0,0 +1,98 @@
...
+
+              define_accessor_and_matcher_aggregator(key, group_keys)
+            else
+              define_accessor_and_matcher_from_config(key)
+            end
+          end
+        end
+
+        def extended(base)
+          base.instance_eval { define_accessors_and_matchers }
```
pirj, line 93: Is it possible to define those once in the module, and include that module with already defined methods?

lib/rubocop/rspec/language/node_pattern.rb
```diff
@@ -6,29 +6,72 @@ module Language
       # Common node matchers used for matching against the rspec DSL
       module NodePattern
         extend RuboCop::NodePattern::Macros
+        extend RuboCop::RSpec::Language
+        extend RuboCop::RSpec::Language::Config
+
+        # define_all_language_methods
```
pirj, line 12: Leftover?

lib/rubocop/rspec/language/node_pattern.rb
```diff
@@ -6,29 +6,72 @@ module Language
...
+        def rspec_all(keyword)
+          all_keywords.include?(keyword)
+        end
 
-        def_node_matcher :let?, Helpers::ALL.block_or_block_pass_pattern
+        protected
 
-        def_node_matcher :include?,
-                         Includes::ALL.send_or_block_or_block_pass_pattern
+        def all_keywords
```
pirj, line 59: This one may be made private.

sl4vr:
Does it make sense to change matcher method names from `"rspec_#{basename}"` to `"#{basename}?"`?

bquorning:
Wow, this PR looks like a tremendous effort 👏 

However, I am having a real hard time reviewing the changes. I see e.g. a revert in the commit history, and a few commits that could probably be extracted to a separate PR (e.g. the `:config` changes to the specs).

Would it be possible to clean it up a bit, to make the changes easier to understand? If not, I guess I’ll just have to try harder 😄

pirj:
The revamp of `lib/rubocop/rspec/language.rb` is a breaking change, so this pull request surely needs to be part of 2.0 release, and can't be merged earlier.

@bquorning @Darhazer WDYT of releasing what we have now since it's been a while anyway? Chances are this will be the penultimate release for major version 1.

pirj:
> Does it make sense to change matcher method names from `"rspec_#{basename}"` to `"#{basename}?"`?

This question was troubling me as well.
Methods like `#rspec_hooks` are used as a union of keywords that can be used to match a single element in the AST, while `#hook?` as a block pattern that matches those, a bigger chunk of the AST. So, those two serve different purposes and are not interchangeable.
Also, it would only add confusion to mix those two. I really like that those are separate. `#rspec_*` are DSL elements definition, and the other - ready-to-use matchers for the usage of the DSL.

pirj:
One way to improve the former is to [use function arguments](https://github.com/rubocop-hq/rubocop-ast/blob/master/docs/modules/ROOT/pages/node_pattern.adoc#arguments-for-predicate-and-function-calls) arguments and thus reduce the number of methods, e.g.:
```diff
- #rspec_all_shared_groups
+ #rspec(:shared_groups, :all)
```
In this case the `:all` can be the default argument, thus reducing this to `#rspec(:shared_groups)`.


And probably get rid of predefined methods in `Language::Config`.
There's just one reason why I dislike the dynamically defined methods for `#rspec_*` - there's no clear way to read their source, and for us and for the other cops authors there is no option to add per-method documentation.
[An example of how those can be defined](https://github.com/rspec/rspec-core/blob/5d87c28bb828ba14b54a0d8cfddf044894009314/lib/rspec/core/example_group.rb#L108).

A couple of snippets of what I propose:
```diff
- (send _ #rspec_runners $(send nil? :be))
+ (send _ #rspec(:runners) $(send nil? :be))
```
```diff
- {#rspec_regular_example_groups #rspec_skipped_example_groups #rspec_regular_examples #rspec_skipped_examples #rspec_pending_examples}
+ {#rspec(:example_groups, :regular) #rspec(:example_groups, :skipped) #rspec(:examples, :regular) #rspec(:examples, :skipped) #rspec(:examples, :pending)}
```

It does not necessarily be a part of this pull request, we can make those adjustments somewhere in 2.0.0-rc2.

pirj:
> this pull request surely needs to be part of 2.0 release

As a consequence of this, we'll have to merge https://github.com/rubocop-hq/rubocop-rspec/pull/226 (uses `Language::Expectations::ALL`) and https://github.com/rubocop-hq/rubocop-rspec/pull/994 (uses `Language::Includes::EXAMPLES`) either before this pull request, or soon after, but before the final 2.0.0, not to wait for 3.0 to turn them on by default.

sl4vr:
@bquorning I've rewritten commit history, so now it looks more straightforward.

There are only couple of changes that can be extracted from PR:
* Documentation comment fix in last commit (it really small, let it be 😅)
* `:config` tags removed from cop specs. But it can be left as well because most of the changes are removal of `subject(:cop) {...}` which is mandatory. Config tags are just not needed there since config context is forced for all cops, so no reason to keep them until any next PR.

Hopefully restructured commit history will make it easier to review

sl4vr:
>There's just one reason why I dislike the dynamically defined methods for #rspec_* - there's no clear way to read their source, and for us and for the other cops authors there is no option to add per-method documentation.
An example of how those can be defined.

@pirj I agree that naming of matchers is pretty obscure. I tried to make them more similar to natural language, but turns out there is a price.

I like the idea with function arguments, but I would probably even use camel-cased names as they're defined in config, like `#rspec(:SharedGroups, :Regular)`, `#rspec(:SharedGroups)`, etc.
Not common ruby syntax for symbols though, but we could use constants `#rspec(SharedGroups, Regular)`.
Just thinking out loud.

> Methods like #rspec_hooks are used as a union of keywords that can be used to match a single element in the AST, while #hook? as a block pattern that matches those, a bigger chunk of the AST. So, those two serve different purposes and are not interchangeable.

Yeah, I was thinking about those matcher naming as well, tbh `hook?` could be `hooks_block?`, etc.
Not necessary though.

pirj:
Sounds reasonable, @sl4vr 

Commit grouping looks good 👍 

Just one thing - please add a CHANGELOG entry. This effort **really** deserves recognition.
It will have an incredible impact on extensibility, and a number of outstanding tickets could be resolved. Thanks a lot for what you did.

I can take care of fixing `StubbedMock` (#226) and `RepeatedIncludeExamples` (#994, merged) to adhere to the new API you've introduced.

marcandre:
A complexity that would be great to do without is that this PR builds methods in pairs: one to get the `Set` of keywords (e.g. `hooks_keywords`) and one to the `that_set.include?(arg)` (e.g. `rspec_hooks`). It would be nice to only have the `Set` (probably best called `rspec_hooks` or `rspec_hooks_set`).

In the last releases, `NodePattern` was extended to use sets directly as matchers when they are a constant, an argument, or a named parameter.

For example:
```ruby
# this PR:
def_node_matcher :rspec_block?, block_pattern('#rspec_all')

# can also be written as:
def_node_matcher :rspec_block?, block_pattern('%rspec_all'), rspec_all: rspec_all_set
```

But that is a bit tedious. It would be easy to modify `NodePattern` so that it finds the value of a unnamed parameter that isn't specified. For example

```ruby
def_node_matcher :rspec_block?, block_pattern('%rspec_all_set')
```

This could call `rspec_all_set` directly. Note the difference of using `#` and `%`: `#foo` calls the method `foo(cur_element)`, while `%foo` calls `foo === cur_element`, and `foo` can be specified at definition time, or at call time. For `Set`s, `===` is the same as `include?`.

It's a trivial change to `NodePattern`, which instead of generating `def rspec_block?(node=self, rspec_all_set:) ...` would generate `def rspec_block?(node=self, rspec_all_set: rspec_all_set) ...`

WDYT?

lib/rubocop/rspec/language/config.rb
```diff
@@ -0,0 +1,95 @@
...
+        def define_accessor_and_matcher(basename, block)
+          define_keyword_accessor "#{basename}_keywords", block
+          define_matcher_method basename
+          private "rspec_#{basename}", "#{basename}_keywords"
+        end
+
+        def define_keyword_accessor(accessor_name, block)
+          define_method accessor_name do
+            instance_variable_get("@#{accessor_name}") ||
+              instance_variable_set("@#{accessor_name}", instance_eval(&block))
```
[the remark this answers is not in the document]
sl4vr: My original idea was to keep `rspec_language_for` protected, but probably it's not needed, then `yield self` and actual block would work just fine.

lib/rubocop/rspec/language/node_pattern.rb
```diff
@@ -6,29 +6,72 @@ module Language
...
+        PATTERN
+
+        def_node_matcher :example?, block_pattern('#rspec_all_examples')
+
+        def_node_matcher :hook?, block_pattern('#rspec_hooks')
+
+        def_node_matcher :let?, <<-PATTERN
+          {
+            #{block_pattern('#rspec_helpers')}
+            #{block_pass_pattern('#rspec_helpers')}
```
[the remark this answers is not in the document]
sl4vr: Makes sense, I didn't even looked into it. So seems it was redundant all the time.

lib/rubocop/rspec/language/node_pattern.rb
```diff
@@ -6,29 +6,72 @@ module Language
...
 
-        def_node_matcher :include?,
-                         Includes::ALL.send_or_block_or_block_pass_pattern
+        def rspec_all(keyword)
+          all_keywords.include?(keyword)
+        end
 
-        def_node_matcher :subject?, Subject::ALL.block_pattern
+        def all_keywords
+          @all_keywords ||= [
```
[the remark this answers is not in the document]
sl4vr: Makes sense to move it to config 👍 

> When done, should be a Set
Could you elaborate?

lib/rubocop/rspec/language/config.rb
```diff
@@ -0,0 +1,95 @@
...
+
+        def define_accessor_and_matcher(basename, block)
+          define_keyword_accessor "#{basename}_keywords", block
+          define_matcher_method basename
+          private "rspec_#{basename}", "#{basename}_keywords"
+        end
+
+        def define_keyword_accessor(accessor_name, block)
+          define_method accessor_name do
+            instance_variable_get("@#{accessor_name}") ||
```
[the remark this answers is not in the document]
sl4vr: Do you suggest to rename it to `memoize` and use eval with string instead, right?

sl4vr:
@marcandre I definitely would love to get rid of pairs of methods and use just memoized sets.

Afaics the code below won't work because it is called on class level when we don't have config yet, `rspec_all_set` is instance method.
```ruby
def_node_matcher :rspec_block?, block_pattern('%rspec_all'), rspec_all: rspec_all_set
```
 I'm not sure, but probably it's possible to modify `NodePattern` to eval not passed keywords at matching time?

But it also seems to me pretty neat to have only one keywords method instead like `rspec(*config_keys)` and collection of dynamically defined memoized sets under the hood. So it remove all of `"rspec_#{method_basename}"`-like methods and make things more clear for everyone in one shot, since it would allow to use config keys like we discussed with @pirj, kinda `#rspec(:ExampleGroups, :Skipped)`. If there would be no performance issues of course.

marcandre:
> Afaics the code below won't work because it is called on class level when we don't have config yet, `rspec_all_set` is instance method.

You're right, sorry; a lambda would work but is even more trouble

> I'm not sure, but probably it's possible to modify NodePattern to eval not passed keywords at matching time?

Yes, that's what I was suggesting. Either by calling the method with the same name as keyword parameter directly, or via a catchall...

> #rspec(:ExampleGroups, :Skipped)

Oh, right. That could be already achieved I think:

```ruby
def rspec(elem, group, subgroup)
  config.structure_keywords(group, subgroup).include? elem
end
```

lib/rubocop/rspec/language/node_pattern.rb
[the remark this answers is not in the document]
pirj: Hold on, but:
```ruby
def block_pass_pattern(keywords_matcher)
  "(send #rspec? #{keywords_matcher} _ block_pass)"
end

def block_pattern(keywords_matcher)
 "(block #rspec? #{keywords_matcher} ...) ... )"
end
```
one doesn't have a block.

pirj:
Speaking of
```diff
- def_node_matcher :rspec_block?, block_pattern('#rspec_all')
+ def_node_matcher :rspec_block?, block_pattern('%rspec_all')
```

> `#foo` calls the method `foo(cur_element)`, while `%foo` calls `foo === cur_element` [...] For `Set`s, `===` is the same as `include?`.

Sounds like a very minimal optimization, given that `#foo` delegates to `include?`. Not much difference code-wise.

> It's a trivial change to `NodePattern`, which instead of generating `def rspec_block?(node=self, rspec_all_set:) ...` would generate `def rspec_block?(node=self, rspec_all_set: rspec_all_set) ...`

You've started this message with "A complexity that would be great to do without". Isn't the complexity usually built out of trivialities? ;)


Wondering if `#rspec_all` returned a set, without checking `include?` itself, will it `===`?


Let me propose a challenge. We merge this pull request as it is now. Call this a 2.0.0-pre, and then send our PRs with simplifications. @marcandre, you have an upper hand as you may also adjust `rubocop-ast` to your liking to make this code easier. Up for such a challenge? Have an idea on the prize in mind?

lib/rubocop/rspec/language/config.rb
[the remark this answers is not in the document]
pirj: This code is hard to understand. It's not documentable. I personally preferred the ugly and verbose, but simpler variant of it.

The idea behind `#rspec(:SharedGroups, :Regular)` is to get rid of this altogether, and not to choose between those two poles.

lib/rubocop/rspec/language/node_pattern.rb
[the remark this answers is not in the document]
sl4vr: Oops, means `block_pass_pattern` is a subset of `send_pattern` actually.
So it's redundant [when used along with `send_pattern` here](https://github.com/sl4vr/rubocop-rspec/blob/b190a22d0877be4cb6a9b2e5bc1145a399ff4b21/lib/rubocop/rspec/language/node_pattern.rb#L45), but not [here](https://github.com/sl4vr/rubocop-rspec/blob/b190a22d0877be4cb6a9b2e5bc1145a399ff4b21/lib/rubocop/rspec/language/node_pattern.rb#L37).

lib/rubocop/rspec/language/node_pattern.rb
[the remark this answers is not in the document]
pirj: I guess @marcandre means that:
```ruby
@all_keywords ||= Set.new([
  all_example_groups_keywords,
  ...
```

This way it's possible to get rid of `.include?(keyword)` and the parameter here:
```ruby
        def rspec_all
          all_keywords
        end
```
There's one spec failing for `controller do`. Didn't dig further.

My point is to merge as is and then exercise our refactoring skills separately. We're in a good position for that with no 1.x releases left between the merge and 2.0.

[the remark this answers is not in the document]
sl4vr: Hmm I was confused because `all_keywords` returns sum of `Set`s which is a `Set` too.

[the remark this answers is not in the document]
pirj: Nice! 👍

sl4vr:
- [x] Refactored minor review suggestions
- [x] Rebased
- [x] Added CHANGELOG entry

lib/rubocop/rspec/language/node_pattern.rb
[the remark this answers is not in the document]
marcandre: My mistake, sorry 😊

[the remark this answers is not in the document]
sl4vr: The only way to memoize keyword `Set`s on config level that I currently see is to monkey-patch `RuboCop::Config`. I think no any facade or similar would work, config should have memoization on `RuboCop` level so that cops would be instantiated with the same instance of `Config`.
In this case it also makes sense to move there content of `RuboCop::Cop::RSpec::Base`.

pirj:
Tested the extension of language config from a user repo, works perfectly:
image

sl4vr:
I've made a bit of refactoring [here](https://github.com/sl4vr/rubocop-rspec/commit/bae41dc095ecf54b2f3f8e3bf8f537a466d32c2f) including changing matchers to use keys from config as it was discussed, like `#rspec(:Hooks)`.

Should I include that commit into PR?

pirj:
> I've made a bit of refactoring

I like it better!

Just in case - is there any noticeable performance difference between `master` vs `#rspec_...` vs `#rspec(...)` when you run it against some massive repository?

marcandre:
>  @marcandre, you have an upper hand as you may also adjust `rubocop-ast` to your liking to make this code easier. Up for such a challenge? Have an idea on the prize in mind?

Still waiting for my previous prize 😅.
I'm finally all green on my `NodePattern` parser, but it's in need of refactoring. I'll keep `%method(...)` in mind, should be easy to add.

sl4vr:
I've run `time bundle exec rubocop --cache false spec/` several times against quite massive repository. Here are approximate results:

1. `master`
```
real    2m58.160s
user    2m35.507s
sys     0m2.049s
```

2. `#rspec_...`
```
real    3m11.287s
user    2m48.998s
sys     0m2.028s
```

3. `#rspec(...)`
```
real    3m11.763s
user    2m50.498s
sys     0m2.110s
```

4. `#rspec(...)` with `def_node_matcher :rspec_all` (yet another refactoring attempt)
```
real    3m18.894s
user    2m54.295s
sys     0m2.314s
```

Difference between 3 and 4:
```ruby
# 3 variant
def rspec_all(keyword)
  rspec_all_keywords.include?(keyword)
end

def rspec_all_keywords
  @rspec_all_keywords ||=
    rspec_keywords(:ExampleGroups)  +
      rspec_keywords(:SharedGroups) +
      rspec_keywords(:Examples)     +
      rspec_keywords(:Hooks)        +
      rspec_keywords(:Helpers)      +
      rspec_keywords(:Subjects)     +
      rspec_keywords(:Expectations) +
      rspec_keywords(:Runners)      +
end

# 4 variant
def_node_matcher :rspec_all, <<-PATTERN
  {#rspec(:ExampleGroups)
  #rspec(:SharedGroups)
  #rspec(:Examples)
  #rspec(:Hooks)
  #rspec(:Helpers)
  #rspec(:Subjects)
  #rspec(:Expectations)
  #rspec(:Runners)}
PATTERN
```

pirj:
> Here are approximate results:

Nice! I think that ~8% is a reasonable tradeoff.
Since there's not much difference I'd go with the option 3 (`#rspec(...)`).

pirj:
Marc-André's idea to read the config once, not per cop might be exactly what will address the slowdown.

sl4vr:
> Since there's not much difference I'd go with the option 3 (#rspec(...)).

Great, I'll rebase and force push `rspec(...)` variant this evening then.

> Marc-André's idea to read-parse the config once, not per cop might be exactly what will address the slowdown.

Yup, that might be it. I already commented that it requires monkey-patching `::Rubocop::Config` [here](https://github.com/rubocop-hq/rubocop-rspec/pull/956#discussion_r472969954). I would love to use refinements, but it doesn't seem to work with inheritance by design, so I would have to include `using RefinedConfig` in every cop class. Nevertheless if it's fine to monkey-patch `::Rubocop::Config` - I can look into that as well.

pirj:
A wild idea - maybe use a singleton for conf methods and delegate to it from `Language::Config` module?

sl4vr:
> A wild idea - maybe use a singleton for conf methods and delegate to it from Language::Config module?

I think it won't work unfortunately because config might be changed if there are nested configuration files.

pirj:
Agree. Let's not overcomplicate then for a negligible performance impact, especially if it comes with a risk of incorrect behaviour.

marcandre:
A solution would be to store the sets in a global hash indexed by the config, but I think it's best to add methods to the main config.

pirj:
@sl4vr 
```diff
- $(send nil? $#{Expectations::ALL.node_pattern_union} ...)
+ $(send nil? $#rspec(:Expectations) ...)
```

Thanks a lot for keeping this PR up to date!
Totally sorry about that, but it's going to be like that for a while until RuboCop 1.0 is released, hopefully [really soon](https://github.com/rubocop-hq/rubocop/milestones/1.0.0).


Meanwhile, @bquorning, @Darhazer, are you happy with this change?

lib/rubocop/cop/rspec/focus.rb
```diff
@@ -22,19 +22,22 @@ module RSpec
...
         MSG = 'Focused spec found.'
 
-        focused = ExampleGroups::FOCUSED + Examples::FOCUSED
-
-        def_node_matcher :focusable_selector?,
-                         (ExampleGroups::GROUPS + ExampleGroups::SKIPPED +
-                          Examples::EXAMPLES + Examples::SKIPPED +
-                          Examples::PENDING).node_pattern_union
+        def_node_matcher :focusable_selector?, <<-PATTERN
+          {#rspec(:ExampleGroups, :Regular) #rspec(:ExampleGroups, :Skipped)
```
pirj, line 26: @marcandre Do you think it makes sense (for other core cops and extension cops) to provide this kind of syntax:
```ruby
def_node_matcher :focusable_selector?, <<-PATTERN
  #rspec.example_groups.regular
PATTERN
```
e.g. so that the Node Pattern's function calls/predicates can be namespaced/chained?

Disclaimer: there's only an aestetic effect, no actual breakthroughs of having such a syntax. By "makes sense" I mean "is it worth the effort of implementing/maintaining this additional complexity".

lib/rubocop/rspec/rspec_config.rb
```diff
@@ -89,10 +89,12 @@ def rspec_keywords_all_all
       end
 
       def rspec_pattern
+        return @rspec_pattern if @rspec_pattern
```
bquorning, line 92: Is this memoization a necessary part of this PR? It’s been discussed previously in #967, and as I recall it, memoizing on the Cop instance doesn’t make much of a speed difference.

sl4vr: I just glimpsed that PR and thought that since I'm moving configuration-related methods from Cop to Config anyway I could add memoization there. `RSpecConfig` is included into `RuboCop::Config` in this PR, so it will be memoization on config level.

Anyway if you think that it's not quite needed to this particular PR, I can rebase without it, thankfully it's a standalone commit.

lib/rubocop/cop/rspec/focus.rb
marcandre: It could be interesting to have namespaces somehow; I'd be enclined to use constants though... `RSpec::ExampleGroups.regular`, assuming `regular` is a singleton_method; and maybe `def_node_matcher` should define a singleton_class method at the same time.

pirj:
@sl4vr Do you have a moment to check if this applies will on `release-2.0` branch? (I've changed the target branch for this PR).

Codeclimate can be fixed later in a follow-up.

pirj:
Looks green, @bquorning

bquorning:
I’ll leave this review for tomorrow or the weekend.

bquorning (commented):
I have extracted a bit of this PR into #1059.

bquorning:
I don’t see any specs describing how how configuration changes which DSL methods are matched. Also, I wonder how easy it will be to configure – does one need to copy the entire existing configuration from default.yml, or is it possible to easily add one or two missing method names to one of the lists?

pirj:
> Also, I wonder how easy it will be to configure – does one need to copy the entire existing configuration from default.yml, or is it possible to easily add one or two missing method names to one of the lists?

@bquorning Luckily enough, just the missing ones need to be added, and it's merged with our base config.

bquorning:
> just the missing ones need to be added, and it's merged with our base config.

Do you think it would be a useful test case?

pirj:
> I don’t see any specs describing how configuration changes which DSL methods are matched.
> Do you think it would be a useful test case?

I've missed this obvious thing 🤦 

Give me a couple hours, I'll rebase and through in a coiple tests showing how the default config can be extended. I remember there was a leak when overriding the config, but I've figured out a way (`dup`?) how to avoid that.

spec/rubocop/rspec/r_spec_config_spec.rb
```diff
@@ -0,0 +1,35 @@
...
+      epic 'great achievements or events is narrated in elevated style' do
+        ballad 'slay Minotaur' do
+          # ...
+        end
+      end
+    RUBY
+  end
+  let(:ast) { RuboCop::ProcessedSource.new(source, RUBY_VERSION.to_f).ast }
+
+  context 'with the default config' do
```
pirj, line 18: @bquorning The test for this whole change.

A real cop spec: https://github.com/rubocop-hq/rubocop-rspec/pull/1007/commits/0cf00f4c26e38e5658039c6ac41d5d01f6b2d922#diff-fb6cde18a3a09cdce77660231314ae226a9fd3605928279010b8dbd3558e518eR271

bquorning (commented):
I think the DSL is fine, but I really don’t like that we monkey patch RuboCop’s Config class. Can we come up with a better way to achieve the same thing?

Also, all of those dynamic methods… I saw that @marcandre has previously suggested using constants, e.g. `RSpec::ExampleGroups.regular`. Is that a dead end, or a real option?

I realise that we have to get our v2.0 ready soon. But I am not sure how much of these internal we are allowed to change before v3.0, if we were to merge this pull request as-is.

marcandre:
> I think the DSL is fine, but I really don’t like that we monkey patch RuboCop’s Config class. Can we come up with a better way to achieve the same thing?
> 
> Also, all of those dynamic methods… I saw that @marcandre has previously suggested using constants, e.g. `RSpec::ExampleGroups.regular`. Is that a dead end, or a real option?

I opened https://github.com/rubocop-hq/rubocop-ast/issues/144, can someone please confirm that this would be helpful here (or propose something else)?

lib/rubocop-rspec.rb
```diff
@@ -31,6 +32,17 @@
 
 require_relative 'rubocop/cop/rspec_cops'
 
+# Include memoized `rspec_keywords_...` methods into RuboCop::Config
+# to allow usage of `rspec_keywords` method and `rspec` matcher.
+# Also include `rspec_pattern` to make `relevant_rubocop_rspec_file?` method
+# of `RuboCop::Cop::RSpec::Base` work.
+
+module RuboCop
+  class Config # rubocop:disable Style/Documentation
```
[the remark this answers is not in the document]
sl4vr: If I remember it right - the only benefit of `RuboCop::Config` monkey patching is config-level memoization of keywords (https://github.com/rubocop-hq/rubocop-rspec/pull/956#discussion_r471712658). I mean if I would use `RuboCop::RSpec::Config` as decorator for `RuboCop::Config`, I would have to create an instance of it everytime I want to call `rspec_keywords` of `RuboCop::RSpec::Language::NodePatterns`.
But I wonder if there's an option to decorate config in one place.

lib/rubocop/cop/rspec/align_left_let_brace.rb
```diff
@@ -40,6 +37,16 @@ def on_new_investigation
...
           end
         end
+
+        private
+
+        def token_aligner
+          RuboCop::RSpec::AlignLetBrace.new(
+            processed_source.ast,
+            :begin,
+            config
```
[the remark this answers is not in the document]
sl4vr: Afaik we have to pass config because it's really dynamic and may differ depending on files under check.

sl4vr:
> > I think the DSL is fine, but I really don’t like that we monkey patch RuboCop’s Config class. Can we come up with a better way to achieve the same thing?
> > Also, all of those dynamic methods… I saw that @marcandre has previously suggested using constants, e.g. `RSpec::ExampleGroups.regular`. Is that a dead end, or a real option?
> 
> I opened [rubocop-hq/rubocop-ast#144](https://github.com/rubocop-hq/rubocop-ast/issues/144), can someone please confirm that this would be helpful here (or propose something else)?

Should I try to adopt those changes and improve DSL?
cc: @pirj

pirj:
Thanks, but I'm on it already. I've made some changes to our config loading in #1063, please take a look. This may affect the implementation of config loading here as well.And I plan to squash some commits to‎ avoid having intermenient implementation in history.                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                        From: Viacheslav MefodinSent: Wednesday, 28 October 2020 11:09To: rubocop-hq/rubocop-rspecReply To: rubocop-hq/rubocop-rspecCc: Phil Pirozhkov; MentionSubject: Re: [rubocop-hq/rubocop-rspec] RSpec aliases runtime matchers (#956)





I think the DSL is fine, but I really don’t like that we monkey patch RuboCop’s Config class. Can we come up with a better way to achieve the same thing?

Also, all of those dynamic methods… I saw that @marcandre has previously suggested using constants, e.g. RSpec::ExampleGroups.regular. Is that a dead end, or a real option?



I opened rubocop-hq/rubocop-ast#144, can someone please confirm that this would be helpful here (or propose something else)?



Should I try to adopt those changes and improve DSL?

cc: @pirj



—You are receiving this because you were mentioned.Reply to this email directly, view it on GitHub, or unsubscribe.

[

{

"@context": "http://schema.org",

"@type": "EmailMessage",

"potentialAction": {

"@type": "ViewAction",

"target": "https://github.com/rubocop-hq/rubocop-rspec/pull/956#issuecomment-717769650",

"url": "https://github.com/rubocop-hq/rubocop-rspec/pull/956#issuecomment-717769650",

"name": "View Pull Request"

},

"description": "View this Pull Request on GitHub",

"publisher": {

"@type": "Organization",

"name": "GitHub",

"url": "https://github.com"

}

}

]

sl4vr:
Alright, let me know if I still can do something.

pirj:
I'm a bit concerned about implementing patterns as:
```ruby
def_node_matcher :example_group?, block_pattern('#ExampleGroups.all')
```
It works if `all` uses `Config` underneath, but:
 - it introduces shared state, and this will break once we decide to run RSpec in parallel in threads (there's some work going on in RSpec on this), since if we'll be stubbing/modifying this `Config`, it makes it thread-unsafe
 - - I would rather not go the `Thread.current` route
 - - neither I'm excited to inject cop's config into `Language::ExampleGroups` etc on cop initialization
 - not really a problem for now, but I can think of cop-specific alias configuration, and the approach with `Config` makes it impossible, as cop's own `config` will be ignored, and only the default configuration will be respected.

The reason for this is that a regular method call, e.g.:
```ruby
def_node_matcher :example_group?, block_pattern('#example_groups_all')
```
has all the context from the current cop, specifically `config`, but the `ExampleGroups.all` has no awareness of where it's called from and has no access to the current cop's `config` or anything else.

I'm also hesitant to add a second optional parameter, `cop`, [to those predicates](https://github.com/rubocop-hq/rubocop-ast/commit/1aab8b199c4aafe5a69e965264e478c4ae03daee#diff-e8fe46b5b20b94dae6087a40efdc9363f65b83262f2b752d18770588c647f112R2292), as this probably will affect other functionality as there's no specific distinction on how they are called.

I've tried implementing `ExampleGroups` as a method in a module that the `Base` cop includes, but it doesn't work for two reasons:
1. `rubocop-ast` seems to parse the constant, and call the method on it
2. 
```ruby
module A
  def self.a
    1
  end
  def self.B
    2
  end
end

> A::a
=> 1
> A::B
NameError: uninitialized constant A::B
```

I'm still fancy to provide [the new syntax](https://github.com/rubocop-hq/rubocop-ast/issues/144) to access DSL configuration in node patterns, but can't yet find a good approach to it.

From my point of view,
```ruby
def_node_matcher :example_group?, block_pattern('#ExampleGroups::Regular.all')
```
looks more self-explanatory, familiar Ruby namespacing than
```ruby
def_node_matcher :example_group?, block_pattern('#example_groups.regular.all')
```

pirj:
The above boils down to:
> Cop: Hey, `CigaretteUtil.can_sell_to?(years_old)`
> CigaretteUtil: hey, I don't know, only you are aware of the legal age, it's in your `config`, can only say some average default, which might be inaccurate

pirj:
I went this route:
```ruby
        def_node_matcher :example_group?, block_pattern('#ExampleGroups.all')

        # TODO: add a note to docs to call `super` if you define on_new_investigation in a cop
        def on_new_investigation
          Config.rspec_language_config = config['RSpec']['Language']
        end

        module Config
          class << self
            attr_accessor :rspec_language_config
          end
        end

        module ExampleGroups
          def self.all(element)
            Regular.all(element)# ||
              # Skipped.all(element) ||
              # Focused.all(element)
          end

          module Regular
            def self.all(element)
              Config.rspec_language_config
                .dig('ExampleGroups', 'Regular')
                .include?(element.to_s)
            end
          end
```

Please speak up if you strongly feel against this, @bquorning @Darhazer @marcandre @sl4vr

bquorning:
> it introduces shared state, and this will break once we decide to run RSpec in parallel in threads (there's some work going on in RSpec on this), since if we'll be stubbing/modifying this `Config`, it makes it thread-unsafe

You mean RuboCop, right?

---

> ```ruby
> > A::a
> => 1
> > A::B
> NameError: uninitialized constant A::B
> ```

```
> A.a
=> 1
> A.B
=> 2
```

But probably I don’t understand the problem.

---

> ```ruby
> def_node_matcher :example_group?, block_pattern('#example_groups.regular.all')
> ```

I think this looks fine.

---

> ```ruby
>         # TODO: add a note to docs to call `super` if you define on_new_investigation in a cop
> ```

Or a cop 🤷 

---

> I went this route:

The line `Config.rspec_language_config = config['RSpec']['Language']` doesn’t look thread safe to me. If we could change an instance variable (on the cop) instead, that would be preferable.

pirj:
```ruby
    def_node_matcher :example_group?, block_pattern('#example_groups.regular.all')
```
> I think this looks fine.

It's not bad, but it's not implemented (yet?) in `rubocop-ast`.

> The line `Config.rspec_language_config = config['RSpec']['Language']` doesn’t look thread safe to me. If we could change an instance variable (on the cop) instead, that would be preferable.

Sure. But `RuboCop::RSpec::Language::ExampleGroups.all` method won't be able to read it anyway. So the only option I can think of is `Thread.current[]`,
But I would like to know why this needs to be done. When RuboCop is being run in parallel there's no harm of setting the same config over again even if `rubocop` runs with `--parallel`.
I'm more worried that if we run [`rspec` in a threaded mode (that doesn't exist yet)](https://github.com/rspec/rspec-rails/issues/2104#issuecomment-658474900), that will be breaking our own specs. Not a big deal if this happens, and I don't see a good reason why would we want to do this anyway, our specs run for ~10s now.
I regret that I've brought this up 😆

pirj:
@bquorning @sl4vr Please take a look at the changes.
I'm going to squash them all into one commit and attribute it to @sl4vr.
`+454 −346` => `+507 −321`
Let me go through all the code review notes once again, and check the remaining todos, but in general, this can be reviewed.

lib/rubocop/rspec/language.rb
```diff
@@ -2,152 +2,200 @@
...
+      # This is used in Dialect and DescribeClass cops to detect RSpec blocks
+      module ALL # :nodoc:
+        def self.all(element)
+          ExampleGroups.all(element) ||
+            Examples.all(element) ||
+            Helpers.all(element) ||
+            Hooks.all(element) ||
+            Includes.all(element) ||
+            SharedGroups.all(element) ||
+            Subjects.all(element)
```
pirj, line 193: I've left out `Expectations.all` and `Runners.all` from here. They don't seem to be needed for our cops (`Dialect` and `DescribeClass`), and it triggered Cyclomatic Complexity cop.

spec/shared/default_rspec_language_config_context.rb
```diff
@@ -0,0 +1,27 @@
+# frozen_string_literal: true
+
+RSpec.shared_context 'with default AllCops.RSpec.Language config', :config do
```
pirj, line 3: This will be needed by custom cops if they have specs for DSL customization.
Can't think why this would be necessary for release 2.0 though.
We don't enforce using alias configuration, and custom cops can continue to use their own configuration until someone decides to follow the mainstream, and would like to contribute by extracting this somewhere to `lib/`.

sl4vr:
I just feel a bit reluctant about that `all` method, which is actually always "all". I mean in method chain variant `block_pattern('#example_groups.regular.all')` should actually be just `block_pattern('#example_groups.regular')`, while we assume only 3 nesting levels of course.

pirj:
@sl4vr Fair enough! I'll change it to `ExampleGroups.all`/`ExampleGroups.regular`/... then, what do you think?

marcandre:
Two things:
1) If you need `#foo.bar.baz` to work, It's easy to implement and it looks reasonable; just open an issue.
2) You can already experiment with a hack that defined a method `example_groups_regular_all` that calls `example_groups.regular.all` like:

```ruby
def block_pattern(arg)
  arg.gsub! /^#((?:\w+\.)+\w)/ do |chain|
    alias_name = chain.gsub('.', '_')
    module_eval "def #{alias_name}(*a); chain(*a); end"
    alias_name 
  end
end
```

lib/rubocop/cop/rspec/align_left_let_brace.rb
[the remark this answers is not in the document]
pirj: Fixed.

lib/rubocop-rspec.rb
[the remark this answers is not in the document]
pirj: Fixed.

lib/rubocop/cop/rspec/hook_argument.rb
```diff
@@ -64,13 +64,11 @@ class HookArgument < Base
         IMPLICIT_MSG = 'Omit the default `%<scope>p` argument for RSpec hooks.'
         EXPLICIT_MSG = 'Use `%<scope>p` for RSpec hooks.'
 
-        def_node_matcher :hook?, Hooks::ALL.node_pattern_union
-
-        def_node_matcher :scoped_hook, <<-PATTERN
-          (block $(send _ #hook? (sym ${:each :example})) ...)
+        def_node_matcher :scoped_block, <<-PATTERN
```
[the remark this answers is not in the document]
pirj: Fixed.

lib/rubocop/rspec/concept.rb
```diff
@@ -4,12 +4,13 @@ module RuboCop
...
     # Wrapper for RSpec DSL methods
     class Concept
+      extend RuboCop::NodePattern::Macros
+      extend Language::NodePattern
       include Language
-      include Language::NodePattern
-      extend NodePattern::Macros
 
-      def initialize(node)
+      def initialize(node, config)
```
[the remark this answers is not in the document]
pirj: Done

bquorning:
It looks like some specs fail when run alone, but pass when running the entire suite:

```
❯ git rev-parse @
ebfbfd56bdf5d16a88c8ad3cd9ecbf5151f23531

❯ bundle exec rspec ./spec/rubocop/rspec/example_group_spec.rb:43 | grep -A10 Failures:
Failures:

  1) RuboCop::RSpec::ExampleGroup exposes examples in scope
     Failure/Error: Language.config['Examples']['Regular'].include?(element.to_s)

     NoMethodError:
       undefined method `[]' for nil:NilClass
     # ./lib/rubocop/rspec/language.rb:90:in `regular'
     # ./lib/rubocop/rspec/language.rb:83:in `all'
     # ./lib/rubocop/rspec/language.rb:48:in `example?'
     # ./lib/rubocop/rspec/example_group.rb:59:in `public_send'

❯ bundle exec rspec spec/rubocop/rspec/hook_spec.rb:78 | grep -A10 Failures:
Failures:

  1) RuboCop::RSpec::Hook#metadata extracts hash metadata
     Failure/Error: Language.config['HookScopes'].include?(element.to_s)

     NoMethodError:
       undefined method `[]' for nil:NilClass
     # ./lib/rubocop/rspec/language.rb:127:in `all'
     # ./lib/rubocop/rspec/hook.rb:47:in `valid_scope?'
     # ./lib/rubocop/rspec/hook.rb:17:in `block in extract_metadata'
     # ./lib/rubocop/rspec/hook.rb:16:in `times'
```

pirj:
@bquorning True. Fixed this by tagging those specs `:config`, adding a cop to be investigated and calling `on_new_investigation` on this `cop`. It's kind of a trick to set `Language` for those specs that typically happens automatically when `expect_offense` conducts an investigation.
Specs in question were Concept specs (Example, ExampleGroup, Hook) that don't call `expect_offense`.

pirj:
@bquorning Both this PR's build and #1007 that I've just rebased onto this are green.

bquorning:
But how come the specs passed when run with the entire suite? Usually it means there is some global state being changed somewhere.

pirj:
There is indeed global state, in `Language.config` that is assigned when the first cop is initialized and starts the investigation.

lib/rubocop/cop/rspec/hooks_before_examples.rb
```diff
@@ -30,8 +30,8 @@ class HooksBeforeExamples < Base
 
         def_node_matcher :example_or_group?, <<-PATTERN
           {
-            #{(Examples::ALL + ExampleGroups::ALL).block_pattern}
-            #{Includes::EXAMPLES.send_pattern}
+            #{block_pattern('{#ExampleGroups.all #Examples.all}')}
+            #{send_pattern('#Includes.example')}
```
bquorning, line 34: I also just now noticed that `Include::EXAMPLES` was changed to `Includes.example`. I wonder if the plural form was better.

spec/rubocop/rspec/hook_spec.rb
```diff
@@ -1,8 +1,10 @@
 # frozen_string_literal: true
 
-RSpec.describe RuboCop::RSpec::Hook do
+RSpec.describe RuboCop::RSpec::Hook, :config do
```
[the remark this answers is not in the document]
bquorning: It seems that they do now :-)

bquorning (commented):
This PR looks *so* much better now. I think we are getting very close to merging.

spec/rubocop/rspec/hook_spec.rb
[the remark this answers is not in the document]
pirj: 😆

lib/rubocop/rspec/language.rb
```diff
@@ -2,152 +2,195 @@
...
+              skipped(element) ||
+              focused(element)
+          end
+
+          def regular(element)
+            Language.config['ExampleGroups']['Regular'].include?(element.to_s)
+          end
 
-      module ExampleGroups
-        GROUPS  = SelectorSet.new(%i[describe context feature example_group])
```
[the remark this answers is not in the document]
pirj: Hm. I lost track of this.
To me, "regular" sounds better. But we at least have to mention this in the update_to_v2 document.

lib/rubocop/rspec/language.rb
```diff
@@ -2,152 +2,195 @@
...
+        def self.all(element)
+          Language.config['Subjects'].include?(element.to_s)
+        end
       end
 
-      module Runners
-        ALL = SelectorSet.new(%i[to to_not not_to])
+      # This is used in Dialect and DescribeClass cops to detect RSpec blocks.
+      # Ommission of Expectations and Runners is a deliberate simplification,
+      # but not a requirement.
```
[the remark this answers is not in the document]
bquorning: Oh, I didn’t mean that they should necessarily be added back. You mentioned it was a deliberate simplification; I just wasn’t sure what that meant.

lib/rubocop/cop/rspec/empty_line_after_subject.rb
```diff
@@ -32,7 +32,7 @@ def on_block(node)
 
         def in_spec_block?(node)
           node.each_ancestor(:block).any? do |ancestor|
-            Examples::ALL.include?(ancestor.method_name)
+            Examples.all(ancestor.method_name)
```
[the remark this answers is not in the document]
bquorning: Returning different types (boolean vs array) from the same method also isn’t great. Maybe we should call `#all?` from within out patterns? Also not really a great option :-/

I think we should just leave it for now.

bquorning (approved):
I… have run out of comments 😅

Thank you for the great work @pirj and @sl4vr. 🎉 

Squash and merge at will.

lib/rubocop/rspec/language.rb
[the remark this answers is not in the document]
pirj: I was on the fence with keeping (and dealing with Complexity cops) and omitting them.
Omitting them would be confusing since for our cops there's no difference, but some third-party cops could rely on ALL to match all of RSpec concepts. Seems better to have them in.

pirj:
@sl4vr Thanks a lot for implementing this. It is a huge amount of work, really, I now understand this after making subtle changes. It's time to claim your [CoM](https://cultofmartians.com/tasks/rubocop-rspec-configuration.html) trophy, but it's not all, you also get this symbolic golden RuboCop statuette:
output-onlinepngtools


@marcandre Thanks a lot for early-stage reviews, incredible suggestions and making changes to `rubocop-ast` that allowed this change to be made. You get a golden RuboCop statuette as well, and I guess it's not your first one!
output-onlinepngtools

marcandre:
Thanks @pirj, you're too kind 😅 
Congrats to @sl4vr, @bquorning and yourself for landing this 🎉

sl4vr:
@pirj thank you for leading and finishing this PR 😅

PR #960, opened 2020-07-10
Don't spend CPU finding the same node twice

bquorning:
As mentioned in https://github.com/rubocop-hq/rubocop-rspec/pull/955#discussion_r451809657

---

* [-] Added tests.
* [-] Updated documentation.

### DIFF
--- a/lib/rubocop/cop/rspec/leading_subject.rb
+++ b/lib/rubocop/cop/rspec/leading_subject.rb
@@ -47,7 +47,7 @@ module RuboCop
             if offending?(sibling)
               msg = format(MSG, offending: sibling.method_name)
               add_offense(node, message: msg) do |corrector|
-                autocorrect(corrector, node)
+                autocorrect(corrector, node, sibling)
               end
             end
 
@@ -57,19 +57,14 @@ module RuboCop
 
         private
 
-        def autocorrect(corrector, node)
-          first_node = find_first_offending_node(node)
+        def autocorrect(corrector, node, sibling)
           RuboCop::RSpec::Corrector::MoveNode.new(
             node, corrector, processed_source
-          ).move_before(first_node)
+          ).move_before(sibling)
         end
 
         def offending?(node)
           let?(node) || hook?(node) || example?(node)
-        end
-
-        def find_first_offending_node(node)
-          node.parent.children.find { |sibling| offending?(sibling) }
         end
 
         def in_spec_block?(node)
--- a/lib/rubocop/cop/rspec/leading_subject.rb
+++ b/lib/rubocop/cop/rspec/leading_subject.rb
@@ -43,19 +43,23 @@ module RuboCop
         end
 
         def check_previous_nodes(node)
-          node.parent.each_child_node do |sibling|
-            if offending?(sibling)
-              msg = format(MSG, offending: sibling.method_name)
-              add_offense(node, message: msg) do |corrector|
-                autocorrect(corrector, node, sibling)
-              end
+          offending_node(node) do |offender|
+            msg = format(MSG, offending: offender.method_name)
+            add_offense(node, message: msg) do |corrector|
+              autocorrect(corrector, node, offender)
             end
-
-            break if offending?(sibling) || sibling.equal?(node)
           end
         end
 
         private
+
+        def offending_node(node)
+          node.parent.each_child_node.find do |sibling|
+            break if sibling.equal?(node)
+
+            yield sibling if offending?(sibling)
+          end
+        end
 
         def autocorrect(corrector, node, sibling)
           RuboCop::RSpec::Corrector::MoveNode.new(
--- a/spec/rubocop/cop/rspec/leading_subject_spec.rb
+++ b/spec/rubocop/cop/rspec/leading_subject_spec.rb
@@ -135,4 +135,29 @@ RSpec.describe RuboCop::Cop::RSpec::LeadingSubject do
       end
     RUBY
   end
+
+  it 'checks also when subject is below a non-offending node' do
+    expect_offense(<<~RUBY)
+      RSpec.describe do
+        def helper_method
+        end
+
+        it { is_expected.to be_present }
+
+        subject { described_class.new }
+        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ Declare `subject` above any other `it` declarations.
+      end
+    RUBY
+
+    expect_correction(<<~RUBY)
+      RSpec.describe do
+        def helper_method
+        end
+
+        subject { described_class.new }
+        it { is_expected.to be_present }
+
+      end
+    RUBY
+  end
 end


lib/rubocop/cop/rspec/leading_subject.rb
```diff
@@ -57,21 +57,16 @@ def check_previous_nodes(node)
 
         private
 
-        def autocorrect(corrector, node)
-          first_node = find_first_offending_node(node)
+        def autocorrect(corrector, node, sibling)
           RuboCop::RSpec::Corrector::MoveNode.new(
             node, corrector, processed_source
-          ).move_before(first_node)
+          ).move_before(sibling)
```
Darhazer, line 63: I wonder does it change autocorrect with multiple offending nodes, or the first_offending_node was recalculated after each autocorrelation?

bquorning: I don’t think I understand your question. Could you add an example?

Darhazer:
```
RSpec.describe User do
        let(:params) { foo } # first_offsending_node
        let(:bar) { baz } # sibling

        subject { described_class.new }
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ Declare `subject` above any other `let` declarations.
      end
```

Previously it would always move before the first offending node. Now in each iteration, it would move before the current offending node. Unless those are re-evaluated. I added the test and it's not broken.
And the reason for this is that we break after the first offending node, so `sibling` and `first_offending_node` are always the same. I missed that detail in the review

Darhazer: Perhaps the readability of the method could be improved as well

```diff
         def check_previous_nodes(node)
-          node.parent.each_child_node do |sibling|
-            if offending?(sibling)
-              msg = format(MSG, offending: sibling.method_name)
-              add_offense(node, message: msg) do |corrector|
-                autocorrect(corrector, node, sibling)
-              end
+          offending_node(node) do |offender|
+            msg = format(MSG, offending: offender.method_name)
+            add_offense(node, message: msg) do |corrector|
+              autocorrect(corrector, node, offender)
             end
-
-            break if offending?(sibling) || sibling.equal?(node)
           end
         end
 
+        def offending_node(node)
+          node.parent.each_child_node.find do |sibling|
+            break if sibling.equal?(node)
+
+            yield sibling if offending?(sibling)
+          end
+        end
+
```

bquorning: I like where this is going.

I tried replacing the last block body with
```rb
break if sibling.equal?(node)
break unless offending?(sibling)

yield sibling
```
and all tests still pass. if feels wrong. Do we need another test case, or is it ok to rewrite `yield sibling if offending?(sibling)`?

lib/rubocop/cop/rspec/leading_subject.rb
```diff
@@ -57,21 +57,16 @@ def check_previous_nodes(node)
...
         def offending?(node)
           let?(node) || hook?(node) || example?(node)
         end
 
-        def find_first_offending_node(node)
-          node.parent.children.find { |sibling| offending?(sibling) }
-        end
-
         def in_spec_block?(node)
           node.each_ancestor(:block).any? do |ancestor|
```
pirj, line 71: Observation:
`any?` it feels weird. `first` should work just fine. Can't imagine `subject(...) { }` being defined in an iterator, neither I'm aware of any special syntax to wrap `subject`, e.g. `isolated { subject(...) { } }`. `first` should work fine for most cases.

lib/rubocop/cop/rspec/leading_subject.rb
pirj: I guess it's `break unless offending?` doing a bad job. Shouldn't it be `next`?
Still, it's something worth adding a red spec. Maybe something like:
```
describe do
  def helper_method # `break unless offending?(sibling) would break here and skip subsequent offences
  end

  it { } # offence!

  subject(:lost_in_space) { ... }
end
```

pirj:
```ruby
node.parent.each_child_node
```
👍 

Observation:
Wondering if `each_child_node` accepts a type (`each_child_node(:block)`), we could narrow down our search to only let/hooks/examples.

lib/rubocop/cop/rspec/leading_subject.rb
```diff
@@ -57,21 +57,16 @@ def check_previous_nodes(node)
...
-          first_node = find_first_offending_node(node)
+        def autocorrect(corrector, node, sibling)
           RuboCop::RSpec::Corrector::MoveNode.new(
             node, corrector, processed_source
-          ).move_before(first_node)
+          ).move_before(sibling)
         end
 
         def offending?(node)
           let?(node) || hook?(node) || example?(node)
```
pirj, line 67: Observation:
Shouldn't `|| example_group?(node)` be added here?
I can think of:
```ruby
describe do
  describe do
    ...
  end

  subject(:i_am_offended_by_a_preceeding_example_group_where_i_am_a_full_fledged_subject) { ... }
end
```

lib/rubocop/cop/rspec/leading_subject.rb
Darhazer: @bquorning nice catch!
There is a missing test, having a non-offending nodes before the subject (like include_context).
```ruby
expect_offense(<<-RUBY)
      RSpec.describe User do
        include_context('from elsewhere')
        let(:params) { foo }
        let(:bz) { bar }

        subject { described_class.new }
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ Declare `subject` above any other `let` declarations.
      end
    RUBY

    expect_correction(<<-RUBY)
      RSpec.describe User do
        include_context('from elsewhere')
        subject { described_class.new }
        let(:params) { foo }
        let(:bz) { bar }

      end
    RUBY
```

whether those should be offending is another matter.

Darhazer: It could be some `def` or any other ruby code actually that goes before the subject

bquorning: Fyi @pirj calling `each_child_node(:block)` fails for the spec containing `let(:user, &args[:build_user])`.

Darhazer: As far as I remember, you can pass multiple node types, like `each_child_node(:block, :blockpass)`

bquorning: I tried that but didn’t make it work :-)

Darhazer: Yeah, the block_pass is an argument, and the node itself is a send.
```
s(:send, nil, :let,
  s(:sym, :user),
  s(:block_pass,
    s(:send,
      s(:send, nil, :args), :[],
      s(:sym, :build_user))))
```
You can do each_child_node(:block, :send). Not sure if is worthy

PR #961, opened 2020-07-10
Extract some shared logic to BlankLineSeparation

bquorning:
In 0289cd892ba5e34b67140a98a985db476e3f10f1, a bunch of logic was extracted into `BlankLineSeparation` module. In this PR I extract another bit.

Please help me with the naming of the method. I am not sure if `check_missing_separating_lines` is the best name.

---

* [-] Added tests.
* [-] Updated documentation.

### DIFF
--- a/lib/rubocop/cop/rspec/empty_line_after_example_group.rb
+++ b/lib/rubocop/cop/rspec/empty_line_after_example_group.rb
@@ -31,13 +31,9 @@ module RuboCop
 
         def on_block(node)
           return unless example_group?(node)
-          return if last_child?(node)
 
-          missing_separating_line(node) do |location|
-            msg = format(MSG, example_group: node.method_name)
-            add_offense(location, message: msg) do |corrector|
-              corrector.insert_after(location.end, "\n")
-            end
+          missing_separating_line_offense(node) do |method|
+            format(MSG, example_group: method)
           end
         end
       end
--- a/lib/rubocop/cop/rspec/empty_line_after_example_group.rb
+++ b/lib/rubocop/cop/rspec/empty_line_after_example_group.rb
@@ -25,7 +25,7 @@ module RuboCop
       #
       class EmptyLineAfterExampleGroup < Cop
         extend AutoCorrector
-        include RuboCop::RSpec::BlankLineSeparation
+        include RuboCop::RSpec::EmptyLineSeparation
 
         MSG = 'Add an empty line after `%<example_group>s`.'
 
--- a/lib/rubocop/cop/rspec/empty_line_after_final_let.rb
+++ b/lib/rubocop/cop/rspec/empty_line_after_final_let.rb
@@ -18,7 +18,7 @@ module RuboCop
       #   it { does_something }
       class EmptyLineAfterFinalLet < Cop
         extend AutoCorrector
-        include RuboCop::RSpec::BlankLineSeparation
+        include RuboCop::RSpec::EmptyLineSeparation
 
         MSG = 'Add an empty line after the last `%<let>s`.'
 
--- a/lib/rubocop/cop/rspec/empty_line_after_final_let.rb
+++ b/lib/rubocop/cop/rspec/empty_line_after_final_let.rb
@@ -20,7 +20,7 @@ module RuboCop
         extend AutoCorrector
         include RuboCop::RSpec::BlankLineSeparation
 
-        MSG = 'Add an empty line after the last `let` block.'
+        MSG = 'Add an empty line after the last `%<let>s`.'
 
         def on_block(node)
           return unless example_group_with_body?(node)
@@ -31,7 +31,8 @@ module RuboCop
           return if last_child?(latest_let)
 
           missing_separating_line(latest_let) do |location|
-            add_offense(location) do |corrector|
+            msg = format(MSG, let: latest_let.method_name)
+            add_offense(location, message: msg) do |corrector|
               corrector.insert_after(location.end, "\n")
             end
           end
--- a/lib/rubocop/cop/rspec/empty_line_after_final_let.rb
+++ b/lib/rubocop/cop/rspec/empty_line_after_final_let.rb
@@ -28,13 +28,9 @@ module RuboCop
           latest_let = node.body.child_nodes.select { |child| let?(child) }.last
 
           return if latest_let.nil?
-          return if last_child?(latest_let)
 
-          missing_separating_line(latest_let) do |location|
-            msg = format(MSG, let: latest_let.method_name)
-            add_offense(location, message: msg) do |corrector|
-              corrector.insert_after(location.end, "\n")
-            end
+          missing_separating_line_offense(latest_let) do |method|
+            format(MSG, let: method)
           end
         end
       end
--- a/lib/rubocop/cop/rspec/empty_line_after_hook.rb
+++ b/lib/rubocop/cop/rspec/empty_line_after_hook.rb
@@ -41,13 +41,9 @@ module RuboCop
 
         def on_block(node)
           return unless hook?(node)
-          return if last_child?(node)
 
-          missing_separating_line(node) do |location|
-            msg = format(MSG, hook: node.method_name)
-            add_offense(location, message: msg) do |corrector|
-              corrector.insert_after(location.end, "\n")
-            end
+          missing_separating_line_offense(node) do |method|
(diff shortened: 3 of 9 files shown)


Darhazer:
How about `missing_separating_line_offense`
I also wonder if we can have a `message` method in order not to initialize a message without having an offense.

Darhazer:
Another option for a cleaner API would be
```
missing_separating_line_offsense(node) do |message|
  message << format(....)
end
```
this way it's obvious that the offense is something that could be registered. But I don't quite like providing the argument just for the sake of modifying it with the message, and yielding or returning the message out of the block is just non-intuitive API

Anyway, those are just a couple of suggestions

lib/rubocop/rspec/blank_line_separation.rb
```diff
@@ -8,6 +8,16 @@ module BlankLineSeparation
       include FinalEndLocation
       include RuboCop::Cop::RangeHelp
 
+      def check_missing_separating_lines(node, message:)
```
pirj, line 11: Will it work with a default value `message: MSG`?

lib/rubocop/rspec/blank_line_separation.rb
```diff
@@ -8,6 +8,16 @@ module BlankLineSeparation
       include FinalEndLocation
```
pirj, line 8: Just a terminological nitpick - is it a *blank* line, or an *empty* line semantically?

lib/rubocop/cop/rspec/empty_line_after_final_let.rb
```diff
@@ -28,13 +28,8 @@ def on_block(node)
           latest_let = node.body.child_nodes.select { |child| let?(child) }.last
 
           return if latest_let.nil?
```
pirj, line 30: `return unless latest_let`?
This #955 has caused an avalanche. :D

lib/rubocop/cop/rspec/empty_line_after_example_group.rb
```diff
@@ -31,14 +31,9 @@ class EmptyLineAfterExampleGroup < Cop
 
         def on_block(node)
           return unless example_group?(node)
```
pirj, line 33: Will it work as
```ruby
example_group?(node) do |example_group|
  msg = format(MSG, example_group: example_group.method_name)
  check_missing_separating_lines(example_group, message: msg)
end
```

pirj: Same above for `EmptyLineAfterExample` and below for `EmptyLineAfterHook`.

lib/rubocop/cop/rspec/empty_line_after_subject.rb
```diff
@@ -22,13 +22,8 @@ class EmptyLineAfterSubject < Cop
```
pirj, line 22: The message does not make a difference between `subject` and `subject!`, while other cops indicate what hook/example/example group method was used.

lib/rubocop/cop/rspec/empty_line_after_hook.rb
```diff
@@ -41,14 +41,9 @@ class EmptyLineAfterHook < Cop
```
pirj, line 41: Wondering where the `s` from `%<hook>s` is, since in the spec I can't see it:
```ruby
        before { do_something }
        ^^^^^^^^^^^^^^^^^^^^^^^ Add an empty line after `before`.
```

lib/rubocop/cop/rspec/empty_line_after_final_let.rb
```diff
@@ -28,13 +28,8 @@ def on_block(node)
           latest_let = node.body.child_nodes.select { |child| let?(child) }.last
```
pirj, line 28: "'Add an empty line after the last `let` block." - isn't "block" redundant in this statement?

lib/rubocop/rspec/blank_line_separation.rb
bquorning: No, that would refer to `BlankLineSeparation::MSG`.

lib/rubocop/rspec/blank_line_separation.rb
bquorning: RuboCop settled on `Empty` instead of `Blank` in https://github.com/rubocop-hq/rubocop/issues/7077

lib/rubocop/cop/rspec/empty_line_after_example_group.rb
bquorning: That would be neat, but we’d have to add a capture in the right place in `RuboCop::RSpec::Language::SelectorSet`. Something to explore in a separate PR, maybe?

lib/rubocop/cop/rspec/empty_line_after_subject.rb
bquorning: Good point. I’m adding a commit to fix this oversight.

lib/rubocop/cop/rspec/empty_line_after_hook.rb
bquorning: `String#%` uses arguments similar to `Kernel#sprintf`, and the `s` refers to the `string` type. See https://ruby-doc.org/core-2.6.6/Kernel.html#method-i-sprintf

lib/rubocop/cop/rspec/empty_line_after_final_let.rb
bquorning: Good point, I’ll fix it. And have the error message reflect whether `let` or `let!` was called.

lib/rubocop/cop/rspec/empty_line_after_example_group.rb
Darhazer: It still could be
```
example_group?(node) do
  msg = format(MSG, example_group: node.method_name)
  check_missing_separating_lines(node, message: msg)
end
```
but then I would remove the `?` from the name. Still, something that could be explored in a separate PR

lib/rubocop/cop/rspec/empty_line_after_final_let.rb
bquorning: I’m not sure #955 is related to your suggestion.

lib/rubocop/rspec/blank_line_separation.rb
bquorning: Adding a commit to rename this file.

lib/rubocop/cop/rspec/empty_line_after_final_let.rb
pirj: Right, I seem to have mixed things up. But there was something that triggered a chain reaction of refactorings.

bquorning:
> How about `missing_separating_line_offense`

I like it.

> I also wonder if we can have a `message` method in order not to initialize a message without having an offense.

https://github.com/rubocop-hq/rubocop/blob/v0.88.0/lib/rubocop/cop/base.rb#L86-L91 says “Cops are discouraged to override this (`#message`); instead pass your message directly”.

I changed the logic to `yield` the method name of the offending node, to let the calling code generate the error message. Please review again.

lib/rubocop/cop/rspec/empty_line_after_example_group.rb
pirj: Up to you. Looks good as is.
There's always an indefinite space for improvement.

lib/rubocop/cop/rspec/empty_line_after_subject.rb
```diff
@@ -16,18 +16,15 @@ module RSpec
       #   let(:foo) { bar }
       class EmptyLineAfterSubject < Cop
         extend AutoCorrector
-        include RuboCop::RSpec::BlankLineSeparation
+        include RuboCop::RSpec::EmptyLineSeparation
 
-        MSG = 'Add empty line after `subject`.'
+        MSG = 'Add empty line after `%<subject>s`.'
```
pirj, line 21: The rest of the messages begin with "Add **an** empty line".

pirj:
Please merge at will 👍

lib/rubocop/cop/rspec/empty_line_after_subject.rb
bquorning: Fixed.

PR #964, opened 2020-07-11
Enable new cops

Darhazer:
After the update, lets enable the new Rubocop cops on our codebase. Almost everything was passing anyway
Note: this will keep all cops added enabled in rubocop edge, so we will get a warning for added cops that would introduce offenses. The alternative would be enabling of cops one by one. But our codebase is already the best, so what's the chance of introducing an offense :smile: 

---

* [x] ~Added tests.~
* [x] ~Added an entry to the `CHANGELOG.md` if the new code introduces user-observable changes.~

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -6,6 +6,7 @@ require:
 AllCops:
   DisplayCopNames: true
   TargetRubyVersion: 2.4
+  NewCops: enable
   Exclude:
     - 'vendor/**/*'
     - 'spec/fixtures/**/*'
--- a/lib/rubocop/cop/rspec/factory_bot/attribute_defined_statically.rb
+++ b/lib/rubocop/cop/rspec/factory_bot/attribute_defined_statically.rb
@@ -39,7 +39,7 @@ module RuboCop
 
           def on_block(node)
             attributes = factory_attributes(node) || []
-            attributes = [attributes] unless attributes.is_a?(Array)
+            attributes = [attributes] unless attributes.is_a?(Array) # rubocop:disable Style/ArrayCoercion, Lint/RedundantCopDisableDirective
 
             attributes.each do |attribute|
               next unless offensive_receiver?(attribute.receiver, node)
--- a/spec/project/changelog_spec.rb
+++ b/spec/project/changelog_spec.rb
@@ -26,7 +26,7 @@ RSpec.describe 'CHANGELOG.md' do
     describe 'link to related issue on github' do
       let(:issues) do
         entries.map do |entry|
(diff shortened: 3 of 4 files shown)


lib/rubocop/cop/rspec/factory_bot/attribute_defined_statically.rb
```diff
@@ -39,7 +39,7 @@ class AttributeDefinedStatically < Cop
 
           def on_block(node)
             attributes = factory_attributes(node) || []
-            attributes = [attributes] unless attributes.is_a?(Array)
+            attributes = [attributes] unless attributes.is_a?(Array) # rubocop:disable Style/ArrayCoercion, Lint/RedundantCopDisableDirective
```
bquorning, line 42: Why RedundantCopDisableDirective?

spec/project/changelog_spec.rb
```diff
@@ -69,7 +69,7 @@
       end
 
       it 'ends with a punctuation' do
-        expect(bodies).to all(match(/[\.\!]$/))
+        expect(bodies).to all(match(/[.!]$/))
```
bquorning, line 72: We could probably use `end_with` instead of `match`.

.rubocop.yml
```diff
@@ -6,6 +6,7 @@ require:
 AllCops:
   DisplayCopNames: true
   TargetRubyVersion: 2.4
+  NewCops: enable
```
pirj, line 9: 🙌

lib/rubocop/cop/rspec/factory_bot/attribute_defined_statically.rb
Darhazer: Because otherwise, it triggers an offense in versions that still do not have this cop, e.g. in the current one

bquorning: Ah, ok. It’s a bit confusing :-)

spec/project/changelog_spec.rb
Darhazer: @bquorning since there is a `end_with` matcher which doesn't support checking that a string ends with either one of the characters, this would become
```diff
-        expect(bodies).to all(match(/[.!]$/))
+       expect(bodies).to all(end_with('.').or(end_with('!')))
```
not sure if it's better though

bquorning: Nah, I was hoping for `end_with('.', '!')`. 

Let’s just leave it as it is.

Darhazer: Yeah, but that actually requires ending with `.!` ;)

PR #507, opened 2017-11-26
Improve performance when user does not override default RSpec Pattern config

walf443:
I noticed that RuboCop::Cop::RSpec::Cop#rspec_pattern is not so fast.
RSpec Pattern setting is not defined in most case, I think.
So change to cache building Regexp for default case.

Here is the stackprof result of running ./bin/rubocop at rubocop repositories on master

```
bundle exec stackprof tmp/stackprof-cpu-rubocop.dump
==================================
  Mode: cpu(1000)
  Samples: 52150 (0.02% miss rate)
  GC: 3229 (6.19%)
==================================
     TOTAL    (pct)     SAMPLES    (pct)     FRAME
      8039  (15.4%)        4264   (8.2%)     RuboCop::AST::Node#each_child_node
      3352   (6.4%)        3352   (6.4%)     block (2 levels) in <class:Node>
      3229   (6.2%)        3229   (6.2%)     (garbage collection)
      5362  (10.3%)        2846   (5.5%)     Parser::Lexer#advance
      2386   (4.6%)        2386   (4.6%)     AST::Node#to_a
      2133   (4.1%)        2133   (4.1%)     Parser::Source::Buffer#slice
      1421   (2.7%)        1061   (2.0%)     Parser::Source::Buffer#line_for_position
      1555   (3.0%)         908   (1.7%)     RuboCop::Cop::Cop#cop_config
       570   (1.1%)         570   (1.1%)     #<Module:0x00007f9dddefffb0>.of
       557   (1.1%)         557   (1.1%)     Parser::Source::Range#initialize
       620   (1.2%)         549   (1.1%)     RuboCop::Cop::RSpec::Cop#rspec_pattern
       651   (1.2%)         538   (1.0%)     RuboCop::ConfigStore#for
       479   (0.9%)         479   (0.9%)     RuboCop::Formatter::ProgressFormatter#report_file_as_mark
      1001   (1.9%)         440   (0.8%)     AST::Node#initialize
       592   (1.1%)         399   (0.8%)     RuboCop::Cop::Cop#cop_name
       429   (0.8%)         398   (0.8%)     RuboCop::Cop::VariableForce#scanned_node?
       561   (1.1%)         397   (0.8%)     Parser::AST::Node#assign_properties
       731   (1.4%)         396   (0.8%)     Parser::Source::Buffer#line_for
     18229  (35.0%)         360   (0.7%)     RuboCop::Cop::Commissioner#on_send
[...]

PR #965, opened 2020-07-13
Memoize Cop#rspec_pattern

marcandre:
Currently #2 in the memory allocated when running RuboCop... And #1 should disappear soon too 😅

See detailed list in https://github.com/whitequark/parser/pull/721

### DIFF
--- a/lib/rubocop/cop/rspec/cop.rb
+++ b/lib/rubocop/cop/rspec/cop.rb
@@ -45,7 +45,7 @@ module RuboCop
         end
 
         def rspec_pattern
-          if rspec_pattern_config?
+          @rspec_pattern ||= if rspec_pattern_config?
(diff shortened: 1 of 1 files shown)


bquorning:
If we are going to memoize anyway, we might as well revert the previous optimisation done in #507 ?

marcandre:
> If we are going to memoize anyway, we might as well revert the previous optimisation done in #507 ?

Good question. I just went for the quickest fix tbh, I didn't delve into the code. The memoization is only per cop though, that code might still execute too much anyways?

bquorning:
> I just went for the quickest fix tbh

Would you prefer that I take over the PR? I’d like to dig a bit more into how we can perform this calculation only once across all cops.

marcandre:
With pleasure. I was running my tests with `bin/rubocop-profile --memory lib/rubocop/cop/mixin`, but you should get similar results for other combinations I imagine.

PR #966, opened 2020-07-15
Also detect pre-subject example groups in `RSpec/LeadingSubject`

pirj:
---

* [x] Added tests.
* [-] Updated documentation.

### DIFF
--- a/lib/rubocop/rspec/language.rb
+++ b/lib/rubocop/rspec/language.rb
@@ -40,6 +40,10 @@ module RuboCop
           "(send #{RSPEC} #{node_pattern_union} ...)"
         end
 
+        def send_or_block_or_block_pass_pattern
+          "{#{send_pattern} #{block_pattern} #{block_pass_pattern}}"
+        end
+
         def node_pattern_union
           "{#{node_pattern}}"
         end
--- a/lib/rubocop/rspec/language/node_pattern.rb
+++ b/lib/rubocop/rspec/language/node_pattern.rb
@@ -23,6 +23,9 @@ module RuboCop
 
         def_node_matcher :let?, Helpers::ALL.block_or_block_pass_pattern
 
+        def_node_matcher :include?,
+                         Includes::ALL.send_or_block_or_block_pass_pattern
+
         def_node_matcher :subject?, Subject::ALL.block_pattern
       end
     end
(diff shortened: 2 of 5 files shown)


lib/rubocop/rspec/language/node_pattern.rb
```diff
@@ -23,6 +23,9 @@ module NodePattern
 
         def_node_matcher :let?, Helpers::ALL.block_or_block_pass_pattern
 
+        def_node_matcher :include?,
+                         Includes::ALL.send_or_block_or_block_pass_pattern
+
         def_node_matcher :subject?, Subject::ALL.block_pattern
```
pirj, line 29: Side note.
Wondering if we should add support for proper detection of:
```ruby
before(:all, &global)
describe(&block)
shared_examples_for('foo', &block)
```

I dislike this syntax, but we've started supporting it step by step.

I'm thinking of a NoBlockPass cop along the way XD

spec/rubocop/cop/rspec/leading_subject_spec.rb
```diff
@@ -160,4 +160,113 @@ def helper_method
...
+        it_behaves_like 'a good citizen', &block
+
+        subject { described_class.new }
+        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ Declare `subject` above any other `it_behaves_like` declarations.
+      end
+    RUBY
+
+    expect_correction(<<-RUBY)
+      RSpec.describe User do
+        block = ->{ }
```
bquorning, line 265: Ideally, the `block` would be below the subject too, right? This makes me thing, do we want to allow _anything_ above the `subject`?

pirj: I knew this question is coming :D
Didn't want to make this example too different from its siblings by making it a `shared_examples_for '...' do |&block|`.

The style guide [is pretty clear about that](https://rspec.rubystyle.guide/#leading-subject):
> When `subject` is used, it should be the first declaration in the example group.

I'd rather not change the behaviour in a radical way here. I can think of e.g. helper methods or custom matchers, some may prefer to have them in a section above subject.

lib/rubocop/rspec/language.rb
```diff
@@ -40,6 +40,10 @@ def send_pattern
           "(send #{RSPEC} #{node_pattern_union} ...)"
         end
 
+        def send_or_block_or_block_pass_pattern
```
Darhazer, line 43: I don't like where this is going, but currently, we do not have a better solution

PR #967, opened 2020-07-15
Optimize Base#relevant_rubocop_rspec_file?

bquorning:
Another take on #965.

The result of `rspec_pattern` is always the same (because it is based on the configuration file), so there is no need to re-calculate it every time it is called.

I tried memozing on the ~Cop~ Base instance, but at least when running the specs, we instantiate ~Cop~ Base way too often. Memoizing on the class level reduces the number of calculations to once per `Base` subclass we have. If we allowed using class variables (`@@rspec_pattern`) we could reduce the number of calculations to one.

This commit reverts 995b4fe5f16cd73feb2625eadd0fad236cfab38e and adds memoization in a class instance variable.

---

* [-] Added tests.
* [-] Updated documentation.

### DIFF
--- a/lib/rubocop/cop/rspec/base.rb
+++ b/lib/rubocop/cop/rspec/base.rb
@@ -21,14 +21,6 @@ module RuboCop
         include RuboCop::RSpec::Language
         include RuboCop::RSpec::Language::NodePattern
 
-        DEFAULT_CONFIGURATION =
-          RuboCop::RSpec::CONFIG.fetch('AllCops').fetch('RSpec')
-
-        DEFAULT_PATTERN_RE = Regexp.union(
-          DEFAULT_CONFIGURATION.fetch('Patterns')
-                               .map(&Regexp.public_method(:new))
-        )
-
         # Invoke the original inherited hook so our cops are recognized
         def self.inherited(subclass) # rubocop:disable Lint/MissingSuper
           RuboCop::Cop::Base.inherited(subclass)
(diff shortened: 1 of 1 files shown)


bquorning:
And of course this doesn‘t work with specs (spec/rubocop/cop/rspec/cop_spec.rb) that change the configuration.

lib/rubocop/cop/rspec/cop.rb
```diff
@@ -41,32 +33,28 @@ def relevant_file?(file)
...
-        end
-
-        def rspec_pattern_config
-          all_cops_config
-            .fetch('RSpec', DEFAULT_CONFIGURATION)
-            .fetch('Patterns')
+            Config.new
+              .for_all_cops
+              .fetch('RSpec', default_configuration)
+              .fetch('Patterns')
```
pirj, line 56: Isn't the configuration merged with the default configuration, can we assume that `AllCops/RSpec/Patterns` is always present?

lib/rubocop/cop/rspec/cop.rb
```diff
@@ -41,32 +33,28 @@ def relevant_file?(file)
         private
 
         def relevant_rubocop_rspec_file?(file)
-          rspec_pattern =~ file
+          self.class.rspec_pattern =~ file
```
marcandre, line 36: Use `match?`

pirj: We're introducing `rubocop-performance` to tackle that https://github.com/rubocop-hq/rubocop-rspec/pull/970

lib/rubocop/cop/rspec/cop.rb
bquorning: You are right. We can just use `ConfigLoader.default_configuration.for_all_cops` here.

bquorning: Still, I’m struggling to figure out how to make this work with the changing configurations in cop_spec.rb

pirj: Just found this https://github.com/rubocop-hq/rubocop/blob/8842e13621faa82c2688e6d5e7cd891e0aa5873c/lib/rubocop/rspec/shared_contexts.rb#L76
Maybe can be of some help?

pirj: I mean it could be part of 2.0, not even post-2.0. We have a bunch of pull requests specifically for 2.0.

bquorning:
Fixed in #1063.

PR #976, opened 2020-07-21
Ignore trailing punctuation in context prefix

elliterate:
By splitting on word boundaries rather than whitespace (the default), we can easily ignore trailing punctuation on the first word of the context description.

Note: I didn't update any documentation because this seemed more like a bug fix than a change in behavior.

---

* [x] Added tests.
* [ ] Updated documentation.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 * Add a new base cop class `::RuboCop::Cop::RSpec::Base`. The old base class `::RuboCop::Cop::RSpec::Cop` is deprecated, and will be removed in the next major release. ([@bquorning][])
 * Add support for subject detection after includes and example groups in `RSpec/LeadingSubject`. ([@pirj][])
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 
 * Add a new base cop class `::RuboCop::Cop::RSpec::Base`. The old base class `::RuboCop::Cop::RSpec::Cop` is deprecated, and will be removed in the next major release. ([@bquorning][])
 * Add support for subject detection after includes and example groups in `RSpec/LeadingSubject`. ([@pirj][])
+* Ignore trailing punctuation in context description prefix. ([@elliterate][])
```
pirj, line 7: An easy fix for red CI: add yourself in the end of the file.

elliterate: Ah, thanks! I just looked at recent changes to the file, which must have been from people with previous involvement. I remember wondering about the empty `[]`. Guess I should have pursued that line of thought a bit further.

PR #977, opened 2020-07-22
Improve cops by using TopLevelGroup

### LINKED ISSUE #948: Use `TopLevelGroup` instead of `TopLevelDescribe`
The newly introduced `TopLevelGroup` is better (see the pull request introducing it) since it not only detects `describe`, but other flavours of top-level example groups.
Consider using `TopLevelGroup` in our own cops.
Deprecate `TopLevelDescribe` and eventually remove in 2.0 (not earlier, since custom cops may depend on it).

pirj:
The previously used `TopLevelDescribe` suffered a few flaws:
 - it was slow: https://github.com/rubocop-hq/rubocop-rspec/pull/925#discussion_r436988718
 - it ignored non-`describe` top-level example groups https://github.com/rubocop-hq/rubocop-rspec/pull/925#issuecomment-641140259

This pull request fixes those problems by using a newly introduced `TopLevelGroup`.
Also fixes a flaw in `TopLevelGroup` when example groups wrapped inside `module` or `class` were missed.

NOT DONE: I'm on the fence if we should deprecate `TopLevelDescribe` before removing it. We most probably don't have much releases left before 2.0, and the majority of people will jump straight to 2.0 skipping that pre-2.0 release. We may add this pull request (and others that replace usages of `TopLevelDescribe` for `TopLevelGroup`) as reference examples for migration.

Fixes #948

---

* [x] Added tests.
* [x] Updated documentation.

### DIFF
--- a/lib/rubocop/cop/rspec/describe_class.rb
+++ b/lib/rubocop/cop/rspec/describe_class.rb
@@ -3,7 +3,7 @@
 module RuboCop
   module Cop
     module RSpec
-      # Check that the first argument to the top level describe is a constant.
+      # Check that the first argument to the top-level describe is a constant.
       #
       # @example
       #   # bad
@@ -22,49 +22,40 @@ module RuboCop
       #   describe "A feature example", type: :feature do
       #   end
       class DescribeClass < Base
-        include RuboCop::RSpec::TopLevelDescribe
+        include RuboCop::RSpec::TopLevelGroup
 
         MSG = 'The first argument to describe should be '\
               'the class or module being tested.'
 
-        def_node_matcher :valid_describe?, <<-PATTERN
-          {
-            (send #rspec? :describe const ...)
-            (send #rspec? :describe)
-          }
-        PATTERN
-
-        def_node_matcher :describe_with_rails_metadata?, <<-PATTERN
-          (send #rspec? :describe !const ...
-            (hash <#rails_metadata? ...>)
-          )
-        PATTERN
-
         def_node_matcher :rails_metadata?, <<-PATTERN
           (pair
             (sym :type)
-            (sym {
-                   :channel :controller :helper :job :mailer :model :request
-                   :routing :view :feature :system :mailbox
-                 }
-            )
+            (sym { :channel :controller :helper :job :mailer :model :request
+                   :routing :view :feature :system :mailbox })
           )
         PATTERN
 
-        def on_top_level_describe(node, (described_value, _))
-          return if shared_group?(root_node)
-          return if valid_describe?(node)
-          return if describe_with_rails_metadata?(node)
-          return if string_constant_describe?(described_value)
+        def_node_matcher :example_group_with_rails_metadata?, <<~PATTERN
+          (send #rspec? :describe ... (hash <#rails_metadata? ...>))
+        PATTERN
 
-          add_offense(described_value)
+        def_node_matcher :not_a_const_described, <<~PATTERN
+          (send #rspec? :describe $[!const !#string_constant?] ...)
+        PATTERN
+
+        def on_top_level_group(top_level_node)
+          return if example_group_with_rails_metadata?(top_level_node.send_node)
+
+          not_a_const_described(top_level_node.send_node) do |described|
+            add_offense(described)
+          end
         end
 
         private
 
-        def string_constant_describe?(described_value)
-          described_value.str_type? &&
-            described_value.value.match?(/^(?:(?:::)?[A-Z]\w*)+$/)
+        def string_constant?(described)
+          described.str_type? &&
+            described.value.match?(/^(?:(?:::)?[A-Z]\w*)+$/)
         end
       end
     end
--- a/lib/rubocop/cop/rspec/describe_method.rb
+++ b/lib/rubocop/cop/rspec/describe_method.rb
@@ -17,16 +17,24 @@ module RuboCop
       #   describe MyClass, '.my_class_method' do
       #   end
       class DescribeMethod < Base
-        include RuboCop::RSpec::TopLevelDescribe
+        include RuboCop::RSpec::TopLevelGroup
 
         MSG = 'The second argument to describe should be the method '\
               "being tested. '#instance' or '.class'."
 
-        def on_top_level_describe(_node, (_, second_arg))
-          return unless second_arg&.str_type?
-          return if second_arg.str_content.start_with?('#', '.')
+        def_node_matcher :second_argument, <<~PATTERN
+          (block
+            (send #rspec? :describe _first_argument $(str _) ...) ...
+          )
+        PATTERN
 
-          add_offense(second_arg)
+        def on_top_level_group(node)
+          second_argument = second_argument(node)
+
+          return unless second_argument
+          return if second_argument.str_content.start_with?('#', '.')
+
+          add_offense(second_argument)
         end
       end
     end
--- a/lib/rubocop/cop/rspec/file_path.rb
+++ b/lib/rubocop/cop/rspec/file_path.rb
@@ -57,34 +57,42 @@ module RuboCop
       #   my_class_spec.rb         # describe MyClass, '#method'
       #
       class FilePath < Base
-        include RuboCop::RSpec::TopLevelDescribe
+        include RuboCop::RSpec::TopLevelGroup
 
         MSG = 'Spec path should end with `%<suffix>s`.'
 
-        def_node_search :const_described?,  '(send _ :describe (const ...) ...)'
+        def_node_matcher :const_described, <<~PATTERN
+          (block
+            $(send #rspec? _example_group $(const ...) $...) ...
+          )
+        PATTERN
+
         def_node_search :routing_metadata?, '(pair (sym :type) (sym :routing))'
 
-        def on_top_level_describe(node, args)
-          return unless const_described?(node) && single_top_level_describe?
-          return if routing_spec?(args)
+        def on_top_level_group(node)
+          return unless top_level_groups.one?
 
-          glob = glob_for(args)
+          const_described(node) do |send_node, described_class, arguments|
+            next if routing_spec?(arguments)
 
-          return if filename_ends_with?(glob)
-
-          add_offense(
-            node,
-            message: format(MSG, suffix: glob)
-          )
+            ensure_correct_file_path(send_node, described_class, arguments)
+          end
         end
 
         private
+
+        def ensure_correct_file_path(send_node, described_class, arguments)
+          glob = glob_for(described_class, arguments.first)
+          return if filename_ends_with?(glob)
+
+          add_offense(send_node, message: format(MSG, suffix: glob))
+        end
 
         def routing_spec?(args)
           args.any?(&method(:routing_metadata?))
         end
 
-        def glob_for((described_class, method_name))
+        def glob_for(described_class, method_name)
           return glob_for_spec_suffix_only? if spec_suffix_only?
 
           "#{expected_path(described_class)}#{name_glob(method_name)}*_spec.rb"
@@ -94,10 +102,10 @@ module RuboCop
           '*_spec.rb'
         end
 
-        def name_glob(name)
-          return unless name&.str_type?
+        def name_glob(method_name)
+          return unless method_name&.str_type?
 
-          "*#{name.str_content.gsub(/\W/, '')}" unless ignore_methods?
+          "*#{method_name.str_content.gsub(/\W/, '')}" unless ignore_methods?
         end
 
         def expected_path(constant)
--- a/lib/rubocop/rspec/top_level_group.rb
+++ b/lib/rubocop/rspec/top_level_group.rb
@@ -25,14 +25,16 @@ module RuboCop
 
       def top_level_groups
         @top_level_groups ||=
-          top_level_nodes.select { |n| example_or_shared_group?(n) }
+          top_level_nodes(root_node).select { |n| example_or_shared_group?(n) }
       end
 
-      def top_level_nodes
-        if root_node.begin_type?
-          root_node.children
+      def top_level_nodes(node)
+        if node.begin_type?
+          node.children
+        elsif node.module_type? || node.class_type?
+          top_level_nodes(node.body)
         else
-          [root_node]
+          [node]
         end
       end
 
--- a/lib/rubocop/rspec/top_level_group.rb
+++ b/lib/rubocop/rspec/top_level_group.rb
@@ -10,22 +10,31 @@ module RuboCop
       def_node_matcher :example_or_shared_group?,
                        (ExampleGroups::ALL + SharedGroups::ALL).block_pattern
 
-      def on_block(node)
-        return unless respond_to?(:on_top_level_group)
-        return unless top_level_group?(node)
+      def on_new_investigation
+        super
 
-        on_top_level_group(node)
-      end
+        return unless root_node
 
-      private
-
-      def top_level_group?(node)
-        top_level_groups.include?(node)
+        top_level_groups.each do |node|
+          example_group?(node, &method(:on_top_level_example_group))
+          on_top_level_group(node)
+        end
       end
 
       def top_level_groups
         @top_level_groups ||=
           top_level_nodes(root_node).select { |n| example_or_shared_group?(n) }
+      end
+
+      private
+
+      # Dummy methods to be overridden in the consumer
+      def on_top_level_example_group; end
+
+      def on_top_level_group; end
+
+      def top_level_group?(node)
+        top_level_groups.include?(node)
       end
 
       def top_level_nodes(node)
--- a/spec/rubocop/cop/rspec/multiple_describes_spec.rb
+++ b/spec/rubocop/cop/rspec/multiple_describes_spec.rb
@@ -3,26 +3,51 @@
 RSpec.describe RuboCop::Cop::RSpec::MultipleDescribes do
   subject(:cop) { described_class.new }
 
-  it 'finds multiple top level describes with class and method' do
+  it 'flags multiple top-level example groups with class and method' do
     expect_offense(<<-RUBY)
       describe MyClass, '.do_something' do; end
-      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ Do not use multiple top level describes - try to nest them.
+      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ Do not use multiple top-level example groups - try to nest them.
       describe MyClass, '.do_something_else' do; end
     RUBY
   end
 
-  it 'finds multiple top level describes only with class' do
+  it 'flags multiple top-level example groups only with class' do
     expect_offense(<<-RUBY)
       describe MyClass do; end
-      ^^^^^^^^^^^^^^^^ Do not use multiple top level describes - try to nest them.
+      ^^^^^^^^^^^^^^^^ Do not use multiple top-level example groups - try to nest them.
       describe MyOtherClass do; end
     RUBY
   end
 
-  it 'skips single top level describe' do
-    expect_no_offenses(<<-RUBY)
-      require 'spec_helper'
+  it 'flags multiple top-level example groups with an arbitrary argument' do
+    expect_offense(<<-RUBY)
+      describe 'MyClass' do; end
+      ^^^^^^^^^^^^^^^^^^ Do not use multiple top-level example groups - try to nest them.
+      describe 'MyOtherClass' do; end
+    RUBY
+  end
 
+  it 'flags multiple top-level example groups aliases' do
+    expect_offense(<<-RUBY)
+      example_group MyClass do; end
+      ^^^^^^^^^^^^^^^^^^^^^ Do not use multiple top-level example groups - try to nest them.
+      feature MyOtherClass do; end
+    RUBY
+  end
+
+  it 'ignores single top-level example group' do
+    expect_no_offenses(<<-RUBY)
+      describe MyClass do
+      end
+    RUBY
+  end
+
+  it 'ignores multiple shared example groups' do
+    expect_no_offenses(<<-RUBY)
+      shared_examples_for 'behaves' do
+      end
+      shared_examples_for 'misbehaves' do
+      end
       describe MyClass do
       end
     RUBY
--- a/spec/rubocop/cop/rspec/nested_groups_spec.rb
+++ b/spec/rubocop/cop/rspec/nested_groups_spec.rb
@@ -35,11 +35,32 @@ RSpec.describe RuboCop::Cop::RSpec::NestedGroups, :config do
     expect(cop.config_to_allow_offenses[:exclude_limit]).to eq('Max' => 4)
   end
 
-  it 'ignores non-spec context methods' do
-    expect_no_offenses(<<-RUBY)
+  it 'flags example groups wrapped in classes' do
+    expect_offense(<<-RUBY)
       class MyThingy
-        context 'this is not rspec' do
-          context 'but it uses contexts' do
+        describe MyClass do
+          context 'when foo' do
+            context 'when bar' do
+              context 'when baz' do
+              ^^^^^^^^^^^^^^^^^^ Maximum example group nesting exceeded [4/3].
+              end
+            end
+          end
+        end
+      end
+    RUBY
+  end
+
+  it 'flags example groups wrapped in modules' do
+    expect_offense(<<-RUBY)
+      module MyNamespace
+        describe MyClass do
+          context 'when foo' do
+            context 'when bar' do
+              context 'when baz' do
+              ^^^^^^^^^^^^^^^^^^ Maximum example group nesting exceeded [4/3].
+              end
+            end
           end
         end
       end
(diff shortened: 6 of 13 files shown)


lib/rubocop/cop/rspec/file_path.rb
```diff
@@ -57,34 +57,44 @@ module RSpec
...
 
-          return if filename_ends_with?(glob)
+          return unless described_class
+          return if routing_spec?(arguments)
 
-          add_offense(
-            node,
-            message: format(MSG, suffix: glob)
-          )
+          ensure_correct_file_path(send_node, described_class, arguments)
```
pirj, line 80: Code climate insisted to break this apart.

spec/rubocop/cop/rspec/multiple_describes_spec.rb
```diff
@@ -3,28 +3,36 @@
...
-
       describe MyClass do
       end
     RUBY
   end
+
+  it 'flags shared example groups' do
+    expect_offense(<<-RUBY)
+      shared_examples_for 'behaves' do
+      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ Do not use multiple top-level example groups - try to nest them.
```
pirj, line 32: This is a bigger change, but I believe a good one.

spec/rubocop/cop/rspec/nested_groups_spec.rb
```diff
@@ -35,11 +54,32 @@
     expect(cop.config_to_allow_offenses[:exclude_limit]).to eq('Max' => 4)
   end
 
-  it 'ignores non-spec context methods' do
```
pirj, line 38: It wasn't ignoring. Just the example didn't have enough nesting to trigger the offence.

lib/rubocop/cop/rspec/describe_class.rb
```diff
@@ -51,13 +51,21 @@ class DescribeClass < Base
...
-          return if describe_with_rails_metadata?(node)
-          return if string_constant_describe?(described_value)
+        def_node_matcher :described, <<~PATTERN
+          (block $(send #{RSPEC} :describe $_ ...) ...)
+        PATTERN
+
+        def on_top_level_group(top_level_node)
+          node, described = described(top_level_node)
+
+          return unless described
```
Darhazer, line 61: Can we have `!nil?` in the matcher and skip this check

lib/rubocop/cop/rspec/describe_method.rb
```diff
@@ -17,16 +17,24 @@ module RSpec
...
-          return if second_arg.str_content.start_with?('#', '.')
+        def_node_matcher :second_argument, <<~PATTERN
+          (block
+            (send #{RSPEC} :describe _first_argument $(str _) ...) ...
+          )
+        PATTERN
 
-          add_offense(second_arg)
+        def on_top_level_group(node)
+          second_argument = second_argument(node)
```
Darhazer, line 32: we implicitly check that this is a `describe` this way. I'm not strictly against it, but since some cops need to operate only on describe, I wonder if we can't customize the on_top_level_group to actually have separate callbacks for on_top_level_describe. We still use the same code, just call different method if it is available and the top level group is a describe. And we don't make any breaking change in the API

spec/rubocop/cop/rspec/multiple_describes_spec.rb
Darhazer: I think it's quite common to define multiple shared examples in a single, helper file?

pirj: Agree, will narrow it back down to example groups only 👍

We may think of this option later on.

lib/rubocop/cop/rspec/describe_class.rb
pirj: I keep forgetting about the yielding feature of a node pattern.

This comment of yours, however, led to a major refactoring 😆

lib/rubocop/cop/rspec/describe_method.rb
pirj: Implemented separate hooks, `on_top_level_example_group` and `on_top_level_group` 👍

lib/rubocop/cop/rspec/describe_class.rb
```diff
@@ -22,49 +22,38 @@ module RSpec
...
-          return if valid_describe?(node)
-          return if describe_with_rails_metadata?(node)
-          return if string_constant_describe?(described_value)
+        def_node_matcher :not_a_const_described, <<~PATTERN
+          (block
+            [  (send #{RSPEC} :describe $[!const !#string_constant?] ...)
+              !(send #{RSPEC} :describe ... (hash <#rails_metadata? ...>)) ]
+            ...
+          )
+        PATTERN
```
pirj, line 44: Collapsed several significantly overlapping node patterns into one.
Moved guard statements into node pattern as well.

There still seems to be some redundancy in this node pattern, similarly to `SubjectStub`:
```
          {
            (block (send nil? :subject (sym $_)) args ...)
            (block (send nil? $:subject) args ...)
          }
```

I'll open a ticket for `rubocop-ast`.

lib/rubocop/cop/rspec/describe_method.rb
```diff
@@ -17,16 +17,24 @@ module RSpec
...
+
+          return unless second_argument
+          return if second_argument.str_content.start_with?('#', '.')
```
pirj, line 35: I'm ok to turn this into a method and pushing this into the node pattern, but decided not to fix everything in one go.

lib/rubocop/cop/rspec/file_path.rb
```diff
@@ -57,34 +57,44 @@ module RSpec
...
+        def_node_matcher :const_described, <<~PATTERN
+          (block
+            $(send #{RSPEC} _example_group $(const ...) $...) ...
+          )
+        PATTERN
+
         def_node_search :routing_metadata?, '(pair (sym :type) (sym :routing))'
 
-        def on_top_level_describe(node, args)
-          return unless const_described?(node) && single_top_level_describe?
```
pirj, line 68: That was the only remaining usage of `single_top_level_describe?`, and I decided to inline it.

lib/rubocop/cop/rspec/file_path.rb
```diff
@@ -57,34 +57,43 @@ module RSpec
...
+            next unless described_class
+            next if routing_spec?(arguments)
 
-          return if filename_ends_with?(glob)
-
-          add_offense(
-            node,
-            message: format(MSG, suffix: glob)
-          )
+            ensure_correct_file_path(send_node, described_class, arguments)
```
pirj, line 78: Code climate insisted on this extraction.

lib/rubocop/rspec/top_level_group.rb
```diff
@@ -10,11 +10,21 @@ module TopLevelGroup
       def_node_matcher :example_or_shared_group?,
                        (ExampleGroups::ALL + SharedGroups::ALL).block_pattern
 
-      def on_block(node)
-        return unless respond_to?(:on_top_level_group)
-        return unless top_level_group?(node)
+      def on_new_investigation
```
pirj, line 13: This is a hair more efficient than `on_block`.

lib/rubocop/rspec/top_level_group.rb
```diff
@@ -10,11 +10,21 @@ module TopLevelGroup
...
+            example_group?(node, &method(:on_top_level_example_group))
+          end
 
-        on_top_level_group(node)
+          on_top_level_group(node) if respond_to?(:on_top_level_group)
+        end
+      end
+
+      def top_level_groups
+        return [] unless root_node
```
pirj, line 24: To satisfy our "empty spec" smoke test. Is there a better way around this?

lib/rubocop/rspec/top_level_group.rb
```diff
@@ -23,16 +33,13 @@ def top_level_group?(node)
...
-          top_level_nodes.select { |n| example_or_shared_group?(n) }
-      end
-
-      def top_level_nodes
-        if root_node.begin_type?
-          root_node.children
+      def top_level_nodes(node)
+        if node.begin_type?
+          node.children
+        elsif node.module_type? || node.class_type?
```
pirj, line 39: Previously,
```ruby
module A
  class B
    RSpec.describe 'C' do
```
was missed.

spec/rubocop/cop/rspec/nested_groups_spec.rb
```diff
@@ -20,6 +20,36 @@
...
+  end
+
+  it 'flags nested example groups inside shared examples' do
+    expect_offense(<<-RUBY)
+      shared_examples_for 'nested like express' do
+        context 'when foo' do
+          context 'when bar' do
+            context 'when baz' do
+              context 'when qux' do
+              ^^^^^^^^^^^^^^^^^^ Maximum example group nesting exceeded [4/3].
```
pirj, line 44: `shared_examples_for` is not counted.
If we count it in, even one of our own specs triggers an offence:
```ruby
RSpec.describe RuboCop::Cop::RSpec::PredicateMatcher, :config do
  context 'when enforced style is `explicit`' do
    shared_examples 'explicit common' do
      context 'when custom matchers are allowed' do
      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ Maximum example group nesting exceeded [4/3].
```

pirj:
Ping @bquorning @Darhazer

lib/rubocop/cop/rspec/describe_class.rb
Darhazer: Is `[ ...]` a new syntax for `{ ... }`? Does it work on the minimum supported rubocop version or should we bump the requirements?

lib/rubocop/cop/rspec/describe_method.rb
Darhazer: We still check implicitly if this is a describe, by using the second_argument matcher

lib/rubocop/rspec/top_level_group.rb
```diff
@@ -10,11 +10,21 @@ module TopLevelGroup
       def_node_matcher :example_or_shared_group?,
                        (ExampleGroups::ALL + SharedGroups::ALL).block_pattern
 
-      def on_block(node)
-        return unless respond_to?(:on_top_level_group)
-        return unless top_level_group?(node)
+      def on_new_investigation
+        top_level_groups.each do |node|
+          if respond_to?(:on_top_level_example_group)
+            example_group?(node, &method(:on_top_level_example_group))
```
Darhazer, line 16: and we can have the same for other methods like describe

lib/rubocop/rspec/top_level_group.rb
Darhazer: return from the `on_new_investigation` if there is no ast?

lib/rubocop/cop/rspec/describe_class.rb
pirj: `[ ]` is logical "and", it's around for a while, just not too frequently used.

I find it too hard to reason about boolean login, this node pattern sounds like:
a block
  with a send of `describe` to `RSpec`/nil?
    where the first argument is
      neither a const not a string resembling a const
      *and* the rest of the arguments don't contain a hash with Rails-like metadata

lib/rubocop/rspec/top_level_group.rb
pirj: Yep, if we'll eventually need them.

Initially, I was going to add code that would call `on_top_level_shared_group`, but that would only count one usage, so decided not to add it just now.

lib/rubocop/rspec/top_level_group.rb
pirj: Good idea, thanks!

pirj: Fixed.

pirj:
Simplified TopLevelGroup by introducing dummy methods.
I've removed `TopLevelGroup`'s checks if those methods exist in the including class, since of `TopLevelGroup` is included, that means that one of those methods is overridden. Hence, we can use dummy methods and call them instead of checking if they exist in the consumer.

pirj:
@bquorning @Darhazer ping

lib/rubocop/cop/rspec/describe_class.rb
Darhazer: I'm confused by the `(send #rspec? :describe ` part in the second argument of `[ ]`

pirj: @Darhazer Why?

It's basically the same as previously (which I forgot to remove during merge resolution):
```ruby
        def_node_matcher :describe_with_rails_metadata?, <<-PATTERN
          (send #rspec? :describe !const ...
            (hash <#rails_metadata? ...>)
          )
        PATTERN
```

`!const` check is not necessary anymore, it's checked in the first part of `[ ]`.

bquorning: I too find this node pattern a bit cryptic. Perhaps a bit more duplication can be justified if it makes this code easier to read?

lib/rubocop/rspec/top_level_group.rb
```diff
@@ -10,28 +10,31 @@ module TopLevelGroup
...
+        top_level_groups.each do |node|
+          example_group?(node, &method(:on_top_level_example_group))
+          on_top_level_group(node)
+        end
       end
 
-      def single_top_level_group?
-        top_level_groups.one?
+      def top_level_groups
+        @top_level_groups ||=
```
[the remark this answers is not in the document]
pirj: I was following [this note](https://github.com/rubocop-hq/rubocop-rspec/pull/932#discussion_r439636213):

> Right now a new cop instance is created for every file

There might have been some changes after the doc has been introduced.

As a quick experiment, I've set a breakpoint in random cop's `initialize` and ran `rubocop`:
```
From: /Users/pirj/source/rubocop-rspec/lib/rubocop/cop/rspec/described_class.rb:63 RuboCop::Cop::RSpec::DescribedClass#initialize:

    61: def initialize(*args)
    62:   require 'pry'; binding.pry
 => 63:   super
    64: end

[1] pry(#<RuboCop::Cop::RSpec::DescribedClass>)>
.
From: /Users/pirj/source/rubocop-rspec/lib/rubocop/cop/rspec/described_class.rb:63 RuboCop::Cop::RSpec::DescribedClass#initialize:

    61: def initialize(*args)
    62:   require 'pry'; binding.pry
 => 63:   super
    64: end

[1] pry(#<RuboCop::Cop::RSpec::DescribedClass>)>
.
From: /Users/pirj/source/rubocop-rspec/lib/rubocop/cop/rspec/described_class.rb:63 RuboCop::Cop::RSpec::DescribedClass#initialize:

    61: def initialize(*args)
    62:   require 'pry'; binding.pry
 => 63:   super
    64: end

[1] pry(#<RuboCop::Cop::RSpec::DescribedClass>)>
.
From: /Users/pirj/source/rubocop-rspec/lib/rubocop/cop/rspec/described_class.rb:63 RuboCop::Cop::RSpec::DescribedClass#initialize:

    61: def initialize(*args)
    62:   require 'pry'; binding.pry
 => 63:   super
    64: end

[1] pry(#<RuboCop::Cop::RSpec::DescribedClass>)>
.
```

@marcandre Can you please shed light on this? Isn't this doc misleading?

I used to think it would be quite hard to reuse cop instances since depending on the directory they may receive different config (due to directory-local `.rubocop.yml`).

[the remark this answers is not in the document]
pirj: The [cop mentioned in the referenced discussion, `SurroundingSpace`](https://github.com/rubocop-hq/rubocop/blob/master/lib/rubocop/cop/mixin/surrounding_space.rb#L59) does reset its state:
```ruby
      def on_new_investigation
        @token_table = nil
```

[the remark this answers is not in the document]
pirj: What's even more strange:
```ruby
        def on_new_investigation
          print "N"
          super
        end

        def initialize(*args)
          print "#"
          super
        end
```
outputs:
```
##.#.#.#.#.#.#.#.#.#.#.#.#.#.#.#.#.#.#.#.#.#C#.#.#.#.#.#.#.#.#.#.#.#.#.#.#.#.#.#.#.#.#.#.#.#.#.#.#.#.#.#.#.#.#.#.#.#.#.#.#.#.#.#.#.#.#.#.#.#.#.#.#.#.#.#.#.#.#.#.#.#.#.#.#.#.#.#.#.#.#.#.#.#.#.#.#.#.#.#.#.#.#.#.#.#.#.#.#.#.#.#.#.#N.#N.#N.#N.#N.#N.#N.#N.#N.#N.#N.#N.#N.#N.#N.#N.#N.#N.#N.#N.#N.#N.#N.#N.#N.#N.#N.#N.#N.#N.#N.#N.#N.#N.#N.#N.#N.#N.#N.#N.#N.#N.#N.#N.#N.#N.#N.#N.#N.#N.#N.#N.#N.#N.#N.#N.#N.#N.#N.#N.#N.#N.#N.#N.#N.#N.#N.#N.#N.#N.#N.#N.#N.#N.#N.#N.#N.#N.#N.#N.#N.#N.#N.#N.#N.#N.#N.#N.#N.#N.#N.#N.#N.#N.#N.#N.#N.#N.#N.#N.#.
```

🤔

lib/rubocop/cop/rspec/file_path.rb
```diff
@@ -57,34 +57,42 @@ module RSpec
...
 
-        def on_top_level_describe(node, args)
-          return unless const_described?(node) && single_top_level_describe?
-          return if routing_spec?(args)
+        def on_top_level_group(node)
+          return unless top_level_groups.one?
 
-          glob = glob_for(args)
+          const_described(node) do |send_node, described_class, arguments|
+            next if routing_spec?(arguments)
```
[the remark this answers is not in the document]
pirj: At the point when we spawn `rubocop-rspec-rails`? :D

lib/rubocop/rspec/top_level_group.rb
```diff
@@ -10,28 +10,31 @@ module TopLevelGroup
...
-        return unless respond_to?(:on_top_level_group)
-        return unless top_level_group?(node)
+      def on_new_investigation
+        return unless root_node
 
-        on_top_level_group(node)
+        top_level_groups.each do |node|
+          example_group?(node, &method(:on_top_level_example_group))
+          on_top_level_group(node)
+        end
```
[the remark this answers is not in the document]
pirj: Why?
https://github.com/rubocop-hq/rubocop/blob/6b36c4490c2d7c41349e4f71721bf0cf91920578/lib/rubocop/cop/base.rb#L67:
```ruby
      # Called before all on_... have been called
      # When refining this method, always call `super`
      def on_new_investigation
        # Typically do nothing here
      end
```

lib/rubocop/rspec/top_level_group.rb
```diff
@@ -25,14 +25,16 @@ def top_level_group?(node)
...
         @top_level_groups ||=
-          top_level_nodes.select { |n| example_or_shared_group?(n) }
+          top_level_nodes(root_node).select { |n| example_or_shared_group?(n) }
       end
 
-      def top_level_nodes
-        if root_node.begin_type?
-          root_node.children
+      def top_level_nodes(node)
+        if node.begin_type?
```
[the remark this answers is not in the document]
pirj: I don't really have a strong preference here, but
```ruby
        case node.type
        when :begin
          node.children
        when :module, :class
          top_level_nodes(node.body)
        else
          [node]
        end
```
feels more like poking around in internals.

lib/rubocop/rspec/top_level_group.rb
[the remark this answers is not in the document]
pirj: I've answered my own question 🤦 

> When refining this method, always call `super`

lib/rubocop/cop/rspec/describe_class.rb
pirj: Good point guys. Split this node pattern into two parts. Looks more readable now.

marcandre: FYI the shiny newly released `rubocop-ast` allows comments in patterns

lib/rubocop/rspec/top_level_group.rb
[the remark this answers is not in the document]
marcandre: @pirj Currently, cops being persisted [is opt-in](https://docs.rubocop.org/rubocop/0.88/v1_upgrade_notes.html#cop-persistence). That being said, I'm wondering if a cop that implements `on_new_investigation` shouldn't be persisted by default.
BTW, different config => new cop, but that's a very rare setup. 
It looks like the cop is initialized a bunch of times before doing the actual run, that seems like a bug somewhere, I'll check it out.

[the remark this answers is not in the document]
pirj: I'm looking at default cop's initializer and struggling to understand if there are any performance/memory usage improvements of reusing cop instances:
```ruby
      def initialize(config = nil, options = nil)
        @config = config || Config.new
        @options = options || { debug: false }
        reset_investigation
      end

      def reset_investigation
        @currently_disabled_lines = @current_offenses = @processed_source = @current_corrector = nil
      end
```

It would make sense if the expensive `Config.new` would be called on each extra cop initialization, but it isn't:
```ruby
      # Called between investigations
      def ready
        return self if self.class.support_multiple_source?

        self.class.new(@config, @options) # <== `@config` is reused
      end
```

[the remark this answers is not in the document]
pirj: Please don't get me wrong, I don't really mind (as a performance-last person) to remove memoization here. We're only calling this `top_level_groups` method once per top-level example group. And those who have more than one, probably deserve to slightly suffer anyway.

I'm more concerned about the confusion that this reuse might deal if once this option will become opt-out rather than opt-in. Cop specs won't be bringing memoization to the light, will they? With all numerous well-unknown RuboCop extensions this can have quite an impact.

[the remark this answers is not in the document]
marcandre: Nothing major indeed. I wanted to speedup the callback cache building in `Commissionner`, but it could be done another way (per class instead of per instance). Some cops *want to* persist, though (there was an issue on that), which is why it's possible to do so.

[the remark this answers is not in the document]
pirj: Also, speaking of per-directory configs. In theory, it's allowed, and I don't see a reason to disrespect this option.

In the practice of working with majestic monoliths composed of Rails engines, in-repo gems and libraries in git submodules it's totally not uncommon to have different styles and different configuration, at least in some directories.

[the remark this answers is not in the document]
pirj: > Some cops want to persist

Got it 👍 That's reasonable, I can imagine some use cases for that.

However, does it make sense to keep this notice for the rest of the cops that don't want and won't persist by default? https://github.com/rubocop-hq/rubocop/blob/6b36c4490c2d7c41349e4f71721bf0cf91920578/docs/modules/ROOT/pages/v1_upgrade_notes.adoc#if-your-class-uses-instance-variables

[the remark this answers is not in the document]
marcandre: > Cop specs won't be bringing memoization to the light, will they? With all numerous well-unknown RuboCop extensions this can have quite an impact.

Yes, that was my thinking too.

> I don't see a reason to disrespect this option.

Absolutely :-) 

> However, does it make sense to keep this notice for the rest of the cops

You mean the recommendation to clear the data with `on_new_investigation`? I guess it's not necessary. I should also be clearer reword the "will be called on different if" to "will not be called unless"...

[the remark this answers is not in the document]
pirj: I would remove the whole "If your class uses instance variables" section, as it's more confusing than helpful considering the current "one source" - "one cop instance" behaviour.

Even for those cops that need to be persisted, I can think of ~global variables~ class variables to persist the data between the runs.

By the way, there's one more thing that can impact those persisted cops - `--parallel`. Do we parallelize by cops or by source?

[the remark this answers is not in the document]
marcandre: => https://github.com/rubocop-hq/rubocop/pull/8440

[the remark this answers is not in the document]
marcandre: > Even for those cops that need to be persisted, I can think of class variables to persist the data between the runs.

Not quite... See https://github.com/rubocop-hq/rubocop/issues/7968 if interested

> By the way, there's one more thing that can impact those persisted cops - --parallel. Do we parallelize by cops or by source?

I imagine it's by source, but we could segregate those cops that need it; I suspect they simply run them without `--parallel` though.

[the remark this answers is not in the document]
pirj: @bquorning As a bottom-line: I wouldn't worry about `@top_level_groups ||=` used here unless any of the cops would decide to opt-in for reusage *and* use `TopLevelGroup` which I highly doubt.

pirj:
Loud thinking

```ruby
describe Weather do
  let(:temp) { 20.celsius }
  context "when it's warm" do
    let(:temp) { 30.celsius } # <== should it count as an additional `let`, or should it collapse with an existing one?
```

marcandre:
> Loud thinking
> 
> ```ruby
> describe Weather do
>   let(:temp) { 20.celsius }
>   context "when it's warm" do
>     let(:temp) { 30.celsius } # <== should it count as an additional `let`, or should it collapse with an existing one?
> ```

I assumed it was one. It makes even less sense to me to count the repeated `let`.

PR #980, opened 2020-07-24
Extract constants in node patterns to node patterns

pirj:
* [x] works on our lower-bound RuboCop 0.87.0

---

* [-] Added tests.
* [-] Updated documentation.

### DIFF
--- a/lib/rubocop/rspec/language/node_pattern.rb
+++ b/lib/rubocop/rspec/language/node_pattern.rb
@@ -7,6 +7,8 @@ module RuboCop
       module NodePattern
         extend RuboCop::NodePattern::Macros
 
+        def_node_matcher :rspec?, '{(const {nil? cbase} :RSpec) nil?}'
+
         def_node_matcher :example_group?, ExampleGroups::ALL.block_pattern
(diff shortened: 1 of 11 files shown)


lib/rubocop/rspec/language/node_pattern.rb
```diff
@@ -7,6 +7,8 @@ module Language
       module NodePattern
         extend RuboCop::NodePattern::Macros
 
+        def_node_matcher :rspec?, '{(const {nil? cbase} :RSpec) nil?}'
```
pirj, line 10: For some reason, I had an idea that this is impossible.

But is it good?
Using `#{RSPEC}` in several node patterns mean we'll compiling this part into Ruby code each time. For `#rspec`, we're compiling once and reusing the already compiled code across other node matchers that use it.

pirj:
Ping @bquorning @Darhazer

lib/rubocop/rspec/language/node_pattern.rb
Darhazer: > For some reason, I had an idea that this is impossible.

Neither did I :joy:

pirj:
Ping @bquorning

PR #982, opened 2020-07-31
Fix Metrics/ClassLength violation

bquorning:
Our TODO file mentioned that the `DescribedClass` class was too long. Looking closer at it, I found a couple of methods (`#full_const_name`, `#collapse_namespace`, `#const_name`, and `#namespace`) working on the isolated problem of determining the “full constant name”, of a const node. This PR extracts those methods into a separate `FullConstNameFinder` module.

The code that I move was introduced in one single commit, bdc8723c65a0521e6df3e0b481592d19fdc000ca, and I wonder if @pirj wouldn’t like to help me with this here PR? 😄 

### TODO

- [ ] Code documentation.
- [ ] Add tests specifically for the extracted module.
- [ ] Consider extracting the code into a separate file so it can be re-used elsewhere.

---

* [ ] Added tests.
* [x] Updated documentation.


lib/rubocop/cop/rspec/described_class.rb
```diff
@@ -156,64 +156,69 @@ def offensive_described_class?(node)
...
-        #   collapse_namespace([:A, :B], [nil, :B, :C) # => [nil, :B, :C]
-        def collapse_namespace(namespace, const)
-          return const if namespace.empty?
-          return const if const.first.nil?
-
-          start = [0, (namespace.length - const.length)].max
-          max = namespace.length
-          intersection = (start..max).find do |shift|
-            namespace[shift, max - shift] == const[0, max - shift]
+          def call(node)
```
pirj, line 166: If not the docs, I would struggle to understand what those methods do.
Do you mind to add docs for `call` as well?

lib/rubocop/cop/rspec/described_class.rb
```diff
@@ -156,64 +156,69 @@ def offensive_described_class?(node)
 
           return if nearest_described_class.equal?(node)
 
-          full_const_name(nearest_described_class) == full_const_name(node)
+          FullConstNameFinder.call(nearest_described_class) ==
+            FullConstNameFinder.call(node)
```
pirj, line 160: This is not very self-explanatory when I read it aloud.

lib/rubocop/cop/rspec/described_class.rb
bquorning: Done

lib/rubocop/cop/rspec/described_class.rb
bquorning: I agree. Please review again.

pirj:
> [ ] Code documentation

Looks pretty good already.

> [ ] Add tests specifically for the extracted module

We don't have much tests for modules, we rely on cops' tests to cover module's behavior, so we're good here from my point of view.

> [ ] Consider extracting the code into a separate file so it can be re-used elsewhere

Did you have some other cops in mind that could use this? I can't think of any.

lib/rubocop/cop/rspec/described_class.rb
```diff
@@ -156,64 +156,83 @@ def offensive_described_class?(node)
...
+            elsif namespace.lvar_type? || namespace.cbase_type?
+              [nil, name]
+            end
+          end
+
+          # @param node [RuboCop::AST::Node]
+          # @return [Array<Symbol>]
+          # @example
+          #   namespace(node) # => [:A, :B, :C]
+          def namespace(node)
```
Darhazer, line 229: I felt that such a code should belong to rubocop-ast. So I checked and there is indeed `parent_module_name`, which seems to be used only by the `DuplicateMethods` cop and seems to be buggy (or not quite doing the same on purpose). Maybe it would be good to fix it, or at least open a ticket?

pirj: Filed https://github.com/rubocop-hq/rubocop-rspec/issues/984

Can't agree more that this namespace juggling doesn't really belong here.

@bquorning It seems that with this find the same result of reducing the class length can be achieved by using `parent_module_name`.

Darhazer: Well, I actually tried that, with a quick fix in `parent_module_name` itself, but it fails on `::MyClass` - still resolving it under `MyNamespace` instead of the global namespace

Issue #908, opened 2020-05-01
Cop Idea: Ensure let/subject are used with symbols not strings


tejasbubane:
And provide an option to use strings.

```ruby
# bad
let('first_name') { 'Adam' }
subject('user') { create_user }

#good
let(:first_name) { 'Adam' }
subject(:user) { create_user } 
```

This was brought up by @Darhazer in [this comment](https://github.com/rubocop-hq/rubocop-rspec/pull/907#discussion_r418586020)

pirj:
I can imagine some cases to cover that are accepted by RSpec:
```
let("item#{num}") { } # bad practice? 
let("multi" \
    "line") { } # interesting to see how this will be addressed later on
```

PR #986, opened 2020-08-03
Allow dstr even when `symbol` style is configured

bquorning:
Relax the `RSpec/VariableDefinition` cop so interpolated and multiline strings are accepted even when configured to enforce the `symbol` style.

It seems weird that we enforce writing e.g. `let(:"foo#{bar}")` instead of `let("foo#{bar}")`. What I think is important to enforce is that people don't mix e.g. `let(:foo)` and `let('bar')`.

See https://github.com/rubocop-hq/rubocop-rspec/pull/910#discussion_r464230835 and #908.

/cc @tejasbubane 

---

* [x] Added tests.
* [x] Updated documentation.

### DIFF
--- a/spec/rubocop/cop/rspec/variable_definition_spec.rb
+++ b/spec/rubocop/cop/rspec/variable_definition_spec.rb
@@ -11,17 +11,15 @@ RSpec.describe RuboCop::Cop::RSpec::VariableDefinition, :config do
       RUBY
     end
 
-    it 'registers an offense for interpolated string' do
-      expect_offense(<<~'RUBY')
+    it 'does not register offense for interpolated string' do
+      expect_no_offenses(<<~'RUBY')
         let("user-#{id}") { 'Adam' }
-            ^^^^^^^^^^^^ Use symbols for variable names.
       RUBY
     end
 
-    it 'registers an offense for multiline string' do
-      expect_offense(<<~'RUBY')
-        let("user"\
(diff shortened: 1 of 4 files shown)


spec/rubocop/cop/rspec/variable_definition_spec.rb
```diff
@@ -11,17 +11,15 @@
...
     end
 
-    it 'registers an offense for multiline string' do
-      expect_offense(<<~'RUBY')
-        let("user"\
-            ^^^^^^^ Use symbols for variable names.
+    it 'does not register offense for multiline string' do
+      expect_no_offenses(<<~'RUBY')
+        let("user" \
             "-foo") { 'Adam' }
```
pirj, line 23: Doesn't fall into an interpolated string category, but I understand that they are both `dstr_type?`, pretty similar AST-wise and innecesarily hard to distinguish:
```
$ ruby-parse -e '"a" \
quote> "b"'
(dstr
  (str "a")
  (str "b"))
```
```
$ ruby-parse -e '"a#{"b"}"'
(dstr
  (str "a")
  (begin
    (str "b")))
```

bquorning: And we shouldn’t recommend converting a multi-line string into a multi-line symbol. Right?

pirj: Indeed!

PR #987, opened 2020-08-03
[Fix #941] Add support for ignoring certain variables in RSpec/VariableName

### LINKED ISSUE #941: RSpec/VariableName: Would appreciate an 'ignore' configuration option
**Feature Request**
A configuration option to specify certain variables or patterns to ignore

**Example**
```yaml
RSpec/VariableName:
  Enabled: true
  EnforcedStyle: "snake_case"
  IgnoreVariables:
    - fooBar
    - /fizzBuzz/
```
**Context**
We use an RSpec plugin called [RSwag](https://github.com/rswag/rswag) that allows the creation of specs & swagger documentation from a single DSL. One of the features is that documented parameters can be populated in tests via `let` statements

eg.
```ruby
describe 'Blogs API - Individual Blog Endpoint' do
  path '/blogs/{id}' do
	parameter(name: :id, in: :path, required: true, type: :string)
    let(:id) { 'foo' }
	# test code ensues
  end
end
```

The issue arises when providing an array as a query parameter
```ruby
describe 'Blogs API - Blog Collection Endpoint' do
  path '/blogs' do
    # making a request to /blogs?id[]=foo&id[]=bar&id[]=baz
    parameter(
      name: :'id[]',
      in: :query,
      required: false,
      type: :array,
      items: { type: :string },
      collectionFormat: :multi,
    )
    let(:'id[]') { ['foo', 'bar', 'baz'] } # sets off the cop
    # test code ensues
  end
end 
```

It's way too much of a corner case to modify the cop to ignore that specific pattern, but I think a generic option to ignore certain things by name or regex is appropriate.

jtannas:
Fixes #941

This PR adds support for ignoring naming conventions on certain RSpec variables via configuration options. Options for `IgnoredVariables` and `IgnoredPatterns` are provided, and are empty by default. This will be helpful for strange cases and for codebases that are iteratively refactoring to apply the rule.

My original thought was to have them combined in a single `Ignore` option which would accept both static values and regex patterns, but the extra complexity of parsing the configuration led me to split it in two.

Edit: `IgnoredVariables` has been removed - see comments below

---

* [x] Added tests.
* [x] Updated documentation.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 * Fix `RSpec/EmptyExampleGroup` to ignore examples groups with examples defined inside iterators. ([@pirj][])
 * Improve `RSpec/NestedGroups`, `RSpec/FilePath`, `RSpec/DescribeMethod`, `RSpec/MultipleDescribes`, `RSpec/DescribeClass`'s top-level example group detection. ([@pirj][])
 * Add detection of `let!` with a block-pass or a string literal to `RSpec/LetSetup`. ([@pirj][])
+* Add `IgnoredPatterns` configuration option to `RSpec/VariableName`. ([@jtannas][])
 
 ## 1.42.0 (2020-07-09)
 
@@ -540,3 +541,4 @@ Compatibility release so users can upgrade RuboCop to 0.51.0. No new features.
 [@harrylewis]: https://github.com/harrylewis
 [@elliterate]: https://github.com/elliterate
 [@mlarraz]: https://github.com/mlarraz
+[@jtannas]: https://github.com/jtannas
--- a/lib/rubocop/cop/rspec/variable_name.rb
+++ b/lib/rubocop/cop/rspec/variable_name.rb
@@ -5,25 +5,43 @@ module RuboCop
     module RSpec
       # Checks that memoized helper names use the configured style.
       #
+      # Variables can be excluded from checking using the `IgnoredPatterns`
+      # option.
+      #
       # @example EnforcedStyle: snake_case (default)
       #   # bad
-      #   let(:userName) { 'Adam' }
-      #   subject(:userName) { 'Adam' }
+      #   subject(:userName1) { 'Adam' }
+      #   let(:userName2) { 'Adam' }
       #
       #   # good
-      #   let(:user_name) { 'Adam' }
-      #   subject(:user_name) { 'Adam' }
+      #   subject(:user_name_1) { 'Adam' }
+      #   let(:user_name_2) { 'Adam' }
       #
       # @example EnforcedStyle: camelCase
       #   # bad
-      #   let(:user_name) { 'Adam' }
-      #   subject(:user_name) { 'Adam' }
+      #   subject(:user_name_1) { 'Adam' }
+      #   let(:user_name_2) { 'Adam' }
       #
       #   # good
-      #   let(:userName) { 'Adam' }
-      #   subject(:userName) { 'Adam' }
+      #   subject(:userName1) { 'Adam' }
+      #   let(:userName2) { 'Adam' }
+      #
+      # @example IgnoredPatterns configuration
+      #
+      #   # rubocop.yml
+      #   # RSpec/VariableName:
+      #   #   EnforcedStyle: snake_case
+      #   #   IgnoredPatterns:
+      #   #     - ^userFood
+      #
+      # @example
+      #   # okay because it matches the `^userFood` regex in `IgnoredPatterns`
+      #   subject(:userFood_1) { 'spaghetti' }
+      #   let(:userFood_2) { 'fettuccine' }
+      #
       class VariableName < Base
         include ConfigurableNaming
+        include IgnoredPattern
         include RuboCop::RSpec::Variable
 
         MSG = 'Use %<style>s for variable names.'
@@ -31,6 +49,7 @@ module RuboCop
         def on_send(node)
           variable_definition?(node) do |variable|
             return if variable.dstr_type? || variable.dsym_type?
+            return if matches_ignored_pattern?(variable.value)
 
             check_name(node, variable.value, variable.loc.expression)
           end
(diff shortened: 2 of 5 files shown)


lib/rubocop/cop/rspec/variable_name.rb
```diff
@@ -14,6 +17,14 @@ module RSpec
       #   let(:user_name) { 'Adam' }
       #   subject(:user_name) { 'Adam' }
       #
+      #   # okay with `userFood` listed in `IgnoredVariables`
```
pirj, line 20: Typically we describe it in a separate `@example`:
```ruby
      # @example `Prefixes` configuration
      #
      #   # .rubocop.yml
      #   # RSpec/ContextWording:
      #   #   Prefixes:
      #   #     - when
      #   #     - with
      #   #     - without
      #   #     - if
      #   #     - unless
      #   #     - for
      #
      # @example
      #   # bad
      #   context 'the display name not present' do
      #     # ...
      #   end
      #
      #   # good
      #   context 'when the display name is not present' do
      #     # ...
      #   end
```

do you mind adding two examples to demonstrate how we expect the configuration of those options?
Otherwise, it may create a false impression that `IgnoredVariables`/`IgnoredPatterns` only relate to `EnforcedStyle: snake_case`.

lib/rubocop/cop/rspec/variable_name.rb
```diff
@@ -24,13 +35,16 @@ module RSpec
       #   subject(:userName) { 'Adam' }
       class VariableName < Base
         include ConfigurableNaming
+        include IgnoredPattern
```
pirj, line 38: 👍

lib/rubocop/cop/rspec/variable_name.rb
```diff
@@ -41,6 +55,10 @@ def on_send(node)
         def message(style)
           format(MSG, style: style)
         end
+
+        def matches_ignored_variable?(value)
```
pirj, line 59: Maybe `value` -> `variable_name`?

spec/rubocop/cop/rspec/variable_name_spec.rb
```diff
@@ -68,6 +72,18 @@
           let("user#{name}") { 'Adam' }
         RUBY
       end
+
+      it 'does not register offense for configured ignoredVariable' do
+        expect_no_offenses(<<~'RUBY')
+          let('ignoredVariable') { 'Adam' }
+        RUBY
+      end
```
pirj, line 80: Please add those examples separate from `context 'when configured for `snake_case`' do`, along with config setup, and preferably each having a balancing example where the same variable name is flagged.

pirj:
Do you think we really need `IgnoredVariables` at all given that it's possible to add `^var_ignored$` to `IgnoredPatterns`?
I have no preference here, but I can't figure out a good use case for it. What did you have in mind, where would it come handy in your practice?

CHANGELOG.md
```diff
@@ -8,6 +8,7 @@
 * Relax `RSpec/VariableDefinition` cop so interpolated and multiline strings are accepted even when configured to enforce the `symbol` style. ([@bquorning][])
 * Fix `RSpec/EmptyExampleGroup` to flag example groups with examples in invalid scopes. ([@mlarraz][])
 * Fix `RSpec/EmptyExampleGroup` to ignore examples groups with examples defined inside iterators. ([@pirj][])
+* Add `IgnoreVariables` and `IgnorePatterns` configuration options to`RSpec/VariableName. ([@jtannas][])
```
pirj, line 11: "to`RSpec/VariableName" - backtick is missing, space is missing.

jtannas:
> Do you think we really need `IgnoredVariables` at all given that it's possible to add `^var_ignored$` to `IgnoredPatterns`?
> I have no preference here, but I can't figure out a good use case for it. What did you have in mind, where would it come handy in your practice?

I suppose we don't need both. My original thought was that the `IgnoredVariables` is more user-friendly for simple cases than using regex patterns. In hindsight though, having redundant configuration options is probably even less user friendly.
I'll drop the `IgnoredVariables` since `IgnoredPatterns` is more flexible/powerful/already written.

spec/rubocop/cop/rspec/variable_name_spec.rb
```diff
@@ -188,4 +188,33 @@
...
   end
+
+  context 'when configured to ignore certain patterns' do
+    context 'when the variable does not match any of the patterns' do
+      let(:cop_config) do
+        { 'EnforcedStyle' => 'snake_case',
+          'IgnoredPatterns' => ['^userFood$', '^userPet$'] }
+      end
+
+      it 'registers an offense' do
```
bquorning, line 199: Sorry for nitpicking, but I’ll suggest removing the innermost `context`s, and use the same `cop_config` for these two examples. E.g.:
```rb
context 'when configured to ignore certain patterhs' do
  let(:cop_config) do
    # …
  end

  it 'registers an offense when not matching any ignored pattern' do
    # …
  end

  it 'registers an offense when matching at least one ignored pattern' do
    # …
  end
end
```

What do you think?

lib/rubocop/cop/rspec/variable_name.rb
```diff
@@ -22,15 +25,31 @@ module RSpec
...
+      #   # rubocop.yml
+      #   # RSpec/VariableName:
+      #   #   EnforcedStyle: snake_case
+      #   #   IgnoredPatterns:
+      #   #     - ^userFood$
+      #
+      # @example
+      #   # okay because it matches the`^userFood$` regex in `IgnoredPatterns`
+      #   let(:userFood) { 'spaghetti' }
+      #   subject(:userFood) { 'spaghetti' }
```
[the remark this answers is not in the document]
jtannas: sounds good - I was using the pattern from the existing examples, so I'll go ahead and make the same fixes to them while I'm in there

bquorning:
Could you squash the 3 commits into one, please? Then we’re ready to merge 👍

Issue #972, opened 2020-07-17
Documentation


bquorning:
I noticed that @bbatsov added three commits, adding some AsciiDoc documentation:

- 0d5031744078ab0787e0f92ff18828743daf6b69
- b15eefbf3d2e7261c38d30637e0eec36620bb055
- 97e8ebbac60845f83eab72d2cf2851a5c629fbb7

I cannot find any related PRs, so probably they were pushed directly to master (please don’t do that again 😄).

It seems we now have almost identical documents in `manual/` and `docs/` folders, but only the former is getting updated. What’s the plan? Should we change our documentation Rake task to update `docs/` and delete the `manual/` folder?

bquorning:
I am asking party because I want to turn [cops_documentation.rake](https://github.com/rubocop-hq/rubocop-rspec/blob/0cad53a9ba5a96161be78a315262d9a8da959782/tasks/cops_documentation.rake#L13) into a class instead having all the code inside a Rake task, and that _almost_ identical code exists in [rubocop](https://github.com/rubocop-hq/rubocop/blob/2d3f95c55d05faab3ad659ff9a56f408e53a2bf3/tasks/cops_documentation.rake#L12), [rubocop-rails](https://github.com/rubocop-hq/rubocop-rails/blob/2f09406560fbece1adced14edd68f7d8f9e3254e/tasks/cops_documentation.rake#L13), and [rubocopo-performance](https://github.com/rubocop-hq/rubocop-performance/blob/8e6fc14ad8825bbc3a27620b9f1fb68524e542a9/tasks/cops_documentation.rake#L13), and I imagine rubocop-rake has, or would benefit from having, similar code.

pirj:
That would be great to unify doc tasks with other repos.
Since it's `docs/` that get deployed to https://docs.rubocop.org/rubocop-rspec/, I guess we can retire `manual/`.

We still have a link in gemspec that points to https://rubocop.readthedocs.io/projects/rspec/en/latest/ though.
Frankly, I have little to no idea how this documentation gets its way there. Somehow via a hook that uses `mkdocs.yml`?

https://www.rubydoc.info/find/github?q=rubocop-rspec returns 404.
I remember `@`lsegal and `@`ioquatix maintain some other doc site, but can't quickly find it.
Anyway, it seems rather unrelated, as such sites parse the code using class docs.

bquorning:
[...]

PR #989, opened 2020-08-06
Use RuboCop's code for generating documentation

bquorning:
Rubocop now has a `CopsDocumentationGenerator` class that can generate consistent documentation across all projects.

With this PR, we will be updating the `docs` folder and no longer updating the `manual` folder. Should we delete it? Which documentation is currently published, and will it work after merging this PR?

---

* [-] Added tests.
* [x] Updated documentation.


bquorning:
As per #972, I think we should now be publishing to https://docs.rubocop.org/rubocop-rspec/ instead of https://rubocop.readthedocs.io/projects/rspec/en/latest/

Darhazer:
> As per #972, I think we should now be publishing to https://docs.rubocop.org/rubocop-rspec/ instead of https://rubocop.readthedocs.io/projects/rspec/en/latest/

Have no idea what to configure there though

rubocop-rspec.gemspec
```diff
@@ -33,7 +33,7 @@ Gem::Specification.new do |spec|
 
   spec.metadata = {
     'changelog_uri' => 'https://github.com/rubocop-hq/rubocop-rspec/blob/master/CHANGELOG.md',
-    'documentation_uri' => 'https://rubocop-rspec.readthedocs.io/'
+    'documentation_uri' => 'https://docs.rubocop.org/rubocop-rspec/'
   }
 
   spec.add_runtime_dependency 'rubocop', '~> 0.87'
```
pirj, line 39: To generate docs, do we depend on 0.89 In development?

Rakefile
```diff
@@ -53,11 +53,11 @@ end
 desc 'Confirm documentation is up to date'
 task confirm_documentation: :generate_cops_documentation do
   _, _, _, process =
-    Open3.popen3('git diff --exit-code manual/')
+    Open3.popen3('git diff --exit-code docs/')
 
   unless process.value.success?
     raise 'Please run `rake generate_cops_documentation` ' \
-          'and add manual/ to the commit.'
+          'and add docs/ to the commit.'
```
pirj, line 60: Wondering if this can/should be rephrased to "stage your doc/ changes".

lib/rubocop/cop/rspec/base.rb
```diff
@@ -29,6 +29,8 @@ class Base < ::RuboCop::Cop::Base
                                .map(&Regexp.public_method(:new))
         )
 
+        exclude_from_registry
```
pirj, line 32: 👍

rubocop-rspec.gemspec
bquorning: Yes

Rakefile
bquorning: I am concerned whether everybody would understand the verb “stage”. And after staging the changes, they _do_ need to also be committed :-)

bbatsov:
> With this PR, we will be updating the docs folder and no longer updating the manual folder. Should we delete it? Which documentation is currently published, and will it work after merging this PR?

Yeah, you can delete the manual folder. Afterwards you can setup redirects in the RTD admin console if you want. 

> Which documentation is currently published, and will it work after merging this PR?

Nothing will happen automatically after merging this PR, as documentation deployments for the new system are currently triggered manually (typically after we cut some new release). I saw that @koic already explained the process in another ticket, and it's already in the docs somewhere - https://docs.rubocop.org/rubocop/0.89/contributing.html#working-on-the-docs Probably we can add there some more instructions down the road.

bquorning:
Force-pushed with updated CHANGELOG.

PR #990, opened 2020-08-06
Add detection of `let!` with a block-pass or a string literal to `RSpec/LetSetup`

pirj:
Found when addressing [this comment](https://github.com/rubocop-hq/rubocop-rspec/pull/863#discussion_r466290915).

* [x] Added tests.
* [-] Updated documentation.

### DIFF
--- a/lib/rubocop/cop/rspec/let_setup.rb
+++ b/lib/rubocop/cop/rspec/let_setup.rb
@@ -35,7 +35,10 @@ module RuboCop
                          ).block_pattern
 
         def_node_matcher :let_bang, <<-PATTERN
-          (block $(send nil? :let! {(sym $_) (str $_)}) ...)
+          {
+            (block $(send nil? :let! {(sym $_) (str $_)}) ...)
+            $(send nil? :let! {(sym $_) (str $_)} block_pass)
+          }
         PATTERN
 
         def_node_search :method_called?, '(send nil? %)'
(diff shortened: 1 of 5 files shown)


spec/smoke_tests/weird_rspec_spec.rb
```diff
@@ -33,6 +33,17 @@
...
+  let('foo') { 1 }
+  let('foo' \
+      'bar') { 1 }
+
+  let!('foo') { 1 }
+  let!('foo' \
+      'bar') { 1 }
+
+  let("foo#{1}") { 1 }
+  let!("foo#{1}") { 1 }
```
Darhazer, line 45: maybe add some blockpasses as well

lib/rubocop/cop/rspec/let_setup.rb
```diff
@@ -35,7 +35,10 @@ class LetSetup < Base
                          ).block_pattern
 
         def_node_matcher :let_bang, <<-PATTERN
-          (block $(send nil? :let! (sym $_)) args ...)
+          {
+            (block $(send nil? :let! {(sym $_) (str $_)}) ...)
+            $(send nil? :let! {(sym $_) (str $_)} block_pass)
+          }
```
pirj, line 41: It might be possible to get rid of this complexity somehow by extracting a part of this. But refactoring can wait while we have more cases with block-pass vs regular.

spec/smoke_tests/weird_rspec_spec.rb
pirj: 👍 Decided to add blockpass here in https://github.com/rubocop-hq/rubocop-rspec/pull/863, which errors out as you said with blockpass, but `LetSetup` doesn't.

Issue #992, opened 2020-08-08
Cop Idea: Detect repeated include_examples in same context


biinari:
Similar to #260, we should detect identical includes of [shared examples](https://relishapp.com/rspec/rspec-core/v/3-9/docs/example-groups/shared-examples) within an example group.

This should be a violation (spotted in [spec for rubocop Style/ConditionalAssignment](https://github.com/rubocop-hq/rubocop/blob/1fc19811eef1f70a103cddee4b34e4e8a863b8c7/spec/rubocop/cop/style/conditional_assignment_assign_to_condition_spec.rb#L724-L725)):

```ruby
RSpec.describe RuboCop::Cop::Style::ConditionalAssignment do
  it_behaves_like('all assignment types', '=')
  it_behaves_like('all assignment types', '+=')
  it_behaves_like('all assignment types', '-=')

  it_behaves_like('all assignment types', '=') # violation: repeated
  it_should_behave_like('all assignment types', '+=') # violation: repeated with different syntax

  include_examples('all assignment types', '-=') # violation: similar but in current context instead of nested context
end
```

`it_behaves_like` and `it_should_behave_like` are identical (both include the examples in a nested context) so should be treated the same.

Although `include_examples` includes the examples in the current context instead of a nested context, I can't think of any cases where one would want to use different forms for the same `shared_examples`.

---

The following should not be violations:

```ruby
context 'with foo' do
  it_behaves_like 'some examples', 'test'

  it_behaves_like 'some examples', 'other' # ok: different parameters
  it_behaves_like 'other examples', 'test' # ok: different shared_examples
end

context 'with bar' do
  it_behaves_like 'some examples', 'test' # ok: in a different context
end
```

pirj:
Nice!
Would you like to hack on it?

biinari:
@pirj yeah, I fancy having a go at it.

PR #993, opened 2020-08-09
Support expect_offense template arguments

biinari:
Pass keyword arguments through to `RuboCop::RSpec::ExpectOffense#expect_offense`. These are used to set template replacements like `%{name}` added in [rubocop #8069](https://github.com/rubocop-hq/rubocop/issues/8069).

Allow code offense templates in specs to use `%{name}`. [`Style/FormatStringToken`](https://docs.rubocop.org/rubocop/cops_style.html#styleformatstringtoken) prefers `%<name>s` to `%{name}` but `expect_offense` does not use `Kernel#format` or `String#%`.

Example usage (from work in progress for #992):

```ruby
[include_examples it_behaves_like it_should_behave_like].each do |include_method|
  it "registers an offense with #{include_method}" do
    expect_offense(<<~RUBY, include_method: include_method)
      describe 'foo' do
        %{include_method} 'an x'
        ^{include_method}^^^^^^^ Repeated include of shared_examples 'an x' on line(s) [3]
        %{include_method} 'an x'
        ^{include_method}^^^^^^^ Repeated include of shared_examples 'an x' on line(s) [2]
      end
    RUBY
  end
end
```

---

* [ ] Added tests.
* [ ] Updated documentation.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -54,3 +54,7 @@ RSpec/ExampleLength:
 RSpec/DescribeClass:
   Exclude:
     - spec/project/**/*.rb
+
+Style/FormatStringToken:
+  Exclude:
+    - spec/rubocop/**/*.rb
--- a/spec/support/expect_offense.rb
+++ b/spec/support/expect_offense.rb
@@ -9,7 +9,7 @@ module ExpectOffense
 
   DEFAULT_FILENAME = 'example_spec.rb'
 
-  def expect_offense(source, filename = DEFAULT_FILENAME)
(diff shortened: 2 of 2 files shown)


pirj:
Thanks!

Can you please add an example usage of this feature?

I'm not sure I completely understand why `FormatStringToken` should be disabled for cops.

The feature was released as part of RuboCop 0.85, and we depend on 0.87+, so 👍

biinari:
No problem @pirj. I've added an example and a bit more explanation for the disabling of the `Style/FormatStringToken` cop. I'm making use of this in the specs for #992, which I'll finish off later.

PR #994, opened 2020-08-10
Add cop RSpec/RepeatedIncludeExample

### LINKED ISSUE #992: Cop Idea: Detect repeated include_examples in same context

biinari:
[Fixes #992]

Detect repeated include of `shared_examples` in same context.

```ruby
# bad
describe 'foo' do
  it_behaves_like 'something'
  it_behaves_like 'something' # bad: repeat
end

# bad
describe 'foo' do
  include_examples('a thing', 'x')
  include_examples('a thing', 'x') # bad: same parameters
end

# good
describe 'foo' do
  it_behaves_like 'something'
  it_behaves_like 'other' # ok: different shared_examples
end

# good
context 'foo' do
  include_examples('a thing', 'x')
  include_examples('a thing', 'y') # ok: different parameters
end

# good
describe 'foo' do
  it_behaves_like 'something'
end
describe 'bar' do
  it_behaves_like 'something' # ok: in a different example group
end

# NOT DETECTED - left for a future improvement
it_behaves_like 'a foo' do
  let(:x) { 1 }
end
it_behaves_like 'a foo' do
  let(:x) { 1 }
end
```
---

* [x] Added tests.
* [x] Updated documentation.


biinari:
I'm not sure if allowing re-assigning variables to avoid repeats is reasonable. Perhaps it should suggest a separate context is used in that case instead.

lib/rubocop/cop/rspec/repeated_include_example.rb
```diff
@@ -0,0 +1,140 @@
...
+      #    # good
+      #    context 'foo' do
+      #      cool = 'neat'
+      #      include_examples 'something', cool
+      #      cool = 'fun'
+      #      include_examples 'something', cool
+      #    end
+      #
+      class RepeatedIncludeExample < Base
+        MSG = 'Repeated include of shared_examples %<name>s on line(s) %<loc>s'
```
pirj, line 58: Maybe `loc` -> `repeats`?

lib/rubocop/cop/rspec/repeated_include_example.rb
```diff
@@ -0,0 +1,140 @@
...
+
+        def_node_matcher :several_include_examples?, <<-PATTERN
+          (begin <#include_examples? #include_examples? ...>)
```
pirj, line 61: Will it match if there's something in between, e.g.
```ruby
context 'foo' do
  it_should_behave_like 'a duck'
  it {}
  it_should_behave_like 'a duck'
end
```

lib/rubocop/cop/rspec/repeated_include_example.rb
```diff
@@ -0,0 +1,140 @@
...
+        PATTERN
+
+        def_node_matcher :include_examples?,
+                         Includes::EXAMPLES.send_or_block_or_block_pass_pattern
```
pirj, line 65: Speaking of blocks and block-passes, what is the behavior for:
```ruby
it_behaves_like 'foo' do
  let(:bar) { 1 }
end
it_behaves_like 'foo' do
  let(:bar) { 2 }
end
```
and
```ruby
it_behaves_like 'foo', &bar
it_behaves_like 'foo', &bar
# those two seem like duplicates, but imagine it has side effects
bar = -> { $incr += 1 }
```

lib/rubocop/cop/rspec/repeated_include_example.rb
biinari: Yes it will, I'll make sure there's a test case to indicate that.

biinari:
Good thinking. I'll restrict this to detecting with literal arguments only.

lib/rubocop/cop/rspec/repeated_include_example.rb
biinari: Will just go for literals / constants now, so a `block_pass` will be allowed without offense.

spec/rubocop/cop/rspec/repeated_include_example_spec.rb
```diff
@@ -0,0 +1,318 @@
...
+        it "allows #{include_method} with different interpolated strings" do
+          expect_no_offenses(<<~RUBY)
+            describe 'foo' do
+              duck = 'quack'
+              goose = 'honk'
+              #{include_method} "quacks like a \#{duck}"
+              #{include_method} "quacks like a \#{goose}"
+            end
+          RUBY
+        end
```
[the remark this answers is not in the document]
biinari: This problem should now be resolved as will only detect statements with literal or constant arguments.

lib/rubocop/cop/rspec/repeated_include_example.rb
biinari: Also removed the matching of blocks as I can't see a need to detect ones that look similar to a static analysis.

lib/rubocop/cop/rspec/repeated_include_example.rb
pirj: Tested, this works fine. Maybe we don't need an example, we would be testing the node pattern children matching.

spec/rubocop/cop/rspec/repeated_include_example_spec.rb
```diff
@@ -0,0 +1,233 @@
...
+              #{include_method} 'two', @@foo
+              #{include_method} 'two', @@foo
+
+              #{include_method} 'two', $foo
+              #{include_method} 'two', $foo
+            end
+          RUBY
+        end
+
+        it "allows repeated #{include_method} with block" do
```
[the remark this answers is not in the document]
pirj: I can think of examples when the block is the same, but it doesn't mean those two included examples will be the same:
```ruby
it_behaves_like 'a foo' do
  let(:x) { rand }
end
```
or
```ruby
it_behaves_like 'a global counter' do
  let(:x) { $x += 1 }
end
```

But most probably this will never be found in the wild, and if it is, it's a rare exception. More likely such occurrences will be the result of an unintentional copy-paste.

And we're in a good position to be able to detect block identity by having its AST, that case where static analysis shine.

Still, if you don't feel like implementing it here, please feel free to leave this for future improvement.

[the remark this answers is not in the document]
biinari: My bad, I forgot to update the PR description. I think it makes sense to focus just on literal arguments for now without checking blocks.

lib/rubocop/cop/rspec/repeated_include_example.rb
```diff
@@ -0,0 +1,104 @@
...
+      #    end
+      #
+      #    describe 'bar' do
+      #      include_examples 'cool stuff'
+      #    end
+      #
+      #    # good
+      #    describe 'foo' do
+      #      it_behaves_like 'a cool', 'thing'
+      #      it_bahaves_like 'a cool', 'person'
```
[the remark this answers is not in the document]
biinari: Good spot!

lib/rubocop/cop/rspec/repeated_include_example.rb
```diff
@@ -0,0 +1,104 @@
...
+            .group_by { |child| signature_keys(child) }
+            .values
+            .reject(&:one?)
+            .flat_map { |items| add_repeated_lines(items) }
+        end
+
+        def literal_include_examples?(node)
+          include_examples?(node) &&
+            node.arguments.all?(&:recursive_literal_or_const?) &&
+            node.arguments.none?(&:block_pass_type?)
```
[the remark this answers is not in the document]
biinari: Of course, thanks!

pirj:
Thank you, @biinari !

PR #1009, opened 2020-08-19
Skip files with shared groups in RSpec/FilePath

### LINKED ISSUE #1008: shared_examples inconsistencies
I've update this gem to v1.43.1 and it started catching `RSpec/FilePath` in my `spec/support/shared_examples` folder, but I found the behavior confusing and inconsistent, some examples:

```ruby
# spec/support/shared_examples/my_class.rb
shared_examples_for MyClass do
   ...
end
```

The code above breaks with `RSpec/FilePath`, it expects `my_class_spec.rb` instead of just `my_class`, but the [rspec docs](https://relishapp.com/rspec/rspec-core/docs/example-groups/shared-examples) and [better specs](https://www.betterspecs.org/#shared) doesn't mention this convention, where does it come from? Looks like a bug to me.

But this is just one problem, if you have this instead:

```ruby
# spec/support/shared_examples/my_class.rb
shared_examples_for "my class" do
   # lots of ofenses
end
```

The code above doesn't break the `RSpec/FilePath` cop but also **doesn't break at cops at all** in the code, this file is just ignored completly by rubocop. If I change `"my class"` to `MyClass` then it catches all the ofenses and along with the `RSpec/FilePath`.

The third problem is when you have a module, like this:

```ruby
# spec/support/shared_examples/my_module_my_class.rb
shared_examples_for MyModule::MyClass do
   ...
end
```

It breaks with `RSpec/FilePath`, as it expects the file to be in `spec/support/shared_examples/my_module/my_class_spec.rb`, but if you move it there, RSpec fails to run:

```
NoMethodError:
  undefined method `shared_examples_for' for main:Object
```

Keeping it short:

- Where does the `_spec.rb` file name convention from `shared_examples` come from? Does it make sense to just break this cop when the shared example is declared using a class and not break when it is a string?
- Why the cops don't catch ofenses when you declared the shared example with a string?
- The module problem is some RSpec config I am missing?

pirj:
https://relishapp.com/rspec/rspec-core/docs/example-groups/shared-examples states that shared groups don't have to have the _spec.rb suffix, and even preferably omit _spec.

Fixes #1008

---

* [x] Added tests.
* [-] Updated documentation.

### DIFF
--- a/lib/rubocop/rspec/top_level_group.rb
+++ b/lib/rubocop/rspec/top_level_group.rb
@@ -16,7 +16,7 @@ module RuboCop
         return unless root_node
 
         top_level_groups.each do |node|
-          example_group?(node, &method(:on_top_level_example_group))
+          on_top_level_example_group(node) if example_group?(node)
           on_top_level_group(node)
         end
       end
@@ -29,9 +29,9 @@ module RuboCop
       private
 
       # Dummy methods to be overridden in the consumer
-      def on_top_level_example_group; end
+      def on_top_level_example_group(_node); end
 
-      def on_top_level_group; end
+      def on_top_level_group(_node); end
 
       def top_level_group?(node)
         top_level_groups.include?(node)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## Master (Unreleased)
 
 * Add `RSpec/RepeatedIncludeExample` cop. ([@biinari][])
+* Fix `RSpec/FilePath` when checking a file with a shared example. ([@pirj][])
 
 ## 1.43.1 (2020-08-17)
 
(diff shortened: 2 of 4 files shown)


pirj:
@fabioxgn Please let me know if this fix resolves the issue you are experiencing.

lib/rubocop/rspec/top_level_group.rb
```diff
@@ -16,7 +16,7 @@ def on_new_investigation
         return unless root_node
 
         top_level_groups.each do |node|
-          example_group?(node, &method(:on_top_level_example_group))
```
pirj, line 19: It turns out `example_group?` was yielding `nil` as there's no capture.

fabioxgn:
> @fabioxgn Please let me know if this fix resolves the issue you are experiencing.

It does! Thanks.

bquorning (approved):

PR #1018, opened 2020-08-22
Fix a false positive for `RSpec/EmptyExampleGroup`

koic:
This PR fixes a false positive for `RSpec/EmptyExampleGroup` when example group with examples defined in all `if` branches.

I found it on Capybara's CI.
https://travis-ci.org/github/teamcapybara/capybara/jobs/719967666#L846-L848

---

* [x] Added tests.
* [ ] Updated documentation.

### DIFF
--- a/spec/rubocop/cop/rspec/empty_example_group_spec.rb
+++ b/spec/rubocop/cop/rspec/empty_example_group_spec.rb
@@ -54,6 +54,62 @@ RSpec.describe RuboCop::Cop::RSpec::EmptyExampleGroup, :config do
     RUBY
   end
 
+  it 'ignores example group with examples defined in `if` branches' do
+    expect_no_offenses(<<~RUBY)
+      describe 'Ruby 2.3 syntax' do
+        version = 2.3
+
+        if RUBY_VERSION >= version
+          it { expect(use_safe_navigation_operator?(code)).to be(true) }
+        else
+          warn 'Ruby < 2.3 is barely supported, please use a newer version for development.'
+        end
+      end
+    RUBY
+  end
+
+  it 'ignores example group with examples but no examples in `if` branches' do
+    expect_no_offenses(<<~RUBY)
+      describe 'Ruby 2.3 syntax' do
+        version = 2.3
+
+        if RUBY_VERSION < version
+          warn 'Ruby < 2.3 is barely supported, please use a newer version for development.'
+        end
+
+        it { expect(newspaper(page)).to have_ads }
+      end
+    RUBY
+  end
+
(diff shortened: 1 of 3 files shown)


pirj:
Changed your code slightly to ignore the case when an example is defined in at least one of the `if` branches:
```ruby
      describe "usage of safe navigation" do
        if RUBY_VERSION >= 2.3
          it { expect(use_safe_navigation_operator?(code)).to be(true) }
        else
          warn 'Ruby < 2.3 is barely supported, please use a newer version for development.'
        end
      end
```

Let me explain why I think it is correct.
Branched `if` is typically the same as several `if`s.
```ruby
      describe "usage of safe navigation" do
        if RUBY_VERSION >= 2.3
          it { expect(use_safe_navigation_operator?(code)).to be(true) }
        end
        if RUBY_VERSION < 2.3
          warn 'Ruby < 2.3 is barely supported, please use a newer version for development.'
        end
      end
```
But we would treat the above as an example group with an example, right (as the first `if` has an example in its only branch).
Since we allow and detect examples in any of the `if` statements and don't require all `if` statements present in the example group to define examples, it would make sense to follow the same approach with `if` branches as if they would represent separate `if` statements.

pirj:
@bquorning @Darhazer quick approval?

Darhazer (approved):

spec/rubocop/cop/rspec/empty_example_group_spec.rb
```diff
@@ -54,6 +54,62 @@
...
+        if RUBY_VERSION < version
+          warn 'Ruby < 2.3 is barely supported, please use a newer version for development.'
+        end
+
+        it { expect(newspaper(page)).to have_ads }
+      end
+    RUBY
+  end
+
+  it 'flags an empty example group with no examples defined in if branches' do
```
bquorning, line 85:
```suggestion
  it 'flags an empty example group with no examples defined in `if` branches' do
```

bquorning (approved):

pirj:
Thanks!

koic:
I would like to thank you all ❤️

marcandre:
If I'm not mistaken, this hasn't been released yet?

pirj:
That is correct, @marcandre, [not yet](https://github.com/rubocop-hq/rubocop-rspec/blob/master/CHANGELOG.md#master-unreleased).

@bquorning There are some fixes, maybe you can cherry-pick and push out a minor release?
There are some new cops as well, but also some pull requests with new cops are hanging, I didn't have time to review them yet.

PR #1020, opened 2020-08-23
Fix false positives in RSpec/EmptyExampleGroup

pirj:
After running the cop against real-world-rspec I've found a number of false positives.

To make out users happy, here are three fixes, for:
 - example groups with pending examples
 - example groups defined inside methods
 - example groups defined inside examples

Before: 529 offenses
After: 291 offences (real ones mostly, and some due to custom example group definitions like in Pundit specs)

---

* [x] Added tests.
* [x] Updated documentation.

### DIFF
--- a/lib/rubocop/cop/rspec/empty_example_group.rb
+++ b/lib/rubocop/cop/rspec/empty_example_group.rb
@@ -160,6 +160,8 @@ module RuboCop
         PATTERN
 
         def on_block(node)
+          return if node.each_ancestor(:def, :defs).any?
+
           example_group_body(node) do |body|
             add_offense(node.send_node) unless examples?(body)
           end
--- a/lib/rubocop/cop/rspec/empty_example_group.rb
+++ b/lib/rubocop/cop/rspec/empty_example_group.rb
@@ -33,6 +33,11 @@ module RuboCop
       #     end
       #   end
       #
+      #   # good
+      #   describe Bacon do
+      #     pending 'will add tests later'
+      #   end
+      #
       # @example configuration
       #
       #   # .rubocop.yml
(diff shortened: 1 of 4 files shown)


lib/rubocop/cop/rspec/empty_example_group.rb
```diff
@@ -88,8 +93,12 @@ class EmptyExampleGroup < Base
         #   @return [Array<RuboCop::AST::Node>] matching nodes
         def_node_matcher :example_or_group_or_include?, <<~PATTERN
           {
+            #{Examples::ALL.send_pattern}
+            #{Examples::SKIPPED.send_pattern}
```
Darhazer, line 97: Why SKIPPED and PENDING do not match the ALL pattern?

pirj: Good catch, thank you. Don't know what I was thinking.

lib/rubocop/cop/rspec/empty_example_group.rb
```diff
@@ -88,8 +93,10 @@ class EmptyExampleGroup < Base
         #   @return [Array<RuboCop::AST::Node>] matching nodes
         def_node_matcher :example_or_group_or_include?, <<~PATTERN
           {
+            #{Examples::ALL.send_pattern}
             #{Examples::ALL.block_pattern}
             #{ExampleGroups::ALL.block_pattern}
+            #{ExampleGroups::SKIPPED.block_pattern}
```
Darhazer, line 99: same question here though: SKIPPED should be part of ALL

bquorning (approved):

PR #1021, opened 2020-08-23
Support rspec aruba integration in RSpec/DescribeClass

### LINKED ISSUE #1017: RSpec/DescribeClass: support for Aruba integration?
[Aruba](https://github.com/cucumber/aruba) is a gem from Cucumber for testing CLI applications, and it has RSpec integration.

From the [rspec docs](https://relishapp.com/philoserf/aruba/docs/rspec/getting-started-with-rspec-and-aruba):

```
To use the simple integration just require aruba/rspec in your
spec_helper.rb. After that you only need to flag your tests with type:
:aruba and some things are set up for.
```

```ruby
RSpec.describe 'Integrate Aruba into RSpec', :type => :aruba do
  context 'when to be or not be...' do
    it { expect(aruba).to be }
  end

  context 'when write file' do
    let(:file) { 'file.txt' }

    before(:each) { write_file file, 'Hello World' }

    it { expect(file).to be_an_existing_file }
    it { expect([file]).to include an_existing_file }
  end
end
```

But since RSpec/DescribeClass has no knowledge of Aruba, it expects a class argument even though it's a feature spec.

Is supporting this Aruba integration a wanted feature? If so, I can work on it - in fact, I already have a working solution.

Rafix02:
As discussed in #1017, @pirj suggested first extracting the Rails-related metadata already being ignored to an `IgnoredMetadata` config option, and then adding `aruba` to the default list.

My implementation treats the config as symbols, i.e. `type: :request` will be ignored, but `'type' => 'request'` won't. I tried making it so the configuration would mandate the typing, but Psych tells me symbols are not allowed, so that's the second best thing I could think of.

`reek` says my method smells, but since I'm unfamiliar with the codebase I couldn't think of anything better... the cops with similar behaviour I could find did more or less the same thing. If anyone can think of a better way, I'm all ears!

Please do also tell me whether you think the updated documentation is lacking in some sense. Should I add more info to the commit messages?

Fixes #1017.

---

* [x] Added tests.
* [x] Updated documentation.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@
 
 * Fix `RSpec/FilePath` when checking a file with a shared example. ([@pirj][])
 * Fix subject nesting detection in `RSpec/LeadingSubject`. ([@pirj][])
+* Add `IgnoredMetadata` configuration option to `RSpec/DescribeClass`. ([@Rafix02][])
 
 ## 1.43.1 (2020-08-17)
 
@@ -563,3 +564,4 @@ Compatibility release so users can upgrade RuboCop to 0.51.0. No new features.
 [@mockdeep]: https://github.com/mockdeep
 [@biinari]: https://github.com/biinari
 [@koic]: https://github.com/koic
+[@Rafix02]: https://github.com/Rafix02
--- a/config/default.yml
+++ b/config/default.yml
@@ -90,6 +90,7 @@ RSpec/DescribeClass:
     - feature
     - system
     - mailbox
+    - aruba
   VersionAdded: '1.0'
   VersionChanged: '1.44'
   StyleGuide: https://www.rubydoc.info/gems/rubocop-rspec/RuboCop/Cop/RSpec/DescribeClass
--- a/config/default.yml
+++ b/config/default.yml
@@ -76,7 +76,22 @@ RSpec/ContextWording:
 RSpec/DescribeClass:
   Description: Check that the first argument to the top-level describe is a constant.
   Enabled: true
+  IgnoredMetadata:
+    type:
+    - channel
+    - controller
+    - helper
+    - job
+    - mailer
+    - model
+    - request
+    - routing
+    - view
+    - feature
+    - system
+    - mailbox
   VersionAdded: '1.0'
+  VersionChanged: '1.44'
   StyleGuide: https://www.rubydoc.info/gems/rubocop-rspec/RuboCop/Cop/RSpec/DescribeClass
 
 RSpec/DescribeMethod:
--- a/lib/rubocop/cop/rspec/describe_class.rb
+++ b/lib/rubocop/cop/rspec/describe_class.rb
@@ -7,7 +7,7 @@ module RuboCop
       #
       # It can be configured to ignore strings when certain metadata is passed.
       #
-      # Ignores Rails `type` metadata by default.
+      # Ignores Rails and Aruba `type` metadata by default.
       #
       # @example `IgnoredMetadata` configuration
       #
--- a/lib/rubocop/cop/rspec/describe_class.rb
+++ b/lib/rubocop/cop/rspec/describe_class.rb
@@ -5,6 +5,19 @@ module RuboCop
     module RSpec
       # Check that the first argument to the top-level describe is a constant.
       #
+      # It can be configured to ignore strings when certain metadata is passed.
+      #
+      # Ignores Rails `type` metadata by default.
+      #
+      # @example `IgnoredMetadata` configuration
+      #
+      #   # .rubocop.yml
+      #   # RSpec/DescribeClass:
+      #   #   IgnoredMetadata:
+      #   #     type:
+      #   #       - request
+      #   #       - controller
+      #
       # @example
       #   # bad
       #   describe 'Do something' do
@@ -27,35 +40,41 @@ module RuboCop
         MSG = 'The first argument to describe should be '\
               'the class or module being tested.'
 
-        def_node_matcher :rails_metadata?, <<-PATTERN
-          (pair
-            (sym :type)
-            (sym { :channel :controller :helper :job :mailer :model :request
-                   :routing :view :feature :system :mailbox })
-          )
-        PATTERN
-
-        def_node_matcher :example_group_with_rails_metadata?, <<~PATTERN
-          (send #rspec? :describe ... (hash <#rails_metadata? ...>))
+        def_node_matcher :example_group_with_ignored_metadata?, <<~PATTERN
+          (send #rspec? :describe ... (hash <#ignored_metadata? ...>))
         PATTERN
 
         def_node_matcher :not_a_const_described, <<~PATTERN
           (send #rspec? :describe $[!const !#string_constant?] ...)
         PATTERN
 
-        def on_top_level_group(top_level_node)
-          return if example_group_with_rails_metadata?(top_level_node.send_node)
+        def_node_matcher :sym_pair?, <<~PATTERN
+          (pair $sym $sym)
+        PATTERN
 
-          not_a_const_described(top_level_node.send_node) do |described|
+        def on_top_level_group(node)
+          return if example_group_with_ignored_metadata?(node.send_node)
+
+          not_a_const_described(node.send_node) do |described|
             add_offense(described)
           end
         end
 
         private
 
+        def ignored_metadata?(node)
+          sym_pair?(node) do |key, value|
+            ignored_metadata[key.value.to_s].to_a.include?(value.value.to_s)
+          end
+        end
+
         def string_constant?(described)
           described.str_type? &&
             described.value.match?(/^(?:(?:::)?[A-Z]\w*)+$/)
+        end
+
+        def ignored_metadata
+          cop_config['IgnoredMetadata'] || {}
         end
       end
     end
--- a/lib/rubocop/cop/rspec/describe_class.rb
+++ b/lib/rubocop/cop/rspec/describe_class.rb
@@ -48,7 +48,7 @@ module RuboCop
           (send #rspec? :describe $[!const !#string_constant?] ...)
         PATTERN
 
-        def_node_matcher :sym_pair?, <<~PATTERN
+        def_node_matcher :sym_pair, <<~PATTERN
           (pair $sym $sym)
         PATTERN
 
@@ -63,7 +63,7 @@ module RuboCop
         private
 
         def ignored_metadata?(node)
-          sym_pair?(node) do |key, value|
+          sym_pair(node) do |key, value|
             ignored_metadata[key.value.to_s].to_a.include?(value.value.to_s)
           end
         end
--- a/spec/rubocop/cop/rspec/describe_class_spec.rb
+++ b/spec/rubocop/cop/rspec/describe_class_spec.rb
@@ -1,8 +1,6 @@
 # frozen_string_literal: true
 
-RSpec.describe RuboCop::Cop::RSpec::DescribeClass do
-  subject(:cop) { described_class.new }
-
+RSpec.describe RuboCop::Cop::RSpec::DescribeClass, :config do
   it 'checks first-line describe statements' do
     expect_offense(<<-RUBY)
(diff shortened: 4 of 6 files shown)


lib/rubocop/cop/rspec/describe_class.rb
```diff
@@ -27,36 +40,37 @@ class DescribeClass < Base
...
+          not_a_const_described(node.send_node) do |described|
             add_offense(described)
           end
         end
 
         private
 
+        def ignored_metadata?(node)
+          ignored_metadata.find { |key, _v| key.to_sym == node.key.value }
+            &.last&.map(&:to_sym)&.include? node.value.value
```
pirj, line 63: Is it the same as?
```ruby
ignored_metadata[node.key.value.to_s].include?(node.value.value.to_s)
```
So you try to avoid string allocation for each metadata key and value?

By the way, what if metadata value is not a string or symbol literal?

```ruby
describe 'A', type: method_call do
```

or
```ruby
describe 'A', type: 'fea' + 'ture' do
```

I'd suggest narrowing this down to string and symbol literals.

config/default.yml
```diff
@@ -76,7 +76,23 @@ RSpec/ContextWording:
 RSpec/DescribeClass:
   Description: Check that the first argument to the top-level describe is a constant.
   Enabled: true
+  IgnoredMetadata:
+    type:
```
pirj, line 80: By the way, it seems to be possible to put a symbol if you surround it by colons:
```yaml
IgnoredMetadata:
  :type:
    - channel
```
Should be sufficient for keys.

Rafix02: By doing that I get the following error:
```
Failure/Error: CONFIG         = YAML.safe_load(CONFIG_DEFAULT.read).freeze

Psych::DisallowedClass:
  Tried to load unspecified class: Symbol
```
 The method can be configured to allow symbols, but I couldn't find anywhere why it's unsafe by default in the first place. What do you think?

manual/cops_rspec.md
```diff
@@ -320,12 +320,26 @@ Prefixes | `when`, `with`, `without` | Array
...
 
+#### `IgnoredMetadata` configuration
+
+```ruby
+# .rubocop.yml
+# RSpec/DescribeClass:
+#   IgnoredMetadata:
+#     type:
+#       - request
+#       - controller
```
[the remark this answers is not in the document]
Rafix02: True. I just couldn't think of any other real possibities, you think `foo: :bar` would be communicative enough?

spec/rubocop/cop/rspec/describe_class_spec.rb
```diff
@@ -120,77 +118,98 @@
     end
   end
 
-  it 'ignores request specs' do
-    expect_no_offenses(<<-RUBY)
-      describe 'my new feature', type: :request do
+  context 'when IgnoredMetadata is configured' do
```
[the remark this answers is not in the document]
Rafix02: I'm not sure if I misunderstood, but following the example in the link I added a spec with an empty config right before the context, having the same metadata as the first spec in it. Does that accomplish the same as you suggested?

lib/rubocop/cop/rspec/describe_class.rb
Rafix02: The only difference is that your snippet would ignore both symbols and strings, while currently the method only ignores symbols.

You're right, the code breaks in both of those cases. I added examples, validated for pairs of symbols and changed the code to your version. You want it to also work for strings? If so, can you think of a simpler way to do it than replicating the condition?

manual/cops_rspec.md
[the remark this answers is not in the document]
pirj: Maybe `type: :policy`, `type: :ability` and `foo: :bar as well?

It actually just took me by surprise why we ignore Rails' metadata. Models, controllers, jobs and helpers should have a class name, why are they ignored? 🤔

spec/rubocop/cop/rspec/describe_class_spec.rb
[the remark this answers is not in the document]
pirj: When you don't specify `cop_config` explicitly, the one from `config/default.yml` is used.

To show that `IgnoredMetadata` config option works, you can keep all the existing examples with the default config, as previously, and:
 - add an example that flags the example group with `foo: :bar` metadata
 - add an example with explicit `IgnoredMetadata: 'foo' => ['bar']` that ignores the same example group

Same as here:
https://github.com/rubocop-hq/rubocop-rspec/blob/master/spec/rubocop/cop/rspec/empty_example_group_spec.rb#L203 - not configured - there's an offence.
https://github.com/rubocop-hq/rubocop-rspec/blob/0b5b672c6f76b7e0ee91072748a7ad7881da1cd9/spec/rubocop/cop/rspec/empty_example_group_spec.rb#L214 - configured - no offence

config/default.yml
pirj: Oh, bummer, didn't know symbols are now allowed. 👍

lib/rubocop/cop/rspec/describe_class.rb
```diff
@@ -27,36 +40,42 @@ class DescribeClass < Base
...
-            (sym :type)
-            (sym { :channel :controller :helper :job :mailer :model :request
-                   :routing :view :feature :system :mailbox })
-          )
-        PATTERN
-
-        def_node_matcher :example_group_with_rails_metadata?, <<~PATTERN
-          (send #rspec? :describe ... (hash <#rails_metadata? ...>))
+        def_node_matcher :example_group_with_ignored_metadata?, <<~PATTERN
+          (send #rspec? :describe ... (hash <#ignored_metadata? ...>))
```
pirj, line 44: Speaking of `describe`, there are more than just it:
```ruby
        GROUPS  = SelectorSet.new(%i[describe context feature example_group])
        SKIPPED = SelectorSet.new(%i[xdescribe xcontext xfeature])
        FOCUSED = SelectorSet.new(%i[fdescribe fcontext ffeature])
```

If you change the entry point from `on_top_level_group` to `on_top_level_example_group`, you'll be skipping shared example groups.

In this case, if you want to remove the limitation of only detecting `describe`, it's as easy as removing `:describe` part from the node pattern above + a spec example that e.g. `context` is flagged, but `shared_examples_for` are not.

Can be handled separately. Just let me know what you'll decide, I can file a separate ticket.

pirj: There's one more place to fix, `not_a_const_described`:
```
(send #rspec? :describe $[!const !#string_constant?] ...)
```

lib/rubocop/cop/rspec/describe_class.rb
```diff
@@ -27,36 +40,42 @@ class DescribeClass < Base
...
 
+        def ignored_metadata?(node)
+          return false unless sym_pair?(node)
+
+          ignored_metadata[node.key.value.to_s]&.include? node.value.value.to_s
+        end
+
+        def sym_pair?(node)
+          node.key.sym_type? && node.value.sym_type?
+        end
```
pirj, line 69: This can be defined as:
```ruby
        def_node_matcher :sym_pair?, <<~PATTERN
          (pair sym sym)
        PATTERN
```
if you like.

pirj:
Looks good, apart from the spec and configuration `@example`.

Tried to rebase and force-push to reduce `+3,562 −60` stats, but for some reason, it fails.

manual/cops_rspec.md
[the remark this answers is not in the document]
pirj: Filed https://github.com/rubocop-hq/rubocop-rspec/issues/1023

lib/rubocop/cop/rspec/describe_class.rb
pirj: Filed https://github.com/rubocop-hq/rubocop-rspec/issues/1025

lib/rubocop/cop/rspec/describe_class.rb
bquorning: Or even
```rb
def_node_matcher :sym_pair?, <<~PATTERN
  (pair $sym $sym)
PATTERN
```
(capturing the key and value) – in which case `#ignored_metadata?` could be reduced to
```rb
def ignored_metadata?(node)
  sym_pair?(node) do |key, value|
    ignored_metadata[key.value.to_s]&.include?(value.value.to_s)
  end
end
```

lib/rubocop/cop/rspec/describe_class.rb
```diff
@@ -27,36 +40,42 @@ class DescribeClass < Base
...
             add_offense(described)
           end
         end
 
         private
 
+        def ignored_metadata?(node)
+          return false unless sym_pair?(node)
+
+          ignored_metadata[node.key.value.to_s]&.include? node.value.value.to_s
```
bquorning, line 64: The `&.` can be avoided by using a local variable, e.g.:
```rb
ignored_values = ignored_metadata.fetch(node.key.value.to_s, [])
ignored_values.include?(node.value.value.to_s)
```

Darhazer: Or by a simple conversion protocol `nil.to_a` results in an empty array, so it won't include anything

lib/rubocop/cop/rspec/describe_class.rb
pirj: Wondering if it needs to be `sym_pair?(node).any? |key, value|`?

manual/cops_rspec.md
```diff
@@ -0,0 +1,3392 @@
+# RSpec
```
bquorning, line 1: Please remove this file from the PR

We recently switched from using the `manual/` folder to using the `docs/` folder, and this file was probably created somewhere in the middle. Sorry about the confusion 😄

pirj (approved):
Code review notes addressed. Rebased, squashed. The build is all green.

@bquorning @Darhazer please take a look.

lib/rubocop/cop/rspec/describe_class.rb
```diff
@@ -27,36 +40,42 @@ class DescribeClass < Base
...
+          (send #rspec? :describe ... (hash <#ignored_metadata? ...>))
         PATTERN
 
         def_node_matcher :not_a_const_described, <<~PATTERN
           (send #rspec? :describe $[!const !#string_constant?] ...)
         PATTERN
 
-        def on_top_level_group(top_level_node)
-          return if example_group_with_rails_metadata?(top_level_node.send_node)
+        def_node_matcher :sym_pair?, <<~PATTERN
```
Darhazer, line 51: niptick: since this is meant to yield the key and value, and not merely return a boolean, I'd remove the `?` from the name

lib/rubocop/cop/rspec/describe_class.rb
```diff
@@ -27,36 +40,42 @@ class DescribeClass < Base
...
+
+          not_a_const_described(node.send_node) do |described|
             add_offense(described)
           end
         end
 
         private
 
+        def ignored_metadata?(node)
+          sym_pair?(node) do |key, value|
```
Darhazer, line 66: I wonder if capturing the whole metadata and using
```
node.pairs.any? do |key, value|
  ignored_metadata[key.value.to_s].to_a.include?(value.value.to_s)
end
```
won't be easier.
I'm fine with either way

Darhazer: Or if we really need to work only on sym types, it would include some filtering, so maybe the matcher version is cleaner

Darhazer (approved):

CHANGELOG.md
```diff
@@ -12,6 +12,7 @@
 
 * Fix `RSpec/FilePath` when checking a file with a shared example. ([@pirj][])
 * Fix subject nesting detection in `RSpec/LeadingSubject`. ([@pirj][])
+* Add `IgnoredMetadata` configuration option to `RSpec/DescribeClass`. ([@Rafix02][])
```
pirj, line 15: Oh

pirj:
Thank you, @Rafix02 !

PR #1027, opened 2020-08-28
Allow context with no args followed by include context

mpospelov:
Relax `RSpec/MissingExampleGroupArgument` to allow `context` with no arguments in case of `include_context`
 
```ruby
context do
  include_context 'when something is different'

  specify { expect(1 + 2).to eq(3) }
end
```
---

* [x] Added tests.
* [x] Updated documentation.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 * Add `RSpec/RepeatedIncludeExample` cop. ([@biinari][])
 * Fix false positives in `RSpec/EmptyExampleGroup`. ([@pirj][])
 * Add `RSpec/StubbedMock` cop. ([@bquorning][], [@pirj][])
+* Relax `RSpec/MissingExampleGroupArgument` cop so `context` with no arguments can be used if `include_context` is followed on next line. ([@mpospelov][])
 
 ## 1.43.2 (2020-08-25)
 
@@ -561,3 +562,4 @@ Compatibility release so users can upgrade RuboCop to 0.51.0. No new features.
 [@jtannas]: https://github.com/jtannas
 [@mockdeep]: https://github.com/mockdeep
 [@biinari]: https://github.com/biinari
+[@mpospelov]: https://github.com/mpospelov
--- a/docs/modules/ROOT/pages/cops_rspec.adoc
+++ b/docs/modules/ROOT/pages/cops_rspec.adoc
@@ -2499,6 +2499,10 @@ end
 
 describe "A feature example" do
 end
+
+context do
+  include_context 'when something is different'
+end
 ----
 
 === References
--- a/lib/rubocop/cop/rspec/missing_example_group_argument.rb
+++ b/lib/rubocop/cop/rspec/missing_example_group_argument.rb
@@ -19,12 +19,25 @@ module RuboCop
       #
       #   describe "A feature example" do
       #   end
+      #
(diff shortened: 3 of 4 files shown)


pirj:
Just wondering, what if there are two `include_context`s?
Is the docstring from `include_context` included in the documentation format RSpec output?

mpospelov:
> Just wondering, what if there are two include_contexts?
this one should be ignored too

> Is the docstring from include_context included in the documentation format RSpec output?

@pirj yeah missed that thing, suddenly it's not included 😞

So I think it's better to duplicate this naming because of output

pirj:
I've seen another option in the wild:
```ruby
RSpec.describe 'A' do
  context with: 'something different' do
    it { }
  end
end
```

Where `with:` is:

```ruby
RSpec.shared_context 'something else' do
  let(:a) { 1 }
end

RSpec.describe 'A' do
  before(with: ->(_) { true }) do
    self.class.include_context self.class.metadata[:with]
    self.class.metadata[:description] = self.class.metadata[:with]
  end

  context '', with: 'something else' do
    it { expect(a).to eq(1) }
  end

  context 'without a' do
    specify do
      expect { a }
        .to raise_error(NameError)
    end
  end
end
```

And, of course, you can put this `before` in the `spec/spec_helper.rb`:
```ruby
RSpec.configure do |config|
  config.before(with: ->(_) { true }) do
    self.class.include_context self.class.metadata[:with]
    self.class.metadata[:description] = self.class.metadata[:with]
  end
end
```

The downside, coincidental, is that RSpec treats the first argument to an example group as a docstring, not matter what, even if it's a Hash intended to represent metadata. So writing it as `context with: 'something else' do` (without the first `''` argument) won't work the same way.

Unfortunately, the setting of `metadata[:description]` doesn't seem to affect the documentation output. I can't remember exacly how it was done, probably example group creating method was overridden.

```ruby
RSpec.describe 'A' do

  def self.context(docstring, **metadata, &block)
    docstring = metadata[:with] if metadata[:with]
    super(docstring, **metadata, &block).tap do |example_group|
      example_group.include_context metadata[:with] if metadata[:with]
    end
  end

  context 'aaaaa', with: 'something else' do
    it { expect(a).to eq(1) }
  end

  context 'without a' do
    specify do
      expect { a }
        .to raise_error(NameError)
    end
  end
end
```
This works. The downside is that it only overrides `context`, but this is semantically sufficient for your case.
As with the previous implementation, you can extract `def context` to a module and `extend` the configuration:
```ruby
module WithContext
  def context(docstring, **metadata, &block)
    docstring = metadata[:with] if metadata[:with]
    super(docstring, **metadata, &block).tap do |example_group|
      example_group.include_context metadata[:with] if metadata[:with]
    end
  end
end
```
```ruby
RSpec.configure do |config|
  config.extend WithContext
end
```

Please also be aware that if you happen to both pass the docstring and `with:`, in the above implementation the docstring will be overridden. So if that's the case for you, you'll have to figure out a way to combine the two.

Also, you may add some trickery around passed parameters, and check if the first one is a hash with needed metadata, then build it on the fly and avoid from this superfluous first empty argument in example group definitions.

I hope you'll find this pattern useful and it fulfils your needs. If you make it a gem, I would happily adopt it in my current project.

Still, I'd prefer to keep the cop intact.

PR #1033, opened 2020-09-19
Add named_only style for RSpec/NamedSubject cop

kuahyeow:
This adds `named_only` style for the RSpec/NamedSubject cop. It is intended as a smaller step to first catch explicit subject references where the `subject(:name) { .. }`  is named.

We keep the default style (I have named it `always`). 

---

* [x] Added tests.
* [x] Updated documentation.

### DIFF
--- a/docs/modules/ROOT/pages/cops_rspec.adoc
+++ b/docs/modules/ROOT/pages/cops_rspec.adoc
@@ -3317,7 +3317,7 @@ end
 | Yes
 | No
 | 1.5.3
-| -
+| <<next>>
 |===
 
 Checks for explicitly referenced test subjects.
@@ -3329,11 +3329,13 @@ name it using `subject(:your_subject_name) { ... }`. Your test subjects
 should be the most important object in your tests so they deserve
 a descriptive name.
 
-This cop can be configured in your configuration using the
-`IgnoreSharedExamples` which will not report offenses for implicit
+This cop can be configured in your configuration using `EnforcedStyle`,
+and `IgnoreSharedExamples` which will not report offenses for implicit
 subjects in shared example groups.
 
 === Examples
+
+==== `EnforcedStyle: always` (default)
 
 [source,ruby]
 ----
@@ -3347,7 +3349,7 @@ RSpec.describe User do
 end
 
 # good
-RSpec.describe Foo do
+RSpec.describe User do
   subject(:user) { described_class.new }
 
   it 'is valid' do
@@ -3356,10 +3358,49 @@ RSpec.describe Foo do
 end
 
 # also good
-RSpec.describe Foo do
+RSpec.describe User do
   subject(:user) { described_class.new }
 
   it { is_expected.to be_valid }
+end
+----
+
+==== `EnforcedStyle: named_only`
+
+[source,ruby]
+----
+# bad
+RSpec.describe User do
+  subject(:user) { described_class.new }
+
+  it 'is valid' do
+    expect(subject.valid?).to be(true)
+  end
+end
+
+# good
+RSpec.describe User do
+  subject(:user) { described_class.new }
+
+  it 'is valid' do
+    expect(user.valid?).to be(true)
+  end
+end
+
+# also good
+RSpec.describe User do
+  subject { described_class.new }
+
+  it { is_expected.to be_valid }
+end
+
+# acceptable
+RSpec.describe User do
+  subject { described_class.new }
+
+  it 'is valid' do
+    expect(subject.valid?).to be(true)
+  end
 end
 ----
 
@@ -3367,6 +3408,10 @@ end
 
 |===
 | Name | Default value | Configurable values
+
+| EnforcedStyle
+| `always`
+| `always`, `named_only`
 
 | IgnoreSharedExamples
 | `true`
--- a/lib/rubocop/cop/rspec/named_subject.rb
+++ b/lib/rubocop/cop/rspec/named_subject.rb
@@ -12,11 +12,11 @@ module RuboCop
       # should be the most important object in your tests so they deserve
       # a descriptive name.
       #
-      # This cop can be configured in your configuration using the
-      # `IgnoreSharedExamples` which will not report offenses for implicit
+      # This cop can be configured in your configuration using `EnforcedStyle`,
+      # and `IgnoreSharedExamples` which will not report offenses for implicit
       # subjects in shared example groups.
       #
-      # @example
+      # @example `EnforcedStyle: always` (default)
       #   # bad
       #   RSpec.describe User do
       #     subject { described_class.new }
@@ -27,7 +27,7 @@ module RuboCop
       #   end
       #
       #   # good
-      #   RSpec.describe Foo do
+      #   RSpec.describe User do
       #     subject(:user) { described_class.new }
       #
       #     it 'is valid' do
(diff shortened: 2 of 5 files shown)


docs/modules/ROOT/pages/cops_rspec.adoc
```diff
@@ -2868,11 +2870,54 @@ RSpec.describe Foo do
...
+  subject(:user) { described_class.new }
+
+  it 'is valid' do
+    expect(subject.valid?).to be(true)
+  end
+end
+
+# good
+RSpec.describe Foo do
+  subject(:user) { described_class.new }
```
pirj, line 2888: The described class is Foo, maybe adjust the subject name to match?

lib/rubocop/cop/rspec/named_subject.rb
```diff
@@ -62,14 +99,45 @@ def on_block(node)
...
             node.each_ancestor(:block).any?(&method(:shared_example?))
         end
+
+        def check_explicit_subject(node)
+          return unless disallow_explicit_subject?(node)
+
+          add_offense(node.loc.selector)
+        end
+
+        def disallow_explicit_subject?(node)
```
pirj, line 119: I personally find it hard to understand what the method does just looking at its name.
If the explicit subject is `user`, why it would be disallowed?

lib/rubocop/cop/rspec/named_subject.rb
```diff
@@ -41,7 +41,44 @@ module RSpec
...
+      #
+      # @example `EnforcedStyle: named_only`
+      #   # bad
+      #   RSpec.describe User do
+      #     subject(:user) { described_class.new }
+      #
+      #     it 'is valid' do
+      #       expect(subject.valid?).to be(true)
+      #     end
+      #   end
```
pirj, line 53: Wondering how the cop will behave with a nested context:
```ruby
RSpec.describe User do
  subject(:user) { described_class.new }
  describe 'age' do
    subject { user.age }
    it 'allows driving a car' do
      expect(subject.can_drive_a_car).to be(true)
    end
  end
end
```

lib/rubocop/cop/rspec/named_subject.rb
```diff
@@ -62,14 +99,45 @@ def on_block(node)
...
+          node
+            .each_ancestor(:block)
+            .lazy
+            .map { |block_node| find_subject(block_node) }
+            .find(&:itself)
+        end
+
+        def find_subject(block_node)
+          block_node.body.child_nodes.find { |send_node| subject?(send_node) }
+        end
```
pirj, line 140: What if the subject is defined in the other branch?
```ruby
RSpec.describe User do
  describe 'age' do
    subject(:age) { user.age }
    it 'allows driving a car' do
      expect(subject.can_drive_a_car).to be(true)
    end
  end

  describe 'foo' do
    it { expect(subject).to be_a(User) } # implicit `subject` is used
    # ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ offence is detected, there's a named subject in an unrelated sibling example group
  end
end
```

pirj (commented):
Thanks, nice addition.

I'm quite concerned about edge cases though. `def_node_search` and `each_ancestor` is a sure way to false positives.

spec/rubocop/cop/rspec/named_subject_spec.rb
```diff
@@ -60,6 +60,96 @@ def foo
...
+            expect(subject).to be_a(User)
+          end
+
+          it "is valid" do
+            expect(subject.valid?).to be(true)
+          end
+
+          it "is new" do
+            expect(subject).to be_new_record
+          end
```
pirj, line 101: This is redundant. I see other examples in this spec have the same duplication. I suggest to remove and only keep:
```ruby
         it "is valid" do
            expect(subject.valid?).to be(true)
          end
```

lib/rubocop/cop/rspec/named_subject.rb
pirj: Actually, I was wrong, no offence is detected.

lib/rubocop/cop/rspec/named_subject.rb
pirj: No offence is detected 👍 

Would you please an example to the spec to indicate that it's as by design?

lib/rubocop/cop/rspec/named_subject.rb
pirj: Maybe call it a `allow_explicit_subject?` to avoid double negation (`unless disallow_...` -> `if allow_...`)?

docs/modules/ROOT/pages/cops_rspec.adoc
kuahyeow: Hmm, it seem we describe a mixture of `Foo` and `User`, so I have made it consistently `describe User`. Hope that's OK.

lib/rubocop/cop/rspec/named_subject.rb
kuahyeow: Thanks! Sure, I have added new example `ignores subject where the closest declaration is not named` to the specs

lib/rubocop/cop/rspec/named_subject.rb
kuahyeow: Have changed it to `allow_explicit_subject?` Let me know if this is clearer !

kuahyeow:
@pirj Thanks for the nice review ! I have added changes as requested, please see each thread above.

lib/rubocop/cop/rspec/named_subject.rb
pirj: I had doubts about such an example, but it works fine too 👍 
```ruby
_         RSpec.describe User do
            describe 'valid' do
              subject(:valid) { described_class.new.valid? }

              it 'is valid' do
                expect(valid).to be(true)
              end
            end

_           describe 'age' do
              it 'allows driving a car' do
                expect(subject.can_drive_a_car).to be(true)
              end
            end
          end
 ```

File: lib/rubocop/cop/rspec/named_subject.rb (hunk 2 above)
Comment (pirj) [reply lib/rubocop/cop/rspec/named_subject.rb]: Much better, thanks!

**Review by pirj — APPROVED** [review]
Thank you!

**Comment by inkstak** [comment]
Nice addition. Any release planned soon?

Should this cop also ignore implicit subjects?
This would be useful when `is_expected` is not recommanded by `RSpec/ImplicitSubject`  and no explicit subject is  defined.

```ruby
RSpec.describe Klass do
  it do
    prepare_something_to_be_true
    expect(subject).to be_something
  end
end
```

lib/rubocop/cop/rspec/named_subject.rb
```diff
@@ -62,14 +99,53 @@ def on_block(node) # rubocop:disable InternalAffairs/NumblockHandler
...
+          style == :always
+        end
+
+        def named_only?(node)
+          style == :named_only &&
+            subject_definition_is_named?(node)
+        end
+
+        def subject_definition_is_named?(node)
+          subject = nearest_subject(node)
```
bquorning, line 133: I wonder about the performance of this two-way ast traversal on each `subject` reference.

pirj: We do this in a couple of other cops, like `EmptyExampleGroup`, `ImplicitBlockExpectation` just to name a few.

pirj: Ping

bquorning: I just tried running [RuboCop's profiler](https://github.com/rubocop/rubocop/blob/master/bin/rubocop-profile) on the specs a large repository I work on. And the performance of these AST traversals seem to be fine.

bquorning (approved):
Needs a rebase to fix changelog conflict (sorry about that)

pirj:
@bquorning I won't be merging this so far, not to interfere with the release.

pirj:
Thank you, @kuahyeow ! 
Please accept our apologies it took two years to merge.

PR #1036, opened 2020-09-30
Use the new Node Pattern multiple term union syntax

pirj:
`rubocop-ast` has recently introduced syntax that allows for matching multiple subsequent terms in a union.

https://github.com/rubocop-hq/rubocop-ast/commit/e28920a56c894bdf5b482d274cbc88e91f54e5a3#diff-9a2615ac7e37917e7fb9e29d3cb4beb0

https://github.com/rubocop-hq/rubocop-ast/issues/79

This change is to demonstrate this syntax in action, and to (subjectively and arguably) simplify the node pattern used.

~I'm targeting this for 2.0 to avoid bumping our dependencies for `rubocop-ast` to 0.7.1.~

---

* [-] Added tests.
* [-] Updated documentation.

### DIFF
--- a/lib/rubocop/cop/rspec/around_block.rb
+++ b/lib/rubocop/cop/rspec/around_block.rb
@@ -31,7 +31,7 @@ module RuboCop
                          'or `%<arg>s.run`.'
 
         def_node_matcher :hook, <<-PATTERN
-          (block {(send nil? :around) (send nil? :around sym)} (args $...) ...)
+          (block (send nil? :around sym ?) (args $...) ...)
         PATTERN
 
         def_node_search :find_arg_usage, <<-PATTERN
--- a/lib/rubocop/cop/rspec/subject_stub.rb
+++ b/lib/rubocop/cop/rspec/subject_stub.rb
@@ -43,10 +43,10 @@ module RuboCop
         #
         #   @yield [Symbol] subject name
         def_node_matcher :subject, <<-PATTERN
-          {
-            (block (send nil? :subject (sym $_)) args ...)
-            (block (send nil? $:subject) args ...)
-          }
+            (block
+              (send nil?
+                {:subject (sym $_) | $:subject}
+              ) args ...)
         PATTERN
 
         # @!method message_expectation?(node, method_name)
--- a/lib/rubocop/cop/rspec/implicit_block_expectation.rb
+++ b/lib/rubocop/cop/rspec/implicit_block_expectation.rb
@@ -22,8 +22,7 @@ module RuboCop
         def_node_matcher :lambda?, <<-PATTERN
           {
             (send (const nil? :Proc) :new)
-            (send nil? :proc)
-            (send nil? :lambda)
+            (send nil? {:proc :lambda})
           }
         PATTERN
 
--- a/lib/rubocop/rspec/hook.rb
+++ b/lib/rubocop/rspec/hook.rb
@@ -6,11 +6,7 @@ module RuboCop
     class Hook < Concept
       def_node_matcher :extract_metadata, <<~PATTERN
         (block
-          {
-            (send _ _ #valid_scope? $...)
-            (send _ _ $...)
-          }
-          ...
+          (send _ _ #valid_scope? ? $...) ...
         )
       PATTERN
 
--- a/rubocop-rspec.gemspec
+++ b/rubocop-rspec.gemspec
@@ -37,6 +37,7 @@ Gem::Specification.new do |spec|
   }
 
   spec.add_runtime_dependency 'rubocop', '~> 0.87'
+  spec.add_runtime_dependency 'rubocop-ast', '~> 0.7.1'
 
   spec.add_development_dependency 'rack'
   spec.add_development_dependency 'rake'


Darhazer:
This syntax is great and I'm sure we can apply it to many more places.
It's OK to bump rubocop-ast dependency as long as it doesn't require bumping rubocop dependency, which should be carefully tested.

bquorning:
> It's OK to bump rubocop-ast dependency

If we do that, let’s depend on 0.7.1 instead of 0.7.0 then. The more recent version [removes the dependency on the strscan gem](https://github.com/rubocop-hq/rubocop-ast/compare/v0.7.0...v0.7.1).

pirj:
Thanks for the support, guys! I went through all the node pattern matchers and have quickly applied this syntax.
If you like it, I'll add `rubocop-ast` to our gemspec, and this becomes mergeable.

I'm a bit over-enthusiastic about it, please take a look at those updated patterns, do you find them readable, or at least more readable than before?

lib/rubocop/cop/rspec/focus.rb
```diff
@@ -30,8 +30,13 @@ class Focus < Base
                           Examples::PENDING).node_pattern_union
 
         def_node_matcher :metadata, <<-PATTERN
-          {(send #rspec? #focusable_selector? <$(sym :focus) ...>)
-           (send #rspec? #focusable_selector? ... (hash <$(pair (sym :focus) true) ...>))}
+          (send #rspec? #focusable_selector?
+            {
+              <$(sym :focus) ...>
```
pirj, line 35: `< >` is necessary, some specs fail if it's removed.
The same for `pending` and `multiple_expectations`: I'm not sure if it's an inadequate test coverage, or for `... hash` it's not needed.

lib/rubocop/cop/rspec/around_block.rb
```diff
@@ -31,7 +31,7 @@ class AroundBlock < Base
                          'or `%<arg>s.run`.'
 
         def_node_matcher :hook, <<-PATTERN
-          (block {(send nil? :around) (send nil? :around sym)} (args $...) ...)
+          (block (send nil? :around {sym | }) (args $...) ...)
```
Darhazer, line 34: isn't there a syntax for optional node, e.g. `sym ?` (note the space is required in order not to turn `sym?` into a predicate

lib/rubocop/cop/rspec/focus.rb
Darhazer: I think this is an example where readability is not  being improved

lib/rubocop/cop/rspec/multiple_expectations.rb
```diff
@@ -54,10 +54,14 @@ class MultipleExpectations < Base
         TRUE = ->(node) { node.true_type? }
 
         def_node_matcher :aggregate_failures?, <<-PATTERN
-          (block {
-              (send _ _ <(sym :aggregate_failures) ...>)
-              (send _ _ ... (hash <(pair (sym :aggregate_failures) %1) ...>))
-            } ...)
+          (block (send _ _
```
Darhazer, line 57: this one also doesn't improve

lib/rubocop/cop/rspec/pending.rb
```diff
@@ -40,10 +40,13 @@ class Pending < Base
         def_node_matcher :skippable?, SKIPPABLE.send_pattern
 
         def_node_matcher :skipped_in_metadata?, <<-PATTERN
-          {
-            (send _ _ <#skip_or_pending? ...>)
-            (send _ _ ... (hash <(pair #skip_or_pending? { true str }) ...>))
-          }
+          (send _ _
```
Darhazer, line 43: maybe we need another type of a helper (to use as a predicate) that takes care of the fact that it might or might not be a hash

lib/rubocop/cop/rspec/stubbed_mock.rb
```diff
@@ -106,10 +106,13 @@ class StubbedMock < Base
...
         def_node_matcher :matcher_with_hash, <<~PATTERN
-          {
-            (send nil? :receive_messages hash)           # receive_messages(foo: 'bar', baz: 'qux')
-            (send nil? :receive_message_chain ... hash)  # receive_message_chain(:foo, bar: 'baz')
-          }
+          (send nil?
+            {
+              :receive_messages hash           # receive_messages(foo: 'bar', baz: 'qux')
+              |
+              :receive_message_chain ... hash  # receive_message_chain(:foo, bar: 'baz')
```
Darhazer, line 113: wonder if it won't be fixed by just having `<hash ...>`

lib/rubocop/cop/rspec/around_block.rb
pirj: Nice! There's no month without learning a new Node Pattern trick.

lib/rubocop/cop/rspec/pending.rb
pirj: Indeed. Reverting this meanhile along with the other two.

lib/rubocop/cop/rspec/stubbed_mock.rb
pirj: Not sure what you mean exactly. `receive_messages`/`receive_message_chain` differ here, and the latter can have positional arguments, while the former can't.
Reverting this as well.

lib/rubocop/cop/rspec/subject_stub.rb
```diff
@@ -43,10 +43,10 @@ class SubjectStub < Base
         #
         #   @yield [Symbol] subject name
         def_node_matcher :subject, <<-PATTERN
-          {
-            (block (send nil? :subject (sym $_)) args ...)
-            (block (send nil? $:subject) args ...)
-          }
+            (block
+              (send nil?
+                {:subject (sym $_) | $:subject}
```
pirj, line 48: Ok, just one real usage left. This one looks better to me than the original.

Darhazer (approved):

PR #1039, opened 2020-10-12
Fix a false positive for `RSpec/RepeatedExampleGroupBody` when `pending` or `skip` have argument(s).

Tietew:
`RSpec/RepeatedExampleGroupBody` does not register offenses when example group body is `skip` or `pending`.

```ruby
describe 'alpha' do
  skip
end

describe 'bravo' do
  skip
end
```

But each of them has a same argument, registers offenses.

```ruby
describe 'alpha' do
^^^^^^^^^^^^^^^^^^^ Repeated describe block body
  skip 'something needed'
end

describe 'bravo' do
^^^^^^^^^^^^^^^^^^^ Repeated describe block body
  skip 'something needed'
end
```

I thinks this is a false positive.
This PR allows this.

---

* [x] Added tests.
* [ ] Updated documentation.

### DIFF
--- a/spec/rubocop/cop/rspec/repeated_example_group_body_spec.rb
+++ b/spec/rubocop/cop/rspec/repeated_example_group_body_spec.rb
@@ -224,6 +224,50 @@ RSpec.describe RuboCop::Cop::RSpec::RepeatedExampleGroupBody do
     RUBY
   end
 
+  it 'skips `skip` and `pending` statements with arguments' do
+    expect_no_offenses(<<-RUBY)
+      describe '#load' do
+        skip 'storage feature needed'
+      end
+
+      describe '#save' do
+        skip 'storage feature needed'
+      end
+
+      describe '#get_foo' do
+        pending 'foo feature is broken'
+      end
+
+      describe '#set_foo' do
+        pending 'foo feature is broken'
+      end
+    RUBY
+  end
+
+  it 'registers offense correctly if `skip` and `pending` have block' do
+    expect_offense(<<-RUBY)
+      describe '#load' do
+      ^^^^^^^^^^^^^^^^^^^ Repeated describe block body on line(s) [5]
+        skip { cool_predicate_method }
+      end
+
+      describe '#save' do
+      ^^^^^^^^^^^^^^^^^^^ Repeated describe block body on line(s) [1]
+        skip { cool_predicate_method }
+      end
+
+      describe '#get_foo' do
+      ^^^^^^^^^^^^^^^^^^^^^^ Repeated describe block body on line(s) [13]
+        pending { cool_predicate_method }
+      end
+
+      describe '#set_foo' do
+      ^^^^^^^^^^^^^^^^^^^^^^ Repeated describe block body on line(s) [9]
+        pending { cool_predicate_method }
+      end
+    RUBY
+  end
+
   it 'registers offense correctly if example groups are separated' do
     expect_offense(<<-RUBY)
       describe 'repeated' do
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,8 +4,8 @@
 
 * Allow `RSpec/ContextWording` to accept multi-word prefixes. ([@hosamaly][])
 * Drop support for ruby 2.4. ([@bquorning][])
-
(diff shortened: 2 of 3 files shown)


spec/rubocop/cop/rspec/repeated_example_group_body_spec.rb
```diff
@@ -226,6 +226,50 @@
...
+        pending 'another reason'
+      end
+    RUBY
+  end
+
+  it 'registers offense correctly if `skip` and `pending` have block' do
+    expect_offense(<<-RUBY)
+      describe 'repeated' do
+      ^^^^^^^^^^^^^^^^^^^^^^ Repeated describe block body on line(s) [5]
+        skip { blah }
```
pirj, line 253: Is this a valid syntax? In a sense that the block seems to be ignored.

Might make sense to remove this example and only keep the one with the string argument?

pirj (approved):
Thanks!

Darhazer (approved):

spec/rubocop/cop/rspec/repeated_example_group_body_spec.rb
Tietew:
``` ruby
skip 'is skipped' do
  some_expect
end
# is same as
it 'is skipped' do
  skip
  some_expect
end
```

``` ruby
pending 'is pending' do
  some_expect
end
# is same as
it 'is pending' do
  pending
  some_expect
end
```

I added this test to represent explicitly that `skip` and `pending` with block cannot to be ignored.

pirj: Aha, you're right, `skip`/`pending` are [also example methods](https://github.com/rspec/rspec-core/blob/5d87c28bb828ba14b54a0d8cfddf044894009314/lib/rspec/core/example_group.rb#L195), and we [do account for that](https://github.com/rubocop-hq/rubocop-rspec/blob/300943ef5ca71d8708685c17c10f3d95b6509b04/lib/rubocop/rspec/language.rb#L87).

spec/rubocop/cop/rspec/repeated_example_group_body_spec.rb
```diff
@@ -226,6 +226,50 @@
...
+      end
+
+      describe 'repeated too' do
+      ^^^^^^^^^^^^^^^^^^^^^^^^^^ Repeated describe block body on line(s) [13]
+        pending { blah }
+      end
+
+      describe 'this is repeated too' do
+      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ Repeated describe block body on line(s) [9]
+        pending { blah }
```
pirj, line 268: What if several skipped examples only contain a comment, e.g.
```ruby
# FIXME: don't forget to write a spec for this
```
?

spec/rubocop/cop/rspec/repeated_example_group_body_spec.rb
```diff
@@ -226,6 +226,50 @@
...
 
+  it 'skips `skip` and `pending` statements with arguments' do
+    expect_no_offenses(<<-RUBY)
+      context 'processed' do
+        skip 'some reason'
+      end
+
+      context 'processed' do
+        skip 'some reason'
+      end
```
pirj, line 237: Now when I look at it, it seems that there's something wrong, and it's the repeated context title.
Do you mind to make them different?
The example with `describe '#load' do` isn't the best reference, it's a miss.

spec/rubocop/cop/rspec/repeated_example_group_body_spec.rb
Tietew: Comments are just ignored on compare due they are not in AST.
This PR didn't target empty example blocks.

spec/rubocop/cop/rspec/repeated_example_group_body_spec.rb
Tietew: I'll try update them to clarify use case.

bquorning:
@pirj Are you happy with this PR as it is? (We can resolve the Changelog conflict on merge).

bquorning (approved):

pirj (approved):
Thanks!

PR #1040, opened 2020-10-13
New Cop: `UnusedImplicitSubject`

marcandre:

* [x] Added tests.
* [x] Updated documentation.


marcandre:
Hmmm. I now realize it contradicts `RSpec/ImplicitSubject` and maybe it should supplement it.

I had read quickly that cop, but got confused by the fact that it deals with two things that aren't related in my mind:
- use of `it_is_expected` vs `expect(subject)`
- number of lines / expectactions / whatever

PR #1041, opened 2020-10-13
`Language::SelectorSet` tweaks

marcandre:
I wanted to access `[:subject, :subject!]` and was surprised to realize I couldn't.

This PR addresses that.

Commits are not directly related, but second one assumes first one. Let me know if I need to split them.

### DIFF
--- a/lib/rubocop/rspec/language.rb
+++ b/lib/rubocop/rspec/language.rb
@@ -7,7 +7,8 @@ module RuboCop
       # Set of method selectors
       class SelectorSet
         def initialize(selectors)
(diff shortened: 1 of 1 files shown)


pirj (approved):
I have no objections.
We've changed a lot in #956, do those changes need to be repeated there as well?

marcandre:
> I have no objections.
> We've changed a lot in #956, do those changes need to be repeated there as well?

No, I don't think so, `rspec_keywords` exposes those values.

Darhazer (approved):
No objections from my side

PR #1042, opened 2020-10-15
Fixes RSpec/ExpectChange to work with nested class names.

### LINKED ISSUE #1037: RSpec/ExpectChange doesn't work well with nested class
We use `RSpec/ExpectChange` with `EnforcedStyle: block`.

This rule works well with not nested class like below:

```
it 'should create new token' do
  expect { subject }.to change(Token, :count).by(1)
end
```

But when we use nested class like `User::Token` instead of `Token`,  this rule doesn't work.
It can be easily reproduced with spec/rubocop/cop/rspec/expect_change_spec.rb.

```
Failures:

  1) RuboCop::Cop::RSpec::ExpectChange with EnforcedStyle `block` flags implicit block expectation syntax with nested class
     Failure/Error: super
     
       Diff:
       @@ -1,5 +1,4 @@
                it do
                  expect(run).to change(User::Token, :count).by(1)
       -                         ^^^^^^^^^^^^^^^^^^^^^^^^^^^ Prefer `change { User::Token.count }`.
                end
       
     # ./spec/support/expect_offense.rb:13:in `expect_offense'
     # ./spec/rubocop/cop/rspec/expect_change_spec.rb:103:in `block (3 levels) in <top (required)>'
```

M-Yamashita01:
fixes #1037

When we use `RSpec/ExpectChange` with `EnforcedStyle: block`,
we expect RuboCop leaves comments to use `change { Token.count }` for all codes below:

1 `expect { subject }.to change(Token, :count).by(1)`
2 `expect { subject }.to change(::Token, :count).by(1)`
3 `expect { subject }.to change(User::Token, :count).by(1)`

However,  RuboCop raises an offence only for the first.

The AST of `expect { subject }.to change(User::Token, :count).by(1)` is like below

```
                (send
                  (block
                    (send nil :expect)
                    (args)
                    (send nil :subject)) :to
                  (send
                    (send nil :change
                      (const
                        (const nil :User) :Token)
                      (sym :count)) :by
                    (int 1))))))))
```

and RSpec::ExpectChange checks it with the Regex below:

```
                      (const
                        (const nil :User) :Token)
```

I fixed the `RSpec::ExpectChange` to match the AST of `change(User::Token, :count)` and added some tests about it. 

---

* [x] Added tests.
* [-] Updated documentation.

### DIFF
--- a/lib/rubocop/cop/rspec/expect_change.rb
+++ b/lib/rubocop/cop/rspec/expect_change.rb
@@ -39,7 +39,7 @@ module RuboCop
 
         # @!method expect_change_with_arguments(node)
         def_node_matcher :expect_change_with_arguments, <<-PATTERN
-          (send nil? :change $_ (sym $_))
+          (send nil? :change $_ ({sym str} $_))
         PATTERN
 
         # @!method expect_change_with_block(node)
@@ -47,7 +47,7 @@ module RuboCop
           (block
             (send nil? :change)
             (args)
-            (send ({const send} nil? $_) $_)
+            (send $_ $_)
           )
         PATTERN
 
@@ -67,9 +67,9 @@ module RuboCop
           return unless style == :method_call
 
           expect_change_with_block(node) do |receiver, message|
-            msg = format(MSG_BLOCK, obj: receiver, attr: message)
+            msg = format(MSG_BLOCK, obj: receiver.source, attr: message)
             add_offense(node, message: msg) do |corrector|
-              replacement = "change(#{receiver}, :#{message})"
+              replacement = "change(#{receiver.source}, :#{message})"
               corrector.replace(node, replacement)
             end
           end
--- a/spec/rubocop/cop/rspec/expect_change_spec.rb
+++ b/spec/rubocop/cop/rspec/expect_change_spec.rb
@@ -23,7 +23,7 @@ RSpec.describe RuboCop::Cop::RSpec::ExpectChange do
       RUBY
     end
 
-    it 'ignores blocks that cannot be converted to obj/attribute pair' do
+    it 'ignores blocks when the method is called with arguments' do
       expect_no_offenses(<<-RUBY)
         it do
           expect { run }.to change { User.sum(:points) }
@@ -31,7 +31,7 @@ RSpec.describe RuboCop::Cop::RSpec::ExpectChange do
       RUBY
     end
 
-    it 'ignores change method of object that happens to receive a block' do
+    it 'ignores when not an expectation' do
       expect_no_offenses(<<-RUBY)
         it do
           Record.change { User.count }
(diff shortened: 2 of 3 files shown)


spec/rubocop/cop/rspec/expect_change_spec.rb
```diff
@@ -67,17 +67,47 @@
...
+        end
+      RUBY
+    end
+
     it 'flags change matcher when receiver is a variable' do
       expect_offense(<<-RUBY)
         it do
-          expect { run }.to change(User, :count)
-                            ^^^^^^^^^^^^^^^^^^^^ Prefer `change { User.count }`.
+          expect { run }.to change(user, :count)
```
pirj, line 103: I'm not sure I completely understand the reason behind this change `User` -> `user`.
`User.count` resembles Rails specs.

pirj: It's very interesting, actually.

It seems that currently we only match consts correctly.
While your example is actually totally legit, e.g.
```ruby
expect { file.upload! }.to change(user.uploads, :count)
```

Do you mind to add something like this as a separate example for both styles?

lib/rubocop/cop/rspec/expect_change.rb
```diff
@@ -64,9 +66,11 @@ def on_block(node)
           return unless style == :method_call
```
pirj, line 66: Since you've touched handling of both styles, `block` and `method_call`, please consider adding examples under `context 'with EnforcedStyle `method_call`' do` as well.

lib/rubocop/cop/rspec/expect_change.rb
```diff
@@ -64,9 +66,11 @@ def on_block(node)
...
 
           expect_change_with_block(node) do |receiver, message|
-            msg = format(MSG_BLOCK, obj: receiver, attr: message)
+            obj = receiver.is_a?(RuboCop::AST::Node) ? receiver.source : receiver
```
pirj, line 69: I have an impression that something is horribly wrong with the node pattern that I've drafted in #1037.
`receiver.is_a?(RuboCop::AST::Node)` shouldn't be needed, and I would expect `.source` to be callable on any match.

Are you up to play with the node pattern a bit more?

Please don't hesitate to ping me if you stumble some weird behaviour.

pirj (commented):
Sorry it took me so long to review. Thanks for your effort!

spec/rubocop/cop/rspec/expect_change_spec.rb
YusukeIwaki: Thank you for your comments, (M-Yamashita is my colleague :) )

When `RSpec/ExpectChange` was introduced, this spec was `change(user, :count)` https://github.com/rubocop-hq/rubocop-rspec/commit/fc57a5cb7537b21e9e8457f419a4cdfbb80aa1e5

And this commit https://github.com/rubocop-hq/rubocop-rspec/commit/971fc7d38d9f4b5ea87b179fdf8c5d0b9c4b85b0 changed `user` to `User`.
However `change(User, :count)` is already tested in another spec ( `it 'flags change matcher without block'` ).
So we thought this is unexpected change and just reverted it (User to user).

> It seems that currently we only match consts correctly.
> While your example is actually totally legit, e.g.
>
> expect { file.upload! }.to change(user.uploads, :count)
> Do you mind to add something like this as a separate example for both styles?

We didn't noticed this pattern, but it's really interesting.
We will try it.

lib/rubocop/cop/rspec/expect_change.rb
YusukeIwaki: > I have an impression that something is horribly wrong with the node pattern that I've drafted in #1037.
receiver.is_a?(RuboCop::AST::Node) shouldn't be needed, and I would expect .source to be callable on any match.

I completely agree with it, and we will study about node pattern with [this](https://docs.rubocop.org/rubocop-ast/node_pattern.html)

pirj:
Do you need any help with this, @YusukeIwaki ?

spec/rubocop/cop/rspec/expect_change_spec.rb
M-Yamashita01: I'm sorry for the late reply.
We added examples for method_call and block styles.

lib/rubocop/cop/rspec/expect_change.rb
M-Yamashita01: We added examples under context with EnforcedStyle method_call.

lib/rubocop/cop/rspec/expect_change.rb
M-Yamashita01: We studied about node pattern and fixed the implementation.

pirj (approved):
Thank you!
I've made a couple of small corrections:
 - now it detects `::User.count`
 - it detects a string message `change(User, 'count')`
 - removed a couple of old duplicated examples

bquorning (approved):

PR #1043, opened 2020-10-17
Add Example Request to Examples Module

smridge:
I started using https://github.com/zipmark/rspec_api_documentation . I understand since this is not part of any core rspec gems, it is entirely debatable whether this should be added or not. The easy option is to exclude the cop. However, I still find it to be a valid cop in keeping spec files clean.

As explained [here](https://github.com/zipmark/rspec_api_documentation), `example` is used as `it`. When using `example`, `do_request` needs to be added within the block. However, there is a more concise method of using [example_request](https://github.com/zipmark/rspec_api_documentation). Unfortunately, this upsets rubo 😞 

Thoughts from anyone would be great. The trend of auto generating docs is certainly a thing, I would be curious on the best way to solve for this.

---

* [ ] Added tests.
* [ ] Updated documentation.

### DIFF
--- a/lib/rubocop/rspec/language.rb
+++ b/lib/rubocop/rspec/language.rb
@@ -93,8 +93,9 @@ module RuboCop
         FOCUSED  = SelectorSet.new(%i[fit fspecify fexample fscenario focus])
         SKIPPED  = SelectorSet.new(%i[xit xspecify xexample xscenario skip])
         PENDING  = SelectorSet.new(%i[pending])
+        API_DOC  = SelectorSet.new(%i[example_request])
 
-        ALL = EXAMPLES + FOCUSED + SKIPPED + PENDING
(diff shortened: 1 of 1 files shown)


pirj:
Not shooting your idea down right away.
For the time being, I suggest you monkey patching this.

We plan to release `rubocop-rspec` 2.0 when `rubocop` releases 1.0 real soon, and it will include #956. It solves a number of similar problems with other gems that extend RSpec syntax.

In a nutshell, all DSL will be configurable, so you'll be able to inject DSL extensions via `.rubocop.yml`. And so could the extensions themselves, so if you opened a pull request to `rspec_api_documentation` to configure their RSpec aliases in `config/default.yml`, and `require:` it from your project-local `.rubocop.yml`, will make `rubocop-rspec` aware of those aliases ⚡

smridge:
Beautiful! I’m excited for this! Thank you for the detailed explanation!

PR #1047, opened 2020-10-20
Relaxes rubocop-ast dependency locking

PhilCoggins:
Relaxes rubocop-ast dependency lock from `~> 0.7.1` to `~> 0.7`, fixes https://github.com/rubocop-hq/rubocop-rspec/issues/1046

---

* [ ] Added tests.
* [ ] Updated documentation.

### DIFF
--- a/rubocop-rspec.gemspec
+++ b/rubocop-rspec.gemspec
@@ -37,7 +37,7 @@ Gem::Specification.new do |spec|
   }
 
   spec.add_runtime_dependency 'rubocop', '~> 0.87'
-  spec.add_runtime_dependency 'rubocop-ast', '~> 0.7.1'
+  spec.add_runtime_dependency 'rubocop-ast', '>= 0.7.1'
 
   spec.add_development_dependency 'rack'
   spec.add_development_dependency 'rake'


rubocop-rspec.gemspec
```diff
@@ -37,7 +37,7 @@ Gem::Specification.new do |spec|
   }
 
   spec.add_runtime_dependency 'rubocop', '~> 0.87'
-  spec.add_runtime_dependency 'rubocop-ast', '~> 0.7.1'
+  spec.add_runtime_dependency 'rubocop-ast', '~> 0.7'
```
pirj, line 40: There was a [specific reason to use 0.7.1 as opposed to 0.7.0](https://github.com/rubocop-hq/rubocop-rspec/pull/1036#issuecomment-701258609). But I guess `~> 0.7` is good either way, and `bundler` will install 0.7.1+ unless 0.7.0 is already installed.

I hope this explicit dependency specification statement will go away when RuboCop and `rubocop-ast` hit 1.0, and we hit 2.0.

pirj (approved):
Thanks!

pirj:
@bquorning Any chance for a minor release?

rubocop-rspec.gemspec
pirj: After giving it another thought, let's write it as `>= 0.7.1`, since 0.7.1 released [this](https://github.com/rubocop-hq/rubocop-ast/pull/127) fix for JRuby.

PhilCoggins:
This seems to fail with rubocop-ast 0.8.0, probably want to hold off on merging until that gets figured out. https://app.circleci.com/pipelines/github/rubocop-hq/rubocop-rspec/579/workflows/7af23933-9e4f-434b-9c88-5992c5f1f60c/jobs/17591

bquorning:
> This seems to fail with rubocop-ast 0.8.0

It only fails against edge rubocop (they’re making ready for v1.0). The other specs pass, also when using rubocop-ast 0.8.0 – e.g. https://app.circleci.com/pipelines/github/rubocop-hq/rubocop-rspec/580/workflows/2d22b1ca-ade8-4e41-a49d-50c80cc3a608/jobs/17602

Issue #1051, opened 2020-10-21
Make Rubocop-RSpec compatible with Rubocop v1.0


bquorning:
RuboCop v1.0 has been released, and RuboCop-RSpec is not compatible. It is expected that the compatible version will be released as RuboCop-RSpec v2.0.0, which will also contain a number of other major changes.

This milestone tracks the work that needs to be done: https://github.com/rubocop-hq/rubocop-rspec/milestone/1

I’d like to tackle this in two steps, one after the other:

1. Compatibility with new RuboCop
2. Code removal and other non-backwards compatible changes.

Please help me identify which PRs in the milestone belong to category 1 and which belong to 2.

I would like to test things out in another branch than `master`. Which, given that all current pull requests are against `master` means that I must merge and test locally. Why? I would like to have some certainty about the direction before we start merging code *and* I would like to keep the master branch clean in case we need to push a bugfix release before we’re ready to release v2.0.

bquorning:
Hmm, actually I think (1) may just be https://github.com/rubocop-hq/rubocop-rspec/pull/1019 and a couple of fixes to default_config_spec.rb.

pirj:
That is very reasonable.
You may actually change the target branch of pull requests for 2.0 to e.g. `release-2.0`
![image](https://user-images.githubusercontent.com/6916/96783570-365ed800-13f6-11eb-9582-5d0bf5576e36.png)

> Hmm, actually I think (1) may just be #1019 and a couple of fixes to default_config_spec.rb.

I think this is correct.

bquorning:
> You may actually change the target branch of pull requests for 2.0 to e.g. `release-2.0`

Ooh nice, I didn’t know that. I’ll do that for our open pull requests on the 2.0 milestone.

bquorning:
Actually, after merging #1019 we could release a v1.99.0 that would be compatible with RuboCop 1.0, while we work on the remaining [PRs in the 2.0 milestone](https://github.com/rubocop-hq/rubocop-rspec/milestone/1).

jesperronn:
> Actually, after merging #1019 we could release a v1.99.0 that would be compatible with RuboCop 1.0, while we work on the remaining [PRs in the 2.0 milestone](https://github.com/rubocop-hq/rubocop-rspec/milestone/1).

Considering semver versioning, I don't think it makes sense to throw in a high `1.x` number. 

The two options I see for version numbering is:

[...]

PR #1052, opened 2020-10-21
Allow rubocop-rspec to install with rubocop 1.0.0

cheister:
Fix for
```
      rubocop-rspec (= 1.44.1) was resolved to 1.44.1, which depends on
        rubocop (~> 0.87)
```
when trying to use it with rubocop 1.0.0


bquorning:
Thanks you for the PR, but we plan to add a few breaking changes along with the RuboCop v1.0 compatibility. See #1051.

PR #1056, opened 2020-10-22
Add mixins folder

bquorning:
Closes #942 

I rebased @mockdeep’s pull request #942 but found that I cannot push to that branch. So, I opened this new pull request.

Original description:

This moves several files to a new `mixins/` folder. I was going to add
some new macros, per [this comment][1], but thought it might clutter
things up a bit to put them in the root folder. I thought of adding a
`macros/` folder, but decided to emulate the [rubocop source][2]
instead.

[1]: https://github.com/rubocop-hq/rubocop-rspec/pull/934#issuecomment-643813427
[2]: https://github.com/rubocop-hq/rubocop/tree/master/lib/rubocop/cop/mixin

---

* [-] Added tests.
* [-] Updated documentation.


lib/rubocop/cop/rspec/mixin/top_level_describe.rb
```diff
@@ -0,0 +1,54 @@
+# frozen_string_literal: true
+
+module RuboCop
+  module Cop
+    module RSpec
+      # Helper methods for top level describe cops
+      module TopLevelDescribe
```
pirj, line 7: I believe we've got rid of this one.

pirj (approved):
All looks good, except for the retired `top_level_describe`.

lib/rubocop/cop/rspec/mixin/top_level_describe.rb
bquorning: You’re absolutely right. I remember thinking about it during the rebase, but I must have forgot. Let’s hope it’s the only thing I forgot 😱

bquorning:
@pirj Please merge at will. #956 will need to be rebased afterwards.

PR #1059, opened 2020-10-23
Use config context for all cop specs

bquorning:
* Add shared context for all cop specs.
* Remove redundant `config` metadata for cop specs.
* Remove redundant `cop` subjects from cop specs.

Extracted from #956.

---

* [-] Added tests.
* [-] Updated documentation.


pirj (approved):
Thanks for extracting! This is for sure completely unrelated to aliases.

bquorning:
🎉 Thanks

PR #1063, opened 2020-10-27
Remove relevance detection code

pirj:
We have had problems with it:
 - it was slow and memory-hungry
 - it uses a non-standard extension of RuboCop's config (`Patterns`), while a comparable `Include` option exists
 - department-specific (e.g. FactoryBot) config was ignored
 - for the same department (i.e. FactoryBot) files to inspect are different for different cops

Close #967
Close #1062
Fixes #1044

It's all interconnected in this PR, here's the path I had to go through:

1. Remove the code that checks file relevance
2. Replace Patterns with Include
3. Adjust config/default.yml
4. Adjust specs, specifying the file for other departments
5. Tweak config shared example to respect department settings
6. Adjust config formatter
7. Fix spec that failed to detect offences in an unrelated file

### TODOs

- [x] extract 7 into a separate PR (extracted to #1064)
- [x] adjust #1013 to mention `Patterns` -> `Include` migration (added a TODO there)

---

Before submitting the PR make sure the following are checked:

* [x] Feature branch is up-to-date with `master` (if not - rebase it).
* [x] Squashed related commits together.
* [x] Added tests.
* [x] Updated documentation.
* [x] Added an entry to the `CHANGELOG.md` if the new code introduces user-observable changes.
* [x] The build (`bundle exec rake`) passes (be sure to run this locally, since it may produce updated documentation that you will need to commit).

If you have modified an existing cop's configuration options:

* [x] Set `VersionChanged` in `config/default.yml` to the next major version.

PR #1064, opened 2020-10-27
Refactor a spec

pirj:
'autocorrect' shared example group does not pass a correct file name to
the cop, and if this cop is tweaked to only inspect factories, this spec
fails because the default file name/path is skipped, since it's not a
factory path.

#1063 depends on this

---

* [x] Added tests.
* [-] Updated documentation.

### DIFF
--- a/.rubocop_todo.yml
+++ b/.rubocop_todo.yml
@@ -10,3 +10,7 @@
 # Configuration parameters: CountComments, CountAsOne.
 Metrics/ClassLength:
   Max: 106
+
+RSpec/ExampleLength:
+  Exclude:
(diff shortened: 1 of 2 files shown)


.rubocop_todo.yml
```diff
@@ -10,3 +10,7 @@
 # Configuration parameters: CountComments, CountAsOne.
 Metrics/ClassLength:
   Max: 106
+
+RSpec/ExampleLength:
+  Exclude:
+    - spec/rubocop/cop/rspec/factory_bot/attribute_defined_statically_spec.rb
```
pirj, line 16: It's ~ [45/30].

bquorning (approved):

PR #1065, opened 2020-10-28
Stop using newly added cops

pirj:
```
lib/rubocop/cop/rspec/multiple_expectations.rb:73:32: W: Lint/ToEnumArguments: Ensure you correctly provided all the arguments.
              expectations_count = to_enum(:find_expectation, node).count
                                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```
([source](https://app.circleci.com/pipelines/github/rubocop-hq/rubocop-rspec/634/workflows/df5cf9d4-b931-4fd2-af3f-36f4e74276ab/jobs/18146))

We'd still love to use OUR new cops while working on our codebase, let's see how this option works when set under RSpec namespace when we introduce the first pending cop.

---

* [-] Added tests.
* [-] Updated documentation.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -8,7 +8,7 @@ require:
 AllCops:
   DisplayCopNames: true
   TargetRubyVersion: 2.4
-  NewCops: enable
+  NewCops: disable
   Exclude:
     - 'vendor/**/*'
     - 'spec/fixtures/**/*'


pirj:
@bquorning Merge at will.

bquorning (approved):

bquorning:
> Lint/ToEnumArguments: Ensure you correctly provided all the arguments.

I ran into this again recently. I think it’s actually a bug – see also https://github.com/rubocop/rubocop/pull/8941#issuecomment-1159797356

bquorning:
> We'd still love to use OUR new cops while working on our codebase, let's see how this option works when set under RSpec namespace when we introduce the first pending cop.

It doesn’t work.

```
Warning: RSpec does not support NewCops parameter.

Supported parameters are:

  - Enabled
  - StyleGuideBaseURL
  - Include
  - Language
```

pirj:
What if we define `NewCops: pending` in `config/default.yml` under `RSpec` like [`rubocop` does in its `config/default.yml`](https://github.com/rubocop/rubocop/blob/dfc6e871f2fdc1d0adeeb9139f67ea0ac163857f/config/default.yml#L111)?

This could fix the error, but I'm still uncertain if it would work as intended.

Another concern is if changing the root `NewCops` in project's `.rubocop.yml` would affect `RSpec/NewCops`. It feels that it should, but again I'm uncertain.

bquorning:
> What if we define `NewCops: pending` in `config/default.yml` under `RSpec` like [`rubocop` does in its `config/default.yml`](https://github.com/rubocop/rubocop/blob/dfc6e871f2fdc1d0adeeb9139f67ea0ac163857f/config/default.yml#L111)?

Nope, that doesn’t work :sad_panda:

PR #1071, opened 2020-11-03
Enable pending cop(s)

pirj:

* [-] Added tests.
* [-] Updated documentation.


bquorning (approved):
Should we mark this as a breaking change in the changelog? (RuboCop did that on their v1.0 release)

pirj:
Our changelog is less sophisticated, we don't have fix/change/feature distinction.
I'm not sure if there is any change in what we've merged to 2.0 so far that is not a breaking change :D
Added an entry to our CHANGELOG 👍

PR #1076, opened 2020-11-06
Fix backtick with a following quote

pirj:
It seems that a quote follwoing the backtick doesn't work well:

image

Looks better now, at least according to [GH's render](https://github.com/rubocop-hq/rubocop-rspec/blob/c04a22d56c998109ee6d1be6b7c16150c2fd52b1/docs/modules/ROOT/pages/upgrade_to_version_2.adoc#use-the-rubocop-standard-include-option-to-filter-inspected-files):
image

[Old one](https://github.com/rubocop-hq/rubocop-rspec/blob/master/docs/modules/ROOT/pages/upgrade_to_version_2.adoc#use-the-rubocop-standard-include-option-to-filter-inspected-files) for comparison:

image

* [x] Updated documentation.

### DIFF
--- a/docs/modules/ROOT/pages/upgrade_to_version_2.adoc
+++ b/docs/modules/ROOT/pages/upgrade_to_version_2.adoc
@@ -91,7 +91,7 @@ RSpec/FactoryBot:
     - property/factories/**/*.rb
 ----
 
-NOTE: Please keep in mind that `Include`'s merge mode is set to override the default settings, so if you intend to add a path while keeping the default paths, you should include the default `Include` paths in your configuration.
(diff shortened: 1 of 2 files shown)


bquorning (changes requested):
What a weird little bug in asciidoc.

https://asciidoctor.org/docs/asciidoc-writers-guide/#replacements says that `'` is converted to the “Typographic apostrophe” `’`. So I think let’s just use that glyph directly and spare asciidoc that work :-)

pirj:
Nice catch, didn't know that.

bquorning (approved):

bquorning:
Hmm, I am not sure how to update the documentation after it has been published 😕

PR #1078, opened 2020-11-08
[UPDATE] gems' version dependency

sfeuga:
In a `Gemfile` if you use Rubocop 1.2, you face an issue to install this gem:
```
rubocop-rspec (~> 2.0) was resolved to 2.0.0, which depends on
      rubocop (~> 1.0)
```

According to [the gem documentation](https://guides.rubygems.org/specification-reference/#required_ruby_version=) I propose to update the version of rubocop in the `rubocop-rspec.gemspec` file.

Tested on `Fedora 32` with `ruby --version ruby 2.7.2p137 (2020-10-01 revision 5445e04352) [x86_64-linux]`.

---

* [x] Added tests. __(No new test needed)__
* [x] Updated documentation. __(No documentation update needed)__

### DIFF
--- a/rubocop-rspec.gemspec
+++ b/rubocop-rspec.gemspec
@@ -36,13 +36,13 @@ Gem::Specification.new do |spec|
     'documentation_uri' => 'https://docs.rubocop.org/rubocop-rspec/'
   }
 
-  spec.add_runtime_dependency 'rubocop', '~> 1.0'
+  spec.add_runtime_dependency 'rubocop', '>= 1.0'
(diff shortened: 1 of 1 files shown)


bquorning:
The existing gemspec allows installing rubocop-rspec 2.0.0 and rubocop 1.2.0 together.

Do you get an error when installing? Could you paste a bit more of the output you get when calling `bundle install`?

sfeuga:
Hi @bquorning Just test it again and it was an issue with my rvm gemset sorry to bother you :facepalm: 
I think we can close this pr.

PR #1079, opened 2020-11-08
Allow us to use `fit`, `fdescribe`, `:focus` etc.

bquorning:
I find it useful sometimes to just slap an `f` in front of `it` or `describe` when I want to run just a few specs.

---

* [-] Added tests.
* [-] Updated documentation.


pirj (approved):
Cool!

PR #1080, opened 2020-11-08
Clean up our spec helper

bquorning:
- Use the same block variable name (`meta`) both times `define_derived_metadata` is called.
- Use `each` instead of `map` when not using the returned array anyway.
- RSpec will add `/spec` to the load path, so we don't need to.

### DIFF
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -1,7 +1,6 @@
 # frozen_string_literal: true
 
 require 'rubocop'
-
 require 'rubocop/rspec/support'
 
 if ENV['COVERAGE'] == 'true'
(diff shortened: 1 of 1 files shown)


spec/spec_helper.rb
```diff
@@ -14,16 +13,17 @@ module SpecHelper
 end
 
 spec_helper_glob = File.expand_path('{support,shared}/*.rb', __dir__)
-Dir.glob(spec_helper_glob).map(&method(:require))
+Dir.glob(spec_helper_glob).sort.each(&method(:require))
```
bquorning, line 16: `Lint/NonDeterministicRequireOrder` complains about `each`, but didn’t notice the `map`…

pirj: I guess our `RSpec/IteratedExpectation` would also miss `map`.
But we have some code in [`RSpec/EmptyExampleGroup`](https://github.com/rubocop-hq/rubocop-rspec/blob/c3fa5ba4ccd5b3cffd58c55ca7fdde63614f2501/lib/rubocop/cop/rspec/empty_example_group.rb#L93) that is sophisticated enough to detect different iterators]((https://github.com/rubocop-hq/rubocop-rspec/blob/c3fa5ba4ccd5b3cffd58c55ca7fdde63614f2501/spec/rubocop/cop/rspec/empty_example_group_spec.rb#L171).

spec/spec_helper.rb
```diff
@@ -14,16 +13,17 @@ module SpecHelper
...
 RSpec.configure do |config|
+  # Set metadata so smoke tests are run on all cop specs
   config.define_derived_metadata(file_path: %r{/spec/rubocop/cop/}) do |meta|
     meta[:type] = :cop_spec
   end
 
   # Include config shared context for all cop specs
-  config.define_derived_metadata(type: :cop_spec) do |metadata|
-    metadata[:config] = true
+  config.define_derived_metadata(type: :cop_spec) do |meta|
```
pirj, line 25: I had doubts, but it matches [a canonical example](https://relishapp.com/rspec/rspec-core/v/3-9/docs/expectation-framework-integration/aggregating-failures#enable-failure-aggregation-globally-using-%60define-derived-metadata%60) :+1:

pirj (approved):
Thanks!

PR #1081, opened 2020-11-08
Use `expect_correction` more

bquorning:
I think we should use `expect_correction` more and soon completely stop using `include_examples 'autocorrect'`.

PR #1082, opened 2020-11-09
Less autocorrect

bquorning:
Follow-up to #1081. Use `expect_correction` more, and use `include_examples 'autocorrect'` less.

---

* [x] Added tests.
* [-] Updated documentation.

### DIFF
--- a/spec/rubocop/cop/rspec/factory_bot/create_list_spec.rb
+++ b/spec/rubocop/cop/rspec/factory_bot/create_list_spec.rb
@@ -9,52 +9,64 @@ RSpec.describe RuboCop::Cop::RSpec::FactoryBot::CreateList do
     let(:enforced_style) { :create_list }
 
     it 'flags usage of n.times with no arguments' do
-      expect_offense(<<-RUBY)
+      expect_offense(<<~RUBY)
         3.times { create :user }
         ^^^^^^^ Prefer create_list.
+      RUBY
+
+      expect_correction(<<~RUBY)
+        create_list :user, 3
       RUBY
     end
 
     it 'flags usage of n.times when FactoryGirl.create is used' do
-      expect_offense(<<-RUBY)
+      expect_offense(<<~RUBY)
         3.times { FactoryGirl.create :user }
         ^^^^^^^ Prefer create_list.
+      RUBY
+
+      expect_correction(<<~RUBY)
+        FactoryGirl.create_list :user, 3
       RUBY
     end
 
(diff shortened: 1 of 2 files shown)


spec/rubocop/cop/rspec/factory_bot/create_list_spec.rb
```diff
@@ -65,105 +77,135 @@
...
-      expect_offense(<<-RUBY)
-       FactoryGirl.create_list(:user, 3) { |user| user.points = rand(1000) }
-                   ^^^^^^^^^^^ Prefer 3.times.
+      expect_offense(<<~RUBY)
+        FactoryGirl.create_list(:user, 3) { |user| user.points = rand(1000) }
+                    ^^^^^^^^^^^ Prefer 3.times.
+      RUBY
+
+      expect_correction(<<~RUBY)
+        3.times { FactoryGirl.create(:user) } { |user| user.points = rand(1000) }
```
pirj, line 201: Unrelated to this change, but it turns out this is a bad example - models are persisted before they are passed to the block, so the change to `points` won't be persisted. https://github.com/thoughtbot/factory_bot/issues/1444
We eventually may want to adjust this to `build`, `user.update` or add `user.save!`

spec/rubocop/cop/rspec/hook_argument_spec.rb
```diff
@@ -62,6 +49,13 @@
         config.after(:each)  { true }
         ^^^^^^^^^^^^^^^^^^^ Omit the default `:each` argument for RSpec hooks.
       RUBY
+
+      expect_correction(<<-RUBY)
+        before { true }
+        after  { true }
+        around { true }
+        config.after  { true }
```
pirj, line 57: Cosmetic: extra space, in the "bad" example as well.

spec/rubocop/cop/rspec/hook_argument_spec.rb
```diff
@@ -152,20 +168,35 @@
...
+
+      expect_correction(<<-RUBY)
+        before(:example) { true }
+        after(:example)  { true }
+        around(:example) { true }
+        config.before(:example) { true }
+      RUBY
     end
 
-    it 'does not flag hooks without default scopes' do
```
pirj, line 157: 👍

spec/rubocop/cop/rspec/hook_argument_spec.rb
bquorning: Fixed

PR #1083, opened 2020-11-09
Lint our YAML files, using yamllint

bquorning:
Thanks to RuboCop, our Ruby files are so fine and consistently formatted. Our YAML files on the other hand…

So I thought we might want to use [yamllint](https://yamllint.readthedocs.io/en/stable/) to lint these files. Custom rules are in the .yamllint file, and you can see the enforced rules in practice on the other changed files in this PR.

I also thought we might want to try out GitHub Actions for this task, instead of the usual CircleCI. We can always move it later, if need be.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -1,9 +1,9 @@
 inherit_from: .rubocop_todo.yml
 
 require:
- - rubocop-performance
- - rubocop-rspec
- - rubocop/cop/internal_affairs
+  - rubocop-performance
+  - rubocop-rspec
+  - rubocop/cop/internal_affairs
 
 AllCops:
   DisplayCopNames: true
--- a/lib/rubocop/rspec/config_formatter.rb
+++ b/lib/rubocop/rspec/config_formatter.rb
@@ -15,7 +15,9 @@ module RuboCop
       end
 
       def dump
-        YAML.dump(unified_config).gsub(EXTENSION_ROOT_DEPARTMENT, "\n\\1")
+        YAML.dump(unified_config)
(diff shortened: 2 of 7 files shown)


.rubocop.yml
```diff
@@ -20,7 +20,7 @@ Layout/HashAlignment:
   Enabled: false
 
 Layout/LineLength:
-  Max: 80 # default: 120
+  Max: 80  # default: 120
```
Darhazer, line 23: requiring 2 spaces before the commend is kinda weird

Darhazer (approved):
This is great :raised_hands:

.rubocop.yml
bquorning: Yeah, I’ll remove that rule.

lib/rubocop/rspec/config_formatter.rb
```diff
@@ -15,7 +15,9 @@ def initialize(config, descriptions)
       end
 
       def dump
-        YAML.dump(unified_config).gsub(EXTENSION_ROOT_DEPARTMENT, "\n\\1")
+        YAML.dump(unified_config)
+          .gsub(EXTENSION_ROOT_DEPARTMENT, "\n\\1")
+          .gsub(/^(\s+)- /, '\1  - ')
```
bquorning, line 20: Add two extra spaces of indentation to every line starting with `- ` or `<some spaces>- `.

pirj: Aha, this is where the flat list indentation was coming from! 💯

pirj (approved):
YAML looks so much better now!

PR #1084, opened 2020-11-09
Move from CircleCI to GitHub Actions

bquorning:
It seems like GHA is slightly faster than CircleCI (https://github.com/rubocop-hq/rubocop-rspec/actions/runs/384097573 vs https://app.circleci.com/pipelines/github/rubocop-hq/rubocop-rspec/772/workflows/590d8153-c0ca-4d87-8c5d-d6ba132e4f40). Being a built-in tool of GitHub, it’s easier to sign in and manage permissions.

We cannot move the Code Climate integration away from CircleCi because I forgot who has the keys to sign in and get a new API key 🤦


.github/workflows/main.yml
```diff
@@ -0,0 +1,89 @@
...
+      - run: bundle exec rake internal_investigation
+
+  jruby:
+    runs-on: ubuntu-latest
+    name: JRuby
+    steps:
+      - uses: actions/checkout@v2
+      - uses: ruby/setup-ruby@v1
+        with:
+          ruby-version: jruby
```
pirj, line 67: This looks very similar to the `rubocop` step, is it possible to add `jruby` there?
```
      matrix:
        ruby: ['2.4', '2.5', '2.6', '2.7', 'jruby']
```

.github/workflows/main.yml
```diff
@@ -0,0 +1,89 @@
...
+      - run: bundle install
+      - run: bundle exec rake internal_investigation
```
pirj, line 69: I think previously we were running specs for JRuby as well:
```
rake internal_investigation spec
```

pirj:
Circle stuck for the second time (that I've noticed) in https://github.com/rubocop-hq/rubocop-rspec/pull/1133
So now is high time for the move.

It seemed to me that we have two (different?) CodeClimate checks:
image

In either case, I believe we can survive without CodeClimate. At least having a CI that works (rather than [pretends to work](https://status.circleci.com/) is preferable.

pirj (approved):

pirj:
Fixed the internal investigation in https://github.com/rubocop-hq/rubocop-rspec/pull/1134

.github/workflows/main.yml
```diff
@@ -0,0 +1,51 @@
...
+  main:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        ruby:
+          - 2.4
+          - 2.5
+          - 2.6
+          - 2.7
+          - 3.0
```
bquorning, line 26:
```suggestion
          - "3.0"
```
Don't let YAML be confused about `3` and `3.0`.

bquorning:
@Darhazer What do you think – is it OK to get rid of Code Climate?

Darhazer:
@bquorning perhaps we can register a new codeclimate under the new organization :joy: 

But I'm fine with moving away from it as well

pirj:
I don't think code climate ever has brought any value in the past to `rubocop-rspec` specifically.

bquorning:
How do you feel about me removing the simplecov gem and `rake coverage` too, then?

pirj:
> removing the `simplecov` gem and `rake coverage` too

I suprised me, it seems `rake coverage` is running as part of our default Rake task, isn't it?
```ruby
task default: %i[build_config coverage
                 internal_investigation
```

The current coverage is
```
All Files (99.69% covered ...
```
with a few outliers:
```
lib/rubocop/cop/rspec/align_left_let_brace.rb | 94.12 %
```

At the same time, `SimpleCov.minimum_coverage` is not set in our `spec_helper.rb`, and I can't remember opening `coverage/index.html`.

Is it causing any problems?

Wondering what it takes to raise our coverage to 100% and set it as a hard constraint.
image

Also wondering why it counts coverage of spec files 😕

bquorning:
> is it OK to get rid of Code Climate?

>> But I'm fine with moving away from it as well

>> I don't think code climate ever has brought any value in the past to `rubocop-rspec` specifically.

Sorry for the delay. I’ll rebase this PR and merge on green.

pirj:
Fantastic, thanks!

bquorning:
All open PRs need to be rebased (or merge origin/master) for CI to pass now.

PR #1088, opened 2020-11-11
Make CreateList cop not safe

### LINKED ISSUE #1087: CreateList is not a safe cop
### Description

rubocop autocorrect this :
```rb
10.times { create(:user, age: rand(10..0)) }
```
to
```
create_list(:user, 10, age: rand(10..0))
```

both obviously do not have the same output as in the corrected version rand is only evaluated once and will attribute its result to all users.

We can "prevent" this behaviour by using the `n` like 10.times { |n| ...}` but this is just a corner case of multiple possibilities where `x.times` is just a legit call

### Expected behavior

this should not be a safe cop

ngouy:
Fix #1087 
Mark CreateList as not safe cop.

---

* [ ] Added tests.
* [x] Updated documentation.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Master (Unreleased)
 
+* Mark `RSpec/FactoryBot/CreateList` as not safe cop. ([@ngouy][])
+
 ## 2.0.0 (2020-11-06)
 
 * Remove deprecated class `::RuboCop::Cop::RSpec::Cop`. ([@bquorning][])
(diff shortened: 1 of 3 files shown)


pirj (changes requested):
Even though this cop **is** correction-unsafe, I'm more inclined to fixing this case rather than marking its otherwise useful auto-correction as unsafe.

ngouy:
How is it possible to detect when content is something that needs to be called x times vs not?

detecting the usage of the block variable is what is currently implemented, but there are tons of possibility other than the one I provider above

if you use `rand`, or `[].sample`, or even a method `next_position` that return a different output at each call

What is your plan match to detect successfully every single possibility?

PR #1094, opened 2020-11-14
Remove seemingly unnecessary spec config file

bquorning:
This file was added in #1063, but the specs seem to run fine without it. I confirmed this using RuboCop v1.0.0, v1.1.0, v1.2.0 and v1.3.0. Do we still need this file @pirj?

---

* [-] Added tests.
* [-] Updated documentation.


pirj (approved):
I'm surprised that everything specs still run fine, as I've spent quite some tome crafting this. But I'm very happy to see this go!
Thanks!

bquorning:
> I've spent quite some time crafting this.

I know you have, and I’m as surprised as you are. Do you recall why you found it necessary at the time? Is it just a side-effect of the RuboCop 1.0 changes and the RuboCop RSpec 2.0 changes that leads to it being possible to remove this file now?

pirj:
> Do you recall why you found it necessary at the time?

Not really. Looking at PR comments and code comments gives me no idea either. I'll learn from this.

PR #1096, opened 2020-11-15
Cop for consistent usage of parens in FactoryBot calls

### LINKED ISSUE #416: Cop idea: [Factory girl] Enforce consistent usage of parentheses in factory girl methods
I'm working on a code base where I have both styles of using factory:
`create(:user, name: 'John')` and `create :user, name: 'John;`
it will be good if one of those styles is required.

Note: parentheses may be required in nested associations no matter the style, like in:
`create :account, user: create(:user, name 'John')`

Liberatys:
Implement a new cop that enables the enforcement of parentheses in factory bot calls.

This cop attempts to solve the "issue" outlined in #416
edit: fixes #416

---

* [x] Added tests.
* [x] Updated documentation.

### DIFF
--- a/config/default.yml
+++ b/config/default.yml
@@ -918,6 +918,16 @@ RSpec/FactoryBot/AttributeDefinedStatically:
   VersionChanged: '2.0'
   Reference: https://www.rubydoc.info/gems/rubocop-rspec/RuboCop/Cop/RSpec/FactoryBot/AttributeDefinedStatically
 
+RSpec/FactoryBot/ConsistentParenthesesStyle:
+  Description: Use a consistent style for parentheses in factory bot calls.
+  Enabled: pending
+  EnforcedStyle: require_parentheses
+  SupportedStyles:
+    - require_parentheses
+    - omit_parentheses
+  VersionAdded: "<<next>>"
+  Reference: https://www.rubydoc.info/gems/rubocop-rspec/RuboCop/Cop/RSpec/FactoryBot/ConsistentParenthesesStyle
+
 RSpec/FactoryBot/CreateList:
   Description: Checks for create_list usage.
   Enabled: true
--- a/docs/modules/ROOT/pages/cops.adoc
+++ b/docs/modules/ROOT/pages/cops.adoc
@@ -103,6 +103,7 @@
 === Department xref:cops_rspec_factorybot.adoc[RSpec/FactoryBot]
 
 * xref:cops_rspec_factorybot.adoc#rspecfactorybot/attributedefinedstatically[RSpec/FactoryBot/AttributeDefinedStatically]
+* xref:cops_rspec_factorybot.adoc#rspecfactorybot/consistentparenthesesstyle[RSpec/FactoryBot/ConsistentParenthesesStyle]
 * xref:cops_rspec_factorybot.adoc#rspecfactorybot/createlist[RSpec/FactoryBot/CreateList]
 * xref:cops_rspec_factorybot.adoc#rspecfactorybot/factoryclassname[RSpec/FactoryBot/FactoryClassName]
 * xref:cops_rspec_factorybot.adoc#rspecfactorybot/syntaxmethods[RSpec/FactoryBot/SyntaxMethods]
--- a/docs/modules/ROOT/pages/cops_rspec_factorybot.adoc
+++ b/docs/modules/ROOT/pages/cops_rspec_factorybot.adoc
@@ -51,6 +51,67 @@ count { 1 }
 
 * https://www.rubydoc.info/gems/rubocop-rspec/RuboCop/Cop/RSpec/FactoryBot/AttributeDefinedStatically
 
+== RSpec/FactoryBot/ConsistentParenthesesStyle
+
+|===
+| Enabled by default | Safe | Supports autocorrection | Version Added | Version Changed
+
+| Pending
+| Yes
+| Yes
+| <<next>>
+| -
+|===
+
+Use a consistent style for parentheses in factory bot calls.
+
+=== Examples
+
+[source,ruby]
+----
+# bad
+create :user
+build(:user)
+create(:login)
+create :login
+----
+
+==== `EnforcedStyle: require_parentheses` (default)
+
+[source,ruby]
+----
+# good
+create(:user)
+create(:user)
+create(:login)
+build(:login)
+----
+
+==== `EnforcedStyle: omit_parentheses`
+
+[source,ruby]
+----
+# good
+create :user
+build :user
+create :login
+create :login
+----
+
+=== Configurable attributes
+
+|===
+| Name | Default value | Configurable values
+
+| EnforcedStyle
+| `require_parentheses`
+| `require_parentheses`, `omit_parentheses`
+|===
+
+=== References
+
+* https://www.rubydoc.info/gems/rubocop-rspec/RuboCop/Cop/RSpec/FactoryBot/ConsistentParenthesesStyle
+
 == RSpec/FactoryBot/CreateList
 
 |===
--- a/lib/rubocop/cop/rspec/factory_bot/consistent_parentheses_style.rb
+++ b/lib/rubocop/cop/rspec/factory_bot/consistent_parentheses_style.rb
@@ -1,0 +1,99 @@
+# frozen_string_literal: true
+
+module RuboCop
+  module Cop
+    module RSpec
+      module FactoryBot
+        # Use a consistent style for parentheses in factory bot calls.
+        #
+        # @example
+        #
+        #   # bad
+        #   create :user
+        #   build(:user)
+        #   create(:login)
+        #   create :login
+        #
+        # @example `EnforcedStyle: require_parentheses` (default)
+        #
+        #   # good
+        #   create(:user)
+        #   create(:user)
+        #   create(:login)
+        #   build(:login)
+        #
+        # @example `EnforcedStyle: omit_parentheses`
+        #
+        #   # good
+        #   create :user
+        #   build :user
+        #   create :login
+        #   create :login
+        #
+        class ConsistentParenthesesStyle < Base
+          extend AutoCorrector
+          include ConfigurableEnforcedStyle
+          include RuboCop::RSpec::FactoryBot::Language
+          include RuboCop::Cop::Util
+
+          def self.autocorrect_incompatible_with
+            [Style::MethodCallWithArgsParentheses]
+          end
+
+          MSG_REQUIRE_PARENS = 'Prefer method call with parentheses'
+          MSG_OMIT_PARENS = 'Prefer method call without parentheses'
+
+          FACTORY_CALLS = RuboCop::RSpec::FactoryBot::Language::METHODS
+
+          # @!method factory_call(node)
+          def_node_matcher :factory_call, <<-PATTERN
+              (send
+                ${#factory_bot? nil?} %FACTORY_CALLS
+              $...)
+          PATTERN
+
+          def on_send(node)
+            return if nested_call?(node) # prevent from nested matching
+
+            factory_call(node) do
+              if node.parenthesized?
+                process_with_parentheses(node)
+              else
+                process_without_parentheses(node)
+              end
+            end
+          end
+
+          def process_with_parentheses(node)
+            return unless style == :omit_parentheses
+
+            add_offense(node.loc.selector,
+                        message: MSG_OMIT_PARENS) do |corrector|
+              remove_parentheses(corrector, node)
+            end
+          end
+
+          def process_without_parentheses(node)
+            return unless style == :require_parentheses
+
+            add_offense(node.loc.selector,
+                        message: MSG_REQUIRE_PARENS) do |corrector|
+              add_parentheses(node, corrector)
+            end
+          end
+
+          def nested_call?(node)
+            node.parent&.send_type?
+          end
+
+          private
+
+          def remove_parentheses(corrector, node)
+            corrector.replace(node.location.begin, ' ')
+            corrector.remove(node.location.end)
+          end
+        end
+      end
+    end
+  end
+end
--- a/spec/rubocop/cop/rspec/factory_bot/consistent_parentheses_style_spec.rb
+++ b/spec/rubocop/cop/rspec/factory_bot/consistent_parentheses_style_spec.rb
@@ -1,0 +1,279 @@
+# frozen_string_literal: true
+
+RSpec.describe RuboCop::Cop::RSpec::FactoryBot::ConsistentParenthesesStyle do
+  let(:cop_config) do
+    { 'EnforcedStyle' => enforced_style }
+  end
+
+  context 'when EnforcedStyle is :enforce_parentheses' do
+    let(:enforced_style) { :require_parentheses }
+
+    context 'with create' do
+      it 'flags the call to use parentheses' do
+        expect_offense(<<~RUBY)
+          create :user
+          ^^^^^^ Prefer method call with parentheses
+        RUBY
+
+        expect_correction(<<~RUBY)
+          create(:user)
+        RUBY
+      end
+    end
+
+    context 'with multiline method calls' do
+      it 'expects parentheses around multiline call' do
+        expect_offense(<<~RUBY)
+          create :user,
+          ^^^^^^ Prefer method call with parentheses
+            username: "PETER",
+            peter: "USERNAME"
+        RUBY
+
+        expect_correction(<<~RUBY)
+          create(:user,
+            username: "PETER",
+            peter: "USERNAME")
+        RUBY
+      end
+    end
+
+    context 'with build' do
+      it 'flags the call to use parentheses' do
+        expect_offense(<<~RUBY)
+          build :user
+          ^^^^^ Prefer method call with parentheses
+        RUBY
+
+        expect_correction(<<~RUBY)
+          build(:user)
+        RUBY
+      end
+    end
+
+    context 'with mixed tests' do
+      it 'flags the call to use parentheses' do
+        expect_offense(<<~RUBY)
+          build_list :user, 10
+          ^^^^^^^^^^ Prefer method call with parentheses
+          build_list "user", 10
+          ^^^^^^^^^^ Prefer method call with parentheses
+          create_list :user, 10
+          ^^^^^^^^^^^ Prefer method call with parentheses
+          build_stubbed :user
+          ^^^^^^^^^^^^^ Prefer method call with parentheses
+          build_stubbed_list :user, 10
+          ^^^^^^^^^^^^^^^^^^ Prefer method call with parentheses
+        RUBY
+
+        expect_correction(<<~RUBY)
+          build_list(:user, 10)
+          build_list("user", 10)
+          create_list(:user, 10)
+          build_stubbed(:user)
+          build_stubbed_list(:user, 10)
+        RUBY
+      end
+    end
+
+    context 'with nested calling' do
+      it 'flags the call to use parentheses' do
+        expect_offense(<<~RUBY)
+          build :user, build(:yester)
+          ^^^^^ Prefer method call with parentheses
+        RUBY
+
+        expect_correction(<<~RUBY)
+          build(:user, build(:yester))
+        RUBY
+      end
+
+      it 'works in a bigger context' do
+        expect_offense(<<~RUBY)
+          context 'with context' do
+            let(:build) { create :user, build(:user) }
+                          ^^^^^^ Prefer method call with parentheses
+
+            it 'test in test' do
+              user = create :user, first: name, peter: miller
+                     ^^^^^^ Prefer method call with parentheses
+            end
+
+            let(:build) { create :user, build(:user, create(:user, create(:first_name))) }
+                          ^^^^^^ Prefer method call with parentheses
+          end
+        RUBY
+
+        expect_correction(<<~RUBY)
+          context 'with context' do
+            let(:build) { create(:user, build(:user)) }
+
+            it 'test in test' do
+              user = create(:user, first: name, peter: miller)
+            end
+
+            let(:build) { create(:user, build(:user, create(:user, create(:first_name)))) }
+          end
+        RUBY
+      end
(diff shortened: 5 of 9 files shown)


Liberatys:
This cop attempts to solve the "issue" outlined in 12940

config/default.yml
```diff
@@ -721,6 +721,16 @@ RSpec/FactoryBot/AttributeDefinedStatically:
   VersionChanged: '2.0'
   StyleGuide: https://www.rubydoc.info/gems/rubocop-rspec/RuboCop/Cop/RSpec/FactoryBot/AttributeDefinedStatically
 
+RSpec/FactoryBot/ConsistentParenthesesStyle:
+  Description: Use a consistent style for parentheses in factory bot calls.
```
pirj, line 725: I suggest limiting this cop to [directories with factory definitions](https://github.com/rubocop-hq/rubocop-rspec/blob/7d048b653b73315509600ae42241bdb03a033ae4/config/default.yml#L716), for other cases, including specs, there's a RuboCop core cop that keeps an eye on parenthesis usages.

Or do you think that another cop is usually disabled for directories with specs?

lib/rubocop/cop/rspec/factory_bot/consistent_parentheses_style.rb
```diff
@@ -0,0 +1,101 @@
...
+        # Use a consistent style for parentheses in factory bot calls
+        #
+        # @example
+        #
+        #   # bad
+        #
+        #   create :user
+        #   build(:user)
+        #   create(:login)
+        #   create :login
```
pirj, line 16: There are other methods, like `atributes_for`, `build_stubbed`, `build_stubbed_list`, `build_list`, `create_list`, `build_pair`, `create_pair` (not sure if `build_stubbed_pair` exists).

lib/rubocop/cop/rspec/factory_bot/consistent_parentheses_style.rb
```diff
@@ -0,0 +1,101 @@
...
+                ${(const nil? {:FactoryGirl :FactoryBot}) nil?} :create (sym $_)
+              $...)
+              (send 
+                ${(const nil? {:FactoryGirl :FactoryBot}) nil?} :build (sym $_)
+              $...)
+            }
+          PATTERN
+
+          def on_send(node)
+            return if nested_call?(node)
```
pirj, line 49: But why not? For performance?

lib/rubocop/cop/rspec/factory_bot/consistent_parentheses_style.rb
```diff
@@ -0,0 +1,101 @@
...
+          def_node_matcher :factory_call, <<-PATTERN
+            {
+              (send
+                ${(const nil? {:FactoryGirl :FactoryBot}) nil?} :create (sym $_)
+              $...)
+              (send 
+                ${(const nil? {:FactoryGirl :FactoryBot}) nil?} :build (sym $_)
+              $...)
+            }
+          PATTERN
```
pirj, line 46: You can simplify this to:
```ruby
          def_node_matcher :factory_call, <<-PATTERN
              (send
                ${(const nil? {:FactoryGirl :FactoryBot}) nil?} {:create :build} (sym $_)
              $...)
          PATTERN
```

lib/rubocop/cop/rspec/factory_bot/consistent_parentheses_style.rb
```diff
@@ -0,0 +1,101 @@
...
+          extend AutoCorrector
+          include ConfigurableEnforcedStyle
+
+          MSG_ENFORCE_PARENS = 'Prefer method call with parentheses'
+          MSG_OMIT_PARENS = 'Prefer method call without parentheses'
+
+          def_node_matcher :factory_call, <<-PATTERN
+            {
+              (send
+                ${(const nil? {:FactoryGirl :FactoryBot}) nil?} :create (sym $_)
```
pirj, line 40: `(sym $_)` limits this to literal symbols:
```
create(:user)
```
however, it will miss other cases:
```
model = :user
create(model)
```
or
```ruby
create('user')
```

lib/rubocop/cop/rspec/factory_bot/consistent_parentheses_style.rb
```diff
@@ -0,0 +1,101 @@
...
+
+            add_offense(node.loc.selector, message: MSG_ENFORCE_PARENS) do |corrector|
+              autocorrect_enforce_parens(corrector, node)
+            end
+          end
+
+          def nested_call?(node)
+            parent = node.parent
+            # prevent from nested matching
+            if parent.respond_to?('method_name')
```
pirj, line 79: Is it possible to use `parent.send_type?`?

pirj (commented):
Wow, thanks a lot, looks really great!

I'm not sure I completely understand if it's really necessary to check call nesting.
Also, the detection can be improved to cover a wider range of FB syntax usages right away.

lib/rubocop/cop/rspec/factory_bot/consistent_parentheses_style.rb
Liberatys: Because create(:default_country, create :default_country) nesting without parens is not a valid call. Ruby forces you to put parens around the nested factory. So checking it only causes issues

Liberatys: Ruby tells me the following: 

`let(:user) { create(:user) }` == valid

`let(:user) { create(:user, create :user) }` == very not valid

For some reason it is expecting a ( in this format

pirj: Right. Sees there's a difference when there is more than one argument.

However, if it's invalid syntax, why should we care about it in the cop?
```
$ ruby-parse -e 'create(:user, create :user)'
(fragment:0):1:22: error: unexpected token tSYMBOL
(fragment:0):1: create(:user, create :user)
(fragment:0):1:                      ^~~~~
```

Liberatys: That is a fair point ^^. But I'd prefer not to iterate on the nested calls because that could cause issues with the cop trying to fix the nested calls.

lib/rubocop/cop/rspec/factory_bot/consistent_parentheses_style.rb
```diff
@@ -0,0 +1,100 @@
...
+        #   build(:login)
+        #
+        class ConsistentParenthesesStyle < Base
+          extend AutoCorrector
+          include ConfigurableEnforcedStyle
+
+          MSG_ENFORCE_PARENS = 'Prefer method call with parentheses'
+          MSG_OMIT_PARENS = 'Prefer method call without parentheses'
+
+          FACTORY_CALLS = %i[create build]
```
pirj, line 37: This seems to be unused.

lib/rubocop/cop/rspec/factory_bot/consistent_parentheses_style.rb
pirj: Ha, I finally got it. Sorry for being a 🐌
So this check is to prevent auto-correction for `create(:user, create(:user))` to a nonsense when `omit_parentheses` is the enforced style.
I'm all for having this check 👍 Thanks for standing your ground.

However, due to how `nested_call?` is implemented, it might still attempt to auto-correct the following for example:
```ruby
puts(1, create(:user))
```

To cover this it will be enough to check `node.parent.send_type?`.

lib/rubocop/cop/rspec/factory_bot/consistent_parentheses_style.rb
```diff
@@ -0,0 +1,100 @@
...
+
+          def_node_matcher :factory_call, <<-PATTERN
+              (send
+                ${(const nil? {:FactoryGirl :FactoryBot}) nil?} {:create :build :create_list 
+                :build_list :create_pair :build_stubbed_list :atributes_for :build_stubbed
+                :build_pair}
```
pirj, line 43: You may simplify the looks of this slightly by [extracting the list of methods to a constant](https://docs.rubocop.org/rubocop-ast/node_pattern.html#const-for-constants):
```
(send
  ${(const nil? {:FactoryGirl :FactoryBot}) nil?} $%FACTORY_CALLS $...
)
```

pirj (approved):
Looks good!
Added a couple of cosmetic notes.
Thanks!

pirj:
[RuboCop is red](https://app.circleci.com/pipelines/github/rubocop-hq/rubocop-rspec/743/workflows/9f4f9b76-0879-4ddd-b6f0-41a6e9a6599d/jobs/19251), I hope you'll find it relatively easy to fix it.

lib/rubocop/cop/rspec/factory_bot/consistent_parentheses_style.rb
Liberatys: 👍

lib/rubocop/cop/rspec/factory_bot/consistent_parentheses_style.rb
Liberatys: I changed the implementation and used your example as a test :D and thank you for being so patient during this discussion ^^

lib/rubocop/cop/rspec/factory_bot/consistent_parentheses_style.rb
```diff
@@ -0,0 +1,99 @@
+# frozen_string_literal: true
+
+module RuboCop
+  module Cop
+    module RSpec
+      module FactoryBot
+        # Use a consistent style for parentheses in factory bot calls
+        #
+        # @example
```
pirj, line 9: As a last touch, I suggest splitting this into two examples, one per style, like e.g. [here](https://github.com/rubocop-hq/rubocop-rspec/blob/master/lib/rubocop/cop/rspec/message_spies.rb#L11).

config/default.yml
```diff
@@ -721,6 +721,16 @@ RSpec/FactoryBot/AttributeDefinedStatically:
...
+  Enabled: pending
+  EnforcedStyle: enforce_parentheses
+  SupportedStyles:
+    - enforce_parentheses
```
Darhazer, line 729: For consistency with other Rubocop's cops I'd suggest to name this `require_parentheses`

lib/rubocop/cop/rspec/factory_bot/consistent_parentheses_style.rb
```diff
@@ -0,0 +1,99 @@
...
+            factory_call(node) do
+              if node.parenthesized?
+                process_with_parentheses(node)
+              else
+                process_without_parentheses(node)
+              end
+            end
+          end
+
+          def process_with_parentheses(node)
```
Darhazer, line 60: I wonder if Rubocop's modules `RequireParentheses` / `OmitParentheses` can't be largely reused. At least they handle much more cases - like multiline method calls, for which we'll also need specs

Darhazer (dismissed):

lib/rubocop/cop/rspec/factory_bot/consistent_parentheses_style.rb
Liberatys: Tests have been added

Liberatys:
@Darhazer Sorry for a long time between activities ^^ I pretty much forgot that this pr exists :D.
I added your concern about the 'nested' call as a spec: https://github.com/rubocop/rubocop-rspec/pull/1096/files#diff-e694e1d15e68306d0a724e35afadbc5a430fd6602ccd827e985f2aefae2935c7R228.

I also added some specs for multiline calls: https://github.com/rubocop/rubocop-rspec/pull/1096/files#diff-e694e1d15e68306d0a724e35afadbc5a430fd6602ccd827e985f2aefae2935c7R160-R175.

🙏🏻 Thank you for the work on rspec and rubocop.

lib/rubocop/cop/rspec/factory_bot/consistent_parentheses_style.rb
```diff
@@ -0,0 +1,107 @@
...
+            [Style::MethodCallWithArgsParentheses]
+          end
+
+          MSG_ENFORCE_PARENS = 'Prefer method call with parentheses'
+          MSG_OMIT_PARENS = 'Prefer method call without parentheses'
+
+          FACTORY_CALLS = Set[:create, :build, :create_list,
+                              :build_list, :create_pair,
+                              :build_stubbed_list, :atributes_for,
+                              :build_stubbed, :build_pair ].freeze
```
bquorning, line 48: Where does this list of method names come from? I found a list of method names at https://www.rubydoc.info/gems/factory_bot/FactoryBot/Syntax/Methods, but that list also includes `attributes_for_list`, `attributes_for_pair` and `build_stubbed_pair` – so I wonder if those are deliberately omitted.

There’s a typo in `:atributes_for`.

Liberatys: Thank you very much ^^ The 3 missing were not left out on purpose. I guess I forgot them when the pr was created.

lib/rubocop/cop/rspec/factory_bot/consistent_parentheses_style.rb
```diff
@@ -0,0 +1,105 @@
...
+                              :build_list, :create_pair,
+                              :build_stubbed_list, :attributes_for,
+                              :attributes_for_list, :attributes_for_pair,
+                              :build_stubbed_pair, :build_stubbed,
+                              :build_pair ].freeze
+
+          # @!method factory_call(node)
+          def_node_matcher :factory_call, <<-PATTERN
+              (send
+                ${(const nil? {:FactoryGirl :FactoryBot}) nil?} %FACTORY_CALLS
```
pirj, line 55: Cosmetic: we have a helper for this in `lib/rubocop/rspec/factory_bot/language.rb` if you like it.
Could be just
```
${#factory_bot? nil?} %FACTORY_CALLS
```
As a free benefit, it also covers `::FactoryBot` calls.

lib/rubocop/cop/rspec/factory_bot/consistent_parentheses_style.rb
```diff
@@ -0,0 +1,105 @@
...
+
+          def nested_call?(node)
+            parent = node.parent
+            # prevent from nested matching
+            parent&.send_type?
+          end
+
+          private
+
+          def autocorrect_remove_parens(corrector, node)
```
pirj, line 97: Cosmetic. Maybe call it `remove_parentheses` to match `add_parentheses`?

lib/rubocop/cop/rspec/factory_bot/consistent_parentheses_style.rb
```diff
@@ -0,0 +1,105 @@
...
+          include RuboCop::Cop::Util
+
+          def self.autocorrect_incompatible_with
+            [Style::MethodCallWithArgsParentheses]
+          end
+
+          MSG_ENFORCE_PARENS = 'Prefer method call with parentheses'
+          MSG_OMIT_PARENS = 'Prefer method call without parentheses'
+
+          FACTORY_CALLS = Set[:create, :build, :create_list,
```
pirj, line 45: Cosmetic: there's a similar (and even more complete) list in `lib/rubocop/cop/rspec/factory_bot/syntax_methods.rb`. Does it make sense to extract it?

spec/rubocop/cop/rspec/factory_bot/consistent_parentheses_style_spec.rb
```diff
@@ -0,0 +1,279 @@
...
+      it 'does not flag as invalid - build' do
+        expect_no_offenses(<<~RUBY)
+          build :user
+        RUBY
+      end
+    end
+
+    it 'works in a bigger context' do
+      expect_offense(<<~RUBY)
+        class Context
```
pirj, line 252: Should this be `RSpec.describe Context do`?

spec/rubocop/cop/rspec/factory_bot/consistent_parentheses_style_spec.rb
```diff
@@ -0,0 +1,279 @@
...
+                   ^^^^^^ Prefer method call without parentheses
+          end
+
+          let(:build) { create(:user, build(:user, create(:user, create(:first_name)))) }
+                        ^^^^^^ Prefer method call without parentheses
+        end
+      RUBY
+
+      expect_correction(<<~RUBY)
+        class Context
```
pirj, line 267: Here as well.

pirj (approved):
Looks superb. Thanks for getting back to this PR!
Please rebase just in case and re-generate docs.
Otherwise, just purely cosmetic notes, feel free to ignore, no showstoppers for me.

config/default.yml
```diff
@@ -841,6 +841,16 @@ RSpec/FactoryBot/AttributeDefinedStatically:
...
   Reference: https://www.rubydoc.info/gems/rubocop-rspec/RuboCop/Cop/RSpec/FactoryBot/AttributeDefinedStatically
 
+RSpec/FactoryBot/ConsistentParenthesesStyle:
+  Description: Use a consistent style for parentheses in factory bot calls.
+  Enabled: pending
+  EnforcedStyle: require_parentheses
+  SupportedStyles:
+    - require_parentheses
+    - omit_parentheses
+  VersionAdded: '2.11'
```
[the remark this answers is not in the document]
Liberatys: noice! :D

docs/modules/ROOT/pages/cops_rspec_factorybot.adoc
```diff
@@ -51,6 +51,57 @@ count { 1 }
...
+| Pending
+| Yes
+| Yes
+| 2.1
+| -
+|===
+
+Use a consistent style for parentheses in factory bot calls.
+
+=== Examples
```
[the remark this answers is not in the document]
Liberatys: 👍🏻

Liberatys:
@pirj I rebased onto upstream, regenerated the docs and implement your suggestions ^^ Thank you for the review :D

docs/modules/ROOT/pages/cops.adoc
```diff
@@ -97,6 +97,7 @@
 === Department xref:cops_rspec_factorybot.adoc[RSpec/FactoryBot]
 
 * xref:cops_rspec_factorybot.adoc#rspecfactorybot/attributedefinedstatically[RSpec/FactoryBot/AttributeDefinedStatically]
+* xref:cops_rspec_factorybot.adoc#rspecfactorybot/consistentparenthesesstyle[RSpec/FactoryBot/ConsistentParenthesesStyle]
```
pirj, line 100: Unrelated to your changes. Anchors for cops in sub-departments (Capybara, FactoryBot) seem broken. Worth looking at in another PR, and probably the problem is in RuboCop repo.

Should be:
```diff
- #rspecfactorybot/consistentparenthesesstyle
+ #rspecfactorybotconsistentparenthesesstyle
```

lib/rubocop/cop/rspec/factory_bot/consistent_parentheses_style.rb
```diff
@@ -0,0 +1,101 @@
...
+          include RuboCop::Cop::Util
+
+          def self.autocorrect_incompatible_with
+            [Style::MethodCallWithArgsParentheses]
+          end
+
+          MSG_ENFORCE_PARENS = 'Prefer method call with parentheses'
+          MSG_OMIT_PARENS = 'Prefer method call without parentheses'
+
+          FACTORY_CALLS = RuboCop::RSpec::FactoryBot::Language::METHODS
```
pirj, line 46: Cosmetic: this module is included, so `METHODS` should be available without the namespace.

pirj:
Thank you! Looks great 🙌

lib/rubocop/cop/rspec/factory_bot/consistent_parentheses_style.rb
Liberatys: This was actually a deliberate choice ^^ only `METHODS` or directly using `METHODS` seemed a bit too magical ^^ by using the namespace I at least understand where the constant is coming from.

Liberatys: (I can of course change it to `METHODS` if you prefer?

Liberatys:
ah 🤦🏻 @pirj Sorry. Should have checked again if the CI passed :/. Thank you very much for fixing it!

lib/rubocop/cop/rspec/factory_bot/consistent_parentheses_style.rb
```diff
@@ -0,0 +1,99 @@
...
+              if node.parenthesized?
+                process_with_parentheses(node)
+              else
+                process_without_parentheses(node)
+              end
+            end
+          end
+
+          def process_with_parentheses(node)
+            return if style == :require_parentheses
```
bquorning, line 68: We’re definitely getting into nitpick territory now, so feel free to ignore.

Shouldn’t this line rather be `return unless style == :omit_parentheses`. That way, the conditional wouldn’t need to change if we ever added a 3rd style.

lib/rubocop/cop/rspec/factory_bot/consistent_parentheses_style.rb
```diff
@@ -0,0 +1,99 @@
...
+          include ConfigurableEnforcedStyle
+          include RuboCop::RSpec::FactoryBot::Language
+          include RuboCop::Cop::Util
+
+          def self.autocorrect_incompatible_with
+            [Style::MethodCallWithArgsParentheses]
+          end
+
+          MSG_ENFORCE_PARENS = 'Prefer method call with parentheses'
+          MSG_OMIT_PARENS = 'Prefer method call without parentheses'
```
bquorning, line 44: The supported styles are called “require”… and “omit”…. Should we rename the …“ENFORCE”… constant to match?

bquorning (approved):

pirj:
Thank you for the contribution, your time and effort, @Liberatys! My apologies for delaying the merge.

PR #1099, opened 2020-11-21
Fix false positive for relative file path runs with long namespaces (fixes #1091)

### LINKED ISSUE #1091: RSpec/FilePath is broken with vscode-ruby after 1.39.0 (2020-05-01)
Details:
- rubocop-rspec version: 1.43.2 (all versions >= 1.39.0 affected)
- Rubocop version: 0.92.0

Related issue with vscode-ruby:
https://github.com/rubyide/vscode-ruby/issues/625

---

The RSpec/FilePath cop does not work with [vscode-ruby](https://github.com/rubyide/vscode-ruby).

It causes issues such as this one (originating from https://github.com/decidim/decidim/pull/6857):
image

Why this happens is that vscode-ruby runs rubocop inside the folder where the file exists. So, in the example case shown in the screenshot it does the following:

```bash
$ cd /home/user/dev/decidim/decidim-assemblies/spec/presenters/decidim/assemblies/admin_log
$ bundle exec rubocop /home/user/dev/decidim/decidim-assemblies/spec/presenters/decidim/assemblies/admin_log/assemblies_type_presenter_spec.rb
Inspecting 1 file
C

Offenses:

assemblies_type_presenter_spec.rb:5:1: C: RSpec/FilePath: Spec path should end with decidim/assemblies/admin_log/assemblies_type_presenter*_spec.rb.
describe Decidim::Assemblies::AdminLog::AssembliesTypePresenter, type: :helper do
^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

```

I was debugging vscode-ruby to find the root cause but I figured it was this commit in rubocop-rspec that has broken the functionality:
https://github.com/rubocop-hq/rubocop-rspec/commit/811ae08ea3a60453e18a8bcb9dfaefafd848f779#diff-5b709885b35566a0c699ad4a3200b9b7acbc28ef640fa56a8e66f6c732420a34

Is there some reasoning behind this change or should vscode-ruby change the way it runs the linter?

This could be fixed for vscode-ruby by changing the following lines in rubocop-rspec:
https://github.com/rubocop-hq/rubocop-rspec/blob/78f94bb3216fa2a6f27a43af47728762caabde89/lib/rubocop/cop/rspec/file_path.rb#L135-L137

To:

```ruby
filename = File.expand_path(processed_source.buffer.name)
```

But maybe this would possibly reintroduce the problems that https://github.com/rubocop-hq/rubocop-rspec/commit/811ae08ea3a60453e18a8bcb9dfaefafd848f779 fixed?

It is hard to make the judgement where the actual bug is without knowing some insight about this change. I don't understand why it has been made.

ahukkanen:
This is a PR that fixes the issue that is further explained at #1091. 

This was broken by this commit included in the PR #869:
https://github.com/rubocop-hq/rubocop-rspec/commit/811ae08ea3a60453e18a8bcb9dfaefafd848f779#diff-5b709885b35566a0c699ad4a3200b9b7acbc28ef640fa56a8e66f6c732420a34

It turned out to be extremely difficult (at least I don't know how after multiple tries and experiments) to fix the issue described at #1091 using the file name globs. Therefore I switched to regular expression based matching which allows a more specific targeting for the case in question.

The removed specs originate from this commit in the PR #917:
https://github.com/rubocop-hq/rubocop-rspec/commit/23ad402fb3bd5149adc456ca3c54204c38fc271f#diff-5bc1941f9b92914876efec6f7fcc70b153e82194d960213a280078c6cb4b0c3a

The removed specs are no longer needed because it now uses absolute paths as it did prior to merging #869. I believe these were also issues introduced by #869.

Fixes #1091

---

* [x] Added tests.
* [ ] Updated documentation.

### DIFF
--- a/lib/rubocop/cop/rspec/file_path.rb
+++ b/lib/rubocop/cop/rspec/file_path.rb
@@ -82,30 +82,39 @@ module RuboCop
         private
 
         def ensure_correct_file_path(send_node, described_class, arguments)
-          glob = glob_for(described_class, arguments.first)
-          return if filename_ends_with?(glob)
+          pattern = pattern_for(described_class, arguments.first)
+          return if filename_ends_with?(pattern)
 
-          add_offense(send_node, message: format(MSG, suffix: glob))
+          # For the suffix shown in the offense message, modify the regular
+          # expression pattern to resemble a glob pattern for clearer error
+          # messages.
+          offense_suffix = pattern.gsub('.*', '*').sub('[^/]', '')
+            .sub('\.', '.')
+          add_offense(send_node, message: format(MSG, suffix: offense_suffix))
         end
 
         def routing_spec?(args)
           args.any?(&method(:routing_metadata?))
         end
 
-        def glob_for(described_class, method_name)
-          return glob_for_spec_suffix_only? if spec_suffix_only?
+        def pattern_for(described_class, method_name)
+          return pattern_for_spec_suffix_only? if spec_suffix_only?
 
-          "#{expected_path(described_class)}#{name_glob(method_name)}*_spec.rb"
+          [
+            expected_path(described_class),
+            name_pattern(method_name),
+            '[^/]*_spec\.rb'
+          ].join
         end
 
-        def glob_for_spec_suffix_only?
-          '*_spec.rb'
+        def pattern_for_spec_suffix_only?
+          '.*_spec\.rb'
         end
 
-        def name_glob(method_name)
+        def name_pattern(method_name)
           return unless method_name&.str_type?
 
-          "*#{method_name.str_content.gsub(/\W/, '')}" unless ignore_methods?
+          ".*#{method_name.str_content.gsub(/\W/, '')}" unless ignore_methods?
         end
 
         def expected_path(constant)
@@ -131,11 +140,9 @@ module RuboCop
           cop_config['IgnoreMethods']
         end
 
-        def filename_ends_with?(glob)
-          filename =
-            RuboCop::PathUtil.relative_path(processed_source.buffer.name)
-              .gsub('../', '')
-          File.fnmatch?("*#{glob}", filename)
+        def filename_ends_with?(pattern)
+          filename = File.expand_path(processed_source.buffer.name)
+          filename.match?("#{pattern}$")
         end
 
         def relevant_rubocop_rspec_file?(_file)
--- a/spec/rubocop/cop/rspec/file_path_spec.rb
+++ b/spec/rubocop/cop/rspec/file_path_spec.rb
@@ -64,6 +64,13 @@ RSpec.describe RuboCop::Cop::RSpec::FilePath do
     RUBY
   end
 
+  it 'registers an offense for a file without the .rb extension' do
+    expect_offense(<<-RUBY, 'spec/models/user_specxrb')
+      describe User do; end
+      ^^^^^^^^^^^^^ Spec path should end with `user*_spec.rb`.
+    RUBY
+  end
+
   it 'ignores shared examples' do
     expect_no_offenses(<<-RUBY, 'spec/models/user.rb')
       shared_examples_for 'foo' do; end
@@ -183,30 +190,26 @@ RSpec.describe RuboCop::Cop::RSpec::FilePath do
     RUBY
   end
 
-  it 'uses relative path' do
-    allow(RuboCop::PathUtil)
-      .to receive(:relative_path).and_return('spec/models/bar_spec.rb')
+  it 'registers an offense for a path containing the class name' do
     expect_offense(<<-RUBY, '/home/foo/spec/models/bar_spec.rb')
       describe Foo do; end
       ^^^^^^^^^^^^ Spec path should end with `foo*_spec.rb`.
     RUBY
   end
 
-  it 'uses relative path for sibling directory project' do
-    allow(RuboCop::PathUtil)
-      .to receive(:relative_path)
-      .and_return('../ext-project/spec/models/bar_spec.rb')
-    expect_no_offenses(<<-RUBY, '/home/ext-project/spec/models/bar_spec.rb')
-      describe Bar do; end
+  it 'detects the path based on a class name with long module' do
+    expect_offense(<<-RUBY, '/home/foo/spec/very/my_class_spec.rb')
+      describe Very::Long::Namespace::MyClass do; end
+      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ Spec path should end with `very/long/namespace/my_class*_spec.rb`.
     RUBY
   end
 
-  it 'uses relative path for different path project' do
-    allow(RuboCop::PathUtil)
-      .to receive(:relative_path)
-      .and_return('../../opt/ext-project/spec/models/bar_spec.rb')
-    expect_no_offenses(<<-RUBY, '/opt/ext-project/spec/models/bar_spec.rb')
-      describe Bar do; end
+  it 'detects the path based the absolute path of the file' do
+    allow(File).to receive(:expand_path).with('my_class_spec.rb').and_return(
+      '/home/foo/spec/very/long/namespace/my_class_spec.rb'
+    )
+    expect_no_offenses(<<-RUBY, 'my_class_spec.rb')
+      describe Very::Long::Namespace::MyClass do; end
     RUBY
   end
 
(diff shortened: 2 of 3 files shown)


lib/rubocop/cop/rspec/file_path.rb
```diff
@@ -82,30 +82,39 @@ def on_top_level_example_group(node)
...
-
-          add_offense(send_node, message: format(MSG, suffix: glob))
+          pattern = pattern_for(described_class, arguments.first)
+          return if filename_ends_with?(pattern)
+
+          # For the suffix shown in the offense message, modify the regular
+          # expression pattern to resemble a glob pattern for clearer error
+          # messages.
+          offense_suffix = pattern.gsub('.*', '*').gsub('[^/]*', '*')
+            .sub('\.rb', '.rb')
```
bquorning, line 92: How about if we substitute as little as possible, so very narrowly matching the regex-specific parts of the string? E.g.
```suggestion
          offense_suffix = pattern.gsub('.*', '*').sub('[^/]', '')
            .sub('\.', '.')
```
What do you think?

lib/rubocop/cop/rspec/file_path.rb
```diff
@@ -131,11 +140,9 @@ def ignore_methods?
...
         end
 
-        def filename_ends_with?(glob)
-          filename =
-            RuboCop::PathUtil.relative_path(processed_source.buffer.name)
-              .gsub('../', '')
-          File.fnmatch?("*#{glob}", filename)
+        def filename_ends_with?(pattern)
+          filename = File.expand_path(processed_source.buffer.name)
+          filename.match?(".*#{pattern}$")
```
bquorning, line 145: `.*` is unnecessary at the beginning (or end) of a regex.

bquorning: Could we move the `$` (which should probably be `\z`?) into the pattern itself? Then this line would become simply `filename.match?(pattern)`…

bquorning (commented):
I agree this looks like a better solution to the problems that #869 and #917 were fixing.

But I wonder if this PR works on Windows (do we need to split by `File::SEPARATOR` and join with `'/'` before calling `#match?`) or case sensitive filesystems.

bquorning:
> * [ ]  Squashed related commits together. - (❓ - couldn't you squash when merging?)

We don’t use GitHub’s “Squash and merge” option, since it fails to create a merge commit.

lib/rubocop/cop/rspec/file_path.rb
ahukkanen: Sure, it's changed now.

lib/rubocop/cop/rspec/file_path.rb
ahukkanen: - You are right about unnecessary `.*` in the beginning. It is removed.
- We cannot move the `$` or `\z` to the `pattern_for` method because it would then appear in the offense messages unless it is also substituted. I think that it makes sense to keep the ending part of the pattern in the filename_**ends_with?** method.
- Regarding `$` or `\z` I guess either one is fine. Is there any case when a single filename can be multiline?

spec/rubocop/cop/rspec/file_path_spec.rb
```diff
@@ -183,30 +183,27 @@
...
   end
 
-  it 'uses relative path for different path project' do
-    allow(RuboCop::PathUtil)
-      .to receive(:relative_path)
-      .and_return('../../opt/ext-project/spec/models/bar_spec.rb')
-    expect_no_offenses(<<-RUBY, '/opt/ext-project/spec/models/bar_spec.rb')
-      describe Bar do; end
+  it 'detects the path based the absolute path of the file' do
+    allow(File).to receive(:expand_path).and_call_original
```
pirj, line 201: Is this needed? If you've experienced an issue with RSpec going nuts about stubbed `File.expand_path`, it's been [fixed in RSpec 3.10](https://github.com/rspec/rspec-support/pull/431/commits/d18bd38f608eff053a32bcbc9d3604f280c8942f#diff-99389396657b227e4704a1f3baef7a701d143906dd68f114dbf1aa42c790c1beR15).

lib/rubocop/cop/rspec/file_path.rb
```diff
@@ -82,30 +82,39 @@ def on_top_level_example_group(node)
...
+        def pattern_for_spec_suffix_only?
+          '.*_spec.rb'
         end
 
-        def name_glob(method_name)
+        def name_pattern(method_name)
           return unless method_name&.str_type?
 
-          "*#{method_name.str_content.gsub(/\W/, '')}" unless ignore_methods?
+          ".*#{method_name.str_content.gsub(/\W/, '')}" unless ignore_methods?
```
pirj, line 117: I don't exactly agree with the following offence:
```
/Users/pirj/source/rspec-dev/repos/rspec-expectations/spec/rspec/expectations/failure_aggregator_spec.rb:2:3: C: RSpec/FilePath: Spec path should end with failure_aggregator*whenusedviaaggregate_failures*_spec.rb.
  RSpec.describe FailureAggregator, "when used via `aggregate_failures`" do
  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```
As I would expect `whenusedviaaggregate_failures` to be more like `when_used_via_aggregate_failures`, but it has nothing to do with this pull request.

pirj (approved):
It's strange that the cop fails to detect an offence in:
```
RSpec.describe Object, "should" do
```
or
```
RSpec.describe Object, "#should" do
```
when inspecting `spec/rspec/expectations/extensions/kernel_spec.rb`.
But it was like that before this change as well.

Yet another case that this cop is missing (both before and after the change) is a namespace:
```
# spec/rspec/expectations/configuration_spec.rb
module RSpec
  module Mocks # Should be Expectations
    RSpec.describe Configuration do
```

Those two are not blockers, but more like ideas if you'd like to follow-up on this.

There's a new legitimate offence detected:
```
/Users/pirj/source/rspec-dev/repos/rspec-expectations/spec/rspec/expectations/syntax_spec.rb:88:3: C: RSpec/FilePath: Spec path should end with expectations*_spec.rb.
  RSpec.describe Expectations do
  ^^^^^^^^^^^^^^^^^^^^^^^^^^^
```
👍 

I couldn't break the cop by testing locally.

Looks good, the change is quite elegant, and besides fixing all the cases is now easier to read. Thanks a lot!

lib/rubocop/cop/rspec/file_path.rb
bquorning: I haven’t double checked this problem, but could you open it as a new issue @pirj ?

pirj: Filed all three deficiencies as tickets 👍

spec/rubocop/cop/rspec/file_path_spec.rb
pirj: If you'd like to update our used RSpec version and remove this like, please go ahead. Preferably in a separate pull request though. But I consider this optional.

ahukkanen: You are right. It's not needed. I guess I was running on older rspec version before.

Looking at my `Gemfile.lock`, it's already using rspec 3.10.0. Looking at the project gemspec, it should be already using the latest version.

I removed the line and tests seem to pass.

lib/rubocop/cop/rspec/file_path.rb
```diff
@@ -82,30 +82,39 @@ def on_top_level_example_group(node)
...
+            expected_path(described_class),
+            name_pattern(method_name),
+            '[^/]*_spec\.rb'
+          ].join
         end
 
-        def glob_for_spec_suffix_only?
-          '*_spec.rb'
+        def pattern_for_spec_suffix_only?
+          '.*_spec.rb'
```
bquorning, line 111: Sorry I hadn’t noticed this one before.

```suggestion
          '.*_spec\.rb'
```

bquorning (commented):
Could you please 1) add a changelog entry, and 2) rebase the PR, squashing all commits down to one?

bquorning:
> But I wonder if this PR works on Windows (do we need to split by `File::SEPARATOR` and join with `'/'` before calling `#match?`) or case sensitive filesystems.

Do we have any Windows users around?

pirj:
Please add a Changelog entry, squash commits and I believe we're ready to merge.
Thanks once again!

lib/rubocop/cop/rspec/file_path.rb
ahukkanen: This is now resolved and I added specs for it.

ahukkanen:
@pirj Changelog added and commits are squashed.

ahukkanen:
> > But I wonder if this PR works on Windows (do we need to split by `File::SEPARATOR` and join with `'/'` before calling `#match?`) or case sensitive filesystems.
> 
> Do we have any Windows users around?

@bquorning I tested this under Windows and it seems to work fine there as is. This is because `processed_source.buffer.name` returns the filename with forward slashes also under Windows.

bquorning (approved):
Thank you 🙏🏼

PR #1108, opened 2020-12-08
Ignore CreateList offense on method call detection

### LINKED ISSUE #1087: CreateList is not a safe cop

ngouy:
As soon as a method call happens in a `n.times { create(:my_factory, ...}` block, nothing ensures that the method will return the same n times, so `create_list` and `create` will not nessecarly have the same output (ae `create(:something, position: rand)`)

fixes #1087 

---

* [x] Added tests.
* [x] Updated documentation.

### DIFF
--- a/docs/modules/ROOT/pages/cops_rspec_factorybot.adoc
+++ b/docs/modules/ROOT/pages/cops_rspec_factorybot.adoc
@@ -79,8 +79,14 @@ This cop can be configured using the `EnforcedStyle` option
 # good
 create_list :user, 3
 
-# good
-3.times { |n| create :user, created_at: n.months.ago }
+# bad
+3.times { create :user, age: 18 }
+
+# good - index is used to alter the created models attributes
+3.times { |n| create :user, age: n }
+
+# good - contains a method call, may return different values
+3.times { create :user, age: rand }
 ----
 
 ==== `EnforcedStyle: n_times`
--- a/lib/rubocop/cop/rspec/factory_bot/create_list.rb
+++ b/lib/rubocop/cop/rspec/factory_bot/create_list.rb
@@ -15,8 +15,14 @@ module RuboCop
         #   # good
         #   create_list :user, 3
         #
-        #   # good
-        #   3.times { |n| create :user, created_at: n.months.ago }
+        #   # bad
+        #   3.times { create :user, age: 18 }
+        #
+        #   # good - index is used to alter the created models attributes
+        #   3.times { |n| create :user, age: n }
+        #
+        #   # good - contains a method call, may return different values
+        #   3.times { create :user, age: rand }
         #
         # @example `EnforcedStyle: n_times`
         #   # bad
@@ -33,13 +39,26 @@ module RuboCop
           MSG_N_TIMES = 'Prefer %<number>s.times.'
           RESTRICT_ON_SEND = %i[create_list].freeze
 
-          # @!method n_times_block_without_arg?(node)
-          def_node_matcher :n_times_block_without_arg?, <<-PATTERN
+          # @!method n_times_block?(node)
+          def_node_matcher :n_times_block?, <<-PATTERN
             (block
               (send (int _) :times)
-              (args)
               ...
             )
+          PATTERN
+
+          # @!method n_times_block_with_arg_and_used?(node)
+          def_node_matcher :n_times_block_with_arg_and_used?, <<-PATTERN
+            (block
+              (send (int _) :times)
+              (args (arg _value))
+                `_value
+            )
+          PATTERN
+
+          # @!method arguments_include_method_call?(node)
+          def_node_matcher :arguments_include_method_call?, <<-PATTERN
+            (send ${nil? #factory_bot?} :create (sym $_) `$(send ...))
           PATTERN
 
           # @!method factory_call(node)
@@ -54,7 +73,10 @@ module RuboCop
 
           def on_block(node)
             return unless style == :create_list
-            return unless n_times_block_without_arg?(node)
+
+            return unless n_times_block?(node)
+            return if n_times_block_with_arg_and_used?(node)
+            return if arguments_include_method_call?(node.body)
             return unless contains_only_factory?(node.body)
 
             add_offense(node.send_node, message: MSG_CREATE_LIST) do |corrector|
--- a/spec/rubocop/cop/rspec/factory_bot/create_list_spec.rb
+++ b/spec/rubocop/cop/rspec/factory_bot/create_list_spec.rb
@@ -53,9 +53,26 @@ RSpec.describe RuboCop::Cop::RSpec::FactoryBot::CreateList do
       RUBY
     end
 
-    it 'ignores n.times with argument' do
+    it 'ignores n.times with n as argument' do
       expect_no_offenses(<<~RUBY)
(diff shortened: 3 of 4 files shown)


spec/rubocop/cop/rspec/factory_bot/create_list_spec.rb
```diff
@@ -55,7 +55,13 @@
 
     it 'ignores n.times with argument' do
```
pirj, line 56: Unfortunately, we have no coverage for the case we pass an argument, but it not a method call, do you mind adding a couple of such cases, probably as a single example?
```ruby
3.times { |n| create :user, :active }
3.times { |n| create :user, password: '123' }
3.times { |n| create :user, :active, password: '123' }
```
```ruby
    it "flags n.times when create call doesn't have method calls" do
      expect_offense(<<~RUBY)
        3.times { |n| create :user, :active }
        ^^^^^^^ Prefer create_list.
        3.times { |n| create :user, password: '123' }
        ^^^^^^^ Prefer create_list.
        3.times { |n| create :user, :active, password: '123' }
        ^^^^^^^ Prefer create_list.
      RUBY
    end
```
fails.

spec/rubocop/cop/rspec/factory_bot/create_list_spec.rb
```diff
@@ -55,7 +55,13 @@
...
       expect_no_offenses(<<~RUBY)
-        3.times { |n| create :user, created_at: n.days.ago }
+        3.times { |n| create :user, position: n }
+      RUBY
+    end
+
+    it 'ignores when a method is called in factory arguments' do
+      expect_no_offenses(<<~RUBY)
+        3.times { create :user, position: rand }
```
pirj, line 64: This is where it gets tricky.
There can be a positional argument, and a hash argument with a method call.

I suggest using child argument matching for that, e.g.:
```
def_node_matcher :has_method_call_in_hash?, '(hash <(pair _ send) ...>)'
def_node_matcher :has_method_call_in_positional_args?, '<send ...>'
```

pirj: A random example you can use for inspiration https://github.com/rubocop-hq/rubocop-rspec/pull/804/files#diff-321b4c4eae6bca30faa33539f9cfb4accf057592521e9a9eaf6a897001c349a8R42

pirj (commented):
Awesome!
I've left a couple of notes.

lib/rubocop/cop/rspec/factory_bot/create_list.rb
```diff
@@ -50,6 +60,7 @@ class CreateList < Base
           def on_block(node)
             return unless style == :create_list
             return unless n_times_block_without_arg?(node)
+            return if block_contains_method_calls?(node.body)
```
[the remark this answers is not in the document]
ngouy: Im just not familiar yet with nodes and pattern matching, all improvements are welcome

spec/rubocop/cop/rspec/factory_bot/create_list_spec.rb
ngouy: why should we differenciate them if anyway they will both discard a potential offence detection

spec/rubocop/cop/rspec/factory_bot/create_list_spec.rb
```diff
@@ -55,7 +55,13 @@
 
     it 'ignores n.times with argument' do
       expect_no_offenses(<<~RUBY)
-        3.times { |n| create :user, created_at: n.days.ago }
```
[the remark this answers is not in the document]
ngouy: the issue I have with this one is that it's unclear if it passes through due to the usage of n or the usage of a method

spec/rubocop/cop/rspec/factory_bot/create_list_spec.rb
ngouy: indeed my bad

lib/rubocop/cop/rspec/factory_bot/create_list.rb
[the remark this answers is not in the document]
pirj: No worries, I'm just suggesting a better name for this node matcher, no implementation changes needed.

spec/rubocop/cop/rspec/factory_bot/create_list_spec.rb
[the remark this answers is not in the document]
pirj: Add a balancing example to make it clear. Do you have this in mind?
```ruby
3.times { |n| create :user, repositories_count: n }
```

spec/rubocop/cop/rspec/factory_bot/create_list_spec.rb
pirj: Aha! I've missed that you're using https://docs.rubocop.org/rubocop-ast/node_pattern.html#for-descendants 👍 

All good and I'm happy if the cop detects:
```ruby
        3.times { |n| create :user, :active, position: 123 }
```
and ignores:
```ruby
        3.times { |n| create :user, :active, position: n }
```

pirj:
Ping, @ngouy. Would you like to revive this PR?

ngouy:
I think so. Give me one week or so.
Pretty busy RN but next week should be good

spec/rubocop/cop/rspec/factory_bot/create_list_spec.rb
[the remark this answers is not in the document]
ngouy: we already have that line 56

ngouy:
we should be good @pirj

spec/rubocop/cop/rspec/factory_bot/create_list_spec.rb
ngouy: that's what the specs are describing so we should be good

spec/rubocop/cop/rspec/factory_bot/create_list_spec.rb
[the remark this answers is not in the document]
pirj: Alright, the old
```ruby
3.times { |n| create :user, position: n.days.ago }
```
passes with no offences, too.

docs/modules/ROOT/pages/cops_rspec_factorybot.adoc
```diff
@@ -79,8 +79,14 @@ This cop can be configured using the `EnforcedStyle` option
 # good
 create_list :user, 3
 
+# bad
+3.times { create :user, age: 18 }
+
 # good
```
pirj, line 85: I suggest clarifying this as:
```ruby
# good - index is used to alter the created models attributes
```

pirj (approved):
Thank you!

bquorning (approved):

PR #1109, opened 2020-12-14
Update `RSpec/Focus` to have auto-correction.

dvandersluis:
`RSpec/Focus` did not have autocorrection so I've added it now. It will remove `focus: true` and `:focus`, and replace `fdescribe`, `fit`, etc. with the corresponding non-focused method.

Since the methods can be defined by the user, there's a small conceit that it looks for a prefix of `f` and only corrects if there is a parallel "regular" method. This means that `focus 'test' do; end` will not be corrected. However, an offense will still be registered, so I think this is okay. Happy to make changes if anyone has a suggestion.

---

* [x] Added tests.
* [x] Updated documentation.

### DIFF
--- a/spec/rubocop/cop/rspec/focus_spec.rb
+++ b/spec/rubocop/cop/rspec/focus_spec.rb
@@ -39,6 +39,26 @@ RSpec.describe RuboCop::Cop::RSpec::Focus do
       pending 'test', meta: true, focus: true do; end
                                   ^^^^^^^^^^^ Focused spec found.
     RUBY
+
+    expect_correction(<<-RUBY)
+      example 'test', meta: true do; end
+      xit 'test', meta: true do; end
+      describe 'test', meta: true do; end
+      RSpec.describe 'test', meta: true do; end
+      it 'test', meta: true do; end
+      xspecify 'test', meta: true do; end
+      specify 'test', meta: true do; end
+      example_group 'test', meta: true do; end
+      scenario 'test', meta: true do; end
+      xexample 'test', meta: true do; end
+      xdescribe 'test', meta: true do; end
+      context 'test', meta: true do; end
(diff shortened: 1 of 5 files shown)


spec/rubocop/cop/rspec/focus_spec.rb
```diff
@@ -137,6 +182,18 @@
...
+      context 'test' do; end
+      it 'test' do; end
+      scenario 'test' do; end
+      example 'test' do; end
+      specify 'test' do; end
+      focus 'test' do; end
+    RUBY
   end
 
   it 'flags rspec example blocks that include `:focus` preceding a hash' do
```
pirj, line 199: Does it correct this as well?

spec/rubocop/cop/rspec/focus_spec.rb
```diff
@@ -137,6 +182,18 @@
...
+    expect_correction(<<-RUBY)
+      describe 'test' do; end
+      RSpec.describe 'test' do; end
+      feature 'test' do; end
+      context 'test' do; end
+      it 'test' do; end
+      scenario 'test' do; end
+      example 'test' do; end
+      specify 'test' do; end
+      focus 'test' do; end
```
pirj, line 195: 👍 for making sure it's not `ocus`!

pirj (commented):
Thank you!

spec/rubocop/cop/rspec/focus_spec.rb
dvandersluis: Oops! Missed that one, updated now.

pirj (approved):
💯

Darhazer (approved):

dvandersluis:
Thanks @pirj, looking forward to not having to fix these manually haha

PR #1112, opened 2020-12-21
Move non-literal attributes to a block

Darhazer:
Solves #1087 by moving all dynamic attributes out of the `create` call and into a block

This is the first step: tests passing after @pirj confirms that this is a direction we want to go, I'll refactor and open the PR for review

---

* [ ] Added tests.
* [ ] Updated documentation.

### DIFF
--- a/spec/rubocop/cop/rspec/factory_bot/create_list_spec.rb
+++ b/spec/rubocop/cop/rspec/factory_bot/create_list_spec.rb
@@ -102,12 +102,12 @@ RSpec.describe RuboCop::Cop::RSpec::FactoryBot::CreateList do
 
     it 'flags usage of n.times with keyword arguments' do
       expect_offense(<<~RUBY)
-        5.times { create :user, :trait, key: val }
+        5.times { create :user, :trait, key: :val }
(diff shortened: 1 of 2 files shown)


spec/rubocop/cop/rspec/factory_bot/create_list_spec.rb
```diff
@@ -142,6 +142,39 @@
...
     end
+
+    it 'flags usage of n.times with dynamic arguments' do
+      expect_offense(<<~RUBY)
+        3.times { create(:user, foo: rand(2), bar: 'baz', joe: doe) }
+        ^^^^^^^ Prefer create_list.
+      RUBY
+
+      expect_correction(<<~RUBY)
+create_list(:user, 3, bar: 'baz') do |user|
```
pirj, line 153: The behaviour is surprisingly different, see https://github.com/thoughtbot/factory_bot/issues/1444

>  The model is saved when it’s created and then it’s passed to the block. If you want to update the model in the block, call `save` in the block.

So even though it resolves one issue with `rand` not being random across the list of records, but it leaves dirty attributes, which are not persisted.

Issue #918, opened 2020-05-26
RSpec/ScatteredLet autocorrect improvements


Darhazer:
I have played recently with ScatteredLet autocorrects and noticed a couple of things.

First, it moves everything after the first let, which changes the order of lets. It would be better to move it after the last consecutive (non-offending let)

---
It used not to move comments - something that most likely applies to all cops that use the `MoveNode` helper. Note that this applies to inline comments, but also to comments directly before the node to be moved. This was fixed in #1115

PR #1115, opened 2021-01-03
Take into account comments when moving nodes

Darhazer:
Improved the `MoveNode` helper to also move the inlined comment, as well as any consecutive comments that are immediately before the moved node.

Initially, I wanted to use the `CommentsHelp` module from RuboCop, but it had few issues (not working with heredocs and taking the preceding comment even if it is just a comment at the end of a line of another statement), so I copied and adapted the module. I'll open PR(s) with fixes for this in RuboCop as well, and eventually, we may end up adapting the whole MoveNode as a build-in autocorrector, but we will have to keep a copy in rubocop-rspec for a while anyway.

This fixes the second point in #918. The first will be tackled separately.

---

* [x] Added tests.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Master (Unreleased)
 
+* Fix `HooksBeforeExamples`, `LeadingSubject`, `LetBeforeExamples` and `ScatteredLet` autocorrection to take into account inline comments and comments immediately before the moved node. ([@Darhazer][])
+
 ## 2.1.0 (2020-12-17)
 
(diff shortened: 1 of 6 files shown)


pirj (approved):
Plain awesome!

Darhazer:
Resolved the code-climate complaint (and simplified the code a lot, as it turned out there is a much easier way to get the associated comments)

pirj:
> Gem::LoadError: You have already activated rake 13.0.1, but your Gemfile requires rake 13.0.3. Prepending `bundle exec`

This should have fixed it https://github.com/rubocop-hq/rubocop-rspec/pull/1114

bquorning (approved):
🎉

PR #1117, opened 2021-01-10
Imrpove performance by limitting on_send calls

Darhazer:
I noticed that Rubocop, since v0.90, has an [optimization for not calling on_send](https://github.com/rubocop-hq/rubocop/pull/8365) based on the expected method_name for the cop.

I applied that to the cops for it was easy (which is about 50% of the cops). Due to the configurable language, it might be hard to do for some cops, and for others, it just requires more refactoring, as the cops check for a method somewhere down the chain. 

I haven't really measured the performance improvement, but a naive test (running it with `time` over our own specs) shows a significant difference.

cc: @fatkodima

---

* [x] ~Added tests.~

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## Master (Unreleased)
 
 * Fix `HooksBeforeExamples`, `LeadingSubject`, `LetBeforeExamples` and `ScatteredLet` autocorrection to take into account inline comments and comments immediately before the moved node. ([@Darhazer][])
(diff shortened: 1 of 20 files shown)


fatkodima (approved):
👍 Yeah, this definitely should make a noticeable difference.

marcandre:
Perfect!
It's always faster to use `RESTRICT_ON_SEND` whenever possible, no need to benchmark it 👍
Thanks!

PR #1119, opened 2021-01-20
[Fix #647] Add OrderedHooks cop

### LINKED ISSUE #647: Cop idea: ordered hooks
A cop to enforce hook ordering:
before :suite
before :context
before :example
around :example
after  :example
after  :context
after  :suite

Darhazer:
No auto-correction at this point since MoveNode doesn't play nicely with each_cons
Some people may prefer `around` hooks to be before `before` hooks, so making it configurable would also be nice.

---

* [x] Added tests.
* [x] Updated documentation.


docs/modules/ROOT/pages/cops_rspec.adoc
```diff
@@ -2973,6 +2973,47 @@ end
...
+| Yes
+| 2.2
+| -
+|===
+
+Checks that before/around/after hooks are defined in the correct order.
+
+If multiple hooks are defined in example group, they should appear in
+the following order:
+before :suite
```
pirj, line 2992: But `before(:suite)` cannot appear in an example group? https://github.com/rspec/rspec-core/blob/e95cfe3724de9b56f973e38f84fdb1b609d88e01/lib/rspec/core/hooks.rb#L197

docs/modules/ROOT/pages/cops_rspec.adoc
```diff
@@ -2973,6 +2973,47 @@ end
...
+before :context
+before :example
+around :example
+after :example
+after :context
+after :suite
```
pirj, line 2998: Same https://github.com/rspec/rspec-core/blob/e95cfe3724de9b56f973e38f84fdb1b609d88e01/lib/rspec/core/hooks.rb#L274

lib/rubocop/cop/rspec/ordered_hooks.rb
```diff
@@ -0,0 +1,86 @@
...
+      #  after { run_cleanup }
+      #
+      class OrderedHooks < Base
+        extend AutoCorrector
+
+        MSG = '`%<hook>s` is supposed to appear before `%<previous>s`' \
+          ' at line %<line>d.'
+
+        EXPECTED_HOOK_ORDER = %i[before around after].freeze
+        EXPECTED_SCOPE_ORDER = %i[suite context each].freeze
```
pirj, line 34: Maybe refer to `:each` as `:example`?

lib/rubocop/cop/rspec/ordered_hooks.rb
```diff
@@ -0,0 +1,86 @@
...
+
+        def on_block(node)
+          return unless example_group_with_body?(node)
+
+          RuboCop::RSpec::ExampleGroup.new(node)
+            .hooks
+            .each_cons(2) { |previous, current| check_order(previous, current) }
```
pirj, line 41: ❤️

pirj (approved):
Looks great! Thank you.
And sorry for the delay in reviewing.

lib/rubocop/cop/rspec/ordered_hooks.rb
```diff
@@ -0,0 +1,86 @@
...
+      #
+      #  # good
+      #  before { run_setup }
+      #  after { run_cleanup }
+      #
+      class OrderedHooks < Base
+        extend AutoCorrector
+
+        MSG = '`%<hook>s` is supposed to appear before `%<previous>s`' \
+          ' at line %<line>d.'
```
bquorning, line 31:
```suggestion
        MSG = '`%<hook>s` is supposed to appear before `%<previous>s` ' \
          'at line %<line>d.'
```
RuboCop (Parser) unfortunately [cannot check this](https://github.com/rubocop-hq/rubocop/issues/6420).

lib/rubocop/cop/rspec/ordered_hooks.rb
```diff
@@ -0,0 +1,86 @@
...
+      #  # bad
+      #  after { run_cleanup }
+      #  before { run_setup }
+      #
+      #  # good
+      #  before { run_setup }
+      #  after { run_cleanup }
+      #
+      class OrderedHooks < Base
+        extend AutoCorrector
```
bquorning, line 28: This can be removed, right?

docs/modules/ROOT/pages/cops_rspec.adoc
```diff
@@ -2973,6 +2973,47 @@ end
...
 * https://www.rubydoc.info/gems/rubocop-rspec/RuboCop/Cop/RSpec/NotToNot
 
+== RSpec/OrderedHooks
+
+|===
+| Enabled by default | Safe | Supports autocorrection | VersionAdded | VersionChanged
+
+| Pending
+| Yes
+| Yes
```
bquorning, line 2983: The PR description says there's no autocorrection yet.
```suggestion
| No
```

lib/rubocop/cop/rspec/ordered_hooks.rb
```diff
@@ -0,0 +1,86 @@
...
+                           previous: format_hook(previous),
+                           line: previous.to_node.loc.line)
+
+          add_offense(current.to_node.send_node, message: message)
+        end
+
+        def format_hook(hook)
+          msg = hook.name.to_s
+          raw_scope_name = hook.send(:scope_name)
+          msg += "(:#{raw_scope_name})" if raw_scope_name
```
bquorning, line 80: Perhaps add the scope only when the hooks are identical? To avoid messages like “`before(:each)` is supposed to appear before `after(:suite)`” where I think a message like “`before` is supposed to appear before `after`” might be sufficient.

lib/rubocop/cop/rspec/ordered_hooks.rb
```diff
@@ -0,0 +1,86 @@
...
+
+        def check_scope_order(previous, current)
+          previous_idx = EXPECTED_SCOPE_ORDER.index(previous.scope)
+          current_idx = EXPECTED_SCOPE_ORDER.index(current.scope)
+
+          if current.name == :after # for after we expect reversed order
+            order_violation(previous, current) if previous_idx < current_idx
+          elsif previous_idx > current_idx
+            order_violation(previous, current)
+          end
```
bquorning, line 65: I was hoping to get a more symmetric comparison of `previous_idx` with `current_idx`. What do you think about e.g.

```suggestion
          case current.name
          when :after # for after we expect reversed order
            order_violation(previous, current) if previous_idx < current_idx
          else
            order_violation(previous, current) if previous_idx > current_idx
          end
```

CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 
 * Fix `HooksBeforeExamples`, `LeadingSubject`, `LetBeforeExamples` and `ScatteredLet` autocorrection to take into account inline comments and comments immediately before the moved node. ([@Darhazer][])
 * Improve rubocop-rspec performance. ([@Darhazer][])
+* Add `RSpec/OrderdHooks` cop. ([@Darhazer][])
```
DaniPB, line 7:
```suggestion
* Add `RSpec/OrderedHooks` cop. ([@Darhazer][])
```

PR #1128, opened 2021-02-05
[Fix #1127] Allow `RSpec/EmptyExampleGroup` to have example group types that should be allowed specified in its configuration

### LINKED ISSUE #1127: `RSpec/EmptyExampleGroup` allowed example group types
I have a special `endpoint` example group, which I have defined as an example group in `RSpec/Language/ExampleGroups/Regular`. However, it is valid for this to have no body, because the `endpoint` method calls `it_behaves_like`, so there are implicit examples, even if it appears blank.

It'd be great if `RSpec/EmptyExampleGroup` could have a configuration to allow me to ignore certain example group "types" so that this can be ignored while other empty example groups can still register offenses.

eg:

```ruby
RSpec.describe API::FooController do
  endpoint :index, path: '/foo'
end

RSpec.describe API::BarController do
  endpoint :show, path: '/bar/:id' do
    it 'does something'
    # ...
  end
end
```

```
spec/test_spec.rb:1:1: C: RSpec/EmptyExampleGroup: Empty example group detected.
RSpec.describe API::FooController do
```

dvandersluis:
Allows certain example group methods to be excluded from `RSpec/EmptyExampleGroup`. Fixes #1127.

---

* [x] Added tests.
* [x] Updated documentation.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Master (Unreleased)
 
+* Allow `RSpec/EmptyExampleGroup` to have example group types that should be allowed specified in its configuration. ([@dvandersluis][])
+
 ## 2.2.0 (2021-02-02)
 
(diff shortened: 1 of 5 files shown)


dvandersluis:
@pirj suggested a better fix in the issue, so I'm closing.

PR #1129, opened 2021-02-07
Fix a false negative for `RSpec/FilePath` when spec file has multiple `describe`.

### LINKED ISSUE #1106: RSpec/FilePath: false negative
The cop fails to detect an offence in:
```ruby
RSpec.describe Object, "should" do
```
and
```ruby
RSpec.describe Object, "#should" do
```
when inspecting `spec/rspec/expectations/extensions/kernel_spec.rb`.
Source: https://github.com/rubocop-hq/rubocop-rspec/pull/1099

M-Yamashita01:
Fix #1106 

RSpec/FilePath does not register offenses when the spec file has multiple `describe`, such as the following code.

```
describe MyClass, '#foo' do
end

describe MyClass, '#bar' do
end
```

But the spec file has a single ‘describe’, registers offenses.

```
describe MyClass, '#foo' do
^^^^^^^^^^^^^^^^^^^^^^^^ Spec path should end with `my_class*foo*_spec.rb`.
end
```

I think this is a false negative, so could you please approve this PR?

---

* [x] Added tests.
* [ ] Updated documentation.

### DIFF
--- a/lib/rubocop/cop/rspec/file_path.rb
+++ b/lib/rubocop/cop/rspec/file_path.rb
@@ -70,8 +70,6 @@ module RuboCop
         def_node_search :routing_metadata?, '(pair (sym :type) (sym :routing))'
 
         def on_top_level_example_group(node)
-          return unless top_level_groups.one?
-
           const_described(node) do |send_node, described_class, arguments|
             next if routing_spec?(arguments)
 
--- a/spec/rubocop/cop/rspec/file_path_spec.rb
+++ b/spec/rubocop/cop/rspec/file_path_spec.rb
@@ -71,7 +71,8 @@ RSpec.describe RuboCop::Cop::RSpec::FilePath do
     RUBY
   end
 
-  it 'registers an offense for a bad path with multiple describe for instance methods' do
+  it 'registers an offense for a bad path' \
+        'with multiple describe for instance methods' do
     expect_offense(<<-RUBY, 'wrong_class_foo_spec.rb')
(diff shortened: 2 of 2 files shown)


pirj:
What would be the suggested file path in this case?

pirj:
I suggest checking the diff of offences resulted by this change against https://github.com/pirj/real-world-rspec, and specifically `rspec-core` and `rspec-expectations` repos, to my best memory they have had several top-level describes in the same file.

Basically, this change would prevent from having multiple top-level example groups. Some prefer to;
```ruby
RSpec.describe A, '#foo' do
end
RSpec.describe A, '#bar do
end
```
and others
```
RSpec.describe A do
  describe '#foo' do
  end
  describe '#bar' do
  end
end
```

I believe both styles are ok.
Even if we want to provide a toolbox to enforce consistency between those two styles, this is unlikely a job for `RSpec/FilePath` cop.

M-Yamashita01:
Thank you for pointing this out.

> I believe both styles are ok.
Even if we want to provide a toolbox to enforce consistency between those two styles, this is unlikely a job for RSpec/FilePath cop.

You are totally right.
I'm sorry, I should have discussed this in the issue beforehand.
So, could I close this Pull Request because we don't think it's necessary?

pirj:
Thank you for the time you put in it.

I guess the PR can be closed. And a PR with a cop that would suggest combining/splitting top-level groups with method specifiers is welcome!

PR #1132, opened 2021-02-13
Add new `RSpec/IdenticalEqualityAssertion` cop

### LINKED ISSUE #1130: Cop idea: Identical equality assertion
An assertion such as this is probably a typo or a flawed test (since it can never fail):

`expect(foo.bar).to eq(foo.bar)`

tejasbubane:
Closes #1130

---

* [x] Added tests.
* [x] Updated documentation.


lib/rubocop/cop/rspec/identical_equality_assertion.rb
```diff
@@ -0,0 +1,35 @@
...
+      #
+      #   # good
+      #   expect(foo.bar).to eq(2)
+      #   expect(foo.bar).to eql(2)
+      #
+      class IdenticalEqualityAssertion < Base
+        MSG = 'Identical equality can never fail.'
+
+        def_node_matcher :equality_check?, <<~PATTERN
+          (send (send nil? :expect $_) :to (send nil? {:eql :eq} $_))
```
pirj, line 22: I suggest adding `be ==` and `be` here. This is optional though, `eq` will cover the majority of cases. Just afraid of people abusing alternative matches to make this pass.

lib/rubocop/cop/rspec/identical_equality_assertion.rb
```diff
@@ -0,0 +1,35 @@
...
+        PATTERN
+
+        def on_send(node)
```
pirj, line 25: Does it make sense to use the `RESTRICT_ON_SEND` trick here?

lib/rubocop/cop/rspec/identical_equality_assertion.rb
```diff
@@ -0,0 +1,35 @@
...
+      #   # bad
+      #   expect(foo.bar).to eq(foo.bar)
+      #   expect(foo.bar).to eql(foo.bar)
+      #
+      #   # good
+      #   expect(foo.bar).to eq(2)
+      #   expect(foo.bar).to eql(2)
+      #
+      class IdenticalEqualityAssertion < Base
+        MSG = 'Identical equality can never fail.'
```
pirj, line 19: This is not correct.
`randomizer.next_rand == randomizer.next_rand` has a chance to fail.
And `factorial.next == factorial.next` will only pass for the first two calls.

Do you mind to change the wording?

pirj (commented):
Since it's a new cop, do you mind to run it against https://github.com/pirj/real-world-rspec?
Usually, I remove all `.rubocop.yml` files in it manually. And each time I wish `rubocop` had an option to ignore config files in the source code.

Looks good in general.
Just a small note regarding the offence message and sanity check against popular repos using RSpec.

Thank you for another contribution!

lib/rubocop/cop/rspec/identical_equality_assertion.rb
Darhazer: :100:

lib/rubocop/cop/rspec/identical_equality_assertion.rb
andyw8: How about:

`Identical expressions on both sides of the equality may indicate a flawed test`.

pirj: 💯

pirj (approved):
Perfect. Thank you!

tejasbubane:
@pirj I ran this cop on `real-world-rspec` and found these offenses:
https://gist.github.com/tejasbubane/e64636d7c7dd97c4b056a09b09a6c61b

pirj:
@tejasbubane 99% look completely legit!

```ruby
expect(true).to eq(true)
```
:laughing: 

Some are (`rspec-core`) testing the RSpec itself, but it's ok to flag them:
```ruby
it("equality") { expect(1).to eq(1) }
```

The only one that is probably a false offence:
```ruby
expect(fqdn_rand(3)).to eql(fqdn_rand(3))
```

Thank you!

BTW, did you feel the pain of removing `.rubocop` YAML files from that repo? :smile:

tejasbubane:
> BTW, did you feel the pain of removing .rubocop YAML files from that repo? 

No I ran with `--only RSpec/IdenticalEqualityAssertion` :)

lib/rubocop/cop/rspec/identical_equality_assertion.rb
```diff
@@ -0,0 +1,39 @@
...
+          (send (send nil? :expect $_) :to
+            {(send nil? {:eql :eq :be} $_)
+             (send (send nil? :be) :== $_)})
+        PATTERN
+
+        def on_send(node)
+          equality_check?(node) do |left, right|
+            return unless left == right
+
+            add_offense(node)
```
bquorning, line 33:
```suggestion
            add_offense(node) if left == right
```

spec/rubocop/cop/rspec/identical_equality_assertion_spec.rb
```diff
@@ -0,0 +1,62 @@
+# frozen_string_literal: true
+
+RSpec.describe RuboCop::Cop::RSpec::IdenticalEqualityAssertion do
+  let(:msg) do
+    'Identical expressions on both sides of the equality '\
+    'may indicate a flawed test.'
+  end
```
bquorning, line 7:
```suggestion
  let(:msg)  { described_class::MSG }
```

But I actually prefer having the message spelled out in each example. Yes, it's duplication, but a spec without method calls and interpolation is faster/easier to understand.

bquorning (commented):
Good stuff. Only nitpick comments.

spec/rubocop/cop/rspec/identical_equality_assertion_spec.rb
tejasbubane: I think we have an internal cop that prevents this:

```
spec/rubocop/cop/rspec/identical_equality_assertion_spec.rb:4:15: C: InternalAffairs/UselessMessageAssertion: Do not specify cop behavior using described_class::MSG.
  let(:msg) { described_class::MSG }
              ^^^^^^^^^^^^^^^^^^^^
```

tejasbubane:
@bquorning Thanks. Fixed as per your comments - except one.

spec/rubocop/cop/rspec/identical_equality_assertion_spec.rb
bquorning: >But I actually prefer having the message spelled out in each example.

I'm not sure if you saw that part of my comment 😅

pirj (approved):
Thanks!

spec/rubocop/cop/rspec/identical_equality_assertion_spec.rb
```diff
@@ -0,0 +1,62 @@
...
+
+  it 'registers an offense when using identical expressions with `eq`' do
+    expect_offense(<<~RUBY)
+      expect(foo.bar).to eq(foo.bar)
+      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ #{msg}
```
bquorning, line 12: I prefer not having the message interpolated in the specs – I also don't think we do it in other specs.

What do you think about e.g.
```suggestion
      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ Identical expressions on both sides of the equality may indicate a flawed test.
```
?

pirj:
>>> ran this cop on real-world-rspec and found these offenses:

>> BTW, did you feel the pain of removing .rubocop YAML files from that repo?

> No I ran with --only RSpec/IdenticalEqualityAssertion :)

Do you mind sharing how you do it, @tejasbubane ?

```
~/source/rubocop-rspec $ rubocop --only RSpec/Rails/AvoidSetupHook ../real-world-rspec/
Error: RuboCop found unsupported Ruby version 2.4 in `TargetRubyVersion` parameter (in /Users/pirj/source/real-world-rspec/fat_free_crm/.rubocop.yml). 2.4-compatible analysis was dropped after version 1.12.
Supported versions: 2.5, 2.6, 2.7, 3.0, 3.1
```
same for `real-world-rspec/camaleon-cms/.rubocop.yml`
After removing them:
```
Error: The `Layout/AlignParameters` cop has been renamed to `Layout/ParameterAlignment`.
(obsolete configuration found in /Users/pirj/source/real-world-rspec/rspec-mocks/.rubocop_rspec_base.yml, please update it)
The `Layout/IndentArray` cop has been renamed to `Layout/FirstArrayElementIndentation`.
(obsolete configuration found in /Users/pirj/source/real-world-rspec/rspec-mocks/.rubocop_rspec_base.yml, please update it)
The `Layout/IndentAssignment` cop has been renamed to `Layout/AssignmentIndentation`.
(obsolete configuration found in /Users/pirj/source/real-world-rspec/rspec-mocks/.rubocop_rspec_base.yml, please update it)
The `Layout/IndentHeredoc` cop has been renamed to `Layout/HeredocIndentation`.
(obsolete configuration found in /Users/pirj/source/real-world-rspec/rspec-mocks/.rubocop_rspec_base.yml, please update it)
The `Style/TrailingCommaInLiteral` cop has been removed. Please use `Style/TrailingCommaInArrayLiteral` and/or `Style/TrailingCommaInHashLiteral` instead.
(obsolete configuration found in /Users/pirj/source/real-world-rspec/rspec-mocks/.rubocop_rspec_base.yml, please update it)
The `Style/MethodMissing` cop has been split into `Style/MethodMissingSuper` and `Style/MissingRespondToMissing`.
(obsolete configuration found in /Users/pirj/source/real-world-rspec/rspec-mocks/.rubocop_rspec_base.yml, please update it)
```

I have those one-liners handy:
```
fd --no-ignore --hidden 'rubocop*.yml' | xargs rm
```
and
```
for repo in $(fd --type d --maxdepth 1); do; (cd "$repo"; git reset --hard HEAD); done
```
But I'd rather run `rubocop --only` and skip those unrelated errors.

spec/rubocop/cop/rspec/identical_equality_assertion_spec.rb
tejasbubane: Changed.

tejasbubane:
@pirj Not sure how I ran with `--only` back then, but this worked when I tried now:

```
$ cat Gemfile     
source 'https://rubygems.org'

gem 'rubocop', path: '../rubocop'
gem 'rubocop-rspec', path: '../rubocop-rspec' # Contains newly built / unreleased cops
```
```
$ cat .rubocop.yml
require:
  - rubocop-rspec

AllCops:
  DisabledByDefault: true

RSpec/IdenticalEqualityAssertion:
  Enabled: true
```
```
$ bundle exec rubocop --config .rubocop.yml   
.....
54137 files inspected, 115 offenses detected

pirj:
Thanks, that works!

bquorning (approved):

PR #1133, opened 2021-02-22
add RSpec/PreferBeforeOverSetup cop

paydaylight:
Added new cop `RSpec/PreferBeforeOverSetup` that checks that rspec's `before` method is used over rspec-rails' `setup`.

* It helps tests be consistent by eliminating `setup` calls without arguments that are basically `before` method calls.
* Setup allows expectations inside a block that are counterintuitive and make tests do two things at a time which is (bad)[https://www.betterspecs.org/#single].

---

* [ ] Added tests.
* [ ] Updated documentation.


lib/rubocop/cop/rspec/prefer_before_over_setup.rb
```diff
@@ -0,0 +1,44 @@
...
+      #   # good
+      #   before do
+      #     allow(foo).to receive(:bar)
+      #   end
+      #
+      class PreferBeforeOverSetup < Base
+        extend AutoCorrector
+
+        MSG = 'Use `before` instead of `setup`.'
+        SPACE = ' '
```
pirj, line 24: I believe this is unnecessary with `frozen_string_literal`.

lib/rubocop/cop/rspec/prefer_before_over_setup.rb
```diff
@@ -0,0 +1,44 @@
...
+      #   setup do
+      #     allow(foo).to receive(:bar)
+      #   end
+      #
+      #   # good
+      #   before do
+      #     allow(foo).to receive(:bar)
+      #   end
+      #
+      class PreferBeforeOverSetup < Base
```
pirj, line 20: Maybe `UseOfSetupHook`?
Reasoning: https://github.com/rubocop-hq/rubocop/issues/7077

lib/rubocop/cop/rspec/prefer_before_over_setup.rb
```diff
@@ -0,0 +1,44 @@
...
+          (block
+            (send _ :setup)
+            (args) _)
+        PATTERN
+
+        def on_block(node)
+          return unless setup_call?(node)
+
+          add_offense(node) do |corrector|
+            setup_block_parts = node.source.split(/ /)
```
pirj, line 36: I believe if you [captured](https://docs.rubocop.org/rubocop-ast/node_pattern.html#for-captures) the `:setup` with `$:setup`, you could just `corrector.replace(setup, 'before'`).

spec/rubocop/cop/rspec/prefer_before_over_setup_spec.rb
```diff
@@ -0,0 +1,49 @@
...
+      expect_offense(<<~RUBY)
+        setup do
+        ^^^^^^^^ Use `before` instead of `setup`.
+          allow(foo).to receive(:bar)
+        end
+      RUBY
+    end
+  end
+
+  context 'when using curly braces block syntax' do
```
pirj, line 23: As far as I'm aware, there's no difference between those two from AST standpoint. And it doesn't make sense to cover them both.

spec/rubocop/cop/rspec/prefer_before_over_setup_spec.rb
```diff
@@ -0,0 +1,49 @@
...
+        }
+      RUBY
+    end
+  end
+
+  context 'when autocorrecting offenses' do
+    it 'replaces `setup` with `before`' do
+      expect(
+        autocorrect_source('setup do foo = 1 end')
+      ).to eq('before do foo = 1 end')
```
pirj, line 46: Please feel free to put this into the same example with `expect_offense`.

spec/rubocop/cop/rspec/prefer_before_over_setup_spec.rb
```diff
@@ -0,0 +1,49 @@
...
+          allow(foo).to receive(:bar)
+        }
+      RUBY
+    end
+  end
+
+  context 'when autocorrecting offenses' do
+    it 'replaces `setup` with `before`' do
+      expect(
+        autocorrect_source('setup do foo = 1 end')
```
pirj, line 45: You may use `expect_correction(<<-RUBY)` that accepts a "good" source as its argument and uses the "bad" from the previously called `expect_offense`.

pirj (commented):
> Setup allows expectations inside a block

`before` allows that as well. :shushing_face: 

Thanks for the contribution. A few small tweaks and I'm happy to accept it!

lib/rubocop/cop/rspec/prefer_before_over_setup.rb
```diff
@@ -0,0 +1,44 @@
+# frozen_string_literal: true
+
+module RuboCop
+  module Cop
+    module RSpec
+      # Checks that tests use rspec's `before` over rspec-rails' `setup` method.
```
pirj, line 6: s/rspec/RSpec/

pirj:
CircleCI status reports for partially degraded service.

lib/rubocop/cop/rspec/prefer_before_over_setup.rb
paydaylight: what about `AvoidSetupHook`?

spec/rubocop/cop/rspec/prefer_before_over_setup_spec.rb
paydaylight: I was thinking about ensuring that implementation stays the same but maybe it is redundant.

spec/rubocop/cop/rspec/avoid_setup_hook_spec.rb
```diff
@@ -0,0 +1,39 @@
...
+        end
+      RUBY
+
+      expect_correction(<<~RUBY)
+        before do
+          allow(foo).to receive(:bar)
+        end
+      RUBY
+    end
+  end
```
pirj, line 38: I believe there's not much of an added value of having this in a separate context.
Just move `expect_correction` to `registers an offense on setup call` and it'll do it.
That is, to my best memory, how most of our specs are written.

lib/rubocop/cop/rspec/prefer_before_over_setup.rb
paydaylight: @pirj not quite got this point

pirj: Please capitalize "RSpec".

lib/rubocop/cop/rspec/prefer_before_over_setup.rb
paydaylight: @pirj For it to work you need to capture the whole node (`$:setup` captures `:send` symbol), anyway thanks for the tip!

CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 ## Master (Unreleased)
 
+* Add `RSpec/PreferBeforeOverSetup` cop. ([@paydaylight][])
```
pirj, line 5: AvoidSetupHook

config/default.yml
```diff
@@ -106,6 +106,12 @@ RSpec/AroundBlock:
   VersionAdded: '1.11'
   StyleGuide: https://www.rubydoc.info/gems/rubocop-rspec/RuboCop/Cop/RSpec/AroundBlock
 
+RSpec/AvoidSetupHook:
+  Description: Checks that tests use RSpec's `before` over rspec-rails' `setup` method.
```
pirj, line 110: `setup` is not an `rspec-rails`', but it rather comes from Rails itself.
See https://github.com/rspec/rspec-rails/blob/ee251b792646beb239f8b614fdc13ec111885a66/lib/rspec/rails/adapters.rb#L106

There might also be a `teardown do` hook that should canonically be defined as `after do`.

pirj (approved):
Looks great, thank you!

config/default.yml
bquorning: rspec-rails or rails proper, doesn't this cop belong in our `RSpec::Rails` namespace?

pirj:
Rebased, moved to `Rails` namespace, changed `VersionAdded` to 2.4, adjusted the Changelog, squashed commits.

pirj:
> * ~Setup allows expectations inside a block that are counterintuitive and make tests do two things at a time which is [bad](https://www.betterspecs.org/#single).

We're more relying on the [the Community RSpec style guide](https://rspec.rubystyle.guide/#expectation-per-example) that has this rule with single expectation relaxed. See https://github.com/rubocop/rspec-style-guide/pull/47 for more info.

Also, `before` hook also allows for expectations inside a block.

pirj:
Checked against [`real-world-rspec`](https://github.com/pirj/real-world-rspec):

```
54136 files inspected, 47 offenses detected, 47 offenses auto-correctable
```

Two offences were off:
```
/Users/pirj/source/real-world-rspec/huginn/spec/lib/huginn_scheduler_spec.rb:8:5: C: [Correctable] RSpec/Rails/AvoidSetupHook: Use before instead of setup.
    stub(@scheduler).setup {}
```
```
/Users/pirj/source/real-world-rspec/rspec-rails/spec/rspec/rails/setup_and_teardown_adapter_spec.rb:15:7: C: [Correctable] RSpec/Rails/AvoidSetupHook: Use before instead of setup.
      group.setup { "baz" }
```

Fixed.

bquorning (approved):

pirj:
Thank you, @paydaylight !

pirj:
BTW would you like to hack on `RSpec/Rails/AvoidTeardownHook` or extend and rename `AvoidSetupHook` to something more generic before we even released it?

PR #1135, opened 2021-03-22
Allow `RSpec/ContextWording` to accept multi-word prefixes

hosamaly:
Allow the `ContextWording` cop to accept multi-word prefixes.

The new logic uses a regular expression. To preserve existing behaviour, configured prefixes are escaped.

---

* [x] Added tests.
* [x] Updated documentation.
* [x] The build (`bundle exec rake`) passes. (Waiting for CI)

### DIFF
--- a/lib/rubocop/cop/rspec/context_wording.rb
+++ b/lib/rubocop/cop/rspec/context_wording.rb
@@ -8,6 +8,7 @@ module RuboCop
       # The default list of prefixes is minimal. Users are encouraged to tailor
       # the configuration to meet project needs. Other acceptable prefixes may
       # include `if`, `unless`, `for`, `before`, `after`, or `during`.
+      # They may consist of multiple words if desired.
       #
       # @see https://rspec.rubystyle.guide/#context-descriptions
       # @see http://www.betterspecs.org/#contexts
@@ -52,7 +53,7 @@ module RuboCop
         private
 
         def bad_prefix?(description)
-          !prefixes.include?(description.split(/\b/).first)
+          !prefix_regex.match?(description)
         end
 
         def joined_prefixes
@@ -65,6 +66,10 @@ module RuboCop
 
         def prefixes
           cop_config['Prefixes'] || []
+        end
+
+        def prefix_regex
+          /^#{Regexp.union(prefixes)}\b/
         end
       end
     end
--- a/spec/rubocop/cop/rspec/context_wording_spec.rb
+++ b/spec/rubocop/cop/rspec/context_wording_spec.rb
@@ -95,5 +95,35 @@ RSpec.describe RuboCop::Cop::RSpec::ContextWording do
         end
       RUBY
     end
+
+    context 'with a multi-word prefix' do
+      let(:cop_config) { { 'Prefixes' => ['assuming that'] } }
+
+      it 'skips descriptions with allowed multi-word prefixes' do
+        expect_no_offenses(<<-RUBY)
+          context 'assuming that display name is present' do
+          end
+        RUBY
+      end
+    end
+
+    context 'with special regex characters' do
+      let(:cop_config) { { 'Prefixes' => ['a$b\d'] } }
+
+      it 'matches the full prefix' do
+        expect_offense(<<-RUBY)
+          context 'a' do
+                  ^^^ Start context description with 'a$b\\d'.
+          end
+        RUBY
+      end
+
+      it 'matches special characters' do
+        expect_no_offenses(<<-RUBY)
+          context 'a$b\\d something' do
+          end
+        RUBY
+      end
+    end
   end
 end
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Master (Unreleased)
 
+* Allow `RSpec/ContextWording` to accept multi-word prefixes. ([@hosamaly][])
+
 ## 2.2.0 (2021-02-02)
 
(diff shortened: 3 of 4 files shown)


lib/rubocop/cop/rspec/context_wording.rb
```diff
@@ -52,7 +53,7 @@ def on_block(node)
         private
 
         def bad_prefix?(description)
-          !prefixes.include?(description.split(/\b/).first)
+          !prefix_regex.match?(description)
```
pirj, line 56: Maybe `prefixes.none? { |prefix| description.start_with?(prefix) }`?

hosamaly: That's what I did at first, but it could then match part of a word, which would cause a failure in the following test case:
https://github.com/rubocop/rubocop-rspec/blob/1d957cc2c00492acd649073aa9f5e10b07c4a4ed/spec/rubocop/cop/rspec/context_wording_spec.rb#L43-L49

pirj: Got it 👍 
`prefixes.none? { |prefix| description.start_with?("#{prefix} ") }` then?

hosamaly: That would fail when the prefix is followed by a character other than space, such as the following test case:

https://github.com/rubocop/rubocop-rspec/blob/1d957cc2c00492acd649073aa9f5e10b07c4a4ed/spec/rubocop/cop/rspec/context_wording_spec.rb#L36-L41

pirj: I guess there's no way around regexp then 😄

spec/rubocop/cop/rspec/context_wording_spec.rb
```diff
@@ -95,5 +95,35 @@
...
+    end
+
+    context 'with special regex characters' do
+      let(:cop_config) { { 'Prefixes' => ['a$b\d'] } }
+
+      it 'matches the full prefix' do
+        expect_offense(<<-RUBY)
+        context 'a' do
+                ^^^ Start context description with 'a$b\\d'.
+        end
```
pirj, line 117: Please indent this and the following a bit more to indicate that this is inside heredoc. Subjectively it's slightly easier to read this way.

lib/rubocop/cop/rspec/context_wording.rb
```diff
@@ -66,6 +67,12 @@ def joined_prefixes
         def prefixes
           cop_config['Prefixes'] || []
         end
+
+        def prefix_regex
+          escaped_prefixes =
+            prefixes.map { |prefix| "#{Regexp.escape(prefix)}\\b" }
+          /^#{escaped_prefixes.join('|')}/
```
pirj, line 74: In theory there's `Regexp.union`, but with `escape`, `/\b/` and `/^/` needed to be added it seems the current approach is unbeatable.

pirj (approved):
Thank you!

pirj:
@Darhazer I can't invite you as a reviewer for some reason.

spec/rubocop/cop/rspec/context_wording_spec.rb
hosamaly: Done.

lib/rubocop/cop/rspec/context_wording.rb
Darhazer: start_with? also accepts regexp, as well as multiple arguments, so I was just wondering as well, why we don't use it

Darhazer (approved):

Darhazer:
@pirj I lost my write access (perhaps due to the migration to the rubocop organisation)

pirj:
@bquorning Can you please bring Maxim back to the team?

lib/rubocop/cop/rspec/context_wording.rb
hosamaly: @Darhazer `start_with?` accepts regular expressions in Ruby 2.5, but not in Ruby 2.4.

lib/rubocop/cop/rspec/context_wording.rb
bquorning, line 74: Don't ignore your intuition, @pirj. `Regexp.union` indeed seems to do escaping, and it adds a `(?: … )` group around the result, so the `/\b/` can be moved to the last line.
```suggestion
          /^#{Regexp.union(prefixes)}\b/
```

hosamaly: Thanks for the suggestion, @bquorning. I had tried to do something similar but failed because I kept the word separator (`\b`) within the escaped prefixes before calling `union`. Yours is much cleaner. Thanks!

bquorning (approved):

lib/rubocop/cop/rspec/context_wording.rb
hosamaly: Resolved using @bquorning's suggestion `/^#{Regexp.union(prefixes)}\b/`

pirj:
Thanks for your contribution, @hosamaly !

lib/rubocop/cop/rspec/context_wording.rb
pirj: > Don't ignore your intuition

That's my primary development driver 😄

Darhazer:
@bquorning sorry for pinging again, somehow I missed the invitation and not it's expired :cry:

PR #1138, opened 2021-04-13
Add new RSpec/DescribeTopLevelGroup cop.

M-Yamashita01:
Added new cop `RSpec/DescribeTopLevelGroup` that checks that the multiple `describe` is split or combined.

In the previous pull request I made, there was a comment asking me to make the cop, so I made this pull request.  
https://github.com/rubocop/rubocop-rspec/pull/1129#issuecomment-775117552

`Splitting: describe_top_level_group`

```
# bad
RSpec.describe MyClass do
  describe '#foo' do
    subject { my_class_instance.foo }
  end

  describe '#bar' do
    subject { my_class_instance.bar }
  end
end

# good
RSpec.describe MyClass do
  describe '#foo' do
    subject { my_class_instance.foo }
  end
end

RSpec.describe MyClass do
  describe '#bar' do
    subject { my_class_instance.bar }
  end
end
```

`Combining: describe_top_level_group`

```
# bad
RSpec.describe MyClass do
  describe '#foo' do
    subject { my_class_instance.foo }
  end
end

RSpec.describe MyClass do
  describe '#bar' do
    subject { my_class_instance.bar }
  end
end

# good
RSpec.describe MyClass do
  describe '#foo' do
    subject { my_class_instance.foo }
  end

  describe '#bar' do
    subject { my_class_instance.bar }
  end
end
```

Let me know if anything is unclear, or could be improved. Thank you.

---

* [x] Added tests.
* [x] Updated documentation.


pirj:
We have a similar cop, https://github.com/rubocop/rubocop-rspec/blob/695d00405b9d470df23ec3503c641eb225832ca8/lib/rubocop/cop/rspec/multiple_describes.rb#L25

It has a good example of what I meant in:

> a cop that would suggest combining/splitting top-level groups with method specifiers

```ruby
      #   # bad
      #   describe MyClass, '.do_something' do
      #   end
      #   describe MyClass, '.do_something_else' do
      #   end
      #
      #   # good
      #   describe MyClass do
      #     describe '.do_something' do
      #     end
      #     describe '.do_something_else' do
      #     end
      #   end
```

Sorry for not making my expectations clear from the beginning.

0. To proceed, I suggest incorporating your changes to `MultipleDescribes` cop. Please keep in mind that the case it should consider is when there's a specifier (second argument to the top-level example group) for a described class.

1. What is the behaviour of the cop with "split" enforced style and a deeply-nested spec with `describe` used for all nesting? Will it suggest to flatten it all to level zero?

2. Also, what about a case where some common code is present for two nested example groups?

```ruby
# Is this still bad?

RSpec.describe MyClass do
  subject(:bar) { ... }
  let(:foo) { ... }
  before { ... }

  def helper
    ...
  end

  RSpec::Matchers.define :be_a_multiple_of do |expected|
    match do |actual|
      actual % expected == 0
    end
  end

  # Should this be extracted to a top-level example group?
  describe '#foo' do
    subject { my_class_instance.foo }
  end

  describe '#bar' do
    subject { my_class_instance.bar }
  end
end
```

3. I'd suggest giving it a think about renaming the cop. "Describes" is a bit more specific. A top-level example group can be a `feature` or some other alias. I think I've even seen `context` used exclusively.

M-Yamashita01:
Thank you for your comment.

> 0. To proceed, I suggest incorporating your changes to MultipleDescribes cop. Please keep in mind that the case it should consider is when there's a specifier (second argument to the top-level example group) for a described class.
> 1. What is the behaviour of the cop with "split" enforced style and a deeply-nested spec with describe used for all nesting? Will it suggest to flatten it all to level zero?

I understand. I will add my changes to MultipleDescribes cop.
And In the case of a split, I will flat it all to level zero.

> 2. Also, what about a case where some common code is present for two nested example groups?

There was a pattern of using a common code. I hadn't noticed it.
In the current cop, it detects "describe '#foo' do".
I think that It seems better to not detect it.

> 3. I'd suggest giving it a think about renaming the cop. "Describes" is a bit more specific. A top-level example group can be a feature or some other alias. I think I've even seen context used exclusively.

I understand. This time I will modify the existing cop without adding a new cop, so I will be careful about the cop name when I add a new cop in the future.

pirj:
> Could I close this Pull Request?

Please feel free to close if it's more convenient to you to send another one.

M-Yamashita01:
I understand. Thank you.

PR #1146, opened 2021-04-28
Bump version to 2.3.0

bquorning:
---

* [-] Added tests.
* [-] Updated documentation.

### DIFF
--- a/docs/antora.yml
+++ b/docs/antora.yml
@@ -1,5 +1,5 @@
 name: rubocop-rspec
 title: RuboCop RSpec
-version: master
+version: '2.3'
 nav:
   - modules/ROOT/nav.adoc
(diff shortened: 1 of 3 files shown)


docs/antora.yml
```diff
@@ -1,5 +1,5 @@
 name: rubocop-rspec
 title: RuboCop RSpec
-version: master
+version: '2.3'
```
Darhazer, line 3: I trust that you know whether it should include the '.0' at the end or not ;)

Darhazer (approved):

docs/antora.yml
bquorning: This is how we usually do it. I don't know more than that 😅

koic: Yeah, Only major and minor versions should be specified. Patch version is not required.
doc-version

PR #1156, opened 2021-05-21
Fix an exception in DescribedClass when accessing a constant on a variable in a namespaced spec

rrosenblum:


pirj (approved):
Great, thank you!

bquorning (approved):

Darhazer (approved):

PR #1158, opened 2021-06-09
Make FilePath check suffix when given a non-constant top-level node

topalovic:
Currently, `RSpec/FilePath` matcher ignores non-constant top-level nodes.

However, the standard feature format assumes string descriptions:

```ruby
RSpec.feature "Some feature" do
end
```

This can make feature files without a proper suffix go unnoticed and then ignored by the runner. Test it by creating a feature file without a suffix, e.g. `spec/features/some_feature.rb`.

This PR makes the matcher more general by matching `_` instead of `const`. If the top-level node is not a constant, we'll simply fall back to the basic `spec_suffix_only?` check, regardless of the actual config flag.

---

* [x] Added tests.
* [x] Updated documentation.

### DIFF
--- a/lib/rubocop/cop/rspec/file_path.rb
+++ b/lib/rubocop/cop/rspec/file_path.rb
@@ -61,10 +61,10 @@ module RuboCop
 
         MSG = 'Spec path should end with `%<suffix>s`.'
 
-        # @!method const_described(node)
-        def_node_matcher :const_described, <<~PATTERN
+        # @!method example_group(node)
+        def_node_matcher :example_group, <<~PATTERN
           (block
-            $(send #rspec? _example_group $(const ...) $...) ...
+            $(send #rspec? _example_group $_ $...) ...
           )
         PATTERN
 
@@ -74,17 +74,17 @@ module RuboCop
         def on_top_level_example_group(node)
           return unless top_level_groups.one?
 
-          const_described(node) do |send_node, described_class, arguments|
+          example_group(node) do |send_node, example_group, arguments|
             next if routing_spec?(arguments)
 
-            ensure_correct_file_path(send_node, described_class, arguments)
+            ensure_correct_file_path(send_node, example_group, arguments)
           end
         end
 
         private
 
-        def ensure_correct_file_path(send_node, described_class, arguments)
-          pattern = pattern_for(described_class, arguments.first)
+        def ensure_correct_file_path(send_node, example_group, arguments)
+          pattern = pattern_for(example_group, arguments.first)
           return if filename_ends_with?(pattern)
 
           # For the suffix shown in the offense message, modify the regular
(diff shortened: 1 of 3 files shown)


lib/rubocop/cop/rspec/file_path.rb
```diff
@@ -61,10 +61,10 @@ class FilePath < Base
 
         MSG = 'Spec path should end with `%<suffix>s`.'
 
-        # @!method const_described(node)
-        def_node_matcher :const_described, <<~PATTERN
+        # @!method example_group(node)
+        def_node_matcher :example_group, <<~PATTERN
           (block
-            $(send #rspec? _example_group $(const ...) $...) ...
+            $(send #rspec? _example_group $(_ ...) $...) ...
```
pirj, line 67: I think `$(_ ...)` can be just `$_`.

pirj (approved):
Hey @topalovic :wave: 

Looks good, thank you!

lib/rubocop/cop/rspec/file_path.rb
topalovic: Good call 👍🏼

lib/rubocop/cop/rspec/file_path.rb
```diff
@@ -61,10 +61,10 @@ class FilePath < Base
 
         MSG = 'Spec path should end with `%<suffix>s`.'
 
-        # @!method const_described(node)
-        def_node_matcher :const_described, <<~PATTERN
+        # @!method example_group(node)
+        def_node_matcher :example_group, <<~PATTERN
           (block
-            $(send #rspec? _example_group $(const ...) $...) ...
+            $(send #rspec? _example_group $_ $...) ...
```
Darhazer, line 67: Perhaps we can capture just the arguments and deconstruct when needed
```
$(send #rspec? _example_group $_ $...)
```
then the usage will be
```
example_group(node) do |send_node, arguments|
```
and inside `ensure_correct_file_path` we would have
```
example_group, method_name, = *arguments
```

pirj: Won't it result in an internal affairs offence?
```ruby
# lib/rubocop/cop/rspec/expect_output.rb
          # rubocop:disable InternalAffairs/NodeDestructuring
          variable_name, _rhs = *node
          # rubocop:enable InternalAffairs/NodeDestructuring
```

pirj:
I'm not sure why CircleCI doesn't pick up the build https://app.circleci.com/pipelines/github/rubocop/rubocop-rspec
Green locally with RuboCop 1.8.1 and Ruby 2.5.7.

lib/rubocop/cop/rspec/file_path.rb
Darhazer: Only when the variable name ends with node :joy:

topalovic: I'm not sure I'm up to snuff wrt "internal affairs" or `parser` intricacies, so I would trust you to refactor the matcher according to your vision :-)

I'm primarily concerned with getting the fix out there, since my own team got burned by trusting RuboCop to warn about feature files without the suffix.

Darhazer: Fair enough

Darhazer (approved):

lib/rubocop/cop/rspec/file_path.rb
bquorning: > I'm primarily concerned with getting the fix out there

It's released now, as v2.4.0.

PR #1163, opened 2021-06-30
Fix default config for sub-departments

### LINKED ISSUE #1160: `RSpec/Rails/AvoidSetupHook` consider `minitest` setup hook as problem
Hi there, I got rails project with a bunch of legacy code, and for some forgotten reason it uses both `rspec` and `minitest` tests

If I run `rubocop` with auto-fix in root folder - it will consider this as a problem and fix it and broke minitest file

```
rubocop test/controllers/clients_controller_test.rb
Inspecting 1 file
C

Offenses:

test/controllers/clients_controller_test.rb:6:3: C: [Correctable] RSpec/Rails/AvoidSetupHook: Use before instead of setup.
  setup do ...
  ^^^^^^^^

1 file inspected, 1 offense detected, 1 offense auto-correctable
```

And after applying fix I got:

 ```
rake test
rake aborted!
NoMethodError: undefined method `before' for ClientsControllerTest:Class
```

I think this cop should ignore files containg minitest tests, maybe by detecting `class ClientsControllerTest < ActionController::TestCase` or something like this

### PR DESCRIPTION
**Pull request by pirj**
It turns out that the default config does not apply for sub-departments.
It is weird on many levels:
 - RSpec DSL configuration didn't work
 - - RSpec DSL doesn't work as intended anyway #1126
 - if users would re-define RSpec DSL, they would have to do it for all sub-departments, as YAML defaults are only for the default config
 - if users would re-define RSpec DSL and use YAML defaults, too, the merging of the resulting config is undetermined

Still, I believe having `Include` to work correctly is the most important.

fixes #1160

---

* [-] Added tests.
* [-] Updated documentation.

### DIFF
--- a/config/default.yml
+++ b/config/default.yml
@@ -1,10 +1,10 @@
 ---
 RSpec:
   Enabled: true
-  Include:
+  Include: &1
     - "**/*_spec.rb"
     - "**/spec/**/*"
-  Language:
+  Language: &2
     inherit_mode:
       merge:
         - Expectations
@@ -745,6 +745,11 @@ RSpec/Yield:
   VersionAdded: '1.32'
   StyleGuide: https://www.rubydoc.info/gems/rubocop-rspec/RuboCop/Cop/RSpec/Yield
 
+RSpec/Capybara:
+  Enabled: true
+  Include: *1
+  Language: *2
+
 RSpec/Capybara/CurrentPathExpectation:
   Description: Checks that no expectations are set on Capybara's `current_path`.
   Enabled: true
@@ -766,6 +771,11 @@ RSpec/Capybara/VisibilityMatcher:
   VersionAdded: '1.39'
   VersionChanged: '2.0'
   StyleGuide: https://www.rubydoc.info/gems/rubocop-rspec/RuboCop/Cop/RSpec/Capybara/VisibilityMatcher
+
+RSpec/FactoryBot:
+  Enabled: true
+  Include: *1
+  Language: *2
 
 RSpec/FactoryBot/AttributeDefinedStatically:
   Description: Always declare attribute values as blocks.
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

**Comment by pirj** [comment]
@ShockwaveNN can you please check that this fix solves your case?
```ruby
# Gemfile
gem 'rubocop-rspec', github: 'rubocop/rubocop-rspec', branch: 'fix-config-for-nested-departments'
```

pirj:
@jonas054 Looks good?

ShockwaveNN:
@pirj Can confirm, this do not broke my minitests

jonas054 (approved):
Looks good as far as I can tell. :+1:

pirj:
Bump

Darhazer (commented):
I don't really understand the changes :man_shrugging: Please rebase anyway :)

pirj:
@Darhazer Rebased 👍 

The explanation of the problem is in https://github.com/rubocop/rubocop-rspec/issues/1160#issuecomment-871500855

> configuration for the RSpec department isn't automatically inherited down to subdepartments like RSpec/Rails. I'm not sure if we could change RuboCop so that kind of inheritance takes place. The other solution would be to add
```ruby
RSpec/Rails:
  Include:
    - "**/*_spec.rb"
    - "**/spec/**/*"
```

> in `config/default.yml` of rubocop-rspec.

config/default.yml
```diff
@@ -1,10 +1,10 @@
 ---
 RSpec:
   Enabled: true
-  Include:
+  Include: &1
```
Darhazer, line 4: It seems that this could be named, e.g. alphanumeric. Might make the file less scary, if it has proper names :)

lib/rubocop/rspec/config_formatter.rb
```diff
@@ -24,6 +25,8 @@ def dump
 
       def unified_config
         cops.each_with_object(config.dup) do |cop, unified|
+          next if SUBDEPARTMENTS.include?(cop)
```
Darhazer, line 28: so those are not really cops, but departments declarations. Nice

Darhazer (approved):

config/default.yml
pirj: I vaguely recall I planned to do so (to mimic the well-known `&default` from `config/database.yml` 😄), but ended up with numbers somehow. Fixed 👍

pirj: Ah, now I recall. It's `bin/build_config` that does that:
```diff
--- a/config/default.yml
+++ b/config/default.yml
@@ -1,10 +1,10 @@
 ---
 RSpec:
   Enabled: true
-  Include: &include
+  Include: &1
     - "**/*_spec.rb"
     - "**/spec/**/*"
-  Language: &language
+  Language: &2
     inherit_mode:
       merge:
         - Expectations
```

Supposedly, `YAML.dump(unified_config)` has no respect for named anchors.

I'll revert, otherwise CI is red.

PR #1166, opened 2021-07-16
Fix doc rspec hook argument

lokhi:
I think there is an issue with the example of EnforcedStyle: each for Rspec/HookArgument.
The before do should be bad imo.

---

* [ ] Added tests.
* [x] Updated documentation.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Changelog
 
 ## Master (Unreleased)
+* Fix each example for `RSpec/HookArgument`. ([@lokhi][])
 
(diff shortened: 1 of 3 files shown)


pirj (approved):
Well spotted :+1:
Confirmation https://github.com/rubocop/rubocop-rspec/blob/e33f8387d87ce73e1fdc8ad8fce659d4762d51d9/spec/rubocop/cop/rspec/hook_argument_spec.rb#L128

To fix CI, you should add yourself to the bottom of the changelog, too.

lokhi:
thanks it's ok

pirj:

PR #1170, opened 2021-07-26
Add SubjectDeclaration cop

### LINKED ISSUE #1157: Cop idea: Consider the `let` naming as `subject` an offence
Consider the `let` naming as `subject` an offence:

```ruby
let(:subject) { described_class.new }

describe '#result' do
  it 'returns forty two' do
    expect(subject.result).to eq 42
  end
end
```

dswij:
Closes #1157 

Adds a new cop `SubjectDeclaration` that considers declaring subject with `let` or `subject` an offense, i.e. `let(:subject)` or `subject(:subject)

---

* [x] Added tests.
* [x] Updated documentation.


pirj:
Nice!
I guess `rake` will pass if you change this:
```diff
 RSpec/SubjectDeclaration:
-  Description: Ensure that subject is defined using subject helper.
+  Description: Ensure that subject is defined using subject helper
   Enabled: pending
-  VersionAdded: '2.5.0'
+  VersionAdded: 2.5.0
```

dswij:
running `rake` seems to suggest that we need to end `Description` field with a period. But it autocorrects and remove the period. This seems a bit odd I think. 

```
Failures:

  1) config/default.yml ends every description with a period
     Failure/Error: expect(cop_configuration('Description')).to all(end_with('.'))

       expected ["Checks for stubbed test subjects.", "Checks that around blocks actually run the test.", "Do not use...k over Rails `setup` method.", "Enforces use of symbolic or numeric value to describe HTTP status."] to all end with "."

          object at index 25 failed to match:
             expected "Ensure that subject is defined using subject helper" to end with "."
     # ./spec/project/default_config_spec.rb:70:in `block (2 levels) in <top (required)>'
```

pirj (approved):
Please accept my apologies for the frustrating experience with periods in cop description.

The cop looks flawless.
Thank you so much for your contribution!

lib/rubocop/cop/rspec/subject_declaration.rb
```diff
@@ -0,0 +1,49 @@
...
+          (send nil? {#Helpers.all} {(sym :subject) (str "subject")} ...)
+        PATTERN
+
+        # @!method redundant_subject?(node)
+        def_node_matcher :redundant_subject?, <<~PATTERN
+          (send nil? {#Subjects.all} {(sym :subject) (str "subject")} ...)
+        PATTERN
+
+        def on_send(node)
+          let_subject?(node) do
```
Darhazer, line 38: Alternatively, you can combine the matchers, and just check the captured node method name to determine the message

spec/rubocop/cop/rspec/subject_declaration_spec.rb
```diff
@@ -0,0 +1,107 @@
...
+        subject!('subject') { 'some subject' }
+        ^^^^^^^^^^^^^^^^^^^ Ambiguous declaration of subject
+      RUBY
+    end
+  end
+
+  context 'when subject helper is used directly' do
+    it 'does not register an offense' do
+      expect_no_offenses(<<~RUBY)
+        subject(:test_subject) { 'some subject' }
```
Darhazer, line 103: for spec compliteness, it would be nice to have the rest of the helpers (subject!, let, let!) tested in a situation in which they do not trigger an offense. For subject there is the case with no name at all, e.g. `subject { 'some subject' }`

spec/rubocop/cop/rspec/subject_declaration_spec.rb
```diff
@@ -0,0 +1,107 @@
...
+
+    it 'is an offense when declared Ambiguously with `subject!`' do
+      expect_offense(<<~RUBY)
+        subject!(:subject) { 'some subject' }
+        ^^^^^^^^^^^^^^^^^^ Ambiguous declaration of subject
+      RUBY
+    end
+  end
+
+  context 'when `subject` is string' do
```
Darhazer, line 52: you can define a shared group and execute it with `:subject` and `'subject'`

Darhazer (commented):
It would be good to at least complete the specs for the valid usages of the subject.

lib/rubocop/cop/rspec/subject_declaration.rb
```diff
@@ -24,24 +24,22 @@ class SubjectDeclaration < Base
...
         PATTERN
 
         def on_send(node)
-          let_subject?(node) do
-            add_offense(node, message: MSG_LET)
-          end
+          offense = offensive_subject_declaration?(node)
+          return unless offense
+
+          add_offense(node, message: message_for(offense))
```
Darhazer, line 36:
```suggestion
          offensive_subject_declaration?(node) do |offense|
            add_offense(node, message: message_for(offense))
          end
```
Also the `?` in the name could be removed, as we are using it in a non-boolean context

lib/rubocop/cop/rspec/subject_declaration.rb
```diff
@@ -24,24 +24,22 @@ class SubjectDeclaration < Base
...
+        end
+
+        private
 
-          redundant_subject?(node) do
-            add_offense(node, message: MSG_REDUNDANT)
-          end
+        def message_for(offense)
+          Helpers.all(offense) ? MSG_LET : MSG_REDUNDANT
```
Darhazer, line 42: we already have the `let?` matcher in the language
```suggestion
          let?(offense) ? MSG_LET : MSG_REDUNDANT
```

Darhazer (approved):
Thank you for the improved specs. I think the implementation could be a bit cleaner, but it's good enough, so it's up to you if you'd like to apply any further suggestions on the code.
As far as I'm concerned, this can be merged (please squash the commits)

lib/rubocop/cop/rspec/subject_declaration.rb
dswij: `let?` matcher only applies to block pattern for `let` without block pass, and the Cop currently does not checks for it. 

I think that the additional complications in node pattern matcher is not worth to use `let?` matcher here.

pirj:
Thanks a lot for the contribution, @dswij ! Clean execution and a useful recruit for RSpec department 💯

PR #1171, opened 2021-07-27
Add rubocop-rake

bquorning:
RuboCop keeps suggesting it on every run, and I cannot think of a reason not to add this gem.

- https://rubygems.org/gems/rubocop-rake
- https://github.com/rubocop/rubocop-rake


pirj (approved):
LGTM!

PR #1178, opened 2021-08-10
Declare autocorrect as unsafe for `ExpectChange`

### LINKED ISSUE #1174: RSpec/ExpectChange false positive when calling a method that returns different results each time it is called
Before:

```ruby
it 'demonstrates the problem' do
  expect { false }.to change { changes_each_call.time }
end

def changes_each_call
  sleep(1)
  OpenStruct.new(time: Time.zone.now)
end
```

After:

```ruby
it 'demonstrates the problem' do
  expect { false }.to change(changes_each_call, :time)
end

# changes_each_call unmodified
```

This results in `changes_each_call` only being called once when `change` is called. The previous behavior would construct two objects, one each time the block is evaluated.

This bit me when `changes_each_call` was the result of an ActiveRecord query which needs to be fetched from the database to see the change occur.

francois-ferrandis:
This cop is unsafe, as illustrated in #1174.

fixes #1174

---

* [ ] Added tests.
* [x] Updated documentation.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Master (Unreleased)
 
+* Declare autocorrect as unsafe for `ExpectChange`. ([@francois-ferrandis][])
 * Fix each example for `RSpec/HookArgument`. ([@lokhi][])
 * Exclude unrelated Rails directories from `RSpec/DescribeClass`. ([@MothOnMars][])
 * Add `RSpec/ExcessiveDocstringSpacing` cop. ([@G-Rath][])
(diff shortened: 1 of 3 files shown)


francois-ferrandis:
@pirj Is that all good? :upside_down_face:

pirj (approved):
> added my name to the contributors list

We value any contribution. Thank you for being vigilant and persistent.

bquorning (approved):

francois-ferrandis:
> > added my name to the contributors list
> 
> We value any contribution. Thank you for being vigilant and persistent.

Thank you for being true to [your word](https://fili.pp.ru/advise-to-open-source-maintainers.html)! Very welcoming attitude, this is great! :pray:

PR #1182, opened 2021-08-13
Fix typo in NestedGroups doc

ybiquitous:
This change aims to fix just a typo in the `RSpec/NestedGroups` document.
See <https://docs.rubocop.org/rubocop-rspec/cops_rspec.html#rspecnestedgroups>

image

---

* [ ] Added tests.
* [x] Updated documentation.


pirj:
Thanks!

PR #1183, opened 2021-08-14
Docs minor fixes

pirj:
## Broken note

A typographic quote in
```
`Include`’s
```
seems to have broken the parsing of backticks. You may see that the text between `Includes` is wrapped into monospace code section, and the second backtick of the second backtick on the last `Includes` is missing.

image

We have had this issue with regular single quotes, that's why we've used a typographic one here. But the problems either affects both, or it started to with Antora version update.

## Broken sub-department cops docs

If you go to https://docs.rubocop.org/rubocop-rspec/2.0/upgrade_to_version_2.html, links to the second-level cops docs are broken. The last one is correct, https://docs.rubocop.org/rubocop-rspec/2.0/cops_rspec.html, while others have anchors, e.g. https://docs.rubocop.org/rubocop-rspec/2.0/upgrade_to_version_2.html#cops_capybara.adoc. This is probably due to `cops_docs.adoc` vs `cops_docs/` name clash.
Suspect: https://docs.antora.org/antora/2.3/page/page-id/#important
Ideally, we would generate `cops_rspec_capybara.adoc` instead of `cops_rspec/capybara.adoc`.

I've tracked it down to this method in RuboCop:
```
      def department_to_basename(department)
          "cops_#{department.downcase}"
```

Any suggestions? I've sent a [PR to RuboCop to update this to change `downcase` to `downcase.tr('/', '_')`](https://github.com/rubocop/rubocop/pull/10013).
Other RuboCop extensions don't seem to make use of sub-departments, and will be unaffected by this change.
Docs in the second commit are generated using the patched RuboCop.

---

* [-] Added tests.
* [x] Updated documentation.
* [-] The build (`bundle exec rake`) passes (be sure to run this locally, since it may produce updated documentation that you will need to commit).


pirj:
Corresponding RuboCop PR is merged. After its release:
 - the build will become green
 - in development, we won't have to point to `github:` in`Gemfile.local`
 - ...
 - and the doc will be fixed!

bquorning (approved):

bquorning:
This PR should not be merged until the next version of RuboCop has been released, right?

pirj:
> This PR should not be merged until the next version of RuboCop has been released, right?

We can merge it, but that would cause complication for contributors.
In any case, contributors with the checked-out `rubocop-rspec` would have to `bundle update --conservative rubocop` to use the fixed task. Or we would have to bump our dependency again. I don't have a preference here, maybe it would make sense to bump the dependency and make everyone's live simpler.

bquorning:
A new version of RuboCop was released today – I have rerun our CI, and everything is green.

If a contributor runs into tests failing because of an old version of RuboCop, I think they will quickly find the cause. If not, we’re here to help (if they ask).

I don’t think this is enough cause to bump the gem’s requirement of RuboCop.

PR #1186, opened 2021-08-27
Introduce LeadingLet cop

kykyi:
This PR introduces a new cop: `RSpec/LeadingLet`. `LeadingLet` has grown out of a desire to place all `let` and `let!` above other blocks in spec setup in order use these as the basic setup, and more specific blocks such as `subject` closer to the action (i.e. the spec they are the subject for!)

This PR leverages *heavily* off of the `LeadingSubject` class, thank you to the folks behind that one. The reason is because the logic is the same, put `let` above all else.

---

* [x] Added tests.
* [x] Updated documentation.


kykyi:
@pirj I like your spec structure suggestion! 

I guess having a spec for `LeadingSubject` and `LeadingLet` is an indication that what is desired is a `cop` where we will be able to specify spec structure, at least for `let`, `let!`, `subject`, `before`, and `after` - omitting `describe`/`context` because as you said they are used interchangeably. 

I might shelf this PR and put some time into a `Layout/SpecStructure`?

kykyi:
`allow edits by maintainers` is on 👍 feel free to push commits

pirj:
> I might shelf this PR and put some time into a Layout/SpecStructure?

That would be great!
I mean, this cop is still good, but if you're up for a bigger one, that's fantastic.

Actually, `Layout/ClassStructure` is smart enough to split definition types to categories, and set the desired expected order of those categories. This means that `context`/`describe` and other example group definitions can be put into the same category, and you may interleave them as you like.

I'd say https://github.com/rubocop/rubocop/blob/master/lib/rubocop/cop/layout/class_structure.rb can contain a lot of what you'll need. Please don't hesitate to extract this to a module in `rubocop`, and we'll bump the version so that the module is present. Or, just copy and paste, and we DRY it out later (however, there's a "100 lines max" limit in our `rubocop.yml` to my best memory).

pirj:
I still can't push to your branch for some reason. `rake generate_cops_documentation` is all that's needed to fix the build.

spec/rubocop/cop/rspec/leading_let_spec.rb
```diff
@@ -0,0 +1,281 @@
...
+  it 'registers an offense when let is below a non-offending node' do
+    expect_offense(<<~RUBY)
+      RSpec.describe do
+        def helper_method
+        end
+
+        it { is_expected.to be_present }
+
+        let(:params) { foo }
+        ^^^^^^^^^^^^^^^^^^^^ Declare all `let` above any other `it` declarations.
```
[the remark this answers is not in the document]
kykyi: The offense will bubble from the first block with comes before the `let`, I have added this spec to assert this:
```
 context 'when the let is below both a hook and a subject' do
    it 'registers an offense on the first offending node' do
      expect_offense(<<-RUBY)
      RSpec.describe User do
        subject { described_class.new}
        before { allow(Foo).to receive(:bar) }

        let(:params) { foo }
        ^^^^^^^^^^^^^^^^^^^^ Declare all `let` above any `subject` declarations.
      end
      RUBY

      expect_correction(<<-RUBY)
      RSpec.describe User do
        let(:params) { foo }
        subject { described_class.new}
        before { allow(Foo).to receive(:bar) }

      end
      RUBY
    end
  end
  ```
  Is this behaviour sensible?

**Comment by kykyi** [comment]
@pirj would still like to get this one in so my team can start using it, and in the meantime will work on the `SpecLayout` PR 🙏 

I have updated from all of your comments and run the `rake` task too

**Comment by pirj** [comment]
To make the CI green:
image

For some weird reason, I still can't push.

File: lib/rubocop/cop/rspec/leading_let.rb [hunk 2]
```diff
@@ -0,0 +1,91 @@
...
+
+module RuboCop
+  module Cop
+    module RSpec
+      # Enforce that let definitions come before all other definitions.
+      #
+      # @example
+      #   # bad
+      #     subject { described_class.new(params) }
+      #     let(:params) { blah }
```
pirj, line 11: Cosmetic: not need to additionally indent those examples:
```ruby
# @example
#   # bad
#   subject { described_class.new(params) }
```
is fine.

spec/rubocop/cop/rspec/leading_let_spec.rb
```diff
@@ -0,0 +1,292 @@
...
+
+      expect_correction(<<-RUBY)
+      RSpec.describe User do
+        let(:params) { foo }
+        subject { described_class.new}
+        before { allow(Foo).to receive(:bar) }
+
+      end
+      RUBY
+    end
```
pirj, line 290: Cosmetic: please indent the code inside heredoc just like in other examples.

pirj (approved):
Looks good to me. Thanks!

pirj:
> get this one in so my team can start using it, and in the meantime will work on the SpecLayout PR

Makes perfect sense 👍

Darhazer:
This conflicts with LeadingSubject. At the very leat, those two should have `autocorrect_incompatible_with` declared, and one should be disabled by default. 

It would be best to have a cop to force the structure, as @pirj suggested (and I'm quite sure we used to have such an issue open, though I can't find it), or perhaps we can have one cop (LeadingDeclaration?) that would force what comes first (let/subject)

pirj:
Now, when I think about this cop, which basically suggests putting `let`s before `subject`.

```ruby
# good
let(:params) { blah }
subject { described_class.new(params) }
```

The cop goes against the [Leading Subject](https://rspec.rubystyle.guide/#leading-subject) guideline:

> When subject is used, it should be the first declaration in the example group.

```ruby
# good
describe Article do
  subject { FactoryBot.create(:some_article) }
  let(:user) { FactoryBot.create(:user) }
```

The [project goals](https://docs.rubocop.org/rubocop-rspec/#project-goals) state that:

> Enforce the guidelines and best practices outlined in the community [RSpec style guide](https://rspec.rubystyle.guide/)

We have [`RSpec/LetBeforeExamples`](https://docs.rubocop.org/rubocop-rspec/cops_rspec.html#rspecletbeforeexamples) cop that would put `let`s to the top of the example group, but below `subject`, and below hooks.

With [`RSpec/LeadingSubject`](https://docs.rubocop.org/rubocop-rspec/cops_rspec.html#rspecleadingsubject) they would allow ordering that meets the style guide while still leaving some space for flexibility of ordering `let`s and hooks.

To be fair, this cop would allow putting `let`s before hooks, and I'm all for this style, but this is not mentioned in the style guide.

Still, I'm more inclined to close this. My sincere apologies for wasting your time and effort, @kykyi.

PR #1198, opened 2021-10-05
fix(ExcessiveDocstringSpacing): ignore heredocs

### LINKED ISSUE #1196: RSpec/ExcessiveDocstringSpacing incorrect behaviour with heredoc syntax
I think this rule can be applied only to one-line descriptions.
It should not correct heredocs, it breaks ruby syntax after auto-correct..

```ruby
# delete/by_topic_spec.rb: C: [Correctable] RSpec/ExcessiveDocstringSpacing: Excessive whitespace.
#  it <<~DESC do
#      ^^^^^

  it <<~DESC do
    does not remove
      another comments reports,
      another comments votes,
      another comments image,
      another comments feeds,
      another comments notifications,
      another comments public feeds,
      another comments subs
    does not nullifies
      another comments notifiers resources
  DESC
```

G-Rath:
Fixes #1196

I'll do a separate PR for fixing `ExampleWording` because that can support reporting against herdocs, just not fixing them automatically.

---

* [x] Added tests.
* [x] Updated documentation.


bquorning (approved):
I wonder if people are using heredocs for their docstrings? Maybe we should have an ExcessiveDocstring cop 😆

Darhazer (approved):
Please rebase (fixing the changelog)

PR #1199, opened 2021-10-05
fix(ExampleWording): do not try and fix heredoc strings

G-Rath:
This was found via #1196; while we can accurately determine if heredoc descriptions violate this rule, the current fixer turns heredocs into invalid syntax, so we shouldn't try and fix them.

---

* [x] Added tests.
* [x] Updated documentation.


pirj (approved):
Fantastic streak!

lib/rubocop/cop/rspec/example_wording.rb
```diff
@@ -60,8 +60,13 @@ def on_block(node)
...
         def add_wording_offense(node, message)
           docstring = docstring(node)
-          add_offense(docstring, message: message) do |corrector|
-            corrector.replace(docstring, replacement_text(node))
+
+          if node.heredoc?
+            add_offense(docstring, message: message)
+          else
+            add_offense(docstring, message: message) do |corrector|
+              corrector.replace(docstring, replacement_text(node))
```
bquorning, line 68: I wonder if this could work as
```suggestion
              corrector.replace(docstring, replacement_text(node)) unless node.heredoc?
```
Also, I am not sure if I even like my own suggestion 😄

bquorning: I just tried @Darhazer’s suggestion and it seems to work – the specs are passing. @G-Rath could you try it again please?

Darhazer: Yup that's pattern is used in some of the rubocop's core cops as well

pirj: I'm wondering if this would affect what is shown in the footer. I would guess if `add_offense` is called with a block, that means it would add to "correctable". While for heredoc it's not the case, and it should only appear as an offence, not as a autocorrectable one.
Please correct me if this assumption is wrong.

G-Rath: ahh *`next`* - I was doing `return` 🤦 

Awesome, was hoping we could do something like this instead of having the if/else.

bquorning (approved):

PR #1205, opened 2021-10-27
Fix a build error for Ruby 3.1.0-dev

koic:
Follow up to https://github.com/ruby/psych/pull/487.

This PR fixes the following Ruby 3.1.0-dev build error.

```console
% ruby -v
ruby 3.1.0dev (2021-10-26T00:30:42Z master 7d4c59203f) [x86_64-darwin19]

% bundle exec rake
bin/build_config
/Users/koic/.rbenv/versions/3.1.0-dev/lib/ruby/3.1.0/psych/visitors/to_ruby.rb:430:in
`visit_Psych_Nodes_Alias': Unknown alias: 1 (Psych::BadAlias)
        from /Users/koic/.rbenv/versions/3.1.0-dev/lib/ruby/3.1.0/psych/visitors/visitor.rb:30:in `visit'
        from /Users/koic/.rbenv/versions/3.1.0-dev/lib/ruby/3.1.0/psych/visitors/visitor.rb:6:in `accept'
        from /Users/koic/.rbenv/versions/3.1.0-dev/lib/ruby/3.1.0/psych/visitors/to_ruby.rb:35:in `accept'
(snip)

        from /Users/koic/.rbenv/versions/3.1.0-dev/lib/ruby/3.1.0/psych/visitors/to_ruby.rb:35:in `accept'
        from /Users/koic/.rbenv/versions/3.1.0-dev/lib/ruby/3.1.0/psych.rb:335:in `safe_load'
        from /Users/koic/.rbenv/versions/3.1.0-dev/lib/ruby/3.1.0/psych.rb:370:in `load'
        from /Users/koic/.rbenv/versions/3.1.0-dev/lib/ruby/3.1.0/psych.rb:671:in `block in load_file'
        from /Users/koic/.rbenv/versions/3.1.0-dev/lib/ruby/3.1.0/psych.rb:670:in `open'
        from /Users/koic/.rbenv/versions/3.1.0-dev/lib/ruby/3.1.0/psych.rb:670:in `load_file'
        from bin/build_config:23:in `<main>'
rake aborted!
Command failed with status (1): [bin/build_config...]
/Users/koic/src/github.com/rubocop/rubocop-rspec/Rakefile:37:in `block
in <top (required)>'
/Users/koic/.rbenv/versions/3.1.0-dev/bin/bundle:23:in `load'
/Users/koic/.rbenv/versions/3.1.0-dev/bin/bundle:23:in `<main>'
Tasks: TOP => default => build_config
(See full trace by running task with --trace)
[...]

PR #1210, opened 2021-11-08
Trigger GitHub Actions workflow on pull requests

bquorning:
The `push` event alone doesn't work when making changed on a forked repository – see how the workflow currently doesn’t run for https://github.com/rubocop/rubocop-rspec/pull/1205. So now the main workflow runs when a pull request is opened, re-opened or "synchronized" which I assume is when you push new changes.

There is some documentation at https://docs.github.com/en/actions/learn-github-actions/events-that-trigger-workflows#pull_request, but I cannot find anything that says exactly that `push` doesn't trigger on a forked repository. But I tried it out in #1209 where it seems to fix the problem.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,6 +1,10 @@
 name: CI
 
-on: push
+on:
+  pull_request:
+  push:
+    branches:
+      - master
 
 jobs:
   confirm_config_and_documentation:


pirj (approved):
RSpec has it [this way](https://github.com/rspec/rspec-core/blob/e36aa2a9ebe68acee3ce05190fc2124947b45925/.github/workflows/ci.yml#L13). Not sure if `branches: '*'` even does anything. Let's see how this goes, LGTM!

PR #1212, opened 2021-11-12
Add AutoCorrect config support to Focus

ni3t:
Background:
- The `RSpec/Focus` cop is very useful in preventing developers from accidentally committing focused specs, which skips all other specs in a suite.
- In certain editors, rubocop can be configured to run on save, with autocorrect enabled, which automatically applies corrections with cops considered "autocorrectable"
- In the case of the `RSpec/Focus` cop, if you are developing a test and want to focus it, and you have autocorrect configured in your editor, when you save the file it will remove any focused specs.

Therefore:
- This PR introduces AutoCorrect configurability to the `RSpec/Focus` cop
- Developers can choose whether they want the `RSpec/Focus` cop to automatically correct violations, or only notify.

---

* [x] Added tests.
* [ ] Updated documentation.

* [na] Added the new cop to `config/default.yml`.
* [na] The cop is configured as `Enabled: pending` in `config/default.yml`.
* [na] The cop documents examples of good and bad code.
* [na] The tests assert both that bad code is reported and that good code is not reported.
* [na] Set `VersionAdded` in `default/config.yml` to the next minor version.

### DIFF
--- a/lib/rubocop/cop/rspec/focus.rb
+++ b/lib/rubocop/cop/rspec/focus.rb
@@ -87,6 +87,10 @@ module RuboCop
           corrector.replace(range,
                             range.source.sub(focus.method_name.to_s, unfocused))
         end
+
+        def should_autocorrect
+          cop_config['AutoCorrect']
+        end
       end
     end
   end
--- a/config/default.yml
+++ b/config/default.yml
@@ -371,8 +371,9 @@ RSpec/FilePath:
 RSpec/Focus:
   Description: Checks if examples are focused.
   Enabled: true
+  AutoCorrect: true
   VersionAdded: '1.5'
-  VersionChanged: '2.1'
+  VersionChanged: '2.2'
   StyleGuide: https://www.rubydoc.info/gems/rubocop-rspec/RuboCop/Cop/RSpec/Focus
 
 RSpec/HookArgument:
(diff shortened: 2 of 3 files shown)


ni3t:
I'm not sure what is necessary to update the docs or if they are auto-updated. Thanks!

lib/rubocop/cop/rspec/focus.rb
```diff
@@ -87,6 +87,10 @@ def correct_send(corrector, focus)
           corrector.replace(range,
                             range.source.sub(focus.method_name.to_s, unfocused))
         end
+
+        def should_autocorrect
```
pirj, line 91: I can't find where this is called from. Either in `rubocop-rspec` or in `rubocop`.
Will specs fail if you remove this method?

pirj:
What happens if you put this in your project's `.rubocop.yml`?
```yaml
RSpec/Focus:
  AutoCorrect: false
```

I remember in some cases RuboCop would emit a warning that `AutoCorrect` is among the cop's configuration options. The option would still work though. I've tried it locally and there is no such warning. That probably was a bug in RuboCop, and it's fixed now.

To me, it seems like a regular RuboCop feature that all cops normally adhere to. And not something that should be covered with individual cops' specs.

`AutoCorrect: true` is the default if the cop supports auto-correction. Just confirmed that practically.

So:
1. The change to `config/default.yml` is redundant - there is no warning without this change if you configure `AutoCorrect` to `false`; `true` is the default anyway
2. `should_autocorrect` in `lib/rubocop/cop/rspec/focus.rb` doesn't seems to do anything
3. Cops do not normally have specs for `Enabled`/`*AutoCorrect`/`Safe` options. `spec/rubocop/cop/rspec/focus_spec.rb` changes are redundant

What does this PR do, can you please elaborate?

pirj:
Ping

PR #1213, opened 2021-11-12
Set `rubygems_mfa_required` in gemspec

dvandersluis:
Ruby gems now allows MFA to be required for pushes: https://guides.rubygems.org/mfa-requirement-opt-in/
This allows users to have confidence that the actual authors were responsible for updates (obviously that doesn't mean that it can't be a malicious update, but at least it's not from someone who got access to a pusher's account).

Follows rubocop/rubocop#10239. Update gemspec to require MFA for privileged operations.


pirj:
@bquorning Good to go?

bquorning (approved):
That’s great!

PR #1214, opened 2021-11-21
Add new `RSpec/FileFixture` cop

leoarnold:
`RSpec` provides the convenience method

```ruby
file_fixture('path/some_file.csv')
```

which (by default) is equivalent to

```ruby
Rails.root.join('spec/fixtures/files/path/some_file.csv')
```

We add a cop which detects and autocorrects verbose `Rails.root`
constructions and works best in combination with [`Rails/RootPathnameMethods`](https://github.com/rubocop/rubocop-rails/pull/587).

---

* [X] Added tests.
* [X] Updated documentation.

* [X] Added the new cop to `config/default.yml`.
* [X] The cop is configured as `Enabled: pending` in `config/default.yml`.
* [X] The cop documents examples of good and bad code.
* [X] The tests assert both that bad code is reported and that good code is not reported.
* [X] Set `VersionAdded` in `default/config.yml` to the next minor version.


changelog/new_add_new_rspecfilefixture_cop.md
```diff
@@ -0,0 +1 @@
+* [#1214](https://github.com/rubocop/rubocop-rails/pull/1214): Add new `RSpec/FileFixture` cop. ([@leoarnold][])
```
pirj, line 1: We're still using the old approach with just one CHANGELOG.md in this repo - it's not a major pain to merge it.

lib/rubocop/cop/rspec/file_fixture.rb
```diff
@@ -0,0 +1,107 @@
...
+      #   "#{Rails.root}/spec/fixtures/path/some_file.csv"
+      #   Rails.root.join('spec/fixtures/path/some_file.csv')
+      #   Rails.root.join('spec', 'fixtures', 'path', 'some_file.csv')
+      #
+      #   "#{Rails.root}/spec/fixtures/files/some_file.csv"
+      #   Rails.root.join('spec/fixtures/files/some_file.csv')
+      #   Rails.root.join('spec', 'fixtures', 'files', 'some_file.csv')
+      #
+      #   # good
+      #   file_fixture('../path/some_file.csv').path
```
pirj, line 33: Is `path` a `Pathname` method? I couldn't find it.

lib/rubocop/cop/rspec/file_fixture.rb
```diff
@@ -0,0 +1,107 @@
...
+        def_node_matcher :file_io?, <<~PATTERN
+          (send
+            (const {nil? cbase} :File) {:binread :binwrite :open :read :write} $...)
+        PATTERN
+
+        # @!method rails_root_join(node)
+        def_node_matcher :rails_root_join, <<~PATTERN
+          (send
+            (send
+              (const {nil? cbase} :Rails) :root) :join (str $_)+)
```
pirj, line 61: What if it's not a literal string passed to `join`? E.g. a string variable or an interpolated string.

spec/rubocop/cop/rspec/file_fixture_spec.rb
```diff
@@ -0,0 +1,145 @@
+# frozen_string_literal: true
+
+RSpec.describe RuboCop::Cop::RSpec::FileFixture, :config do
+  it 'registers an offense when using ' \
+     '`"\#{Rails.root}/spec/fixtures/files/some_file.csv"`' do
+    expect_offense(<<~RUBY)
+      "\#{Rails.root}/spec/fixtures/files/some_file.csv"
```
pirj, line 7: I think it's possible to write heredoc as `<<~'RUBY'` to avoid escaping the hash.

spec/rubocop/cop/rspec/file_fixture_spec.rb
```diff
@@ -0,0 +1,145 @@
+# frozen_string_literal: true
+
+RSpec.describe RuboCop::Cop::RSpec::FileFixture, :config do
+  it 'registers an offense when using ' \
+     '`"\#{Rails.root}/spec/fixtures/files/some_file.csv"`' do
```
pirj, line 5: This docstring just duplicates what the example itself tells. Does it make sense to write a free-form text explaining this case?

config/default.yml
```diff
@@ -353,6 +353,12 @@ RSpec/ExpectOutput:
   VersionAdded: '1.10'
   StyleGuide: https://www.rubydoc.info/gems/rubocop-rspec/RuboCop/Cop/RSpec/ExpectOutput
 
+RSpec/FileFixture:
```
pirj, line 356: We have Rails sub-department.

pirj (commented):
Looks good in general.
Before looking more closely: do you think this cop will suit `rubocop-rails` more?
Even though it's related to testing, it would also register offences in Minitest tests, wouldn't it?
And it's strictly for Rails.

leoarnold:
@pirj You're right. I thought RSpec had it's own `file_fixture` helper, but it's in `rspec-rails` and it literally calls `activesupport`:

https://github.com/rspec/rspec-rails/blob/b4b8271a61077af94dc2e24ef09b6312515399db/lib/rspec/rails/fixture_file_upload_support.rb#L42

I'll close here and reopen on `rubocop-rails`.

PR #1215, opened 2021-11-23
Add new `RSpec/FactoryBot/SyntaxMethods` cop.

leoarnold:
`FactoryBot` provides a mixin called
[`FactoryBot::Syntax::Methods'](https://github.com/thoughtbot/factory_bot/blob/master/GETTING_STARTED.md#rspec)
which allows to replace

```ruby
FactoryBot.create(:bar)
FactoryBot.build(:bar)
FactoryBot.attributes_for(:bar)
```

with

```
create(:bar)
build(:bar)
attributes_for(:bar)
```

The auto-correction is unsafe as there is no way to tell just from syntax
whether a given example group includes the mixin or not.

---

* [x] Added tests.
* [x] Updated documentation.

### DIFF
--- a/config/default.yml
+++ b/config/default.yml
@@ -816,6 +816,13 @@ RSpec/FactoryBot/FactoryClassName:
   VersionChanged: '2.0'
   StyleGuide: https://www.rubydoc.info/gems/rubocop-rspec/RuboCop/Cop/RSpec/FactoryBot/FactoryClassName
 
+RSpec/FactoryBot/SyntaxMethods:
+  Description: Use shorthands from `FactoryBot::Syntax::Methods` in your specs.
+  Enabled: pending
+  SafeAutoCorrect: false
+  VersionAdded: "<<next>>"
+  StyleGuide: https://www.rubydoc.info/gems/rubocop-rspec/RuboCop/Cop/RSpec/FactoryBot/SyntaxMethods
+
 RSpec/Rails:
   Enabled: true
   Include: *1
--- a/docs/modules/ROOT/pages/cops_rspec_factorybot.adoc
+++ b/docs/modules/ROOT/pages/cops_rspec_factorybot.adoc
@@ -157,3 +157,66 @@ end
 === References
 
 * https://www.rubydoc.info/gems/rubocop-rspec/RuboCop/Cop/RSpec/FactoryBot/FactoryClassName
+
+== RSpec/FactoryBot/SyntaxMethods
+
+|===
+| Enabled by default | Safe | Supports autocorrection | Version Added | Version Changed
+
+| Pending
+| Yes
+| Yes (Unsafe)
+| <<next>>
+| -
+|===
+
+Use shorthands from `FactoryBot::Syntax::Methods` in your specs.
+
+=== Safety
+
+The auto-correction is marked as unsafe because the cop
+cannot verify whether you already include
+`FactoryBot::Syntax::Methods` in your test suite.
+
+If you're using Rails, add the following configuration to
+`spec/support/factory_bot.rb` and be sure to require that file in
+`rails_helper.rb`:
+
+[source,ruby]
+----
+RSpec.configure do |config|
+  config.include FactoryBot::Syntax::Methods
+end
+----
+
+If you're not using Rails:
+
+[source,ruby]
+----
+RSpec.configure do |config|
+  config.include FactoryBot::Syntax::Methods
+
+  config.before(:suite) do
+    FactoryBot.find_definitions
+  end
+end
+----
+
+=== Examples
+
+[source,ruby]
+----
+# bad
+FactoryBot.create(:bar)
+FactoryBot.build(:bar)
+FactoryBot.attributes_for(:bar)
+
+# good
+create(:bar)
+build(:bar)
+attributes_for(:bar)
+----
+
+=== References
+
+* https://www.rubydoc.info/gems/rubocop-rspec/RuboCop/Cop/RSpec/FactoryBot/SyntaxMethods
--- a/lib/rubocop/cop/rspec/capybara/feature_methods.rb
+++ b/lib/rubocop/cop/rspec/capybara/feature_methods.rb
@@ -42,6 +42,7 @@ module RuboCop
         #   end
         class FeatureMethods < Base
           extend AutoCorrector
+          include InsideExampleGroup
 
           MSG = 'Use `%<replacement>s` instead of `%<method>s`.'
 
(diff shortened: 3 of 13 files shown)


lib/rubocop/cop/rspec/mixin/inside_example_group.rb
```diff
@@ -0,0 +1,27 @@
+# frozen_string_literal: true
+
+module RuboCop
+  module Cop
+    module RSpec
+      # Helps you identify whether a given node
+      # is within an example group or not.
+      module InsideExampleGroup
+        def inside_example_group?(node)
```
pirj, line 9: Will adding `private` above all methods break things?

lib/rubocop/cop/rspec/factory_bot/syntax_methods.rb
```diff
@@ -0,0 +1,111 @@
...
+
+module RuboCop
+  module Cop
+    module RSpec
+      module FactoryBot
+        # Use the shorthands from `FactoryBot::Syntax::Methods` in your specs.
+        #
+        # The auto-correction is marked as unsafe because the cop
+        # cannot verify whether you already include
+        # `FactoryBot::Syntax::Methods` in your test suite.
```
pirj, line 11: Fair enough 👍

lib/rubocop/cop/rspec/factory_bot/syntax_methods.rb
```diff
@@ -0,0 +1,111 @@
+# frozen_string_literal: true
+
+module RuboCop
+  module Cop
+    module RSpec
+      module FactoryBot
+        # Use the shorthands from `FactoryBot::Syntax::Methods` in your specs.
+        #
+        # The auto-correction is marked as unsafe because the cop
```
pirj, line 9: There's a recently added `@safety` tag. I believe it adds a comment to the docs in a specific way. Example https://github.com/rubocop/rubocop/blob/783cb69643ab9a40fee9e8917aa557299be5b59a/lib/rubocop/cop/lint/number_conversion.rb#L21

lib/rubocop/cop/rspec/capybara/feature_methods.rb
```diff
@@ -60,13 +61,6 @@ class FeatureMethods < Base
             {#{MAP.keys.map(&:inspect).join(' ')}}
           PATTERN
 
-          # @!method spec?(node)
-          def_node_matcher :spec?, <<-PATTERN
-            (block
-              (send #rspec? {:describe :feature} ...)
-            ...)
-          PATTERN
```
pirj, line 68: Thanks! This was missing other example group aliases.

lib/rubocop/cop/rspec/factory_bot/syntax_methods.rb
```diff
@@ -0,0 +1,111 @@
...
+
+              add_offense(crime_scene(node), message: message) do |corrector|
+                corrector.remove(offense(node))
+              end
+            end
+          end
+
+          private
+
+          def crime_scene(node)
```
pirj, line 94: I think I've first seen `evidence` used for the first time in RuboCop ecosystem, and now `crime_scene`. Those are perfect names to describe what those methods do.

lib/rubocop/cop/rspec/factory_bot/syntax_methods.rb
```diff
@@ -0,0 +1,111 @@
...
+
+          # @!method factory_bot_method?(node)
+          def_node_matcher :factory_bot_method?, <<~PATTERN
+            (send (const {nil? cbase} :FactoryBot) $%RESTRICT_ON_SEND ...)
+          PATTERN
+
+          def on_send(node)
+            return unless inside_example_group?(node)
+
+            factory_bot_method?(node) do |method|
```
pirj, line 83: The flow here is quite complex.

We find method calls for methods from a list.
For each, we check if it's called inside an example group. To me, that code that you've extracted to a concern is far from being obvious in what it does.
We check the receiver to be `FactoryBot`.

We already know that `node.method_name` is in the list. Will it be sufficient to check if `node.receiver` matches `(const {nil? cbase} :FactoryBot)`?

Does it make sense to make this check before walking up the tree in a quest for example groups?

config/default.yml
```diff
@@ -816,6 +816,13 @@ RSpec/FactoryBot/FactoryClassName:
   VersionChanged: '2.0'
   StyleGuide: https://www.rubydoc.info/gems/rubocop-rspec/RuboCop/Cop/RSpec/FactoryBot/FactoryClassName
 
+RSpec/FactoryBot/SyntaxMethods:
+  Description: Use the shorthands from `FactoryBot::Syntax::Methods` in your specs.
+  Enabled: pending
+  SafeAutoCorrect: false
+  VersionAdded: "<<next>>"
```
pirj, line 823: We didn't use that notion in the past, but it's quite a useful trick.

pirj (approved):
Looks good as is. A few cosmetic notes, and a couple of optional suggestions.
Thanks!

lib/rubocop/cop/rspec/factory_bot/syntax_methods.rb
leoarnold: Had to remove it because `rake default` said `@safety` is not known.

lib/rubocop/cop/rspec/mixin/inside_example_group.rb
leoarnold: Added.

lib/rubocop/cop/rspec/factory_bot/syntax_methods.rb
pirj: Wondering how it works 🤔 
What if you `bundle update rubocop`?

leoarnold: I added the `.yardopts` file from `rubocop-performance` and now it works.

docs/modules/ROOT/pages/cops_rspec_factorybot.adoc
```diff
@@ -157,3 +157,66 @@ end
...
+| Pending
+| Yes
+| Yes (Unsafe)
+| <<next>>
+| -
+|===
+
+Use shorthands from `FactoryBot::Syntax::Methods` in your specs.
+
+=== Safety
```
pirj, line 175: :kissing: :ok_hand: perfect, thank you!

Darhazer (approved):

pirj:
Thank you for the contribution!

lib/rubocop/cop/rspec/factory_bot/syntax_methods.rb
```diff
@@ -0,0 +1,107 @@
...
+            build_pair
+            build_stubbed
+            build_stubbed_list
+            build_stubbed_pair
+            create
+            create_list
+            create_pair
+            generate
+            generate_list
+            null
```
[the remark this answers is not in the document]
leoarnold: `FactoryBot::Syntax::Methods.instance_methods.sort` :wink:

PR #1211, opened 2021-11-10
Add Ruby 3.1 to the matrix

pirj:
Follow-up to #1205

---

Before submitting the PR make sure the following are checked:

* [x] Feature branch is up-to-date with `master` (if not - rebase it).
* [x] Squashed related commits together.
* [-] Added tests.
* [-] Updated documentation.
* [-] Added an entry to the `CHANGELOG.md` if the new code introduces user-observable changes.
* [x] The build (`bundle exec rake`) passes (be sure to run this locally, since it may produce updated documentation that you will need to commit).

PR #1223, opened 2021-12-28
Add Ruby 3.1 release to the matrix

pirj:
3.1.0 was released https://www.ruby-lang.org/en/news/2021/12/25/ruby-3-1-0-released/

Follow-up to #1211, but to avoid occidental failures with `ruby-head` that supposedly tracks the ongoing development of a potentially breaking Ruby 3.2.

---

* [-] Added tests.
* [-] Updated documentation.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -28,6 +28,7 @@ jobs:
           - 2.7
           - "3.0"
           - "3.1"
+          - ruby-head
           - jruby
         task:
           - internal_investigation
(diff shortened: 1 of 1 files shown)


.github/workflows/main.yml
```diff
@@ -27,7 +27,7 @@ jobs:
           - 2.6
           - 2.7
           - "3.0"
-          - ruby-head
+          - "3.1"
```
bquorning, line 30: I suggest that we keep ruby-head too, so we can spot (and perhaps report on) build errors.
```suggestion
          - "3.1"
          - ruby-head
```

bquorning (approved):

PR #1232, opened 2022-01-16
Make RSpec/SubjectStub stricter

### LINKED ISSUE #1230: Isn't the SubjectStub cop particularly ambiguous about the rules for nested subjects?
cf: #1228 

The `SubjectStub` cop prohibits stubs even on nested subjects as an inclusion of its rules.
However, the way the warning message is issued and the strictness of the rule are unclear, and it seems to be a source of confusion for users.

For example, the following should be offensive, but are not detected as offensive.

```
describe '#example' do
  subject(:foo) { described_class.new }
  context 'when I shake things up' do
    subject(:bar) { described_class.new }

    before do
      allow(foo).to receive(:wow) # => Is this offensive?
      # allow(bar).to receive(:wow)  #=> If you remove comment out, it becomes offensive.
    end

    it 'tries to trick rubocop-rspec' do
      expect(bar).to eql(foo)
    end
end
```

It is unlikely that such code would be written intentionally, but it could be unintentionally mixed in, and it should be detected by rubocop's checks.

In addition, there are cases where stubbing a subject is not detected as offensive, such as when nesting by an anonymous subject is not detected as offensive.
It is desirable that these intentions or regularities are clearly stated in the description if possible.

pirj:
fixes #1230

Make the cop stricter by also considering stubbing of subjects defined in parent example groups:
```ruby
RSpec.describe Foo do
  subject(:foo) { ... }
  
  describe '#bar' do
    subject(:bar) { foo.bar }

    specify do
      allow(foo).to receive(:baz).and_return(:baz) # <= this wasn't previously considered as an offence
    end
  end
end
```

### Comparison

Running it against [`real-world-rspec`](https://github.com/pirj/real-world-rspec):

Before:
```
55577 files inspected, 1686 offenses detected
rubocop --only RSpec/SubjectStub ~/source/real-world-rspec  1537.45s user 139.80s system 167% cpu 16:42.65 total
```

After:
```
55577 files inspected, 1752 offenses detected
rubocop --only RSpec/SubjectStub ../real-world-rspec  1521.27s user 129.19s system 191% cpu 14:20.72 total
```

~🎉 due to minor optimizations, the cop is now 2x faster.~ - no changes

Less than 100 new offences, details below.

### Newly Discovered Real-World Offences

Basically all in the same category - `subject` is redefined in a child example group, and the subject from parent example group is stubbed (cases 1-5).

Only the last case 6 takes care of redefining the previously defined `subject` with `let`. ✅ - we ignore this.

1.
```ruby
RSpec.describe API::Helpers::Caching, :use_clean_rails_redis_caching do
  subject(:instance) { Class.new.include(described_class, Grape::DSL::Headers).new }

  describe "#cache_action_if" do
    subject do
      instance.cache_action_if(conditional, cache_key, **kwargs) do
        return_value
      end
    end

    it "caches the block" do
      expect(instance).to receive(:cache_action).with(cache_key, **kwargs)

      subject
    end
```

2.
```ruby
RSpec.describe Banzai::ReferenceParser::MergeRequestParser do
  subject(:parser) { described_class.new(Banzai::RenderContext.new(merge_request.target_project, user)) }


  describe '#can_read_reference?' do
    subject { parser.can_read_reference?(user, merge_request) }

    it 'calls #can? only once' do
      expect(parser).to receive(:can?).once

      2.times { parser.can_read_reference?(user, merge_request) }
    end
```

3.
```ruby
RSpec.describe Gitlab::AlertManagement::Payload::Base do
  subject(:parsed_payload) { payload_class.new(project: project, payload: raw_payload) }

  describe '#alert_params' do
    subject { parsed_payload.alert_params }
      before do
        allow(parsed_payload).to receive_messages(stubs)
      end
```

4.
```ruby
RSpec.describe ::Gitlab::Ci::Pipeline::Logger do
  subject(:logger) { described_class.new(project: project) }

  describe '#commit' do
    subject(:commit) { logger.commit(pipeline: pipeline, caller: 'source') }

    before do
      allow(logger).to receive(:current_monotonic_time) { Time.current.to_i }
    end
```

5.
```ruby
RSpec.describe RuboCop::Formatter::JSONFormatter do
  subject(:formatter) { described_class.new(output) }

  describe '#hash_for_file' do
    subject(:hash) { formatter.hash_for_file(file, offenses) }

    before do
      count = 0
      allow(formatter).to receive(:hash_for_offense) do
        count += 1
      end
    end
```

6. ⚠️ ~`return_item` is redefined from `subject` to `let` - seems to be a false positive~ ✅ fixed, we ignore this now
```ruby
RSpec.describe Spree::ReturnItem, type: :model do
  describe "reception_status state_machine" do
    subject(:return_item) { create(:return_item) }

    context 'when transitioning to :received' do
      let(:return_item) { create(:return_item) }
      subject { return_item.receive! }

      # StateMachines has some "smart" code for guessing how many arguments to
      # send to the callback methods that interferes with rspec-mocks.
      around { |e| without_partial_double_verification { e.call } }

      it 'calls #attempt_accept, #check_unexchange, and #process_inventory_unit!' do
        expect(return_item).to receive(:attempt_accept)
        expect(return_item).to receive(:check_unexchange)
        expect(return_item).to receive(:process_inventory_unit!)

        subject
      end
```

---

* [x] Added tests.
* [x] Updated documentation.

### DIFF
--- a/lib/rubocop/cop/rspec/subject_stub.rb
+++ b/lib/rubocop/cop/rspec/subject_stub.rb
@@ -90,7 +90,7 @@ module RuboCop
         def_node_matcher :message_expectation?, <<-PATTERN
           (send
             {
-              (send nil? { :expect :allow } (send nil? {% :subject}))
+              (send nil? { :expect :allow } (send nil? %))
               (send nil? :is_expected)
             }
             #Runners.all
@@ -120,7 +120,7 @@ module RuboCop
             name = subject(child)
             next unless name
 
-            outer_example_group = child.each_ancestor.find do |a|
+            outer_example_group = child.each_ancestor(:block).find do |a|
               example_group?(a)
             end
 
@@ -132,12 +132,11 @@ module RuboCop
         def find_subject_expectations(node, subject_names = [], &block)
           subject_names = @explicit_subjects[node] if @explicit_subjects[node]
 
-          expectation_detected = (subject_names + [:subject]).any? do |name|
-            message_expectation?(node, name)
-          end
+          names = Set[*[*subject_names, :subject]]
+          expectation_detected = message_expectation?(node, names)
           return yield(node) if expectation_detected
 
-          node.each_child_node do |child|
+          node.each_child_node(:send, :block, :begin) do |child|
             find_subject_expectations(child, subject_names, &block)
           end
         end
--- a/lib/rubocop/cop/rspec/subject_stub.rb
+++ b/lib/rubocop/cop/rspec/subject_stub.rb
@@ -53,27 +53,35 @@ module RuboCop
 
         MSG = 'Do not stub methods of the object under test.'
 
-        # @!method subject(node)
+        # @!method subject?(node)
         #   Find a named or unnamed subject definition
         #
         #   @example anonymous subject
-        #     subject(parse('subject { foo }').ast) do |name|
+        #     subject?(parse('subject { foo }').ast) do |name|
         #       name # => :subject
         #     end
         #
         #   @example named subject
-        #     subject(parse('subject(:thing) { foo }').ast) do |name|
+        #     subject?(parse('subject(:thing) { foo }').ast) do |name|
         #       name # => :thing
         #     end
         #
         #   @param node [RuboCop::AST::Node]
         #
         #   @yield [Symbol] subject name
-        def_node_matcher :subject, <<-PATTERN
-            (block
-              (send nil?
-                {:subject (sym $_) | $:subject}
-              ) args ...)
+        def_node_matcher :subject?, <<-PATTERN
+          (block
+            (send nil?
+              {:subject (sym $_) | $:subject}
+            ) args ...)
+        PATTERN
+
+        # @!method let?(node)
+        #   Find a memoized helper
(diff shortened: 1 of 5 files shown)


pirj:
Some cases are no longer detected, probably due to mistaken optimizations:
```ruby
  RSpec.describe Spree.user_class, type: :model do
    context "reporting" do
      def load_orders
        allow(subject).to receive(:spree_orders).and_return(double(complete: orders))
      end
    end
  end
```

I'm checking.

lib/rubocop/cop/rspec/subject_stub.rb
```diff
@@ -136,7 +136,7 @@ def find_subject_expectations(node, subject_names = [], &block)
           expectation_detected = message_expectation?(node, names)
           return yield(node) if expectation_detected
 
-          node.each_child_node(:send, :block, :begin) do |child|
+          node.each_child_node(:send, :def, :block, :begin) do |child|
```
bquorning, line 139: Could you add spec coverage for the `:def` addition please?

bquorning (commented):
I’m fine with this change too :-)

bquorning (approved):

PR #1236, opened 2022-02-18
Add FactoryAssociation cop

Benaaaaa:
Add a FactoryAssociation cop which checks for 'associations' in factories that specify a ':factory' option and tells you to use explicit build reference which optimises performance as cascading factories are not saved unless needed to. 

Pics for cascading factory calls 
before:
old_factory
after:
new_factory

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 * Add bad example to `RSpec/SubjectStub` cop. ([@oshiro3][])
 * Replace non-styleguide cops `StyleGuide` attribute with `Reference`. ([@pirj][])
 * Fix `RSpec/SubjectStub` to disallow stubbing of subjects defined in parent example groups. ([@pirj][])
+* Add `RSpec/FactoryBot/FactoryAssociation` cop. ([@Benaaaaa][])
 
 ## 2.7.0 (2021-12-26)
 
@@ -667,3 +668,4 @@ Compatibility release so users can upgrade RuboCop to 0.51.0. No new features.
 [@leoarnold]: https://github.com/leoarnold
 [@harry-graham]: https://github.com/harry-graham
 [@oshiro3]: https://github.com/oshiro3
+[@Benaaaaa]: https://github.com/Benaaaaa
--- a/config/default.yml
+++ b/config/default.yml
@@ -831,6 +831,11 @@ RSpec/FactoryBot/CreateList:
   VersionChanged: '2.0'
   Reference: https://www.rubydoc.info/gems/rubocop-rspec/RuboCop/Cop/RSpec/FactoryBot/CreateList
 
+RSpec/FactoryBot/FactoryAssociation:
+  Description: Checks for 'association' in factories that specify a ':factory' option. 
+  Enabled: pending
+  VersionAdded: '2.8.1'
+
 RSpec/FactoryBot/FactoryClassName:
   Description: Use string value when setting the class attribute explicitly.
   Enabled: true
--- a/lib/rubocop/cop/rspec/factory_bot/factory_association.rb
+++ b/lib/rubocop/cop/rspec/factory_bot/factory_association.rb
@@ -1,0 +1,101 @@
+# frozen_string_literal: true
+
+require 'rubocop'
+
+module RuboCop
+  module Cop
+    module RSpec
+      module FactoryBot
+        # This cop looks for 'association' in factories that specify a
+        # ':factory' option and tells you to use explicit build reference
+        # which improves the performance as cascading factories
+        # will not be saved unless needed to
+        #
+        # @example
+        #   #bad
+        #    FactoryBot.define do
+        #      factory :book do
+        #        title { 'Lord of the Rings' }
+        #        association :author
+        #      end
+        #    end
+        #
+        #    FactoryBoy.define do
+        #      factory :book do
+        #        title {'Lord of the Rings'}
+        #        author { association :author }
+        #      end
+        #    end
+        #
+        #    #good
+        #    FactoryBot.define do
+        #      factory :book do
+        #        title { 'Lord of the Rings' }
+        #        author { build(:author) }
+        #      end
+        #    end
+        #
+        #    FactoryBot.define do
+        #      factory :author do
+        #        name { 'J. R. R. Tolkien' }
+        #      end
+        #    end
+        class FactoryAssociation < ::RuboCop::Cop::Base
+          extend AutoCorrector
+
+          MSG = 'Use %<association>s { build(:%<factory>s) } instead'
+
+          # @!method association_definition(node)
+          def_node_matcher :association_definition, <<~PATTERN
+            (send nil? :association (:sym $_) (hash (pair (sym :factory) (:sym $_))) ?)
+          PATTERN
+
+          # @!method inline_association_definition(node)
+          def_node_matcher :inline_association_definition, <<~PATTERN
+            (block (send nil? $_) (args) (send nil? :association (sym $_)))
+          PATTERN
+
+          def on_block(node)
+            inline_association_definition(node) do |association, factory|
+              message = format(MSG, association: association.to_s,
+                                    factory: factory.to_s)
+
+              add_offense(node, message: message) do |corrector|
+                corrector.replace(node, replacement(association, factory))
+              end
+            end
+          end
+
+          def on_send(node)
+            return unless @current_corrector.empty?
+
+            association_definition(node) do |association, factory|
+              factory = [association] if factory.empty?
+
+              message = format(MSG, association: association.to_s,
+                                    factory: factory.first.to_s)
+
+              add_offense(node, message: message) do |corrector|
+                corrector.replace(node, replacement(association, factory.first))
+              end
+            end
+          end
+
+          private
+
+          def expression(node)
+            if node.block_type?
+              inline_association_definition(node)
+            else
+              association_definition(node).flatten
+            end
+          end
+
+          def replacement(association, factory)
+            "#{association} { build(:#{factory}) }"
+          end
+        end
+      end
+    end
+  end
+end
--- a/lib/rubocop/cop/rspec_cops.rb
+++ b/lib/rubocop/cop/rspec_cops.rb
@@ -6,6 +6,7 @@ require_relative 'rspec/capybara/visibility_matcher'
 
 require_relative 'rspec/factory_bot/attribute_defined_statically'
 require_relative 'rspec/factory_bot/create_list'
+require_relative 'rspec/factory_bot/factory_association'
 require_relative 'rspec/factory_bot/factory_class_name'
 require_relative 'rspec/factory_bot/syntax_methods'
 
--- a/spec/rubocop/cop/rspec/factory_bot/factory_association_spec.rb
+++ b/spec/rubocop/cop/rspec/factory_bot/factory_association_spec.rb
@@ -1,0 +1,110 @@
+# frozen_string_literal: true
+
+RSpec.describe RuboCop::Cop::RSpec::FactoryBot::FactoryAssociation,
+               :config do
+
+  let(:message) do
+    "Use #{association_name} { build(:#{factory_name}) } instead"
+  end
+
+  context 'when association name and factory name are the same' do
+    let(:association_name) { 'foo' }
+    let(:factory_name) { 'foo' }
+
+    it 'autocorrects method' do
+      expect_offense(<<~RUBY)
+        association :foo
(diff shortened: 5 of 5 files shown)


pirj:
> optimises performance as cascading factories are not saved unless needed to

Can you please elaborate on "unless needed to"?
How it would be possible to create a book in such a way that the author is created, too, and not just built?

```ruby
        #    FactoryBot.define do
        #      factory :book do
        #        title { 'Lord of the Rings' }
        #        author { build(:author) }
        #      end
        #    end
        #
        #    FactoryBot.define do
        #      factory :author do
        #        name { 'J. R. R. Tolkien' }
        #      end
        #    end
```

I see the hardcoded usage of a specific strategy is factories as an anti-pattern, as it prevents certain flexibility.

If your intention is to persist the immediate factory and skip saving all (potentially deeply nested) associations, I'd rather defined a [custom strategy](https://github.com/thoughtbot/factory_bot/blob/master/GETTING_STARTED.md#custom-strategies), e.g. `create_shallow`.

Benaaaaa:
> > optimises performance as cascading factories are not saved unless needed to
> 
> Can you please elaborate on "unless needed to"? How it would be possible to create a book in such a way that the author is created, too, and not just built?
> 
> ```ruby
>         #    FactoryBot.define do
>         #      factory :book do
>         #        title { 'Lord of the Rings' }
>         #        author { build(:author) }
>         #      end
>         #    end
>         #
>         #    FactoryBot.define do
>         #      factory :author do
>         #        name { 'J. R. R. Tolkien' }
>         #      end
>         #    end
> ```
> 
> I see the hardcoded usage of a specific strategy is factories as an anti-pattern, as it prevents certain flexibility.
> 
> If your intention is to persist the immediate factory and skip saving all (potentially deeply nested) associations, I'd rather defined a [custom strategy](https://github.com/thoughtbot/factory_bot/blob/master/GETTING_STARTED.md#custom-strategies), e.g. `create_shallow`.

I'll try to go step by step just to be as detailed as possible. So imagine the scenario where we have the `book` and `author` models and factories defined as above. If we want to test out our `book` model for let's say validations, for most of the time we will be using `build(:book)` instead of the `create(:book)` as those tests are usually faster. In this scenario this is how `build(:book)` will actually look like: `book(id: nil, title: 'Lord of the Rings', author_id: nil)` but in the background the book is still valid, because if you type the `book.valid?` it will return 'true'. You can do the `book.save` and it will save it meaning it will look sth like this: `book(id: nil, title: 'Lord of the Rings', author_id: 1)`.
In case we define our author association this way: `association :author` what `build(:book)` will do is it will `create(:author)` instead of just building it. The book model will look like this: `book(id: nil, title: 'Lord of the Rings', author_id: 1)`. 
So I believe this also answers the question of creating books while creating authors as well. So if we use the `create(:book)` what it will do is `build(:book)` check the `book.valid?` and then do the `book.save` while in the background doing the `author.save`. 
So in essence all this does is actually offers you more flexibility as building a book doesn't mean creating an author but building an author as well, while creating a book means creating an author.

pirj:
> test out our book model for let's say validations, for most of the time we will be using build(:book)

For validations, in most cases `build_stubbed` is sufficient.

What if there's a validation that checks something about the author?
```ruby
class Book < ApplicationRecord
  belongs_to :author
  validates :author_signed_the_agreement

  def author_signed_the_agreement
    errors[:base] << "Author has not signed the agreement, can't add a book!" unless author.agreements.signed.any?
  end
end
```

I believe in this case, with your `build(:author)` approach, `book = create(:book); book.save!` won't work. Because the Author record is expected to be persisted before the Book is.

Moreover, if you have proper foreign key constraints set, and you should, `create(:book)` would blow up saying there's a foreign key violation, as your Author record is not present in the database.

Benaaaaa:
> > test out our book model for let's say validations, for most of the time we will be using build(:book)
> 
> For validations, in most cases `build_stubbed` is sufficient.
> 
> What if there's a validation that checks something about the author?
> 
> ```ruby
> class Book < ApplicationRecord
>   belongs_to :author
>   validates :author_signed_the_agreement
> 
>   def author_signed_the_agreement
>     errors[:base] << "Author has not signed the agreement, can't add a book!" unless author.agreements.signed.any?
>   end
> end
> ```
> 
> I believe in this case, with your `build(:author)` approach, `book = create(:book); book.save!` won't work. Because the Author record is expected to be persisted before the Book is.
> 
> Moreover, if you have proper foreign key constraints set, and you should, `create(:book)` would blow up saying there's a foreign key violation, as your Author record is not present in the database.

Sorry, I wasn't precise enough. If we are calling `create(:book)` it will first save the author and then save the book. So if you are doing `create(:book)` you don't need to do the `book.save!` because it will do it by itself. What I meant by the whole 
`book.valid?` `book.save` is that is what is happening in the background of the `create(:book)` call. But just to be as open as possible: if we call the `build(:book)` it will only build the author, but if we call `create(:book)` it will actually persist the author as well. If I am maybe still not being precise enough I can open a console and show you how it works. 

The concern with the author validation seems legit, I think it still will work through that. Here is code of some local testing:
```ruby
#migration
class CreateTestBook < ActiveRecord::Migration[6.1]
  def change
    create_table :test_books do |t|
      t.belongs_to :author, null:false, foreign_key: { on_delete: :cascade }

      t.string  :title, null:false
      
      t.timestamps
    end
  end
end

#model
class TestBook < ApplicationRecord
  belongs_to :author

  validate :author_name_is_correct
  validates :title, presence: true

  def author_name_is_correct
    errors[:base] << 'Author has no name' unless author.first_name
  end
end

#factory
FactoryBot.define do
  factory :test_book do
    title { 'Lord of the Rings' }

    author { build(:author) }
  end
end

[3] pry(main)> book = FactoryBot.build(:test_book)
=> #<TestBook:0x00007f81ef23ac08 id: nil, author_id: nil, title: "Lord of the Rings", created_at: nil, updated_at: nil>
[4] pry(main)> book.author
=> #<Author:0x00007f81ef2393d0 id: nil, first_name: "Oren", last_name: "Macejkovic", created_at: nil, updated_at: nil>
[5] pry(main)> book.author.test_books
=> []
[6] pry(main)> book.valid?
=> true
[7] pry(main)> book.save
  TRANSACTION (0.4ms)  BEGIN
  Author Create (1.3ms)  INSERT INTO "authors" ("first_name", "last_name", "created_at", "updated_at") VALUES ($1, $2, $3, $4) RETURNING "id"  [["first_name", "Oren"], ["last_name", "Macejkovic"], ["created_at", "2022-02-21 11:47:42.934544"], ["updated_at", "2022-02-21 11:47:42.934544"]]
  TestBook Create (0.7ms)  INSERT INTO "test_books" ("author_id", "title", "created_at", "updated_at") VALUES ($1, $2, $3, $4) RETURNING "id"  [["author_id", 3], ["title", "Lord of the Rings"], ["created_at", "2022-02-21 11:47:42.958221"], ["updated_at", "2022-02-21 11:47:42.958221"]]
  TRANSACTION (6.8ms)  COMMIT
=> true
```
How it works is `book.author` exists because it is built. So if our author factory is properly built for passing validations any `author` object will be passing the validations it needs. And then simply calling the `book.save` will actually first save the `author` and then the `book` meaning the `author` will be persisted before the `book`

pirj:
What happens if you `FactoryBot.create(:book)` with the above code?

Benaaaaa:
> What happens if you `FactoryBot.create(:book)` with the above code?

Pretty much the same as if you do the `book = FactoryBot.build(:book)` and then do `book.save` ( since this is how the process works in the background) :

```ruby
[1] pry(main)> FactoryBot.create(:test_book)
  TRANSACTION (0.4ms)  BEGIN
  Author Create (0.9ms)  INSERT INTO "authors" ("first_name", "last_name", "created_at", "updated_at") VALUES ($1, $2, $3, $4) RETURNING "id"  [["first_name", "Karolyn"], ["last_name", "Schuster"], ["created_at", "2022-02-21 12:21:52.466066"], ["updated_at", "2022-02-21 12:21:52.466066"]]
  TestBook Create (1.7ms)  INSERT INTO "test_books" ("author_id", "title", "created_at", "updated_at") VALUES ($1, $2, $3, $4) RETURNING "id"  [["author_id", 7], ["title", "Lord of the Rings"], ["created_at", "2022-02-21 12:21:52.479134"], ["updated_at", "2022-02-21 12:21:52.479134"]]
  TRANSACTION (41.2ms)  COMMIT
=> #<TestBook:0x00007f81ebf580b0
 id: 5,
 author_id: 7,
 title: "Lord of the Rings",
 created_at: Mon, 21 Feb 2022 12:21:52.479134000 UTC +00:00,
 updated_at: Mon, 21 Feb 2022 12:21:52.479134000 UTC +00:00>
```

Benaaaaa:
Update, it seems that the cop in the end is unnecessary as in FactoryBot version 5.0.0 was introduced implicit using of the same build strategy as their parent objects. No idea why this wasn't working immediately on some of the projects. More on the link: https://github.com/thoughtbot/factory_bot/blob/main/GETTING_STARTED.md#build-strategies-1

pirj:
> It seems that the cop in the end is unnecessary

That slipped my mind completely, too. I still avoid using `build` where possible, choosing between `build_stubbed` and `create`.

Do you think there's anything actionable left here, or the PR can be closed?

Benaaaaa:
I think it can be closed. Thank you for your time and efforts to go into depths of discussion 😄

Benaaaaa:
Upon further inspection and using palkans [test-prof](https://github.com/test-prof/test-prof) I realised that using the `author { build(:author) }` is actually still faster than using `association :author` even though the strategy was introduced in 5.0 to make it work the same. I recommend you test it out in one of your local applications to see the difference. `FPROF=flamegraph` gives a very good visual representation while `FPROF=1` gives a very good numbers representation of factory calls. The only problem that this runs into, which might make it an unsafe cop is if we have a restriction on our `book` that `author_id` can't be null this will fail the `FactoryBot.create(:book)` because `book.valid?` will return `false` and you won't be able to call `book.save`

Benaaaaa:
Upon further inspection and using palkans [test-prof](https://github.com/test-prof/test-prof) I realised that using the `author { build(:author) }` is actually still faster than using `association :author` even though the strategy was introduced in 5.0 to make it work the same. I recommend you test it out in one of your local applications to see the difference. `FPROF=flamegraph` gives a very good visual representation while `FPROF=1` gives a very good numbers representation of factory calls. There seem to be some problems with factory_bot/rails where it doesn't work properly and some specs are failing, for now the only situation that we ran into is where we have a factory with an 2 associations that both specify the same factory, where for some reason it won't make the interviewer, but I think that problem is something else, not this cop.

In the end, it seems that this does work better than using anything else when using an associations in our  factories and using `build`. I understand that you are trying to use `build_stubbed` much more than `build`, but a lot of people are still using `build` and if their associations are defined as such it just makes the whole test suite run much faster

pirj:
If for some reason `association :author`'s performance is worse while it should be the same as `author { build(:author) }`, I suggest raising an issue in FactoryBot repo along with a reproducible case and flamegraphs/profiling results.

I don't find it convincing to add a cop that would recommend changing the spec notation for a specific version of FactoryBot. It is possible, as RuboCop has a `TargetRubyVersion` setting, `rubocop-rails` has `TargetRailsVersion`, but I'm opposed to introduce `TargetFactoryBotVersion`.

It is getting even more complicated if FactoryBot fixes this in, e.g. 6.3.0. Should we recommend using `association` again? How do we know if it's a stylistic mistake, or it was done due to the use of FactoryBot 5.0 - 6.2 that had this performance difference, or it is a deliberate choice to use `author { build(:author) }`?

From my experience on large test suites (tens of hours of suite run time), the time spent in factories code is ~50% of the whole run time. There are multiple ways to overcome this. `test-prof`'s `let_it_be`/`before_all` is one, and a [patch to reuse nested models](https://github.com/test-prof/test-prof/discussions/73#discussioncomment-205671) just to name a few.

Benaaaaa:
Yeah, so in further investigation I've realised that the difference actually only shows when calling `FactoryBot.create`. In the `association :author` case what it does is it uses 2 save points: 1 for the author and then 1 for the book so going sth like:
`book.new => author.create => book.save`, while in the other one it uses 1 savepoint for both author and book, by going like 
`book.new => author.build => book.save` ( which saves both the author and the book into db). 
As you mentioned this might be something that actually should be adressed to FactoryBot. Also thanks for taking your time and providing the `let_it_be` and `before_all` methods, checked them out and they seem to be great for speeding up some tests 😄

Benaaaaa:
Nope, you can close it 🙂

Issue #16, opened 2014-07-30
Use `be` matchers


geniou:
check for things like `eq(true)` and propose to use `be_true` / `be_truthy`

https://www.relishapp.com/rspec/rspec-expectations/v/3-0/docs/built-in-matchers/be-matchers

ochagata:
can this be closed?

geniou:
@ochagata I think this hasn't be done yet snd its still a valid idea. So I would love to have this one day but at the moment I don't have the time to do it on my own.

backus:
This is a bit tough to enforce since the semantics of `#==` (backing `eq`), `#eql?` (backing `eql`), and `#equal?` (backing `be`) are all different. Specifically, `#==` generally means "comparison with coercion" so even if I write `expect(foo.bar).to eq(true)` it does not mean that either `expect(foo.bar).to eql(true)` or `expect(foo.bar).to be(true)` are certain.

I'm leaving this issue open because I need to do some thinking about how much more opinionated I would like to make rubocop-rspec. IMO if you write some object with this behavior:

```ruby
foo == true     # => true
foo.eql?(true) # => false
```

you are probably doing something weird. With that said things become more hairy if we are talking about something like `eq(1)` vs `be(1)`. I would be ok with flagging `eql(1)` -> `be(1)` since `eql?` should not involve coercion. It is totally valid though to say something like `expect(Score.new(1)).to eq(1)`

PR #1238, opened 2022-02-23
Add new `RSpec/BeEq` cop

### LINKED ISSUE #244: Identity comparison vs. value comparison
Perhaps mostly a pet peeve, but I would like to enforce the following styles:

1. Use `be_nil` instead of `be(nil)` or `eq(nil)`.
2. Use `be(true)` instead of `eq(true)`.
3. Use `be(false)` instead of `eq(false)`.

Maybe also

4. Use `to be_nil` instead of `not_to be`.
5. Use `not_to be_nil` instead of `to be`.

bquorning:
Like we have `RSpec/BeEql` to recommend using `be` instead of `eql` where object identity comparison is preferable to comparing with the `#eql?` method, this new `RSpec/BeEq` cop recommends using `be` instead of `eq` where identity comparison is preferable over using the `#==` method.

This cop recommends using `be(nil)` instead of `eq(nil)`, `be(true)` instead of `eq(true)`, and `be(false)` instead of `eq(false)`.

Partially fixes #244.

---

* [x] Added tests.
* [x] Updated documentation.


lib/rubocop/cop/rspec/be_eq.rb
```diff
@@ -0,0 +1,45 @@
...
+      #
+      class BeEq < Base
+        extend AutoCorrector
+
+        MSG = 'Prefer `be` over `eq`.'
+        RESTRICT_ON_SEND = %i[to not_to to_not].freeze
+
+        # @!method eq_type_with_identity(node)
+        def_node_matcher :eq_type_with_identity, <<-PATTERN
+          (send _ #Runners.all $(send nil? :eq {true false nil_type?}))
```
Darhazer, line 32: I just wonder why in the BeNil cop you opted for just `(send nil? :be nil)` and here you're checking that it's being passed to a runner (to/not_to)

Darhazer (approved):

lib/rubocop/cop/rspec/be_eq.rb
bquorning: I stole it from [the BeEql cop](https://github.com/rubocop/rubocop-rspec/blob/766bc1b30a17ade4d36cee8dd310c71e66d5e8a9/lib/rubocop/cop/rspec/be_eql.rb#L44-L47), which applies it only `to` and specifically not to `not_to` and `to_not`. Since the `BeEq` cop can (in my opinion) apply also to negative matchers, maybe I should just change the pattern to `(send nil? :eq {true false nil_type?})`?

pirj, line 32: Cosmetic: can `nil_type?` be just `nil` (not `nil?`)? `BeEql` can be simplified, too.

pirj, line 32: Is there a specific reason why you don't include symbols and integers?

pirj (approved):
Looks good, thank you!

lib/rubocop/cop/rspec/be_eq.rb
bquorning: Yep, that works. Thanks.

bquorning: The comment https://github.com/rubocop/rubocop-rspec/issues/16#issuecomment-237363015 mentions value objects where it’s valid to override `#==` and have an expectation as e.g. `expect(Score.new(1)).to eq(1)`.

bquorning: Fixed, squashed and pushed.

PR #1239, opened 2022-02-23
Add new `RSpec/BeNil` cop

### LINKED ISSUE #244: Identity comparison vs. value comparison

bquorning:
Check that `be_nil` is used instead of `be(nil)`.

RSpec has a built-in `be_nil` matcher specifically for expecting `nil`. For consistent specs, we recommend using that instead of `be(nil)`.

Partially fixes #244.

---

* [x] Added tests.
* [x] Updated documentation.


Darhazer (approved):

pirj (approved):
Looks good, thank you!

marcandre:
Note that it's not 100% safe.

```
expect(OpenStruct.new(nil?: true)).to be_nil # => pass
expect(OpenStruct.new(nil?: true)).to be nil # => fail
```

More importantly, I see absolutely no reason to prefer `be_nil` to `be nil`. The later is stricter, simpler and requires less knowledge of `rspec`.

pirj:
We couldn't [figure out a real example](https://github.com/rubocop/rubocop-rspec/issues/244#issuecomment-1048556473).

bquorning:
> I see absolutely no reason to prefer `be_nil` to `be nil`. The later is stricter, simpler and requires less knowledge of `rspec`.

Well, that _is_ true 😅 Should we consider making the cop configurable to users can enforce either one style or the other?

PR #1241, opened 2022-02-26
Exclude initializers from DescribeClass cop

ixti:
Exclude `spec/initializers` from being enforced by RSpec/DescribeClass cop.

---

* [ ] Added tests.
* [ ] Updated documentation.

### DIFF
--- a/config/default.yml
+++ b/config/default.yml
@@ -195,6 +195,7 @@ RSpec/DescribeClass:
   Enabled: true
   Exclude:
     - "**/spec/features/**/*"
+    - "**/spec/initializers/**/*"
     - "**/spec/requests/**/*"
     - "**/spec/routing/**/*"
     - "**/spec/system/**/*"


ixti:
@bquorning WDYT? If you think it might be merged, I'll update PR to match requirements.

bquorning:
Is the spec/initializers/ folder a “standard” folder in Rails? If we start adding non-standard folders we might end up with a very long list 😅 @pirj what do you think?

pirj:
> we might end up with a very long list

Indeed. I haven’t seen initializers specs. Do you mind sharing an example, @ixti?
Are there such specs in any of ‘real-world-r spec’ repos?

ixti:
`spec/initializers` is not standard (rspec-rails does not create it), but it's a good idea to have those anyway. Here are real-world example:

* https://gitlab.com/gitlab-org/gitlab/-/tree/master/spec/initializers
* https://github.com/discourse/discourse/tree/main/spec/initializers

I've found myself in need to cover initializers with tests after I've caused minor issues in production twice in a row by making some typos in a block that was executed in specific cases only:

``` ruby
Gitlab::Experiment.configure do |config|
  config.nested_behaviour = lambda do |...|
    # this block is called only when experiment is executed from within another experiment.
  end
end
```

ixti:
I'm totally OK to close this PR unmerged, though.
Probably it's pretty "edge case" indeed, and I can go with inline one-by-one disables in my case.

pirj:
﻿
Wondering if there’s an experiment focused spec, integrated, where a spec for this could belong.

For discourse, there’s just one spec. And the initializer has a note:

> Enabling `must_approve_users` on an existing site is odd

And it is what the spec covers, an “odd” case.
Approving users upon site setting change appears to be business logic, and even though it’s hooked in as an event listener, putting it into an initializer is unexpected.

Checking the very first https://gitlab.com/gitlab-org/gitlab/-/blob/master/spec/initializers/00_rails_disable_joins_spec.rb, it seems that it tests some kind of an activerecord extension. Does it belong to initializers? Is the extension itself, ‘DisableJoins’, defined in initializers? Doesn’t it belong to ‘lib’?

ixti:
> Wondering if there’s an experiment focused spec, integrated, where a spec for this could belong.

Yeah. That's what I ended up with :D

``` ruby
RSpec.describe ApplicationExperiment, type: :experiment do
  it "prevents nested experiments" do
    stub_const(
      "InnerExperiment",
      Class.new(described_class) do
        default_rollout :percent, include_control: false
        variant(:foo) { 123 }
      end,
    )

    stub_const(
      "OuterExperiment",
      Class.new(described_class) do
        default_rollout :percent, include_control: false
        variant(:bar) { InnerExperiment.new(:inner).run }
      end,
    )

    aggregate_failures do
      expect { experiment(:outer).run }.to raise_error(Gitlab::Experiment::NestingError)
      expect { experiment(:inner).run }.not_to raise_error
    end
  end
end
```

As ApplicationExperiment is the one other experiments inherit, it's safe to check this behaviour there ;D


> And it is what the spec covers, an “odd” case.

Yup. After openning this PR I've realized that odd and edge cases can go with one-off inline exclusion.

ixti:
And I can't talk for gitlab's initializers. But I agree, it's better to keep initializers as logic-free as possible.

ixti:
Thanks, @pirj and @bquorning, for taking your time.
:heart:

PR #1246, opened 2022-03-05
Add RSpec/VerifiedDoubleReference cop

t3h2mas:
Enforces consistent style for verified double class references. The preferred style is configurable. I left the `Enabled` as `false` because I figure users may want to opt-in to using the cop.

`constant` is the default style since code editors should provide support for:
* displaying warnings for missing or invalid references
* viewing the definition of the doubled class

**Examples**:

`EnforcedStyle: constant`

```ruby
# bad
let(:foo) do
  instance_double('ClassName', method_name: 'returned_value')
end

# good
let(:foo) do
  instance_double(ClassName, method_name: 'returned_value')
end
```

`EnforcedStyle: string`

```ruby
# bad
let(:foo) do
  instance_double(ClassName, method_name: 'returned_value')
end

# good
let(:foo) do
  instance_double('ClassName', method_name: 'returned_value')
end
```

---
* [x] Added tests.
* [x] Updated documentation.


config/default.yml
```diff
@@ -761,6 +761,16 @@ RSpec/VariableName:
   VersionChanged: '1.43'
   Reference: https://www.rubydoc.info/gems/rubocop-rspec/RuboCop/Cop/RSpec/VariableName
 
+RSpec/VerifiedDoubleReference:
+  Description: Checks for consistent verified double reference style.
+  Enabled: false
```
pirj, line 766: I'd suggest [`pending`, this special status](https://docs.rubocop.org/rubocop/versioning.html) was introduced specifically for this purpose.

docs/modules/ROOT/pages/cops_rspec.adoc
```diff
@@ -4300,6 +4300,69 @@ let(:userFood_2) { 'fettuccine' }
...
+
+=== Examples
+
+==== `EnforcedStyle: const`
+
+[source,ruby]
+----
+# bad
+let(:foo) do
+  instance_double("ClassName", method_name: 'returned_value')
```
pirj, line 4328: Cosmetic: make both double quotes or both single quotes?

spec/rubocop/cop/rspec/verified_double_reference_spec.rb
```diff
@@ -0,0 +1,41 @@
...
+      { 'EnforcedStyle' => 'const' }
+    end
+
+    it 'does not flag a violation when using a const reference' do
+      expect_no_offenses('instance_double(ClassName)')
+    end
+
+    it 'flags a violation when using a string reference' do
+      expect_offense(<<-RUBY)
+        instance_double("ClassName")
```
pirj, line 15: Unrelated to this line specifically. What if an instance variable is used?
```ruby
instance_double(class_name)
```

pirj (approved):
Looks good, thank you!

lib/rubocop/cop/rspec/verified_double_reference.rb
```diff
@@ -0,0 +1,72 @@
...
+      #   let(:foo) do
+      #     instance_double('ClassName', method_name: 'returned_value')
+      #   end
+      class VerifiedDoubleReference < Base
+        include ConfigurableEnforcedStyle
+
+        MSG = 'Prefer `%<style>s` references for verfied doubles.'
+
+        SUPPORTED_STYLES = %w[const string].freeze
+        RESTRICT_ON_SEND = %i[instance_double instance_spy class_double].freeze
```
bquorning, line 40: I just now remembered that I have a custom cop similar to this in one of our projects. 😄 

In my version, the list of methods is
```
:mock_model, :stub_model,
:instance_double, :object_double, :class_double,
:instance_spy, :object_spy, :class_spy
```
What do you think? Can we expand the list without problems?

lib/rubocop/cop/rspec/verified_double_reference.rb
```diff
@@ -0,0 +1,72 @@
...
+
+        # @!method verified_double(node)
+        def_node_matcher :verified_double, <<-PATTERN
+          (send nil? {:instance_double :instance_spy :class_double} $_ ...)
```
bquorning, line 44: I _think_ that when `RESTRICT_ON_SEND` is set, you don’t need to repeat the method names here – so e.g. `(send nil? _ $_ ...)` should ok. @pirj can you confirm or debunk that idea?

lib/rubocop/cop/rspec/verified_double_reference.rb
```diff
@@ -0,0 +1,72 @@
...
+      #   # good
+      #   let(:foo) do
+      #     instance_double('ClassName', method_name: 'returned_value')
+      #   end
+      class VerifiedDoubleReference < Base
+        include ConfigurableEnforcedStyle
+
+        MSG = 'Prefer `%<style>s` references for verfied doubles.'
+
+        SUPPORTED_STYLES = %w[const string].freeze
```
bquorning, line 39: I don’t think this constant does anything (and we need to remove it from a couple of our existing gems). I think the `ConfigurableEnforcedStyle` module goes looking in the `SupportedStyles` part of the cop’s configuration.

lib/rubocop/cop/rspec/verified_double_reference.rb
```diff
@@ -0,0 +1,72 @@
...
+
+        def preferred_style?(expectation)
+          map_type(expectation.type).equal?(style)
+        end
+
+        def map_type(type)
+          return :string if type == :str
+
+          type
+        end
```
bquorning, line 68: Did you consider having an actual map (Hash instance) instead of a `#map_type` method?

spec/rubocop/cop/rspec/verified_double_reference_spec.rb
```diff
@@ -0,0 +1,41 @@
...
+                        ^^^^^^^^^^^ Prefer `const` references for verfied doubles.
```
bquorning, line 16: I think users would know better how to resolve the error is we didn’t use `const` in the instructions.
```suggestion
                        ^^^^^^^^^^^ Prefer `constant` references for verfied doubles.
```

lib/rubocop/cop/rspec/verified_double_reference.rb
```diff
@@ -0,0 +1,72 @@
...
+      #   end
+      #
+      #   # good
+      #   let(:foo) do
+      #     instance_double('ClassName', method_name: 'returned_value')
+      #   end
+      class VerifiedDoubleReference < Base
+        include ConfigurableEnforcedStyle
+
+        MSG = 'Prefer `%<style>s` references for verfied doubles.'
```
bquorning, line 37:
```suggestion
        MSG = 'Prefer `%<style>s` references for verified doubles.'
```

lib/rubocop/cop/rspec/verified_double_reference.rb
pirj: Fair enough. This is reasonable to avoid the redundant check with a spec covering that unrelated methods are not flagged.

bquorning: Otherwise, interpolating `RESTRICT_ON_SEND.map(&:inspect).join(' ')` is a way to avoid duplication.

spec/rubocop/cop/rspec/verified_double_reference_spec.rb
t3h2mas: @pirj It will flag a violation. My thought is instance variables to classes could be used by overriding the cop when necessary. Open to opinions. I am making assumptions about how common instance variable class references are used in tests anecdotally.

An alternative would be to allow `lvar` nodes for one or both styles

lib/rubocop/cop/rspec/verified_double_reference.rb
t3h2mas: @bquorning added

lib/rubocop/cop/rspec/verified_double_reference.rb
t3h2mas: Done

lib/rubocop/cop/rspec/verified_double_reference.rb
t3h2mas: I did consider it. Moved to a map now that I'm using `constant` as the style name

lib/rubocop/cop/rspec/verified_double_reference.rb
t3h2mas: Thanks 😅

bquorning (approved):
This looks good. Thank you for the contribution ❤️ 

By the way, I think that adding autocorrection should be a relatively minor change. Would you like to give it a go?

t3h2mas:
> This looks good. Thank you for the contribution ❤️
> 
> By the way, I think that adding autocorrection should be a relatively minor change. Would you like to give it a go?

@bquorning added autocorrect. Let me know if you have feedback

spec/rubocop/cop/rspec/verified_double_reference_spec.rb
pirj: I'm on the fence with this.
Could you please run the cop against https://github.com/pirj/real-world-rspec to see if there are legitimate use cases that would trigger the cop?

lib/rubocop/cop/rspec/verified_double_reference.rb
```diff
@@ -0,0 +1,93 @@
...
+          verified_double(node) do |class_reference|
+            return correct_style_detected if preferred_style?(class_reference)
+
+            message = format(MSG, style: style)
+            expression = class_reference.loc.expression
+
+            add_offense(expression, message: message) do |corrector|
+              violation = class_reference.children.last.to_s
+              corrector.replace(expression, correct_style(violation))
+            end
```
pirj, line 66: Adding https://github.com/rubocop/rubocop/blob/1e55b1aa5e4c5eaeccad5d61f08b7930ed6bc341/lib/rubocop/cop/internal_affairs/style_detected_api_use.rb#L42 would pacify the internal cop.

spec/rubocop/cop/rspec/verified_double_reference_spec.rb
bquorning: From real-world-rspec, here are a few cases I found that result in errors (from `Hash#fetch`) when running this new cop:

- https://github.com/rspec/rspec-mocks/blob/7823b0b542456cfd6193293e3f19e01b7eb934ee/spec/rspec/mocks/formatting_spec.rb#L80-L92
- https://github.com/rspec/rspec-mocks/blob/7823b0b542456cfd6193293e3f19e01b7eb934ee/spec/rspec/mocks/spy_spec.rb#L66-L68
- https://github.com/rspec/rspec-mocks/blob/7823b0b542456cfd6193293e3f19e01b7eb934ee/spec/rspec/mocks/verifying_doubles/construction_spec.rb#L61
- https://github.com/diaspora/diaspora/blob/5d81555ae1009fd35ef1f14473d8cf0956d91d6f/spec/integration/federation/receive_federation_messages_spec.rb#L115
- https://github.com/thoughtbot/administrate/blob/dcc137567b75580a21401558dec3ce0294b794fc/spec/administrate/views/fields/date_time/_index_spec.rb#L6-L12
- https://github.com/instructure/canvas-lms/blob/050120c81d3c6ee362e10d0244f4780f488ecd0d/spec/controllers/gradebooks_controller_spec.rb#L3133

lib/rubocop/cop/rspec/verified_double_reference.rb
t3h2mas: @pirj I think it's worth questioning how stable the RESTRICT_ON_SEND behavior is in rubocop. If the feature is part of the public api, then is rubocop responsible for the tests? I imagine it is considered public since it is shown in a development doc example on https://docs.rubocop.org/rubocop/development.html

I guess the downside to my thinking is if and when there's a breaking change in rubocop.

spec/rubocop/cop/rspec/verified_double_reference_spec.rb
t3h2mas: @bquorning Do you mind sharing the setup you used to run the spec against the real-world-rspec? I was about to dig into issues with `bundle exec rubocop -c ./config/default.yml --only RSpec/VerifiedDoubleReference ../real-world-rspec`

It's now my opinion that this cop should allow local and instance variables by default. This could be taken further to allow anything that isn't the opposite style of the current EnforcedStyle. This should handle the `(array)` from https://github.com/rspec/rspec-mocks/blob/7823b0b542456cfd6193293e3f19e01b7eb934ee/spec/rspec/mocks/formatting_spec.rb#L80-L92.

An alternative could be to add a configuration option for both instance and local variables that determines whether they are allowed. These options would default to true.

lib/rubocop/cop/rspec/verified_double_reference.rb
t3h2mas: For the sake of time, feel free to share a testing approach if you strongly disagree.

My initial thought would be a partial spy on `VerifiedDoubleReference#on_send`

spec/rubocop/cop/rspec/verified_double_reference_spec.rb
bquorning: > Do you mind sharing the setup you used to run the spec against the real-world-rspec? I was about to dig into issues with `bundle exec rubocop -c ./config/default.yml --only RSpec/VerifiedDoubleReference ../real-world-rspec`

I did almost that, but found that the cloned projects’ rubocop configs were causing various problems, so I removed them first.
```
rm -v ../real-world-rspec/**/.rubocop.yml
bundle exec rubocop --only RSpec/VerifiedDoubleReference -- ../real-world-rspec/
```

pirj: > should allow any node type that isn't the opposite style of the current EnforcedStyle

:+1:

lib/rubocop/cop/rspec/verified_double_reference.rb
pirj: I can imagine that RuboCop introduces a new `--skip-performance-optimizations` flag that would not respect `RESTRICT_ON_SEND` when used, without us noticing.

So even though it's mentioned in the public docs

pirj: Edited :point_up:

lib/rubocop/cop/rspec/verified_double_reference.rb
```diff
@@ -0,0 +1,119 @@
...
+        NODE_STYLE_MAP = {
+          str: :string,
+          const: :constant
+        }.freeze
+
+        # @!method verified_double(node)
+        def_node_matcher :verified_double, <<~PATTERN
+          (send
+            nil?
+            %RESTRICT_ON_SEND # unpacks constant
```
pirj, line 68: Cosmetic: % is redundant.
This is [a standard syntax](﻿﻿https://docs.rubocop.org/rubocop-ast/node_pattern.html#const-or-const-for-constants), so I believe a comment is unnecessary.

lib/rubocop/cop/rspec/verified_double_reference.rb
```diff
@@ -0,0 +1,119 @@
...
+            $_ # class reference to inspect
```
pirj, line 69: Cosmetic: can be written as `$_class_reference` if you want to make it clear what it captures exactly.

spec/rubocop/cop/rspec/verified_double_reference_spec.rb
t3h2mas: @prij @bquorning updated to now allow any node type that isn't the opposite style of the current EnforcedStyle

When running against real-world-rspec there are no cop errors, and get this nice output:
> 56602 files inspected, 529 offenses detected, 529 offenses auto-correctable

pirj (approved):
`CHANGELOG.md` needs rebasing, otherwise looks good.

Thanks for the contribution!

t3h2mas:
@pirj rebased and made the cosmetic suggestions

spec/rubocop/cop/rspec/verified_double_reference_spec.rb
```diff
@@ -0,0 +1,90 @@
...
+  end
+end
+
+def aligned_offense(offender, offense)
+  offense_start = offender.index('(')
+  aligned = offender + "\n#{' ' * (offense_start + 1)}#{offense}"
+  <<~RUBY
+    #{aligned}
+  RUBY
+end
```
bquorning, line 90: I understand your pain 😬 We actually have a (well-hidden) helper for these kinds of specs with variable-length indentation – it’s documented at https://github.com/rubocop/rubocop/blob/5b6e3a2d8751a434079f5ce359bf34f358d6c2ea/lib/rubocop/rspec/expect_offense.rb#L75-L102

If you like, I can change the specs to not use `#aligned_offense`. Either I can push a commit to your branch (if possible), or I can do it after merge.

bquorning: Turns out the diff isn’t so bad:
```diff
diff --git a/spec/rubocop/cop/rspec/verified_double_reference_spec.rb b/spec/rubocop/cop/rspec/verified_double_reference_spec.rb
index 930a9145..a90a8e8d 100644
--- a/spec/rubocop/cop/rspec/verified_double_reference_spec.rb
+++ b/spec/rubocop/cop/rspec/verified_double_reference_spec.rb
@@ -13,12 +13,10 @@ RSpec.shared_examples 'verified double references' do |verified_double|
     end
 
     it 'flags a violation when using a string reference' do
-      expect_offense(
-        aligned_offense(
-          "#{verified_double}('ClassName')",
-          '^^^^^^^^^^^ Prefer `constant` references for verified doubles.'
-        )
-      )
+      expect_offense(<<~RUBY, verified_double: verified_double)
+        %{verified_double}('ClassName')
+        _{verified_double} ^^^^^^^^^^^ Prefer `constant` references for verified doubles.
+      RUBY
 
       expect_correction(<<~RUBY)
         #{verified_double}(ClassName)
@@ -40,12 +38,10 @@ RSpec.shared_examples 'verified double references' do |verified_double|
     end
 
     it 'flags a violation when using a constant reference' do
-      expect_offense(
-        aligned_offense(
-          "#{verified_double}(ClassName)",
-          '^^^^^^^^^ Prefer `string` references for verified doubles.'
-        )
-      )
+      expect_offense(<<~RUBY, verified_double: verified_double)
+        %{verified_double}(ClassName)
+        _{verified_double} ^^^^^^^^^ Prefer `string` references for verified doubles.
+      RUBY
 
       expect_correction(<<~RUBY)
         #{verified_double}('ClassName')
@@ -80,11 +76,3 @@ RSpec.describe RuboCop::Cop::RSpec::VerifiedDoubleReference do
     end
   end
 end
-
-def aligned_offense(offender, offense)
-  offense_start = offender.index('(')
-  aligned = offender + "\n#{' ' * (offense_start + 1)}#{offense}"
-  <<~RUBY
-    #{aligned}
-  RUBY
-end
```
(e.g. copy to clipboard and apply with `pbpaste | git apply` if on MacOS)

t3h2mas: @bquorning Nice! Applied

bquorning:
Almost there. Could you run `bundle exec rake` and fix the issue? I think it’s something about a newline in the config file.

t3h2mas:
@bquorning done. FYI, after rebasing from master today, these changes are undone when generating documentation: 
```
diff --git a/docs/modules/ROOT/pages/cops_rspec.adoc b/docs/modules/ROOT/pages/cops_rspec.adoc
index 04d969f..9eec09f 100644
--- a/docs/modules/ROOT/pages/cops_rspec.adoc
+++ b/docs/modules/ROOT/pages/cops_rspec.adoc
@@ -349,7 +349,7 @@ end
 | Name | Default value | Configurable values
 
 | Exclude
-| `spec/spec_helper.rb`, `spec/rails_helper.rb`, `+spec/support/**/*.rb+`
+| `spec/spec_helper.rb`, `spec/rails_helper.rb`, `spec/support/**/*.rb`
 | Array
 |===
 
@@ -1431,7 +1431,7 @@ expect(name).to eq("John")
 | Name | Default value | Configurable values
 
 | Exclude
-| `+spec/routing/**/*+`
+| `spec/routing/**/*`
 | Array
 |===
 
diff --git a/docs/modules/ROOT/pages/cops_rspec_factorybot.adoc b/docs/modules/ROOT/pages/cops_rspec_factorybot.adoc
index c8175da..57401bf 100644
--- a/docs/modules/ROOT/pages/cops_rspec_factorybot.adoc
+++ b/docs/modules/ROOT/pages/cops_rspec_factorybot.adoc
@@ -43,7 +43,7 @@ count { 1 }
 | Name | Default value | Configurable values
 
 | Include
-| `spec/factories.rb`, `+spec/factories/**/*.rb+`, `+features/support/factories/**/*.rb+`
+| `spec/factories.rb`, `spec/factories/**/*.rb`, `features/support/factories/**/*.rb`
 | Array
 |===
 
@@ -100,7 +100,7 @@ create_list :user, 3
 | Name | Default value | Configurable values
 
 | Include
-| `+**/*_spec.rb+`, `+**/spec/**/*+`, `spec/factories.rb`, `+spec/factories/**/*.rb+`, `+features/support/factories/**/*.rb+`
+| `+**/*_spec.rb+`, `+**/spec/**/*+`, `spec/factories.rb`, `spec/factories/**/*.rb`, `features/support/factories/**/*.rb`
 | Array
 
 | EnforcedStyle
@@ -150,7 +150,7 @@ end
 | Name | Default value | Configurable values
 
 | Include
-| `spec/factories.rb`, `+spec/factories/**/*.rb+`, `+features/support/factories/**/*.rb+`
+| `spec/factories.rb`, `spec/factories/**/*.rb`, `features/support/factories/**/*.rb`
 | Array
 |===
 ```

Possibly related to https://github.com/rubocop/rubocop-rspec/commit/1f1a8a9ddbc8c5a1cbf80dbad69d3e6ad18091c2?

This blocks `bundle exec rake` from completing naturally

**Comment by bquorning** [comment]
Probably you are not using RuboCop >= v1.26.0 locally? CI does, and passes 🎉 

@pirj, a final review?

File: spec/rubocop/cop/rspec/verified_double_reference_spec.rb [hunk 14]
```diff
@@ -0,0 +1,78 @@
...
+                     "#{verified_double}('ClassName')",
+                     'string'
+  end
+
+  it 'does not flag a violation when reference is not a supported style' do
+    expect_no_offenses(<<~RUBY)
+      klass = Array
+      instance_double(klass)
+
+
```
bquorning, line 61:
```suggestion
```

spec/rubocop/cop/rspec/verified_double_reference_spec.rb
```diff
@@ -0,0 +1,78 @@
...
+        #{verified_double}('ClassName')
+      RUBY
+    end
+
+    include_examples 'detects style',
+                     "#{verified_double}('ClassName')",
+                     'string'
+  end
+
+  it 'does not flag a violation when reference is not a supported style' do
```
bquorning, line 56: This example should be run just once, i.e. not be a shared example.

lib/rubocop/cop/rspec/verified_double_reference.rb
```diff
@@ -0,0 +1,120 @@
...
+
+        def preferred_style?(class_reference)
+          class_reference_type = class_reference.type
+
+          return true unless suspicious?(class_reference_type)
+
+          as_style(class_reference_type).equal?(style)
+        end
+
+        def suspicious?(class_reference_node)
```
bquorning, line 100: The parameter `class_reference_node` seems off – we don’t pass a node to this method. The same argument is passed to `#as_style` where it is called `type` instead.

t3h2mas: @bquorning changed map from `NODE_STYLE_MAP` -> `REFERENCE_TYPE_STYLES`

Updated the parameter for both `#as_style` and` #suspicious?` to `reference_type`

Let me know what you think about these changes.

CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 # Changelog
 
 ## Master (Unreleased)
+* Add `RSpec/VerifiedDoubleReference` cop. ([@t3h2mas][])
 
 * Fix a false positive for `RSpec/EmptyExampleGroup` when expectations in case statement. ([@ydah][])
```
```suggestion

* Add `RSpec/VerifiedDoubleReference` cop. ([@t3h2mas][])
* Fix a false positive for `RSpec/EmptyExampleGroup` when expectations in case statement. ([@ydah][])
```

spec/rubocop/cop/rspec/verified_double_reference_spec.rb
```diff
@@ -0,0 +1,78 @@
...
+        #{verified_double}('ClassName')
+      RUBY
+    end
+
+    include_examples 'detects style',
+                     "#{verified_double}('ClassName')",
+                     'string'
+  end
+
+end
```
bquorning, line 56: CI (or `bundle exec rake`) complains about this one:
```suggestion
  end
end
```

bquorning: Sorry @t3h2mas, I force-pushed on your branch to resolve this issue. CI is green now 🎉

lib/rubocop/cop/rspec/verified_double_reference.rb
```diff
@@ -0,0 +1,120 @@
...
+
+        private
+
+        def preferred_style?(class_reference)
+          class_reference_type = class_reference.type
+
+          return true unless suspicious?(class_reference_type)
+
+          as_style(class_reference_type).equal?(style)
+        end
```
pirj, line 98: Cosmetic. Can also be written as:
```ruby
def opposed_style?(class_reference)
  class_reference_style = REFERENCE_TYPE_STYLES[class_reference.type] # nil, :string, or :constant

  return false unless class_reference_style # Reference is not in the supported style list. No enforcement

  class_reference_style != style
end
```

My nitpick is that `preferred_style?` returns `true` when "reference is not in the supported style list", when it's not semantially a preferred style, it's an ignored style.

This would allow getting rid of the two methods below, and only requires a one-line change in the consumer:
```diff
- return correct_style_detected if preferred_style?(class_reference)
+ return correct_style_detected unless opposed_style?(class_reference)
```

Still, leaving it up to you, the current approach is quite comprehendible, too.

spec/rubocop/cop/rspec/verified_double_reference_spec.rb
```diff
@@ -0,0 +1,77 @@
...
+    end
+
+    it 'does not flag a violation when using a constant reference' do
+      expect_no_offenses("#{verified_double}(ClassName)")
+    end
+
+    it 'flags a violation when using a string reference' do
+      expect_offense(<<~RUBY, verified_double: verified_double)
+        %{verified_double}('ClassName')
+        _{verified_double} ^^^^^^^^^^^ Prefer `constant` references for verified doubles.
```
pirj, line 18: Why is "constant" in backticks?
Would "Use a literal constant to refer to a class for verified doubles" be an easier to understand offence hint?

spec/rubocop/cop/rspec/verified_double_reference_spec.rb
```diff
@@ -0,0 +1,77 @@
+# frozen_string_literal: true
+
+RSpec.shared_examples 'verified double references' do |verified_double|
```
pirj, line 3: Cosmetic. Is there a reason to have this shared example as a gobal, shared one? Or is it sufficient to nest it under `RSpec.describe RuboCop::Cop::RSpec::VerifiedDoubleReference do`?

spec/rubocop/cop/rspec/verified_double_reference_spec.rb
```diff
@@ -0,0 +1,77 @@
...
+    end
+
+    include_examples 'detects style',
+                     "#{verified_double}('ClassName')",
+                     'string'
+  end
+end
+
+RSpec.describe RuboCop::Cop::RSpec::VerifiedDoubleReference do
+  described_class::RESTRICT_ON_SEND.each do |verified_double|
```
pirj, line 58: I strive to refrain from referencing constants from spec code. Imagine, you remove `class_double` from `RESTRICT_ON_SEND`, and the spec remains green.

spec/rubocop/cop/rspec/verified_double_reference_spec.rb
```diff
@@ -0,0 +1,77 @@
...
+    describe verified_double do
+      it_behaves_like 'verified double references', verified_double
```
pirj, line 60: Since this only includes shared examples, and those shared examples are exclusively used from here, I believe it makes sense to just inline those shared examples.

pirj (approved):
Looks good overall. Only styling notes, not a blocker to merge.

Thanks a lot for your contribution!

lib/rubocop/cop/rspec/verified_double_reference.rb
t3h2mas: Great suggestion. I appreciate that it clarifies the behavior and drops the LOC.

Updated

spec/rubocop/cop/rspec/verified_double_reference_spec.rb
t3h2mas: Removed the backticks, updated the language to be a compromise between the prior language and your suggestion.

spec/rubocop/cop/rspec/verified_double_reference_spec.rb
t3h2mas: Handled by inlining the shared examples

spec/rubocop/cop/rspec/verified_double_reference_spec.rb
t3h2mas: Good point, don't want to break public behavior. Copied the array into the test.

spec/rubocop/cop/rspec/verified_double_reference_spec.rb
t3h2mas: Fine with me, inlined

t3h2mas:
@pirj back to you for review

spec/rubocop/cop/rspec/verified_double_reference_spec.rb
```diff
@@ -1,10 +1,24 @@
...
+    stub_model
+  ]
+
+  # described_class is overwritten in the inner describe block
+  # rubocop:disable RSpec/DescribedClass
+  let(:cop_class) { RuboCop::Cop::RSpec::VerifiedDoubleReference }
+  # rubocop:enable RSpec/DescribedClass
 
+  verified_doubles.each do |verified_double|
+    describe verified_double do
```
bquorning, line 21: If you give a string argument to `describe` – e.g. just `describe verified_double.to_s do`, then described_class does not get overwritten, and lines 15-18 can be removed.

lib/rubocop/cop/rspec/verified_double_reference.rb
```diff
@@ -73,7 +73,7 @@ class VerifiedDoubleReference < Base
 
         def on_send(node)
           verified_double(node) do |class_reference|
-            return correct_style_detected unless opposing_style?(class_reference)
+            break correct_style_detected unless opposing_style?(class_reference)
```
bquorning, line 76: A bit of a hack to change the keyword because the line got too long 😄

spec/rubocop/cop/rspec/verified_double_reference_spec.rb
pirj: Fixed with
```diff
- verified_doubles = %i[
+ verified_doubles = %w[
 ```

**Comment by pirj** [comment]
@t3h2mas Thank you!
I dared to make a couple of final cosmetic touches.

**Comment by pirj** [comment]
@bquorning For some reason, I get dirty files after running `rake generate_cops_documentation` locally, even after `bundle update`:
image
But CI is silent about any discrepancies. Can you reproduce this?

File: lib/rubocop/cop/rspec/verified_double_reference.rb (hunk 25 above)
Comment (pirj) [reply lib/rubocop/cop/rspec/verified_double_reference.rb:76]: Even `next` would probably work, so there's a one-character space for improvement left 😆

**Comment by bquorning** [comment]
What does `bundle exec rubocop -V` return @pirj. Maybe you have a `Gemfile.local` preventing you from using RuboCop >= v1.26.0?

**Comment by pirj** [comment]
@bquorning Ah, exactly. It was pointed to a local stale `rubocop` repo. All good now, thank you!

File: spec/rubocop/cop/rspec/verified_double_reference_spec.rb (hunk 24 above)
Comment (t3h2mas) [reply spec/rubocop/cop/rspec/verified_double_reference_spec.rb:16]: TIL 🙌

**Comment by schmijos** [comment]
This cop breaks a stubbing pattern:

```rb
stub_const('Sentry', instance_double('Sentry'))
```

The following doesn't work anymore in environments where `Sentry` is not available.

```rb
stub_const('Sentry', instance_double(Sentry))
```

pirj:
@schmijos I don't quite understand what you are trying to achieve with:
```ruby
stub_const('Sentry', instance_double('Sentry'))
```
Can you please elaborate?

Just by looking at it, and with the information you've provided, it seems that you are attempting to assign a regular, non-verified double to a class constant. Is that what you want, really? How do you use `Sentry` in your code afterwards? Do you stub its methods?

I see two problems here. The first is that even if `Sentry` was available, after this statement it wouldn't be pointing to a class `Sentry`, but rather an instance of that class. So you would better be off with a `class_double('Sentry')`.

And if your intention is to disregard all calls to `Sentry` made inside the code under test, you might want to use a ["null double"](https://relishapp.com/rspec/rspec-mocks/v/3-9/docs/basics/null-object-doubles):
```ruby
stub_const('Sentry', double('Sentry').as_null_object)
```

In case if `RSpec/VerifiedDoubles` complains that you should use `instance_double` instead of `double('Sentry').as_null_object`, it makes sense to either:
1. Fix `RSpec/VerifiedDoubles` to ignore `double` chained with `as_null_object`
2. Use a symbol as the class name `double(:Sentry).as_null_object` and set `IgnoreSymbolicNames` to `true`
3. Use a nameless double `double.as_null_object` as the cop [ignores them by default](https://github.com/rubocop/rubocop-rspec/blob/af2fd8a216b0ed7846372672a9eadf5db87126b2/config/default.yml#L788)

PR #1251, opened 2022-03-20
Make `RSpec/BeNil` cop configurable

bquorning:
Make `RSpec/BeNil` cop configurable with a :be_nil and a :be style.

Inspired by @marcandre's comment at https://github.com/rubocop/rubocop-rspec/pull/1239#issuecomment-1067532133

> I see absolutely no reason to prefer `be_nil` to `be nil`. The latter is stricter, simpler and requires less knowledge of rspec.

---

* [x] Added tests.
* [x] Updated documentation.

### DIFF
--- a/config/default.yml
+++ b/config/default.yml
@@ -155,9 +155,14 @@ RSpec/BeEql:
   Reference: https://www.rubydoc.info/gems/rubocop-rspec/RuboCop/Cop/RSpec/BeEql
 
 RSpec/BeNil:
-  Description: Check that `be_nil` is used instead of `be(nil)`.
+  Description: Ensures a consistent style is used when matching `nil`.
   Enabled: pending
+  EnforcedStyle: be_nil
+  SupportedStyles:
+    - be
+    - be_nil
   VersionAdded: 2.9.0
+  VersionChanged: 2.10.0
   Reference: https://www.rubydoc.info/gems/rubocop-rspec/RuboCop/Cop/RSpec/BeNil
 
 RSpec/BeforeAfterAll:
--- a/lib/rubocop/cop/rspec/be_nil.rb
+++ b/lib/rubocop/cop/rspec/be_nil.rb
@@ -3,24 +3,39 @@
 module RuboCop
   module Cop
     module RSpec
-      # Check that `be_nil` is used instead of `be(nil)`.
+      # Ensures a consistent style is used when matching `nil`.
       #
-      # RSpec has a built-in `be_nil` matcher specifically for expecting `nil`.
-      # For consistent specs, we recommend using that instead of `be(nil)`.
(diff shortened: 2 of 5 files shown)


config/default.yml
```diff
@@ -155,9 +155,14 @@ RSpec/BeEql:
   Reference: https://www.rubydoc.info/gems/rubocop-rspec/RuboCop/Cop/RSpec/BeEql
 
 RSpec/BeNil:
-  Description: Check that `be_nil` is used instead of `be(nil)`.
+  Description: Ensures a consistent style is used when matching `nil`.
   Enabled: pending
+  EnforcedStyle: be_nil
```
pirj, line 160: On `real-world-rspec`:
```
$ rg --no-ignore --hidden --stats 'be_nil'
10881 matches

$ rg --no-ignore --hidden --stats 'be[ \(]nil'
1290 matches
```
The latter includes text matches in strings and comments (e.g. "cannot be nil when locking the attribute").

So 👍 for enforcing `be_nil`.

config/default.yml
```diff
@@ -155,9 +155,14 @@ RSpec/BeEql:
...
+  SupportedStyles:
+    - be
+    - be_nil
   VersionAdded: 2.9.0
+  VersionChanged: 3.0.0
```
pirj, line 165: There's [another active PR targeting 2.10](https://github.com/rubocop/rubocop-rspec/pull/1246/files#diff-3653213e3a153018a75cc7b2536b1d0217a6d8d75188d71cf45441812f789698R771). The enforced style here remains the same.
Should this go to 2.10, too?

pirj (approved):
Thank you!

bquorning:
Let’s get #1246 merged first, before merging this PR.

lib/rubocop/cop/rspec/be_nil.rb
```diff
@@ -3,34 +3,68 @@
...
       #
-      # RSpec has a built-in `be_nil` matcher specifically for expecting `nil`.
-      # For consistent specs, we recommend using that instead of `be(nil)`.
+      # You can either use the more specific `be_nil` matcher, or the more
+      # generic `be` matcher with a `nil` argument.
       #
-      # @example
+      # This cop can be configured using the `EnforcedStyle` option
       #
+      # @example `EnforcedStyle: be_nil`
```
koic, line 13: Adding `(default)` would be more user-friendly :-)

```suggestion
      # @example `EnforcedStyle: be_nil` (default)
```

bquorning: 👍🏼 It would be nice if we had a way to check that documentation includes example for all different `EnforcedStyle`s _and_ that it marks which of them is the default.

PR #1252, opened 2022-03-21
[Fix #1237] Fix `Capybara/CurrentPathExpectation` autocorrect incompatible with `Style/TrailingCommaInArguments` autocorrect

### LINKED ISSUE #1237: Capybara/CurrentPathException autocorrect incompatible with Style/TrailingComma autocorrect
rubocop version: 1.18.3
rubocop-rspec version: 2.4.0

Example code:

```ruby
# original code
expect(current_path).to eq(
  some_path(
    record_id: record.id
  )
)
```

If Style/TrailingComma (in vanilla rubocop) is enabled for methods, it adds a comma after `record.id` and after the closed parenthesis.

Also, Capybara/CurrentPathExpectation adds `, ignore_query: true` to the `have_current_path` params. If both are fixed with the same autocorrect, the following is the result:

```ruby
# after autocorrect
expect(page).to have_current_path(
  some_path(
    record_id: record.id
  ),, ignore_query: true
)
```

which is a syntax error (the double-comma) and still doesn't have the trailing comma.

A workaround is to run the Style cops first, then the RSpec cops, or vice-versa - just not in the same command

ydah:
Fixes https://github.com/rubocop/rubocop-rspec/issues/1237

This PR fixes `Capybara/CurrentPathExpectation` autocorrect incompatible with `Style/TrailingCommaInArguments` autocorrect when code like the following:

```ruby
expect(current_path).to eq(
  some_path(
    record_id: record.id
  )
)
```

---

* [x] Added tests.
* [ ] Updated documentation.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,9 +1,10 @@
 # Changelog
 
 ## Master (Unreleased)
-* Add `RSpec/VerifiedDoubleReference` cop. ([@t3h2mas][])
 
 * Fix a false positive for `RSpec/EmptyExampleGroup` when expectations in case statement. ([@ydah][])
+* Add `RSpec/VerifiedDoubleReference` cop. ([@t3h2mas][])
+* Make `RSpec/BeNil` cop configurable with a `be_nil` style and a `be` style. ([@bquorning][])
 
 ## 2.9.0 (2022-02-28)
 
--- a/config/default.yml
+++ b/config/default.yml
@@ -155,9 +155,14 @@ RSpec/BeEql:
   Reference: https://www.rubydoc.info/gems/rubocop-rspec/RuboCop/Cop/RSpec/BeEql
 
 RSpec/BeNil:
-  Description: Check that `be_nil` is used instead of `be(nil)`.
+  Description: Ensures a consistent style is used when matching `nil`.
   Enabled: pending
+  EnforcedStyle: be_nil
+  SupportedStyles:
+    - be
+    - be_nil
   VersionAdded: 2.9.0
+  VersionChanged: 2.10.0
   Reference: https://www.rubydoc.info/gems/rubocop-rspec/RuboCop/Cop/RSpec/BeNil
 
 RSpec/BeforeAfterAll:
--- a/docs/modules/ROOT/pages/cops_rspec.adoc
+++ b/docs/modules/ROOT/pages/cops_rspec.adoc
@@ -283,15 +283,19 @@ expect(foo).to be(nil)
 | Yes
 | Yes
 | 2.9.0
-| -
+| 2.10.0
 |===
 
-Check that `be_nil` is used instead of `be(nil)`.
+Ensures a consistent style is used when matching `nil`.
 
-RSpec has a built-in `be_nil` matcher specifically for expecting `nil`.
-For consistent specs, we recommend using that instead of `be(nil)`.
+You can either use the more specific `be_nil` matcher, or the more
+generic `be` matcher with a `nil` argument.
+
+This cop can be configured using the `EnforcedStyle` option
 
 === Examples
+
+==== `EnforcedStyle: be_nil` (default)
 
 [source,ruby]
 ----
@@ -301,6 +305,27 @@ expect(foo).to be(nil)
 # good
 expect(foo).to be_nil
 ----
+
+==== `EnforcedStyle: be`
+
+[source,ruby]
+----
+# bad
+expect(foo).to be_nil
+
+# good
+expect(foo).to be(nil)
+----
+
+=== Configurable attributes
+
+|===
+| Name | Default value | Configurable values
+
+| EnforcedStyle
+| `be_nil`
+| `be`, `be_nil`
+|===
 
 === References
 
--- a/lib/rubocop/cop/rspec/be_nil.rb
+++ b/lib/rubocop/cop/rspec/be_nil.rb
@@ -3,24 +3,39 @@
 module RuboCop
   module Cop
     module RSpec
-      # Check that `be_nil` is used instead of `be(nil)`.
+      # Ensures a consistent style is used when matching `nil`.
       #
-      # RSpec has a built-in `be_nil` matcher specifically for expecting `nil`.
-      # For consistent specs, we recommend using that instead of `be(nil)`.
+      # You can either use the more specific `be_nil` matcher, or the more
+      # generic `be` matcher with a `nil` argument.
       #
-      # @example
+      # This cop can be configured using the `EnforcedStyle` option
       #
+      # @example `EnforcedStyle: be_nil` (default)
       #   # bad
       #   expect(foo).to be(nil)
       #
       #   # good
       #   expect(foo).to be_nil
       #
+      # @example `EnforcedStyle: be`
+      #   # bad
+      #   expect(foo).to be_nil
+      #
+      #   # good
+      #   expect(foo).to be(nil)
+      #
       class BeNil < Base
         extend AutoCorrector
+        include ConfigurableEnforcedStyle
 
-        MSG = 'Prefer `be_nil` over `be(nil)`.'
-        RESTRICT_ON_SEND = %i[be].freeze
+        BE_MSG = 'Prefer `be(nil)` over `be_nil`.'
+        BE_NIL_MSG = 'Prefer `be_nil` over `be(nil)`.'
+        RESTRICT_ON_SEND = %i[be be_nil].freeze
+
+        # @!method be_nil_matcher?(node)
+        def_node_matcher :be_nil_matcher?, <<-PATTERN
+          (send nil? :be_nil)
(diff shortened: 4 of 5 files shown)


lib/rubocop/cop/rspec/capybara/current_path_expectation.rb
```diff
@@ -103,11 +103,22 @@ def add_ignore_query_options(corrector, node)
...
-              ', ignore_query: true'
+              "#{comma} ignore_query: true"
             )
           end
+
+          def expectation_last_child_comma?(node, last_child)
+            trailing_config = config.for_cop('Style/TrailingCommaInArguments')
+            return true unless trailing_config['Enabled']
+
+            trailing_config['EnforcedStyleForMultiline'] == 'no_comma' ||
```
bquorning, line 119: I think this is too tight a coupling between RuboCop-RSpec and RuboCop’s internals.

Can we get away with just defining an `.autocorrect_incompatible_with` method instead?

bquorning, line 119: I think this is too tight a coupling between RuboCop-RSpec and RuboCop’s internals.

Can we get away with just defining an `.autocorrect_incompatible_with` method instead?

CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 ## Master (Unreleased)
 
 * Fix a false positive for `RSpec/EmptyExampleGroup` when expectations in case statement. ([@ydah][])
+* Fix `Capybara/CurrentPathExpectation` autocorrect incompatible with `Style/TrailingComma` autocorrect. ([@ydah][])
```
```suggestion
* Fix `Capybara/CurrentPathExpectation` autocorrect incompatible with `Style/TrailingCommaInArguments` autocorrect. ([@ydah][])
```

```suggestion
* Fix `Capybara/CurrentPathExpectation` autocorrect incompatible with `Style/TrailingCommaInArguments` autocorrect. ([@ydah][])
```

lib/rubocop/cop/rspec/capybara/current_path_expectation.rb
ydah: It is better that way. Thank you so much.
Use `.autocorrect_incompatible_with` method.

CHANGELOG.md
ydah: Oops sorry, will update PR.

bquorning (commented):
The added FileHelper is almost identical to spec helpers in [rubocop](https://github.com/rubocop/rubocop/blob/2d8ea1502f369b6de62c48bcbb9bd2e88001011f/spec/support/file_helper.rb) and in [rubocop-rails](https://github.com/rubocop/rubocop-rails/blob/a098c4f27ef39cf75a3209419a5555b2931946ac/spec/support/file_helper.rb), and the "cli spec behavior" shared context is almost identical to shared contexts in [rubocop](https://github.com/rubocop/rubocop/blob/2d8ea1502f369b6de62c48bcbb9bd2e88001011f/spec/support/cli_spec_behavior.rb) and [rubocop-rails](https://github.com/rubocop/rubocop-rails/blob/a098c4f27ef39cf75a3209419a5555b2931946ac/spec/support/cli_spec_behavior.rb).

I wonder if we should push for that code to be moved into [the shared contexts in rubocop's lib folder](https://github.com/rubocop/rubocop/blob/2d8ea1502f369b6de62c48bcbb9bd2e88001011f/lib/rubocop/rspec/shared_contexts.rb) so they can shared across repositories.

cc @koic, you probably have an opinion on that matter.

lib/rubocop/cop/rspec/capybara/current_path_expectation.rb
```diff
@@ -52,6 +52,19 @@ class CurrentPathExpectation < Base
...
           PATTERN
 
+          # Monkey patching for `Style/TrailingCommaInArguments` of RuboCop core
+          # rubocop:disable Style/ClassAndModuleChildren
+          class Style::TrailingCommaInArguments
+            def self.autocorrect_incompatible_with
+              [RSpec::Capybara::CurrentPathExpectation]
+            end
+          end
+          # rubocop:enable Style/ClassAndModuleChildren
```
bquorning, line 62: The wording tell me this code was inspired by rubocop-rails which has the "inverse" autocorrect_incompatible_with definition inside the cop, as here.

In rubocop-rspec we have these inverse definitions [here](https://github.com/rubocop/rubocop-rspec/blob/b56d28749734b2985799d2abbf810c91b874ff03/lib/rubocop-rspec.rb#L40-L50). Not sure if that is better or worse, but let’s at least have it be consistent 😄

ydah: Oops, sorry. I had overlooked that💦
I moved the inverse definitions. Thank you so much!

lib/rubocop-rspec.rb
```diff
@@ -37,16 +37,26 @@
...
         def self.autocorrect_incompatible_with
           [RSpec::AlignLeftLetBrace, RSpec::AlignRightLetBrace]
         end
       end
     end
+
+    module Style
+      class TrailingCommaInArguments
+        def self.autocorrect_incompatible_with
+          [RSpec::Capybara::CurrentPathExpectation]
```
Darhazer, line 54: doesn't we override whatever is set in TrailingCommaInArguments, potentially breaking other extensions or future cops? Shouldn't we declare the incompatibility only in our own cop? If the incompatibility has to be declared in both cops, should we include a module which would then append to whatever `super` returns?

cc: @bquorning

bquorning: You’re right.

ydah: You are right.
I tried declaring incompatibility only in our own cop, but it did not work out as expected.

So I'm trying to solve this problem in the following way:
> If the incompatibility has to be declared in both cops, should we include a module which would then append to whatever super returns?

Result of running RSpec when we declared incompatibility only in our own cop:
```
bundle exec rspec spec/rubocop/cli/autocorrect_spec.rb:35
Run options: include {:focus=>true, :locations=>{"./spec/rubocop/cli/autocorrect_spec.rb"=>[35]}}

Randomized with seed 40281

RuboCop::CLI --autocorrect
  when corrects `RSpec/Capybara/CurrentPathExpectation` with `Style/TrailingCommaInArguments`
    autocorrects be compatible with each other (FAILED - 1)

Failures:

  1) RuboCop::CLI --autocorrect when corrects `RSpec/Capybara/CurrentPathExpectation` with `Style/TrailingCommaInArguments` autocorrects be compatible with each other
     Failure/Error:
       expect(File.read('spec/example.rb')).to eq(<<-RUBY)
         expect(page).to have_current_path(
           some_path(
             id: id,
           ), ignore_query: true
         )
       RUBY
     
       expected: "        expect(page).to have_current_path(\n          some_path(\n            id: id,\n          ), ignore_query: true\n        )\n"
            got: "        expect(page).to have_current_path(\n          some_path(\n            id: id,\n          ),, ignore_query: true\n        )\n"
     
       (compared using ==)
     
       Diff:
       @@ -1,6 +1,6 @@
                expect(page).to have_current_path(
                  some_path(
                    id: id,
       -          ), ignore_query: true
       +          ),, ignore_query: true
                )
       
     # ./spec/rubocop/cli/autocorrect_spec.rb:40:in `block (3 levels) in <top (required)>'
     # ./spec/support/cli_spec_behavior.rb:21:in `block (2 levels) in <top (required)>'
     # ./vendor/bundle/ruby/2.6.0/gems/rubocop-1.25.1/lib/rubocop/rspec/shared_contexts.rb:30:in `block (4 levels) in <top (required)>'
     # ./vendor/bundle/ruby/2.6.0/gems/rubocop-1.25.1/lib/rubocop/rspec/shared_contexts.rb:30:in `chdir'
     # ./vendor/bundle/ruby/2.6.0/gems/rubocop-1.25.1/lib/rubocop/rspec/shared_contexts.rb:30:in `block (3 levels) in <top (required)>'
     # ./vendor/bundle/ruby/2.6.0/gems/rubocop-1.25.1/lib/rubocop/rspec/shared_contexts.rb:7:in `block (2 levels) in <top (required)>'

Finished in 0.13613 seconds (files took 1.13 seconds to load)
1 example, 1 failure

Failed examples:

rspec ./spec/rubocop/cli/autocorrect_spec.rb:35 # RuboCop::CLI --autocorrect when corrects `RSpec/Capybara/CurrentPathExpectation` with `Style/TrailingCommaInArguments` autocorrects be compatible with each other

Randomized with seed 40281
```

ydah:
I accidentally Close. Sorry.
We are creating the PR again: https://github.com/rubocop/rubocop-rspec/pull/1253

ydah:
@bquorning @Darhazer 
I accidentally Close. Sorry.
I'm creating the PR again: https://github.com/rubocop/rubocop-rspec/pull/1253

PR #1253, opened 2022-03-25
[Fix #1237] Fix `Capybara/CurrentPathExpectation` autocorrect incompatible with `Style/TrailingCommaInArguments` autocorrect

### LINKED ISSUE #1237: Capybara/CurrentPathException autocorrect incompatible with Style/TrailingComma autocorrect

ydah:
I accidentally Close: https://github.com/rubocop/rubocop-rspec/pull/1252 Sorry.

Fixes https://github.com/rubocop/rubocop-rspec/issues/1237

This PR fixes `Capybara/CurrentPathExpectation` autocorrect incompatible with `Style/TrailingCommaInArguments` autocorrect when code like the following:

```ruby
expect(current_path).to eq(
  some_path(
    record_id: record.id
  )
)
```

---

* [x] Added tests.
* [ ] Updated documentation.

### DIFF
--- a/lib/rubocop-rspec.rb
+++ b/lib/rubocop-rspec.rb
@@ -37,16 +37,21 @@ require_relative 'rubocop/cop/rspec_cops'
 # We have to register our autocorrect incompatibilities in RuboCop's cops
 # as well so we do not hit infinite loops
 
-module RuboCop
-  module Cop
-    module Layout
-      class ExtraSpacing # rubocop:disable Style/Documentation
-        def self.autocorrect_incompatible_with
-          [RSpec::AlignLeftLetBrace, RSpec::AlignRightLetBrace]
-        end
-      end
+RuboCop::Cop::Layout::ExtraSpacing.singleton_class.prepend(
+  Module.new do
+    def autocorrect_incompatible_with
+      super.push(RuboCop::Cop::RSpec::AlignLeftLetBrace)
+      .push(RuboCop::Cop::RSpec::AlignRightLetBrace)
     end
   end
-end
+)
+
+RuboCop::Cop::Style::TrailingCommaInArguments.singleton_class.prepend(
+  Module.new do
+    def autocorrect_incompatible_with
+      super.push(RuboCop::Cop::RSpec::Capybara::CurrentPathExpectation)
+    end
+  end
+)
 
 RuboCop::AST::Node.include(RuboCop::RSpec::Node)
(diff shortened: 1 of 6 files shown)


lib/rubocop-rspec.rb
```diff
@@ -37,16 +37,36 @@
...
+        end
+
+        def autocorrect_incompatible_with_extend(*args)
+          TrailingCommaInArguments.autocorrect_incompatible_with_extend(*args)
+        end
+
+        class << self
+          alias autocorrect_incompatible_with_org autocorrect_incompatible_with
+          alias autocorrect_incompatible_with autocorrect_incompatible_with_ext
+        end
```
bquorning, line 65: Instead of an alias-method chain, we could do module-prepending which I think is cleaner:
```rb
RuboCop::Cop::Layout::ExtraSpacing.singleton_class.prepend(Module.new do
  def autocorrect_incompatible_with
    [RSpec::AlignLeftLetBrace, RSpec::AlignRightLetBrace]
  end
end)

RuboCop::Cop::Style::TrailingCommaInArguments.singleton_class.prepend(Module.new do
  def autocorrect_incompatible_with
    [RSpec::Capybara::CurrentPathExpectation]
  end
end)
```

Of course this is very much repetitive, so perhaps later we should consider adding a class method to `Cop` that can add all this code with a single call. But definitely not for this PR though.

Darhazer: IMO it should be `append` and concating to the `super`, otherwise it will still be just replaced by the original code

bquorning: 🤦🏼 Of course, I’m not sure how I managed to write down only half of my thoughts before commenting. Thanks for catching that @Darhazer. What I meant was something along the lines of
```rb
RuboCop::Cop::Layout::ExtraSpacing.singleton_class.prepend(Module.new do
  def autocorrect_incompatible_with
    super.push(RSpec::AlignLeftLetBrace).push(RSpec::AlignRightLetBrace)
  end
end)

RuboCop::Cop::Style::TrailingCommaInArguments.singleton_class.prepend(Module.new do
  def autocorrect_incompatible_with
    super.push(RSpec::Capybara::CurrentPathExpectation)
  end
end)

ydah: Thanks. I'll use module-prepending, as it certainly feels cleaner that way.

ydah:
I updated this PR. Thank you so much!

Darhazer (approved):

Issue #988, opened 2020-08-04
Rubocop RSpec Defaults (Metrics/BlockLength) - Thoughts?


luke-hill:
This in theory is a tiny PR. But first I wanted to get your take on it as contributors.

In Rubocop RSpec, every project I have ever used always adds 2 overrides, or ignores.
```
Metrics/BlockLength:
  Exclude:
    - *.gemspec # For whatever the name of your gem may be
    - spec/**/_spec.rb # nearly all spec files go over 25 lines
```

Should we add new upper limits / ignores for these 2 specific files. I personally think we should and it would save me overriding these in every project I work in.

pirj:
Sure, I agree. Also CanCan abilities. And factories especially if nested `factory`s are used.

Can we override `BlockLength` in our config without breaking it for the rest of the directories?

I'm not sure we should add gemspec inclusion here, but pretty sure you didn't mean it, just added it as another example.

luke-hill:
I actually did sorry. Most gemspecs are over 25 lines long. When you factor in all the options that are configured.

For each example of me creating a gem I have this override in. Also to my knowledge having an override which doesn't exist doesn't crash the system.

pirj:
```yaml
Metrics/BlockLength:
  Exclude:
    - *.gemspec
```
this is better be discussed in https://github.com/rubocop-hq/rubocop/issues then.

Otherwise, a pull request is appreciated.
So, the scenario is:
1. A repo locally excludes `BlockLength` for `app/abilities`
2. `rubocop` gem excludes `*.gemspec` in its `config/default.yml`
3. `rubocop-rspec` excludes `spec/**/_spec.rb` in its `config/default.yml`
it all works together and excludes all of that.

Chances are this may work by default, and in this case, it's enough to send two pulls, one -0 +1 to `rubocop`, another -0 +4 here, and fix it once and for all.
If it doesn't - I encourage you to look at [`inherit_mode`](https://github.com/rubocop-hq/rubocop/blob/5bcf201535d36a417c703555866c33aedf6eadb9/docs/modules/ROOT/pages/configuration.adoc#merging-arrays-using-inherit_mode).

luke-hill:
I missed this response. Sorry. I'll actually look into doing this soon if you think it's a valid inclusion.

I agree the gemspec one should go to rubocop proper. And the rubocop-rspec should have exclude for Metrics/BlockLength.

Sorry for missing this all this time.

luke-hill:
[...]

PR #1256, opened 2022-03-29
Exclude all spec blocks from Metrics/BlockLength

luke-hill:
This was something that was originally discussed in https://github.com/rubocop/rubocop-rspec/issues/988

I'll probably need some more support from @pirj to just confirm I've done everything right (I think I may have missed some things)
It's likely some other maintainers may want to discuss this (As I couldn't spot any other examples of non-rspec cops in the default.yml

---

* [-] Added tests.
* [-] Updated documentation.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 * Drop Ruby 2.5 support. ([@ydah][])
 * Add new `RSpec/ChangeByZero` cop. ([@ydah][])
 * Improve `RSpec/ExpectChange` to detect namespaced and top-level constants. ([@M-Yamashita01][])
+* Introduce an amendment to `Metrics/BlockLength` to exclude spec files. ([@luke-hill][])
 
 ## 2.10.0 (2022-04-19)
 
@@ -686,3 +687,4 @@ Compatibility release so users can upgrade RuboCop to 0.51.0. No new features.
 [@ydah]: https://github.com/ydah
 [@t3h2mas]: https://github.com/t3h2mas
 [@M-Yamashita01]: https://github.com/M-Yamashita01
+[@luke-hill]: https://github.com/luke-hill
--- a/config/default.yml
+++ b/config/default.yml
@@ -110,6 +110,14 @@ RSpec:
       - subject
       - subject!
 
+Metrics/BlockLength:
+  inherit_mode:
+    merge:
+      - Exclude
+  Exclude:
+    - "**/*_spec.rb"
+    - "**/spec/**/*"
+
 RSpec/AlignLeftLetBrace:
   Description: Checks that left braces for adjacent single line lets are aligned.
   Enabled: false
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -51,7 +51,6 @@ Metrics/BlockLength:
     - rubocop-rspec.gemspec
     - Rakefile
     - '**/*.rake'
-    - spec/**/*.rb
 
 Naming/FileName:
   Exclude:
--- a/lib/rubocop/rspec/config_formatter.rb
+++ b/lib/rubocop/rspec/config_formatter.rb
@@ -8,6 +8,7 @@ module RuboCop
     class ConfigFormatter
       EXTENSION_ROOT_DEPARTMENT = %r{^(RSpec/)}.freeze
       SUBDEPARTMENTS = %(RSpec/Capybara RSpec/FactoryBot RSpec/Rails)
+      AMENDMENTS = %(Metrics/BlockLength)
       COP_DOC_BASE_URL = 'https://www.rubydoc.info/gems/rubocop-rspec/RuboCop/Cop/RSpec/'
 
       def initialize(config, descriptions)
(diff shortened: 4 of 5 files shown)


config/default.yml
```diff
@@ -110,6 +110,12 @@ RSpec:
       - subject
       - subject!
 
+Metrics/BlockLength:
+  Enabled: true
+  VersionAdded: 'TBD'
+  Exclude:
+    - 'spec/**/_spec.rb'
```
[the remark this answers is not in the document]
luke-hill: 🙈

config/default.yml
```diff
@@ -114,7 +114,7 @@ Metrics/BlockLength:
   Enabled: true
   VersionAdded: 'TBD'
   Exclude:
-    - 'spec/**/_spec.rb'
+    - 'spec/**/*_spec.rb'
```
pirj, line 117: Maybe follow the same naming scheme as a common `Include` one above?
```
- "**/*_spec.rb"
- "**/spec/**/*"
```

This has one additional benefit - shared examples/contexts (e.g. `spec/support/shared_examples/it_behaves_like_everyone.rb`) would be excluded, too, even though they typically don't end with `_spec.rb`.

config/default.yml
```diff
@@ -114,7 +114,7 @@ Metrics/BlockLength:
   Enabled: true
   VersionAdded: 'TBD'
```
pirj, line 115: We can use `'<<next>>'` like [`rubocop-rails` does](https://github.com/rubocop/rubocop-rails/commit/0e06a0ee424b3ca7f7f83b25d9335116e6d376ec#diff-3653213e3a153018a75cc7b2536b1d0217a6d8d75188d71cf45441812f789698R881)?

config/default.yml
```diff
@@ -110,6 +110,12 @@ RSpec:
       - subject
       - subject!
 
+Metrics/BlockLength:
+  Enabled: true
```
pirj, line 114: I don't think we need to explicitly enable it. Even though, probably it's disabled in user project's `.rubocop.yml` and this would override this, I can think of cases when common rules are extracted to a gem and it's `require:`'d in `.rubocop.yml` before or after `rubocop-rspec`, and this may inadvertently enable the disabled cop.

config/default.yml
pirj: Do we even need this? 🤔

config/default.yml
```diff
@@ -110,6 +110,12 @@ RSpec:
       - subject
       - subject!
 
+Metrics/BlockLength:
```
pirj, line 113: Unfortunately enough, this doesn't seem to work. I've checked by using this branch on a project and removing the explicit `Exclude` for specs from its `.rubocop.yml`.

This seems to work like this for [`rubocop-rails`'s `Style/AndOr` override](https://github.com/rubocop/rubocop-rails/blob/944394b9a45c82137cd9af8852ef43b3cd472265/config/default.yml#L981).

I can think of [`inherit_mode`](https://docs.rubocop.org/rubocop/1.27/configuration.html#merging-arrays-using-inherit_mode), and that the user project's configuration `Exclude` overrides the one from `config/default.yml`.

pirj:
> As I couldn't spot any other examples of non-rspec cops in the default.yml

[`rubocop-rails` does a similar thing](https://github.com/rubocop/rubocop-rails/blob/944394b9a45c82137cd9af8852ef43b3cd472265/config/default.yml#L981).

config/default.yml
```diff
@@ -110,6 +110,14 @@ RSpec:
...
+  inherit_mode:
```
pirj, line 114: `rubocop-rails` goes as far as [setting `Exclude` to `merge` globally](https://github.com/rubocop/rubocop-rails/blob/944394b9a45c82137cd9af8852ef43b3cd472265/config/default.yml#L3).

luke-hill:
Hi there. I've not been monitoring this as I've been off for a while.

Thanks for jumping on this. Evidently it was a bit bigger haha :D

pirj (approved):

CHANGELOG.md
```diff
@@ -8,6 +8,7 @@
 * Add `RSpec/VerifiedDoubleReference` cop. ([@t3h2mas][])
 * Make `RSpec/BeNil` cop configurable with a `be_nil` style and a `be` style. ([@bquorning][])
 * Fix `Capybara/CurrentPathExpectation` autocorrect incompatible with `Style/TrailingCommaInArguments` autocorrect. ([@ydah][])
+* Introduce an amendment to `Metrics/BlockLength` to exclude spec files. ([@luke-hill][])
```
luke-hill, line 11: Should this be in master not 2.10?

pirj: Bad rebase. Can you please fix it?

luke-hill:
This I believe is now rtm @pirj - Thanks for all your help fixing up this!

pirj:
The build is green after merging in a fix for a couple of internal affairs offences.
@bquorning @Darhazer I believe this small change is such a relief for our users who previously had to exclude `/spec/*` from `Metrics/BlockLength` in their `.rubocop.yml` on every new project.

Darhazer (approved):

bquorning (approved):

pirj:
@luke-hill Thanks!

Issue #249, opened 2016-12-07
Cop idea: change example length from counting lines to statements


jcoyne:
Example:
```ruby
    it "spawns CreateWorkJobs for each work" do
      expect(CreateWorkJob).to receive(:perform_later).with(user,
                                                            "GenericWork",
                                                            {
                                                              keyword: [],
                                                              title: ['File One'],
                                                              resource_type: ["Article"],
                                                              uploaded_files: ['1']
                                                            },
                                                            child_log).and_return(true)
      expect(CreateWorkJob).to receive(:perform_later).with(user,
                                                            "GenericWork",
                                                            {
                                                              keyword: [],
                                                              title: ['File Two'],
                                                              resource_type: ["Image"],
                                                              uploaded_files: ['2']
                                                            },
                                                            child_log).and_return(true)
      subject
    end
```

This has 3 statements, which seems like a reasonable number for an example. But due to the way it's formatted (which I believe contributes to good readability *) it has 17 lines, which causes the cop to fail. Would it be possible to count the number of statements instead of lines?

\* Kevlin Henney - Seven Ineffective Coding Habits of Many Programmers https://vimeo.com/97329157#t=14:36

backus:
I think line length and statement complexity are probably both decent heuristics with edge cases that will be obviously wrong in some cases. What do you think abuot a separate cop instead like `RSpec/ExampleComplexity` that might do something like this?

jfelchner:
Now that I'm looking at this I'm wondering if just using rubocop's built-in `PerceivedComplexity`, `AbcSize`, `CyclomaticComplexity` cops might be good enough here.

jfelchner:
FWIW, rubocop's line number cops now allow exceptions for hashes, arrays, and heredocs.  Perhaps it's abstracted enough that this cop could take advantage of it as well. See [here](https://github.com/rubocop-hq/rubocop/blob/master/lib/rubocop/cop/mixin/code_length.rb#L54)

Darhazer:
Since ExampleLength supports the CountAsOne option for arrays and hashes, perhaps we can close this?

pirj:
Makes sense 👍 @bquorning ?

bquorning:
[...]

PR #1257, opened 2022-03-30
Count heredocs as one line in ExampleLength

bquorning:
This changes only the *internal* RuboCop configuration. End-users will see no change.

Inspired by #249.

---

* [ ] Added tests.
* [ ] Updated documentation.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -60,7 +60,9 @@ Naming/FileName:
     - lib/rubocop-rspec.rb
 
 RSpec/ExampleLength:
-  Max: 30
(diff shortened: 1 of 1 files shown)


pirj (approved):
Thanks!
The format is a bit strange, I was thinking it to be a list like:
```yaml
CountAsOne:
 - heredoc
```
but docs/specs point otherwise.

bquorning:
Yeah, yaml has two different syntaxes for lists. I’ve changed to use the one you suggested @pirj, which we also use almost everywhere else.

Issue #1100, opened 2020-11-26
Cop idea: prefer not_to change over change.by(0)


Darhazer:
`change.by(0)` indicates a no change. We can create a cop to enforce the negation.
```
# bad 
expect { action }.to change(User, :count).by(0)
expect { action }.to change { User.count }.by(0)
# good 
expect { action }.not_to change(User, :count)
expect { action }.not_to change { User.count }
```

Note that this is not auto-correctable in case there are chained matchers. You need to define a negated matcher first.

januszm:
> You need to define a negated matcher first.

In relation to this comment from 2017 https://github.com/rspec/rspec-expectations/issues/664#issuecomment-307804053 , perhaps it's a high time to include this  `define_negated_matcher :not_change, :change` in RSpec by default with the change merged here.

januszm:
Right, I did not express myself precisely, rather the point is that `rubocop-rspec` could provide such a matcher, since after turning on this Cop by default, in practice, it requires it for chained "change" by 0. But that's another story, I assume this module doesn't and won't add any extensions to RSpec.
My point is, this now requires developers / teams to extend their apps with a custom negated match. This is already something that could come with the gem (whichever is right).

pirj:
> this module doesn't and won't add any extensions to RSpec

Indeed.

> This is already something that could come with the gem (whichever is right).

It's one line in `spec/spec_helper.rb` vs one line in `Gemfile` + one in `Gemfile.lock`.

PR #1265, opened 2022-04-23
Add new `RSpec/ChangeByZero` cop

ydah:
Resolves: #1100

This cop enforces the use of `not_to change` over `to change.by(0)`

## example

```ruby
# bad
expect { run }.to change(Foo, :bar).by(0)
expect { run }.to change { Foo.bar }.by(0)
expect { run }.to change(Foo, :bar).by(0).and change(Foo, :baz).by(0)
expect { run }.to change { Foo.bar }.by(0).and change { Foo.baz }.by(0)

# good
expect { run }.not_to change(Foo, :bar)
expect { run }.not_to change { Foo.bar }
expect { run }.to not_change(Foo, :bar).and not_change(Foo, :baz)
expect { run }.to not_change { Foo.bar }.and not_change { Foo.baz }
```

---

* [x] Added tests.
* [x] Updated documentation.


spec/rubocop/cop/rspec/change_by_zero_spec.rb
```diff
@@ -0,0 +1,69 @@
...
+
+    expect_correction(<<-RUBY)
+      it do
+        expect{ foo }.not_to change(User, :count)
+      end
+    RUBY
+  end
+
+  it 'does not register an offense when using' \
+    '`to change(object, :attribute).by(not 0)`' do
```
pirj, line 20: This code is backtick-quoted, I'd think it means that literally `not 0`.
To be honest, I even checked if `not 0` would run and what it evaluates to 😄 
WDYT of "does not register an offense when the argument to `by` is not zero"?

spec/rubocop/cop/rspec/change_by_zero_spec.rb
```diff
@@ -0,0 +1,69 @@
+# frozen_string_literal: true
+
+RSpec.describe RuboCop::Cop::RSpec::ChangeByZero do
+  it 'registers an offense when using `to change(object, :attribute).by(0)`' do
```
pirj, line 4: This docstring says `object`, but the cop only inspects if it's a literal class.

pirj (commented):
Thank you for another contribution!

A few notes below.
Additionally, I've ran the cop against https://github.com/pirj/real-world-rspec, and it errored out on a couple of files:
```
An error occurred while RSpec/ChangeByZero cop was inspecting /Users/pirj/source/real-world-rspec/canvas-lms/spec/models/content_export_spec.rb:325:30.
undefined method `end_pos' for nil:NilClass
/Users/pirj/source/rubocop-rspec/lib/rubocop/cop/rspec/change_by_zero.rb:67:in `autocorrect'
```

```
expect { @ce.save! }.to change(DelayedMessage, :count).by 0
```

Checked a few cases - seem to be related to `by` with no parentheses.

And this confirms that this weird `by(0)` pattern is really used in the wild 🎉

lib/rubocop/cop/rspec/change_by_zero.rb
```diff
@@ -0,0 +1,73 @@
...
+        def_node_matcher :expect_change_with_arguments, <<-PATTERN
+          (send nil? :change $_ (sym $_))
+        PATTERN
+
+        # @!method expect_change_with_block(node)
+        def_node_matcher :expect_change_with_block, <<-PATTERN
+          (block
+            (send nil? :change)
+            (args)
+            (send ({const send} nil? $_) $_)
```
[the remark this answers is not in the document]
Darhazer: I believe those are copied from `ExpectChange` cop so it would suffer from the same issue

lib/rubocop/cop/rspec/change_by_zero.rb
```diff
@@ -0,0 +1,73 @@
...
+
+          add_offense(node.parent.loc.expression) do |corrector|
+            autocorrect(corrector, node)
+          end
+        end
+
+        def change_by_zero?(node)
+          _, method_name, args = *node.parent
+          method_name == :by && args.int_type? && args.value.zero?
+        end
```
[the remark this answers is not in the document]
ydah: Thank you so much! It looks simpler, and I will be using the one you suggested.

lib/rubocop/cop/rspec/change_by_zero.rb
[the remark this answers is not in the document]
ydah: Yes, It had that issue and will correct it.

lib/rubocop/cop/rspec/change_by_zero.rb
```diff
@@ -0,0 +1,73 @@
...
+
+module RuboCop
+  module Cop
+    module RSpec
+      # Prefer `not_to change` over `to change.by(0)`.
+      #
+      # @example
+      #   # bad
+      #   expect { run }.to change(Foo, :bar).by(0)
+      #   expect { run }.to change { Foo.bar }.by(0)
```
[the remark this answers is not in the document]
ydah: > Presently, the cop registers an offence:
and for some reason only for the last change.

This was a mistake in my code. My apologies.
I've considered the options you suggested:

Compound expectations are no longer subject to autocorrect, but I improve to suggest the use of negated matchers.

spec/rubocop/cop/rspec/change_by_zero_spec.rb
```diff
@@ -0,0 +1,69 @@
...
+    expect_offense(<<-RUBY)
+      it do
+        expect{ foo }.to change(User, :count).by(0)
```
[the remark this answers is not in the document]
ydah: Indeed it is. I'll fix it.

spec/rubocop/cop/rspec/change_by_zero_spec.rb
```diff
@@ -0,0 +1,69 @@
...
+        expect{ foo }.to change(User, :count).by(1)
+      end
+    RUBY
+  end
+
+  it 'does not register an offense when using' \
+    '`to change(object, :attribute).by(0)` with chained matchers' do
+    expect_no_offenses(<<-RUBY)
+      it do
+        expect{ foo }.to change(User, :count).by(0).and_return(1)
```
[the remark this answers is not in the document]
ydah: I thought it was certainly better to delete it, so I will.

spec/rubocop/cop/rspec/change_by_zero_spec.rb
ydah: "does not register an offense when the argument to by is not zero"  looks better. I'll fix it.

pirj (approved):
Just perfect, thank you so much! 🎉 

Tested on `real-world-rspec` with no single hiccup:
```
181 files inspected, 409 offenses detected, 290 offenses corrected
```

lib/rubocop/cop/rspec/change_by_zero.rb
```diff
@@ -0,0 +1,87 @@
...
+          (send
+            (block
+              (send nil? :change)
+              (args)
+              (send (...) $_)) :by
+            (int 0))
+        PATTERN
+
+        def on_send(node)
+          expect_change_with_arguments(node) do
```
Darhazer, line 53: minor: the cop will be faster if we RESTRICT_ON_SEND to `change` and then check the pattern against `node.parent` (as we will be getting the inner send node)

Darhazer (approved):

lib/rubocop/cop/rspec/change_by_zero.rb
bquorning: > the cop will be faster if we RESTRICT_ON_SEND to […]

Can we write a cop to verify the presence of RESTRICT_ON_SEND when `on_send` is defined?

ydah: Thank you so much. I'll define RESTRICT_ON_SEND to `change`.

ydah:
I add RESTRICT_ON_SEND to `change`.  Improved speed for cop. 
I updated this PR! Thank you so much.

bquorning (approved):

lib/rubocop/cop/rspec/change_by_zero.rb
ydah: > Can we write a cop to verify the presence of RESTRICT_ON_SEND when on_send is defined?

I tried to create a PR https://github.com/rubocop/rubocop/pull/10783 but it is now closed.
The reason is that there was a judgment that there may be more noise than detection efficiency.

Issue #740, opened 2019-02-12
Proposal: Avoid repetitive definitions of `CustomTransform` in `RSpec/FilePath`


aeroastro:
## Summary

I would like to achieve the following 2 things for `RSpec/FilePath`.
I am willing to implement the solution when we reach an agreement on this issue.

### Avoid repetitive `CustomTransform` definitions for single inflection.

When using `ActiveSupport::Inflector`, we have to write only 1 definition for 1 word with special inflection. On the other hand, when using `rubocop-rspec`, we have to define multiple definitions (all the composite words derived from a special word). This makes development a little bit less productive.

For example, let's assume that we use a special word like `PvP` in a project.

When using ActiveSupport::Inflector, all we have to do is write a single rule on `PvP`.

```ruby
ActiveSupport::Inflector.inflections(:en) do |inflect|
  inflect.acronym 'PvP'
end
```

When we use the current `CustomTransform`, we have to define all the transforms as follows. 

```yaml
RSpec/FilePath:
  CustomTransform:
    PvP: pvp
    SyncPvP: sync_pvp
    AsyncPvP: async_pvp
    PvPOverPvP: pvp_over_pvp
    PvPsController: pvps_controller
```

This is mainly because the exact-match substitution is handled before snake casing.

https://github.com/rubocop-hq/rubocop-rspec/blob/d69442f17d874cff8f5c11faabd929534c8fc717/lib/rubocop/cop/rspec/file_path.rb#L83-L89

I would like to let the following config just work as `ActiveSupport::Inflector` config does.

```yaml
RSpec/FilePath:
  CustomTransform:
    PvP: pvp
```

### Share the same settings as ActiveSupport::Inflector has

I am not so inclined to achieve this as the first one, but, hopefully, I would like to share the same config as we have in Rails application. This motivation is not applicable to non-Rails user. Therefore, I would like to use less-invasive way to achieve this.

In our project, we are using `require` directive in config YAML to require `config/initializers/inflections.rb`, where `ActiveSupport::Inflector` config exists, but there may be a more sophisticated intuitive way.

## How to solve this issue

IMHO, I believe that `ActiveSupport::Inflector`, which is famous for its sophisticated string handling, can solve this issue. Since this idea has both pros and cons, I have set up this issue.

https://api.rubyonrails.org/classes/ActiveSupport/Inflector.html

[...]

PR #1181, opened 2021-08-12
Bump minimum RuboCop dependency to fix DSL configuration merging

pirj:
fixes #1126

Related fix in RuboCop https://github.com/rubocop/rubocop/pull/9952, [released in `rubocop` 1.19.0](https://github.com/rubocop/rubocop/releases/tag/v1.19.0).

---

Before submitting the PR make sure the following are checked:

* [x] Feature branch is up-to-date with `master` (if not - rebase it).
* [x] Squashed related commits together.
* [-] Added tests.
* [-] Updated documentation.
* [x] Added an entry to the `CHANGELOG.md` if the new code introduces user-observable changes.
* [x] The build (`bundle exec rake`) passes (be sure to run this locally, since it may produce updated documentation that you will need to commit).

PR #1266, opened 2022-04-23
Make `RSpec/FilePath` support ActiveSupport inflections, if defined

jeromedalbert:
Fixes https://github.com/rubocop/rubocop-rspec/issues/740. Implementation is based on the suggestion in this comment: https://github.com/rubocop/rubocop-rspec/issues/740#issuecomment-462932812.

This is especially useful for Rails applications (or any application using ActiveSupport inflections) that define mixed-case custom acronyms, for example:

```ruby
# config/initializers/inflections.rb
ActiveSupport::Inflector.inflections do |inflect|
  inflect.acronym('PvP')
end

# app/models/pvp_class.rb instead of app/models/pv_p_class.rb, thanks to the custom acronym
class PvPClass
  ...
end
```

It is also great to be re-using ActiveSupport's inflection configuration instead of having to repeat it in Rubocop as mentioned in https://github.com/rubocop/rubocop-rspec/issues/740.

Other common custom mixed-case acronyms could be StatsD, OAuth, RESTful, etc.

---

* [x] Added tests.
* [ ] Updated documentation.

### DIFF
--- a/lib/rubocop/cop/rspec/file_path.rb
+++ b/lib/rubocop/cop/rspec/file_path.rb
@@ -130,6 +130,13 @@ module RuboCop
         end
 
         def camel_to_snake_case(string)
+          if defined?(ActiveSupport::Inflector)
+            if File.exist?('./config/initializers/inflections.rb')
+              require './config/initializers/inflections'
+            end
+            return ActiveSupport::Inflector.underscore(string)
+          end
+
           string
             .gsub(/([^A-Z])([A-Z]+)/, '\1_\2')
             .gsub(/([A-Z])([A-Z][^A-Z\d]+)/, '\1_\2')
--- a/spec/rubocop/cop/rspec/file_path_spec.rb
+++ b/spec/rubocop/cop/rspec/file_path_spec.rb
@@ -246,6 +246,27 @@ RSpec.describe RuboCop::Cop::RSpec::FilePath do
     end
   end
 
+  context 'when ActiveSupport Inflector is defined', order: :defined do
+    before { require 'active_support/inflector' }
+
+    it 'registers an offense for a bad path when there is no custom acronym' do
+      expect_offense(<<-RUBY, 'pvp_class_foo_spec.rb')
+        describe PvPClass, 'foo' do; end
+        ^^^^^^^^^^^^^^^^^^^^^^^^ Spec path should end with `pv_p_class*foo*_spec.rb`.
+      RUBY
+    end
+
+    it 'does not register an offense when class name contains custom acronym' do
+      ActiveSupport::Inflector.inflections do |inflect|
+        inflect.acronym('PvP')
+      end
+
+      expect_no_offenses(<<-RUBY, 'pvp_class_foo_spec.rb')
+        describe PvPClass, 'foo' do; end
+      RUBY
+    end
+  end
+
   context 'when configured with IgnoreMethods' do
     let(:cop_config) { { 'IgnoreMethods' => true } }
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 * Drop Ruby 2.5 support. ([@ydah][])
 * Add new `RSpec/ChangeByZero` cop. ([@ydah][])
+* Make `RSpec/FilePath` support ActiveSupport inflections, if defined. ([@jeromedalbert][])
 
 ## 2.10.0 (2022-04-19)
 
@@ -684,3 +685,4 @@ Compatibility release so users can upgrade RuboCop to 0.51.0. No new features.
 [@oshiro3]: https://github.com/oshiro3
 [@ydah]: https://github.com/ydah
 [@t3h2mas]: https://github.com/t3h2mas
+[@jeromedalbert]: https://github.com/jeromedalbert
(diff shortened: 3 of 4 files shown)


lib/rubocop/cop/rspec/file_path.rb
```diff
@@ -130,6 +130,13 @@ def expected_path(constant)
         end
 
         def camel_to_snake_case(string)
+          if defined?(ActiveSupport::Inflector)
+            if File.exist?('./config/initializers/inflections.rb')
+              require './config/initializers/inflections'
+            end
```
jeromedalbert, line 136: Inflection configuration must be loaded before `ActiveSupport::Inflector.underscore` to take acronyms and other inflection rules into account.

That configuration is typically found in the `config/initializers/inflections.rb` file that is generated by default in new Rails apps.

pirj:
Nice!
Do you mind throwing in a spec that would indicate how it behaves in action?

pirj (approved):
Thank you!

pirj:
The failing example is flaky, due to order dependency. If the example that define the "PvP" acronym runs first, that the other one fails.
I suggest adding metadata to the example to work this around:
```ruby
context 'when ActiveSupport Inflector is defined', order: :defined do
```

lib/rubocop/cop/rspec/file_path.rb
bquorning, line 136: I’m afraid this code is a bit slow. Could we do the `defined?`/`File.exists?`/`require` thing just _once_ and store the result perhaps a class variable?

lib/rubocop/cop/rspec/file_path.rb
```diff
@@ -130,6 +130,13 @@ def expected_path(constant)
         end
 
         def camel_to_snake_case(string)
+          if defined?(ActiveSupport::Inflector)
+            if File.exist?('./config/initializers/inflections.rb')
+              require './config/initializers/inflections'
```
bquorning, line 135: It feels like this file path should be configurable.

spec/rubocop/cop/rspec/file_path_spec.rb
```diff
@@ -246,6 +246,27 @@ class Foo
...
+      expect_offense(<<-RUBY, 'pvp_class_foo_spec.rb')
+        describe PvPClass, 'foo' do; end
+        ^^^^^^^^^^^^^^^^^^^^^^^^ Spec path should end with `pv_p_class*foo*_spec.rb`.
+      RUBY
+    end
+
+    it 'does not register an offense when class name contains custom acronym' do
+      ActiveSupport::Inflector.inflections do |inflect|
+        inflect.acronym('PvP')
+      end
```
bquorning, line 262: I wonder if we couldn’t do
```rb
ensure
  ActiveSupport::Inflector.inflections.clear(:acronyms)
```
instead of having the `order :defined` on the context block.

bquorning (commented):
I am a bit on the fence about adding Rails specific logic into RuboCop-RSpec, so not giving a 👍🏼 right away. I’ll think about it some more.

jeromedalbert:
@bquorning 

> I’ll think about it some more.

Any update on your thoughts? I should probably not address the review comments if this PR gets closed for ideological reasons.

bquorning:
It would be neat if we could configure this cop with an “inflector block”, i.e. `->(string) { ActiveSupport::Inflector.underscore(string) }`, but I am not sure we can do that via YAML. @pirj what do you think?

pirj:
@bquorning I was a secret admirer/dreamer of a plain Ruby RuboCop configuration, especially after our [DSL configuration extension](https://github.com/rubocop/rubocop-rspec/pull/956) and some [later problems](https://github.com/rubocop/rubocop-rspec/pull/1181) with [merging arrays in YAML](https://github.com/rubocop/rubocop/issues/9325).

Did you have something like:
```ruby
RuboCop::Cop::RSpec::FilePath.add_custom_transform  { |string| ActiveSupport::Inflector.underscore(string) }
```
in mind? I'm all for it. :smiling_imp: 

cc @Darhazer

lib/rubocop/cop/rspec/file_path.rb
```diff
@@ -130,6 +130,13 @@ def expected_path(constant)
         end
 
         def camel_to_snake_case(string)
+          if defined?(ActiveSupport::Inflector)
```
cdhagmann, line 133: There are other libraries that can be used to set inflectors: 
 - [dry-inflector](https://dry-rb.org/gems/dry-inflector/0.1/)
 - [zeitwerk](https://github.com/fxn/zeitwerk/blob/main/lib/zeitwerk/inflector.rb)

I would suggest providing a script field in the YAML files that could be run such that `String#underscore` could be defined if desired. 

```yaml
Rspec/FilePath:
  LoadInfector: |-
    require 'active_support/inflector'
    require './config/initializers/inflections'
```

Alternatively, we could add a field for the Inflector gem and/or definition file:

```yaml
Rspec/FilePath:
  UseInflectorGem: ActiveSupport
  InflectorDefinition: ./config/initializers/inflections.rb
```

Doing either of these should then allow `if string.respond_to?(:underscore)` to be used instead of hard-coding a library and a file location.

cdhagmann: Another alternative may simple be to make sure the file name matches the locations in which the class is defined. (e.g. if MyHTMLModule::XMLReader is defined in `app/services/my_html_module/xml_reader.rb` that this cop expects it spec to be in `spec/services/my_html_module/xml_reader_spec.rb` regardless of inflection rules.

jeromedalbert:
This PR is stale, closing it.

PR #1267, opened 2022-04-26
Drop Ruby 2.5 support

ydah:
Follow up https://github.com/rubocop/rubocop/pull/10577

---

~* [ ] Added tests.~
~* [ ] Updated documentation.~

~* [ ] Added the new cop to `config/default.yml`.~
~* [ ] The cop is configured as `Enabled: pending~` in `config/default.yml`.~
~* [ ] The cop documents examples of good and bad code.~
~* [ ] The tests assert both that bad code is reported and that good code is not reported.~
~* [ ] Set `VersionAdded` in `default/config.yml` to the next minor version.~

~* [ ] Set `VersionChanged` in `config/default.yml` to the next major version.~

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Master (Unreleased)
 
+* Drop Ruby 2.5 support. ([@ydah][])
+
 ## 2.10.0 (2022-04-19)
 
 * Fix a false positive for `RSpec/EmptyExampleGroup` when expectations in case statement. ([@ydah][])
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -23,7 +23,6 @@ jobs:
     strategy:
       matrix:
         ruby:
-          - 2.5
           - 2.6
           - 2.7
           - "3.0"
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -6,7 +6,7 @@ require:
 
 AllCops:
   DisplayCopNames: true
-  TargetRubyVersion: 2.5
+  TargetRubyVersion: 2.6
   NewCops: disable
   Exclude:
     - 'vendor/**/*'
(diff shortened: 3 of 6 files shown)


ydah:
At least the problem I am encountering is that the `TargetRubyVersion` in `rubocop.yml` points to 2.5, so it looks like CI is not passing.

https://github.com/rubocop/rubocop-rspec/runs/6180012648?check_suite_focus=true

> Error: RuboCop found unsupported Ruby version 2.5 in `TargetRubyVersion` parameter (in .rubocop.yml). 2.5-compatible analysis was dropped after version 1.28.

CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 ## Master (Unreleased)
 
+* **(Breaking)** Drop Ruby 2.5 support. ([@ydah][])
```
bquorning, line 5: I wouldn’t call it a breaking change (which would force us to do a major version release). Because of the `required_ruby_version` in our gemspec, those of our users that are still using Ruby 2.5 are simply unable to install the new version – and nothing will break for them.

bquorning (changes requested):

CHANGELOG.md
ydah: I thought that was indeed true, so I deleted it. Thank you so much.

bquorning (approved):
When I saw this PR I thought “we only _just_ removed support for Ruby 2.4”. But no, that was almost exactly [one year ago](https://github.com/rubocop/rubocop-rspec/pull/1140). 😄 

Thank you for the PR @ydah 🙏🏼

bquorning:
I have removed the two Ruby 2.5-related branch protection rules
- internal_investigation, Ruby 2.5 (gem)
- spec, Ruby 2.5 (gem)

PR #1277, opened 2022-05-18
Fix regression in ExpectChange

### LINKED ISSUE #1276: ExpectChange in 2.11.0 makes problematic corrections
```ruby
expect { a.action }.to change { model.reload.attribute }.from(1).to(0) 
```

gets corrected to

```ruby
expect { a.action }.to change(model.reload, :attribute).from(1).to(0)
```

which is not equivalent (because reload is only called once instead of twice) and can cause the spec to fail.

A second issue is `An error occurred while RSpec/ExpectChange cop was inspecting` occurs when you have something like this:

```ruby
expect { described_class.action }.not_to change { model }
```

The issue here appears to be that the `change { model }` block has no method chain and it can't parse it? This actually allowed us to discover a spec that was testing incorrectly so I don't know if there's a legitimate use of the matcher in this way but either way I'm guessing it should at least be parseable.

pirj:
The cop was triggered by `change { user.reload.name }`

fixes #1276

---

* [x] Added tests.
* [-] Updated documentation.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Master (Unreleased)
 
+* Fix a regression in `RSpec/ExpectChange` flagging chained method calls. ([@pirj][])
+
 ## 2.11.0 (2022-05-18)
 
 * Drop Ruby 2.5 support. ([@ydah][])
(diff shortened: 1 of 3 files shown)


bquorning (approved):

pirj:
@bquorning Should we do a patch release?

atanbt:
@pirj yes I just confirmed that branch fixes the issue

Issue #984, opened 2020-08-02
Refactor `RSpec/DescribedClass` to use `parent_module_name` from `rubocop-ast`


pirj:
There is [`parent_module_name` in `rubocop-ast`](https://github.com/rubocop-hq/rubocop-ast/blob/e2df776f6aa4133413e2de0e20b4a3295b12988f/lib/rubocop/ast/node.rb#L333), which seems to do the same as the `namespace` method from `RSpec/DescribedClass`.

**WARNING**: `parent_module_name` is used only by the `DuplicateMethods` cop and seems to be buggy (or not quite doing the same that we want on purpose).

Originally https://github.com/rubocop-hq/rubocop-rspec/pull/982/files#r464095483

PR #1287, opened 2022-05-26
Add test to ensure wrong namespace is caught

### LINKED ISSUE #1107: RSpec/FilePath misses a mismatched namespace
The cop is missing a mismatched namespace:

```ruby
# spec/rspec/expectations/configuration_spec.rb
module RSpec
  module Mocks # Should be Expectations
    RSpec.describe Configuration do
```
#1099

tejasbubane:
As the test is passing, I think the issue does not exist anymore so we can close #1107 

---

* [x] Added tests.
* [ ] Updated documentation.

### DIFF
--- a/spec/rubocop/cop/rspec/file_path_spec.rb
+++ b/spec/rubocop/cop/rspec/file_path_spec.rb
@@ -206,6 +206,13 @@ RSpec.describe RuboCop::Cop::RSpec::FilePath do
     RUBY
   end
 
+  it 'registers an offense for path with incorrect namespace' do
+    expect_offense(<<-RUBY, '/home/foo/spec/very/long/my_class_spec.rb')
+      describe Very::Medium::MyClass do; end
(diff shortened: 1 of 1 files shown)


pirj (commented):
This example is nice to have, but the problem in #1107 is that the following example doesn't pass:
```ruby
  it 'registers an offense for path with incorrect namespace' do
    expect_offense(<<-RUBY, '/home/foo/spec/very/long/my_class_spec.rb')
      module Very
        module Medium
          describe MyClass do; end
          ^^^^^^^^^^^^^^^^ Spec path should end with `very/medium/my_class*_spec.rb`.
        end
      end
    RUBY
  end
```

Would you like to hack on it?

pirj:
There's a `namespace` helper in `lib/rubocop/cop/rspec/described_class.rb` that may come handy.
#984 is a related ticket to make use of a similar helper from `rubocop-ast`.

pirj:
Ping @tejasbubane

pirj:
Superseded by #1323

PR #1289, opened 2022-05-30
Move enforced styles to top in docs

pirj:
Coming from https://github.com/rubocop/rubocop-rspec/pull/1279#discussion_r884678237

> but should we always have the default on top, or should we have them alphabeticall? 

This PR suggests default on top.

---

* [-] Added tests.
* [x] Updated documentation.

### DIFF
--- a/docs/modules/ROOT/pages/cops_rspec.adoc
+++ b/docs/modules/ROOT/pages/cops_rspec.adoc
@@ -1528,17 +1528,6 @@ This cop can be configured using the `EnforcedStyle` option.
 
 === Examples
 
-==== `EnforcedStyle: block`
-
-[source,ruby]
-----
-# bad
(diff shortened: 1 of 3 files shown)


Darhazer (approved):
Thank you. It's nice that only 2 cops were affected :)

bquorning (approved):
We have no way of enforcing this automatically, do we?

pirj:
> enforcing this automatically

Nothing comes to mind. We don't have checks on generated docs.
Originally, it's coming from the cops' sources, but from comments, so a custom cop wouldn't probably cut it.

Darhazer:
> We have no way of enforcing this automatically, do we?

Not yet. From the same #1279 I also suggested to have a way to enforce the presence of the doc for each supported style.

PR #1291, opened 2022-06-01
Fix ImplicitSubject within third-party aliases

### LINKED ISSUE #1290: RSpec/ImplicitSubject does not use example methods defined by third-party gems
I use saharspecs and its `its_call` alias :

```
its_call(1) do
  is_expected
    .to  send_message(Foo, :method).once
    .and change { whatever }
end
```

I've add saharspec aliases to rubocop and  RSpec/ImplicitSubject is set to accept single statement

```
inherit_gem:
  saharspec: config/rubocop-rspec.yml

[...]

RSpec/ImplicitSubject:
  EnforcedStyle: single_statement_only
```

It report offenses on Implicit subject:

```
RSpec/ImplicitSubject: Don't use implicit subject. (https://www.rubydoc.info/gems/rubocop-rspec/RuboCop/Cop/RSpec/ImplicitSubject)
    is_expected
    ^^^^^^^^^^^
```

I've got the same error when I set RSpec/ImplicitSubject to accept single line

```
its_call(1) { is_expected.to send_message(Foo, :method).once }
```
```
RSpec/ImplicitSubject:
  EnforcedStyle: single_line_only
```
```
RSpec/ImplicitSubject: Don't use implicit subject. (https://www.rubydoc.info/gems/rubocop-rspec/RuboCop/Cop/RSpec/ImplicitSubject)
  its_call(1) { is_expected.to send_message(Foo, :method).once }
                ^^^^^^^^^^^
```

inkstak:
Fix #1290

---

* [x] Added tests.
* [ ] Updated documentation.

### DIFF
--- a/spec/rubocop/cop/rspec/implicit_subject_spec.rb
+++ b/spec/rubocop/cop/rspec/implicit_subject_spec.rb
@@ -85,6 +85,15 @@ RSpec.describe RuboCop::Cop::RSpec::ImplicitSubject do
         end
       RUBY
     end
+
+    it 'allows `is_expected` with an alias' do
+      other_cops.dig('RSpec', 'Language', 'Includes', 'Examples')
+        .push('epic')
+
+      expect_no_offenses(<<-RUBY)
+        epic { is_expected.to be_good }
+      RUBY
+    end
   end
 
   context 'with EnforcedStyle `single_statement_only`' do
@@ -93,6 +102,17 @@ RSpec.describe RuboCop::Cop::RSpec::ImplicitSubject do
     it 'allows `is_expected` in multi-line example with single statement' do
       expect_no_offenses(<<-RUBY)
         it 'expect subject to be used' do
+          is_expected.to be_good
+        end
+      RUBY
+    end
+
+    it 'allows `is_expected` in single statement with an alias' do
+      other_cops.dig('RSpec', 'Language', 'Includes', 'Examples')
+        .push('epic')
+
+      expect_no_offenses(<<-RUBY)
+        epic do
           is_expected.to be_good
         end
       RUBY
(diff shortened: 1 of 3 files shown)


spec/rubocop/cop/rspec/implicit_subject_spec.rb
```diff
@@ -98,6 +107,17 @@ def permits(actions)
       RUBY
     end
 
+    it 'allows `is_expected` in single statement with an alias' do
```
pirj, line 110: The root problem was that Includes.examples were not considered, so maybe it makes sense to specify this, e.g. "with an alias to Includes.examples".

On a different thought, `its_call` doesn't seems to be a very fair alias to e.g. `it_behaves_like`, as the latter creates an example group.
Perhaps it would make sense to change aliasing in `saharspec` to `RSpec/Language/Examples/Regular` [just like we have it for `its`](https://github.com/rubocop/rubocop-rspec/blob/bb046ebb8ede99ea9273240918b6b22a516a3dbe/config/default.yml#L48).
This would also make this PR redundant, and would also 

See https://github.com/palkan/action_policy/pull/138 for more, but there is some specifics to how e.g. `success` is used there, without a block to justify adding it as an alias to Include.examples.

What do you think?

ka8725: @pirj do you ask if the following config is equal to the one that I made in the PR to saharspec?

```yml
RSpec:
  Language:
    Examples:
      Regular:
        - its_block
        - its_call
        - its_map
      Skipped:
        - xits_block
        - xits_call
        - xits_map
      Focused:
        - fits_block
        - fits_call
        - fits_map
```

I think so, I checked this config and it works well for all my scenarios.

pirj:
I'm closing this, as this can be fixed on `saharspec` side. Thanks everyone involved!
@inkstak Please accept my apologies for the initially wrong suggestion that sank some of your time.

PR #1293, opened 2022-06-03
Better explanation for MessageSpies

kuboon:
Current example simply switches `receive` and `have_received` but they are not same function.

### DIFF
--- a/lib/rubocop/cop/rspec/message_spies.rb
+++ b/lib/rubocop/cop/rspec/message_spies.rb
@@ -12,22 +12,23 @@ module RuboCop
       #
       #   # bad
       #   expect(foo).to receive(:bar)
-      #   subject
+      #   do_something
       #
       #   # good
       #   allow(foo).to receive(:bar) # or use instance_spy
-      #   subject
+      #   do_something
       #   expect(foo).to have_received(:bar)
       #
       # @example `EnforcedStyle: receive`
       #
       #   # bad
-      #   subject
+      #   allow(foo).to receive(:bar)
+      #   do_something
       #   expect(foo).to have_received(:bar)
       #
       #   # good
       #   expect(foo).to receive(:bar)
-      #   subject
+      #   do_something
       #
       class MessageSpies < Base
         include ConfigurableEnforcedStyle
--- a/lib/rubocop/cop/rspec/message_spies.rb
+++ b/lib/rubocop/cop/rspec/message_spies.rb
@@ -12,17 +12,22 @@ module RuboCop
       #
       #   # bad
       #   expect(foo).to receive(:bar)
+      #   subject
       #
       #   # good
+      #   allow(foo).to receive(:bar) # or use instance_spy
+      #   subject
       #   expect(foo).to have_received(:bar)
       #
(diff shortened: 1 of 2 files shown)


lib/rubocop/cop/rspec/message_spies.rb
```diff
@@ -12,17 +12,22 @@ module RSpec
       #
       #   # bad
       #   expect(foo).to receive(:bar)
+      #   subject
```
pirj, line 15: I'm not convinced that the use of `subject` is the best example.
Harvesting the tickets related to the cop, I've seen an example with `do_something`.

lib/rubocop/cop/rspec/message_spies.rb
```diff
@@ -12,17 +12,22 @@ module RSpec
...
       #
       #   # good
+      #   allow(foo).to receive(:bar) # or use instance_spy
```
pirj, line 18: This addition makes sense 👍

lib/rubocop/cop/rspec/message_spies.rb
```diff
@@ -12,17 +12,22 @@ module RSpec
...
+      #   subject
       #   expect(foo).to have_received(:bar)
       #
       # @example `EnforcedStyle: receive`
       #
       #   # bad
+      #   subject
```
pirj, line 25: Does it make sense to add an `allow` here as well?

lib/rubocop/cop/rspec/message_spies.rb
```diff
@@ -12,17 +12,22 @@ module RSpec
...
       #   # bad
       #   expect(foo).to receive(:bar)
+      #   subject
       #
       #   # good
+      #   allow(foo).to receive(:bar) # or use instance_spy
+      #   subject
       #   expect(foo).to have_received(:bar)
       #
       # @example `EnforcedStyle: receive`
```
pirj, line 22: Side note: I've run the cop against real-world-rspec with both settings:
`EnforcedStyle: receive`: 57497 files inspected, 1881 offences detected
`EnforcedStyle: have_received`: 57497 files inspected, 26290 offenses detected
That leads me to a thought that `to receive` is more popular, despite the cop being enabled by default and the default enforced style.

I would rather not shove the style that is my personal preference.
@bquorning, a question to you as a supporter of the current enforced style. Do you think it is a sufficient reason to change the default enforced style, or it's fine to keep the existing state of affairs?

kuboon:
Thanks for reviews.
I'll fix them later.

lib/rubocop/cop/rspec/message_spies.rb
bquorning: To recap, my preference for spies (`to_receive`) came after reading about [4-phase testing](https://thoughtbot.com/blog/four-phase-test), and after realizing that `expect … to_receive … and_return` is [confusing](https://github.com/rubocop/rubocop-rspec/issues/221).

I realize by the number of tickets that it’s a controversial style, and I don’t want to force controversial on people. But I do fear that if disabled, most people will never discover the cop.

pirj:
Thank you!

Issue #1248, opened 2022-03-08
Cop idea: use more specific matchers in Capybara


boris-petrov:
It would be nice to have a cop that warns for usages like:

```rb
expect(page).to have_css('button', exact_text: 'a')
expect(page).to have_css('button[disabled]', exact_text: 'b')
expect(page).to have_css('button:not([disabled])', exact_text: 'c')

expect(page).to have_css('a', exact_text: 'd')
expect(page).to have_css('a.cls', exact_text: 'e')

expect(page).to have_css('table')
expect(page).to have_css('table.cls')
```

And suggest replacement with the following:

```rb
expect(page).to have_button('a')
expect(page).to have_button('b', disabled: true)
expect(page).to have_button('c')

expect(page).to have_link('d')
expect(page).to have_link('e', class: 'cls')

expect(page).to have_table
expect(page).to have_table(class: 'cls')
```

Of course, there are many cases that can't be changed like that (for example when generic attributes are checked like `button[data-foo=1]`) but at least "obvious" code could be checked.

pirj:
Would you like to contribute such a cop, @boris-petrov ?

boris-petrov:
@pirj - I doubt I'll have the time to do it any time soon. :(

ydah:
If don't mind, I'll try contribute on this cop.

pirj:
That would be fantastic, @ydah !

ydah:
Sorry for the delay... PR is now open!

PR #1294, opened 2022-06-05
Add new `RSpec/Capybara/SpecificMatcher` cop

ydah:
Resolves: https://github.com/rubocop/rubocop-rspec/issues/1248

This cop checks for there is a more specific matcher offered by Capybara.

### @example
```ruby
# bad
expect(page).to have_css('button')
expect(page).to have_css('button', exact_text: 'foo')
expect(page).to have_css('button[disabled]', exact_text: 'foo')
expect(page).to have_css('a')
expect(page).to have_css('a.cls', exact_text: 'foo')
expect(page).to have_css('table')
expect(page).to have_css('table.cls')

# good
expect(page).to have_button
expect(page).to have_button('foo')
expect(page).to have_button('foo', disabled: true)
expect(page).to have_link
expect(page).to have_link('foo', class: 'cls')
expect(page).to have_table
expect(page).to have_table(class: 'cls')
```

---

* [x] Added tests.
* [x] Updated documentation.

### DIFF
--- a/docs/modules/ROOT/pages/cops_rspec_capybara.adoc
+++ b/docs/modules/ROOT/pages/cops_rspec_capybara.adoc
@@ -106,6 +106,45 @@ end
 
 * https://www.rubydoc.info/gems/rubocop-rspec/RuboCop/Cop/RSpec/Capybara/FeatureMethods
 
+== RSpec/Capybara/SpecificMatcher
+
+|===
+| Enabled by default | Safe | Supports autocorrection | Version Added | Version Changed
+
+| Pending
+| Yes
+| No
+| 2.12
+| -
+|===
+
+Checks for there is a more specific matcher offered by Capybara.
+
+=== Examples
+
+[source,ruby]
+----
+# bad
+expect(page).to have_selector('button')
+expect(page).to have_no_selector('button.cls')
+expect(page).to have_css('button')
+expect(page).to have_no_css('a.cls', exact_text: 'foo')
+expect(page).to have_css('table.cls')
+expect(page).to have_css('select')
+
+# good
+expect(page).to have_button
+expect(page).to have_no_button(class: 'cls')
+expect(page).to have_button
+expect(page).to have_no_link('foo', class: 'cls')
+expect(page).to have_table(class: 'cls')
+expect(page).to have_select
+----
+
+=== References
+
+* https://www.rubydoc.info/gems/rubocop-rspec/RuboCop/Cop/RSpec/Capybara/SpecificMatcher
+
 == RSpec/Capybara/VisibilityMatcher
 
 |===
--- a/lib/rubocop/cop/rspec/capybara/specific_matcher.rb
+++ b/lib/rubocop/cop/rspec/capybara/specific_matcher.rb
@@ -1,0 +1,77 @@
+# frozen_string_literal: true
+
+module RuboCop
+  module Cop
+    module RSpec
+      module Capybara
+        # Checks for there is a more specific matcher offered by Capybara.
+        #
+        # @example
+        #
+        #   # bad
+        #   expect(page).to have_selector('button')
+        #   expect(page).to have_no_selector('button.cls')
+        #   expect(page).to have_css('button')
+        #   expect(page).to have_no_css('a.cls', exact_text: 'foo')
+        #   expect(page).to have_css('table.cls')
+        #   expect(page).to have_css('select')
+        #
+        #   # good
+        #   expect(page).to have_button
+        #   expect(page).to have_no_button(class: 'cls')
+        #   expect(page).to have_button
+        #   expect(page).to have_no_link('foo', class: 'cls')
+        #   expect(page).to have_table(class: 'cls')
+        #   expect(page).to have_select
+        #
+        class SpecificMatcher < Base
+          MSG = 'Prefer `%<good_matcher>s` over `%<bad_matcher>s`.'
+          RESTRICT_ON_SEND = %i[have_selector have_no_selector have_css
+                                have_no_css].freeze
+          SPECIFIC_MATCHER = {
+            'button' => 'button',
+            'a' => 'link',
+            'table' => 'table',
+            'select' => 'select'
+          }.freeze
+
+          # @!method first_argument(node)
+          def_node_matcher :first_argument, <<-PATTERN
+            (send nil? _ (str $_) ... )
+          PATTERN
+
+          def on_send(node)
+            return unless (arg = first_argument(node))
+            return unless (matcher = specific_matcher(arg))
+            return if acceptable_pattern?(arg)
+
+            add_offense(node, message: message(node, matcher))
+          end
+
+          private
+
+          def specific_matcher(arg)
+            splitted_arg = arg[/^\w+/, 0]
+            SPECIFIC_MATCHER[splitted_arg]
+          end
+
+          def acceptable_pattern?(arg)
+            arg.match?(/\[.+=\w+\]/)
+          end
+
+          def message(node, matcher)
+            format(MSG,
+                   good_matcher: good_matcher(node, matcher),
+                   bad_matcher: node.method_name)
+          end
+
+          def good_matcher(node, matcher)
+            node.method_name
+              .to_s
+              .gsub(/selector|css/, matcher.to_s)
+          end
+        end
+      end
+    end
+  end
+end
--- a/spec/rubocop/cop/rspec/capybara/specific_matcher_spec.rb
+++ b/spec/rubocop/cop/rspec/capybara/specific_matcher_spec.rb
@@ -1,0 +1,86 @@
+# frozen_string_literal: true
+
+RSpec.describe RuboCop::Cop::RSpec::Capybara::SpecificMatcher do
+  it 'does not register an offense for abstract matcher when ' \
+    'first argument is not a replaceable element' do
+    expect_no_offenses(<<-RUBY)
+      expect(page).to have_selector('article')
+      expect(page).to have_no_selector('body')
+      expect(page).to have_css('tbody')
+    RUBY
+  end
+
+  it 'does not register an offense for abstract matcher when ' \
+    'first argument is not an element' do
+    expect_no_offenses(<<-RUBY)
+      expect(page).to have_no_css('.a')
+      expect(page).to have_selector('#button')
+      expect(page).to have_no_selector('[table]')
+    RUBY
+  end
+
+  it 'registers an offense when using `have_selector`' do
+    expect_offense(<<-RUBY)
+    expect(page).to have_selector('button')
+                    ^^^^^^^^^^^^^^^^^^^^^^^ Prefer `have_button` over `have_selector`.
+    expect(page).to have_selector('a')
+                    ^^^^^^^^^^^^^^^^^^ Prefer `have_link` over `have_selector`.
+    expect(page).to have_selector('table')
+                    ^^^^^^^^^^^^^^^^^^^^^^ Prefer `have_table` over `have_selector`.
+    expect(page).to have_selector('select')
+                    ^^^^^^^^^^^^^^^^^^^^^^^ Prefer `have_select` over `have_selector`.
+    RUBY
+  end
+
+  it 'registers an offense when using `have_no_selector`' do
+    expect_offense(<<-RUBY)
+    expect(page).to have_no_selector('button')
+                    ^^^^^^^^^^^^^^^^^^^^^^^^^^ Prefer `have_no_button` over `have_no_selector`.
+    RUBY
+  end
+
+  it 'registers an offense when using `have_css`' do
+    expect_offense(<<-RUBY)
+    expect(page).to have_css('button')
+                    ^^^^^^^^^^^^^^^^^^ Prefer `have_button` over `have_css`.
+    RUBY
+  end
+
+  it 'registers an offense when using `have_no_css`' do
+    expect_offense(<<-RUBY)
+    expect(page).to have_no_css('button')
+                    ^^^^^^^^^^^^^^^^^^^^^ Prefer `have_no_button` over `have_no_css`.
+    RUBY
+  end
+
+  it 'registers an offense when using abstract matcher and other args' do
+    expect_offense(<<-RUBY)
+    expect(page).to have_css('button', exact_text: 'foo')
+                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ Prefer `have_button` over `have_css`.
+    RUBY
+  end
+
+  it 'registers an offense when using abstract matcher with state' do
+    expect_offense(<<-RUBY)
+    expect(page).to have_css('button[disabled]', exact_text: 'foo')
+                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ Prefer `have_button` over `have_css`.
+    expect(page).to have_css('button:not([disabled])', exact_text: 'bar')
+                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ Prefer `have_button` over `have_css`.
+    RUBY
+  end
+
+  it 'does not register an offense for abstract matcher when ' \
+    'first argument is element with nonreplaceable attributes' do
+    expect_no_offenses(<<-RUBY)
+      expect(page).to have_css('button[foo=bar]')
+      expect(page).to have_css('button[foo-bar=baz]', exact_text: 'foo')
+    RUBY
+  end
+
+  it 'does not register an offense for abstract matcher when ' \
+    'first argument is dstr' do
+    expect_no_offenses(<<-'RUBY')
+      expect(page).to have_css(%{a[href="#{foo}"]}, text: "bar")
+    RUBY
+  end
+end
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 * Fix incorrect path suggested by `RSpec/FilePath` cop when second argument contains spaces. ([@tejasbubane][])
 * Fix autocorrect for EmptyLineSeparation. ([@johnny-miyake][])
+* Add new `RSpec/Capybara/SpecificMatcher` cop. ([@ydah][])
 
 ## 2.11.1 (2022-05-18)
 
--- a/config/default.yml
+++ b/config/default.yml
@@ -832,6 +832,12 @@ RSpec/Capybara/FeatureMethods:
   VersionChanged: '2.0'
   Reference: https://www.rubydoc.info/gems/rubocop-rspec/RuboCop/Cop/RSpec/Capybara/FeatureMethods
 
+RSpec/Capybara/SpecificMatcher:
+  Description: Checks for there is a more specific matcher offered by Capybara.
+  Enabled: pending
+  VersionAdded: '2.12'
+  Reference: https://www.rubydoc.info/gems/rubocop-rspec/RuboCop/Cop/RSpec/Capybara/SpecificMatcher
+
 RSpec/Capybara/VisibilityMatcher:
   Description: Checks for boolean visibility in Capybara finders.
   Enabled: true
--- a/docs/modules/ROOT/pages/cops.adoc
+++ b/docs/modules/ROOT/pages/cops.adoc
@@ -93,6 +93,7 @@
 
 * xref:cops_rspec_capybara.adoc#rspeccapybara/currentpathexpectation[RSpec/Capybara/CurrentPathExpectation]
 * xref:cops_rspec_capybara.adoc#rspeccapybara/featuremethods[RSpec/Capybara/FeatureMethods]
+* xref:cops_rspec_capybara.adoc#rspeccapybara/specificmatcher[RSpec/Capybara/SpecificMatcher]
 * xref:cops_rspec_capybara.adoc#rspeccapybara/visibilitymatcher[RSpec/Capybara/VisibilityMatcher]
 
 === Department xref:cops_rspec_factorybot.adoc[RSpec/FactoryBot]
--- a/lib/rubocop/cop/rspec_cops.rb
+++ b/lib/rubocop/cop/rspec_cops.rb
@@ -2,6 +2,7 @@
 
 require_relative 'rspec/capybara/current_path_expectation'
 require_relative 'rspec/capybara/feature_methods'
+require_relative 'rspec/capybara/specific_matcher'
 require_relative 'rspec/capybara/visibility_matcher'
 
 require_relative 'rspec/factory_bot/attribute_defined_statically'


lib/rubocop/cop/rspec/capybara/specific_matcher.rb
```diff
@@ -0,0 +1,73 @@
+# frozen_string_literal: true
+
+module RuboCop
+  module Cop
+    module RSpec
+      module Capybara
+        # Checks for if there is a more specific matcher offered by Capybara.
```
pirj, line 7: One of "for" or "if" sound redundant.

lib/rubocop/cop/rspec/capybara/specific_matcher.rb
```diff
@@ -0,0 +1,73 @@
...
+          SPECIFIC_MATCHER = {
+            button: 'button',
+            a: 'link',
+            table: 'table'
+          }.freeze
+          MSG_HAVE_CSS = 'Use `have_%<matcher>s` instead of `have_css`.'
+          MSG_HAVE_NO_CSS = 'Use `have_no_%<matcher>s` ' \
+                              'instead of `have_no_css`.'
+          MSG_HAS_CSS = 'Use `has_%<matcher>s` instead of `has_css`.'
+          MSG_HAS_NO_CSS = 'Use `has_no_%<matcher>s` instead of `has_no_css`.'
```
pirj, line 40: What are the cases when the use of `has_...` matchers forms a correct English sentence?

spec/rubocop/cop/rspec/capybara/specific_matcher_spec.rb
```diff
@@ -0,0 +1,335 @@
...
+  it 'does not register an offense for `have_css` when elements with' \
+    'not element' do
+    expect_no_offenses(<<-RUBY)
+      expect(page).to have_css('.a')
+      expect(page).to have_css('#button')
+      expect(page).to have_css('[table]')
+    RUBY
+  end
+
+  it 'does not register an offense for `have_no_css` when elements with' \
```
pirj, line 22: Does it make sense to combine examples of different matchers (`have_css`, `have_no_css`, `has_css`, `has_no_css`) into one example?

spec/rubocop/cop/rspec/capybara/specific_matcher_spec.rb
```diff
@@ -0,0 +1,335 @@
...
+      expect(page).to has_no_css('.a')
+      expect(page).to has_no_css('#button')
+      expect(page).to has_no_css('[table]')
+    RUBY
+  end
+
+  it "registers an offense when using `have_css('button')`" do
+    expect_offense(<<-RUBY)
+    expect(page).to have_css('button')
+                    ^^^^^^^^ Use `have_button` instead of `have_css`.
```
pirj, line 79: It feels that the offence relates to the whole matcher, not just the method name. It will probably have to change anyway for autocorrection.

pirj (commented):
Thank you!

To me, specs are overly verbose.
If we'd add another dimension, e.g. a check for symbols (`have_css(:button)`), it feels that we would have to double the number of examples. Is there a way to structure the spec in a way to reduce the current and prevent the fututre combinatoric explosion?

docs/modules/ROOT/pages/cops_rspec_capybara.adoc
```diff
@@ -106,6 +106,48 @@ end
...
+expect(page).to have_link('foo', class: 'cls')
+expect(page).to have_table
+expect(page).to have_table(class: 'cls')
+
+=== Examples
+
+[source,ruby]
+----
+# bad
+----
```
pirj, line 145: This seems to be broken.

lib/rubocop/cop/rspec/capybara/specific_matcher.rb
```diff
@@ -0,0 +1,73 @@
...
+module RuboCop
+  module Cop
+    module RSpec
+      module Capybara
+        # Checks for if there is a more specific matcher offered by Capybara.
+        #
+        # @example
+        #
+        #   # bad
+        # expect(page).to have_css('button')
```
pirj, line 12: I believe it's due to the lack of indentation here.

lib/rubocop/cop/rspec/capybara/specific_matcher.rb
```diff
@@ -0,0 +1,73 @@
...
+        # expect(page).to have_css('button', exact_text: 'foo')
+        # expect(page).to have_css('button[disabled]', exact_text: 'foo')
+        # expect(page).to have_css('a')
+        # expect(page).to have_css('a.cls', exact_text: 'foo')
+        # expect(page).to have_css('table')
+        # expect(page).to have_css('table.cls')
```
pirj, line 18: A more generic `have_selector` needs to be checked, too.

pirj:
Tested with `real-world-rspec`:
```
57497 files inspected, 64 offenses detected
```
No errors 👍

lib/rubocop/cop/rspec/capybara/specific_matcher.rb
pirj: 2x much offences detected with `have_selector` included:
```
﻿﻿57497 files inspected, 167 offenses detected
```

lib/rubocop/cop/rspec/capybara/specific_matcher.rb
ydah: Indeed it is. Remove the redundant if.

lib/rubocop/cop/rspec/capybara/specific_matcher.rb
ydah: Oh, `has_...` method ends with `? ` at the end.
https://www.rubydoc.info/gems/capybara/Capybara/Node/Matchers#has_css%3F-instance_method

I've improved the definition of this message because each time the number of supported methods increases, the definition of this message also increases.

spec/rubocop/cop/rspec/capybara/specific_matcher_spec.rb
ydah: spec structure was creating redundancy, so I reviewed the structure and organized the tests. What do you think?

spec/rubocop/cop/rspec/capybara/specific_matcher_spec.rb
ydah: Indeed it is. I have modified it so that it is applied to the whole.

docs/modules/ROOT/pages/cops_rspec_capybara.adoc
ydah: Oops, sorry. I hadn't noticed...

lib/rubocop/cop/rspec/capybara/specific_matcher.rb
ydah: Thank you very much. Fixed the indentation!

lib/rubocop/cop/rspec/capybara/specific_matcher.rb
ydah: Oh! Added `have_selector`。

spec/rubocop/cop/rspec/capybara/specific_matcher_spec.rb
```diff
@@ -0,0 +1,85 @@
...
+      expect(page).to have_css('tbody')
+    RUBY
+  end
+
+  it 'does not register an offense for abstract matcher when elements with' \
+    'not element' do
+    expect_no_offenses(<<-RUBY)
+      expect(page).to have_no_css('.a')
+      expect(page).to has_css?('#button')
+      expect(page).to has_no_css?('[table]')
```
pirj, line 18: Is this a correct usage of `has_css?`?
In docs, I see this method used as a node matcher, not as an RSpec predicate matcher:
```ruby
expect(page.has_css?('p#foo')).to be true
```
vs
```
expect(page).to have_css('p#foo')
```

I would rather prefer only checking the RSpec matcher, and let something like `rspec-capybara` to take care of `has_css?`.

spec/rubocop/cop/rspec/capybara/specific_matcher_spec.rb
```diff
@@ -0,0 +1,85 @@
...
+                    ^^^^^^^^^^^^^^^^^^^^^^ Use `#have_table` instead of `#have_selector`.
+    expect(page).to have_selector('select')
+                    ^^^^^^^^^^^^^^^^^^^^^^^ Use `#have_select` instead of `#have_selector`.
+    RUBY
+  end
+
+  it 'registers an offense when using `#have_none_of_selectors`' do
+    expect_offense(<<-RUBY)
+    expect(page).to have_none_of_selectors('button')
+                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ Use `#have_no_button` instead of `#have_none_of_selectors`.
```
pirj, line 38: Is there a `have_no_button`?

spec/rubocop/cop/rspec/capybara/specific_matcher_spec.rb
```diff
@@ -0,0 +1,85 @@
...
+      expect(page).to have_no_css('.a')
+      expect(page).to has_css?('#button')
+      expect(page).to has_no_css?('[table]')
+    RUBY
+  end
+
+  it 'registers an offense when using `#have_selector`' do
+    expect_offense(<<-RUBY)
+    expect(page).to have_selector('button')
+                    ^^^^^^^^^^^^^^^^^^^^^^^ Use `#have_button` instead of `#have_selector`.
```
pirj, line 25: There is no ambiguity to the subject of the offence. We point at the method call. Is it necessary to tell the method name in the offence message again?

spec/rubocop/cop/rspec/capybara/specific_matcher_spec.rb
```diff
@@ -0,0 +1,85 @@
...
+    expect(page).to have_selector('a')
+                    ^^^^^^^^^^^^^^^^^^ Use `#have_link` instead of `#have_selector`.
```
pirj, line 27: I couldn't quickly find any examples in our code where we prefix methods with a `#` in offence messages. Is it necessary?

spec/rubocop/cop/rspec/capybara/specific_matcher_spec.rb
```diff
@@ -0,0 +1,85 @@
...
+    expect(page).to have_selector('table')
+                    ^^^^^^^^^^^^^^^^^^^^^^ Use `#have_table` instead of `#have_selector`.
+    expect(page).to have_selector('select')
+                    ^^^^^^^^^^^^^^^^^^^^^^^ Use `#have_select` instead of `#have_selector`.
+    RUBY
+  end
+
+  it 'registers an offense when using `#have_none_of_selectors`' do
+    expect_offense(<<-RUBY)
+    expect(page).to have_none_of_selectors('button')
```
pirj, line 37: What if the `have_none_of_selectors` is used with mutiple arguments, and 'button' is not the first one?
Even if the button is the first one, `have_*_of_selectors` accept multiple locators, while other methods just one.
There is no direct substitute.

lib/rubocop/cop/rspec/capybara/specific_matcher.rb
```diff
@@ -0,0 +1,69 @@
...
+            return unless (matcher = specific_matcher(node))
+
+            add_offense(node, message: message(node, matcher))
+          end
+
+          private
+
+          def specific_matcher(node)
+            SPECIFIC_MATCHER.fetch(
+              node.first_argument.children.first.split(/\W/).first.to_sym, nil
```
pirj, line 49: Cosmetic: with lengthy constructs like `node.first_argument.children.first`, it may make sense to consider using node pattern.

lib/rubocop/cop/rspec/capybara/specific_matcher.rb
```diff
@@ -0,0 +1,69 @@
...
+            return unless node.first_argument.str_type?
+            return unless (matcher = specific_matcher(node))
+
+            add_offense(node, message: message(node, matcher))
+          end
+
+          private
+
+          def specific_matcher(node)
+            SPECIFIC_MATCHER.fetch(
```
pirj, line 48: Is `fetch` with a default `nil` same as just `SPECIFIC_MATCHER[...]`?

spec/rubocop/cop/rspec/capybara/specific_matcher_spec.rb
ydah: I intended what is defined here `have_no_button`.
https://github.com/teamcapybara/capybara/blob/51bca8300136ba2dcf45e92883a07845105f2b3a/lib/capybara/rspec/matchers.rb#L162-L169

spec/rubocop/cop/rspec/capybara/specific_matcher_spec.rb
ydah: Oops, that's right.
When I ran the `new_cop` command, it generated one with a prefix of `#`, so I completely thought this was the standard.

```console
bundle exec rake 'new_cop[Foo/Bar]' 
```

```ruby
        MSG = 'Use `#good_method` instead of `#bad_method`.'
```

spec/rubocop/cop/rspec/capybara/specific_matcher_spec.rb
ydah: I would have preferred more descriptive, but it certainly seems redundant in this case.

spec/rubocop/cop/rspec/capybara/specific_matcher_spec.rb
ydah: Oh, yes, indeed. I think I'll exclude `have_*_of_selectors` as it seems to be outside the realm of this Cop.

lib/rubocop/cop/rspec/capybara/specific_matcher.rb
ydah: `SPECIFIC_MATCHER[...]` changed to use.

spec/rubocop/cop/rspec/capybara/specific_matcher_spec.rb
ydah: > I would rather prefer only checking the RSpec matcher, and let something like rspec-capybara to take care of has_css?.

It certainly looks better that way.

lib/rubocop/cop/rspec/capybara/specific_matcher.rb
```diff
@@ -0,0 +1,71 @@
...
+  module Cop
+    module RSpec
+      module Capybara
+        # Checks for there is a more specific matcher offered by Capybara.
+        #
+        # @example
+        #
+        #   # bad
+        #   expect(page).to have_selector('button')
+        #   expect(page).to have_none_of_selectors('button.cls')
```
pirj, line 13: Leftover?

lib/rubocop/cop/rspec/capybara/specific_matcher.rb
```diff
@@ -0,0 +1,71 @@
...
+          SPECIFIC_MATCHER = {
+            button: 'button',
+            a: 'link',
+            table: 'table',
+            select: 'select'
+          }.freeze
+
+          # @!method first_argument(node)
+          def_node_search :first_argument, <<-PATTERN
+            (send nil? _ (str $_) ?... )
```
pirj, line 39: 👏 Today I learned `?...` is possible. I remember struggling to write a node matcher that would accept many or no elements. 👍

lib/rubocop/cop/rspec/capybara/specific_matcher.rb
```diff
@@ -0,0 +1,71 @@
...
+          PATTERN
+
+          def on_send(node)
+            return unless node.first_argument.str_type?
```
pirj, line 43: This seems redundant now - the node patter would only match for a string.

lib/rubocop/cop/rspec/capybara/specific_matcher.rb
```diff
@@ -0,0 +1,71 @@
...
+            return unless (matcher = specific_matcher(node))
+
+            add_offense(node, message: message(node, matcher))
+          end
+
+          private
+
+          def specific_matcher(node)
+            first_argument(node) do |args|
+              return SPECIFIC_MATCHER[args.first.split(/\W/).first.to_sym]
```
pirj, line 53: Is `return` necessary here?

lib/rubocop/cop/rspec/capybara/specific_matcher.rb
```diff
@@ -0,0 +1,71 @@
...
+        #   expect(page).to have_button
+        #   expect(page).to have_button(class: 'cls')
+        #   expect(page).to have_button
+        #   expect(page).to have_no_link('foo', class: 'cls')
+        #   expect(page).to have_table(class: 'cls')
+        #   expect(page).to have_select
+        #
+        class SpecificMatcher < Base
+          MSG = 'Use `%<good_matcher>s`.'
+          RESTRICT_ON_SEND = %i[have_selector have_css have_no_css].freeze
```
pirj, line 29: I think `have_no_selector` exists, too.

spec/rubocop/cop/rspec/capybara/specific_matcher_spec.rb
```diff
@@ -0,0 +1,64 @@
...
+  it 'does not register an offense for abstract matcher when elements with' \
+    'no matcher specified in detail' do
+    expect_no_offenses(<<-RUBY)
+      expect(page).to have_selector('article')
+      expect(page).to have_none_of_selectors('body')
+      expect(page).to have_css('tbody')
+    RUBY
+  end
+
+  it 'does not register an offense for abstract matcher when elements with' \
```
pirj, line 13: `\` combines this into a string without adding a space:
> does not register an offense for abstract matcher when elements withnot element

pirj (approved):
A few cosmetic changes, and good to go from my point of view.
Thanks a lot for your effort, @ydah !

lib/rubocop/cop/rspec/capybara/specific_matcher.rb
bquorning: Or the "good" example `expect(page).to have_button(class: 'cls')` probably needs to be changed.

bquorning: Hmm, it seems `have_none_of_selectors` causing offense is not covered by specs. Perhaps it should be?

pirj: We decided not to deal with `have_any_of_selectors`/`have_none_of_selectors`/`have_all_of_selectors`, as they accept several locators, while those matchers that we currently inspect accept just one.
This can be quite hard to do right. Related discussion: https://github.com/rubocop/rubocop-rspec/pull/1294#discussion_r890823552

The cop might see an improvement on this front, and e.g. autocorrection in a follow-up.

lib/rubocop/cop/rspec/capybara/specific_matcher.rb
bquorning: The specs pass if I change `?...` to `...` and `args.first` to `args` below. Do we need to add some spec examples where there are no args?

lib/rubocop/cop/rspec/capybara/specific_matcher.rb
bquorning: You can omit `return` if you use `def_node_matcher` instead of `def_node_search` above. With `def_node_matcher` another possible implementation is
```rb
args = first_argument(node)
SPECIFIC_MATCHER[args.first.split(/\W/).first.to_sym]
```

lib/rubocop/cop/rspec/capybara/specific_matcher.rb
ydah: Oops, sorry. I fixed it.

lib/rubocop/cop/rspec/capybara/specific_matcher.rb
ydah: Indeed is it, I'll fix it.

lib/rubocop/cop/rspec/capybara/specific_matcher.rb
ydah: Finally, the `SPECIFIC_MATCHER` key was made into a string and the method was improved as follows

```ruby
def specific_matcher(node)
  args = first_argument(node)
  SPECIFIC_MATCHER[args.first.split(/\W/).first]
end
```

lib/rubocop/cop/rspec/capybara/specific_matcher.rb
ydah: Oh, sure. I had forgotten about that.

spec/rubocop/cop/rspec/capybara/specific_matcher_spec.rb
ydah: Oops, I'll add that space.

lib/rubocop/cop/rspec/capybara/specific_matcher.rb
ydah: Certainly this place is `... ` is also working.

Here I certainly wanted to match for node the case of passing one or more arguments to `have_css`.
But I found out that `...` is fine. I may have been mistaken.

spec/rubocop/cop/rspec/capybara/specific_matcher_spec.rb
```diff
@@ -0,0 +1,71 @@
...
+    'no matcher specified in detail' do
+    expect_no_offenses(<<-RUBY)
+      expect(page).to have_selector('article')
+      expect(page).to have_no_selector('body')
+      expect(page).to have_css('tbody')
+    RUBY
+  end
+
+  it 'does not register an offense for abstract matcher when elements with ' \
+    'not element' do
```
bquorning, line 14: I’m not sure what “when elements with not element” means.

bquorning (approved):
Looks good so far, just a few comments left.

spec/rubocop/cop/rspec/capybara/specific_matcher_spec.rb
ydah: Improved wording. What do you think?

https://github.com/rubocop/rubocop-rspec/blob/6fb113cdc3dc298f95359a5519fd6050280ce98f/spec/rubocop/cop/rspec/capybara/specific_matcher_spec.rb#L4-L5

https://github.com/rubocop/rubocop-rspec/blob/6fb113cdc3dc298f95359a5519fd6050280ce98f/spec/rubocop/cop/rspec/capybara/specific_matcher_spec.rb#L13-L14

lib/rubocop/cop/rspec/capybara/specific_matcher.rb
```diff
@@ -0,0 +1,70 @@
...
+          def on_send(node)
+            return unless (matcher = specific_matcher(node))
+
+            add_offense(node, message: message(node, matcher))
+          end
+
+          private
+
+          def specific_matcher(node)
+            arg = first_argument(node).first[/^\w+/, 0]
```
bquorning, line 52: Last thing: If we use `def_node_matcher` instead of `def_node_search`, we could drop the `.first` here, right? Would it cause any problems?

ydah: That is correct. Thank you very much.

bquorning (approved):
🎉

lib/rubocop/cop/rspec/capybara/specific_matcher.rb
```diff
@@ -0,0 +1,70 @@
...
+        #   # good
+        #   expect(page).to have_button
+        #   expect(page).to have_no_button(class: 'cls')
+        #   expect(page).to have_button
+        #   expect(page).to have_no_link('foo', class: 'cls')
+        #   expect(page).to have_table(class: 'cls')
+        #   expect(page).to have_select
+        #
+        class SpecificMatcher < Base
+          MSG = 'Use `%<good_matcher>s`.'
```
Darhazer, line 28: While we don't have hard rules on messages, for alternative matchers we usually use `Prefer %<good_matcher>s over %<bad_matcher>s.`

Darhazer (commented):
The original issue mentions that usages like `button[data-foo=1]` can't be replaced, but it seems that we do not whitelist the replacable options

lib/rubocop/cop/rspec/capybara/specific_matcher.rb
```diff
@@ -0,0 +1,70 @@
...
+          def on_send(node)
+            return unless (matcher = specific_matcher(node))
+
+            add_offense(node, message: message(node, matcher))
+          end
+
+          private
+
+          def specific_matcher(node)
+            arg = first_argument(node)[/^\w+/, 0]
```
pirj, line 52: This errors out on this:
```ruby
# real-world-rspec/administrate/spec/support/have_flash_matcher.rb
module Features
  def have_flash(text, options = {})
    options.reverse_merge!(type: :notice)
    have_css(".flash-#{options[:type]}", text: text)
  end
end
```

The first argument is a `dstr`, not `str`.

A few similar cases appear if you run it against `real-world-rspec`:
```ruby
# real-world-rspec/administrate/spec/administrate/views/fields/url/_index_spec.rb
    expect(rendered).to have_css(
      %{a[href="#{product.image_url}"]},
      text: product.image_url,
    )
```

I'd suggest adding an example to prevent regressions after fixing.

pirj: It's fine to ignore everything apart from `str` I believe.

lib/rubocop/cop/rspec/capybara/specific_matcher.rb
Darhazer: I see how It could be confusing, when only some instances of have_selector will register an offense, and the suggested matchers would be different. We can either keep it as it is now, or have something like 'Use have_link instead of have_selector when checking for links'

ydah: Take the form `Prefer %<good_matcher>s over %<bad_matcher>s.`.

lib/rubocop/cop/rspec/capybara/specific_matcher.rb
ydah: Oops, I had not noticed that pattern. Thanks, I've decided to ignore the `dstr`.

ydah:
I overlooked the fact in the original issue that a form like `button[data-foo=1]` should be whitelisted.
I whitelisted that one as well and added spec.

lib/rubocop/cop/rspec/capybara/specific_matcher.rb
```diff
@@ -0,0 +1,77 @@
...
+
+          private
+
+          def specific_matcher(arg)
+            splitted_arg = arg[/^\w+/, 0]
+            SPECIFIC_MATCHER[splitted_arg]
+          end
+
+          def acceptable_pattern?(arg)
+            arg.match?(/\[[\w|\W]+=\w+\]/)
```
bquorning, line 59: `\w` and `\W` together match anything 😄 
```suggestion
            arg.match?(/\[.+=\w+\]/)
```

ydah: Oh. It's my bad habit... 😢 Sorry.

Drenmi:
This is great! Thank you for your work, @ydah. 🙇

pirj:
Final check ✅ 
```
57497 files inspected, 169 offenses detected
```

Thanks a lot, @ydah !

PR #1302, opened 2022-06-20
Improved documentation for `RSpec/ExpectActual`

ydah:
This PR is improves documentation for `RSpec/ExpectActual`.

For example, if you create a gem by `$ bundle gem hoge`, the following rspec code will be created.

```ruby
it "does something useful" do
  expect(false).to eq(true)
end
```

At first glance, it seems to be autocorrected, but when both expected and actual values are literal, no autocorrection is performed.
We'll add this as an explanation for cop.

---

* [-] Added tests.
* [x] Updated documentation.

### DIFF
--- a/lib/rubocop/cop/rspec/expect_actual.rb
+++ b/lib/rubocop/cop/rspec/expect_actual.rb
@@ -5,6 +5,8 @@ module RuboCop
     module RSpec
       # Checks for `expect(...)` calls containing literal values.
       #
+      # Autocorrection is performed when the expected is not a literal.
+      #
       # @example
       #   # bad
       #   expect(5).to eq(price)
--- a/docs/modules/ROOT/pages/cops_rspec.adoc
+++ b/docs/modules/ROOT/pages/cops_rspec.adoc
@@ -1483,6 +1483,8 @@ end
 
 Checks for `expect(...)` calls containing literal values.
 
+Autocorrection is performed when the expected is not a literal.
+
 === Examples
 
 [source,ruby]


lib/rubocop/cop/rspec/expect_actual.rb
```diff
@@ -61,6 +65,8 @@ def on_send(node)
           expect_literal(node) do |actual, matcher, expected|
             add_offense(actual.source_range) do |corrector|
               next unless SUPPORTED_MATCHERS.include?(matcher)
+
+              p literal?(expected)
```
pirj, line 69: Leftover?

lib/rubocop/cop/rspec/expect_actual.rb
```diff
@@ -5,6 +5,10 @@ module Cop
     module RSpec
       # Checks for `expect(...)` calls containing literal values.
       #
+      # Autocorrect is supported, but when both expected and actual values
```
pirj, line 8: WDYT of "Autocorrection is performed when the expected is not a literal"?

lib/rubocop/cop/rspec/expect_actual.rb
ydah: Oops, Sorry. I deleted it.

lib/rubocop/cop/rspec/expect_actual.rb
ydah: Thank you so much.That looks better. Changed wording.

pirj (approved):

bquorning (approved):

PR #1303, opened 2022-06-20
Document aggregate_failures behavior in MultipleExpectations cop

edgibbs:
Rspec/MultipleExpectations supports :aggregate_failures by simply ignoring
any blocks enclosed in them. The only issue is that they don't really
appear in any of the docs so the behavior is mostly hidden. This update
merely adds two inline examples to RSpec/MultipleExpectations to allow
for greater discoverability.

---

* [ ] Added tests.
* [x] Updated documentation.

### DIFF
--- a/lib/rubocop/cop/rspec/multiple_expectations.rb
+++ b/lib/rubocop/cop/rspec/multiple_expectations.rb
@@ -31,6 +31,26 @@ module RuboCop
       #     end
       #   end
       #
+      # @example `aggregate_failures: true` (default)
+      #
+      #  # good - the cop ignores when RSpec aggregates failures
+      #   describe UserCreator do
+      #     it 'builds a user', :aggregate_failures do
+      #       expect(user.name).to eq("John")
+      #       expect(user.age).to eq(22)
+      #     end
+      #   end
+      #
+      # @example `aggregate_failures: false`
+      #
+      #  # Detected as an offense
+      #   describe UserCreator do
+      #     it 'builds a user', aggregate_failures: false do
+      #       expect(user.name).to eq("John")
+      #       expect(user.age).to eq(22)
+      #     end
+      #   end
+      #
       # @example configuration
       #
       #   # .rubocop.yml
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 * Add new `RSpec/Capybara/SpecificMatcher` cop. ([@ydah][])
 * Fixed false offense detection in `FactoryBot/CreateList` when a n.times block is including method calls in the factory create arguments. ([@ngouy][])
 * Fix error in `RSpec/RSpec/FactoryBot/CreateList` cop for empty block. ([@tejasbubane][])
+* Update `RSpec/MultipleExpectations` cop documentation with examples of aggregate_failures use. ([@edgibbs][])
 
 ## 2.11.1 (2022-05-18)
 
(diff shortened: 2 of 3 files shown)


lib/rubocop/cop/rspec/multiple_expectations.rb
```diff
@@ -31,6 +31,26 @@ module RSpec
       #     end
       #   end
       #
+      # @example with aggregate_failures (defaults to true)
```
pirj, line 34: I suggest:
```
@example `aggregate_failures: true` (default)
```

lib/rubocop/cop/rspec/multiple_expectations.rb
```diff
@@ -31,6 +31,26 @@ module RSpec
...
+      #
+      #  # not flagged by rubocop
```
pirj, line 36: "No offense" or just "good", or "good - the cop ignores when RSpec aggregates failures".

lib/rubocop/cop/rspec/multiple_expectations.rb
```diff
@@ -31,6 +31,26 @@ module RSpec
...
+      #   describe UserCreator do
+      #     it 'builds a user', :aggregate_failures do
+      #       expect(user.name).to eq("John")
+      #       expect(user.age).to eq(22)
+      #     end
+      #   end
+      #
+      # @example with aggregate_failures: false
+      #
+      #  # flagged by rubocop
```
pirj, line 46: "Detected as an offense" or just "bad".

lib/rubocop/cop/rspec/multiple_expectations.rb
```diff
@@ -31,6 +31,26 @@ module RSpec
...
+      #
+      #  # not flagged by rubocop
+      #   describe UserCreator do
+      #     it 'builds a user', :aggregate_failures do
+      #       expect(user.name).to eq("John")
+      #       expect(user.age).to eq(22)
+      #     end
+      #   end
+      #
+      # @example with aggregate_failures: false
```
pirj, line 44:
```
@example `aggregate_failures: false`
```

pirj (approved):
Thank you!
Some cosmetic notes below.

You may want to adjust the `@example configuration` section below if you like. I'd replace "not flagged by rubocop" to "good - two offenses are considered tolerable".

lib/rubocop/cop/rspec/multiple_expectations.rb
edgibbs: 👍🏾 Thanks for the context. I was trying to find a similar convention in the existing examples. I made the change. The only question was if it made sense to inject a `when` so that it it the rendered document heading shows up as:

when`aggregate_failures: true` (default) 

versus

`aggregate_failures: true` (default)
```
 @example when `aggregate_failures: true` (default)
```

lib/rubocop/cop/rspec/multiple_expectations.rb
edgibbs: Updated with change.

lib/rubocop/cop/rspec/multiple_expectations.rb
edgibbs: Updated with change.

lib/rubocop/cop/rspec/multiple_expectations.rb
edgibbs: Updated with change.

lib/rubocop/cop/rspec/multiple_expectations.rb
pirj: Skimming through existing cops' docs, it seems that we don't add "when", even though it improves readability. [Example](https://docs.rubocop.org/rubocop-rspec/cops_rspec.html#examples-16). I'd stick with the tradition.

pirj (approved):
Thanks!

PR #1305, opened 2022-06-24
Pass range as a positional arg

pirj:
As introduced in https://github.com/rubocop/rubocop/pull/10748

Since RuboCop with this PR wasn't released yet, the CI will be red.

---

* [-] Added tests.
* [-] Updated documentation.
* [x] The build (`bundle exec rake`) passes (be sure to run this locally, since it may produce updated documentation that you will need to commit). - locally, against the `rubocop` PR

### DIFF
--- a/lib/rubocop/cop/rspec/empty_hook.rb
+++ b/lib/rubocop/cop/rspec/empty_hook.rb
@@ -36,11 +36,9 @@ module RuboCop
         def on_block(node)
           empty_hook?(node) do |hook|
             add_offense(hook) do |corrector|
-              range = range_with_surrounding_space(
-                range: node.loc.expression,
-                side: :left
+              corrector.remove(
+                range_with_surrounding_space(node.loc.expression, side: :left)
               )
-              corrector.remove(range)
             end
           end
         end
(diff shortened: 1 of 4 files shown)


bquorning (approved):

bquorning:
I think this PR needs a mention in the changelog, just like #1181.

We should also ask:
1. Do we want to incorporate more newly added RuboCop features, when we do the requirement bump anyway?
2. Do we want to do the requirement just now, or is this a PR that we should wait a month before merging/releasing?

pirj:
[Released in RuboCop 1.31.0](https://github.com/rubocop/rubocop/releases/tag/v1.31.0).

pirj:
> Do we want to incorporate more newly added RuboCop features, when we do the requirement bump anyway?

Yes. But I'm not aware of any.

> wait a month before merging/releasing

Why? Not to force users into `rubocop` 1.31.0 and to wait for 1.31.1 that would probably include bugfixes?

bquorning:
> Why? Not to force users into `rubocop` 1.31.0 and to wait for 1.31.1 that would probably include bugfixes?

🤷🏼 I guess there’s no good reason to wait 😄

PR #1308, opened 2022-06-30
Mark RSpec/VerifiedDoubleReference as unsafe

Drowze:
It's unsafe as the constant might not exist, e.g.:

```ruby
# consider that Asdfasdf is not a defined constant
let(:object) { instance_double("Asdfasdf") }

it do
  expect(object).to receive(:hello)
  object.hello
end
```

If we ran RSpec/VerifiedDoubleReference's autocorrect on the above code, our previously passing test will fail with `NameError: uninitialized constant Asdfasdf`

Disclaimer: I understand this is a bad use of `instance_double` 🤷‍♂️ . Found this issue when trying to apply rubocop-rspec on a codebase I'm maintaining though.

---

* [ ] Added tests.
* [x] Updated documentation.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
 * Fixed false offense detection in `FactoryBot/CreateList` when a n.times block is including method calls in the factory create arguments. ([@ngouy][])
 * Fix error in `RSpec/RSpec/FactoryBot/CreateList` cop for empty block. ([@tejasbubane][])
 * Update `RSpec/MultipleExpectations` cop documentation with examples of aggregate_failures use. ([@edgibbs][])
+* Declare autocorrect as unsafe for `RSpec/VerifiedDoubleReference`. ([@Drowze][])
 
 ## 2.11.1 (2022-05-18)
 
@@ -704,3 +705,4 @@ Compatibility release so users can upgrade RuboCop to 0.51.0. No new features.
 [@johnny-miyake]: https://github.com/johnny-miyake
 [@ngouy]: https://github.com/ngouy
 [@edgibbs]: https://github.com/edgibbs
+[@Drowze]: https://github.com/Drowze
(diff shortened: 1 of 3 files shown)


bquorning:
This PR is about `instance_double("ConsoleNotifier")` being autocorrected to `instance_double(ConsoleNotifier)`. While both are valid uses, the latter may fail if `ConsoleNotifier` is an undefined constant. Hence, the autocorrect is unsafe.

Drowze:
@pirj I was referring specifically to the case where the instance_double is referring to a non-existing constant. It is valid code, but the sole purpose of `instance_double` is to verify we're not stubbing a method that would not exist on an instance of the constant we're referring to - If the constant does not exist, in my opinion that's a bad use of `instance_double` and should be replaced by a simple double.

(unless you're doing TDD and writing tests for a constant that does not exist *yet* - I guess this is a reasonable scenario...)

According [its docs](https://relishapp.com/rspec/rspec-mocks/v/3-9/docs/verifying-doubles/using-an-instance-double), RSpec even has a configuration option to raise an error in case you're referring to a constant that doesn't exist:
```ruby
RSpec.configure do |config|
  config.mock_with :rspec do |mocks|

    # This option should be set when all dependencies are being loaded
    # before a spec run, as is the case in a typical spec helper. It will
    # cause any verifying double instantiation for a class that does not
    # exist to raise, protecting against incorrectly spelt names.
    mocks.verify_doubled_constant_names = true

  end
end
```

pirj:
Please accept my apologies for the rushed judgement. The previous discussion was on a different topic.
The autocorrection for undefined (not yet defined as in TDD) is indeed incorrect.
It's not all about Rails/Zeitwerk, so the selective code loading case is valid, too.
Yet another argument is that `verify_doubled_constant_names` is `false` by default in RSpec, and this, to my best knowledge, won't change in the upcoming 4.0 release, so it's the case for the next few years if not forever.

All good, sorry to bother. Thanks for the contribution, @Drowze!

PR #1309, opened 2022-06-30
Bump version to 2.12.0

bquorning:
We have a few changes since the last release (v2.11.1), so I thought it might be time for a new release.

* Fix incorrect path suggested by `RSpec/FilePath` cop when second argument contains spaces.
* Fix autocorrect for EmptyLineSeparation.
* Add new `RSpec/Capybara/SpecificMatcher` cop.
* Fixed false offense detection in `FactoryBot/CreateList` when a n.times block is including method calls in the factory create arguments.
* Fix error in `RSpec/RSpec/FactoryBot/CreateList` cop for empty block.
* Update `RSpec/MultipleExpectations` cop documentation with examples of aggregate_failures use.
* Declare autocorrect as unsafe for `RSpec/VerifiedDoubleReference`.
* Add new `RSpec/Rails/HaveHttpStatus` cop.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Master (Unreleased)
 
+## 2.12.0 (2022-07-02)
+
 * Fix incorrect path suggested by `RSpec/FilePath` cop when second argument contains spaces. ([@tejasbubane][])
 * Fix autocorrect for EmptyLineSeparation. ([@johnny-miyake][])
 * Add new `RSpec/Capybara/SpecificMatcher` cop. ([@ydah][])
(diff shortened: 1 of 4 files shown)


pirj:
It's a bit late on my side to make good judgements, but I have a feeling that #1308 needs to be re-discussed.
WDYT if we release without it?

bquorning:
👍🏼 Let’s chat in #1308, and feel free to put a “request changes” status on this PR.

pirj (approved):
Thank you!

Maybe also add #1307, as it's ready and green?

bquorning:
Yep, let’s add that pr as well.

PR #1311, opened 2022-07-03
[Fix #1310] Fix a false positive for `RSpec/Capybara/SpecificMatcher`

### LINKED ISSUE #1310: [RSpec/Capybara/SpecificMatcher] False-positive with `have_table`
```rb
expect(page).to have_css('table tbody.some-class tr:last-child')
```

This results in a `Prefer have_table over have_css.` warning however in this case I have sub-matchers and that can't be replaced with `have_table`.

Otherwise the rule works perfect, great work!

cc @ydah

ydah:
Fix: #1310

Updated to not offense in the case of non-replaceable css selectors.

```ruby
expect(page).to have_css('button body')
expect(page).to have_css('a,h1')
expect(page).to have_css('table>tr')
expect(page).to have_css('select+option')
```

---

* [x] Added tests.
* [-] Updated documentation.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Master (Unreleased)
 
+* Fix a false positive for `RSpec/Capybara/SpecificMatcher`. ([@ydah][])
+
 ## 2.12.0 (2022-07-02)
 
 * Fix incorrect path suggested by `RSpec/FilePath` cop when second argument contains spaces. ([@tejasbubane][])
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 ## Master (Unreleased)
 
+* Fix a suggestions for safer conversions for `Lint/NonAtomicFileOperation`. ([@ydah][])
```
pirj, line 5: This seems to be from another PR.

pirj (approved):
Thank you for the quick fix!

CHANGELOG.md
ydah: Oops, sorry. Fixed it.

bquorning (approved):

PR #1316, opened 2022-07-04
Add autocorrect support for `RSpec/EmptyExampleGroup`

r7kamura:
This is an unsafe cop, but I think it could be useful when we wany to automatically remove unnecessary test code.

I have used this feature in our own application and found it so useful that I decided to submit this pull request.

---

* [x] Added tests.
* [x] Updated documentation.

### DIFF
--- a/lib/rubocop/cop/rspec/empty_example_group.rb
+++ b/lib/rubocop/cop/rspec/empty_example_group.rb
@@ -36,6 +36,10 @@ module RuboCop
       #     pending 'will add tests later'
       #   end
       class EmptyExampleGroup < Base
+        extend AutoCorrector
+
+        include RangeHelp
+
         MSG = 'Empty example group detected.'
 
         # @!method example_group_body(node)
@@ -135,7 +139,11 @@ module RuboCop
           return if node.each_ancestor(:block).any? { |block| example?(block) }
 
           example_group_body(node) do |body|
-            add_offense(node.send_node) if offensive?(body)
+            next unless offensive?(body)
+
+            add_offense(node.send_node) do |corrector|
+              corrector.remove(removed_range(node))
+            end
           end
         end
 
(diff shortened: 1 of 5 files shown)


lib/rubocop/cop/rspec/empty_example_group.rb
```diff
@@ -135,7 +137,11 @@ def on_block(node)
           return if node.each_ancestor(:block).any? { |block| example?(block) }
 
           example_group_body(node) do |body|
-            add_offense(node.send_node) if offensive?(body)
+            next unless offensive?(body)
+
+            add_offense(node.send_node) do |corrector|
+              corrector.remove(node)
+            end
```
r7kamura, line 144: The implementation is so simple: just delete the entire ExampleGroup.

spec/rubocop/cop/rspec/empty_example_group_spec.rb
```diff
@@ -19,6 +19,20 @@
         it { should be_true }
       end
     RUBY
+
+    expect_correction(<<~RUBY)
+      describe Foo do
+      #{'  '}
```
pirj, line 25: Is this the leftover indentation whitespace?
Would `range_by_whole_lines` help removing the empty group along with its lines, and potentially with a following/leading separator newline?

r7kamura: Oh, that's a nice feature! I used it right away and it looks very good.
I added https://github.com/rubocop/rubocop-rspec/commit/bd0f3bce51cbe65b2d93da7fb23a55fa64033e84 (then rebased 2 commits & force-pushed).

pirj (approved):
Fantastic, thank you!

bquorning (approved):

pirj:
Thank you, @r7kamura !

Issue #208, opened 2016-09-12
Cop idea: add a rule when there is `it` without any expectations


yuriidanyliak:
Rubocop complains when there is too many expectations but says nothing when we have `it` with some code not including any `expect`s. Maybe there is some sense to implement such rule?

yuriidanyliak:
By the way, I can create corresponding PR, if there is some sense to add this rule.

bquorning:
I think such a cop makes perfect sense. Just remember to allow these:

 1. `it` without any block.
 2. `it` with a call to `skip`.
 3. `it` with a call to `pending`.

See http://www.relishapp.com/rspec/rspec-core/v/3-5/docs/pending-and-skipped-examples/pending-examples and http://www.relishapp.com/rspec/rspec-core/v/3-5/docs/pending-and-skipped-examples/skip-examples

backus:
If this cop did exist it would have to be disabled by default. For example, maybe 80% of the specs for rubocop-rspec use this `expect_violation` helper method I added which dramatically simplifies things. Seems like this cop would flag these

bquorning:
What if the cop is configurable with a list of methods? E.g. `expext`, `skip`, `pending`, `expect_violation`…

backus:
That would make it much more manageable I think. Checkout the `CustomInclude` method that `EmptyExampleGroup` uses: https://github.com/backus/rubocop-rspec/blob/master/lib/rubocop/cop/rspec/empty_example_group.rb#L76

bquorning:
@yuriidanyliak Do you still want to take a stab at implementing this cop?

yuriidanyliak:
@bquorning yeah I'll try to implement such cop.

yuriidanyliak:
Just wrote simplest version of cop. Going to make it more flexible and create a PR.

pirj:
There are cases when this will result in false positives:
```
def expect_processed_value
  expect(process(value))
end

def surprise_the_crowd
  eq(100)
end

it 'is hard to detect an expectation here' do
  expect_processed_value.to surprise_the_crowd
end
```

dgollahon:
As @backus pointed out, we would probably need to disable by default--there is a lot of room for false positives. A configurable whitelist probably covers most cases though.

PR #1321, opened 2022-07-08
Add `RSpec/NoExpectationExample`

### LINKED ISSUE #1319: Cop idea: `NoExpectations`
When I thought about it, I actually was surprised nobody came with it yet :)
So, basically it wold be cool (I'd do it myself, but don't have enough time atm) to have a cop checking whether an example has no expectations at all. It is actually more frequent problem than one might imagine. Some examples might include:
* lazy "it just works" check (`it { subject }` -- it should be at least `expected.to_not raise_error`)
* homemade matcher-less check, e.g. example code which just `raise Something` to signal fail (or worse `puts "something"`) instead of choosing proper matcher
* just accidental copy-paste or rewrite errors, leaving expectations out of the code, or commenting them
* the one where I came with idea: we had a code looking like this
```ruby
it {
  expect(something).to receive(:method).and_return(value)
  subject 
}
```
rspec-rubocop was unhappy (because you shouldn't mix mock and expectation), and suggested to use `allow`. The code author changed it to `allow`... And everything was green, and it is not immediately obvious that the code tests nothing, "it was just small stylistic check suggested by rubocop".

As there is already `MultipleExpectations` cop, it seems that the opposite check wouldn't be too hard to implement, right?..

r7kamura:
This cop checks that each example has any expectation.

fixes #1319

## Example

```ruby
# bad
it do
  a?
end

# good
it do
  expect(a?).to be_truthy
end
```

## Background

Trying it out on my Rails project, I found the following bad cases by this Cop:

```ruby
it do
end

it 'succeeds' do
  subject
end

xit 'returns 200' do
end
```

then I corrected the code like this:

```diff
- it do
- end

it 'succeeds' do
-   subject
+   expect { subject }.not_to raise_error
end

- xit 'returns 200' do
- end
+ xit 'returns 200'
```

---

* [x] Added tests.
* [x] Updated documentation.


r7kamura:
Sorry I haven't read the related issues at all. I see that we had the same discussion a few days ago and even 10 years ago. It's just a coincidence that I created this pull request today.

- https://github.com/rspec/rspec-core/issues/759
- https://github.com/rubocop/rubocop-rspec/issues/208
- https://github.com/rubocop/rubocop-rspec/issues/1319

lib/rubocop/cop/rspec/no_expectation_example.rb
```diff
@@ -0,0 +1,144 @@
...
+          cop_config.fetch('ExampleMethodNames', []).map(&:to_sym).to_set
+        end
+
+        # @param [RuboCop::AST::Node] node
+        # @return [Boolean]
+        def expectation_method_call?(node)
+          return false unless node.is_a?(::RuboCop::AST::SendNode)
+
+          expectation_method_names.include?(node.method_name)
+        end
```
r7kamura, line 106: I've implemented it so that only configured methods are considered as expectation, but it might be better to put in a mechanism to judge like `method_name.start_with?('expect')`.

This would create a naming convention for the `expect` wrapper methods as discussed at https://github.com/rubocop/rubocop-rspec/issues/208#issuecomment-497893002.

> There are cases when this will result in false positives:
> 
> ```
> def expect_processed_value
>   expect(process(value))
> end
> 
> def surprise_the_crowd
>   eq(100)
> end
> 
> it 'is hard to detect an expectation here' do
>   expect_processed_value.to surprise_the_crowd
> end
> ```

lib/rubocop/cop/rspec/no_expectation_example.rb
```diff
@@ -0,0 +1,144 @@
...
+          end
+        end
+
+        private
+
+        # @param [RuboCop::AST::Node] node
+        # @return [RuboCop::AST::BlcokNode, nil]
+        def argument_block_of(node)
+          block = node.parent
+          if block.is_a?(::RuboCop::AST::BlockNode) && block.send_node == node
```
koic, line 84: It looks like it will be replaced with `block_type?` instead of a generic method like `is_a?`.

```suggestion
          if block.block_type? && block.send_node == node
```

lib/rubocop/cop/rspec/no_expectation_example.rb
```diff
@@ -0,0 +1,144 @@
...
+
+        # @return [Set<Symbol>]
+        def example_method_names_from_config
+          cop_config.fetch('ExampleMethodNames', []).map(&:to_sym).to_set
+        end
+
+        # @param [RuboCop::AST::Node] node
+        # @return [Boolean]
+        def expectation_method_call?(node)
+          return false unless node.is_a?(::RuboCop::AST::SendNode)
```
koic, line 103: Ditto.

```suggestion
          return false unless node.send_type?
```

config/default.yml
```diff
@@ -614,6 +614,15 @@ RSpec/NestedGroups:
   VersionChanged: '1.10'
   Reference: https://www.rubydoc.info/gems/rubocop-rspec/RuboCop/Cop/RSpec/NestedGroups
 
+RSpec/NoExpectationExample:
+  Description: Checks if an example includes any expectation.
+  Enabled: pending
+  ExampleMethodNames: []
```
pirj, line 620: We have a "global" setting for this https://github.com/rubocop/rubocop-rspec/blame/master/config/default.yml#L36

lib/rubocop/cop/rspec/no_expectation_example.rb
```diff
@@ -0,0 +1,144 @@
...
+        ]
+
+        MSG = 'No expectation found in this example.'
+
+        # @param [RuboCop::AST::Node] node
+        def on_send(node)
+          return unless example_method_names.include?(node.method_name)
+
+          block = argument_block_of(node)
+          return unless block
```
Darhazer, line 70: it seems like this would be better implement in an `on_block` and then check if the send_note of that block is an example ?

lib/rubocop/cop/rspec/no_expectation_example.rb
```diff
@@ -0,0 +1,144 @@
...
+      #   # good
+      #   it do
+      #     expect(a?).to be_truthy
+      #   end
+      class NoExpectationExample < Base
+        extend AutoCorrector
+
+        include RangeHelp
+
+        DEFAULT_EXAMPLE_METHOD_NAMES = ::Set[
```
Darhazer, line 34: This should be already defined in the `Language` module. Perhaps check out the MultipleExpectation cop for an example how to count the expectations in an example

lib/rubocop/cop/rspec/no_expectation_example.rb
Darhazer: Note that rubocop-rspec already provides a way to add methods to be considered as expectation, by listing them under Language/Expectations config

r7kamura:
Thanks for the review comments! I learned a lot of information I didn't know 👍 
I have made some improvements and hope you will take another look.

lib/rubocop/cop/rspec/no_expectation_example.rb
```diff
@@ -0,0 +1,90 @@
...
+      # @example
+      #
+      #   # bad
+      #   it do
+      #     a?
+      #   end
+      #
+      #   # good
+      #   it do
+      #     expect(a?).to be_truthy
```
koic, line 29: That just my two cents. Method whose name ends with `?`, generally return boolean type. So, `be true` would be more suitable as an example than `be_truthy`.

```suggestion
      #     expect(a?).to be(true)
```

r7kamura: Yeah, I used to think so too (`a?` should always return true or false), but Rails, for example, is known not to have such policy.

- https://github.com/rails/rails/pull/19424
    - Example of `ActionDispatch::Request#xhr?` returning Integer

koic: For https://github.com/rails/rails/pull/19424, the API is already widespread, so maybe a trade-off with breaking changes may have been considered. Proposal that will change return value of published API will be cautious.

r7kamura: Yes, that's true, and there's no reason not to include this suggested change. I'll add this change, thanks 👍

.rubocop.yml
```diff
@@ -81,6 +88,12 @@ RSpec/DescribeClass:
   Exclude:
     - spec/project/**/*.rb
 
+RSpec/MultipleExpectations:
+  Max: 2
```
pirj, line 92: Why?

lib/rubocop/cop/rspec/no_expectation_example.rb
```diff
@@ -0,0 +1,90 @@
...
+        # @return [Boolean]
+        def example_method_call?(node)
+          Examples.all(node.method_name)
+        end
+
+        # Recursively checks if the given node includes any expectation.
+        # @param [RuboCop::AST::Node] node
+        # @return [Boolean]
+        def including_any_expectation?(node)
+          if !node.is_a?(::RuboCop::AST::Node)
```
pirj, line 68: What else can it be?

pirj (commented):
I don't feel that this cop should implement autocorrection. WDYT?

.rubocop.yml
r7kamura: A normal unit test has only one expectation,  but a cop test case is special and has one of the following three patterns:

1. `expect_no_offenses(...)`
2. `expect_offense(...)`
3. `expect_offense(...); expect_correction(...)`

To allow for the 3rd pattern, I set this `Max: 2`.

r7kamura:
> I don't feel that this cop should implement autocorrection. WDYT?

I'm on the fence about that.

`RSpec/EmptyExampleGroup`, which has a similar concept, implements autocorrection, so I thought it might be a good idea for this cop to do the same. Sometimes this autocorrection is useful, sometimes it is not. This cop is marked as `Safe: false`, so it will not be unintentionally autocorrected unless users force it. But thoughtlessly erasing the code is indeed dangerous...

I actually used this cop's autocorrection in my Rails app because I had to change a large amount of code there, but I am thinking that maybe this autocorrection should be removed.

.rubocop.yml
pirj: [RSpec style guide's guideline regarding single expectation per example](﻿﻿﻿https://rspec.rubystyle.guide/#expectation-per-example) is less strict.
Since this is not directly related to introducing this new cop, I suggest extracting this into a separate PR. And I'm still uncertain if we'd really like to constrain ourselves to the ~one~ two expectations style.

pirj:
> `RSpec/EmptyExampleGroup`, which has a similar concept, implements autocorrection, so I thought it might be a good idea for this cop to do the same

I think an empty example group has more chances to be a deliberate choice than the no expectations one. The latter is most likely a mistake.
For me, removing the empty example group doesn't seem to be an ideal autocorrection. But for the no expectations, it isn't for sure. Taking example from the [`NoExpectations` ticket](https://github.com/rubocop/rubocop-rspec/issues/1319):
```ruby
it { subject } # Just checks if calling `subject` doesn't raise any exceptions
```
```ruby
it { validates_presence_of(:attribute) } # Oops, missing `is_expected.to`
```
It would be quite odd to remove those examples.


> This cop is marked as Safe: false

I have to agree with this, unfortunately, with the `Safe` definition as "indicates whether the cop can yield false positives (by design) or not". It is unfortunate, though, that this cop would in most cases won't be run, and won't detect legitimate offences. Sadly, I don't have an idea how to overcome this.

> that maybe this autocorrection should be removed

I fully support that.


Let me quickly run the cop against several codebases in [`real-world-rspec`](https://github.com/pirj/real-world-rspec) to see how severe and frequent false positives are.

.rubocop.yml
Darhazer: I think that the point is:
This PR introduces a cop that would complain about our own specs
So it also adds our expectations helper to the Language configuration
Previously MultipleExpectations would count expectations per example as 0, as it was not aware of those helpers. 
Now that it’s aware, it’s triggering an offence 
Hence also the tweak in configuration

pirj: Makes total sense, thanks.

pirj:
> 57497 files inspected, 24233 offenses detected

I briefly went through, and tried categorizing cases that look like false positive/suspicious:

1. Legacy RSpec expectation notation:
```ruby
# /Users/pirj/source/real-world-rspec/loomio/spec/models/user_spec.rb
    it "returns the stored name if deactivated_at is nil" do
      user.name.should_not == 'Deleted user'
    end
```
```ruby
# /Users/pirj/source/real-world-rspec/rspec-mocks/spec/rspec/mocks/configuration_spec.rb
          it "does not warn about the should syntax" do
            RSpec.should_not_receive(:deprecate)
            Object.new.should_not_receive(:bees)
          end
```

2. Might be legitimate offences
```ruby
# /Users/pirj/source/real-world-rspec/mastodon/spec/controllers/auth/registrations_controller_spec.rb
    it 'does nothing if user already exists' do
      Fabricate(:account, username: 'test')
      subject # Frankly, I'm struggling to understand where `subject` is defined
    end
```

```ruby
# /Users/pirj/source/real-world-rspec/refinerycms/pages/spec/models/refinery/page_meta_data_spec.rb
      it 'meta_description' do
        page.respond_to?(:meta_description) # should be `expect(page).to respond_to(:meta_description)` ?
      end
```

3. Expectation helper method
```ruby
    it 'tracks when sign in is nil' do
      user.update(current_sign_in_at: nil)
      sign_in user, scope: :user
      get :show

      expect_updated_sign_in_at(user)
    end
```

4. Skipped examples
```ruby
      it "updates the requested expense" do
        expense = Expense.create! valid_attributes
        put :update, {:id => expense.to_param, :expense => new_attributes}, valid_session
        expense.reload
        skip("Add assertions for updated state")
      end
```

5. Rails/Minitest-style assertions
```ruby
# /Users/pirj/source/real-world-rspec/publify/publify_core/spec/controllers/admin/articles_controller_spec.rb
      it "sends notifications on create" do
        u = create(:user, notify_via_email: true, notify_on_new_articles: true)
        u.save!
        ActionMailer::Base.deliveries.clear
        emails = ActionMailer::Base.deliveries

        post :create, params: { "article" => base_article }

        assert_equal(1, emails.size)
        assert_equal(u.email, emails.first.to[0])
      end
```

6. Alternative mocking library (e.g. Mocha)
```ruby
# /Users/pirj/source/real-world-rspec/puppet/acceptance/lib/puppet/acceptance/classifier_utils_spec.rb
  it "provides a handle to the classifier service" do
    handle.expects(:perform_request).with(Net::HTTP::Get, '/hi', {})
    handle.get('/hi')
  end
```

7. No explicit `expect { ... }.not_to raise_error`
```ruby
# /Users/pirj/source/real-world-rspec/rspec-core/spec/rspec/core/backtrace_formatter_spec.rb
      it "deals gracefully with a security error" do
        Metadata.instance_eval { @relative_path_regex = nil }
        with_safe_set_to_level_that_triggers_security_errors do
          self.formatter.__send__(:backtrace_line, __FILE__)
          # on some rubies, this doesn't raise a SecurityError; this test just
          # assures that if it *does* raise an error, the error is caught inside
        end
      end
```

8. Esoteric - definition of examples inside examples
```ruby
# /Users/pirj/source/real-world-rspec/rspec-core/spec/rspec/core/configuration_spec.rb
      it 'successfully invokes the block' do
        RSpec.describe("group") { it "example 1" do; end}
        example = RSpec.world.example_groups.first.examples.first
        expect(example.metadata[:new_key]).to eq(:new_value)
      end
```

9. Missing configuration for well-known expectations
```ruby
    it 'does not register any offenses' do
      expect_no_offenses(<<~RUBY, 'foo.rb')
        # cop will not read these contents
        gem('rubocop')
        gem('rubocop')
      RUBY
    end
```
Can we put this:
```yaml
RSpec:
  Language:
    Expectations:
      - expect_correction
      - expect_no_offenses
      - expect_offense
```
into some additional YAML file so `rubocop` and other extensions could just include this, and this would be used in both this cop and `MultipleExpectations`?

It looks like an immense effort to fix this all. With `Safe: false`, I believe it's ok to keep the cop as it is.
I wouldn't even care about the long-deprecated monkey-patching `object.should` RSpec syntax and Minitest-style assertions.

lib/rubocop/cop/rspec/no_expectation_example.rb
r7kamura: For example, if the parent node is `s(:send, nil, :it)`, its children are `[nil, :it]`, so it can be nil or a Symbol.

lib/rubocop/cop/rspec/no_expectation_example.rb
```diff
@@ -0,0 +1,60 @@
...
+        # @param [RuboCop::AST::Node] node
+        # @return [Boolean]
+        def_node_search(
+          :including_any_expectation?,
+          send_pattern('#Expectations.all')
+        )
+
+        # @param [RuboCop::AST::BlockNode] node
+        def on_block(node)
+          return unless example_method_call?(node)
```
pirj, line 44: I think we have an [`example?` helper method](https://github.com/rubocop/rubocop-rspec/blame/master/lib/rubocop/rspec/language.rb#L41). Would it work here?

pirj (approved):
Looks good, thanks!

spec/rubocop/cop/rspec/no_expectation_example_spec.rb
```diff
@@ -0,0 +1,101 @@
...
+        ^^^^^^^^^^^^^^^^^ No expectation found in this example.
+      RUBY
+    end
+  end
+
+  context 'with block-less example in block' do
+    it 'registers no offenses' do
+      expect_no_offenses(<<~RUBY)
+        RSpec.describe Foo do
+          pending 'not implemented'
```
Darhazer, line 96: perhaps we should skip checking of pending or skipped examples in the first place.

config/default.yml
```diff
@@ -61,9 +61,14 @@ RSpec:
       Pending:
         - pending
     Expectations:
+      - are_expected
```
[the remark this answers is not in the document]
bquorning: I haven’t personally used the ‘should’ syntax since RSpec 3 was released, but from the RSpec documentation it sounds like it’s not being removed, even in v4. So I guess we should support it?

@pirj I imagine you might know about official RSpec plans?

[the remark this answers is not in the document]
pirj: All forms of `should` [will be removed from RSpec 4](https://github.com/rspec/rspec-core/pull/2803) with no replacement.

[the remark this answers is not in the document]
pirj: Personally, I would not keep the burden of supporting `should`, as it was [soft-deprecated nine years ago and was disabled in the generated config since then](http://rspec.info/blog/2013/07/the-plan-for-rspec-3/#zero_monkey_patching_mode):
> For RSpec 3, we considered the idea of disabling the old syntax by default, forcing users to opt-in to use it.

But I still see people use it, even those who joined software development less than nine years ago :smile:

Darhazer (approved):

pirj:
Thank you, @r7kamura !

PR #1322, opened 2022-07-08
Add an RSpec 4 pre-release CI job

pirj:
RSpec 4 will eventually be released. Since we're checking RSpec style, why don't we use RSpec 4 at least in an optional job to help test it out?

Below what had to be done.

### Move dev dependencies to Gemfile

A similar (unmerged) `rspec-support` PR https://github.com/rspec/rspec-support/pull/517

Reasoning (https://bundler.io/v2.2/man/gemfile.5.html#GEMSPEC):

> The .gemspec file is ... where you specify the dependencies your gem needs *to run*.

Even though it contradicts the very existence of add_development_dependency.

Discussions on this topic:
- https://github.com/rubygems/rubygems/discussions/5065
- https://github.com/rubygems/rubygems/issues/1104

### Refactored our CI workflow

Extracted `edge` RuboCop in separate job(s).

### RSpec 4

It's complicated and not an obvious situation with shared contexts.
Previously, they were auto-included using matching metadata. Now, only explicitly (see changes to the `spec/spec_helper.rb`). This revealed two issues:

1. Those shared examples that define metadata themselves, e.g. `shared_examples 'config', :config do` would be included first. Since we're overriding the one defined in `rubocop` with ours, to set `other_cops` with a proper RSpec DSL configuration, there is only one way of doing that in a non-Yoda order as we have to do now:
```ruby
  config.include_context 'with default RSpec/Language config', :config # Intuituvely, this should go below
  config.include_context 'config', :config
```
is to remove `:config` from the definition in `rubocop`.
But this makes it complicated, since well-known extensions depend on this metadata, and custom cops' specs, too.
This needs to be somehow fixed on RSpec 4 side.

2. There is no straightforward way now to define a satellite example group for specs, like e.g. 'smoke test' one. There is no such thing as `config.include_examples`. However, `config.include_context` exists, and it's an alias that has different semantics but identical behaviour.

### DIFF
--- a/.rspec
+++ b/.rspec
@@ -1,3 +1,2 @@
 --require spec_helper
---color
 --format documentation
--- a/Gemfile
+++ b/Gemfile
@@ -6,7 +6,7 @@ gemspec
 
 gem 'rack'
 gem 'rake'
-gem 'rspec', '>= 3.4'
+gem 'rspec', '~> 3.11'
 gem 'rubocop-performance', '~> 1.7'
 gem 'rubocop-rake', '~> 0.6'
 gem 'yard'
--- a/Gemfile
+++ b/Gemfile
@@ -4,6 +4,13 @@ source 'https://rubygems.org'
 
 gemspec
 
+gem 'rack'
+gem 'rake'
+gem 'rspec', '>= 3.4'
+gem 'rubocop-performance', '~> 1.7'
+gem 'rubocop-rake', '~> 0.6'
+gem 'yard'
+
 local_gemfile = 'Gemfile.local'
 
 if File.exist?(local_gemfile)
--- a/spec/smoke_tests/weird_rspec_spec.rb
+++ b/spec/smoke_tests/weird_rspec_spec.rb
@@ -86,7 +86,7 @@ RSpec.describe 'Weirdness' do
   end
 
   it_behaves_like :something
-  it_should_behave_like :something
+  it_should_behave_like :something if RSpec::Core::Version::STRING < '4.0'
 
   it_behaves_like :something do
     let(:foo) { 'bar' }
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -24,11 +24,7 @@ RSpec.configure do |config|
   config.order = :random
 
   # Run focused tests with `fdescribe`, `fit`, `:focus` etc.
-  config.filter_run focus: true
-  config.run_all_when_everything_filtered = true
-
-  # Forbid RSpec from monkey patching any of our objects
-  config.disable_monkey_patching!
+  config.filter_run_when_matching :focus
 
   # We should address configuration warnings when we upgrade
   config.raise_errors_for_deprecations!
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -38,6 +38,10 @@ RSpec.configure do |config|
   config.raise_on_warning = true
 
   config.include(ExpectOffense)
+
+  config.include_context 'with default RSpec/Language config', :config
+  config.include_context 'config', :config
+  config.include_context 'smoke test', type: :cop_spec
 end
 
 $LOAD_PATH.unshift(File.join(File.dirname(__FILE__), '..', 'lib'))
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -61,3 +61,24 @@ jobs:
           ruby-version: "${{ matrix.ruby }}"
           bundler-cache: true
       - run: bundle exec rake ${{ matrix.task }}
+
+  rspec4:
+    runs-on: ubuntu-latest
+    name: RSpec 4
+    steps:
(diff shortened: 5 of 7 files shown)


spec/smoke_tests/weird_rspec_spec.rb
```diff
@@ -86,7 +86,7 @@
   end
 
   it_behaves_like :something
-  it_should_behave_like :something
+  # it_should_behave_like :something
```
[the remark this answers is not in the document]
Darhazer: To make sure rubocop-rspec doesn't crash when inspecting specs

pirj:
Strangely, it passes locally just fine.

Gemfile
```diff
@@ -4,6 +4,15 @@ source 'https://rubygems.org'
 
 gemspec
 
+gem 'rack'
+gem 'rake'
+%w[rspec rspec-core rspec-expectations rspec-mocks rspec-support].each do |repo|
```
bquorning, line 9: Could we put this into a Gemfile.local on CI? Like how we test against RuboCop master branch.

spec/smoke_tests/weird_rspec_spec.rb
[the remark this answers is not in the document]
pirj: But how?
Those specs are run by `rspec` as part of the suite.
But we explicitly exclude them from the inspection:
```
AllCops:
  DisplayCopNames: true
  TargetRubyVersion: 2.6
  NewCops: disable
  Exclude:
    - 'vendor/**/*'
    - 'spec/fixtures/**/*'
    - 'tmp/**/*'
    - 'spec/smoke_tests/**/*.rb'
```

If some of our cops wouldn't be capable of inspecting any of those, how would we know?
cc @bquorning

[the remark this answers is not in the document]
Darhazer: We exclude them from rubocop inspection, e.g. so rake internal_investigation doesn't fail, but we have an explicit spec that says
```
expect { inspect_source(source, file_name) }.not_to raise_error
expect { autocorrect_source(source, file_name) }.not_to raise_error
```
on those files, so it checks if trying to inspect or correct those will raise an error, without caring of specific offenses found there

Gemfile
pirj: Done 👍

spec/smoke_tests/weird_rspec_spec.rb
```diff
@@ -86,7 +86,7 @@
   end
 
   it_behaves_like :something
-  it_should_behave_like :something
+  it_should_behave_like :something if RSpec::Core::Version::STRING < '4.0'
```
pirj, line 89: As we not only parse those weird specs, but also actually run them with `rake spec`, we have to conditionally call `it_should_behave_like` as this alias has been removed from RSpec 4.

bquorning (approved):
Fine by me.

Perhaps add a link to the RSpec 4 changelogs?

pirj:
>  add a link to the RSpec 4 changelogs?

@bquorning Do you mean in code comments somewhere? I've added them to the [commit message](https://github.com/rubocop/rubocop-rspec/pull/1322/commits/15b1a9a580dad6098f39bab9b3939aaf94fcdc0f).

.rspec
```diff
@@ -1,3 +1,2 @@
 --require spec_helper
---color
```
pirj, line 2: [color](https://github.com/rspec/rspec-core/commit/04078317ba6577699d06cf4dccf014254dcde7a6)

spec/spec_helper.rb
```diff
@@ -24,11 +24,7 @@ module SpecHelper
   config.order = :random
 
   # Run focused tests with `fdescribe`, `fit`, `:focus` etc.
-  config.filter_run focus: true
-  config.run_all_when_everything_filtered = true
```
pirj, line 28: https://github.com/rspec/rspec-core/pull/2845

spec/spec_helper.rb
```diff
@@ -24,11 +24,7 @@ module SpecHelper
...
-
-  # Forbid RSpec from monkey patching any of our objects
-  config.disable_monkey_patching!
```
pirj, line 31: https://github.com/rspec/rspec-core/pull/2803

spec/spec_helper.rb
```diff
@@ -38,6 +34,10 @@ module SpecHelper
   config.raise_on_warning = true
 
   config.include(ExpectOffense)
+
+  config.include_context 'with default RSpec/Language config', :config
+  config.include_context 'config', :config
```
pirj, line 39: After [this PR](https://github.com/rubocop/rubocop/pull/10806/files#diff-a328b4f030d5897fe22c01bb96a3bfbcc8bcc6a11f936d1e7a2db73a721d0eb8R52) is merged (and released, and we bump our `rubocop` dependency), this line becomes redundant.

PR #1323, opened 2022-07-08
Fix FilePath cop missing expanded namespaces

sl4vr:
Fixes https://github.com/rubocop/rubocop-rspec/issues/1107.
Tests taken from https://github.com/rubocop/rubocop-rspec/pull/1287.

In order to handle expanded namespaces, `namespace` helper method from `RSpec/DescribedClass` cop is used and moved into standalone mixin.

I haven't dug into it, but at first glance it seems that [parent_module_name from rubocop-ast](https://github.com/rubocop-hq/rubocop-ast/blob/e2df776f6aa4133413e2de0e20b4a3295b12988f/lib/rubocop/ast/node.rb#L333) returns namespace of concrete node while `namespace` helper method just looks or all ancestor nodes.
AFAIU `parent_module_name` works fine with class definitions, but to get proper namespace for `describe MyClass` we have to call it on `s(:block, s(:send, nil, :describe, s(:const, nil, :MyClass)), s(:args), nil)` node and not on `s(:const, nil, :MyClass)` node. Which is doable, but it's way more easier to use `namespace` helper method from `RSpec/DescribedClass` cop. Which in turn raises the question: is https://github.com/rubocop/rubocop-rspec/issues/984 really actual?

---

* [x] Added tests.
* [x] Updated documentation.

### DIFF
--- a/spec/rubocop/cop/rspec/file_path_spec.rb
+++ b/spec/rubocop/cop/rspec/file_path_spec.rb
@@ -213,10 +213,21 @@ RSpec.describe RuboCop::Cop::RSpec::FilePath do
     RUBY
   end
 
-  it 'registers an offense for path based on a class name with long module' do
+  it 'registers an offense for path with incorrect collapsed namespace' do
     expect_offense(<<-RUBY, '/home/foo/spec/very/my_class_spec.rb')
       describe Very::Long::Namespace::MyClass do; end
       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ Spec path should end with `very/long/namespace/my_class*_spec.rb`.
+    RUBY
+  end
+
+  it 'registers an offense for path with incorrect expanded namespace' do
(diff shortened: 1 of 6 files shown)


spec/rubocop/cop/rspec/file_path_spec.rb
```diff
@@ -220,6 +220,24 @@
     RUBY
   end
 
+  it 'registers an offense for path with incorrect collapsed namespace' do
```
pirj, line 223: Isn't this the same as the example above?

pirj (approved):
Looks great, thank you!
And welcome back 👋

spec/rubocop/cop/rspec/file_path_spec.rb
sl4vr: Of course I just missed that 😅 
Let's keep this since it has moar constants 🙂

bquorning (approved):

PR #1324, opened 2022-07-11
Fix RSpec/ChangeByZero with compound operators

BrianHawley:
Previously, it didn't recognize compound expressions using `&` or `|`.
This led it to correct part of the expressions as regular code, but the
generated code was malformed.

Now it recognizes compound expressions using the operators.

---

* [x] Added tests.
* [ ] Updated documentation.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 * Fix `RSpec/FilePath` cop missing mismatched expanded namespace. ([@sl4vr][])
 * Add new `AllowConsecutiveOneLiners` (default true) option for `Rspec/EmptyLineAfterHook` cop. ([@ngouy][])
 * Add autocorrect support for `RSpec/EmptyExampleGroup`. ([@r7kamura][])
+* Fix `RSpec/ChangeByZero` with compound expressions using `&` or `|` operators. ([@BrianHawley][])
 
 ## 2.12.1 (2022-07-03)
 
@@ -718,3 +719,4 @@ Compatibility release so users can upgrade RuboCop to 0.51.0. No new features.
 [@edgibbs]: https://github.com/edgibbs
 [@Drowze]: https://github.com/Drowze
 [@akiomik]: https://github.com/akiomik
+[@BrianHawley]: https://github.com/BrianHawley
--- a/lib/rubocop/cop/rspec/change_by_zero.rb
+++ b/lib/rubocop/cop/rspec/change_by_zero.rb
@@ -79,7 +79,7 @@ module RuboCop
         end
 
         def compound_expectations?(node)
-          %i[and or].include?(node.parent.method_name)
+          %i[and or & |].include?(node.parent.method_name)
         end
 
         def autocorrect(corrector, node)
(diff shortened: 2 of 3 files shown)


BrianHawley:
@pirj you reviewed and merged the previous PRs for this cop. Care to take a look at this change?

pirj (approved):
Fantastic, thanks!

BrianHawley:
I just thought of another bad correction. Give me a second.

pirj (changes requested):
RSpec wouldn't allow `.not_to change { }.by(...)` to my best knowledge.

BrianHawley:
@pirj you're right, it looks like rspec complains about that now. I remember it allowing that back in the day, so it's good to see it complaining now. This gem probably doesn't support rspec versions that old, so I'll remove it.

BrianHawley:
@pirj see https://github.com/rspec/rspec-expectations/blob/main/Changelog.md#300beta2--2014-02-17 for details.

BrianHawley:
@pirj do you still want me to undo the `not_to` and `to_not` support? As it is, with that support, it changes code that rspec complains about to semantically equivalent code that it wouldn't complain about. It might help those unfortunates who still have to do rspec 3 updates in this day and age.

pirj:
> if rubocop-rspec supports rspec 2.x or earlier

It's a good question. I would say we don't, as **probably** those with RSpec 2.x are on early Ruby 2.0/2.1 which we don't support. Might be Ruby version update is an easier task than RSpec 2 -> 3 migration.

But not to be overly defensive with our code, I'd remove this, and wait for someone to complain about this deficiency. We can reconsider if this happens.

BrianHawley:
@pirj I just made it skip the bad correction in that case, so it at least doesn't introduce logic errors.

pirj (approved):
Thanks!

Darhazer:
Wow I haven't seens specs using that syntax. Do we need a cop to force using `and` and `or` over `&` and `|`

pirj:
I don’t know.
Those are aliases to or and and.
I’d check existing codebases if they make use of them.
Some may prefer it, and I personally kind of like how it looks like.

Darhazer:
> Some may prefer it, and I personally kind of like how it looks like.

That's why the configurable enforced styles are ;)

BrianHawley:
The `&` and `|` operators are better to use sometimes because their precedence is a better match for the circumstances, and they don't get caught up in enforced parentheses like `and` and `or` are (as I am subjected to at work). If you make a cop to prevent their use, make sure that there's an option to enforce their use, or at least to disable it.

pirj:
My apologies to let the discussion drift. @BrianHawley you are really welcome to open an issue or a PR for the new cop, and to add some reasoning. I'll do my best to find some proof of usage of `&`/`|` in the wild. I confess knowing about this syntax, and being frustrated with `and`/`or` at times, and never giving `&`/`|` a shot. Might be the cop, even with the opposite enforced style could spread the word about such an alternative.

pirj:
Thank you, @BrianHawley !

PR #1325, opened 2022-07-17
Add new `RSpec/SortMetadata` cop

leoarnold:
This cop sorts the RSpec metadata alphabetically:

```ruby
  # bad
  describe 'Something', :b, :a
  context 'Something', foo: 'bar', baz: true
  it 'works', :b, :a, foo: 'bar', baz: true

  # good
  describe 'Something', :a, :b
  context 'Something', baz: true, foo: 'bar'
  it 'works', :a, :b, baz: true, foo: 'bar'
```

---

* [x] Added tests.
* [x] Updated documentation.


lib/rubocop/cop/rspec/sort_tags.rb
```diff
@@ -0,0 +1,333 @@
...
+      #   describe 'Something', :b, :a
+      #   context 'Something', foo: 'bar', baz: true
+      #   it 'works', :b, :a, foo: 'bar', baz: true
+      #
+      #   # good
+      #   describe 'Something', :a, :b
+      #   context 'Something', :baz, foo: 'bar'
+      #   it 'works', :a, :b, :baz, foo: 'bar'
+      #
+      class SortTags < Base
```
pirj, line 41: The first thing that strikes me is the term "tags". RSpec usually refers to it as "metadata". There are mentions of "tag", too, and there's even a [doc page named after "tag"](https://relishapp.com/rspec/rspec-core/v/3-11/docs/command-line/tag-option), I believe ["metadata"](https://relishapp.com/rspec/rspec-core/v/3-11/docs/metadata/user-defined-metadata) is used more often.

lib/rubocop/cop/rspec/sort_tags.rb
```diff
@@ -0,0 +1,333 @@
...
+          skip
+          specify
+          xcontext
+          xdescribe
+          xexample
+          xfeature
+          xit
+          xscenario
+          xspecify
+        ].to_set.freeze
```
pirj, line 94: Wondering if this would work if someone decides to configure `RSpec/Language` and add e.g. an alias to an example,  e.g. `this { fails }`.

lib/rubocop/cop/rspec/sort_tags.rb
```diff
@@ -0,0 +1,333 @@
...
+              pairs
+            end
+          end
+        end
+
+        private_constant :CrimeSceneCleaner, :CrimeSceneInvestigator, :Deputy
+      end
+    end
+  end
+end
```
pirj, line 333: Even though topics of sorting and converting from symbols to hash are closely related, would you agree that considering the length of this, it would better be two separate cops?

I believe there's a metric that limits the cop's LoC to 100 lines. Somehow you've managed to trick it, perhaps Deputies are the reason. Still, it's quite large to comprehend quickly.

spec/rubocop/cop/rspec/sort_tags_spec.rb
```diff
@@ -0,0 +1,279 @@
+# frozen_string_literal: true
+
+RSpec.describe RuboCop::Cop::RSpec::SortTags do
+  it 'does not register an offense when using only symbol tags ' \
+     'in alphabetical order' do
+    expect_no_offenses(<<~RUBY)
+      RSpec.describe 'Something', :a, :b do
```
pirj, line 7: Do you mind adding a couple of weird examples, e.g. multi-line metadata:
```ruby
RSpec.describe 'Something',
               :foo, :bar,
               baz: 'goo' do
```

and some weird values:
```ruby
RSpec.describe 'Something', foo: ->(x) { bar(x) }
```
Would the cop work correctly?

lib/rubocop/cop/rspec/sort_tags.rb
```diff
@@ -0,0 +1,333 @@
...
+      #   describe 'Something', :a, :b
+      #   context 'Something', :baz, foo: 'bar'
+      #   it 'works', :a, :b, :baz, foo: 'bar'
+      #
+      class SortTags < Base
+        extend AutoCorrector
+        include ConfigurableEnforcedStyle
+        include RangeHelp
+
+        HOOKS = %i[
```
Darhazer, line 46: We have those defined in the Language module (and that allows user to extend with their custom macros)

lib/rubocop/cop/rspec/sort_tags.rb
```diff
@@ -0,0 +1,333 @@
...
+          before
+          append_before
+          around
+          prepend_after
+          after
+          append_after
+        ].to_set.freeze
+
+        # Examples, ExampleGroups, SharedGroups, and Hooks
+        RESTRICT_ON_SEND = %i[
```
Darhazer, line 57: Because of the configurable language, it's hard to really restrict the send.

lib/rubocop/cop/rspec/sort_tags.rb
```diff
@@ -0,0 +1,333 @@
...
+        def investigate(tags)
+          return if tags.none?
+
+          investigator = CrimeSceneInvestigator.new(tags, style, self)
+          return unless (yellow_tape = investigator.yellow_tape)
+
+          add_offense(yellow_tape, message: investigator.report) do |corrector|
+            cleaner = CrimeSceneCleaner.new(tags, style)
+            corrector.replace(range(tags), cleaner.cleanup)
+          end
```
Darhazer, line 138: :heart: the naming of those classes

lib/rubocop/cop/rspec/sort_tags.rb
```diff
@@ -0,0 +1,333 @@
...
+            end
+          end
+
+          def include_hash?
+            return false unless @hash_node
+            return !hash_sorted? unless @style == :prefer_symbols
+
+            hash_pairs.any? do |pair_node|
+              pair_node.children.last.true_type?
+            end
```
Darhazer, line 238:
```suggestion
            hash_pairs.any? do |pair_node|
              pair_node.value.true_type?
            end
```
Or even
```suggestion
            hash_pairs.map(&:value).any?(&:true_type?)
```
or if you preserve the origal hash note, and not the pairs
```suggestion
            hash_node.each_value.any?(&:true_type?)
```
This applies to other parts of the code. Use the provided `key` and `value` methods instead of refering first and last children

Darhazer (commented):
I love the idea of this cop :heart: 

I haven't had the time to fully review, but left some initial comments. Most of all, we should strive to support the Language module and it's extensibility and not restrict the cop to the RSpec build-in methods.

spec/rubocop/cop/rspec/sort_tags_spec.rb
leoarnold: In the multi-line case, what would be your expected offense and expected correction?

pirj: There should be no difference. I'm just slightly concerned that multi-line might not work as expected, and if it's a major pain to make it work, it would make sense to just ignore it.

lib/rubocop/cop/rspec/sort_metadata.rb
```diff
@@ -0,0 +1,143 @@
...
+        def on_send(node)
+          rspec_metadata(node) do |symbols, pairs|
+            investigate(symbols, pairs.flatten)
+          end
+        end
+
+        private
+
+        def investigate(symbols, pairs)
+          return if symbols.none? && pairs.none?
```
bquorning, line 107: If `symbols.none? && pairs.none?` then line 108 will also return, right? So perhaps we can just remove line 107 altogether?

lib/rubocop/cop/rspec/sort_tags.rb
leoarnold: I added a spec for configured language. Does that meet your requirements?

pirj:
Speaking of the name, I think we can refer to the upcoming [Cop naming guidelines](https://github.com/rubocop/rubocop/pull/10725/files#diff-d215eaa340849c3e72defb154136bcb8e2238d47e34236eaabd7e09b9b334e1bR40). I'd suggest `RSpec/UnorderedMetadata`.

lib/rubocop/cop/rspec/sort_tags.rb
pirj: Muy bien!

spec/rubocop/cop/rspec/sort_metadata_spec.rb
```diff
@@ -0,0 +1,206 @@
...
+
+  context 'when using custom RSpec language ' \
+          'and adjusting the RuboCop RSpec language configuration' do
+    let(:config) do
+      RuboCop::ConfigLoader.merge_with_default(
+        RuboCop::Config.new(YAML.safe_load(language_config)), ''
+      )
+    end
+
+    let(:language_config) do
```
pirj, line 168: We only check language config in one cop's specs:
```ruby
    before do
      other_cops.dig('RSpec', 'Language', 'Includes', 'Examples')
        .push('it_has_special_behavior')
    end
```
but I love your approach better 👍

lib/rubocop/cop/rspec/sort_metadata.rb
```diff
@@ -0,0 +1,94 @@
...
+        def replacement(symbols, pairs)
+          (sort_symbols(symbols) + sort_pairs(pairs)).map(&:source).join(', ')
+        end
+
+        def sorted?(symbols, pairs)
+          symbols == sort_symbols(symbols) && pairs == sort_pairs(pairs)
+        end
+
+        def sort_pairs(pairs)
+          pairs.sort_by { |pair| pair.key.value.to_s }
```
pirj, line 85:
```
An error occurred while RSpec/SortMetadata cop was inspecting /Users/pirj/source/real-world-rspec/rspec-core/spec/rspec/core/metadata_spec.rb:51:14.
undefined method `value' for s(:lvar, :key):RuboCop::AST::Node
/Users/pirj/source/rubocop-rspec/lib/rubocop/cop/rspec/sort_metadata.rb:85:in `block in sort_pairs'
/Users/pirj/source/rubocop-rspec/lib/rubocop/cop/rspec/sort_metadata.rb:85:in `each'
/Users/pirj/source/rubocop-rspec/lib/rubocop/cop/rspec/sort_metadata.rb:85:in `sort_by'
```

[This line](https://github.com/rspec/rspec-core/blob/814c1c1ecf79d2191718b2795b6d12d9d522f60b/spec/rspec/core/metadata_spec.rb#L51):
```ruby
RSpec.describe("group", key => {})
```

Indeed, both keys and values can be strange up to the extend when they are unsortable.

lib/rubocop/cop/rspec/sort_metadata.rb
```diff
@@ -0,0 +1,94 @@
...
+        end
+
+        def sort_symbols(symbols)
+          symbols.sort_by(&:value)
```
pirj, line 89: To my best knowledge, in theory, keys can be strings, too. And this would fail comparing symbols to strings.
However, this doesn't seem to manifest on several big codebases.

spec/rubocop/cop/rspec/sort_metadata_spec.rb
pirj: I really mean it.
But if you decided to go with `other_cops.dig.push` approach, this definition becomes redundant.

pirj:
Checked against [`real-world-rspec`](https://github.com/pirj/real-world-rspec):
```
57497 files inspected, 230 offenses detected, 230 offenses autocorrectable

1 error occurred:
An error occurred while RSpec/SortMetadata cop was inspecting /Users/pirj/source/real-world-rspec/rspec-core/spec/rspec/core/metadata_spec.rb:51:14.
```

pirj (approved):
Looks great, thank you!

lib/rubocop/cop/rspec/sort_metadata.rb
```diff
@@ -0,0 +1,94 @@
...
+        def sorted?(symbols, pairs)
+          symbols == sort_symbols(symbols) && pairs == sort_pairs(pairs)
+        end
+
+        def sort_pairs(pairs)
+          pairs.sort_by { |pair| pair.key.source.downcase }
+        end
+
+        def sort_symbols(symbols)
+          symbols.sort_by { |symbol| symbol.value.to_s.downcase }
```
leoarnold, line 89: This could fail, e.g.

```ruby
describe "Something", :a, "B", variable do
end
```

Maybe sort by `.value.to_s` for `sym` and `str`, and `.source` otherwise.

lib/rubocop/cop/rspec/sort_metadata.rb
```diff
@@ -0,0 +1,100 @@
...
+        PATTERN
+
+        def on_block(node)
+          if (block_var = rspec_configure(node))
+            metadata_in_block(node).each do |receiver, symbols, pairs|
+              investigate(symbols, pairs.flatten) if receiver == block_var
+            end
+          elsif (symbols, pairs = rspec_metadata(node))
+            investigate(symbols, pairs.flatten)
+          end
```
Darhazer, line 51: One nice thing about `def_node_matcher` is that It yields the match (if there is one; and the node if it matches, but there are no explicit captures), or doesn't yield if there is no match. So that could be rewritten as follows

```suggestion
          rspec_configure(node) do |block_var|
            metadata_in_block(node).each do |receiver, symbols, pairs|
              investigate(symbols, pairs.flatten) if receiver == block_var
            end
          end
          
          rspec_metadata(node) do |symbols, pairs|
            investigate(symbols, pairs.flatten)
          end
```
Furthermore, the `def_node_matcher` may receive additional arguments (with the use of `%`) so you can add a parameter to `metadata_in_block`:
```
         def_node_search :metadata_in_block, <<~PATTERN
           (send (lvar %) #Hooks.all _ ${send str sym}* (hash $...)?)
         PATTERN
```
And then the code will become:
```
          rspec_configure(node) do |block_var|
            metadata_in_block(node, block_var) do |symbols, pairs|
              investigate(symbols, pairs.flatten)
            end
          end
          
          rspec_metadata(node) do |symbols, pairs|
            investigate(symbols, pairs.flatten)
          end
```

spec/rubocop/cop/rspec/sort_metadata_spec.rb
```diff
@@ -0,0 +1,217 @@
...
+
+    expect_correction(<<~RUBY)
+      shared_examples 'a difficult situation', :a, 'B' do |x, y|
+      end
+
+      include_examples 'a difficult situation', 'value', 'another value'
+    RUBY
+  end
+
+  it 'does not register an offense ' \
```
Darhazer, line 130: I wonder why is that? It still would be nice to have it sorted

Darhazer (commented):
I'll review in more details later. For now I've left a note on matcher usage, that you may (or may not) apply at your own discretion; also left a question

spec/rubocop/cop/rspec/sort_metadata_spec.rb
bquorning: I agree, we should flag multiline offenses too. I don’t know if we should attempt to autocorrect them though.

spec/rubocop/cop/rspec/sort_metadata_spec.rb
```diff
@@ -0,0 +1,217 @@
...
+      it 'Something', :a, :b, baz: true, foo: 'bar' do
+      end
+    RUBY
+  end
+
+  it 'registers an offense when using mixed metadata, ' \
+     'but only the symbol keys are in alphabetical order' do
+    expect_offense(<<~RUBY)
+      it 'Something', :a, :b, foo: 'bar', baz: true do
+                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ Sort metadata alphabetically.
```
bquorning, line 74: When seeing this case, I wondered if we should have a separate offense for the symbols and for the hash metadata? In a case like `:y, :x, a: "b"`, would users feel that we ask them to write `a: "b", :x, :y` (which wouldn’t work)?

Maybe I’m overreacting and there is no problem at all.

spec/rubocop/cop/rspec/sort_metadata_spec.rb
leoarnold: In the multi line case, when I provide the range, it will only put carets under the first line of the metadata, even though the range goes further. Is that intentional or am I doing something wrong?

bquorning:
Hi @leoarnold, I’m a bit eager to merge this PR and get a release ready before I’m going on vacation – so I added a commit with @Darhazer’s code suggestion, plus support for multiline offenses (even if only the 1st line gets marked). I hope that’s ok 😅

Eventually, we’ll probably squash the two commits into just one before merging.

leoarnold:
@bquorning I recognize that it is unclear "whose code" it is if you open a PR on somebody else's repository. Nevertheless I am very unhappy with such conduct. I don't understand why there is any objective requirement to rush this PR. And it kinda feels like you are putting pressure on other people solely based on constraints of your private life plans.

bquorning:
> solely based on constraints of your private life plans.

Much open source work is to a certain degree constrained by the maintainers’ private life plans 😄

But @leoarnold, I am sorry for putting undue pressure on you. You are correct that there is no objective requirement to rush this PR. I guess I was getting carried away by this new cop being great and me feeling we were getting close to cutting a new release.

> I am very unhappy with such conduct.

I am sorry to hear that, and please feel free to revert my changes.

Darhazer (approved):

lib/rubocop/cop/rspec/sort_tags.rb
pirj: I see that this cop only suggests ordering, right? Changing from `:foo` to `foo: true` and the other way around was handy, too. Do you plan to submit it as yet another cop?

pirj:
@leoarnold Is there anything left you want to adjust in this PR before it gets merged?

Darhazer:
@leoarnold I understand your frustration. Though personally I was more frustrated at the times when my PRs were ignored (one in PHPUnit will turn [6 this year](https://github.com/sebastianbergmann/phpunit-mock-objects/pull/285), despite the efforts of the community to [push](https://github.com/sebastianbergmann/phpunit/issues/3999) it [forward](https://github.com/sebastianbergmann/phpunit/pull/4000). 

Realistically, if we'd merge your original PR, and then add the changes as a separate commit (as they are done) - the end result - code-wise - would be the same. 

The difference is that you didn't got the chance to apply - or reply to - the suggestions yourself. I'm sorry about that. I hope you'd accept our - all of the rubocop-rspec maintainers - appology for this rushed decision. 

I'm not merging this for the sole reason to give you the chance to respond (well, you could do that in a closed PR as well). Just responding with `I'm happy for this PR to be merged` would be enough. And I thank you for this great contribution :bow:

Darhazer:
Also I just noticed there is no changelog entry for the new cop

bquorning:
As mentioned earlier, I really wish to see this pull request merged. But we have not heard back from the original contributor in several weeks, and I think moving forward now can’t be considered unnecessary rushing.

I am not going to repeat the mistake of pushing (let alone force-pushing) to a contributor’s fork of the repository. Instead, I consider opening a new pull request with @leoarnold’s commit and three other commits with subsequent changes. I hope this is an acceptable solution 🙏🏼

pirj:
Superseded by #1378

@leoarnold I encourage you to revive the original part of this PR related to changing the style `focus: true` <-> `:focus` as a separate cop.

PR #1328, opened 2022-07-22
[Fix #1327] Fix a false positive for `RSpec/Capybara/SpecificMatcher`

### LINKED ISSUE #1327: [RSpec/Capybara/SpecificMatcher] Matching on attributes
AFAIK there's not a way to specify attributes with specific matchers:

`have_selector('a[data-disabled]', text: "Add Thing")`

This violates RSpec/Capybara/SpecificMatcher, but I don't think it's possible to express this with `have_link`, so  I think it's a false positive.

ydah:
Fix: #1327

---

* [x] Added tests.
* [-] Updated documentation.

### DIFF
--- a/lib/rubocop/cop/rspec/capybara/specific_matcher.rb
+++ b/lib/rubocop/cop/rspec/capybara/specific_matcher.rb
@@ -28,6 +28,24 @@ module RuboCop
           MSG = 'Prefer `%<good_matcher>s` over `%<bad_matcher>s`.'
           RESTRICT_ON_SEND = %i[have_selector have_no_selector have_css
                                 have_no_css].freeze
+          COMMON_OPTIONS = %w[
+            above below left_of right_of near count minimum maximum between text
+            id class style visible obscured exact exact_text normalize_ws match
+            wait filter_set focused
+          ].freeze
+          HAVE_BUTTON_OPTIONS = COMMON_OPTIONS + %w[
+            disabled name value title type
+          ].freeze
+          HAVE_LINK_OPTIONS = COMMON_OPTIONS + %w[
+            href alt title download
+          ].freeze
+          HAVE_TABLE_OPTIONS = COMMON_OPTIONS + %w[
+            caption with_cols cols with_rows rows
+          ].freeze
+          HAVE_SELECT_OPTIONS = COMMON_OPTIONS + %w[
+            disabled name placeholder options enabled_options disabled_options
+            selected with_selected multiple with_options
+          ].freeze
           SPECIFIC_MATCHER = {
             'button' => 'button',
             'a' => 'link',
@@ -44,6 +62,7 @@ module RuboCop
             return unless (arg = first_argument(node))
             return unless (matcher = specific_matcher(arg))
             return if acceptable_pattern?(arg)
+            return unless specific_matcher_option?(arg, matcher)
 
             add_offense(node, message: message(node, matcher))
           end
@@ -56,7 +75,28 @@ module RuboCop
           end
 
           def acceptable_pattern?(arg)
-            arg.match?(/\[.+=\w+\]/) || arg.match?(/[ >,+]/)
+            arg.match?(/[ >,+]/)
+          end
+
+          def specific_matcher_option?(arg, matcher)
+            # If `button[foo-bar_baz=foo][bar][baz]`:
+            # extract ["foo-bar_baz", "bar", "baz"]
+            attributes = arg.scan(/\[(.*?)(?:=.*?)?\]/).flatten
+            return true if attributes.empty?
+
+            attributes.all? do |attr|
+              specific_matcher_options(matcher).include?(attr)
+            end
+          end
+
+          def specific_matcher_options(matcher)
+            case matcher
+            when 'button' then HAVE_BUTTON_OPTIONS
+            when 'link' then HAVE_LINK_OPTIONS
+            when 'table' then HAVE_TABLE_OPTIONS
+            when 'select' then HAVE_SELECT_OPTIONS
+            else []
+            end
           end
 
           def message(node, matcher)
--- a/spec/rubocop/cop/rspec/capybara/specific_matcher_spec.rb
+++ b/spec/rubocop/cop/rspec/capybara/specific_matcher_spec.rb
@@ -21,59 +21,145 @@ RSpec.describe RuboCop::Cop::RSpec::Capybara::SpecificMatcher do
 
   it 'registers an offense when using `have_selector`' do
     expect_offense(<<-RUBY)
-    expect(page).to have_selector('button')
-                    ^^^^^^^^^^^^^^^^^^^^^^^ Prefer `have_button` over `have_selector`.
-    expect(page).to have_selector('a')
-                    ^^^^^^^^^^^^^^^^^^ Prefer `have_link` over `have_selector`.
-    expect(page).to have_selector('table')
-                    ^^^^^^^^^^^^^^^^^^^^^^ Prefer `have_table` over `have_selector`.
-    expect(page).to have_selector('select')
-                    ^^^^^^^^^^^^^^^^^^^^^^^ Prefer `have_select` over `have_selector`.
+      expect(page).to have_selector('button')
+                      ^^^^^^^^^^^^^^^^^^^^^^^ Prefer `have_button` over `have_selector`.
+      expect(page).to have_selector('a')
+                      ^^^^^^^^^^^^^^^^^^ Prefer `have_link` over `have_selector`.
+      expect(page).to have_selector('table')
+                      ^^^^^^^^^^^^^^^^^^^^^^ Prefer `have_table` over `have_selector`.
+      expect(page).to have_selector('select')
+                      ^^^^^^^^^^^^^^^^^^^^^^^ Prefer `have_select` over `have_selector`.
     RUBY
   end
 
   it 'registers an offense when using `have_no_selector`' do
     expect_offense(<<-RUBY)
-    expect(page).to have_no_selector('button')
-                    ^^^^^^^^^^^^^^^^^^^^^^^^^^ Prefer `have_no_button` over `have_no_selector`.
+      expect(page).to have_no_selector('button')
+                      ^^^^^^^^^^^^^^^^^^^^^^^^^^ Prefer `have_no_button` over `have_no_selector`.
     RUBY
   end
 
   it 'registers an offense when using `have_css`' do
     expect_offense(<<-RUBY)
-    expect(page).to have_css('button')
-                    ^^^^^^^^^^^^^^^^^^ Prefer `have_button` over `have_css`.
+      expect(page).to have_css('button')
+                      ^^^^^^^^^^^^^^^^^^ Prefer `have_button` over `have_css`.
     RUBY
   end
 
   it 'registers an offense when using `have_no_css`' do
     expect_offense(<<-RUBY)
-    expect(page).to have_no_css('button')
-                    ^^^^^^^^^^^^^^^^^^^^^ Prefer `have_no_button` over `have_no_css`.
+      expect(page).to have_no_css('button')
+                      ^^^^^^^^^^^^^^^^^^^^^ Prefer `have_no_button` over `have_no_css`.
     RUBY
   end
 
   it 'registers an offense when using abstract matcher and other args' do
     expect_offense(<<-RUBY)
-    expect(page).to have_css('button', exact_text: 'foo')
-                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ Prefer `have_button` over `have_css`.
+      expect(page).to have_css('button', exact_text: 'foo')
+                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ Prefer `have_button` over `have_css`.
+    RUBY
+  end
+
+  %i[above below left_of right_of near count minimum maximum between
+     text id class style visible obscured exact exact_text normalize_ws
+     match wait filter_set focused disabled name value
+     title type].each do |attr|
+    it 'registers an offense for abstract matcher when ' \
+       "first argument is element with replaceable attributes #{attr} " \
+       'for `have_button`' do
+      expect_offense(<<-RUBY, attr: attr)
+        expect(page).to have_css("button[#{attr}=foo]")
+                        ^^^^^^^^^^^^^^^^^^{attr}^^^^^^^ Prefer `have_button` over `have_css`.
+        expect(page).to have_css("button[#{attr}]")
+                        ^^^^^^^^^^^^^^^^^^{attr}^^^ Prefer `have_button` over `have_css`.
+      RUBY
+    end
+  end
+
+  %i[above below left_of right_of near count minimum maximum between text id
+     class style visible obscured exact exact_text normalize_ws match wait
+     filter_set focused href alt title download].each do |attr|
+    it 'registers an offense for abstract matcher when ' \
+       "first argument is element with replaceable attributes #{attr} " \
+       'for `have_link`' do
+      expect_offense(<<-RUBY, attr: attr)
+        expect(page).to have_css("a[#{attr}=foo]")
+                        ^^^^^^^^^^^^^{attr}^^^^^^^ Prefer `have_link` over `have_css`.
+        expect(page).to have_css("a[#{attr}]")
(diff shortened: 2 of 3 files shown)


lib/rubocop/cop/rspec/capybara/specific_matcher.rb
```diff
@@ -59,6 +84,28 @@ def acceptable_pattern?(arg)
             arg.match?(/\[.+=\w+\]/) || arg.match?(/[ >,+]/)
           end
 
+          def specific_matcher_option?(arg, matcher)
+            attribute = arg.slice(/\[[\w\W]*\]/)&.gsub!(/[\[\]]*/, '')
```
pirj, line 88: Cosmetic: does it have to be a bang-`gsub`?
I'm too bad with regexp, do you mind adding a short note what this does exactly?

lib/rubocop/cop/rspec/capybara/specific_matcher.rb
```diff
@@ -59,6 +84,28 @@ def acceptable_pattern?(arg)
...
+            if attribute
```
pirj, line 89: Cosmetic suggestion:
```ruby
return true unless attribute

specific_matcher_options(matcher).include?(attribute)
```

spec/rubocop/cop/rspec/capybara/specific_matcher_spec.rb
```diff
@@ -60,20 +60,72 @@
...
   end
 
   it 'does not register an offense for abstract matcher when ' \
     'first argument is element with nonreplaceable attributes' do
     expect_no_offenses(<<-RUBY)
-      expect(page).to have_css('button[foo=bar]')
-      expect(page).to have_css('button[foo-bar=baz]', exact_text: 'foo')
+    expect(page).to have_css('button[data_disabled]')
+    expect(page).to have_css('button[foo=bar]')
+    expect(page).to have_css('button[foo-bar=baz]', exact_text: 'foo')
```
pirj, line 128: Cosmetic: I really prefer it indented as above.

spec/rubocop/cop/rspec/capybara/specific_matcher_spec.rb
```diff
@@ -60,20 +60,72 @@
...
   end
 
+  %i[above below left_of right_of near count minimum maximum between text id
+     class style visible obscured exact exact_text normalize_ws match wait
+     filter_set focused disabled name value title type name
+     disabled].each do |attr|
+    it 'registers an offense for abstract matcher when ' \
+      "first argument is button element with #{attr}" do
+      expect_offense(<<-RUBY, attr: attr)
+        expect(page).to have_css("button[#{attr}]")
```
pirj, line 70: I don't get it, do we check that we flag `have_css("button[with]")` as an offence? Is this a legitimate syntax?

lib/rubocop/cop/rspec/capybara/specific_matcher.rb
```diff
@@ -28,6 +28,30 @@ class SpecificMatcher < Base
...
+          HAVE_BUTTON_OPTIONS = %w[above below left_of right_of near count
+                                   minimum maximum between text id class
+                                   style visible obscured exact exact_text
+                                   normalize_ws match wait filter_set focused
+                                   disabled name value title type name
+                                   disabled].freeze
+          HAVE_LINK_OPTIONS = %w[above below left_of right_of near count
+                                 minimum maximum between text id class style
+                                 visible obscured exact exact_text normalize_ws
+                                 match wait filter_set focused href href
```
pirj, line 40: Double `href`.
What makes me wonder if there are other duplicates.

lib/rubocop/cop/rspec/capybara/specific_matcher.rb
ydah: It did not have to be bang-gsub. I have written a description of what I do at regex.

lib/rubocop/cop/rspec/capybara/specific_matcher.rb
ydah: It's certainly better that way. Fixed.

spec/rubocop/cop/rspec/capybara/specific_matcher_spec.rb
ydah: Oops, fixed.

spec/rubocop/cop/rspec/capybara/specific_matcher_spec.rb
ydah: This cop only offense attributes that can be replaced by specific matcher. I think the `it` explanation was confusing. Improved description. How do you like it?

lib/rubocop/cop/rspec/capybara/specific_matcher.rb
ydah: Oops, I reviewed the other definitions as well.

ydah: I was referring to the capybara error message. but it appears to have been a duplicate suggestion display.
https://github.com/teamcapybara/capybara/pull/2572

pirj:
I'm curious, should this be offensive or not?
```ruby
expect(page).to have_css("a[href=google.com]")
```
Specs currently tell it's not, but that
```ruby
expect(page).to have_css("a[href]")
```
is.

Same for e.g.
```ruby
expect(page).to have_css("button[name=click-me]")
```
vs
```ruby
expect(page).to have_css("button[name]")
```

Is it common to use `name`/`href` without a value in selectors?

ydah:
Indeed it is. Thank you so much.

I've considered offense both of the following:

```ruby
expect(page).to have_css("a[href=google.com]")
expect(page).to have_css("a[href]")
```

How about this?

pirj:
Nice!

> I've considered offense both of the following:
```ruby
expect(page).to have_css("a[href=google.com]")
expect(page).to have_css("a[href]")
```

I could only find the first one is specs.
```ruby
expect(page).to have_css("select[#{attr}=foo]")
```
Do you mind adding the second one, too? I can imaging e.g `visible` or `disabled` can be used without the value.
I suggest keeping them in the same examples, e.g.:
```ruby
        expect(page).to have_css("table[#{attr}=foo]")
                        ^^^^^^^^^^^^^^^^^{attr}^^^^^^^ Prefer `have_table` over `have_css`.
        expect(page).to have_css("table[#{attr}]")
                        ^^^^^^^^^^^^^^^^^{attr}^^^ Prefer `have_table` over `have_css`.
```

ydah:
It certainly looks better that way. I've added a test.

pirj (approved):
Thank you!

lib/rubocop/cop/rspec/capybara/specific_matcher.rb
```diff
@@ -56,7 +79,31 @@ def specific_matcher(arg)
           end
 
           def acceptable_pattern?(arg)
-            arg.match?(/\[.+=\w+\]/) || arg.match?(/[ >,+]/)
+            arg.match?(/[ >,+]/)
+          end
+
+          def specific_matcher_option?(arg, matcher)
+            # If `button[foo-bar_baz=foo]`, extract the "foo-bar_baz".
+            attribute = arg.slice(/\[[\w\W]*\]/)
```
bquorning, line 87: Two questions:
1. Can there be more than one attribute within the brackets?
2. Would a regular expression (with capture) be easier to use here?

ydah: > 1. Can there be more than one attribute within the brackets?

The following code will result in a syntax error.
```ruby
expect(page).to have_css("a[href="http://example.com" disabled]")
=> Failure/Error: expect(page).to have_css('a[href="http://example.com" disabled]')
     
     Nokogiri::CSS::SyntaxError:
       unexpected ' ' after '#<Nokogiri::CSS::Node:0x00007fddf305d998>'
     # (eval):3:in `_racc_do_parse_c'
     # (eval):3:in `do_parse'
     # ./spec/capybara/sandbox_spec.rb:8:in `block (2 levels) in <top (required)>'
```

However, multiple attributes can be specified as shown below, so I've added tests for multiple attributes.
```ruby
expect(page).to have_css('a[href="http://example.com"][disabled]')
```

ydah: >2. Would a regular expression (with capture) be easier to use here?

Indeed it is. Thank you so much. I updated it.

lib/rubocop/cop/rspec/capybara/specific_matcher.rb
```diff
@@ -56,7 +79,31 @@ def specific_matcher(arg)
...
+            arg.match?(/[ >,+]/)
+          end
+
+          def specific_matcher_option?(arg, matcher)
+            # If `button[foo-bar_baz=foo][bar][baz]`:
+            # extract "foo-bar_baz][bar][baz"
+            attributes, = arg.match(/\[([\w\W]*)\]/)&.captures
+            return true unless attributes
+
+            attributes.split('][').all? do |attr|
```
bquorning, line 91: How about e.g.
```suggestion
            # If `button[foo-bar_baz=foo][bar][baz]`:
            # extract ["foo-bar_baz", "bar", "baz"]
            attributes = arg.scan(/\[(.*?)\]/).flatten
            return true if attributes.empty?

            attributes.all? do |attr|
```

ydah: I was having a little trouble there, but it looks to have improved  and better way to me. Thank you so much.

bquorning (approved):

lib/rubocop/cop/rspec/capybara/specific_matcher.rb
```diff
@@ -28,6 +28,28 @@ class SpecificMatcher < Base
           MSG = 'Prefer `%<good_matcher>s` over `%<bad_matcher>s`.'
           RESTRICT_ON_SEND = %i[have_selector have_no_selector have_css
                                 have_no_css].freeze
+          HAVE_BUTTON_OPTIONS = %w[
```
Darhazer, line 31: Perhaps we can extract `COMMON_OPTIONS` and then extend for the different elements

lib/rubocop/cop/rspec/capybara/specific_matcher.rb
```diff
@@ -56,7 +79,31 @@ def specific_matcher(arg)
...
+          def specific_matcher_options(matcher)
+            case matcher
+            when 'button'
+              HAVE_BUTTON_OPTIONS
+            when 'link'
+              HAVE_LINK_OPTIONS
+            when 'table'
+              HAVE_TABLE_OPTIONS
+            when 'select'
+              HAVE_SELECT_OPTIONS
```
Darhazer, line 105: should we default to [] if the matcher is not in the list (even though some other part might guarantee it's one of those)

lib/rubocop/cop/rspec/capybara/specific_matcher.rb
```diff
@@ -56,7 +79,31 @@ def specific_matcher(arg)
...
 
           def acceptable_pattern?(arg)
-            arg.match?(/\[.+=\w+\]/) || arg.match?(/[ >,+]/)
+            arg.match?(/[ >,+]/)
+          end
+
+          def specific_matcher_option?(arg, matcher)
+            # If `button[foo-bar_baz=foo][bar][baz]`:
+            # extract ["foo-bar_baz", "bar", "baz"]
+            attributes = arg.scan(/\[(.*?)\]/).flatten
```
bquorning, line 88: If we want to go all in on regex:
```suggestion
            attributes = arg.scan(/\[(.*?)(?:=.*?)?\]/).flatten
```
Then we can delete the `.split('=').first` in line 92.

lib/rubocop/cop/rspec/capybara/specific_matcher.rb
ydah: It certainly looks better that way. Thank you very much.

lib/rubocop/cop/rspec/capybara/specific_matcher.rb
ydah: I added this because it is an unlikely handling, but it certainly seemed like it would be better to have.

lib/rubocop/cop/rspec/capybara/specific_matcher.rb
ydah: It is clearly better that way. Thank you so much.

spec/rubocop/cop/rspec/capybara/specific_matcher_spec.rb
```diff
@@ -21,62 +21,148 @@
...
+                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ Prefer `have_button` over `have_css`.
+      expect(page).to have_css('button:not([disabled])', exact_text: 'bar')
+                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ Prefer `have_button` over `have_css`.
     RUBY
   end
 
   it 'does not register an offense for abstract matcher when ' \
     'first argument is element with nonreplaceable attributes' do
     expect_no_offenses(<<-RUBY)
+      expect(page).to have_css('button[data_disabled]')
```
Darhazer, line 149: `[data-disabled]` is more realistic usage

Darhazer (approved):

bquorning (approved):

spec/rubocop/cop/rspec/capybara/specific_matcher_spec.rb
ydah: Indeed it is. I change to `[data-disabled]`

bquorning (approved):

lib/rubocop/cop/rspec/capybara/specific_matcher.rb
```diff
@@ -28,6 +28,24 @@ class SpecificMatcher < Base
...
           RESTRICT_ON_SEND = %i[have_selector have_no_selector have_css
                                 have_no_css].freeze
+          COMMON_OPTIONS = %w[
+            above below left_of right_of near count minimum maximum between text
+            id class style visible obscured exact exact_text normalize_ws match
+            wait filter_set focused
+          ].freeze
+          HAVE_BUTTON_OPTIONS = COMMON_OPTIONS + %w[
+            disabled name value title type
+          ].freeze
```
bquorning, line 38: I just realized that this way, `HAVE_BUTTON_OPTIONS` is not frozen. We’d have to do e.g.
```rb
HAVE_BUTTON_OPTIONS = (COMMON_OPTIONS + %w[
  disabled name value title type
]).freeze
```
which really doesn’t seem very elegant.

Perhaps we should just have one big SPECIFIC_MATCHER_OPTIONS hash instead? What do you think @ydah?

```rb
SPECIFIC_MATCHER_OPTIONS = {
  'button' => COMMON_OPTIONS + %w[
    disabled name value title type
  ],
  'link' => COMMON_OPTIONS + %w[
    href alt title download
  ],
  'table' => COMMON_OPTIONS + %w[
    caption with_cols cols with_rows rows
  ],
  'select' => COMMON_OPTIONS + %w[
    disabled name placeholder options enabled_options disabled_options
    selected with_selected multiple with_options
  ]
}.freeze
```

BrianHawley: You'd still have to freeze the hash values, but that can be done like this:
```ruby
SPECIFIC_MATCHER_OPTIONS = {
  #...
}.each_value(&:freeze).freeze
```

bquorning: In that case, I think I’d prefer calling `freeze` directly on the created arrays anyway:
```rb
SPECIFIC_MATCHER_OPTIONS = {
  'button' => (
    COMMON_OPTIONS + %w[disabled name value title type]
  ).freeze,
  'link' => (
    COMMON_OPTIONS + %w[href alt title download]
  ).freeze,
  'table' => (
    COMMON_OPTIONS + %w[caption with_cols cols with_rows rows]
  ).freeze,
  'select' => (
    COMMON_OPTIONS + %w[
      disabled name placeholder options enabled_options
      disabled_options selected with_selected multiple with_options
    ]
  ).freeze
}.freeze
```

pirj: My apologies for merging this. I've just realized there are more comments added after I last refreshed the page.
I hope it's not something that can't be fixed in a follow-up.

pirj:
Thank you for the fix, @ydah !

PR #1333, opened 2022-07-23
Sort Changelog contributors list alphabetically

bquorning:
Now, new contributors no longer have to consider if their handle should be added at the top, the bottom or in the middle.

---

* [x] Added tests.
* [ ] Updated documentation.


pirj (approved):
Awesome, thank you!

PR #1336, opened 2022-07-26
Deprecate `IgnoredPatterns` option in favor of the `AllowedPatterns` options

ydah:
Follow up https://github.com/rubocop/rubocop/pull/10555

This PR obsoletes the `IgnoredPatterns` option and replaces it with the `AllowedPatterns` option.

---

* [x] Added tests.
* [x] Updated documentation.

### DIFF
--- a/config/default.yml
+++ b/config/default.yml
@@ -794,9 +794,10 @@ RSpec/VariableName:
   SupportedStyles:
     - snake_case
     - camelCase
+  AllowedPatterns: []
   IgnoredPatterns: []
   VersionAdded: '1.40'
-  VersionChanged: '1.43'
+  VersionChanged: '2.13'
   Reference: https://www.rubydoc.info/gems/rubocop-rspec/RuboCop/Cop/RSpec/VariableName
 
 RSpec/VerifiedDoubleReference:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@
 * Add support for numblocks to `RSpec/AroundBlock`, `RSpec/EmptyLineAfterHook`, `RSpec/ExpectInHook`, `RSpec/HookArgument`, `RSpec/HooksBeforeExamples`, `RSpec/IteratedExpectation`, and `RSpec/NoExpectationExample`. ([@ydah][])
 * Fix incorrect documentation URLs when using `rubocop --show-docs-url`. ([@r7kamura][])
 * Add `AllowedGroups` configuration option to `RSpec/NestedGroups`. ([@ydah][])
+* Deprecate `IgnoredPatterns` option in favor of the `AllowedPatterns` options. ([@ydah][])
 
 ## 2.12.1 (2022-07-03)
 
@@ -666,7 +667,6 @@ Compatibility release so users can upgrade RuboCop to 0.51.0. No new features.
 [@hosamaly]: https://github.com/hosamaly
 [@jaredbeck]: https://github.com/jaredbeck
 [@jaredmoody]: https://github.com/jaredmoody
-[@jawshooah]: https://github.com/jawshooah
 [@jeffreyc]: https://github.com/jeffreyc
 [@jfragoulis]: https://github.com/jfragoulis
 [@johnny-miyake]: https://github.com/johnny-miyake
--- a/config/obsoletion.yml
+++ b/config/obsoletion.yml
@@ -1,0 +1,14 @@
+#
+# Configuration of obsolete/deprecated cops used by `ConfigObsoletion`.
+#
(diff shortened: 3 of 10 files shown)


ydah:
https://github.com/rubocop/rubocop/pull/10831 is required to resolve the build errors.

config/default.yml
```diff
@@ -778,9 +778,10 @@ RSpec/VariableName:
   SupportedStyles:
     - snake_case
     - camelCase
+  AllowedPatterns: []
   IgnoredPatterns: []
```
Darhazer, line 782: should we remove the obsolete configuration from the default?

Darhazer (commented):
> https://github.com/rubocop/rubocop/pull/10831 is required to resolve the build errors.

Does it also mean that we should bump the minimum required version? The AllowedPatterns is available since 1.28 and we already depend on 1.31, but if deprecation is broken, perhaps we should wait for the next release and also add dependency on it?

ydah:
> Does it also mean that we should bump the minimum required version? The AllowedPatterns is available since 1.28 and we already depend on 1.31, but if deprecation is broken, perhaps we should wait for the next release and also add dependency on it?

Yes. That is the recognition.
https://github.com/rubocop/rubocop/pull/10831 has not been released yet, so it looks like we will have to wait for the next release.

config/default.yml
ydah: Obsolete parameters are still available.
It seems to me that it would be better to keep both available for backward compatibility until the complete removal of the corresponding options at https://github.com/rubocop/rubocop, which will come someday.

pirj:
> > https://github.com/rubocop/rubocop/pull/10831 is required to resolve the build errors.

> Does it also mean that we should bump the minimum required version?

But it's not released yet. This is why `CI / spec Ruby 2.7 (edge)` (`rubocop` `master`) passes, but other builds don't.

Let's wait for the fix to be released and bump the dependency.

ydah:
Now we can incorporate this change:
- https://github.com/rubocop/rubocop/releases/tag/v1.33.0

Darhazer:
we should also bump the minimum required rubocop version

ydah:
@Darhazer Sorry for the delay in responding. I updated this PR.

bquorning (approved):

bquorning:
@pirj / @Darhazer I’d like extra eyes on this PR. 🙏🏼

Darhazer (approved):

config/default.yml
jdelStrother: In the meantime, is there a good way of suppressing this on every run?
```
Warning: obsolete parameter `IgnoredPatterns` (for `RSpec/VariableName`) found in vendor/bundle/ruby/3.1.0/gems/rubocop-rspec-2.13.2/config/default.yml
`IgnoredPatterns` has been renamed to `AllowedPatterns`.
Warning: obsolete parameter `IgnoredPatterns` (for `RSpec/VariableName`) found in .rubocop.yml
`IgnoredPatterns` has been renamed to `AllowedPatterns`.
```

ydah: Can you provide .rubocop.yml?

ydah:
```
Warning: obsolete parameter `IgnoredPatterns` (for `RSpec/VariableName`) found in .rubocop.yml
`IgnoredPatterns` has been renamed to `AllowedPatterns`.
```

This warning can be suppressed by modifying .rubocop.yml as follows.

### before
```yaml
RSpec/VariableName:
  IgnoredPatterns: [...]
```

### after
```yaml
RSpec/VariableName:
  AllowedPatterns: [...] 
```

ydah:
```
Warning: obsolete parameter `IgnoredPatterns` (for `RSpec/VariableName`) found in vendor/bundle/ruby/3.1.0/gems/rubocop-rspec-2.13.2/config/default.yml
`IgnoredPatterns` has been renamed to `AllowedPatterns`.
```

This is a guess, but I think the following settings may be included in the .rubocop.yml file.

```yaml
inherit_from: vendor/bundle/ruby/3.1.0/gems/rubocop-rspec-2.13.2/config/default.yml
```

jdelStrother: Yeah, I get the warnings with our .rubocop.yml like:

```yml
inherit_gem:
  rubocop-rspec:
    - config/default.yml

require:
  - rubocop-rspec

AllCops:
  TargetRubyVersion: 3.1
  NewCops: enable
```

I still get the warnings if I append this to the file - 

```yaml
RSpec/VariableName:
  AllowedPatterns: [] 
```

However, looking into it - I'm not so sure we need the inherit_gem bit. Seems like if I remove it we still get the rspec defaults, but don't get the deprecation warnings.

Sorry for the noise!

Issue #1326, opened 2022-07-18
[RSpec/Capybara/SpecificMatcher] False-negative for `click_link`


boris-petrov:
```rb
find('#some-id a.some-class').click
```

This could be changed to:

```rb
find('#some-id').click_link(class: 'some-class')
# or
find_by_id('some-id').click_link(class: 'some-class')
```

I'm still not sure if that's better but I wanted to put that for discussion here.

cc @ydah, @pirj, @bquorning

ydah:
```ruby
find('#some-id a.some-class').click

find('#some-id').click_link(class: 'some-class')
```

In the above case, it looked good as `RSpec/Capybara/SpecificActions`.
ref: https://rubydoc.info/github/jnicklas/capybara/master/Capybara%2FNode%2FActions:click_link

ydah:
```ruby
# bad
find('#some-id')

# good
find_by_id('some-id')
```

And in the above case, `RSpec/Capybara/SpecificFinders` ?
ref: https://rubydoc.info/github/jnicklas/capybara/Capybara/Node/Finders

PR #1340, opened 2022-07-29
Add new `RSpec/Capybara/SpecificFinders` cop

ydah:
Resolve part of the issue: https://github.com/rubocop/rubocop-rspec/issues/1326

This cop checks for there is a more specific finder offered by Capybara.

```ruby
# bad
find('#some-id')

# good
find_by_id('some-id')
```

---

* [x] Added tests.
* [x] Updated documentation.

### DIFF
--- a/lib/rubocop/cop/rspec/capybara/specific_finders.rb
+++ b/lib/rubocop/cop/rspec/capybara/specific_finders.rb
@@ -1,0 +1,86 @@
+# frozen_string_literal: true
+
+module RuboCop
+  module Cop
+    module RSpec
+      module Capybara
+        # Checks if there is a more specific finder offered by Capybara.
+        #
+        # @example
+        #   # bad
+        #   find('#some-id')
+        #   find('[visible][id=some-id]')
+        #
+        #   # good
+        #   find_by_id('some-id')
+        #   find_by_id('some-id', visible: true)
+        #
+        class SpecificFinders < Base
+          extend AutoCorrector
+
+          include RangeHelp
+
+          MSG = 'Prefer `find_by` over `find`.'
+          RESTRICT_ON_SEND = %i[find].freeze
+
+          # @!method find_argument(node)
+          def_node_matcher :find_argument, <<~PATTERN
+            (send _ :find (str $_) ...)
+          PATTERN
+
+          def on_send(node)
+            find_argument(node) do |arg|
+              next if CssSelector.multiple_selectors?(arg)
+
+              on_attr(node, arg) if attribute?(arg)
+              on_id(node, arg) if CssSelector.id?(arg)
+            end
+          end
+
+          private
+
+          def on_attr(node, arg)
+            return unless (id = CssSelector.attributes(arg)['id'])
+
+            register_offense(node, replaced_argments(arg, id))
+          end
+
+          def on_id(node, arg)
+            register_offense(node, "'#{arg.to_s.delete('#')}'")
+          end
+
+          def attribute?(arg)
+            CssSelector.attribute?(arg) &&
+              CssSelector.common_attributes?(arg)
+          end
+
+          def register_offense(node, arg_replacemenet)
+            add_offense(offense_range(node)) do |corrector|
+              corrector.replace(node.loc.selector, 'find_by_id')
+              corrector.replace(node.first_argument.loc.expression,
+                                arg_replacemenet)
+            end
+          end
+
+          def replaced_argments(arg, id)
+            options = to_options(CssSelector.attributes(arg))
+            options.empty? ? id : "#{id}, #{options}"
+          end
+
+          def to_options(attrs)
+            attrs.each.map do |key, value|
+              next if key == 'id'
+
+              "#{key}: #{value}"
+            end.compact.join(', ')
+          end
+
+          def offense_range(node)
+            range_between(node.loc.selector.begin_pos,
+                          node.loc.end.end_pos)
+          end
+        end
+      end
+    end
+  end
+end
--- a/lib/rubocop/cop/rspec/capybara/specific_matcher.rb
+++ b/lib/rubocop/cop/rspec/capybara/specific_matcher.rb
@@ -37,29 +37,26 @@ module RuboCop
             'select' => 'select',
             'input' => 'field'
           }.freeze
-          COMMON_OPTIONS = %w[
-            above below left_of right_of near count minimum maximum between text
-            id class style visible obscured exact exact_text normalize_ws match
-            wait filter_set focused
-          ].freeze
           SPECIFIC_MATCHER_OPTIONS = {
             'button' => (
-              COMMON_OPTIONS + %w[disabled name value title type]
+              CssSelector::COMMON_OPTIONS + %w[disabled name value title type]
             ).freeze,
             'link' => (
-              COMMON_OPTIONS + %w[href alt title download]
+              CssSelector::COMMON_OPTIONS + %w[href alt title download]
             ).freeze,
             'table' => (
-              COMMON_OPTIONS + %w[caption with_cols cols with_rows rows]
+              CssSelector::COMMON_OPTIONS + %w[
+                caption with_cols cols with_rows rows
+              ]
             ).freeze,
             'select' => (
-              COMMON_OPTIONS + %w[
+              CssSelector::COMMON_OPTIONS + %w[
                 disabled name placeholder options enabled_options
                 disabled_options selected with_selected multiple with_options
               ]
             ).freeze,
             'field' => (
-              COMMON_OPTIONS + %w[
+              CssSelector::COMMON_OPTIONS + %w[
                 checked unchecked disabled valid name placeholder
                 validation_message readonly with type multiple
               ]
@@ -77,12 +74,13 @@ module RuboCop
           PATTERN
 
           def on_send(node)
-            return unless (arg = first_argument(node))
-            return unless (matcher = specific_matcher(arg))
-            return if acceptable_pattern?(arg)
-            return unless specific_matcher_option?(node, arg, matcher)
+            first_argument(node) do |arg|
+              next unless (matcher = specific_matcher(arg))
+              next if CssSelector.multiple_selectors?(arg)
+              next unless specific_matcher_option?(node, arg, matcher)
 
-            add_offense(node, message: message(node, matcher))
+              add_offense(node, message: message(node, matcher))
+            end
           end
 
           private
@@ -97,26 +95,24 @@ module RuboCop
           end
 
           def specific_matcher_option?(node, arg, matcher)
-            # If `button[foo-bar_baz=foo][bar][baz]`:
-            # extract ["foo-bar_baz", "bar", "baz"]
-            attributes = arg.scan(/\[(.*?)(?:=.*?)?\]/).flatten
-            return true if attributes.empty?
-            return false unless replaceable_matcher?(node, matcher, attributes)
+            attrs = CssSelector.attributes(arg).keys
+            return true if attrs.empty?
+            return false unless replaceable_matcher?(node, matcher, attrs)
 
-            attributes.all? do |attr|
+            attrs.all? do |attr|
               SPECIFIC_MATCHER_OPTIONS.fetch(matcher, []).include?(attr)
             end
           end
 
-          def replaceable_matcher?(node, matcher, attributes)
+          def replaceable_matcher?(node, matcher, attrs)
             case matcher
-            when 'link' then replaceable_to_have_link?(node, attributes)
+            when 'link' then replaceable_to_have_link?(node, attrs)
             else true
             end
           end
 
-          def replaceable_to_have_link?(node, attributes)
-            option?(node, :href) || attributes.include?('href')
+          def replaceable_to_have_link?(node, attrs)
+            option?(node, :href) || attrs.include?('href')
           end
 
           def message(node, matcher)
--- a/lib/rubocop/cop/rspec/mixin/css_selector.rb
+++ b/lib/rubocop/cop/rspec/mixin/css_selector.rb
@@ -14,6 +14,24 @@ module RuboCop
         module_function
 
         # @param selector [String]
+        # @return [Boolean]
+        # @example
+        #   id?('#some-id') # => true
+        #   id?('.some-class') # => false
+        def id?(selector)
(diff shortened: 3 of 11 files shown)


lib/rubocop/cop/rspec/mixin/css_selector.rb
```diff
@@ -0,0 +1,38 @@
...
+        def id?(selector)
+          selector.start_with?('#')
+        end
+
+        # @param selector [String]
+        # @return [Boolean]
+        # @example
+        #   multiple_specified?('a.cls b#id') # => true
+        #   multiple_specified?('a.cls') # => false
+        def multiple_specified?(selector)
```
Darhazer, line 22: I think this needs a better name. Like:
`multiple?`
`multiple_selectors?`
`multiple_elements?`
`multiple_matches`?

Darhazer (commented):
I think such a cop has to have a configurable enforced style, e.g. one may prefer to enforce the `find('#id')` style

ydah:
The name of this cop is taken from `Capybara::Node::Finders`, so I am a bit worried about the EnforcedStyle setting if we are going to extend it as this cop to methods like `find_field` for example.

Ref: https://www.rubydoc.info/gems/capybara/Capybara/Node/Finders#find_field-instance_method

Or should I reconsider the name and make it a cop that only monitors find<-->find_by?

spec/rubocop/cop/rspec/capybara/specific_finders_spec.rb
```diff
@@ -0,0 +1,62 @@
...
+      find_by_id('some-id')
+    RUBY
+  end
+
+  it 'does not register an offense when using `find` ' \
+    'with argument is id with multiple matcher' do
+    expect_no_offenses(<<~RUBY)
+      find('#some-id body')
+      find('#some-id,h1')
+      find('#some-id,h2')
```
Darhazer, line 58: Those two cover the same case

spec/rubocop/cop/rspec/capybara/specific_finders_spec.rb
```diff
@@ -0,0 +1,62 @@
...
+      find('#some-id+option')
```
Darhazer, line 59: what about the case of `find('[id=something]')`

Darhazer: Also is it possible to write `find('h2#some')` ?

Darhazer (commented):
The implementation is good and I love the CssSelector extraction

It may conflict with #1345 
While we wait for 1345 to be merged, I've left some questions

spec/rubocop/cop/rspec/capybara/specific_finders_spec.rb
ydah: This was a typo for the following case. I fixed it.

https://github.com/rubocop/rubocop-rspec/blob/246e30e2c4cfdbb5762c0c6ec8475c356893273a/spec/rubocop/cop/rspec/capybara/specific_finders_spec.rb#L98

spec/rubocop/cop/rspec/capybara/specific_finders_spec.rb
ydah: https://github.com/rubocop/rubocop-rspec/blob/246e30e2c4cfdbb5762c0c6ec8475c356893273a/spec/rubocop/cop/rspec/capybara/specific_finders_spec.rb#L39-L48

ydah: I have summarized them below. What do you think?

https://github.com/rubocop/rubocop-rspec/blob/246e30e2c4cfdbb5762c0c6ec8475c356893273a/spec/rubocop/cop/rspec/capybara/specific_finders_spec.rb#L39-L77

lib/rubocop/cop/rspec/capybara/specific_finders.rb
```diff
@@ -0,0 +1,93 @@
...
+            end
+          end
+
+          def on_id(node, arg)
+            add_offense(offense_range(node)) do |corrector|
+              replace_good_method(corrector, node)
+              corrector.replace(node.first_argument.loc.expression,
+                                "'#{arg.delete('#')}'")
+            end
+          end
```
Darhazer, line 61: Perhaps those could be DRYed

```suggestion
          def on_attr(node, arg)
            id = attribute_value(arg, 'id')
            return if id.empty?

            register_offense(node, replaced_argments(arg, id))
          end

          def on_id(node, arg)
            register_offense(node, arg.to_s.delete('#'))
          end
          
          def register_offense(node, arg_replacemenet)
            add_offense(offense_range(node)) do |corrector|
              replace_good_method(corrector, node)
              corrector.replace(node.first_argument.loc.expression, arg_replacemenet)
            end
          end
```

lib/rubocop/cop/rspec/capybara/specific_finders.rb
```diff
@@ -0,0 +1,93 @@
...
+            attrs.map do |attr|
+              key, value = attr.split('=')
+              next if key == 'id'
+
+              "#{key}: #{option_value(value)}"
+            end.compact.join(', ')
+          end
+
+          def option_value(value)
+            value ? "'#{value}'" : true
```
Darhazer, line 78: is it possible for the `value` to be `false` (or would it be always a string if present?)

lib/rubocop/cop/rspec/capybara/specific_finders.rb
```diff
@@ -0,0 +1,93 @@
...
+          end
+
+          def replace_good_method(corrector, node)
+            corrector.replace(node.loc.selector, 'find_by_id')
```
Darhazer, line 82: I don't think this deserves separate method. It could be inlined, especially in the DRYed version

lib/rubocop/cop/rspec/mixin/css_selector.rb
```diff
@@ -0,0 +1,84 @@
...
+          selector.start_with?('[')
+        end
+
+        # @param selector [String]
+        # @return [Array<String>]
+        # @example
+        #   attributes('a[foo-bar_baz]') # => ["foo-bar_baz"]
+        #   attributes('button[foo][bar]') # => ["foo", "bar"]
+        #   attributes('table[foo=bar]') # => ["foo=bar"]
+        def attributes(selector)
```
Darhazer, line 38: I wonder if it won't make for a better API to return a hash, like
{foo-bar_baz: true, foo: true, baz: true, foo: 'bar'} respectively (from all the examples above). Then you don't really need the _keys and _value methods

lib/rubocop/cop/rspec/capybara/specific_finders.rb
ydah: Indeed it is, and we remove this method to match the DRYed version.

lib/rubocop/cop/rspec/capybara/specific_finders.rb
ydah: Thank you very much.  I move it to css_selector as a normalization and modified it. How do you think it?

lib/rubocop/cop/rspec/mixin/css_selector.rb
ydah: It looks very good. I fixed it.

ydah:
Conflicts in CHANGELOG.md have been resolved.

lib/rubocop/cop/rspec/mixin/css_selector.rb
```diff
@@ -0,0 +1,87 @@
...
+        # @return [Array<String>]
+        # @example
+        #   attributes('a[foo-bar_baz]') # => {"foo-bar_baz=>true}
+        #   attributes('button[foo][bar]') # => {"foo"=>true, "bar"=>true}
+        #   attributes('table[foo=bar]') # => {"foo"=>"'bar'"}
+        def attributes(selector)
+          selector.scan(/\[(.*?)\]/).flatten.map do |attr|
+            key, value = attr.split('=')
+            [key, normalize_value(value)]
+          end.to_h
```
Darhazer, line 42:
```suggestion
          selector.scan(/\[(.*?)\]/).flatten.to_h do |attr|
            key, value = attr.split('=')
            [key, normalize_value(value)]
          end
```
`to_h` accepts a block to construct the key/value pair, no need of map.to_h

lib/rubocop/cop/rspec/mixin/css_selector.rb
```diff
@@ -0,0 +1,87 @@
...
+        # @return [Boolean]
+        # @example
+        #   common_attributes?('a[focused]') # => true
+        #   common_attributes?('button[focused][visible]') # => true
+        #   common_attributes?('table[id=some-id]') # => true
+        #   common_attributes?('h1[invalid]') # => false
+        def common_attributes?(selector)
+          attributes(selector).keys.all? do |attr|
+            COMMON_OPTIONS.include?(attr)
+          end
```
Darhazer, line 55: Less known option would be
```suggestion
          attributes(selector).keys.difference(COMMON_OPTIONS).empty?
```

Darhazer (approved):

lib/rubocop/cop/rspec/mixin/css_selector.rb
ydah: It is certainly better that way. Thank you very much.

ydah:
I updated this PR. Thank you so much.

Darhazer (approved):

lib/rubocop/cop/rspec/capybara/specific_matcher.rb
```diff
@@ -37,11 +38,6 @@ class SpecificMatcher < Base
...
             'input' => 'field'
           }.freeze
-          COMMON_OPTIONS = %w[
-            above below left_of right_of near count minimum maximum between text
-            id class style visible obscured exact exact_text normalize_ws match
-            wait filter_set focused
-          ].freeze
           SPECIFIC_MATCHER_OPTIONS = {
             'button' => (
               COMMON_OPTIONS + %w[disabled name value title type]
```
bquorning, line 43: I am not sure why, but I would prefer adding (part of) the namespace in front of the constant name, e.g.:
```suggestion
              CssSelector::COMMON_OPTIONS + %w[disabled name value title type]
```

lib/rubocop/cop/rspec/mixin/css_selector.rb
```diff
@@ -0,0 +1,85 @@
+# frozen_string_literal: true
+
+module RuboCop
+  module Cop
+    module RSpec
+      # Helps parsing css selector.
+      module CssSelector
```
bquorning, line 7: I really like how you’ve grouped common functionality together in this re-usable module.

But I think we should consider not including this module, but instead make the methods “module functions” using `module_function` (or `extend self`, depending on our RuboCop configuration). I say this because the methods are _very_ specialized, and because none of them depend on the cop context, but operate entirely on the single method argument being passed to them.

bquorning (commented):
If possible, I would like to have 2 commits instead of one: 1st commit to extract `CssSelector` from the `Capybara::SpecificMatcher` class, and the 2nd commit to add the `Capybara::SpecificMatcher` class. What do you think?

lib/rubocop/cop/rspec/capybara/specific_finders.rb
```diff
@@ -0,0 +1,82 @@
...
+
+          include CssSelector
+          include RangeHelp
+
+          MSG = 'Prefer `find_by` over `find`.'
+          RESTRICT_ON_SEND = %i[find].freeze
+
+          # @!method find_argument(node)
+          def_node_matcher :find_argument, <<~PATTERN
+            (send _ :find (str $_)...)
```
pirj, line 29: Cosmetic: add whitespace before `...`

pirj (approved):
Looks good from my perspective, and a clean refactoring. Thank you.

Just in case, do you mind running it against https://github.com/pirj/real-world-rspec and checking offences?

lib/rubocop/cop/rspec/capybara/specific_matcher.rb
ydah: Indeed it is. I add namespace in front of the constant name.

lib/rubocop/cop/rspec/mixin/css_selector.rb
ydah: That certainly seems like a better. If I am mistaken in my approach, please point out. I changed it to `module_function`.

lib/rubocop/cop/rspec/capybara/specific_finders.rb
ydah: Thank you so much. I fixed it.

ydah:
@bquorning 
> If possible, I would like to have 2 commits instead of one

It certainly looks like better that the commits should be separated, so I am splitting the commits into two.

@pirj 
> Just in case, do you mind running it against https://github.com/pirj/real-world-rspec and checking offences?

I would appreciate it if you could running it. Thank you so much.

pirj:
```
57497 files inspected, 336 offenses detected, 336 offenses autocorrectable
```
All offences seem legit.

bquorning (approved):
Thank you 👍🏼

lib/rubocop/cop/rspec/capybara/specific_finders.rb
```diff
@@ -0,0 +1,86 @@
...
+            add_offense(offense_range(node)) do |corrector|
+              corrector.replace(node.loc.selector, 'find_by_id')
+              corrector.replace(node.first_argument.loc.expression,
+                                arg_replacemenet)
+            end
+          end
+
+          def replaced_argments(arg, id)
+            options = to_options(CssSelector.attributes(arg))
+            options.empty? ? id : "#{id}, #{options}"
```
bquorning, line 67: @ydah Can you think of a case where `options` does not end up empty?

lib/rubocop/cop/rspec/capybara/specific_finders.rb
```diff
@@ -0,0 +1,86 @@
...
+          end
+
+          def to_options(attrs)
+            attrs.each.map do |key, value|
+              next if key == 'id'
```
bquorning, line 72: Likewise, there are no spec examples reaching this code where `key` is different from `"id"`.

PR #1341, opened 2022-07-29
Add documentation on cases that does not support autocorrection in cop

ydah:
This PR added an explanation of the cases that does not support autocorrection.

---

* [x] Added tests.
* [x] Updated documentation.

### DIFF
--- a/lib/rubocop/cop/rspec/capybara/current_path_expectation.rb
+++ b/lib/rubocop/cop/rspec/capybara/current_path_expectation.rb
@@ -14,14 +14,20 @@ module RuboCop
         # which ensures that preceding actions (like `click_link`) have
         # completed.
         #
+        # This cop does not support autocorrection in some cases.
+        #
         # @example
         #   # bad
         #   expect(current_path).to eq('/callback')
-        #   expect(page.current_path).to match(/widgets/)
         #
         #   # good
-        #   expect(page).to have_current_path("/callback")
-        #   expect(page).to have_current_path(/widgets/)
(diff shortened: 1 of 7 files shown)


lib/rubocop/cop/rspec/capybara/current_path_expectation.rb
```diff
@@ -14,13 +14,19 @@ module Capybara
...
         # @example
         #   # bad
         #   expect(current_path).to eq('/callback')
-        #   expect(page.current_path).to match(/widgets/)
         #
         #   # good
         #   expect(page).to have_current_path("/callback")
+        #
+        #   # bad (not supported autocorrction)
+        #   expect(page.current_path).to match(/widgets/)
```
Darhazer, line 27: This is being corrected. The one that is not, according to the specs, if where the match is against a variable

ydah: I guess I was mistaken. Thank you very much. I have changed the case to use variables.

pirj (approved):
Looks good, thank you!
I'm not confident if the current wording is correct. I'd go with "does not support autocorrection" or "no autocorrection support".

Might be it would make sense to mention why the the autocorrection is not supported, is it because those cases are unsafe, or is because it's not implemented yet. But this can be a topic for a follow-up.

ydah:
I tried to use _does not support autocorrection_  . What do you think?

Darhazer (approved):

PR #1344, opened 2022-08-01
[Fix #1292] Update `RSpec/MultipleExpectations` cop to search for conditional branches and check the number of `expect`.

### LINKED ISSUE #1292: Don't offense `MultipleExpectations` for `expect`s inside conditions
Hello.

It's probably not common or not easy to implement, but I'd like to not receive an offense from `MultipleExpectations` when `expect`s are in `if`-`else`.

```ruby
shared_examples 'something' do |variable:|
  specify do
    if variable
      expect(subject).to all(part_matching)
    else
      expect(subject).not_to include(part_matching)
    end
  end
end
```

ydah:
Resolve: #1292

```ruby
# good
describe Foo do
  it 'some expectation examples defined in `if` branches' do
    if foo
      expect(foo).to be(true)
    elsif bar
      expect(bar).to be(true)
    else
      expect(baz).to be(true)
    end
  end
end

# bad
describe Foo do
  it 'multiple expectation examples defined in `if` branches' do
    if foo
      expect(foo).to be(true)
      expect(bar).to be(true)
    elsif bar
      expect(bar).to be(true)
    else
      expect(baz).to be(true)
    end
  end
end
```

I would like to address this in a separate PR, as putting the response to which this [comment](https://github.com/rubocop/rubocop-rspec/issues/1292#issuecomment-1143906962) refers in a PR would bloat the PR and make it not a PR for one purpose. What do you think?

---

* [x] Added tests.
* [-] Updated documentation.

### DIFF
--- a/lib/rubocop/cop/rspec/multiple_expectations.rb
+++ b/lib/rubocop/cop/rspec/multiple_expectations.rb
@@ -90,16 +90,13 @@ module RuboCop
 
         def on_block(node)
           return unless example?(node)
-
           return if example_with_aggregate_failures?(node)
 
-          expectations_count = to_enum(:find_expectation, node).count
+          total = expectation_count(node)
+          return if total <= max_expectations
 
-          return if expectations_count <= max_expectations
-
-          self.max = expectations_count
-
-          flag_example(node, expectation_count: expectations_count)
+          self.max = total
+          flag_example(node, total: total)
         end
 
         private
@@ -116,23 +113,40 @@ module RuboCop
             .find { |block_node| aggregate_failures?(block_node, ANYTHING) }
         end
 
-        def find_expectation(node, &block)
-          yield if expect?(node) || aggregate_failures_block?(node)
+        def expectation_count(node, count = 0)
+          # do not search  inside of aggregate_failures blocks
+          return count if aggregate_failures_block?(node)
 
-          # do not search inside of aggregate_failures block
-          return if aggregate_failures_block?(node)
-
-          node.each_child_node do |child|
-            find_expectation(child, &block)
+          if node.if_type? || node.case_type?
+            return expectation_count_max(node, count)
           end
+
+          expectation_count_aggregate(node, count)
         end
 
-        def flag_example(node, expectation_count:)
+        def expectation_count_max(node, count)
+          node.each_child_node.map do |child|
+            expectation_count(child, count)
+          end.max
+        end
+
+        def expectation_count_aggregate(node, count)
+          node.each_child_node do |child|
+            count = if expect?(child) || aggregate_failures_block?(child)
+                      expectation_count(child, count + 1)
+                    else
+                      expectation_count(child, count)
+                    end
+          end
+          count
+        end
+
+        def flag_example(node, total:)
           add_offense(
             node.send_node,
             message: format(
               MSG,
-              total: expectation_count,
+              total: total,
               max: max_expectations
             )
           )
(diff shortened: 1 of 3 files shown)


lib/rubocop/cop/rspec/multiple_expectations.rb
```diff
@@ -116,23 +113,40 @@ def find_aggregate_failures(example_node)
...
+            return expectation_count_max(node, count)
+          end
+
+          expectation_count_aggregate(node, count)
+        end
 
-          # do not search inside of aggregate_failures block
-          return if aggregate_failures_block?(node)
+        def expectation_count_max(node, count)
+          node.each_child_node.map do |child|
```
pirj, line 128: Would `node.each_child_node.max do |child|` work?

lib/rubocop/cop/rspec/multiple_expectations.rb
```diff
@@ -116,23 +113,40 @@ def find_aggregate_failures(example_node)
...
+            expectation_count(child, count)
+          end.max
+        end
 
+        def expectation_count_aggregate(node, count)
           node.each_child_node do |child|
-            find_expectation(child, &block)
+            count = if expect?(child) || aggregate_failures_block?(child)
```
pirj, line 135: Should `aggregate_failures` count towards the limit on its own?

pirj, line 135: If `expect?(child)` is `true`, is it possible that `expectation_count` returns (well, adds to the passed `count`) something else than 1?
Can this be replaced with just `count + 1`?

Those are questions, and not suggestions, because I'm uncertain.

lib/rubocop/cop/rspec/multiple_expectations.rb
```diff
@@ -116,23 +113,40 @@ def find_aggregate_failures(example_node)
             .find { |block_node| aggregate_failures?(block_node, ANYTHING) }
         end
 
-        def find_expectation(node, &block)
-          yield if expect?(node) || aggregate_failures_block?(node)
+        def expectation_count(node, count = 0)
+          # do not search  inside of aggregate_failures blocks
```
pirj, line 117: Cosmetic: double space

pirj (commented):
In #1292 I was hinting that alternative styles exist.
I'm uncertain if the support for conditional branches is a necessity for this cop, given the added complexity.

ydah:
> I'm uncertain if the support for conditional branches is a necessity for this cop, given the added complexity.

Umm. I am beginning to think that this may indeed be the case.

While I understand the idea of not wanting expect in conditional branches to be count, the other side of me does not want to write conditional branches in the test code.

Would it be better to be able to switch by an option?
That way complex ones can be confined if you want to allow for conditional branching.

pirj:
When writing strict spec code that only allows a single expectation per example has a goal of immediately seeing what failed without the need to investigate and debug.
Conditionals in examples undermines that goal.

I'm more inclined to close the original ticket.

Darhazer:
Personally I'd ban conditional in specs - except maybe for skipping some examples in a shared examples group, though even then I'd rather have smaller subsets of examples and including explicitly the ones that are expected

pirj:
Please accept my sincere apologies for the confusing initial "PR is welcome" in the ticket, @ydah , and for the time you've put into this.

I'm more inclined to close this and the ticket at this point.

ydah:
I don't see a problem. @pirj :) 
I too have been discussing this here and was thinking that this PR might as well be closed. (I remembered that I am someone who personally would not want to write a conditional branch on  spec.)

Thank you for your time @pirj @Darhazer .

PR #1345, opened 2022-08-01
Fix a false positive for `RSpec/Capybara/SpecificMatcher` when may not have a `href`

ydah:
This PR is fix a false positive for `RSpec/Capybara/SpecificMatcher` when may not have a `href`.

For example, the following href cannot be obtained with `have_link`.
```html
<a class='cls'>Hello world!</a>
```

If you try to retrieve it with the following code,
you will not be able to retrieve it.
```ruby
expect(page).to have_link(class: "cls")
```

`a` tags without an `href` are not links, they are placeholders for links.
So you cannot get it with `have_link`.

Since the test code does not know if the actually has a `href` or not, it is not offense for matchers that do not specify a href.

Refs: https://togithub.com/teamcapybara/capybara/issues/379#issuecomment-1339947

---

* [x] Added tests.
* [x] Updated documentation.

### DIFF
--- a/lib/rubocop/cop/rspec/capybara/specific_matcher.rb
+++ b/lib/rubocop/cop/rspec/capybara/specific_matcher.rb
@@ -12,7 +12,7 @@ module RuboCop
         #   expect(page).to have_selector('button')
         #   expect(page).to have_no_selector('button.cls')
         #   expect(page).to have_css('button')
-        #   expect(page).to have_no_css('a.cls', exact_text: 'foo')
+        #   expect(page).to have_no_css('a.cls', href: 'http://example.com')
         #   expect(page).to have_css('table.cls')
         #   expect(page).to have_css('select')
         #   expect(page).to have_css('input', exact_text: 'foo')
@@ -21,7 +21,7 @@ module RuboCop
         #   expect(page).to have_button
         #   expect(page).to have_no_button(class: 'cls')
         #   expect(page).to have_button
-        #   expect(page).to have_no_link('foo', class: 'cls')
+        #   expect(page).to have_no_link('foo', class: 'cls', href: 'http://example.com')
         #   expect(page).to have_table(class: 'cls')
         #   expect(page).to have_select
         #   expect(page).to have_field('foo')
@@ -71,11 +71,16 @@ module RuboCop
             (send nil? _ (str $_) ... )
           PATTERN
 
+          # @!method option?(node)
+          def_node_search :option?, <<-PATTERN
+            (pair (sym %) _)
+          PATTERN
+
           def on_send(node)
             return unless (arg = first_argument(node))
             return unless (matcher = specific_matcher(arg))
             return if acceptable_pattern?(arg)
-            return unless specific_matcher_option?(arg, matcher)
+            return unless specific_matcher_option?(node, arg, matcher)
 
             add_offense(node, message: message(node, matcher))
           end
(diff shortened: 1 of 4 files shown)


lib/rubocop/cop/rspec/capybara/specific_matcher.rb
```diff
@@ -71,11 +71,16 @@ class SpecificMatcher < Base
             (send nil? _ (str $_) ... )
           PATTERN
 
+          # @!method options(node)
+          def_node_search :options, <<-PATTERN
+            (pair (sym $_) _)
+          PATTERN
```
Darhazer, line 77: You can turn this into
```suggestion
          def_node_search :option?, <<-PATTERN
            (pair (sym %) _)
          PATTERN
```
And then use 
```
option?(node, :href) || attributes.include?('href')
```

lib/rubocop/cop/rspec/capybara/specific_matcher.rb
```diff
@@ -91,17 +96,24 @@ def acceptable_pattern?(arg)
...
           end
 
-          def specific_matcher_option?(arg, matcher)
+          def specific_matcher_option?(node, arg, matcher)
             # If `button[foo-bar_baz=foo][bar][baz]`:
             # extract ["foo-bar_baz", "bar", "baz"]
             attributes = arg.scan(/\[(.*?)(?:=.*?)?\]/).flatten
             return true if attributes.empty?
+            return false if matcher == 'link' &&
+              !replaceable_to_have_link?(node, attributes)
```
Darhazer, line 105: IMO this would be easier to read if it is
```suggestion
            return false unless replaceable_matcher?(node, matcher, attributes)
```
and this will serve as a placeholder for other matchers, even though right now it would have implementation only for links
```
  def replaceable_matcher?(node, matcher, attributes)
    case matcher
    when 'link' then replaceable_to_have_link?(node, attributes)
    else true
    end
```

spec/rubocop/cop/rspec/capybara/specific_matcher_spec.rb
```diff
@@ -96,15 +96,35 @@
...
+       'for `have_link` without `href`' do
+      expect_no_offenses(<<-RUBY, attr: attr)
         expect(page).to have_css("a[#{attr}=foo]")
-                        ^^^^^^^^^^^^^{attr}^^^^^^^ Prefer `have_link` over `have_css`.
         expect(page).to have_css("a[#{attr}]")
-                        ^^^^^^^^^^^^^{attr}^^^ Prefer `have_link` over `have_css`.
+      RUBY
+    end
+
+    it 'registers an offense for abstract matcher when ' \
```
Darhazer, line 109: For spec completeness we need a[href=] with no other attributes

lib/rubocop/cop/rspec/capybara/specific_matcher.rb
ydah: That looks very much better. Thank you!

lib/rubocop/cop/rspec/capybara/specific_matcher.rb
ydah: It certainly looks good for increased readability. Thank you very much.

spec/rubocop/cop/rspec/capybara/specific_matcher_spec.rb
ydah: Indeed it is. I have added a test: https://github.com/rubocop/rubocop-rspec/blob/906aef1e2e6d1a13379a8326b6d63c3e7dd79594/spec/rubocop/cop/rspec/capybara/specific_matcher_spec.rb#L132-L141

Darhazer (approved):

PR #1350, opened 2022-08-03
[Fix #1337] Update the cop to raise error for insufficient descriptions.

### LINKED ISSUE #1337: Cop Idea: prevent "it works" description
I see too many this:

```ruby
it "works" do
  # ...
end
```

Specs are not just here to prevent bugs, but are, as the name implies, specifications. In other words, a documentation we can use as developer to understand what the code is supposed to do.

So it could be nice to enforce a description of "what it does". Preventing "it works" would be a good start 😋 

```ruby
# bad

it "works" do
  # ...
end

# good

it "marks the task as done" do
  # ...
end
```

akrox58:
Per issue #1337, I tried to add a generic way to the ExampleWordingCop to call out insufficient descriptions. For now, I've used "works" and "throws errors" as that is a phrase I see used often as well. I didn't use a new cop since insufficient descriptions seemed to be along the same lines as "should not start with should or it" as they are both description related.

---

* [X] Added tests.
* [X] Updated documentation.

* [X] Set `VersionChanged` in `config/default.yml` to the next major version.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 * Add `require_implicit` style to `RSpec/ImplicitSubject`. ([@r7kamura][])
 * Fix a false positive for `RSpec/Capybara/SpecificMatcher` when `have_css("a")` without attribute. ([@ydah][])
+* Update `RSpec/ExampleWording` cop to raise error for insufficient descriptions. ([@akrox58][])
 
 ## 2.13.2 (2022-09-23)
 
@@ -650,6 +651,7 @@ Compatibility release so users can upgrade RuboCop to 0.51.0. No new features.
 [@abrom]: https://github.com/abrom
 [@ahukkanen]: https://github.com/ahukkanen
 [@akiomik]: https://github.com/akiomik
+[@akrox58]: https://github.com/akrox58
 [@AlexWayfer]: https://github.com/AlexWayfer
 [@andrykonchin]: https://github.com/andrykonchin
 [@andyw8]: https://github.com/andyw8
--- a/config/default.yml
+++ b/config/default.yml
@@ -194,7 +194,7 @@ RSpec/ChangeByZero:
   Description: Prefer negated matchers over `to change.by(0)`.
   Enabled: pending
   VersionAdded: '2.11'
-  VersionChanged: '2.13'
+  VersionChanged: "<<next>>"
   Reference: https://www.rubydoc.info/gems/rubocop-rspec/RuboCop/Cop/RSpec/ChangeByZero
   NegatedMatcher: ~
 
@@ -383,8 +383,10 @@ RSpec/ExampleWording:
     have: has
     HAVE: HAS
   IgnoredWords: []
+  DisallowedExamples:
+    - works
   VersionAdded: '1.0'
-  VersionChanged: '1.2'
+  VersionChanged: '2.13'
   StyleGuide: https://rspec.rubystyle.guide/#should-in-example-docstrings
   Reference: https://www.rubydoc.info/gems/rubocop-rspec/RuboCop/Cop/RSpec/ExampleWording
 
--- a/lib/rubocop/cop/rspec/example_wording.rb
+++ b/lib/rubocop/cop/rspec/example_wording.rb
@@ -6,11 +6,16 @@ module RuboCop
       # Checks for common mistakes in example descriptions.
       #
       # This cop will correct docstrings that begin with 'should' and 'it'.
+      # This cop will also look for insufficient examples and call them out.
       #
       # @see http://betterspecs.org/#should
       #
       # The autocorrect is experimental - use with care! It can be configured
       # with CustomTransform (e.g. have => has) and IgnoredWords (e.g. only).
+      #
+      # Use the DisallowedExamples setting to prevent unclear or insufficient
+      # descriptions. Please note that this config will not be treated as
+      # case sensitive.
       #
       # @example
       #   # bad
@@ -30,11 +35,21 @@ module RuboCop
       #   it 'does things' do
       #   end
       #
+      # @example `DisallowedExamples: ['works']` (default)
+      #   # bad
+      #   it 'works' do
+      #   end
+      #
+      #   # good
+      #   it 'marks the task as done' do
+      #   end
       class ExampleWording < Base
         extend AutoCorrector
 
         MSG_SHOULD = 'Do not use should when describing your tests.'
         MSG_IT     = "Do not repeat 'it' when describing your tests."
+        MSG_INSUFFICIENT_DESCRIPTION = 'Your example description is ' \
+                                       'insufficient.'
 
         SHOULD_PREFIX = /\Ashould(?:n't)?\b/i.freeze
         IT_PREFIX     = /\Ait /i.freeze
@@ -53,11 +68,19 @@ module RuboCop
               add_wording_offense(description_node, MSG_SHOULD)
             elsif message.match?(IT_PREFIX)
               add_wording_offense(description_node, MSG_IT)
+            else
+              check_and_handle_insufficient_examples(description_node)
             end
           end
         end
 
         private
+
+        def check_and_handle_insufficient_examples(description)
+          if insufficient_examples.include?(preprocess(text(description)))
+            add_wording_offense(description, MSG_INSUFFICIENT_DESCRIPTION)
+          end
+        end
 
         def add_wording_offense(node, message)
           docstring = docstring(node)
@@ -112,6 +135,15 @@ module RuboCop
 
         def ignored_words
           cop_config.fetch('IgnoredWords', [])
+        end
+
+        def insufficient_examples
+          examples = cop_config.fetch('DisallowedExamples', [])
+          examples.map! { |example| preprocess(example) }
+        end
+
+        def preprocess(message)
+          message.strip.squeeze(' ').downcase
         end
       end
     end
--- a/spec/rubocop/cop/rspec/example_wording_spec.rb
+++ b/spec/rubocop/cop/rspec/example_wording_spec.rb
@@ -187,4 +187,65 @@ RSpec.describe RuboCop::Cop::RSpec::ExampleWording do
       end
     RUBY
   end
+
+  it 'flags an unclear description' do
+    expect_offense(<<-'RUBY')
+      it "works" do
+          ^^^^^ Your example description is insufficient.
+      end
+    RUBY
+  end
+
+  it 'flags an unclear description despite extra spaces' do
+    expect_offense(<<-'RUBY')
+      it "  works    " do
+          ^^^^^^^^^^^ Your example description is insufficient.
+      end
+    RUBY
+  end
+
+  it 'flags an unclear description despite uppercase and lowercase strings' do
+    expect_offense(<<-'RUBY')
+      it "WOrKs " do
+          ^^^^^^ Your example description is insufficient.
+      end
+    RUBY
+  end
+
+  context 'when `DisallowedExamples: Workz`' do
+    let(:cop_config) { { 'DisallowedExamples' => ['Workz'] } }
+
+    it 'finds a valid sentence across two lines' do
+      expect_no_offenses(<<-'RUBY')
+        it "workz " \
+          "totally fine " do
+        end
+      RUBY
+    end
+
+    it 'finds an invalid example across two lines' do
+      expect_offense(<<-'RUBY')
+        it "workz" \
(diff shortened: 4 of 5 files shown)


lib/rubocop/cop/rspec/example_wording.rb
```diff
@@ -29,14 +30,34 @@ module RSpec
...
       class ExampleWording < Base
         extend AutoCorrector
 
         MSG_SHOULD = 'Do not use should when describing your tests.'
         MSG_IT     = "Do not repeat 'it' when describing your tests."
+        MSG_INSUFFICIENT_DESCRIPTION = 'Your test should be more descriptive.'
 
         SHOULD_PREFIX = /\Ashould(?:n't)?\b/i.freeze
         IT_PREFIX     = /\Ait /i.freeze
+        INSUFFICIENT_EXAMPLES = ['works', 'throws error'].freeze
```
ydah, line 60: It would be better if it could be changed with a configurable option, as suggested in https://github.com/rubocop/rubocop-rspec/issues/1337#issuecomment-1196650696

lib/rubocop/cop/rspec/example_wording.rb
```diff
@@ -52,6 +73,9 @@ def on_block(node)
               add_wording_offense(description_node, MSG_SHOULD)
             elsif message.match?(IT_PREFIX)
               add_wording_offense(description_node, MSG_IT)
+            elsif INSUFFICIENT_EXAMPLES.include?(message.strip.gsub(/  +/, ' '))
```
bquorning, line 76:
```suggestion
            elsif INSUFFICIENT_EXAMPLES.include?(message.strip.squeeze(' '))
```

spec/rubocop/cop/rspec/example_wording_spec.rb
```diff
@@ -187,4 +187,29 @@
       end
     RUBY
   end
+
+  it 'flags an unclear description' do
+    expect_offense(<<-'RUBY')
+      it "throws error" \
+          ^^^^^^^^^^^^^^^ Your test should be more descriptive.
+        "with #{object}" do
```
pirj, line 195: Those should be examples of what the cop should ignore.
The full message is "throws error with a message". Why is this an offence?

It seems that:
 - we don't consider articles e.g. "throws an error"
 - we don't inspect complete messages, but just the first part of it

pirj (changes requested):

lib/rubocop/cop/rspec/example_wording.rb
akrox58: Thank you for the correction!

lib/rubocop/cop/rspec/example_wording.rb
akrox58: DisallowedExamples:
    - works
    
My concern with adding it as a configurable option is the possibility of how big the list can get 😅 Are we anticipating it to get really big eventually?

spec/rubocop/cop/rspec/example_wording_spec.rb
akrox58: Thanks for bringing up the article part! you are right that we don't consider articles! I went down a rabbit hole of trying to ignore the articles and strip the phrase to just ["throws error"] from ["throws error, throws an error"] etc... But then I realized there could be grammatical errors that get missed from it. so I'm a little confused at what to do. I removed the phrase for the time being though. 

as for #2, I thought include? checks for whole phrases. so we will ignore "throws an error with a message" but the cop will sound an alarm for "throws error". Is there something I'm missing?

lib/rubocop/cop/rspec/example_wording.rb
ydah: I think each project would have different examples of settings you want to disallowed, so wouldn't it be better to allow each project to configure its own settings?

lib/rubocop/cop/rspec/example_wording.rb
```diff
@@ -29,11 +30,21 @@ module RSpec
       #   # good
       #   it 'does things' do
       #   end
+      #
+      # @example
```
ydah, line 34: It would be good to have a good case/bad case according to the setting values of the configuration options.

lib/rubocop/cop/rspec/example_wording.rb
```diff
@@ -112,6 +126,14 @@ def custom_transform
         def ignored_words
           cop_config.fetch('IgnoredWords', [])
         end
+
+        def insufficient_examples
+          cop_config.fetch('DisallowedExamples', [])
+        end
+
+        def clean_msg(message)
+          message.strip.squeeze(' ').downcase
```
ydah, line 135: If I format it with `downcase`, won't it always mismatch if there are uppercase letters in the `DisallowedExample` setting?

akrox58: good callout! we should not allow uppercase in the DisallowedExample setting! Updating!

akrox58: great callout. I've added a comment that the words in this config will not be treated as case sensitive!

lib/rubocop/cop/rspec/example_wording.rb
akrox58: I think I addressed it with my latest changes, but let me know if you were expecting something else!

lib/rubocop/cop/rspec/example_wording.rb
akrox58: ah gotcha!!! makes sense!!

lib/rubocop/cop/rspec/example_wording.rb
ydah: I think it looks good.
Also, would it look good to add the following test?

```ruby
context 'when `DisallowedExamples: Workz`' do
  let(:cop_config) { { 'DisallowedExamples' => ['Workz'] } }

  it 'flags an unclear description' do
    expect_offense(<<-'RUBY')
      it "workz" \
          ^^^^^^^^ Your test should be more descriptive.
        "with #{object}" do
      end
    RUBY
  end

  it 'flags an unclear description despite extra spaces' do
    expect_offense(<<-'RUBY')
      it "  workz    " \
          ^^^^^^^^^^^^^^ Your test should be more descriptive.
        "with #{object}" do
      end
    RUBY
  end

  it 'flags an unclear description despite uppercase and lowercase strings' do
    expect_offense(<<-'RUBY')
      it "WOrKz " \
          ^^^^^^^^^ Your test should be more descriptive.
        "with #{object}" do
      end
    RUBY
  end
end
```

lib/rubocop/cop/rspec/example_wording.rb
ydah: I think the part you added is an explanation of the case where 'it' is repeated. I thinking of the https://github.com/rubocop/rubocop-rspec/pull/1350#discussion_r938271855 changes. For your information.

lib/rubocop/cop/rspec/example_wording.rb
```diff
@@ -21,19 +26,29 @@ module RSpec
...
       #   end
+      #
+      # @example
+      #   # bad
+      #   it 'works' do
+      #   end
+      #
+      #   # good
+      #   it 'marks the task as done' do
+      #   end
```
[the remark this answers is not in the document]
akrox58: Ahhh sorry typo :/

lib/rubocop/cop/rspec/example_wording.rb
akrox58: @ydah  I thought I saw a notification on updating the test cases! that was a good idea! have added that too! 
Appreciate your patience on this! 
PR is ready for review again 😅

ydah: Oops, I thought I had added a comment about the proposed test, but it disappeared. Thank you so much!

ydah (approved):
Looks good to me, thank you.

akrox58:
Thanks @ydah! How do I get this merged? Do I need to wait for the other reviewers to approve?

pirj (changes requested):
I would expect the following to pass, but it fails, why?
```ruby
    expect_no_offenses(<<-'RUBY')
      it "works" \
        "totally fine" do
      end
    RUBY
```

akrox58:
> I would expect the following to pass, but it fails, why?
> 
> ```ruby
>     expect_no_offenses(<<-'RUBY')
>       it "works" \
>         "totally fine" do
>       end
>     RUBY
> ```

Looks like I didn't handle that edge case! updated the check to check for the entire description instead, and added some more test cases! Thanks for the catch!

pirj (approved):
I've checked against https://github.com/pirj/real-world-rspec, no errors, and the cop "just works" 😄 

```
...
/Users/pirj/source/real-world-rspec/sharetribe/spec/models/person_spec.rb:338:9: C: [Correctable] RSpec/ExampleWording: Your test should be more descriptive.
    it 'works' do
        ^^^^^
/Users/pirj/source/real-world-rspec/solidus/core/spec/lib/spree/core/testing_support/factories/address_factory_spec.rb:41:11: C: [Correctable] RSpec/ExampleWording: Your test should be more descriptive.
      it 'works' do
          ^^^^^

57497 files inspected, 273 offenses detected, 273 offenses autocorrectable
```

Looks great, thank you for the contribution!

config/default.yml
```diff
@@ -368,8 +368,10 @@ RSpec/ExampleWording:
     have: has
     HAVE: HAS
   IgnoredWords: []
+  DisallowedExamples:
+    - works
```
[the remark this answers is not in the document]
akrox58: @pirj Is there something I would need to do to address this?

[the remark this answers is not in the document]
pirj: No, it is fine. Just thinking out loud to get an opinion of maintainers.
For me, this change is totally fine as is.

bquorning:

pirj:
Rebased and green @bquorning @Darhazer

CHANGELOG.md
```diff
@@ -38,6 +38,7 @@
 * Fix a false positive for `RSpec/Capybara/SpecificMatcher` when pseudo-classes. ([@ydah][])
 * Fix a false negative for `RSpec/SubjectStub` when the subject is declared with the `subject!` method and called by name. ([@eikes][])
 * Support `Array.new(n)` on `RSpec/FactoryBot/CreateList` cop. ([@r7kamura][])
+* Update `RSpec/ExampleWording` cop to raise error for insufficient descriptions. ([@akrox58][])
```
bquorning, line 41: Could you move this up to the `(Unreleased)` section of the Changelog, please?

lib/rubocop/cop/rspec/example_wording.rb
```diff
@@ -113,6 +135,15 @@ def custom_transform
         def ignored_words
           cop_config.fetch('IgnoredWords', [])
         end
+
+        def insufficient_examples
+          examples = cop_config.fetch('DisallowedExamples', [])
+          examples.map! { |example| preprocess(example) }
```
bquorning, line 141: Won't this line allocate (at least) 3 String objects on each `on_block` call? I don’t know if it would be possible to memoize this at class level.

bquorning (commented):
> This cop will also look for insufficient examples and call them out.

> Use the DisallowedExamples setting to prevent unclear or insufficient descriptions.

> `MSG_INSUFFICIENT_DESCRIPTION = 'Your test should be more descriptive.'`

I think we use a more clear terminology. Internally in rspec-core, the first argument to `it` is called `desc`. So perhaps

> This cop will also look for insufficient example descriptions and call them out.

> Use the DisallowedExampleDescriptions setting to prevent unclear or insufficient example descriptions.

> `MSG_INSUFFICIENT_DESCRIPTION = 'Your example description is insufficient.'`

lib/rubocop/cop/rspec/example_wording.rb
pirj: Since, in theory, different `DisallowedExamples` can be configured in different directories, it would be incorrect to cache this at the class level.
I've made the change to cache it in an instance variable, as I guess the same instance will be used for a single investigation.
We also need to [nullify the instance variable in `on_new_investigation`](https://docs.rubocop.org/rubocop-rspec/upgrade_to_version_2.html#always-call-super-from-on_new_investigation-in-your-cops). Is it worth it?

bquorning: No, probably not worth it. Thanks for investigating.

bquorning (approved):

pirj:
Thank you for the contribution and your time, @akrox58 !

PR #1357, opened 2022-08-05
Add `RSpec/ClassCheck` cop

r7kamura:
It would be nice to have a new cop similar to the `Style/ClassCheck` cop.

- https://docs.rubocop.org/rubocop/cops_style.html#styleclasscheck

---

* [x] Added tests.
* [x] Updated documentation.


lib/rubocop/cop/rspec/class_check.rb
```diff
@@ -0,0 +1,75 @@
...
+            node.loc.selector,
+            opposite_method_name(node.method_name)
+          )
+        end
+
+        def format_message(node)
+          format(
+            MSG,
+            current: node.method_name,
+            preferred: opposite_method_name(node.method_name)
```
ydah, line 61: How about the following?
```suggestion
            preferred: style
```

lib/rubocop/cop/rspec/class_check.rb
```diff
@@ -0,0 +1,64 @@
...
+      #   expect(var).to be_kind_of(String)
+      #
+      class ClassCheck < Base
+        extend AutoCorrector
+        include ConfigurableEnforcedStyle
+
+        MSG = 'Prefer `%<preferred>s` over `%<current>s`.'
+
+        RESTRICT_ON_SEND = %i[
+          be_a
```
Darhazer, line 33: there are also `be_an` and `be_a_kind_of`. While be_a / be_an is being used depending on the argument (e.g. they are the same style), it would be nice in the case of be_kind_of style both to be corrected; and also to support the be_a_kind_of  style.

r7kamura: Good catch. I'll support them too 👍

ydah:
The CI that is failing now is being handled by the following PR Please wait a moment:
https://github.com/rubocop/rubocop-rspec/pull/1362

lib/rubocop/cop/rspec/class_check.rb
```diff
@@ -0,0 +1,88 @@
...
+          corrector.replace(node.loc.selector, style)
+        end
+
+        def format_message(node)
+          format(
+            MSG,
+            current: node.method_name,
+            preferred: style
+          )
+        end
```
bquorning, line 72: I think I am [directly contradicting](https://github.com/rubocop/rubocop-rspec/pull/1357#discussion_r938711003) @ydah when I say that I don’t like using `style` in line 63 and 70. The name of the style and the preferred literal code are _incidentally_ identical, but they are two different concepts. While it’s not a big deal here (we probably won’t change the style names soon) it just doesn’t look right to me.

⬆️ This is just my opinion, not necessarily the truth 😄

bquorning (commented):
Nice addition to our collection of cops. I just had a couple of comments.

lib/rubocop/cop/rspec/class_check.rb
r7kamura: Certainly, you have a point.

One of the reasons I chose to use `style` here is that it is better to calculate preferred method names based on `style`, rather than based on `node`. I think we can probably agree on this point. On top of that, if we want to make it explicit that style and method name are not the same concept, we should put a map from style to preferred method name, even if it seems a bit redundant.

bquorning: I agree that the map would be the best approach. And having it map from style to preferred code (e.g. `{ be_a: 'be_a', be_kind_of: 'be_kind_of' }`) does seem a bit redundant.

bquorning (approved):

Darhazer (commented):
Since there is no restriction on the receiver, only on the method name, such a cop is prone to false positives, e.g. `ResponseMock.be_a(sucessful_response)` would be flagged. Even though there is a low chance for people to use such names, it won't hurt to check that the receiver is nil

spec/rubocop/cop/rspec/class_check_spec.rb
```diff
@@ -0,0 +1,115 @@
...
+        RUBY
+      end
+    end
+  end
+
+  context 'when EnforcedStyle is `be_kind_of`' do
+    context 'when `be_kind_of` is used' do
+      let(:cop_config) do
+        { 'EnforcedStyle' => 'be_kind_of' }
+      end
```
Darhazer, line 60: You can move this config in the parent context, so you don't need to repeat it in the next ones

r7kamura: oh, I forgot to move them. Thank you! 👍

Darhazer (approved):
Thank you

.rubocop.yml
```diff
@@ -106,6 +106,8 @@ RSpec/BeNil:
   Enabled: true
 RSpec/ChangeByZero:
   Enabled: true
+RSpec/ClassCheck:
+  Enabled: true
```
pirj, line 110: Is there a reason this cop is enabled, not `pending`?

pirj: My apologies, I've misread this.

Issue #1356, opened 2022-08-05
Feature Request: Support suffix for RSpec/ContextWording


okuramasafumi:
https://docs.rubocop.org/rubocop-rspec/cops_rspec.html#rspeccontextwording

Currently only `prefix` is supported for `RSpec/ContextWording`, but for some languages, only prefix cannot make description sound natural. I'm a native speaker of Japanese so here I'd like to take Japanese as an example.

```ruby
description '#some_method' do
  # Bad
  '条件1' do
    # When condition 1 is fulfilled
  end

  # Good
  '条件1が満たされるとき' do
    # When condition 1 is fulfilled
  end
end
```

Here, the better context for me is the one that ends with 'とき', meaning `if` or `when` in Japanese.
In general, when rubocop-rspec supports suffix for context wording, it can help writing context in other languages than English. And in some situations, we need to write contexts in our native languages, not English. In addition, `RSpec/ContextWording` doesn't mean that it only supports prefix.
In English, suffix is not important in terms of context. So, the default value for suffix can be empty. With empty default value, we can benefit from restricting suffix while it doesn't effect existing users.

ydah:
It is important to note that in the use case you are proposing, suffixes and prefixes need to behavior differently.
The `RSpec/ContextWording` prefix match checking is strict as follows.

```ruby
# for example - `Prefixes: [when]`
# good
context 'when something' do
end

# bad
context 'whenever something' do
end
```

In Japanese, each word is not separated by a space. Therefore, to check words such as those in the example, the following will work differently for prefix and suffix.

```ruby
# for example - `Suffixes: [if]`
# good
context 'something if' do
end

# good
context 'something serif' do
end
```

> In English, suffix is not important in terms of context. So, the default value for suffix can be empty.

However, as mentioned above, in English, suffixes is not be contextually important, so I think you should be fine.

PR #1358, opened 2022-08-05
Add `Suffixes` configuration option to `RSpec/ContextWording`

ydah:
Resolve: #1356

---

* [x] Added tests.
* [x] Updated documentation.


docs/modules/ROOT/pages/cops_rspec.adoc
```diff
@@ -524,6 +530,34 @@ context 'when the display name is not present' do
...
+
+[source,ruby]
+----
+# bad
+context 'when this test is failed' do
+  # ...
+end
+
+# good
+context 'when this test is success' do
```
pirj, line 551: I'd really prefer this to be in an RTL language, as with LTR it's hard to imagine anyone willing to constrain descriptions to be suffixed with a limited set of words.

bquorning:
Should we consider changing the cop to be configured with regular expressions instead of prefixes and suffixes? Then people can use them however they like. The current defaults would need to be changed to e.g.

```yaml
RSpec/ContextWording:
  Prefixes:
    - "^when\b"
    - "^with\b"
    - "^without\b"
```

ydah:
Certainly we might consider changing the cop to be composed of regular expressions.

Would it be better to allow regular expressions to be set with names like `AllowedPattern` instead of the option names `Prefixes` and `Suffixes`?
In that case, it seems to me that it would be better to make `Prefixes` obsoleted but not removed, and migrate slowly while ensuring backward compatibility.

pirj:
I'd keep `Prefixes` and introduced a `AllowedPatterns` to support both suffixes and prefixes. WDYT?

ydah:
The following PR is needed:
- https://github.com/rubocop/rubocop-rspec/pull/1362

ydah:
Conflicts have been resolved.

docs/modules/ROOT/pages/cops_rspec.adoc
bquorning: I agree, an RTL language example would work much better here.

spec/rubocop/cop/rspec/context_wording_spec.rb
```diff
@@ -1,7 +1,9 @@
 # frozen_string_literal: true
 
 RSpec.describe RuboCop::Cop::RSpec::ContextWording do
-  let(:cop_config) { { 'Prefixes' => %w[when with] } }
+  let(:cop_config) do
+    { 'Prefixes' => %w[when with], 'AllowedPatterns' => %w[] }
```
bquorning, line 5:
```suggestion
    { 'Prefixes' => %w[when with], 'AllowedPatterns' => [] }
```

spec/rubocop/cop/rspec/context_wording_spec.rb
```diff
@@ -125,5 +127,41 @@
...
+        { 'Prefixes' => %w[when with], 'AllowedPatterns' => [/success$/] }
+      end
+
+      it 'finds context without `success` at the ending' do
+        expect_offense(<<-RUBY)
+          context 'when the display name not present' do
+                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ Context description is not descriptions of allowed patterns.
+          end
+        RUBY
+      end
```
bquorning, line 142: I find this behavior confusing. The string starts with `when`, which is an allowed prefix.

I may not have thought this through, but it seems like an offense should be called only when **none** of the `Prefixes` match **and none** of the `AllowedPatterns` match.

spec/rubocop/cop/rspec/context_wording_spec.rb
```diff
@@ -125,5 +127,41 @@
         RUBY
       end
     end
+
+    context "when `AllowedPatterns: ['success']`" do
+      let(:cop_config) do
+        { 'Prefixes' => %w[when with], 'AllowedPatterns' => [/success$/] }
```
bquorning, line 133: An RTL example would work better, I think.

bquorning (changes requested):
I think we also need a spec for when Prefixes is empty, but AllowedPatterns is not.

docs/modules/ROOT/pages/cops_rspec.adoc
ydah: If it is an RTL language, it appears that if /olloeh$/ is specified, it will correctly check if it ends in olloeh as an RTL language.

```ruby
# bad
context '!sknaht olleh'
end

# good
context 'olleh !sknaht'
end
```

How about using Japanese as an example, as in the case of the original Issue?

spec/rubocop/cop/rspec/context_wording_spec.rb
```diff
@@ -125,5 +129,72 @@
...
+          context '条件を満たす' do
+                  ^^^^^^^^ Context description should match '(?-mix:とき$)'.
+          end
+        RUBY
+      end
+
+      it 'finds shared_context without `とき` at the ending' do
+        expect_offense(<<-RUBY)
+          shared_context '条件を満たす' do
+                         ^^^^^^^^ Context description should match '(?-mix:とき$)'.
```
ydah, line 149: It seems that when multibyte characters are used here, the caret does not display properly according to the number of characters. This seems to be another issue.

bquorning: We use [unicode-display_width](https://github.com/janlelis/unicode-display_width) which “Determines the monospace display width of a string in Ruby.” Perhaps we don’t use it properly in RuboCop?

> This seems to be another issue.

Yep. I think just adding enough carets to make the specs pass in the correct approach for now 😅

bquorning (commented):
I am sorry for stalling this pull request, but there are still a few things I would like to change 😅

I think we should stop treating the Prefixes as anything special. I think if we override the `#allowed_patterns` method to include the prefixes, we can get rid of `PREFIX_MSG`, `BOTH_MSG` and a bunch of methods. E.g.

```rb
def allowed_patterns
  @allowed_patterns ||= begin
    prefix_regexes = prefixes.map { |pre| /^#{Regexp.escape(pre)}\b/ }
    super.concat(prefix_regexes)
  end
end
```

(Without the memoization it seems that the list grows unbounded. I didn’t have time to check _why_ that happens)

Another thing to be aware of is that “[#inspect actually produces the more natural version of the [Regex] string than #to_s](https://ruby-doc.org/core-2.7.5/Regexp.html#method-i-inspect)”. So, by using `"'#{value.inspect}'"` instead of `"'#{value}'"`, we can have offense messages with `'/とき$/'` instead of `'(?-mix:とき$)'`. (That's also why I chose to `map` the `Regexp.escape`d prefixes – that way they look better in offense messages).

This is of course up for discussion. I’m looking forward to hearing your opinion.

ydah:
> I am sorry for stalling this pull request, but there are still a few things I would like to change 😅

If it makes it better, even if it takes longer, it's worth it ;-)
I sincerely appreciate your always thoughtful reviews.

The approach of overriding the `#allowed_patterns` method without special attention to Prefixes seems so good.
It also keeps the code very simple and seems to improve the outlook of the `ContextWording` class, which was a bit bloated.
This would also unify the messages and make judging very simple.

I also think it would be a good idea to use `Object#inspect` in the violation message.
In that case, there is no need to further quoted string like a `#{"'v.inspect'"}`, so I put the result of `Object#inspect` directly into the offense message, what do you think?

lib/rubocop/cop/rspec/context_wording.rb
```diff
@@ -35,41 +35,71 @@ module RSpec
...
 
         private
 
-        def bad_prefix?(description)
-          !prefix_regex.match?(description)
+        def allowed_patterns
+          @allowed_patterns ||= begin
+            prefix_regexes = prefixes.map { |pre| /^#{Regexp.escape(pre)}\b/ }
+            super.concat(prefix_regexes)
+          end
```
bquorning, line 84: I was looking into _why_ the memoization was necessary, and found that `cop_config["AllowedMethods"]` is the same object as `Array(cop_config["AllowedMethods"])`, and that `concat` is a mutating method. And then I found rubocop/rubocop#10953 with your name on it 😄

bquorning: Perhaps we should follow the new pattern and do e.g.
```rb
def allowed_patterns
  super + prefix_regexes
end

def prefix_regexes
  @prefix_regexes ||= begin
    prefixes = Array(cop_config.fetch('Prefixes', []))
    prefixes.map { |pre| /^#{Regexp.escape(pre)}\b/ }
  end
end
```

ydah: Wow. almost made the same mistake again😅 Thanks for catching that.

bquorning (approved):
Thank you so much @ydah 🙏🏼

config/default.yml
```diff
@@ -212,8 +212,9 @@ RSpec/ContextWording:
     - when
     - with
     - without
+  AllowedPatterns: []
```
Darhazer, line 215: Should we have default allowed patterns, matching the prefixes, and deprecate the prefixes option?

spec/rubocop/cop/rspec/context_wording_spec.rb
```diff
@@ -13,15 +15,15 @@
   it 'finds context without `when` at the beginning' do
     expect_offense(<<-RUBY)
       context 'the display name not present' do
-              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ Start context description with 'when', or 'with'.
+              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ Context description should match /^when\\b/, or /^with\\b/.
```
Darhazer, line 18: with love to see an example with 3 patterns

Darhazer (approved):

config/default.yml
ydah: I am certainly beginning to think that is a good idea.
@bquorning @pirj What do you think?

spec/rubocop/cop/rspec/context_wording_spec.rb
ydah: I updated it
https://github.com/rubocop/rubocop-rspec/blob/c017ffd20d2b78f7fda8b129d1aa85ad80edac70/spec/rubocop/cop/rspec/context_wording_spec.rb#L15-L21

Darhazer: Thank you

lib/rubocop/cop/rspec/context_wording.rb
```diff
@@ -35,41 +35,72 @@ module RSpec
...
+      #   # good
+      #   context '条件を満たすとき' do
+      #     # ...
+      #   end
+      #
       class ContextWording < Base
-        MSG = 'Start context description with %<prefixes>s.'
+        include AllowedPattern
+
+        MSG = 'Context description should match %<patterns>s.'
```
pirj, line 62: I'm uncertain that I'm happy with this change.
From my perspective, the cognitive gap between "Start description with" and "Description should match /@$&#%&$$/" is noticeable.

Wondering if it's easy to use the old message when only `Prefixes` are defined, and the new one when `AllowedPatterns` are defined, too.
Or a mix of like "Context description should either start with %<prefixes> or match %<patterns>".

pirj (commented):
Happy to see the cop works for RTL now, too.
Just a small optional subjective note, maybe an idea for a follow-up if it gets some support.

PR #1362, opened 2022-08-16
Fix `InternalAffairs/NumblockHandler` offenses

ydah:
In this PR, corrections were made to the following internal cop additions where they were in offense.
- https://github.com/rubocop/rubocop/pull/10915

This PR correction of each offense for each cop in separate commits, just as we did with the original PR.

---

* [-] Squashed related commits together.
* [x] Added tests.
* [-] Updated documentation.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@
 * Fix a false positive for `RSpec/Capybara/SpecificMatcher` when may not have a `href` by `have_link`. ([@ydah][])
 * Add `NegatedMatcher` configuration option to `RSpec/ChangeByZero`. ([@ydah][])
 * Add new `RSpec/Capybara/SpecificFinders` cop. ([@ydah][])
+* Add support for numblocks to `RSpec/AroundBlock`, `RSpec/EmptyLineAfterHook`, `RSpec/ExpectInHook`, `RSpec/HookArgument`, `RSpec/HooksBeforeExamples`, `RSpec/IteratedExpectation`, and `RSpec/NoExpectationExample`. ([@ydah][])
 
 ## 2.12.1 (2022-07-03)
 
--- a/lib/rubocop/cop/rspec/around_block.rb
+++ b/lib/rubocop/cop/rspec/around_block.rb
@@ -30,9 +30,14 @@ module RuboCop
         MSG_UNUSED_ARG = 'You should call `%<arg>s.call` ' \
                          'or `%<arg>s.run`.'
 
-        # @!method hook(node)
-        def_node_matcher :hook, <<-PATTERN
+        # @!method hook_block(node)
+        def_node_matcher :hook_block, <<-PATTERN
           (block (send nil? :around sym ?) (args $...) ...)
+        PATTERN
+
+        # @!method hook_numblock(node)
+        def_node_matcher :hook_numblock, <<-PATTERN
+          (numblock (send nil? :around sym ?) ...)
         PATTERN
 
         # @!method find_arg_usage(node)
@@ -41,12 +46,18 @@ module RuboCop
         PATTERN
 
         def on_block(node)
-          hook(node) do |(example_proxy)|
+          hook_block(node) do |(example_proxy)|
             if example_proxy.nil?
               add_no_arg_offense(node)
             else
               check_for_unused_proxy(node, example_proxy)
             end
+          end
+        end
+
+        def on_numblock(node)
+          hook_numblock(node) do
+            check_for_numblock(node)
           end
         end
 
@@ -66,6 +77,17 @@ module RuboCop
           add_offense(
             proxy,
             message: format(MSG_UNUSED_ARG, arg: name)
+          )
+        end
+
+        def check_for_numblock(block)
+          find_arg_usage(block) do |usage|
+            return if usage.include?(s(:lvar, :_1))
+          end
+
+          add_offense(
+            block.children.last,
+            message: format(MSG_UNUSED_ARG, arg: :_1)
           )
         end
       end
--- a/lib/rubocop/cop/rspec/expect_in_hook.rb
+++ b/lib/rubocop/cop/rspec/expect_in_hook.rb
@@ -26,7 +26,7 @@ module RuboCop
         # @!method expectation(node)
         def_node_search :expectation, send_pattern('#Expectations.all')
 
-        def on_block(node)
+        def on_block(node) # rubocop:disable InternalAffairs/NumblockHandler
           return unless hook?(node)
           return if node.body.nil?
 
@@ -35,6 +35,8 @@ module RuboCop
                         message: message(expect, node))
           end
         end
+
+        alias on_numblock on_block
 
         private
 
--- a/lib/rubocop/cop/rspec/factory_bot/create_list.rb
+++ b/lib/rubocop/cop/rspec/factory_bot/create_list.rb
@@ -71,7 +71,7 @@ module RuboCop
             (send {nil? #factory_bot?} :create_list (sym _) (int $_) ...)
           PATTERN
 
-          def on_block(node)
+          def on_block(node) # rubocop:todo InternalAffairs/NumblockHandler
             return unless style == :create_list
 
             return unless n_times_block?(node)
--- a/lib/rubocop/cop/rspec/hook_argument.rb
+++ b/lib/rubocop/cop/rspec/hook_argument.rb
@@ -66,11 +66,13 @@ module RuboCop
 
         # @!method scoped_hook(node)
         def_node_matcher :scoped_hook, <<-PATTERN
-          (block $(send _ #Hooks.all (sym ${:each :example})) ...)
+          ({block numblock} $(send _ #Hooks.all (sym ${:each :example})) ...)
         PATTERN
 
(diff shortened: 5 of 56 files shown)


bquorning:
Can you briefly explain why the cop in some cases is disabled, and in some cases you can just add an `alias`? It has something to do with the number or type of arguments being passed to the block?

ydah:
For the following cop that I added aliases to, I targeted the ones that take block arguments.
- https://github.com/rubocop/rubocop-rspec/pull/1362/commits/d6b9ba6666909e5d911246a83ccaaf5b47b276a0
- https://github.com/rubocop/rubocop-rspec/pull/1362/commits/09bf5c17e1d18eee43e3871d75a16a8e7a2c9721

All cop that use `on_block` are subject to `InternalAffairs/NumblockHandler`, so most of the cop that check describe and context are in offense. But I don't think they should be represented as numblock in the first place, so I disabled them.

bquorning (approved):
@Darhazer, would you have a look as well?

lib/rubocop/cop/rspec/expect_in_hook.rb
```diff
@@ -26,7 +26,7 @@ class ExpectInHook < Base
         # @!method expectation(node)
         def_node_search :expectation, send_pattern('#Expectations.all')
 
-        def on_block(node)
+        def on_block(node) # rubocop:disable InternalAffairs/NumblockHandler
```
Darhazer, line 29: again, hook could be a numblock

lib/rubocop/cop/rspec/factory_bot/create_list.rb
```diff
@@ -71,7 +71,7 @@ class CreateList < Base
             (send {nil? #factory_bot?} :create_list (sym _) (int $_) ...)
           PATTERN
 
-          def on_block(node)
+          def on_block(node) # rubocop:disable InternalAffairs/NumblockHandler
```
Darhazer, line 74: create_list yields, so it could be used with a numblock. Though I guess the cop doesn't really support that. Perhaps we have to open an issue to add support

lib/rubocop/cop/rspec/hook_argument.rb
```diff
@@ -72,7 +72,7 @@ class HookArgument < Base
         # @!method unscoped_hook(node)
         def_node_matcher :unscoped_hook, '(block $(send _ #Hooks.all) ...)'
 
-        def on_block(node)
+        def on_block(node) # rubocop:disable InternalAffairs/NumblockHandler
```
Darhazer, line 75: hook could be a numblock

lib/rubocop/cop/rspec/hooks_before_examples.rb
```diff
@@ -36,7 +36,7 @@ class HooksBeforeExamples < Base
           }
         PATTERN
 
-        def on_block(node)
+        def on_block(node) # rubocop:disable InternalAffairs/NumblockHandler
```
Darhazer, line 39: hook could be a numblock

lib/rubocop/cop/rspec/iterated_expectation.rb
```diff
@@ -33,7 +33,7 @@ class IteratedExpectation < Base
           (send (send nil? :expect (lvar %)) :to ...)
         PATTERN
 
-        def on_block(node)
+        def on_block(node) # rubocop:disable InternalAffairs/NumblockHandler
```
Darhazer, line 36: Perhaps we need to add numblock handling here as well

lib/rubocop/cop/rspec/yield.rb
```diff
@@ -26,7 +26,7 @@ class Yield < Base
         # @!method block_call?(node)
         def_node_matcher :block_call?, '(send (lvar %) :call ...)'
 
-        def on_block(node)
+        def on_block(node) # rubocop:disable InternalAffairs/NumblockHandler
```
Darhazer, line 29: this also needs support for numblocks

lib/rubocop/cop/rspec/factory_bot/create_list.rb
ydah: Indeed it is. we will mark it as a rubocop:todo comment.

lib/rubocop/cop/rspec/yield.rb
ydah: Perhaps, but I think that this cop considers an offense when passing a block argument, so it is always a not offense with numbered parameters.

ydah:
I update this PR. Thank you so much.

lib/rubocop/cop/rspec/around_block.rb
```diff
@@ -29,19 +29,26 @@ class AroundBlock < Base
         MSG_NO_ARG     = 'Test object should be passed to around block.'
         MSG_UNUSED_ARG = 'You should call `%<arg>s.call` ' \
                          'or `%<arg>s.run`.'
+        MSG_NUMBLOCK = 'You should call `_1.call` ' \
```
Darhazer, line 32: `MSG_UNUSED_ARG` covers this, if you pass `_1` as an `arg`

lib/rubocop/cop/rspec/iterated_expectation.rb
```diff
@@ -28,6 +28,13 @@ class IteratedExpectation < Base
           )
         PATTERN
 
+        # @!method each_numblock?(node)
+        def_node_matcher :each_numblock?, <<-PATTERN
+          (numblock
+            (send ... :each) 1 $(...)
```
Darhazer, line 34: hm, what does the 1 means in the pattern 🤔

Darhazer (commented):
Since this upgrades some of the cops to handle numb locks, it deserves a changeling entry

lib/rubocop/cop/rspec/iterated_expectation.rb
ydah: It is possible that my matcher is not written in the right way.
If the code is as follows:
```ruby
it 'validates users' do
  [user1, user2, user3].each { expect(_1).to be_valid }
end
```

It is expressed in AST as follows:
```
s(:numblock,
  s(:send,
    s(:array,
      s(:send, nil, :user1),
      s(:send, nil, :user2),
      s(:send, nil, :user3)), :each), 1, // 👈
  s(:send,
    s(:send, nil, :expect,
      s(:lvar, :_1)), :to,
    s(:send, nil, :be_valid)))
```

so adjusting for this `1` 🤔

ydah: It is changed to `_`. What do you think?

ydah:
I update this PR. Thank you so much.

lib/rubocop/cop/rspec/around_block.rb
ydah: Indeed it is. change to `MSG_UNUSED_ARG `

lib/rubocop/cop/rspec/iterated_expectation.rb
Darhazer: Hah, it seems `1` indicates the number of arguments (the highest numbered argument) used.
```
[user1, user2, user3].each { expect(_1).to be_valid(_2) }
```
produces `2` instead
```
[user1, user2, user3].each { expect(_1).to be_valid(_3) }
```
results in a 3.

if we care that just _1 is used, then the original version is fine. If we don't care, then it's `_`

CHANGELOG.md
```diff
@@ -13,6 +13,7 @@
 * Fix a false positive for `RSpec/Capybara/SpecificMatcher` when may not have a `href` by `have_link`. ([@ydah][])
 * Add `NegatedMatcher` configuration option to `RSpec/ChangeByZero`. ([@ydah][])
 * Add new `RSpec/Capybara/SpecificFinders` cop. ([@ydah][])
+* Add support numblocks for `RSpec/AroundBlock`, `RSpec/EmptyLineAfterHook`, `RSpec/ExpectInHook`, `RSpec/HookArgument`, `RSpec/HooksBeforeExamples` `RSpec/IteratedExpectation` and `RSpec/NoExpectationExample`. ([@ydah][])
```
Darhazer, line 16: `Add support numblocks for` -> `Add support for numblocks to`

Darhazer: `RSpec/HooksBeforeExamples` `RSpec/IteratedExpectation` and `RSpec/NoExpectationExample` ->
`RSpec/HooksBeforeExamples`,`RSpec/IteratedExpectation`, and `RSpec/NoExpectationExample`

Darhazer (commented):
Final finishing touches and we're ready to go 🚀

CHANGELOG.md
ydah: Thank you very much. I've corrected those.

lib/rubocop/cop/rspec/iterated_expectation.rb
ydah: I see. I understand very well, thank you.
we are going to use `_` here.

Darhazer (approved):
Thank you 🙇  🚀 :shipit:

bquorning (approved):
Looks fine as far as I can tell.

Issue #110, opened 2016-06-16
RSpec/DescribeClass applies to view specs when using config.infer_spec_type_from_file_location!


jcoyne:
#90 only fixed it for when the spec is explictly tagged as being a view type.

backus:
@jcoyne unless you have an idea of how to statically determine this configuration I'm going to close this as WONTFIX since the cop cannot know your RSpec configuration

andyw8:
I agree, since inference is disabled by default after RSpec 3.

PR #1365, opened 2022-08-19
Add `RSpec/Rails/InferredSpecType` cop

r7kamura:
In many Rails apps, I often see test code like `RSpec.describe MyModel, type: :model` even though `config.infer_spec_type_from_file_location!` is enabled.

This makes it difficult to distinguish whether they are being overwritten as needed or written even though they are not needed.

To solve this problem, it woule be nice to have a cop that issues offenses for redundant type metadata.

- https://relishapp.com/rspec/rspec-rails/docs/directory-structure

---

* [x] Added tests.
* [x] Updated documentation.

### DIFF
--- a/config/default.yml
+++ b/config/default.yml
@@ -981,6 +981,29 @@ RSpec/Rails/HaveHttpStatus:
   VersionAdded: '2.12'
   Reference: https://www.rubydoc.info/gems/rubocop-rspec/RuboCop/Cop/RSpec/Rails/HaveHttpStatus
 
+RSpec/Rails/InferredSpecType:
+  Description: Identifies redundant spec type.
+  Enabled: pending
+  Safe: false
+  VersionAdded: '2.14'
+  Reference: https://www.rubydoc.info/gems/rubocop-rspec/RuboCop/Cop/RSpec/Rails/InferredSpecType
+  Inferences:
+    channels: channel
+    controllers: controller
+    features: feature
+    generator: generator
+    helpers: helper
+    jobs: job
+    mailboxes: mailbox
+    mailers: mailer
+    models: model
+    requests: request
+    integration: request
+    api: request
+    routing: routing
+    system: system
+    views: view
+
 RSpec/Rails/HttpStatus:
   Description: Enforces use of symbolic or numeric value to describe HTTP status.
   Enabled: true
--- a/lib/rubocop/cop/rspec/rails/inferred_spec_type.rb
+++ b/lib/rubocop/cop/rspec/rails/inferred_spec_type.rb
@@ -73,14 +73,9 @@ module RuboCop
           #   @return [RuboCop::AST::PairNode, nil]
           def_node_matcher :describe_with_type, <<~PATTERN
             (block
-              (send
-                { (const nil? :RSpec) | nil? }
-                _
-                _
-                _*
-                ({ hash | kwargs }
-                  <$(pair (sym :type) (sym _)) ...>
-                )
+              (send #rspec? #ExampleGroups.all
+                ...
+                (hash <$(pair (sym :type) sym) ...>)
               )
               ...
             )
--- a/lib/rubocop/cop/rspec/rails/inferred_spec_type.rb
+++ b/lib/rubocop/cop/rspec/rails/inferred_spec_type.rb
@@ -1,0 +1,135 @@
+# frozen_string_literal: true
+
+module RuboCop
+  module Cop
+    module RSpec
+      module Rails
+        # Identifies redundant spec type.
+        #
+        # @safety
+        #   This cop is marked as unsafe because
+        #   `config.infer_spec_type_from_file_location!` may not be enabled.
+        #
+        # @example
+        #   # bad
+        #   # spec/models/user_spec.rb
+        #   RSpec.describe User, type: :model do
+        #   end
+        #
+        #   # good
+        #   # spec/models/user_spec.rb
+        #   RSpec.describe User do
+        #   end
+        #
+        #   # good
+        #   # spec/models/user_spec.rb
+        #   RSpec.describe User, type: :common do
+        #   end
+        #
+        # @example `Inferences` configuration
+        #   # .rubocop.yml
+        #   # RSpec/InferredSpecType:
+        #   #   Inferences:
+        #   #     services: service
+        #
+        #   # bad
+        #   # spec/services/user_spec.rb
+        #   RSpec.describe User, type: :service do
+        #   end
+        #
+        #   # good
+        #   # spec/services/user_spec.rb
+        #   RSpec.describe User do
+        #   end
+        #
+        #   # good
+        #   # spec/services/user_spec.rb
+        #   RSpec.describe User, type: :common do
+        #   end
+        class InferredSpecType < Base
+          extend AutoCorrector
+
+          MSG = 'Remove redundant spec type.'
+
+          # @param [RuboCop::AST::BlockNode] node
+          def on_block(node)
+            return unless example_group?(node)
+
+            pair_node = describe_with_type(node)
+            return unless pair_node
+            return unless inferred_type?(pair_node)
+
+            removable_node = detect_removable_node(pair_node)
+            add_offense(removable_node) do |corrector|
+              autocorrect(corrector, removable_node)
+            end
+          end
+          alias on_numblock on_block
+
+          private
+
+          # @!method describe_with_type(node)
+          #   @param [RuboCop::AST::BlockNode] node
+          #   @return [RuboCop::AST::PairNode, nil]
+          def_node_matcher :describe_with_type, <<~PATTERN
+            (block
+              (send
+                { (const nil? :RSpec) | nil? }
+                _
+                _
+                _*
+                ({ hash | kwargs }
+                  <$(pair (sym :type) (sym _)) ...>
+                )
+              )
+              ...
+            )
+          PATTERN
+
+          # @param [RuboCop::AST::Corrector] corrector
+          # @param [RuboCop::AST::Node] node
+          def autocorrect(corrector, node)
+            corrector.remove(
+              node.location.expression.with(
+                begin_pos: node.left_sibling.location.expression.end_pos
+              )
+            )
+          end
+
+          # @param [RuboCop::AST::PairNode] node
+          # @return [RuboCop::AST::Node]
+          def detect_removable_node(node)
+            if node.parent.pairs.size == 1
+              node.parent
+            else
+              node
+            end
+          end
+
+          # @return [String]
+          def file_path
+            processed_source.file_path
+          end
+
+          # @param [RuboCop::AST::PairNode] node
+          # @return [Boolean]
+          def inferred_type?(node)
+            inferred_type_from_file_path.inspect == node.value.source
+          end
+
+          # @return [Symbol, nil]
+          def inferred_type_from_file_path
+            inferences.find do |prefix, type|
+              break type.to_sym if file_path.include?("spec/#{prefix}/")
+            end
+          end
+
+          # @return [Hash]
+          def inferences
+            cop_config['Inferences'] || {}
+          end
+        end
+      end
+    end
+  end
+end
--- a/lib/rubocop/cop/rspec/rails/inferred_spec_type.rb
+++ b/lib/rubocop/cop/rspec/rails/inferred_spec_type.rb
@@ -6,6 +6,11 @@ module RuboCop
       module Rails
         # Identifies redundant spec type.
         #
+        # After setting up rspec-rails, you will have enabled
+        # `config.infer_spec_type_from_file_location!` by default in
+        # spec/rails_helper.rb. This cop works in conjunction with this config.
+        # If you disable this config, disable this cop as well.
+        #
         # @safety
         #   This cop is marked as unsafe because
         #   `config.infer_spec_type_from_file_location!` may not be enabled.
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -136,3 +136,5 @@ RSpec/Rails/AvoidSetupHook:
   Enabled: true
 RSpec/Rails/HaveHttpStatus:
   Enabled: true
+RSpec/Rails/InferredSpecType:
+  Enabled: true
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 * Add `AllowedPatterns` configuration option to `RSpec/NoExpectationExample`.  ([@ydah][])
 * Improve `RSpec/NoExpectationExample` cop to ignore examples skipped or pending via metatada.  ([@pirj][])
 * Add `RSpec/FactoryBot/ConsistentParenthesesStyle` cop. ([@Liberatys][])
+* Add `RSpec/Rails/InferredSpecType` cop. ([@r7kamura][])
 
 ## 2.13.2 (2022-09-23)
 
--- a/docs/modules/ROOT/pages/cops.adoc
+++ b/docs/modules/ROOT/pages/cops.adoc
@@ -113,5 +113,6 @@
 * xref:cops_rspec_rails.adoc#rspecrails/avoidsetuphook[RSpec/Rails/AvoidSetupHook]
 * xref:cops_rspec_rails.adoc#rspecrails/havehttpstatus[RSpec/Rails/HaveHttpStatus]
 * xref:cops_rspec_rails.adoc#rspecrails/httpstatus[RSpec/Rails/HttpStatus]
+* xref:cops_rspec_rails.adoc#rspecrails/inferredspectype[RSpec/Rails/InferredSpecType]
 
 // END_COP_LIST
--- a/docs/modules/ROOT/pages/cops_rspec_rails.adoc
+++ b/docs/modules/ROOT/pages/cops_rspec_rails.adoc
@@ -121,3 +121,81 @@ it { is_expected.to have_http_status :error }
 === References
 
 * https://www.rubydoc.info/gems/rubocop-rspec/RuboCop/Cop/RSpec/Rails/HttpStatus
+
+== RSpec/Rails/InferredSpecType
+
+|===
+| Enabled by default | Safe | Supports autocorrection | Version Added | Version Changed
+
+| Pending
+| No
+| Yes (Unsafe)
+| 2.14
+| -
+|===
+
+Identifies redundant spec type.
+
+=== Safety
+
+This cop is marked as unsafe because
+`config.infer_spec_type_from_file_location!` may not be enabled.
+
+=== Examples
+
+[source,ruby]
+----
+# bad
+# spec/models/user_spec.rb
+RSpec.describe User, type: :model do
+end
+
+# good
+# spec/models/user_spec.rb
+RSpec.describe User do
+end
+
+# good
+# spec/models/user_spec.rb
+RSpec.describe User, type: :common do
+end
+----
+
+==== `Inferences` configuration
+
+[source,ruby]
+----
+# .rubocop.yml
+# RSpec/InferredSpecType:
+#   Inferences:
+#     services: service
+
+# bad
+# spec/services/user_spec.rb
+RSpec.describe User, type: :service do
+end
+
+# good
+# spec/services/user_spec.rb
+RSpec.describe User do
+end
+
+# good
+# spec/services/user_spec.rb
+RSpec.describe User, type: :common do
+end
+----
+
+=== Configurable attributes
+
+|===
+| Name | Default value | Configurable values
+
+| Inferences
+| `{"channels"=>"channel", "controllers"=>"controller", "features"=>"feature", "generator"=>"generator", "helpers"=>"helper", "jobs"=>"job", "mailboxes"=>"mailbox", "mailers"=>"mailer", "models"=>"model", "requests"=>"request", "integration"=>"request", "api"=>"request", "routing"=>"routing", "system"=>"system", "views"=>"view"}`
+| 
+|===
+
+=== References
+
+* https://www.rubydoc.info/gems/rubocop-rspec/RuboCop/Cop/RSpec/Rails/InferredSpecType
--- a/docs/modules/ROOT/pages/cops_rspec_rails.adoc
+++ b/docs/modules/ROOT/pages/cops_rspec_rails.adoc
@@ -136,6 +136,11 @@ it { is_expected.to have_http_status :error }
 
 Identifies redundant spec type.
 
+After setting up rspec-rails, you will have enabled
+`config.infer_spec_type_from_file_location!` by default in
+spec/rails_helper.rb. This cop works in conjunction with this config.
+If you disable this config, disable this cop as well.
+
 === Safety
 
 This cop is marked as unsafe because
--- a/lib/rubocop/cop/rspec_cops.rb
+++ b/lib/rubocop/cop/rspec_cops.rb
@@ -20,6 +20,7 @@ begin
 rescue LoadError
   # Rails/HttpStatus cannot be loaded if rack/utils is unavailable.
 end
+require_relative 'rspec/rails/inferred_spec_type'
 
 require_relative 'rspec/align_left_let_brace'
 require_relative 'rspec/align_right_let_brace'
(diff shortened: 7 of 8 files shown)


lib/rubocop/cop/rspec/rails/inferred_spec_type.rb
```diff
@@ -0,0 +1,127 @@
...
+
+          # @param [RuboCop::AST::PairNode] node
+          # @return [Boolean]
+          def inferred_type?(node)
+            inferred_type_from_file_path.inspect == node.value.source
+          end
+
+          # @return [Symbol, nil]
+          def inferred_type_from_file_path
+            INFERENCES.find do |inference|
```
bquorning, line 119: Would it be possible and safe to use `RSpec::Rails::DIRECTORY_MAPPINGS` instead? ([ref](https://github.com/rspec/rspec-rails/blob/6b4969ef8e20cbd21c51840c158b91f90630041c/lib/rspec/rails/configuration.rb#L28)).

r7kamura: Indeed, I wrote this definition based on that part of rspec-rails. 
By "use `RSpec::Rails::DIRECTORY_MAPPINGS`" do you mean that we should `require "rspec-rails"` and use it here? Or does it mean that we should try to define a Hash of the same format?

bquorning: > do you mean that we should `require "rspec-rails"` and use it here?

Yes, much like [how the Rails/HttpStatus cop is loaded](https://github.com/rubocop/rubocop-rspec/blob/master/lib/rubocop/cop/rspec_cops.rb#L16-L20).

Perhaps first we should hear the opinion of @rubocop/rubocop-rspec if it’s a good idea.

pirj: Directory mappings can be reconfigured in `RSpec.configure`, and using the defaults will break the principle of the least surprise. I don't think there's a good way of loading RSpec configuration in RuboCop's runtime. E.g. we ended up duplicating aliases configuration in `RSpec/Language` in `.rubocop.yml`.

There's also an `@api private` note in RSpec Rails, meaning it can change any moment. And we would have to support that. And reasonably, we can't add an explicit runtime dependency constraint on specific RSpec Rails versions, so once RSpec Rails change it, this will might stop working properly.

bquorning: Could we offload the configuration to our users? And then simply copy-paste the default values into config/default.yml?

pirj:
Those two comments suggest that automatic inferring is legacy [[1](https://github.com/rspec/rspec-rails/pull/1541#issuecomment-177342647), [2](https://github.com/rspec/rspec-rails/issues/2507#issuecomment-1107794265)].
I don't have an idea if `infer_spec_type_from_file_location!` will be deprecated or removed and when.
`define_derived_metadata` provides the same, and at some point it may appear in RSpec Rails' `spec/rails_helper.rb` generator and would replace `infer_spec_type_from_file_location!`.

With this in mind, I'm on the fence with the decision what to do with this cop, as it doesn't seem to enforce the routine that RSpec suggests.

bquorning:

r7kamura:
Then, it might be better to provide two styles, one that omits the type whenever possible according to `infer_spec_type_from_file_location!` and another that always specifies it explicitly.

lib/rubocop/cop/rspec/rails/inferred_spec_type.rb
```diff
@@ -0,0 +1,127 @@
...
+        #   # spec/models/user_spec.rb
+        #   RSpec.describe User, type: :model
+        #
+        #   # good
+        #   # spec/models/user_spec.rb
+        #   RSpec.describe User
+        #
+        #   # good
+        #   # spec/models/user_spec.rb
+        #   RSpec.describe User, type: :request
```
pirj, line 25: Maybe s/request/common to avoid cognitive dissonance?

lib/rubocop/cop/rspec/rails/inferred_spec_type.rb
```diff
@@ -0,0 +1,127 @@
...
+        #
+        class InferredSpecType < Base
+          extend AutoCorrector
+
+          # @return [Array<Hash>]
+          INFERENCES = [
```
pirj, line 31: I'd make this configurable so that the cop could detect a redundant `type: :service` in `spec/service/` specs.

lib/rubocop/cop/rspec/rails/inferred_spec_type.rb
```diff
@@ -0,0 +1,127 @@
...
+            { prefix: 'spec/api/', type: :request },
+            { prefix: 'spec/routing/', type: :routing },
+            { prefix: 'spec/system/', type: :system },
+            { prefix: 'spec/views/', type: :view }
+          ].freeze
+
+          MSG = 'Remove redundant spec type.'
+
+          RESTRICT_ON_SEND = %i[
+            describe
```
pirj, line 52: Technically, any example group alias can be used there.

lib/rubocop/cop/rspec/rails/inferred_spec_type.rb
```diff
@@ -0,0 +1,127 @@
...
+            (send
+              { (const nil? :RSpec) | nil? }
+              _
+              _
+              _*
+              ({ hash | kwargs }
+                (pair ...)*
+                $(pair (sym :type) (sym _))
+                (pair ...)*
+              )
```
pirj, line 82: I think that this can be simplified down to `(hash <$(pair (sym :type) sym) ...>))` with [match in any order node pattern feature](https://docs.rubocop.org/rubocop-ast/node_pattern.html#for-match-in-any-order) as we do in e.g. lib/rubocop/cop/rspec/focus.rb

config/default.yml
```diff
@@ -954,6 +954,13 @@ RSpec/Rails/HaveHttpStatus:
   VersionAdded: '2.12'
   Reference: https://www.rubydoc.info/gems/rubocop-rspec/RuboCop/Cop/RSpec/Rails/HaveHttpStatus
 
+RSpec/Rails/InferredSpecType:
+  Description: Identifies redundant spec type.
+  Enabled: pending
+  Safe: false
+  VersionAdded: '2.13'
```
pirj, line 961: `<<next>>`, since it would be 2.14 already

lib/rubocop/cop/rspec/rails/inferred_spec_type.rb
```diff
@@ -0,0 +1,127 @@
...
+          ].freeze
+
+          MSG = 'Remove redundant spec type.'
+
+          RESTRICT_ON_SEND = %i[
+            describe
+          ].freeze
+
+          # @param [RuboCop::AST::SendNode] node
+          def on_send(node)
```
pirj, line 56: Is it sufficient to only check ones that are passed to `on_top_level_group`?

config/default.yml
r7kamura: That's right. It's been a bit long time since I first made this pull request.
`<<next>>` is only mentioned at rubocop gem, so I thought that was the only convention there, but can it also be used rubocop-rspec gem?

lib/rubocop/cop/rspec/rails/inferred_spec_type.rb
r7kamura: `#on_top_level_example_group` seems to solve several problems, but I'm stuck with a problem that autocorrect not working properly when using it. It seems that the node obtained from `#on_top_level_example_group` refers to a different buffer instance from the corrector expects. Is there any workaround for this?

```diff
diff --git a/lib/rubocop/cop/rspec/rails/inferred_spec_type.rb b/lib/rubocop/cop/rspec/rails/inferred_spec_type.rb
index 56f3005..374b759 100644
--- a/lib/rubocop/cop/rspec/rails/inferred_spec_type.rb
+++ b/lib/rubocop/cop/rspec/rails/inferred_spec_type.rb
@@ -49,14 +49,12 @@ module RuboCop
         class InferredSpecType < Base
           extend AutoCorrector
 
-          MSG = 'Remove redundant spec type.'
+          include TopLevelGroup
 
-          RESTRICT_ON_SEND = %i[
-            describe
-          ].freeze
+          MSG = 'Remove redundant spec type.'
 
-          # @param [RuboCop::AST::SendNode] node
-          def on_send(node)
+          # @param [RuboCop::AST::BlockNode] node
+          def on_top_level_example_group(node)
             pair_node = describe_with_type(node)
             return unless pair_node
             return unless inferred_type?(pair_node)
@@ -70,17 +68,20 @@ module RuboCop
           private
 
           # @!method describe_with_type(node)
-          #   @param [RuboCop::AST::SendNode] node
+          #   @param [RuboCop::AST::BlockNode] node
           #   @return [RuboCop::AST::PairNode, nil]
           def_node_matcher :describe_with_type, <<~PATTERN
-            (send
-              { (const nil? :RSpec) | nil? }
-              _
-              _
-              _*
-              ({ hash | kwargs }
-                <$(pair (sym :type) (sym _)) ...>
+            (block
+              (send
+                { (const nil? :RSpec) | nil? }
+                _
+                _
+                _*
+                ({ hash | kwargs }
+                  <$(pair (sym :type) (sym _)) ...>
+                )
               )
+              ...
             )
           PATTERN
 ```

```
$ bundle exec rspec spec/rubocop/cop/rspec/rails/inferred_spec_type_spec.rb

Randomized with seed 2324

RuboCop::Cop::RSpec::Rails::InferredSpecType
  with necessary type in keyword arguments
    does not register any offense
  with redundant type and other Hash metadata
    autocorrects offense (FAILED - 1)
  with redundant type in Hash arguments
    autocorrects offense (FAILED - 2)
  with redundant type and receiver-less describe
    autocorrects offense (FAILED - 3)
  with redundant type and other Symbol metadata
    autocorrects offense (FAILED - 4)
  with Inferrences configuration
    autocorrects offense (FAILED - 5)
  with redundant type in keyword arguments
    autocorrects offense (FAILED - 6)
  with default configuration
    does not crash on smoke test: spec/smoke_tests/factory_bot_spec.rb
    does not crash on smoke test: spec/smoke_tests/empty_spec.rb
    does not crash on smoke test: spec/smoke_tests/weird_rspec_spec.rb
    does not crash on smoke test: spec/smoke_tests/no_tests_spec.rb

Failures:

  1) RuboCop::Cop::RSpec::Rails::InferredSpecType with redundant type and other Hash metadata autocorrects offense
     Failure/Error:
       corrector.remove(
         node.location.expression.with(
           begin_pos: node.left_sibling.location.expression.end_pos
         )
       )
     
     RuntimeError:
       Correction target buffer 4240 name:"/path/to/project/spec/models/user_spec.rb" is not current 4260 name:"/path/to/project/spec/models/user_spec.rb" under investigation
     # ./lib/rubocop/cop/rspec/rails/inferred_spec_type.rb:91:in `autocorrect'
     # ./lib/rubocop/cop/rspec/rails/inferred_spec_type.rb:64:in `block in on_top_level_example_group'
     # ./lib/rubocop/cop/rspec/rails/inferred_spec_type.rb:63:in `on_top_level_example_group'
     # ./lib/rubocop/cop/rspec/mixin/top_level_group.rb:14:in `block in on_new_investigation'
     # ./lib/rubocop/cop/rspec/mixin/top_level_group.rb:13:in `each'
     # ./lib/rubocop/cop/rspec/mixin/top_level_group.rb:13:in `on_new_investigation'
     # ./spec/rubocop/cop/rspec/rails/inferred_spec_type_spec.rb:51:in `block (3 levels) in <top (required)>'

  2) RuboCop::Cop::RSpec::Rails::InferredSpecType with redundant type in Hash arguments autocorrects offense
     Failure/Error:
       corrector.remove(
         node.location.expression.with(
           begin_pos: node.left_sibling.location.expression.end_pos
         )
       )
     
     RuntimeError:
       Correction target buffer 4280 name:"/path/to/project/spec/models/user_spec.rb" is not current 4300 name:"/path/to/project/spec/models/user_spec.rb" under investigation
     # ./lib/rubocop/cop/rspec/rails/inferred_spec_type.rb:91:in `autocorrect'
     # ./lib/rubocop/cop/rspec/rails/inferred_spec_type.rb:64:in `block in on_top_level_example_group'
     # ./lib/rubocop/cop/rspec/rails/inferred_spec_type.rb:63:in `on_top_level_example_group'
     # ./lib/rubocop/cop/rspec/mixin/top_level_group.rb:14:in `block in on_new_investigation'
     # ./lib/rubocop/cop/rspec/mixin/top_level_group.rb:13:in `each'
     # ./lib/rubocop/cop/rspec/mixin/top_level_group.rb:13:in `on_new_investigation'
     # ./spec/rubocop/cop/rspec/rails/inferred_spec_type_spec.rb:36:in `block (3 levels) in <top (required)>'

  3) RuboCop::Cop::RSpec::Rails::InferredSpecType with redundant type and receiver-less describe autocorrects offense
     Failure/Error:
       corrector.remove(
         node.location.expression.with(
           begin_pos: node.left_sibling.location.expression.end_pos
         )
       )
     
     RuntimeError:
       Correction target buffer 4320 name:"/path/to/project/spec/models/user_spec.rb" is not current 4340 name:"/path/to/project/spec/models/user_spec.rb" under investigation
     # ./lib/rubocop/cop/rspec/rails/inferred_spec_type.rb:91:in `autocorrect'
     # ./lib/rubocop/cop/rspec/rails/inferred_spec_type.rb:64:in `block in on_top_level_example_group'
     # ./lib/rubocop/cop/rspec/rails/inferred_spec_type.rb:63:in `on_top_level_example_group'
     # ./lib/rubocop/cop/rspec/mixin/top_level_group.rb:14:in `block in on_new_investigation'
     # ./lib/rubocop/cop/rspec/mixin/top_level_group.rb:13:in `each'
     # ./lib/rubocop/cop/rspec/mixin/top_level_group.rb:13:in `on_new_investigation'
     # ./spec/rubocop/cop/rspec/rails/inferred_spec_type_spec.rb:81:in `block (3 levels) in <top (required)>'

  4) RuboCop::Cop::RSpec::Rails::InferredSpecType with redundant type and other Symbol metadata autocorrects offense
     Failure/Error:
       corrector.remove(
         node.location.expression.with(
           begin_pos: node.left_sibling.location.expression.end_pos
         )
       )
     
     RuntimeError:
       Correction target buffer 4360 name:"/path/to/project/spec/models/user_spec.rb" is not current 4380 name:"/path/to/project/spec/models/user_spec.rb" under investigation
     # ./lib/rubocop/cop/rspec/rails/inferred_spec_type.rb:91:in `autocorrect'
     # ./lib/rubocop/cop/rspec/rails/inferred_spec_type.rb:64:in `block in on_top_level_example_group'
     # ./lib/rubocop/cop/rspec/rails/inferred_spec_type.rb:63:in `on_top_level_example_group'
     # ./lib/rubocop/cop/rspec/mixin/top_level_group.rb:14:in `block in on_new_investigation'
     # ./lib/rubocop/cop/rspec/mixin/top_level_group.rb:13:in `each'
     # ./lib/rubocop/cop/rspec/mixin/top_level_group.rb:13:in `on_new_investigation'
     # ./spec/rubocop/cop/rspec/rails/inferred_spec_type_spec.rb:66:in `block (3 levels) in <top (required)>'

  5) RuboCop::Cop::RSpec::Rails::InferredSpecType with Inferrences configuration autocorrects offense
     Failure/Error:
       corrector.remove(
         node.location.expression.with(
           begin_pos: node.left_sibling.location.expression.end_pos
         )
       )
     
     RuntimeError:
       Correction target buffer 4400 name:"/path/to/project/spec/services/user_spec.rb" is not current 4420 name:"/path/to/project/spec/services/user_spec.rb" under investigation
     # ./lib/rubocop/cop/rspec/rails/inferred_spec_type.rb:91:in `autocorrect'
     # ./lib/rubocop/cop/rspec/rails/inferred_spec_type.rb:64:in `block in on_top_level_example_group'
     # ./lib/rubocop/cop/rspec/rails/inferred_spec_type.rb:63:in `on_top_level_example_group'
     # ./lib/rubocop/cop/rspec/mixin/top_level_group.rb:14:in `block in on_new_investigation'
     # ./lib/rubocop/cop/rspec/mixin/top_level_group.rb:13:in `each'
     # ./lib/rubocop/cop/rspec/mixin/top_level_group.rb:13:in `on_new_investigation'
     # ./spec/rubocop/cop/rspec/rails/inferred_spec_type_spec.rb:104:in `block (3 levels) in <top (required)>'

  6) RuboCop::Cop::RSpec::Rails::InferredSpecType with redundant type in keyword arguments autocorrects offense
     Failure/Error:
       corrector.remove(
         node.location.expression.with(
           begin_pos: node.left_sibling.location.expression.end_pos
         )
       )
     
     RuntimeError:
       Correction target buffer 4440 name:"/path/to/project/spec/models/user_spec.rb" is not current 4460 name:"/path/to/project/spec/models/user_spec.rb" under investigation
     # ./lib/rubocop/cop/rspec/rails/inferred_spec_type.rb:91:in `autocorrect'
     # ./lib/rubocop/cop/rspec/rails/inferred_spec_type.rb:64:in `block in on_top_level_example_group'
     # ./lib/rubocop/cop/rspec/rails/inferred_spec_type.rb:63:in `on_top_level_example_group'
     # ./lib/rubocop/cop/rspec/mixin/top_level_group.rb:14:in `block in on_new_investigation'
     # ./lib/rubocop/cop/rspec/mixin/top_level_group.rb:13:in `each'
     # ./lib/rubocop/cop/rspec/mixin/top_level_group.rb:13:in `on_new_investigation'
     # ./spec/rubocop/cop/rspec/rails/inferred_spec_type_spec.rb:21:in `block (3 levels) in <top (required)>'

Finished in 0.06853 seconds (files took 0.58951 seconds to load)
11 examples, 6 failures

Failed examples:

rspec ./spec/rubocop/cop/rspec/rails/inferred_spec_type_spec.rb:44 # RuboCop::Cop::RSpec::Rails::InferredSpecType with redundant type and other Hash metadata autocorrects offense
rspec ./spec/rubocop/cop/rspec/rails/inferred_spec_type_spec.rb:29 # RuboCop::Cop::RSpec::Rails::InferredSpecType with redundant type in Hash arguments autocorrects offense
rspec ./spec/rubocop/cop/rspec/rails/inferred_spec_type_spec.rb:74 # RuboCop::Cop::RSpec::Rails::InferredSpecType with redundant type and receiver-less describe autocorrects offense
rspec ./spec/rubocop/cop/rspec/rails/inferred_spec_type_spec.rb:59 # RuboCop::Cop::RSpec::Rails::InferredSpecType with redundant type and other Symbol metadata autocorrects offense
rspec ./spec/rubocop/cop/rspec/rails/inferred_spec_type_spec.rb:97 # RuboCop::Cop::RSpec::Rails::InferredSpecType with Inferrences configuration autocorrects offense
rspec ./spec/rubocop/cop/rspec/rails/inferred_spec_type_spec.rb:14 # RuboCop::Cop::RSpec::Rails::InferredSpecType with redundant type in keyword arguments autocorrects offense

Randomized with seed 2324
```

File: config/default.yml (hunk 6 above)
Comment (pirj) [reply config/default.yml]: We're mimicking this convention, since not uncommon that PRs miss their initially targeted releases.
However, we don't formally describe this practice. I couldn't find this mentioned in rubocop, can you please help me find it, apart from the new cop generator spec? I'll be happy to document this properly.

File: lib/rubocop/cop/rspec/rails/inferred_spec_type.rb (hunk 7 above)
Comment (pirj) [reply lib/rubocop/cop/rspec/rails/inferred_spec_type.rb]:
```ruby
        RSpec.describe User do
          describe 'inner', type: :model do
```
Makes total sense to me 👍 
This is a sufficient argument for me to drop the idea to only check top-level example groups.

config/default.yml
```diff
@@ -955,6 +955,29 @@ RSpec/Rails/HaveHttpStatus:
   VersionAdded: '2.12'
   Reference: https://www.rubydoc.info/gems/rubocop-rspec/RuboCop/Cop/RSpec/Rails/HaveHttpStatus
 
+RSpec/Rails/InferredSpecType:
+  Description: Identifies redundant spec type.
+  Enabled: pending
+  Safe: false
+  VersionAdded: '2.14'
+  Reference: https://www.rubydoc.info/gems/rubocop-rspec/RuboCop/Cop/RSpec/Rails/InferredSpecType
+  Inferrences:
```
pirj, line 964: Please correct me if I'm mistaken, this option needs metadata to prevent from [being overriden by the user setting](https://docs.rubocop.org/rubocop/1.36/configuration.html#inheritance-of-hashes-vs-other-types):

> Other types, such as AllCops / Include (an array), are overridden by the child setting.

> However, advanced users can still merge arrays using the inherit_mode setting. See "Merging arrays using inherit_mode" below.

Please see [an example of how we do this for the language config](https://github.com/rubocop/rubocop-rspec/blob/87f3ef5e6b3df4a388901905f53f3bba7f0d3a13/config/default.yml#L10).

[Semi-related issue in `rubocop` repo](https://github.com/rubocop/rubocop/issues/9325).

I don't think there is a way to check that the defaults are still there from our specs, since we just stub `cop_config` and override, too.

pirj (approved):
~Needs a small change to `config/default.yml` to prevent overriding the default list of inferences. Otherwise -~

Just perfect, thank you so much!

config/default.yml
pirj, line 964: My language checker suggests that the noun has single "r", while the verb/proverb - a double "rr".

pirj: > Inferences is a Hash

I've missed that fact altogether.
Thanks for double-checking with a test project!
Please accept my apologies for bothering.

lib/rubocop/cop/rspec/rails/inferred_spec_type.rb
pirj: I've simplified the pattern slightly

lib/rubocop/cop/rspec/rails/inferred_spec_type.rb
```diff
@@ -0,0 +1,130 @@
+# frozen_string_literal: true
+
+module RuboCop
+  module Cop
+    module RSpec
+      module Rails
+        # Identifies redundant spec type.
```
bquorning, line 7: Should we add a little more information about `infer_spec_type_from_file_location!` here? I think the `@safety` notice is not enough on its own.

pirj: @r7kamura do you have time to address this last note?

bquorning (approved):

pirj:
Thank you, @r7kamura !

mvz:
> Those two comments suggest that automatic inferring is legacy [[1](https://github.com/rspec/rspec-rails/pull/1541#issuecomment-177342647), [2](https://github.com/rspec/rspec-rails/issues/2507#issuecomment-1107794265)]. I don't have an idea if `infer_spec_type_from_file_location!` will be deprecated or removed and when. `define_derived_metadata` provides the same, and at some point it may appear in RSpec Rails' `spec/rails_helper.rb` generator and would replace `infer_spec_type_from_file_location!`.
> 
> With this in mind, I'm on the fence with the decision what to do with this cop, as it doesn't seem to enforce the routine that RSpec suggests.

Was this consideration resolved? I'm reluctant to let this cop autocorrect my specs because of this.

davideluque:
Note that view specs are usually described as a string (e.g. `RSpec.describe "users/new.html.erb"`).

#90 fixed an issue with the `RSpec/DescribeClass` cop - when argument `type: :view` is present, the cop should not register the `RSpec/DescribeClass offense`.

In #110 it is discussed that even though the `infer_spec_type_from_file_location!` is active, the cop still registers an `RSpec/DescribeClass` offense for view specs.

Since rubocop can't read RSpec's configuration, either:

- Keep `type: :view` and disable the `RSpec/Rails/InferredSpecType` for view specs

```
RSpec/Rails/InferredSpecType:
  Exclude:
    - 'spec/views/**'
```

or,

- Disable `RSpec/DescribeClass` for view specs

```
RSpec/DescribeClass:
  Exclude:
    - 'spec/views/**'
```

Open to hearing other approaches to this problem. Maybe a test could be added to cover future changes in these two cops.

mvz:
@davideluque I have simply disabled the entire InferredSpecType cop, since spec type inference is supposed to be deprecated.

knapo:
Personally I'm quite confused that you're adding a Rails specific thing to `rubocop-rspec`. `infer_spec_type_from_file_location!` comes from `rspec-rails` and not everyone uses ruby with rails. IMO you should introduce `rubocop-rspec-rails` to define such cops.

knapo:
Fair enough! Thanks for the explanation @pirj

PR #1367, opened 2022-08-22
Fix cop header comment

ydah:
The following unification/fixes were implemented.
- Line break after `@example`
- Good/Bad -> good/bad
- better -> good
- Add empty comment line at the end of header comment
- Fixed indentation errors
- Fixed function header errors

The indentation will also be modified on the document as follows.

### after

---

* [-] Added tests.
* [x] Updated documentation.

### DIFF
--- a/docs/modules/ROOT/pages/cops_rspec.adoc
+++ b/docs/modules/ROOT/pages/cops_rspec.adoc
@@ -1993,8 +1993,7 @@ Checks for before/around/after hooks that come after an example.
 
 [source,ruby]
 ----
-# Bad
-
+# bad
 it 'checks what foo does' do
   expect(foo).to be
 end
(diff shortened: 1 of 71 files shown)


bquorning:
While all these changes look 👍🏼 and consistent, I cannot help but think we need … a linter :-) Do we have anything available, or would we need to build something ourselves?

Also @ydah can you show _why_ the changes are necessary with e.g. screenshots of how it renders before after indentation change, etc.?

ydah:
> While all these changes look 👍🏼 and consistent, I cannot help but think we need … a linter :-) Do we have anything available, or would we need to build something ourselves?

I too would like to leave this kind of work to the linter ;-)
But, I have no idea of a good Linter right now.

bquorning (approved):

PR #1372, opened 2022-08-28
Add `require_implicit` style to `RSpec/ImplicitSubject`

r7kamura:
Although the explicit style is recommended by default, some users may prefer the implicit style. For such users, it would be better if there is an option to make the style consistent with the explicit style, rather than asking them to disable this cop.

---

* [x] Added tests.
* [x] Updated documentation.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Master (Unreleased)
 
+* Add `require_implicit` style to `RSpec/ImplicitSubject`. ([@r7kamura][])
+
 ## 2.13.2 (2022-09-23)
 
 * Fix an error for `RSpec/Capybara/SpecificFinders` with no parentheses. ([@ydah][])
--- a/config/default.yml
+++ b/config/default.yml
@@ -492,8 +492,9 @@ RSpec/ImplicitSubject:
     - single_line_only
     - single_statement_only
     - disallow
+    - require_implicit
   VersionAdded: '1.29'
-  VersionChanged: '1.30'
+  VersionChanged: '2.13'
   Reference: https://www.rubydoc.info/gems/rubocop-rspec/RuboCop/Cop/RSpec/ImplicitSubject
 
 RSpec/InstanceSpy:
--- a/lib/rubocop/cop/rspec/implicit_subject.rb
+++ b/lib/rubocop/cop/rspec/implicit_subject.rb
@@ -42,12 +42,45 @@ module RuboCop
       #   # good
       #   it { expect(subject).to be_truthy }
       #
+      # @example `EnforcedStyle: require_implicit`
+      #   # bad
+      #   it { expect(subject).to be_truthy }
+      #
+      #   # good
+      #   it { is_expected.to be_truthy }
+      #
+      #   # bad
+      #   it do
+      #     expect(subject).to be_truthy
+      #   end
+      #
+      #   # good
+      #   it do
+      #     is_expected.to be_truthy
+      #   end
+      #
+      #   # good
+      #   it { expect(named_subject).to be_truthy }
+      #
       class ImplicitSubject < Base
         extend AutoCorrector
         include ConfigurableEnforcedStyle
 
-        MSG = "Don't use implicit subject."
-        RESTRICT_ON_SEND = %i[is_expected should should_not].freeze
+        MSG_REQUIRE_EXPLICIT = "Don't use implicit subject."
+
+        MSG_REQUIRE_IMPLICIT = "Don't use explicit subject."
+
+        RESTRICT_ON_SEND = %i[
+          expect
+          is_expected
+          should
+          should_not
+        ].freeze
+
+        # @!method explicit_unnamed_subject?(node)
+        def_node_matcher :explicit_unnamed_subject?, <<-PATTERN
+          (send nil? :expect (send nil? :subject))
+        PATTERN
 
         # @!method implicit_subject?(node)
         def_node_matcher :implicit_subject?, <<-PATTERN
@@ -55,8 +88,7 @@ module RuboCop
         PATTERN
 
         def on_send(node)
-          return unless implicit_subject?(node)
-          return if valid_usage?(node)
+          return unless invalid?(node)
 
           add_offense(node) do |corrector|
             autocorrect(corrector, node)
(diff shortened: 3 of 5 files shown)


r7kamura:
I think the name of the style "always" is debatable. Since this cop is called ImplicitSubject, I named it "always" as in "always use implicit subject."

config/default.yml
```diff
@@ -479,8 +479,9 @@ RSpec/ImplicitSubject:
     - single_line_only
     - single_statement_only
     - disallow
+    - always
```
Darhazer, line 482: The rest of the styles either disallow the usage of implicit subject or allow it in certain situations. This doesn't just allows the usage, but requires it. Note that single_line_only doesn't require the use of `is_expected` in single line examples, it merely allows it. So if we are to add this option, I'd go with something like `require_implicit`. Or it may be extracted in separate option (`require_implicit`) and then `always` will be one that either always allow (require_implicit being false) or always requires (if set to true)

lib/rubocop/cop/rspec/implicit_subject.rb
```diff
@@ -42,21 +42,36 @@ module RSpec
...
-        RESTRICT_ON_SEND = %i[is_expected should should_not].freeze
+        RESTRICT_ON_SEND = %i[
+          expect
+          is_expected
+          should
+          should_not
+        ].freeze
+
+        # @!method explicit_subject?(node)
+        def_node_matcher :explicit_subject?, <<-PATTERN
```
Darhazer, line 64: Note that this way we won't flag usages of a named subject, e.g. code like
```
subject(:instance) { described_class.new }

it { expect(instance).to be_valid }
```

will still be allowed

config/default.yml
r7kamura: `require_implicit` seems to be more preferable. Changed it that way 👍

lib/rubocop/cop/rspec/implicit_subject.rb
```diff
@@ -42,21 +42,36 @@ module RSpec
       #   # good
       #   it { expect(subject).to be_truthy }
       #
+      # @example `EnforcedStyle: require_implicit`
+      #   # bad
+      #   it { expect(subject).to be_truthy }
```
pirj, line 47: As far as I understand, it would also apply to multi-line, multi-statement examples?

pirj, line 47: Another thing that is unsettling about this bad example is that it is using an unnamed subject. I understand that other examples do, too.
I'm uncertain how to clarify that even if the subject is named, it will be enforced to be turned into `is_expected`.

r7kamura:
I added a multi-line example and then an example about named subject.

I figured that if named subject is used, it would be better to ignore it. Or should we also flag named subject as an offense and prompt to convert to unnamed subject?

This is my opinion, but I personally think it would be better to keep this cop only to decide whether `is_expected` should be used or `expect(subject)` should be used. Rather than having this cop decide what the named subject should be, it would be better to have a separate cop that decides whether the named subject should be recommended or not, and have it decide there.

Darhazer:
I tend to agree that if there is named subject in use, it's better to be left explicit, as you can't use the name with the implicit syntax, and there might be a reason you have chosen a name.

pirj:
> if there is named subject in use, it's better to be left explicit, as you can't use the name with the implicit syntax, and there might be a reason you have chosen a name.

Furthermore, there can be more than one:

```ruby
subject(:foo) { ... }
context '' do
  subject(:bar) { ... }
  it do
    expect(foo).to ...
  end
```
and we'll end up with a task to properly detect if the used explicit subject is also the right implicit one.

lib/rubocop/cop/rspec/implicit_subject.rb
```diff
@@ -66,32 +94,70 @@ def on_send(node)
...
-        def valid_usage?(node)
-          example = node.ancestors.find { |parent| example?(parent) }
-          return false if example.nil?
-
-          example.method?(:its) || allowed_by_style?(example)
+        def message(_node)
+          if style == :require_implicit
+            "Don't use explicit subject."
+          else
+            "Don't use implicit subject."
```
bquorning, line 113: I am not sure why, but the convention is to store message strings in `MSG_*` constants.

lib/rubocop/cop/rspec/implicit_subject.rb
```diff
@@ -66,32 +94,70 @@ def on_send(node)
...
-          corrector.replace(node.loc.selector, replacement)
         end
 
-        def valid_usage?(node)
-          example = node.ancestors.find { |parent| example?(parent) }
-          return false if example.nil?
-
-          example.method?(:its) || allowed_by_style?(example)
+        def message(_node)
+          if style == :require_implicit
```
bquorning, line 110: Just to make the implementation parallel with `#invalid?`:
```suggestion
          case style
          when :require_implicit
```
…if RuboCop allows it.

lib/rubocop/cop/rspec/implicit_subject.rb
```diff
@@ -66,32 +94,70 @@ def on_send(node)
...
+          implicit_subject?(node) && !its?(node)
+        end
+
+        def invalid_to_single_line_only?(node)
+          implicit_subject?(node) && !its?(node) && !single_line?(node)
+        end
+
+        def invalid_to_single_statement_only?(node)
+          implicit_subject?(node) && !its?(node) && !single_statement?(node)
+        end
```
bquorning, line 144: It requires a lot of "jumping" with the eyes to read this implementation. Would it be ok to have each of these one-line methods inlined into the `#invalid?` method?

r7kamura: Unfortunately we can't do that because it violates both `Metrics/AbcSize` and `Metrics/CyclomaticComplexity` cops. This is why I splitted it into these methods.

lib/rubocop/cop/rspec/implicit_subject.rb
r7kamura: Oh, that's right. I'll change it like that 👍

lib/rubocop/cop/rspec/implicit_subject.rb
bquorning: The only other alternative I can think of is
```rb
def invalid?(node)
  send("invalid_to_#{style}?", node)
end
``` 
which I don’t really like. So let’s keep it as it is then 😄 👍🏼

bquorning:

lib/rubocop/cop/rspec/implicit_subject.rb
Darhazer: We have separate cop for dealing with wether subject is named or not. If the subject is named however, it would be forced to be `is_expected`

lib/rubocop/cop/rspec/implicit_subject.rb
```diff
@@ -66,32 +98,71 @@ def on_send(node)
...
+          else
+            MSG_REQUIRE_EXPLICIT
+          end
         end
 
-        def allowed_by_style?(example)
+        def invalid?(node)
           case style
+          when :require_implicit
+            invalid_to_require_implicit?(node)
```
Darhazer, line 125: We need some better naming. E.g. 
invalid_to_require_implicit? -> explicit_unnamed_subject? (directly, without intermediate method, as the name is quite clear)
invalid_to_disallow? -> invalid_implicit_subject?
invalid_to_single_line_only? -> invalid_implicit_subject? && multiline?
invalid_to_single_statement_only? -> invalid_implicit_subject? && multi_statement?

Darhazer: if the && creates too much complexity we can move those to methods, but my point is to name them closer to what they are

lib/rubocop/cop/rspec/implicit_subject.rb
```diff
@@ -66,32 +98,71 @@ def on_send(node)
...
+            invalid_to_single_statement_only?(node)
+          end
+        end
+
+        def invalid_to_require_implicit?(node)
+          explicit_unnamed_subject?(node)
+        end
+
+        def invalid_to_disallow?(node)
+          implicit_subject?(node) && !its?(node)
```
Darhazer, line 140: this definition should be reused across the other invalids for sure

lib/rubocop/cop/rspec/implicit_subject.rb
r7kamura: OK, as I said at [another comment](https://github.com/rubocop/rubocop-rspec/pull/1372#discussion_r967707599) we still need the mapping methods to avoid some Metrics cops, so I decided to rename the methods from style-derived names to ones that describe whay they do.

lib/rubocop/cop/rspec/implicit_subject.rb
```diff
@@ -66,32 +98,67 @@ def on_send(node)
...
+          case style
+          when :require_implicit
+            MSG_REQUIRE_IMPLICIT
+          else
+            MSG_REQUIRE_EXPLICIT
+          end
         end
 
-        def allowed_by_style?(example)
+        def invalid?(node)
```
Darhazer, line 122: Wouldn't it be simpler, if this was tge `alloed_by_style?` implementation. Sure, it now need both the example and the node (for the require_implicit) style; but it also takes care of finding the example, and ignoring it if it's `its`. If the reason to go in this direction is to enforce the implicit usage in `its` - then this is not shown in the tests

spec/rubocop/cop/rspec/implicit_subject_spec.rb
```diff
@@ -183,4 +183,46 @@ def permits(actions)
       RUBY
     end
   end
+
+  context 'with EnforcedStyle `require_implicit`' do
```
Darhazer, line 187: For spec completeness, we need a no-offense example of using a named subject (e.g. `expect(instance).to be_good`. Also a block usage example (e.g. `expect { subject }.to change(goodness, :count)`)

CHANGELOG.md
```diff
@@ -8,6 +8,7 @@
 * Fix a false positive for `RSpec/NoExpectationExample` with pending using `skip` or `pending` inside an example. ([@ydah][])
 * Exclude `have_text` and `have_content` that raise `ArgumentError` with `RSpec/Capybara/VisibilityMatcher` where `:visible` is an invalid option. ([@ydah][])
 * Fix a false negative for `RSpec/Capybara/VisibilityMatcher` with negative matchers. ([@ydah][])
+* Add `require_implicit` style to `RSpec/ImplicitSubject`. ([@r7kamura][])
```
Darhazer, line 11: You have to move the changelog entry under master

Darhazer (approved):
I believe it's good enough to be merged. Further improvements are always possible, and also welcome; but as a feature that can be used, and has enough of a test coverage - let's :rocket:

bquorning (approved):

PR #1373, opened 2022-08-30
Support `Array.new(n)` on `RSpec/FactoryBot/CreateList` cop

r7kamura:
rubocop-performance's `Performance/TimesMap` cop autocorrects `n.times` into `Array.new(n)`, so it would be nice if this cop supports autocorrection from `Array.new(n) { create(:a) }` to `create_list(:a, n)`.

---

* [x] Added tests.
* [ ] Updated documentation.

### DIFF
--- a/lib/rubocop/cop/rspec/factory_bot/create_list.rb
+++ b/lib/rubocop/cop/rspec/factory_bot/create_list.rb
@@ -40,20 +40,23 @@ module RuboCop
           MSG_N_TIMES = 'Prefer %<number>s.times.'
           RESTRICT_ON_SEND = %i[create_list].freeze
 
-          # @!method n_times_block?(node)
-          def_node_matcher :n_times_block?, <<-PATTERN
+          # @!method array_new_or_n_times_block?(node)
+          def_node_matcher :array_new_or_n_times_block?, <<-PATTERN
             (block
-              (send (int _) :times)
+              {
+                (send (const {nil? | cbase} :Array) :new (int _)) |
+                (send (int _) :times)
+              }
               ...
             )
           PATTERN
 
-          # @!method n_times_block_with_arg_and_used?(node)
-          def_node_matcher :n_times_block_with_arg_and_used?, <<-PATTERN
+          # @!method block_with_arg_and_used?(node)
+          def_node_matcher :block_with_arg_and_used?, <<-PATTERN
             (block
-              (send (int _) :times)
+              _
               (args (arg _value))
-                `_value
+              `_value
             )
           PATTERN
 
@@ -75,8 +78,8 @@ module RuboCop
           def on_block(node) # rubocop:todo InternalAffairs/NumblockHandler
             return unless style == :create_list
 
-            return unless n_times_block?(node)
-            return if n_times_block_with_arg_and_used?(node)
+            return unless array_new_or_n_times_block?(node)
+            return if block_with_arg_and_used?(node)
             return unless node.body
             return if arguments_include_method_call?(node.body)
             return unless contains_only_factory?(node.body)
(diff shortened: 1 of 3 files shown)


lib/rubocop/cop/rspec/factory_bot/create_list.rb
```diff
@@ -57,6 +57,31 @@ class CreateList < Base
             )
           PATTERN
 
+          # @!method array_new_block?(node)
```
Darhazer, line 60: I guess we can merge the patterns, as they differ by the receiver

lib/rubocop/cop/rspec/factory_bot/create_list.rb
```diff
@@ -213,6 +244,14 @@ def call_replacement(node)
               replacement
             end
 
+            def count_from(node)
+              if node.receiver.int_type?
+                node.receiver
+              else
+                node.send_node.arguments.first
```
Darhazer, line 251: there's also `first_argument` method for send nodes ;)

lib/rubocop/cop/rspec/factory_bot/create_list.rb
```diff
@@ -40,20 +40,23 @@ class CreateList < Base
           MSG_N_TIMES = 'Prefer %<number>s.times.'
           RESTRICT_ON_SEND = %i[create_list].freeze
 
-          # @!method n_times_block?(node)
-          def_node_matcher :n_times_block?, <<-PATTERN
+          # @!method array_new_or_n_times_block?(node)
+          def_node_matcher :array_new_or_n_times_block?, <<-PATTERN
```
Darhazer, line 44: I'd use some more generic name for this, like `correctable_block?` but I leave this up to you

Darhazer (approved):

lib/rubocop/cop/rspec/factory_bot/create_list.rb
r7kamura: I thought of that kind of name once myself, but I don't think it is necessarily correctable only by this condition.

If we create a method called `#correctable_block?`, the implementation should look something like this:

```ruby
def correctable_block?(node)
  style == :create_list &&
    array_new_or_n_times_block?(node) &&
    node.body &&
    !arguments_include_method_call?(node.body) &&
    contains_only_factory?(node.body)
end
```

ydah:
Perhaps, but it seems that adding an entry to CHANGELOG.md is not included in the pull request.

r7kamura:
I added a missing entry to CHANGELOG.md. Thanks!

lib/rubocop/cop/rspec/factory_bot/create_list.rb
bquorning: I think I learned once that method names containing "or" or "and" should be avoided, and implemented as two methods instead. Would there be significant impact (readability, performance or other) if we chose to keep `#n_times_block?` and added an `#array_new_block?` matcher?

Darhazer: Originally they were split, but since we’re so similar, I suggested having just one. We haven’t come up with a good name though

bquorning: Ah, sorry about that. I really should look at the previous conversations before I comment 😅

lib/rubocop/cop/rspec/factory_bot/create_list.rb
```diff
@@ -213,6 +216,14 @@ def call_replacement(node)
               replacement
             end
 
+            def count_from(node)
+              if node.receiver.int_type?
+                node.receiver
+              else
+                node.send_node.first_argument
+              end.source
```
bquorning, line 224: Can we give a name to the thing we call `source` on? That might allow me to think a bit less about what the method does 😅 
```suggestion
              foo = if node.receiver.int_type?
                node.receiver
              else
                node.send_node.first_argument
              end
              foo.source
```

r7kamura: OK, I'll use the name `count_node` for this variable 👍 

```ruby
  3.times
# ^ this node

  Array.new(3)
          # ^ or this node
```

bquorning (approved):
@Darhazer Feel free to merge if you’re still happy with the changes.

PR #1375, opened 2022-09-05
Catches SubjectStub violations: subject!(:foo) with expect(foo).to receive

eikes:
The following code should be caught:

```ruby
      describe Foo do
        subject!(:foo) { described_class.new }

        it 'uses bang subject! by name' do
          expect(foo).to receive(:bar)
        end
      end
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@
 * Add `AllowedPatterns` configuration option to `RSpec/ContextWording`. ([@ydah][])
 * Add `RSpec/ClassCheck` cop. ([@r7kamura][])
 * Fix a false positive for `RSpec/Capybara/SpecificMatcher` when pseudo-classes. ([@ydah][])
+* Fix a false negative for `RSpec/SubjectStub` when the subject is declared with the `subject!` method and called by name. ([@eikes][])
 
 ## 2.12.1 (2022-07-03)
 
@@ -656,6 +657,7 @@ Compatibility release so users can upgrade RuboCop to 0.51.0. No new features.
 [@dswij]: https://github.com/dswij
 [@dvandersluis]: https://github.com/dvandersluis
 [@edgibbs]: https://github.com/edgibbs
+[@eikes]: https://github.com/eikes
 [@eitoball]: https://github.com/eitoball
 [@elebow]: https://github.com/elebow
 [@EliseFitz15]: https://github.com/EliseFitz15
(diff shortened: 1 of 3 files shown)


ydah:
Please add an entry to the CHANGELOG.md as this is a change that users can observe.

Darhazer:
I guess it also needs to account for `let!` in the subject_overrides. Also it would be great to update it to use the `Language` module, as this is a cop that would not work with any custom subject implementations.

On a side note, @pirj I think majority of cops that capture the argument are not relying on the customizable language, but work only with the build in language (e.g. `subject`, `subject!` in the case). That's something we may think of addressing for 3.0

CHANGELOG.md
```diff
@@ -18,6 +18,7 @@
 * Add `AllowedGroups` configuration option to `RSpec/NestedGroups`. ([@ydah][])
 * Deprecate `IgnoredPatterns` option in favor of the `AllowedPatterns` options. ([@ydah][])
 * Add `AllowedPatterns` configuration option to `RSpec/ContextWording`. ([@ydah][])
+* Fix a false negative for `RSpec/SubjectStub` when the subject is declared with the `subject!` method and called by name. ([@eikes][])
```
pirj, line 21: You have to add yourself to a list of contributors at the end of this file.

pirj (approved):
No objections.

We can certainly improve `subject!` support in the future in other cops.

lib/rubocop/cop/rspec/subject_stub.rb
```diff
@@ -72,7 +72,7 @@ class SubjectStub < Base
         def_node_matcher :subject?, <<-PATTERN
           (block
             (send nil?
-              {:subject (sym $_) | $:subject}
+              {:subject (sym $_) | :subject! (sym $_) | $:subject | $:subject! }
```
ydah, line 75: How about the following?
```suggestion
              { #Subjects.all (sym $_) | $#Subjects.all }
```

ydah: Oops. Something is wrong with the suggestions 🤔.

```ruby
        def_node_matcher :subject?, <<-PATTERN
          (block
            (send nil?
              { #Subjects.all (sym $_) | $#Subjects.all }
            ) args ...)
        PATTERN
```

eikes: Done!

CHANGELOG.md
eikes: Done!

Darhazer (approved):
Thank you.
Can you also rebase this :pray:

PR #1376, opened 2022-09-06
Add support for subject! method to `RSpec/SubjectDeclaration`

ydah:
In this PR, follow up on the following comments:

Refs: https://github.com/rubocop/rubocop-rspec/pull/1375#pullrequestreview-1096722970

---

* [x] Added tests.
* [-] Updated documentation.

### DIFF
--- a/lib/rubocop/cop/rspec/subject_declaration.rb
+++ b/lib/rubocop/cop/rspec/subject_declaration.rb
@@ -25,7 +25,7 @@ module RuboCop
 
         # @!method offensive_subject_declaration?(node)
         def_node_matcher :offensive_subject_declaration?, <<~PATTERN
-          (send nil? ${#Subjects.all #Helpers.all} {(sym :subject) (str "subject")} ...)
+          (send nil? ${#Subjects.all #Helpers.all} {(sym #Subjects.all) (str #Subjects.all)} ...)
         PATTERN
 
         def on_send(node)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,9 @@
 - Update `config/default.yml` removing deprecated option to make the config correctable by users. ([@ignaciovillaverde])
 - Do not attempt to auto-correct example groups with `include_examples` in `RSpec/LetBeforeExamples`. ([@pirj])
 - Add new `RSpec/SortMetadata` cop. ([@leoarnold])
+* Add `require_implicit` style to `RSpec/ImplicitSubject`. ([@r7kamura][])
+* Fix a false positive for `RSpec/Capybara/SpecificMatcher` when `have_css("a")` without attribute. ([@ydah][])
+* Add support for subject! method to `RSpec/SubjectDeclaration`. ([@ydah][])
 
 ## 2.13.2 (2022-09-23)
 
--- a/spec/rubocop/cop/rspec/subject_declaration_spec.rb
+++ b/spec/rubocop/cop/rspec/subject_declaration_spec.rb
@@ -53,6 +53,8 @@ RSpec.describe RuboCop::Cop::RSpec::SubjectDeclaration do
 
   include_examples 'flag unclear subject declaration', :subject
   include_examples 'flag unclear subject declaration', 'subject'
+  include_examples 'flag unclear subject declaration', :subject!
+  include_examples 'flag unclear subject declaration', 'subject!'
 
   context 'when subject helper is used directly' do
     it 'does not register an offense on named subject' do


spec/rubocop/cop/rspec/named_subject_spec.rb
```diff
@@ -63,27 +63,47 @@ def foo
...
-              expect(subject.valid?).to be(true)
-                     ^^^^^^^ Name your test subject if you need to reference it explicitly.
-            end
-
-            it "is new" do
-              expect(subject).to be_new_record
-                     ^^^^^^^ Name your test subject if you need to reference it explicitly.
+            context 'when subject!' do
+              it "is a User" do
+                expect(subject!).to be_a(User)
```
Darhazer, line 94: well, even if you define your subject with `subject!`, you'd still reference it as `subject` in tests.

ydah: I may have misunderstood something, I separated `it`.

ydah: If I have misunderstood, please point it out. I separated the `it` subject and made the `subject` and `subject!` explicit.

Darhazer: I mean that you can't really use `subject!` in an expectation:

> Failure/Error: expect(subject!).to eq('foo')
       `subject!` is not available from within an example (e.g. an `it` block) or from constructs that run in the scope of an example (e.g. `before`, `let`, etc). It is only available on an example group (e.g. a `describe` or `context` block).

You can still declare it as `subject!` or `subject!(:foo)`, but then in the actual test you'll be using `subject` or `foo`. So we don't need to change how we search for subject usage, but the declarations of a subject should use the language module, unless there is specific reason they want to ignore `subject!`

ydah: Ahh! It sure is. I was making a big mistake...
Thank you very much. So it turns out that no fixes is needed for `RSpec/NamedSubject`.

Darhazer: Though if you define `subject(:subject!)` it works

lib/rubocop/cop/rspec/subject_declaration.rb
```diff
@@ -25,7 +25,7 @@ class SubjectDeclaration < Base
 
         # @!method offensive_subject_declaration?(node)
         def_node_matcher :offensive_subject_declaration?, <<~PATTERN
-          (send nil? ${#Subjects.all #Helpers.all} {(sym :subject) (str "subject")} ...)
+          (send nil? ${#Subjects.all #Helpers.all} {(sym #Subjects.all) (str {"subject" | "subject!"})} ...)
```
Darhazer, line 28: I don't think subject!(:subject!)` is technically offensive, as it defines the `subject!` instance method. Otherwise you can't use `subject!` in the spec.
Though this is open for interpretation. @pirj WDYT?

pirj: It is unclear to me why would anyone use `subject!(:subject!)`, but if they would, I would consider it just as offensive as `subject(:subject)` or a nameless subject. Those two would be defined in different scopes, example group and example ones.

bquorning:

lib/rubocop/cop/rspec/subject_declaration.rb
Darhazer: I'm not sure what to do with this.
On one hand, it extends the cop, so if someone says  in the language config `instance` is a subject, it would look for `let(:instance)`. But it won't trigger on `let("instance")`. Perhaps it should be
```suggestion
          (send nil? ${#Subjects.all #Helpers.all} {(sym #Subjects.all) (str #Subjects.all.map(&:to_s))} ...)
```

bquorning (approved):

lib/rubocop/cop/rspec/subject_declaration.rb
```diff
@@ -25,7 +25,7 @@ class SubjectDeclaration < Base
 
         # @!method offensive_subject_declaration?(node)
         def_node_matcher :offensive_subject_declaration?, <<~PATTERN
-          (send nil? ${#Subjects.all #Helpers.all} {(sym :subject) (str "subject")} ...)
+          (send nil? ${#Subjects.all #Helpers.all} {(sym #Subjects.all) (str #Subjects.all.map(&:to_s))} ...)
```
pirj, line 28:
```suggestion
          (send nil? ${#Subjects.all #Helpers.all} {(sym #Subjects.all) (str #Subjects.all)} ...)
```

Surprisingly, it works like this, too, at least specs do pass.

ydah: Sorry for the late reply. Thanks for the correction.

pirj:
Thank you, @ydah !

PR #1378, opened 2022-09-10
Add new `RSpec/SortMetadata` cop

bquorning:
This cop sorts the RSpec metadata alphabetically:

```ruby
  # bad
  describe 'Something', :b, :a
  context 'Something', foo: 'bar', baz: true
  it 'works', :b, :a, foo: 'bar', baz: true

  # good
  describe 'Something', :a, :b
  context 'Something', baz: true, foo: 'bar'
  it 'works', :a, :b, baz: true, foo: 'bar'
```

---

* [x] Added tests.
* [x] Updated documentation.

Closes #1325


lib/rubocop/cop/rspec/sort_metadata.rb
```diff
@@ -0,0 +1,102 @@
...
+              investigate(symbols, pairs.flatten)
+            end
+          end
+
+          rspec_metadata(node) do |symbols, pairs|
+            investigate(symbols, pairs.flatten)
+          end
+        end
+
+        alias on_numblock on_block
```
bquorning, line 56: @Darhazer I have no idea if this is correct. Tbh I think this message could be quite a bit clearer:

> lib/rubocop/cop/rspec/sort_metadata.rb:44:9: C: InternalAffairs/NumblockHandler: Define on_numblock to handle blocks with numbered arguments.

pirj:
Frankly speaking, I found this part of the original PR less useful than the `key: true` <-> `:key` cop.
Last time I ran the cop on `real-world-rspec`, sorting was doubtfully useful at best.

pirj:
> Which original PR do you mean?

[This part](https://github.com/rubocop/rubocop-rspec/compare/52c205d301bfadf6e0cfac812c3ed62dbab5ca80..dfc40a173f0617f05b10d52259acaa6404d2c660#diff-72172d797d98f9a8586056c578546fdefc8b55e8bfb5b6d2431246ddbbd49f91L4179-L4207) of #1325 that was later removed by the author. [Here](https://github.com/rubocop/rubocop-rspec/pull/1325#discussion_r922896282) I suggested extracting that part into a separate cop, but that other cop slipped through the cracks.

pirj:
> I suggested extracting that part into a separate cop

The reason I've suggested extracting to another cop was my opinion that a cop that would transform `focus: true` to `:focus` would be more usable than the one performing a doubtfully useful alphabetical sorting of e.g. `:request, :public, :ab_test`.
In most of the cases what this cop detected in `real-world-rspec`, I would personally add a `rubocop:disable`.

spec/rubocop/cop/rspec/sort_metadata_spec.rb
```diff
@@ -0,0 +1,222 @@
...
+      end
+    RUBY
+
+    expect_correction(<<~RUBY)
+      it 'Something', :a, :b, baz: true, foo: 'bar' do
+      end
+    RUBY
+  end
+
+  it 'registers an offense when using mixed metadata ' \
```
Darhazer, line 84: Aren't we missing an example that we sort them independently, even though  a purely alphabetical order will suggest that the symbols metadata should be last, like:

```
expect_offense(<<~RUBY)
       it 'Something', :c, :z, foo: 'bar', baz: true do
                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ Sort metadata alphabetically.
       end
     RUBY
```

Darhazer: Actually [this ones](https://github.com/rubocop/rubocop-rspec/pull/1378/files#diff-4f710edcf279102cf88580e810d52f474333e8a181c14ed2592e854ae7a54c86R130) covers the case

Darhazer (approved):
Please rebase ;)

Darhazer:
On a side note, I think we need to release a bug-fix release before merging the new cop

config/default.yml
```diff
@@ -766,6 +766,12 @@ RSpec/SingleArgumentMessageChain:
   VersionChanged: '1.10'
   Reference: https://www.rubydoc.info/gems/rubocop-rspec/RuboCop/Cop/RSpec/SingleArgumentMessageChain
 
+RSpec/SortMetadata:
+  Description: Sort RSpec metadata alphabetically.
+  Enabled: pending
+  VersionAdded: 2.13.0
```
Darhazer, line 772: It will be 2.14 now

bquorning: Fixed

docs/modules/ROOT/pages/cops_rspec.adoc
```diff
@@ -4422,6 +4422,39 @@ allow(foo).to receive("bar.baz")
...
 
+== RSpec/SortMetadata
+
+|===
+| Enabled by default | Safe | Supports autocorrection | Version Added | Version Changed
+
+| Pending
+| Yes
+| Yes
+| 2.13.0
```
Darhazer, line 4433: perhaps docs weren't regenerated or they do not handle <<next>> ? Not sure how this is supposed to work with the docs?

pirj: Not regenerated looking at the red "Confirm config and documentation". I think `<<next>>` goes to docs, too, and during the release we manually replace it in bulk with the release version.

bquorning: Eh, I forgot to run `rake generate_cops_documentation`. It works nice with `<<next>>` too. My mistake.

Darhazer (approved):
Merge before there are other conflicts 🙏

PR #1380, opened 2022-09-10
CI: User newer Rubies & reword job names

bquorning:

---

* [ ] Added tests.
* [ ] Updated documentation.


ydah:
I too think the After looks better and easy to see😀

Darhazer (approved):

bquorning:
There we go: As I have renamed some of the required checks, so all open pull requests need to be rebased.

Issue #1390, opened 2022-09-20
[RSpec/Capybara/VisibilityMatcher] False-negative for `have_no_css`


boris-petrov:
```rb
expect(page).to have_no_css('#some-id', visible: false)
```

This should trigger a warning but doesn't.

PR #1391, opened 2022-09-20
Fix a false negative for `RSpec/Capybara/VisibilityMatcher` with negative matcher

ydah:
This PR fixes a problem with `RSpec/Capybara/VisibilityMatcher`
It is difficult to separate the PRs, and we have divided them into two commits in the PR and as a changelog, but please let us know if we need to separate the PRs.

### @ commit 1 https://github.com/rubocop/rubocop-rspec/pull/1391/commits/9778492811425e750a061871d15a87c3a70f4581
If `:visible` is specified for `have_text` and its alias method, `have_content`, it will result in an `ArgumentError` as follows.
Therefore, This PR is excluded `have_text` and `have_content` for `RSpec/Capybara/VisibilityMatcher`.

```command
      Failure/Error: expect(page).to have_text('a', visible: true)

      ArgumentError:
        Invalid option(s) :visible, should be one of :count, :minimum, :maximum, :between, :wait, :exact, :normalize_ws
      # ./spec/capybara/sandbox_spec.rb:20:in `block (2 levels) in <top (required)>'

      Failure/Error: expect(page).to have_content('a', visible: true)

      ArgumentError:
        Invalid option(s) :visible, should be one of :count, :minimum, :maximum, :between, :wait, :exact, :normalize_ws
      # ./spec/capybara/sandbox_spec.rb:68:in `block (2 levels) in <top (required)>'
```

### @ commit 2 https://github.com/rubocop/rubocop-rspec/pull/1391/commits/b8883bc9e64918cb0bdd34c2c61b66a7a5c134b3
Fix: #1390

---

* [-] Squashed related commits together.
* [x] Added tests.
* [-] Updated documentation.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 * Fix an error for `RSpec/Capybara/SpecificFinders` with no parentheses. ([@ydah][])
 * Fix a false positive for `RSpec/NoExpectationExample` with pending using `skip` or `pending` inside an example. ([@ydah][])
 * Exclude `have_text` and `have_content` that raise `ArgumentError` with `RSpec/Capybara/VisibilityMatcher` where `:visible` is an invalid option. ([@ydah][])
+* Fix a false negative for `RSpec/Capybara/VisibilityMatcher` with negative matchers. ([@ydah][])
 
 ## 2.13.1 (2022-09-12)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 * Fix an error for `RSpec/Capybara/SpecificFinders` with no parentheses. ([@ydah][])
 * Fix a false positive for `RSpec/NoExpectationExample` with pending using `skip` or `pending` inside an example. ([@ydah][])
+* Exclude `have_text` and `have_content` that raise `ArgumentError` with `RSpec/Capybara/VisibilityMatcher` where `:visible` is an invalid option. ([@ydah][])
 
 ## 2.13.1 (2022-09-12)
 
--- a/lib/rubocop/cop/rspec/capybara/visibility_matcher.rb
+++ b/lib/rubocop/cop/rspec/capybara/visibility_matcher.rb
@@ -29,18 +29,20 @@ module RuboCop
         class VisibilityMatcher < Base
           MSG_FALSE = 'Use `:all` or `:hidden` instead of `false`.'
           MSG_TRUE = 'Use `:visible` instead of `true`.'
-          CAPYBARA_MATCHER_METHODS = %i[
-            have_selector
-            have_css
-            have_xpath
-            have_link
-            have_button
-            have_field
-            have_select
-            have_table
-            have_checked_field
-            have_unchecked_field
-          ].freeze
+          CAPYBARA_MATCHER_METHODS = %w[
+            button
+            checked_field
+            css
+            field
+            link
+            select
+            selector
+            table
+            unchecked_field
+            xpath
+          ].flat_map do |element|
+            ["have_#{element}".to_sym, "have_no_#{element}".to_sym]
+          end
 
           RESTRICT_ON_SEND = CAPYBARA_MATCHER_METHODS
 
--- a/lib/rubocop/cop/rspec/capybara/visibility_matcher.rb
+++ b/lib/rubocop/cop/rspec/capybara/visibility_matcher.rb
@@ -40,8 +40,6 @@ module RuboCop
             have_table
             have_checked_field
             have_unchecked_field
-            have_text
-            have_content
           ].freeze
 
           RESTRICT_ON_SEND = CAPYBARA_MATCHER_METHODS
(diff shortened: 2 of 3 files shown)


CHANGELOG.md
```diff
@@ -3,6 +3,9 @@
 ## Master (Unreleased)
 
 * Fix an error for `RSpec/Capybara/SpecificFinders` with no parentheses. ([@ydah][])
+* Add support for subject! method to `RSpec/SubjectDeclaration`. ([@ydah][])
```
Darhazer, line 6: This comes from a different branch, doesn't it?

lib/rubocop/cop/rspec/capybara/visibility_matcher.rb
```diff
@@ -30,18 +30,11 @@ class VisibilityMatcher < Base
...
-            have_link
-            have_button
-            have_field
-            have_select
-            have_table
-            have_checked_field
-            have_unchecked_field
-            have_text
-            have_content
+            have_selector have_css have_xpath have_link have_button
```
Darhazer, line 33: might be a good idea to have an alphabetical order of the values

spec/rubocop/cop/rspec/capybara/visibility_matcher_spec.rb
```diff
@@ -35,10 +35,24 @@
                                                 ^^^^^^^^^^^^^^ Use `:all` or `:hidden` instead of `false`.
       expect(page).to have_unchecked_field('cat', visible: false)
                                                   ^^^^^^^^^^^^^^ Use `:all` or `:hidden` instead of `false`.
-      expect(page).to have_text('My homepage', visible: false)
-                                               ^^^^^^^^^^^^^^ Use `:all` or `:hidden` instead of `false`.
-      expect(page).to have_content('Success', visible: false)
+      expect(page).to have_no_css('.profile', visible: false)
```
Darhazer, line 38: I start to wonder if we shouldn't generate the test in a loop over the said selectors

Darhazer (commented):
Having those in one PR is fine 👍 
Please fix the changelog (and take a look at the other comments, though they are not required changes)

CHANGELOG.md
ydah: Oops, I delete it.

lib/rubocop/cop/rspec/capybara/visibility_matcher.rb
ydah: In a simple reordering, `have_no_**` comes in the middle, so I reordered `have_**` and `have_no_**` respectively

spec/rubocop/cop/rspec/capybara/visibility_matcher_spec.rb
ydah: It is indeed a bit long, so I cut out the negative matcher separately. WDYT?

CHANGELOG.md
```diff
@@ -3,6 +3,8 @@
 ## Master (Unreleased)
 
 * Fix an error for `RSpec/Capybara/SpecificFinders` with no parentheses. ([@ydah][])
+* Exclude methods that raise `ArgumentError` with `RSpec/Capybara/VisibilityMatcher` where `:visible` is an invalid option. ([@ydah][])
```
Darhazer, line 6: Perhaps it's better to list which elements were excluded.

lib/rubocop/cop/rspec/capybara/visibility_matcher.rb
```diff
@@ -30,18 +30,11 @@ class VisibilityMatcher < Base
...
-            have_link
-            have_button
-            have_field
-            have_select
-            have_table
-            have_checked_field
-            have_unchecked_field
-            have_text
-            have_content
+            have_button have_checked_field have_css have_field have_link
```
Darhazer, line 33: WDYT of such an approach:
```
%w[button
   checked_field
   css
   field 
   link
   select
   selector
   table
   unchecked_field
   xpath].flat_map { |element| ["have_#{element}", "have_no_#{element}"]}
```
That way we'd never forget to add the no_ version

CHANGELOG.md
```diff
@@ -3,6 +3,8 @@
...
+* Fix a false negative for `RSpec/Capybara/VisibilityMatcher` with negative matcher. ([@ydah][])
```
Darhazer, line 7:
```suggestion
* Fix a false negative for `RSpec/Capybara/VisibilityMatcher` with negative matchers. ([@ydah][])
```

spec/rubocop/cop/rspec/capybara/visibility_matcher_spec.rb
Darhazer: Works for me

CHANGELOG.md
ydah: That looks better. I have modified it to list it.

lib/rubocop/cop/rspec/capybara/visibility_matcher.rb
ydah: I had to make it a symbol, so I modified it a bit, how do you like it?

CHANGELOG.md
Darhazer: I think now you have to entries for the same commit

ydah: Oops, looks like the second commit contained an unnecessary entry. Fixed it.

Darhazer (approved):

Issue #1395, opened 2022-09-24
RSpec/ContextWording: incorrect sample about AllowedPatterns


takaki:
It is written in source code comment 
```yaml
RSpec/ContextWording:
  AllowedPatterns:
    - /とき$/
```
But it does not working.
It should be instead of 
```
    - とき$
```

PR #1396, opened 2022-09-24
Fix confusing AllowedPatterns example in `RSpec/ContextWording`

ydah:
Fix: https://github.com/rubocop/rubocop-rspec/issues/1395

---

* [-] Added tests.
* [x] Updated documentation.


Darhazer:
Shouldn't we also fix it in the specs?
```
{ 'Prefixes' => [], 'AllowedPatterns' => [/とき$/] }
```
I realized that the specs work because Regexp.new(/foo/) returns /foo/. In the specs we provide regexp, and not string. But it would be closer to the reality if we don't do that and let the `AllowedPatterns` parse strings and turns them into regexps

lib/rubocop/cop/rspec/context_wording.rb
```diff
@@ -92,7 +92,9 @@ def bad_pattern?(description)
         end
 
         def expect_patterns
-          inspected = allowed_patterns.map(&:inspect)
+          inspected = allowed_patterns.map do |pattern|
+            pattern.inspect.gsub(/^"|"$/, '/')
```
ydah, line 96: If the conversion process for display here was not changed, it  displayed as follows.
https://github.com/ydah/rubocop-rspec/blob/c017ffd20d2b78f7fda8b129d1aa85ad80edac70/spec/rubocop/cop/rspec/context_wording_spec.rb#L170-L171

```ruby
          context 'this is an incorrect context' do
                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ Context description should match "patterns", or /^when\\b/.
````

ydah: If you see a need to separate commits or add entries to the CHANGELOG.md, please point out.

koic: `^`  and `$` are used, do you expect multiline? If not, `\A` and `\z` should be used.

Darhazer (approved):

lib/rubocop/cop/rspec/context_wording.rb
ydah: Indeed it is, thank you very much. I updated this PR.

bquorning (approved):

PR #1397, opened 2022-09-25
Add Truffleruby head to CI

gogainda:


bquorning:
Hi @gogainda. Can you confirm whether `rubocop` itself is being tested against truffleruby? I don’t think we should expect `rubocop-rspec` to (always) pass the tests if `rubocop` isn’t expected to.

pirj:
> truffleruby-head: spec (pull_request) Successful in **5m**
> jruby: spec (pull_request) Successful in **1m**
> ruby-head: spec (pull_request) Successful in **23s** 

I'm more inclined to close this.
Along with removing the JRuby check, which I can't even recall we have ever had any problem with.

If the truffleruby team has interest in having rubocop-rspec as a regression testing subject, I guess they could add it in their CI.

[Closed RSpec Core PR](https://github.com/rspec/rspec-core/pull/2712).

We're happy, however, to accept changes to support running RuboCop RSpec on TruffleRuby if needed (e.g. [this merged RSpec Core PR](https://github.com/rspec/rspec-core/pull/2946)). But it seems that since the jobs are green, there's nothing needed to be done on this front.

PR #1400, opened 2022-09-30
Mark RSpec/ReturnFromStub autocorrect as unsafe (fixes #1399)

### LINKED ISSUE #1399: RSpec/ReturnFromStub yields false positives and makes unsafe autocorrections
Example:
```ruby
expect(helper).not_to receive(:facet_display_value_sort) { [] }
```
Triggers:
```
C: [Correctable] RSpec/ReturnFromStub: Use and_return for static values.
```
Which when run with `rubocop -a` (safe autocorrect) produces the code:
```ruby
expect(helper).not_to receive(:facet_display_value_sort).and_return([])
```
Which results in the error:
```
Failure/Error: expect(helper).not_to receive(:facet_display_value_sort).and_return([])
RuntimeError:
   `and_return` is not supported with negative message expectations
```

See also: #947

ryanfb:
This marks the autocorrects introduced by RSpec/ReturnFromStub as unsafe. Fixes #1399.

---

* [X] Added tests.
* [X] Updated documentation.

* [X] Set `VersionChanged` in `config/default.yml` to the next major version.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Master (Unreleased)
 
+* Declare autocorrect as unsafe for `RSpec/ReturnFromStub`. ([@ryanfb][])
 * Add `require_implicit` style to `RSpec/ImplicitSubject`. ([@r7kamura][])
 
 ## 2.13.2 (2022-09-23)
@@ -731,6 +732,7 @@ Compatibility release so users can upgrade RuboCop to 0.51.0. No new features.
 [@rrosenblum]: https://github.com/rrosenblum
 [@rspeicher]: https://github.com/rspeicher
 [@RST-J]: https://github.com/RST-J
+[@ryanfb]: https://github.com/ryanfb
 [@schmijos]: https://github.com/schmijos
 [@seanpdoyle]: https://github.com/seanpdoyle
 [@sl4vr]: https://github.com/sl4vr
--- a/config/default.yml
+++ b/config/default.yml
@@ -725,11 +725,12 @@ RSpec/ReturnFromStub:
   Description: Checks for consistent style of stub's return setting.
   Enabled: true
   EnforcedStyle: and_return
+  SafeAutoCorrect: false
   SupportedStyles:
     - and_return
     - block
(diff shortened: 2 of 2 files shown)


ydah:
It seems to me that a better approach to the issue mentioned at https://github.com/rubocop/rubocop-rspec/issues/1399 would be to not make it an offense in the case of negative message expectations. Please point out if we have missed any cases where something cannot be easily changed.

pirj:
https://docs.rubocop.org/rubocop/usage/autocorrect.html#safe-autocorrect:

> The definition of safety is that the cop doesn’t generate false positives. On top of that there’s SafeAutoCorrect that might be set to false in cases where only the autocorrect performed by a cop is unsafe, but that the offense detection logic is safe. To sum it up:

>    Safe (true/false) - indicates whether the cop can yield false positives (by design) or not.

>    SafeAutoCorrect (true/false) - indicates whether the autocorrect a cop does is safe (equivalent) by design. If a cop is unsafe its autocorrect automatically becomes unsafe as well.

There's a bug in the cop, and it yields false positives.
If you're having troubles with the cop, temporarily disable it in your `.rubocop_todo.yml` until its detection is fixed.

ryanfb:
Let me know if you'd like me to re-PR this to mark `Safe: false` instead of `SafeAutoCorrect: false`, but I don't have the time or familiarity with rubocop internals to fix the underlying bug.

ydah:
They are reporting it as an Issue and I will attempt to fix it tomorrow. Of course, if others are willing to try, I'm sure it will be resolved more quickly👍

pirj:
I see no real issue, as the [block implementation](https://relishapp.com/rspec/rspec-mocks/v/3-9/docs/configuring-responses/block-implementation) is at least useless with negation just as `and_return` is.

> will attempt to fix it tomorrow

Please don't until we figure out why such style is used, and if it is a mistake, or it has some real intent.

ydah:
I shouldn't have used the word "fix". Sorry.
I wanted to express my willingness to check the Issue to see the real problem and resolve it.

PR #1401, opened 2022-10-02
Fix a false positive for `RSpec/Capybara/SpecificMatcher` when `have_css("a")` without attribute

ydah:
This PR is fix a false positive for `RSpec/Capybara/SpecificMatcher` when may not have a `href`.

We fixed a similar correction at https://github.com/rubocop/rubocop-rspec/pull/1345, but the following cases remained false positive.

```
expect(page).to have_css("a")
```

---

* [x] Added tests.
* [-] Updated documentation.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## Master (Unreleased)
 
 * Add `require_implicit` style to `RSpec/ImplicitSubject`. ([@r7kamura][])
+* Fix a false positive for `RSpec/Capybara/SpecificMatcher` when `have_css("a")` without attribute. ([@ydah][])
 
 ## 2.13.2 (2022-09-23)
 
--- a/lib/rubocop/cop/rspec/capybara/specific_matcher.rb
+++ b/lib/rubocop/cop/rspec/capybara/specific_matcher.rb
@@ -96,7 +96,6 @@ module RuboCop
 
           def specific_matcher_option?(node, arg, matcher)
             attrs = CssSelector.attributes(arg).keys
-            return true if attrs.empty?
             return false unless replaceable_matcher?(node, matcher, attrs)
 
             attrs.all? do |attr|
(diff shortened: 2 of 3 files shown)


lib/rubocop/cop/rspec/capybara/specific_matcher.rb
```diff
@@ -96,8 +96,8 @@ def specific_matcher(arg)
 
           def specific_matcher_option?(node, arg, matcher)
             attrs = CssSelector.attributes(arg).keys
-            return true if attrs.empty?
             return false unless replaceable_matcher?(node, matcher, attrs)
+            return true if attrs.empty?
```
Darhazer, line 100: since attrs.all? will return true on an empty list, I guess we can just remove this line

CHANGELOG.md
```diff
@@ -10,6 +10,7 @@
 * Fix a false positive for `RSpec/NoExpectationExample` with pending using `skip` or `pending` inside an example. ([@ydah][])
 * Exclude `have_text` and `have_content` that raise `ArgumentError` with `RSpec/Capybara/VisibilityMatcher` where `:visible` is an invalid option. ([@ydah][])
 * Fix a false negative for `RSpec/Capybara/VisibilityMatcher` with negative matchers. ([@ydah][])
+* Fix a false positive for `RSpec/Capybara/SpecificMatcher` when `have_css("a")` without attribute. ([@ydah][])
```
Darhazer, line 13: this should be under master

ydah: oops, moved to master

lib/rubocop/cop/rspec/capybara/specific_matcher.rb
ydah: indeed it is, thank you so much. I removed it.

Darhazer (approved):

foucist:
Looks like this fixes the case where I use have_css("input.border-red-600") to check the border was set to a color

PR #1402, opened 2022-10-02
Add new `RSpec/Capybara/SpecificActions` cop

ydah:
Add new `RSpec/Capybara/SpecificActions` cop

Fix: https://github.com/rubocop/rubocop-rspec/issues/1326

We would like to eventually add autocorrect, but as with `RSpec/Capybara/SpecificMatcher`, the checks are complex and it would be better to introduce them step by step.
What do you think? I would like to hear your opinion.

---

* [x] Added tests.
* [x] Updated documentation.

### DIFF
--- a/lib/rubocop/cop/rspec/capybara/specific_actions.rb
+++ b/lib/rubocop/cop/rspec/capybara/specific_actions.rb
@@ -1,0 +1,130 @@
+# frozen_string_literal: true
+
+module RuboCop
+  module Cop
+    module RSpec
+      module Capybara
+        # Checks for there is a more specific actions offered by Capybara.
+        #
+        # @example
+        #
+        #   # bad
+        #   find('a').click
+        #   find('button.cls').click
+        #   find('a', exact_text: 'foo').click
+        #   find('div button').click
+        #
+        #   # good
+        #   click_link
+        #   click_button(class: 'cls')
+        #   click_link(exact_text: 'foo')
+        #   find('div').click_button
+        #
+        class SpecificActions < Base
+          MSG = "Prefer `%<good_action>s` over `find('%<selector>s').click`."
+          RESTRICT_ON_SEND = %i[click].freeze
+          SPECIFIC_ACTION = {
+            'button' => 'button',
+            'a' => 'link'
+          }.freeze
+
+          # @!method click_on_selector(node)
+          def_node_matcher :click_on_selector, <<-PATTERN
+            (send _ :find (str $_) ...)
+          PATTERN
+
+          # @!method option?(node)
+          def_node_search :option?, <<-PATTERN
+            (pair (sym %) _)
+          PATTERN
+
+          def on_send(node)
+            click_on_selector(node.receiver) do |arg|
+              next unless supported_selector?(arg)
+              next unless (selector = last_selector(arg))
+              next unless (action = specific_action(selector))
+              next unless specific_action_option?(node.receiver, arg, action)
+              next unless specific_action_pseudo_classes?(arg)
+
+              range = offense_range(node, node.receiver)
+              add_offense(range, message: message(action, selector))
+            end
+          end
+
+          private
+
+          def specific_action(selector)
+            SPECIFIC_ACTION[last_selector(selector)]
+          end
+
+          def specific_action_option?(node, arg, action)
+            attrs = CssSelector.attributes(arg).keys
+            return false unless replaceable_action?(node, action, attrs)
+
+            attrs.all? do |attr|
+              CssSelector.specific_options?(action, attr)
+            end
+          end
+
+          def specific_action_pseudo_classes?(arg)
+            CssSelector.pseudo_classes(arg).all? do |pseudo_class|
+              replaceable_pseudo_class?(pseudo_class, arg)
+            end
+          end
+
+          def replaceable_pseudo_class?(pseudo_class, arg)
+            unless CssSelector.specific_pesudo_classes?(pseudo_class)
+              return false
+            end
+
+            case pseudo_class
+            when 'not()' then replaceable_pseudo_class_not?(arg)
+            else true
+            end
+          end
+
+          def replaceable_pseudo_class_not?(arg)
+            arg.scan(/not\(.*?\)/).all? do |not_arg|
+              CssSelector.attributes(not_arg).values.all? do |v|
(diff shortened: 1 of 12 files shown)


lib/rubocop/cop/rspec/capybara/specific_actions.rb
```diff
@@ -0,0 +1,136 @@
...
+
+              range = offense_range(node, receiver)
+              add_offense(range, message: message(action, selector))
+            end
+          end
+
+          def receiver(node)
+            return unless (receiver = node.receiver)
+
+            receiver if receiver.method?(:find)
```
Darhazer, line 59: the method name is later checked in the first_argument matcher as well. I'd suggest using just the matcher as following:

lib/rubocop/cop/rspec/capybara/specific_actions.rb
```diff
@@ -0,0 +1,136 @@
...
+          RESTRICT_ON_SEND = %i[click].freeze
+          SPECIFIC_ACTION = {
+            'button' => 'button',
+            'a' => 'link'
+          }.freeze
+
+          # @!method first_argument(node)
+          def_node_matcher :first_argument, <<-PATTERN
+            (send _ :find (str $_) ...)
+          PATTERN
```
Darhazer, line 34:
```suggestion
          # @!method first_argument(node)
          def_node_matcher :click_on_selector, <<-PATTERN
            $(send _ :find (str $_) ...)
          PATTERN
```

lib/rubocop/cop/rspec/capybara/specific_actions.rb
```diff
@@ -0,0 +1,136 @@
...
+
+          # @!method option?(node)
+          def_node_search :option?, <<-PATTERN
+            (pair (sym %) _)
+          PATTERN
+
+          def on_send(node)
+            return unless (receiver = receiver(node))
+
+            first_argument(receiver) do |arg|
```
Darhazer, line 44:
```suggestion
            click_on_selector(node.receiver) do |receiver, arg|
```

lib/rubocop/cop/rspec/capybara/specific_actions.rb
```diff
@@ -0,0 +1,136 @@
...
+              return unless supported_selector?(arg)
+              return unless (selector = last_selector(arg)[/^\w+/, 0])
+              return unless (action = specific_action(selector))
+              next unless specific_action_option?(receiver, arg, action)
```
Darhazer, line 48: the mix between return and next is confusing

lib/rubocop/cop/rspec/capybara/specific_actions.rb
```diff
@@ -0,0 +1,136 @@
...
+
+          def replaceable_pseudo_class_not?(arg)
+            arg.scan(/not\(.*?\)/).all? do |not_arg|
+              CssSelector.attributes(not_arg).values.all? do |v|
+                v.is_a?(TrueClass) || v.is_a?(FalseClass)
+              end
+            end
+          end
+
+          def replaceable_action?(node, action, attrs)
```
Darhazer, line 100: perhaps we need to start extracting common parts of the cops

ydah: Indeed it is. I extracting common parts of the cops. These seemed like different methods from CssSelector, so we carved them out into a separate module as CapybaraHelp.

ydah: We have it as a separate commit, but if it needs to be squashed, please point it out.

lib/rubocop/cop/rspec/capybara/specific_actions.rb
```diff
@@ -0,0 +1,80 @@
...
+
+          # @!method click_on_selector(node)
+          def_node_matcher :click_on_selector, <<-PATTERN
+            (send _ :find (str $_) ...)
+          PATTERN
+
+          def on_send(node)
+            click_on_selector(node.receiver) do |arg|
+              next unless supported_selector?(arg)
+              next unless (selector = last_selector(arg)[/^\w+/, 0])
```
Darhazer, line 41: what is this part doing?
I find it confusing that specific_action then again want the last_selector. Can we extract the action in one go?

lib/rubocop/cop/rspec/capybara/specific_actions.rb
```diff
@@ -0,0 +1,80 @@
...
+        #   find('div button').click
+        #
+        #   # good
+        #   click_link
+        #   click_button(class: 'cls')
+        #   click_link(exact_text: 'foo')
+        #   find('div').click_button
+        #
+        class SpecificActions < Base
+          include CapybaraHelp
```
bquorning, line 24: I was wondering if we could (like with the CssSelector module) not include it, but instead just call methods on the module (using `module_function`)? My hope is that it would be easier to follow the code that way. What do you think?

lib/rubocop/cop/rspec/capybara/specific_actions.rb
ydah: In the case of multiple selectors given to `find`, click_button and click_link are substituted if the last selector element divided by a space character is `button` or `a`. The replacement is possible.
example: `find('div button')`
So here we get the last selector given as an argument to `find`.

lib/rubocop/cop/rspec/capybara/specific_actions.rb
```diff
@@ -0,0 +1,80 @@
...
+          def specific_action(selector)
+            SPECIFIC_ACTION[last_selector(selector)]
+          end
+
+          def supported_selector?(selector)
+            !selector.match?(/[>,+~]/)
+          end
+
+          def last_selector(arg)
+            arg.split(' ')[-1]
```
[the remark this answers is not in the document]
ydah: Indeed it is. We will consider adding it after this.

lib/rubocop/cop/rspec/capybara/specific_actions.rb
ydah: In the `CapybaraHelp` class, I had defined `def_node_matcher`, so I did not make it a `module_function`. Certainly from a code readability standpoint, `module_function` is better, so we changed it.

bquorning (approved):

Darhazer (approved):

PR #1405, opened 2022-10-04
Add new `RSpec/Capybara/NegationMatcher` cop

ydah:
Resolve: https://github.com/rubocop/rubocop-rspec/issues/378

---

* [x] Added tests.
* [x] Updated documentation.

### DIFF
--- a/config/default.yml
+++ b/config/default.yml
@@ -869,6 +869,16 @@ RSpec/Capybara/FeatureMethods:
   VersionChanged: '2.0'
   Reference: https://www.rubydoc.info/gems/rubocop-rspec/RuboCop/Cop/RSpec/Capybara/FeatureMethods
 
+RSpec/Capybara/NegationMatcher:
+  Description: Enforces use of `have_no_*` or `not_to` for negated expectations.
+  Enabled: pending
+  VersionAdded: '2.14'
+  EnforcedStyle: not_to
+  SupportedStyles:
+    - have_no
+    - not_to
+  Reference: https://www.rubydoc.info/gems/rubocop-rspec/RuboCop/Cop/RSpec/Capybara/NegationMatcher
+
 RSpec/Capybara/SpecificFinders:
   Description: Checks if there is a more specific finder offered by Capybara.
   Enabled: pending
--- a/lib/rubocop/cop/rspec/capybara/negation_matcher.rb
+++ b/lib/rubocop/cop/rspec/capybara/negation_matcher.rb
@@ -1,0 +1,106 @@
+# frozen_string_literal: true
+
+module RuboCop
+  module Cop
+    module RSpec
+      module Capybara
+        # Enforces use of `have_no_*` or `not_to` for negated expectations.
+        #
+        # @example EnforcedStyle: not_to (default)
+        #   # bad
+        #   expect(page).to have_no_selector
+        #   expect(page).to have_no_css('a')
+        #
+        #   # good
+        #   expect(page).not_to have_selector
+        #   expect(page).not_to have_css('a')
+        #
+        # @example EnforcedStyle: have_no
+        #   # bad
+        #   expect(page).not_to have_selector
+        #   expect(page).not_to have_css('a')
+        #
+        #   # good
+        #   expect(page).to have_no_selector
+        #   expect(page).to have_no_css('a')
+        #
+        class NegationMatcher < Base
+          extend AutoCorrector
+          include ConfigurableEnforcedStyle
+
+          MSG = 'Use `expect(...).%<runner>s %<matcher>s`.'
+          CAPYBARA_MATCHERS = %w[
+            selector css xpath text title current_path link button
+            field checked_field unchecked_field select table
+            sibling ancestor
+          ].freeze
+          POSITIVE_MATCHERS =
+            Set.new(CAPYBARA_MATCHERS) { |element| :"have_#{element}" }.freeze
+          NEGATIVE_MATCHERS =
+            Set.new(CAPYBARA_MATCHERS) { |element| :"have_no_#{element}" }
+              .freeze
+          RESTRICT_ON_SEND = (POSITIVE_MATCHERS + NEGATIVE_MATCHERS).freeze
+
+          # @!method not_to?(node)
+          def_node_matcher :not_to?, <<~PATTERN
+            (send ... :not_to
+              (send nil? %POSITIVE_MATCHERS ...))
+          PATTERN
+
+          # @!method have_no?(node)
+          def_node_matcher :have_no?, <<~PATTERN
+            (send ... :to
+              (send nil? %NEGATIVE_MATCHERS ...))
+          PATTERN
+
+          def on_send(node)
+            return unless offense?(node.parent)
+
+            matcher = node.method_name.to_s
+            add_offense(offense_range(node),
+                        message: message(matcher)) do |corrector|
+              corrector.replace(node.parent.loc.selector, replaced_runner)
+              corrector.replace(node.loc.selector,
+                                replaced_matcher(matcher))
+            end
+          end
+
+          private
+
+          def offense?(node)
+            (style == :have_no && not_to?(node)) ||
+              (style == :not_to && have_no?(node))
+          end
+
+          def offense_range(node)
+            node.parent.loc.selector.with(end_pos: node.loc.selector.end_pos)
+          end
+
+          def message(matcher)
+            format(MSG,
+                   runner: replaced_runner,
+                   matcher: replaced_matcher(matcher))
+          end
+
+          def replaced_runner
+            case style
+            when :have_no
+              'to'
+            when :not_to
+              'not_to'
+            end
+          end
+
+          def replaced_matcher(matcher)
+            case style
+            when :have_no
+              matcher.sub('have_', 'have_no_')
+            when :not_to
+              matcher.sub('have_no_', 'have_')
+            end
+          end
+        end
+      end
+    end
+  end
+end
--- a/spec/rubocop/cop/rspec/capybara/negation_matcher_spec.rb
+++ b/spec/rubocop/cop/rspec/capybara/negation_matcher_spec.rb
@@ -1,0 +1,97 @@
+# frozen_string_literal: true
+
+RSpec.describe RuboCop::Cop::RSpec::Capybara::NegationMatcher, :config do
+  let(:cop_config) { { 'EnforcedStyle' => enforced_style } }
+
+  context 'with EnforcedStyle `have_no`' do
+    let(:enforced_style) { 'have_no' }
+
+    %i[selector css xpath text title current_path link button
+       field checked_field unchecked_field select table
+       sibling ancestor].each do |matcher|
+      it 'registers an offense when using ' \
+         '`expect(...).not_to have_#{matcher}`' do
+        expect_offense(<<~RUBY, matcher: matcher)
+          expect(page).not_to have_#{matcher}
+                       ^^^^^^^^^^^^^{matcher} Use `expect(...).to have_no_#{matcher}`.
+          expect(page).not_to have_#{matcher}('a')
+                       ^^^^^^^^^^^^^{matcher} Use `expect(...).to have_no_#{matcher}`.
+        RUBY
+
+        expect_correction(<<~RUBY)
+          expect(page).to have_no_#{matcher}
+          expect(page).to have_no_#{matcher}('a')
+        RUBY
+      end
+
+      it 'does not register an offense when using ' \
+         '`expect(...).to have_no_#{matcher}`' do
+        expect_no_offenses(<<~RUBY)
+          expect(page).to have_no_#{matcher}
+        RUBY
+      end
+    end
+
+    it 'registers an offense when using ' \
+       '`expect(...).not_to have_text` with heredoc' do
+      expect_offense(<<~RUBY)
+        expect(page).not_to have_text(exact_text: <<~TEXT)
+                     ^^^^^^^^^^^^^^^^ Use `expect(...).to have_no_text`.
+          foo
+        TEXT
+      RUBY
+
+      expect_correction(<<~RUBY)
+        expect(page).to have_no_text(exact_text: <<~TEXT)
+          foo
+        TEXT
+      RUBY
+    end
+  end
+
+  context 'with EnforcedStyle `not_to`' do
+    let(:enforced_style) { 'not_to' }
+
+    %i[selector css xpath text title current_path link button
+       field checked_field unchecked_field select table
+       sibling ancestor].each do |matcher|
+      it 'registers an offense when using ' \
+         '`expect(...).to have_no_#{matcher}`' do
+        expect_offense(<<~RUBY, matcher: matcher)
+          expect(page).to have_no_#{matcher}
+                       ^^^^^^^^^^^^{matcher} Use `expect(...).not_to have_#{matcher}`.
+          expect(page).to have_no_#{matcher}('a')
+                       ^^^^^^^^^^^^{matcher} Use `expect(...).not_to have_#{matcher}`.
+        RUBY
+
+        expect_correction(<<~RUBY)
+          expect(page).not_to have_#{matcher}
+          expect(page).not_to have_#{matcher}('a')
+        RUBY
+      end
+
+      it 'does not register an offense when using ' \
+         '`expect(...).not_to have_#{matcher}`' do
+        expect_no_offenses(<<~RUBY)
+          expect(page).not_to have_#{matcher}
+        RUBY
+      end
+
+      it 'registers an offense when using ' \
+         '`expect(...).to have_no_text` with heredoc' do
+        expect_offense(<<~RUBY)
+          expect(page).to have_no_text(exact_text: <<~TEXT)
+                       ^^^^^^^^^^^^^^^ Use `expect(...).not_to have_text`.
+            foo
+          TEXT
+        RUBY
+
+        expect_correction(<<~RUBY)
+          expect(page).not_to have_text(exact_text: <<~TEXT)
+            foo
+          TEXT
+        RUBY
+      end
+    end
+  end
+end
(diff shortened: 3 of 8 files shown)


lib/rubocop/cop/rspec/capybara/negation_matcher.rb
```diff
@@ -0,0 +1,104 @@
...
+
+          MSG = 'Use `%<good>s` instead of `%<bad>s`.'
+          CAPYBARA_MATCHERS = %w[
+            selector css xpath text title current_path link button
+            field checked_field unchecked_field select table
+            sibling ancestor
+          ].freeze
+          POSITIVE_MATCHERS = CAPYBARA_MATCHERS.flat_map do |element|
+            ["have_#{element}".to_sym]
+          end.freeze
```
Darhazer, line 39: Why using flat_map with single elements?
```suggestion
          POSITIVE_MATCHERS = CAPYBARA_MATCHERS.map do |element|
            "have_#{element}".to_sym
          end.freeze
```

lib/rubocop/cop/rspec/capybara/negation_matcher.rb
```diff
@@ -0,0 +1,104 @@
...
+          NEGATIVE_MATCHERS = CAPYBARA_MATCHERS.flat_map do |element|
+            ["have_no_#{element}".to_sym]
+          end.freeze
```
Darhazer, line 42:
```suggestion
          NEGATIVE_MATCHERS = CAPYBARA_MATCHERS.map do |element|
            "have_no_#{element}".to_sym
          end.freeze
```

Darhazer (commented):
For spec completeness, can you add:
Case with heredoc for have_text (usually those are trickier for auto-correction, though we have some nice helpers to deal with those)

Test with argument, e.g. expect(page).not_to have_css('a') being corrected to expect(page).to have_no_css('a')

lib/rubocop/cop/rspec/capybara/negation_matcher.rb
ydah: Oops, you're right. I fixed it.

lib/rubocop/cop/rspec/capybara/negation_matcher.rb
```diff
@@ -0,0 +1,104 @@
...
+          NEGATIVE_MATCHERS = CAPYBARA_MATCHERS.map do |element|
+            "have_no_#{element}".to_sym
+          end.freeze
+          RESTRICT_ON_SEND = (POSITIVE_MATCHERS + NEGATIVE_MATCHERS).freeze
+
+          # @!method not_to?(node)
+          def_node_matcher :not_to?, <<~PATTERN
+            (send ... :not_to
+              (send nil? {#{POSITIVE_MATCHERS.map(&:inspect).join(' ')}} ...))
```
pirj, line 48: Can `{#{POSITIVE_MATCHERS.map(&:inspect).join(' ')}}` be simplified just to `%POSITIVE_MATCHERS` by [making them a Set](https://docs.rubocop.org/rubocop-ast/node_pattern.html#const-or-const-for-constants)?

spec/rubocop/cop/rspec/capybara/negation_matcher_spec.rb
```diff
@@ -0,0 +1,97 @@
...
+    let(:enforced_style) { 'have_no' }
+
+    %i[selector css xpath text title current_path link button
+       field checked_field unchecked_field select table
+       sibling ancestor].each do |matcher|
+      it 'registers an offense when using ' \
+        '`expect(...).not_to have_#{matcher}`' do
+        expect_offense(<<~RUBY, matcher: matcher)
+          expect(page).not_to have_#{matcher}
+                       ^^^^^^^^^^^^^{matcher} Use `expect(...).not_to have_#{matcher}` instead of `expect(...).to have_no_#{matcher}`.
```
pirj, line 16: It seems there's duplication here. `^^^^^^` already highlight what we consider bad.

lib/rubocop/cop/rspec/capybara/negation_matcher.rb
```diff
@@ -0,0 +1,104 @@
...
+                   good: format_message(node.parent.loc.selector.source,
+                                        node.loc.selector.source),
+                   bad: format_message(replaced_target, replaced_matcher(node)))
+          end
+
+          def format_message(target, matcher)
+            "expect(...).#{target} #{matcher}"
+          end
+
+          def replaced_target
```
pirj, line 89: We refer to this a "runner" in `RuboCop::RSpec::Language`. I'm not aware if there's an official RSpec term for `to`/`not_to`.

lib/rubocop/cop/rspec/capybara/negation_matcher.rb
```diff
@@ -0,0 +1,104 @@
...
+            style == :have_no ? 'to' : 'not_to'
+          end
+
+          def replaced_matcher(node)
+            if style == :have_no
+              node.loc.selector.source.gsub('have_', 'have_no_')
+            else
+              node.loc.selector.source.gsub('_no', '')
```
pirj, line 97: Maybe `.gsub('have_no_', 'have_')` to avoid some edge cases?

pirj (approved):
Looks good.
Thank you for another useful contribution, @ydah!
Small suggestions below.

Would the cop tolerate compound expectations like:
```ruby
expect(page)
  .to have_no_css('foo')
  .and have_no_css('bar')
```

And does it even make sense to insist on using `not_to` for compound expectations? (Similar to I believe #1349).

pirj:
I'm running the cop against `real-world-rspec`, and getting this error message:
```
$ rubocop --only RSpec/Capybara/NegationMatcher ../real-world-rspec
Resolving dependencies...
undefined method `merge' for [{"cops"=>["Layout/SpaceAroundOperators", "Style/SpaceAroundOperators"], "parameters"=>"MultiSpaceAllowedForOperators", "reason"=>"If your intention was to allow extra spaces for alignment, please use `AllowForAlignment: true` instead.\n"}, {"cops"=>"Style/Encoding", "parameters"=>["EnforcedStyle", "SupportedStyles", "AutoCorrectEncodingComment"], "reason"=>"`Style/Encoding` no longer supports styles. The \"never\" behavior is always assumed."}, {"cops"=>"Style/IfUnlessModifier", "parameters"=>"MaxLineLength", "reason"=>"`Style/IfUnlessModifier: MaxLineLength` has been removed. Use `Layout/LineLength: Max` instead\n"}, {"cops"=>"Style/WhileUntilModifier", "parameters"=>"MaxLineLength", "reason"=>"`Style/WhileUntilModifier: MaxLineLength` has been removed. Use `Layout/LineLength: Max` instead\n"}, {"cops"=>"AllCops", "parameters"=>"RunRailsCops", "reason"=>"Use the following configuration instead:\nRails:\n  Enabled: true"}, {"cops"=>"Layout/CaseIndentation", "parameters"=>"IndentWhenRelativeTo", "alternative"=>"EnforcedStyle"}, {"cops"=>["Lint/BlockAlignment", "Layout/BlockAlignment", "Lint/EndAlignment", "Layout/EndAlignment", "Lint/DefEndAlignment", "Layout/DefEndAlignment"], "parameters"=>"AlignWith", "alternative"=>"EnforcedStyleAlignWith"}, {"cops"=>"Rails/UniqBeforePluck", "parameters"=>"EnforcedMode", "alternative"=>"EnforcedStyle"}, {"cops"=>"Style/MethodCallWithArgsParentheses", "parameters"=>"IgnoredMethodPatterns", "alternative"=>"AllowedPatterns"}, {"cops"=>["Performance/Count", "Performance/Detect"], "parameters"=>"SafeMode", "reason"=>"`SafeMode` has been removed. Use `SafeAutoCorrect` instead."}, {"cops"=>"Bundler/GemComment", "parameters"=>"Whitelist", "alternative"=>"IgnoredGems"}, {"cops"=>["Lint/SafeNavigationChain", "Lint/SafeNavigationConsistency", "Style/NestedParenthesizedCalls", "Style/SafeNavigation", "Style/TrivialAccessors"], "parameters"=>"Whitelist", "alternative"=>"AllowedMethods"}, {"cops"=>"Style/IpAddresses", "parameters"=>"Whitelist", "alternative"=>"AllowedAddresses"}, {"cops"=>"Naming/HeredocDelimiterNaming", "parameters"=>"Blacklist", "alternative"=>"ForbiddenDelimiters"}, {"cops"=>"Naming/PredicateName", "parameters"=>"NamePrefixBlacklist", "alternative"=>"ForbiddenPrefixes"}, {"cops"=>"Naming/PredicateName", "parameters"=>"NameWhitelist", "alternative"=>"AllowedMethods"}, {"cops"=>["Metrics/BlockLength", "Metrics/MethodLength"], "parameters"=>"ExcludedMethods", "alternative"=>"IgnoredMethods", "severity"=>"warning"}, {"cops"=>"Lint/Debugger", "parameters"=>"DebuggerReceivers", "reason"=>"`DebuggerReceivers` is no longer necessary, method receivers should be specified in `DebuggerMethods` instead.", "severity"=>"warning"}, {"cops"=>["Layout/IndentationWidth", "Layout/LineLength", "Lint/UnreachableLoop", "Naming/MethodName", "Style/MethodCallWithArgsParentheses"], "parameters"=>"IgnoredPatterns", "alternative"=>"AllowedPatterns", "severity"=>"warning"}]:Array
/Users/pirj/.rvm/gems/ruby-3.0.3@rubocop-rspec/gems/rubocop-1.31.1/lib/rubocop/config_obsoletion.rb:53:in `block (2 levels) in load_rules'
```

is it just me? 🤔

ydah:
> I'm running the cop against real-world-rspec, and getting this error message:

It seems to me that the RuboCop version must be 1.33 or higher.

config/default.yml
```diff
@@ -867,6 +867,16 @@ RSpec/Capybara/FeatureMethods:
   VersionChanged: '2.0'
   Reference: https://www.rubydoc.info/gems/rubocop-rspec/RuboCop/Cop/RSpec/Capybara/FeatureMethods
 
+RSpec/Capybara/NegationMatcher:
+  Description: Enforces use of `have_no_*` or `not_to` for negated expectations.
+  Enabled: pending
+  VersionAdded: '2.14'
+  EnforcedStyle: have_no
```
pirj, line 874: With enforced style `have_no`:
```
...
    expect(page).not_to have_selector("div[data-hook='taxon_children']")
                 ^^^^^^^^^^^^^^^^^^^^
57497 files inspected, 1569 offenses detected, 1569 offenses autocorrectable
```
with `not_to`:
```
...
      expect(find('.payment-info')).to have_no_css('img')
                                    ^^^^^^^^^^^^^^
57497 files inspected, 205 offenses detected, 205 offenses autocorrectable
```

Despite my subjective personal preference, I'd suggest setting the default enforced style to what is causing fewer offences, unless we're enforcing some recommendation from Capybara to prefer `have_no_*` matchers.

I could find this:

> Capybara's RSpec matchers, however, are smart enough to handle either form. The two following statements are functionally equivalent:
```
expect(page).not_to have_xpath('a')
expect(page).to have_no_xpath('a')
```

lib/rubocop/cop/rspec/capybara/negation_matcher.rb
ydah: I see. I named this place `target` because `to`/`not_to` belongs to `ExpectationTarget`.
If we are going to name it `runner` I would like to change it, but I would like to hear your opinion.

https://www.rubydoc.info/gems/rspec-expectations/2.14.4/RSpec/Expectations/ExpectationTarget

pirj:
```ruby
expect(something) # => ExpectationTarget wrapping something
```

I agree that the receiver of `to`/`not_to` is an expectation target. However, what is the `to`/`not_to` itself? From [its description](https://www.rubydoc.info/gems/rspec-expectations/2.14.4/RSpec%2FExpectations%2FExpectationTarget:to):
> #to(matcher = nil, message = nil, &block) ⇒ Boolean
> Runs the given expectation, passing if matcher returns true.

i'd say it's a runner if it runs something. However, it's not referenced as such anywhere.

Furthermore, it might add confusion with an unrelated `RSpec::Core::Runner`.

My only point is the internal consistency, since we call it runner in `RuboCop::RSpec::Language`, possibly by mistake.

Leaving this up to you to decide, I'm fine either way.

lib/rubocop/cop/rspec/capybara/negation_matcher.rb
ydah: Thank you so much. I fixed to use `Set`.

lib/rubocop/cop/rspec/capybara/negation_matcher.rb
ydah: Indeed it is. I updated it.

config/default.yml
ydah: Indeed it is. We made sure to default to `not_to`.

spec/rubocop/cop/rspec/capybara/negation_matcher_spec.rb
ydah: Here, it was `Use bad instead of good` when it should have been `Use good instead of bad` to begin with.

spec/rubocop/cop/rspec/capybara/negation_matcher_spec.rb
```diff
@@ -0,0 +1,97 @@
...
+    let(:enforced_style) { 'have_no' }
+
+    %i[selector css xpath text title current_path link button
+       field checked_field unchecked_field select table
+       sibling ancestor].each do |matcher|
+      it 'registers an offense when using ' \
+        '`expect(...).not_to have_#{matcher}`' do
+        expect_offense(<<~RUBY, matcher: matcher)
+          expect(page).not_to have_#{matcher}
+                       ^^^^^^^^^^^^^{matcher} Use `expect(...).to have_no_#{matcher}` instead of `expect(...).not_to have_#{matcher}`.
```
pirj, line 16: I still think that the "instead of ..." part is redundant. What makes you feel to keep it?

lib/rubocop/cop/rspec/capybara/negation_matcher.rb
```diff
@@ -0,0 +1,105 @@
...
+        #   # good
+        #   expect(page).to have_no_selector
+        #   expect(page).to have_no_css('a')
+        #
+        class NegationMatcher < Base
+          extend AutoCorrector
+          include ConfigurableEnforcedStyle
+
+          MSG = 'Use `%<good>s` instead of `%<bad>s`.'
+          CAPYBARA_MATCHERS = %w[
```
pirj, line 32: If you `%i[` here, `to_sym` below becomes redundant, and will possibly allow to squeeze those statements into single lines.

Darhazer: to_sym is because of prefixing with have_ and have_no

pirj: My bad.

lib/rubocop/cop/rspec/capybara/negation_matcher.rb
```diff
@@ -0,0 +1,105 @@
...
+
+          def replaced_target
+            style == :have_no ? 'to' : 'not_to'
+          end
+
+          def replaced_matcher(node)
+            if style == :have_no
+              node.loc.selector.source.gsub('have_', 'have_no_')
+            else
+              node.loc.selector.source.gsub('have_no_', 'have_')
```
bquorning, line 98:
```suggestion
              node.loc.selector.source.sub('have_', 'have_no_')
            else
              node.loc.selector.source.sub('have_no_', 'have_')
```

lib/rubocop/cop/rspec/capybara/negation_matcher.rb
```diff
@@ -0,0 +1,105 @@
...
+              corrector.replace(node.parent.loc.selector, replaced_target)
+              corrector.replace(node.loc.selector, replaced_matcher(node))
+            end
+          end
+
+          private
+
+          def offense?(node)
+            (style == :have_no && not_to?(node)) ||
+              (style == :not_to && have_no?(node))
```
bquorning, line 71: It seems like the `style` check here, inside `#replaced_target` and inside `#replaced_matcher` is implemented in three different ways – the two latter use “else” instead of an explicit `style == :not_to`.

How about using e.g.
```rb
if style == :have_no
  ...
else
  ...
end
```
or
```rb
case style
when :have_no
  ...
when :not_to
  ...
end
```
in all three places?

lib/rubocop/cop/rspec/capybara/negation_matcher.rb
ydah: Indeed it is. I will change it to `runner` as I think it is better to keep internal consistency.

spec/rubocop/cop/rspec/capybara/negation_matcher_spec.rb
ydah: I was thinking something more descriptive would be better, but I'll remove it here as it may just be redundant.

lib/rubocop/cop/rspec/capybara/negation_matcher.rb
ydah: I unified with the style expressed in case.

bquorning (approved):

lib/rubocop/cop/rspec/capybara/negation_matcher.rb
```diff
@@ -0,0 +1,105 @@
...
+            end
+          end
+
+          def replaced_matcher(node)
+            case style
+            when :have_no
+              node.loc.selector.source.sub('have_', 'have_no_')
+            when :not_to
+              node.loc.selector.source.sub('have_no_', 'have_')
+            end
```
Darhazer, line 99: Isn't it the same as
```suggestion
            when :have_no
              node.method_name.sub('have_', 'have_no_')
            when :not_to
              node.method_name.sub('have_no_', 'have_')
            end
```

pirj:
Thank you, @ydah !

Issue #1385, opened 2022-09-12
`RSpec/NoExpectationExample` false positives


boris-petrov:
```rb
scenario 'test' do
  method
end

private

def method
  expect(1).to eq(1)
end
```

This gives `RSpec/NoExpectationExample` but it shouldn't. :)

ShockwaveNN:
Similar problem then expect in `after` block

Something like this:

```ruby
RSpec.describe do
  after do
    expect('File exists').not_to be_nil
  end

  it 'foo' do
    p 'crete_file'
  end

  it 'bar' do
    p 'crete_file another way'
  end
end
```

annaswims:
It also does not accept `Capybara::Node::Matchers#assert_text` as an assertion in an example like:
```ruby
    it "is not authorized" do
      assert_text("You are not authorized to perform this action.")
    end
```

ShockwaveNN:
A lot of edge cases for single cop, maybe it should be disabled by default

annaswims:
Thanks! I missed [the docs](https://docs.rubocop.org/rubocop-rspec/cops_rspec.html#rspecnoexpectationexample)

Darhazer:
I think we have to account for the expectations in hooks that are in scope. We have the `ExpectInHook` that would disallow such usage, but it would be frustrating for users to disable that cop just to have `NoExpectationExample` complain again.

Maybe we also should add the capybara expectations by default to the language (perhaps behind a config option). I'm on a fense for this one, as it would affect things like `MultipleExpectations` - perhaps the style for capybara scenarios is a bit different. 

@rubocop/rubocop-rspec WDYT?

bquorning:
I think this would be Koch easier to check with dynamic analysis that with static analysis. @pirj Is this something that might be built into RSpec itself one day?

pirj:
No, it's not possible in RSpec, too, since rspec-mocks/rspec-expectations are not the only way of making expectations, e.g. `assert_nil` being one I frequently see. There was a discussion about this somewhere in RSpec issue tracker, but I can't quickly find it.

Darhazer:
I have reviewed a codebase with the cop applied, and now I think that we may add an `AllowedPatterns` configuration, so people can just whitelist `expect_` and `assert_` methods without needing to add them to Language. Though having not them in language also forces people to think how to better express them, e.g. creating custom matchers, instead of merely methods that run expectations

AlexWayfer:
[...]

PR #1408, opened 2022-10-10
Add `AllowedPatterns` configuration option to `RSpec/NoExpectationExample`

ydah:
Fix: https://github.com/rubocop/rubocop-rspec/issues/1385

This cop can be customized allowed expectation methods pattern with `AllowedPatterns`. \Aexpect_ and \Aassert_ are allowed by default.

### @ example `AllowedPatterns: [^expect_]`
```ruby
# bad
it do
  not_expect_something
end

# good
it do
  expect_something
end
```

---

* [x] Added tests.
* [x] Updated documentation.

### DIFF
--- a/lib/rubocop/cop/rspec/no_expectation_example.rb
+++ b/lib/rubocop/cop/rspec/no_expectation_example.rb
@@ -28,7 +28,35 @@ module RuboCop
       #     expect(a?).to be(true)
       #   end
       #
+      # This cop can be customized with an allowed expectation methods pattern
+      # with an `AllowedPatterns` option. ^expect_ and ^assert_ are allowed
+      # by default.
+      #
+      # @example `AllowedPatterns` configuration
+      #
+      #   # .rubocop.yml
+      #   # RSpec/NoExpectationExample:
+      #   #   AllowedPatterns:
+      #   #     - ^expect_
+      #   #     - ^assert_
+      #
+      # @example
+      #   # bad
+      #   it do
+      #     not_expect_something
+      #   end
+      #
+      #   # good
+      #   it do
+      #     expect_something
+      #   end
+      #
+      #   it do
+      #     assert_something
+      #   end
+      #
       class NoExpectationExample < Base
+        include AllowedPattern
         MSG = 'No expectation found in this example.'
 
         # @!method regular_or_focused_example?(node)
@@ -41,26 +69,28 @@ module RuboCop
           }
         PATTERN
 
-        # @!method including_any_expectation?(node)
+        # @!method includes_expectation?(node)
         # @param [RuboCop::AST::Node] node
         # @return [Boolean]
-        def_node_search(
-          :including_any_expectation?,
-          send_pattern('#Expectations.all')
-        )
+        def_node_search :includes_expectation?, <<~PATTERN
+          {
+            #{send_pattern('#Expectations.all')}
+            (send nil? `#matches_allowed_pattern?)
+          }
+        PATTERN
 
-        # @!method including_any_skip_example?(node)
+        # @!method includes_skip_example?(node)
         # @param [RuboCop::AST::Node] node
         # @return [Boolean]
-        def_node_search :including_any_skip_example?, <<~PATTERN
+        def_node_search :includes_skip_example?, <<~PATTERN
           (send nil? {:pending :skip} ...)
         PATTERN
 
         # @param [RuboCop::AST::BlockNode] node
         def on_block(node)
           return unless regular_or_focused_example?(node)
-          return if including_any_expectation?(node)
-          return if including_any_skip_example?(node)
+          return if includes_expectation?(node)
+          return if includes_skip_example?(node)
 
           add_offense(node)
         end
--- a/spec/rubocop/cop/rspec/no_expectation_example_spec.rb
+++ b/spec/rubocop/cop/rspec/no_expectation_example_spec.rb
@@ -137,6 +137,53 @@ RSpec.describe RuboCop::Cop::RSpec::NoExpectationExample do
     end
   end
 
+  context 'when `AllowedPatterns: [^expect_]`' do
+    let(:cop_config) { { 'AllowedPatterns' => ['^expect_'] } }
+
+    context 'when only allowed pattern methods are used' do
+      it 'registers no offenses' do
+        expect_no_offenses(<<~RUBY)
+          it { expect_something }
+        RUBY
+      end
+    end
+
+    context 'when allowed pattern methods and other method are used' do
+      it 'registers no offenses' do
+        expect_no_offenses(<<~RUBY)
+          it do
+            do_something
+            expect_something
+            do_something
+          end
+        RUBY
+      end
+    end
+
+    context 'when nested allowed pattern methods and other method are used' do
+      it 'registers no offenses' do
+        expect_no_offenses(<<~RUBY)
+          it do
+            do_something
+            some_patterns.each do
+              expect_something
+            end
+            do_something
+          end
+        RUBY
+      end
+    end
+
+    context 'when only not allowed pattern methods are used' do
+      it 'does something' do
+        expect_offense(<<~RUBY)
+          it { not_expect_something }
+          ^^^^^^^^^^^^^^^^^^^^^^^^^^^ No expectation found in this example.
+        RUBY
+      end
+    end
+  end
+
   context 'when Ruby 2.7', :ruby27 do
     context 'with no expectation example with it' do
       it 'registers an offense' do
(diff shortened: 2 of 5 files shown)


lib/rubocop/cop/rspec/no_expectation_example.rb
```diff
@@ -28,7 +28,34 @@ module RSpec
...
+      # This cop can be customized allowed expectation methods pattern
+      # with `AllowedPatterns`. \Aexpect_ and \Aassert_ are allowed by default.
+      #
+      # @example `AllowedPatterns` configuration
+      #
+      #   # .rubocop.yml
+      #   # RSpec/NoExpectationExample:
+      #   #   AllowedPatterns:
+      #   #     - \Aexpect_
+      #   #     - \Aassert_
```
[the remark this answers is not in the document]
pirj: By the way, should it be double escaped? I see [just `\A` as a suggestion in RuboCop's own 'config/default.yml` here](https://github.com/rubocop/rubocop/blob/d1270c468cdb9a211229e36aef8d568ae3bfe607/config/default.yml#L2786). I couldn't find any other examples there.

I don't have a strong preference, apart from double escaping.
But just to make sure the default config's double escaping is correct, I'd test this change live, just by running `rubocop` on a code that has `expect_`/`allow_` to see if the cop ignores those, as we stub the config in our spec.

lib/rubocop/cop/rspec/no_expectation_example.rb
```diff
@@ -61,11 +88,22 @@ def on_block(node)
...
 
           add_offense(node)
         end
 
         alias on_numblock on_block
+
+        private
+
+        def including_any_allowed_pattern?(node)
+          node.each_descendant(:send) do |send_node|
```
pirj, line 101: Does this to find allowed expectations in iterators and other nested statements?
Wondering how this affects performance. We run this only once on an example node, right? Shouldn't be a problem then.

pirj: Is nesting worth an additional example in specs?

pirj, line 101: Wondering if `node.each_descendant(:send).find { |send_node| ... }` would work to avoid returning from a block.

lib/rubocop/cop/rspec/no_expectation_example.rb
```diff
@@ -28,7 +28,34 @@ module RSpec
       #     expect(a?).to be(true)
       #   end
       #
+      # This cop can be customized allowed expectation methods pattern
+      # with `AllowedPatterns`. \Aexpect_ and \Aassert_ are allowed by default.
```
pirj, line 32:
```suggestion
      # This cop can be customized with an allowed expectation methods pattern
      # with an `AllowedPatterns` option. \Aexpect_ and \Aassert_ are allowed by default.
```

pirj (approved):
Thank you!
Looks good.
I'll skip running this on `real-world-rspec` myself, but highly encourage someone to do this prior to the release.

lib/rubocop/cop/rspec/no_expectation_example.rb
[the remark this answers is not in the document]
ydah: Indeed it is. I will change it to ``^``.

> By the way, should it be double escaped? I see [just \A as a suggestion in RuboCop's own 'config/default.yml` here](https://github.com/rubocop/rubocop/blob/d1270c468cdb9a211229e36aef8d568ae3bfe607/config/default.yml#L2786). I couldn't find any other examples there.

Oh, I didn't escape there, but I think it may have been escaped by ConfigFormatter.

lib/rubocop/cop/rspec/no_expectation_example.rb
ydah: Thank you very much. It certainly looks better as you suggested, so I will make the change.

ydah: I tried it on a large project at hand, and as shown below, it seems to be slowing down a bit, but does not seem to be a major performance issue.

9481 files inspected, no offenses detected
bundle exec rubocop --only RSpec/NoExpectationExample  17.42s user 11.90s system 106% cpu 27.638 total

9481 files inspected, no offenses detected
bundle exec rubocop --only RSpec/NoExpectationExample  19.12s user 13.86s system 89% cpu 36.810 total

ydah: In the case of the following codes, the ASTs will be those indicated in the comments below them, respectively.

```ruby
it { expect_something }
# node
# s(:block,
#   s(:send, nil, :it),
#   s(:args),
#   s(:send, nil, :expect_something))    <--- send node

it do
  do_something
  expect_something
  do_something
end
# node
s(:block,
#   s(:send, nil, :it),
#   s(:args),
#   s(:begin,    <--- Nodes are nested
#     s(:send, nil, :do_something),    <--- send node
#     s(:send, nil, :expect_something),    <--- send node
#     s(:send, nil, :do_something)))    <--- send node
```

I was going to check target all of the send_node methods and use regular expressions to see if any are specified in AllowedPatterns.

pirj: Wondering if the usage of ["`" node pattern descendent matching syntax](https://docs.rubocop.org/rubocop-ast/node_pattern.html#for-descendants) along with [dynamic checks](https://docs.rubocop.org/rubocop-ast/node_pattern.html#to-call-functions) would change this result.
In any case, I don't see this as a blocker, can be a theoretical performance improvement in some follow-up PR, or a ticket.

pirj:
Running the cop against `real-world-rspec`:
```
57497 files inspected, 9464 offenses detected
```
before the change:
```
57497 files inspected, 23997 offenses detected
```

Fantastic!

lib/rubocop/cop/rspec/no_expectation_example.rb
```diff
@@ -61,11 +89,20 @@ def on_block(node)
           return unless regular_or_focused_example?(node)
           return if including_any_expectation?(node)
           return if including_any_skip_example?(node)
+          return if including_any_allowed_pattern?(node)
```
Darhazer, line 92: I'd rather modify the including_any_expectation? method to account for the allowed patterns

ydah: @Darhazer @pirj 
Fixed to extend `including_any_expectation?` How do you like it? Could you review it again?

spec/rubocop/cop/rspec/no_expectation_example_spec.rb
```diff
@@ -137,6 +137,39 @@
...
+        RUBY
+      end
+    end
+
+    context 'with allowed pattern methods and other method are used' do
+      it 'registers no offenses' do
+        expect_no_offenses(<<~RUBY)
+          it do
+            do_something
+            expect_something
```
Darhazer, line 156: perhaps you can demonstrate the recursiveness of the node_search with nested blocks for instance

spec/rubocop/cop/rspec/no_expectation_example_spec.rb
```diff
@@ -137,6 +137,39 @@
     end
   end
 
+  context 'when `AllowedPatterns: [^expect_]`' do
+    let(:cop_config) { { 'AllowedPatterns' => ['^expect_'] } }
+
+    context 'with only allowed pattern methods are used' do
```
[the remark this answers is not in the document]
ydah: I was using `with` to match the other tests, but it certainly looks like `when` is fine.

spec/rubocop/cop/rspec/no_expectation_example_spec.rb
ydah: Tests were added. How do you like it?

pirj:
Nice!

While looking at the cop, I think as the next step it would make sense to also make it tolerant to `:skip`/`:pending`/`skip: true`/`pending: true` metadata, and `skip`/`pending` example aliases.

Darhazer:
> While looking at the cop, I think as the next step it would make sense to also make it tolerant to :skip/:pending/skip: true/pending: true metadata, and skip/pending example aliases.

To the best of my knowledge, it already handles skip/pending calls. Perhaps it does not handle the metadata. Would you mind checkng and opening an issue>

spec/rubocop/cop/rspec/no_expectation_example_spec.rb
Darhazer: Thank you :bow:

Darhazer (approved):

lib/rubocop/cop/rspec/no_expectation_example.rb
```diff
@@ -41,26 +69,28 @@ class NoExpectationExample < Base
...
         # @param [RuboCop::AST::Node] node
         # @return [Boolean]
-        def_node_search(
-          :including_any_expectation?,
-          send_pattern('#Expectations.all')
-        )
+        def_node_search :includes_expectation?, <<~PATTERN
+          {
+            #{send_pattern('#Expectations.all')}
+            (send nil? `#matches_allowed_pattern?)
```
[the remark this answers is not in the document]
ydah: Indeed it is. Thank you so much.
I open PR: https://github.com/rubocop/rubocop-rspec/pull/1434

PR #1410, opened 2022-10-13
Enable some new cops

bquorning:
I know we have have RuboCop’s new cops disabled (see #1065) because they were a bit unstable, and we want to test on CI against the latest version of RuboCop _without_ getting hit by bugs in new cops. But a while back I added a couple of new cops (`Layout/LineContinuationSpacing` and `Layout/LineEndStringConcatenationIndentation`), and I really miss having them enabled in this repository. So here they are (plus one extra “easy” cop that caused offense).

---

* [ ] Added tests.
* [ ] Updated documentation.


pirj (approved):
Looks great

PR #1413, opened 2022-10-19
Ignore groups with include_examples for autocorrection by LetBeforeExamples

### LINKED ISSUE #1392: RSpec/LetBeforeExamples is not a safe autocorrection
Hello,

I have found an instance where the autocorrection for `RSpec/LetBeforeExamples` changes the behavior.  I believe this cop should be marked as an unsafe autocorrection.

The issue is when a `RSpec.shared_examples` block defines a `let` with the same name as one defined closer to the main examples:

```ruby
# frozen_string_literal: true

Box = Struct.new(:value)

RSpec.shared_examples 'when overriding value' do
  let(:value) { 1 }
end

RSpec.describe 'RSpec/LetBeforeExamples' do
  include_examples 'when overriding value'

  let(:value) { 2 }

  it { expect(value).to eq(2) }
end
```

In this case, the cop will move the `let(:value)` above the `include_examples` call, which changes the behavior.

See example repo here: https://github.com/gburgett/rubocop-let-before-examples

pirj:
fixes #1392

Moving `let` above `include_examples` that defines the same variable may break the spec as the order of definition changes.

---

* [x] Added tests.
* [-] Updated documentation.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,11 +7,12 @@
 * Update `RSpec/ExampleWording` cop to raise error for insufficient descriptions. ([@akrox58][])
 * Add new `RSpec/Capybara/NegationMatcher` cop. ([@ydah][])
 * Add `AllowedPatterns` configuration option to `RSpec/NoExpectationExample`.  ([@ydah][])
-* Improve `RSpec/NoExpectationExample` cop to ignore examples skipped or pending via metatada.  ([@pirj][])
+* Improve `RSpec/NoExpectationExample` cop to ignore examples skipped or pending via metadata.  ([@pirj][])
 * Add `RSpec/FactoryBot/ConsistentParenthesesStyle` cop. ([@Liberatys][])
 * Add `RSpec/Rails/InferredSpecType` cop. ([@r7kamura][])
 * Add new `RSpec/Capybara/SpecificActions` cop. ([@ydah][])
 * Update `config/default.yml` removing deprecated option to make the config correctable by users. ([@ignaciovillaverde][])
+* Do not attempt to auto-correct example groups with `include_examples` in `RSpec/LetBeforeExamples`. ([@pirj][])
 
 ## 2.13.2 (2022-09-23)
 
--- a/lib/rubocop/cop/rspec/let_before_examples.rb
+++ b/lib/rubocop/cop/rspec/let_before_examples.rb
@@ -43,6 +43,14 @@ module RuboCop
           }
         PATTERN
 
+        # @!method include_examples?(node)
+        def_node_matcher :include_examples?, <<~PATTERN
+          {
+            #{block_pattern(':include_examples')}
+            #{send_pattern(':include_examples')}
+          }
+        PATTERN
+
         def on_block(node) # rubocop:disable InternalAffairs/NumblockHandler
           return unless example_group_with_body?(node)
 
@@ -50,6 +58,10 @@ module RuboCop
         end
 
         private
+
+        def example_group_with_include_examples?(body)
+          body.children.any? { |sibling| include_examples?(sibling) }
+        end
 
         def multiline_block?(block)
           block.begin_type?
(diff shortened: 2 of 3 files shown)


Darhazer:
It would be nice if we can conditionally mark the auto-correct as unsafe / run some auto-corrections only if rubocop was run with unsafe corrections enabled (so it would output `N more could be corrected if you run with -A`

lib/rubocop/cop/rspec/let_before_examples.rb
```diff
@@ -43,14 +43,27 @@ class LetBeforeExamples < Base
...
+        def_node_matcher :include_examples?, <<~PATTERN
+          {
+            #{block_pattern(':include_examples')}
+            #{send_pattern(':include_examples')}
+          }
+        PATTERN
+
         def on_block(node) # rubocop:disable InternalAffairs/NumblockHandler
           return unless example_group_with_body?(node)
+          return if example_group_with_included_examples?(node.body)
```
Darhazer, line 56: shouldn't we disable just the correction, not the offence?

spec/rubocop/cop/rspec/let_before_examples_spec.rb
```diff
@@ -41,10 +41,23 @@
     RUBY
   end
 
-  it 'flags `let` after `include_examples`' do
+  it 'ignores `let` after `include_examples`' do
+    # NOTE: include_examples may define the same variable as `let`,
+    # and changing the order may break the spec due to order dependency
+    # if `let` is moved above.
+    expect_no_offenses(<<-RUBY)
```
Darhazer, line 48: Looking at the condition `if if multiline_block?(node.body)` in the original code, I have the feeling that this may fail

```
RSpec.describe User do
         include_examples('should be BEFORE let as it defines `let(:foo)`, too')
       end
     RUBY
```

lib/rubocop/cop/rspec/let_before_examples.rb
```diff
@@ -43,14 +43,27 @@ class LetBeforeExamples < Base
...
 
           check_let_declarations(node.body) if multiline_block?(node.body)
         end
 
         private
 
+        def example_group_with_included_examples?(body)
+          body.children.find { |sibling| include_examples?(sibling) }
```
Darhazer, line 64: since we are not interested in the actual include_examples node, I'd replace `find` for `any?`

pirj:
> N more could be corrected if you run with -A

Great idea! I'd use that.

lib/rubocop/cop/rspec/let_before_examples.rb
pirj: Very well noticed.

spec/rubocop/cop/rspec/let_before_examples_spec.rb
pirj: I couldn't make it fail. As I've moved the check below `multiline_block?`, there should be no issues.

CHANGELOG.md
```diff
@@ -7,8 +7,9 @@
 * Update `RSpec/ExampleWording` cop to raise error for insufficient descriptions. ([@akrox58][])
 * Add new `RSpec/Capybara/NegationMatcher` cop. ([@ydah][])
 * Add `AllowedPatterns` configuration option to `RSpec/NoExpectationExample`.  ([@ydah][])
-* Improve `RSpec/NoExpectationExample` cop to ignore examples skipped or pending via metatada.  ([@pirj][])
```
bquorning, line 10: :tada:

bquorning (approved):

CHANGELOG.md
```diff
@@ -7,8 +7,9 @@
...
+* Improve `RSpec/NoExpectationExample` cop to ignore examples skipped or pending via metadata.  ([@pirj][])
 * Add `RSpec/FactoryBot/ConsistentParenthesesStyle` cop. ([@Liberatys][])
+* Ignore example groups with `include_context` in `RSpec/LetBeforeExamples`. ([@pirj][])
```
Darhazer, line 12: isn't it `include_examples` ? Also it ignores them only for auto-correct. So something along the lines `Do not attempt to auto-correct example groups with `include_examples` ) would be better. It would be good to link to the actual issue for context

Darhazer (approved):

pirj:
@Darhazer I've started filing a RuboCop ticket and found that if you run `rubocop -a`, you get this:

```
1 file inspected, 1 offense detected, 1 more offense can be corrected with `rubocop -A`
```

PR #1414, opened 2022-10-19
Add `RSpec/FactoryBot/AssociationStyle` cop

r7kamura:
In FactoryBot, there are 2 ways to define associations.

```ruby
# implicit style
factory :article do
  user
end

# explicit style
factory :article do
  association :user
end
```

These styles are documented at the following section:

- https://github.com/thoughtbot/factory_bot/blob/ec71611954d07419c9b668be03641332443dcd78/GETTING_STARTED.md#associations

(Stricyly speaking, there is also inline style, but it behaves differently from the above 2 styles.)

It would be nice to have a cop for consistent use of only one of the styles. This Pull Request has been created for this purpose.

---

* [x] Added tests.
* [x] Updated documentation.

### DIFF
--- a/config/default.yml
+++ b/config/default.yml
@@ -913,6 +913,22 @@ RSpec/FactoryBot:
   Include: *1
   Language: *2
 
+RSpec/FactoryBot/AssociationStyle:
+  Description: Use consistent style to define associations.
+  Enabled: pending
+  Safe: false
+  Include:
+    - spec/factories.rb
+    - spec/factories/**/*.rb
+    - features/support/factories/**/*.rb
+  VersionAdded: "<<next>>"
+  EnforcedStyle: implicit
+  SupportedStyles:
+    - explicit
+    - implicit
+  NonImplicitAssociationMethodNames: ~
+  Reference: https://www.rubydoc.info/gems/rubocop-rspec/RuboCop/Cop/RSpec/FactoryBot/AssociationStyle
+
 RSpec/FactoryBot/AttributeDefinedStatically:
   Description: Always declare attribute values as blocks.
   Enabled: true
--- a/docs/modules/ROOT/pages/cops_rspec_factorybot.adoc
+++ b/docs/modules/ROOT/pages/cops_rspec_factorybot.adoc
@@ -1,5 +1,74 @@
 = RSpec/FactoryBot
 
+== RSpec/FactoryBot/AssociationStyle
+
+|===
+| Enabled by default | Safe | Supports autocorrection | Version Added | Version Changed
+
+| Pending
+| No
+| Yes (Unsafe)
+| <<next>>
+| -
+|===
+
+Use consistent style to define associations.
+
+=== Safety
+
+This cop may cause false-positives in `EnforcedStyle: explicit`
+case. It recognizes any method call that has no arguments as an
+implicit association but it might be a user-defined trait call.
+
+=== Examples
+
+==== EnforcedStyle: implicit (default)
+
+[source,ruby]
+----
+# bad
+association :user
+
+# good
+user
+----
+
+==== EnforcedStyle: explicit
+
+[source,ruby]
+----
+# bad
+user
+
+# good
+association :user
+
+# good (NonImplicitAssociationMethodNames: ['email'])
+email
+----
+
+=== Configurable attributes
+
+|===
+| Name | Default value | Configurable values
+
+| Include
+| `spec/factories.rb`, `+spec/factories/**/*.rb+`, `+features/support/factories/**/*.rb+`
+| Array
+
+| EnforcedStyle
+| `implicit`
+| `explicit`, `implicit`
+
+| NonImplicitAssociationMethodNames
+| `<none>`
+| 
+|===
+
+=== References
+
+* https://www.rubydoc.info/gems/rubocop-rspec/RuboCop/Cop/RSpec/FactoryBot/AssociationStyle
+
 == RSpec/FactoryBot/AttributeDefinedStatically
 
 |===
--- a/lib/rubocop/cop/rspec/factory_bot/association_style.rb
+++ b/lib/rubocop/cop/rspec/factory_bot/association_style.rb
@@ -1,0 +1,254 @@
+# frozen_string_literal: true
+
+module RuboCop
+  module Cop
+    module RSpec
+      module FactoryBot
+        # Use consistent style to define associations.
+        #
+        # @safety
+        #   This cop may cause false-positives in `EnforcedStyle: explicit`
+        #   case. It recognizes any method call that has no arguments as an
+        #   implicit association but it might be a user-defined trait call.
+        #
+        # @example EnforcedStyle: implicit (default)
+        #   # bad
+        #   association :user
+        #
+        #   # good
+        #   user
+        #
+        # @example EnforcedStyle: explicit
+        #   # bad
+        #   user
+        #
+        #   # good
+        #   association :user
+        #
+        #   # good (NonImplicitAssociationMethodNames: ['email'])
+        #   email
+        class AssociationStyle < Base # rubocop:disable Metrics/ClassLength
+          extend AutoCorrector
+
+          include ConfigurableEnforcedStyle
+
+          DEFAULT_NON_IMPLICIT_ASSOCIATION_METHOD_NAMES = %w[
+            association
+            sequence
+            skip_create
+            traits_for_enum
+          ].freeze
+
+          RESTRICT_ON_SEND = %i[factory trait].freeze
+
+          # @param node [RuboCop::AST::SendNode]
+          # @return [void]
+          def on_send(node)
+            bad_associations_in(node).each do |association|
+              add_offense(
+                association,
+                message: "Use #{style} style to define associations."
+              ) do |corrector|
+                autocorrect(corrector, association)
+              end
+            end
+          end
+
+          private
+
+          # @!method explicit_association?(node)
+          #   @param node [RuboCop::AST::SendNode]
+          #   @return [Boolean]
+          def_node_matcher :explicit_association?, <<~PATTERN
+            (send nil? :association sym ...)
+          PATTERN
+
+          # @!method implicit_association?(node)
+          #   @param node [RuboCop::AST::SendNode]
+          #   @return [Boolean]
+          def_node_matcher :implicit_association?, <<~PATTERN
+            (send nil? !#non_implicit_association_method_name? ...)
+          PATTERN
+
+          # @!method factory_option_matcher(node)
+          #   @param node [RuboCop::AST::SendNode]
+          #   @return [Array<Symbol>, Symbol, nil]
+          def_node_matcher :factory_option_matcher, <<~PATTERN
+            (send
+              nil?
+              :association
+              ...
+              (hash
+                <
+                  (pair
+                    (sym :factory)
+                    {
+                      (sym $_) |
+                      (array (sym $_)*)
+                    }
+                  )
+                  ...
+                >
+              )
+            )
+          PATTERN
+
+          # @!method trait_arguments_matcher(node)
+          #   @param node [RuboCop::AST::SendNode]
+          #   @return [Array<Symbol>, nil]
+          def_node_matcher :trait_arguments_matcher, <<~PATTERN
+            (send nil? :association _ (sym $_)* ...)
+          PATTERN
+
+          # @!method trait_option_matcher(node)
+          #   @param node [RuboCop::AST::SendNode]
+          #   @return [Array<Symbol>, nil]
+          def_node_matcher :trait_option_matcher, <<~PATTERN
+            (send
+              nil?
+              :association
+              ...
+              (hash
+                <
+                  (pair
+                    (sym :traits)
+                    (array (sym $_)*)
+                  )
+                  ...
+                >
+              )
+            )
+          PATTERN
+
+          # @param corrector [RuboCop::Cop::Corrector]
+          # @param node [RuboCop::AST::SendNode]
+          def autocorrect(corrector, node)
+            case style
+            when :explicit
+              autocorrect_to_explicit_style(corrector, node)
+            when :implicit
+              autocorrect_to_implicit_style(corrector, node)
+            end
+          end
+
+          # @param corrector [RuboCop::Cop::Corrector]
+          # @param node [RuboCop::AST::SendNode]
+          # @return [void]
+          def autocorrect_to_explicit_style(corrector, node)
+            arguments = [
+              ":#{node.method_name}",
+              *node.arguments.map(&:source)
+            ]
+            corrector.replace(node, "association #{arguments.join(', ')}")
+          end
+
+          # @param corrector [RuboCop::Cop::Corrector]
+          # @param node [RuboCop::AST::SendNode]
+          # @return [void]
+          def autocorrect_to_implicit_style(corrector, node)
+            source = node.first_argument.value.to_s
+            options = options_for_autocorrect_to_implicit_style(node)
+            unless options.empty?
+              rest = options.map { |option| option.join(': ') }.join(', ')
+              source += " #{rest}"
+            end
+            corrector.replace(node, source)
+          end
+
+          # @param node [RuboCop::AST::SendNode]
+          # @return [Boolean]
+          def autocorrectable_to_implicit_style?(node)
+            node.arguments.one?
+          end
+
+          # @param node [RuboCop::AST::SendNode]
+          # @return [Boolean]
+          def bad?(node)
+            case style
+            when :explicit
+              implicit_association?(node)
+            when :implicit
+              explicit_association?(node)
+            end
+          end
+
+          # @param node [RuboCop::AST::SendNode]
+          # @return [Array<RuboCop::AST::SendNode>]
+          def bad_associations_in(node)
+            children_of_factory_block(node).select do |child|
+              bad?(child)
+            end
+          end
+
+          # @param node [RuboCop::AST::SendNode]
+          # @return [Array<RuboCop::AST::Node>]
+          def children_of_factory_block(node)
+            block = node.parent
+            return [] unless block
+            return [] unless block.block_type?
+            return [] unless block.body
+
+            if block.body.begin_type?
+              block.body.children
+            else
+              [block.body]
+            end
+          end
+
+          # @param node [RuboCop::AST::SendNode]
+          # @return [Array<Symbol>]
+          def factory_names_from_explicit(node)
+            trait_names = trait_names_from_explicit(node)
+            factory_names = Array(factory_option_matcher(node))
+            result = factory_names + trait_names
+            if factory_names.empty? && !trait_names.empty?
+              result.prepend(node.first_argument.value)
+            end
+            result
+          end
+
+          # @param method_name [Symbol]
+          # @return [Boolean]
+          def non_implicit_association_method_name?(method_name)
+            non_implicit_association_method_names.include?(method_name.to_s)
+          end
+
+          # @return [Array<String>]
+          def non_implicit_association_method_names
+            DEFAULT_NON_IMPLICIT_ASSOCIATION_METHOD_NAMES +
+              (cop_config['NonImplicitAssociationMethodNames'] || [])
+          end
+
+          # @param node [RuboCop::AST::SendNode]
+          # @return [Hash{Symbol => String}]
+          def options_from_explicit(node)
+            return {} unless node.last_argument.hash_type?
+
+            node.last_argument.pairs.inject({}) do |options, pair|
+              options.merge(pair.key.value => pair.value.source)
+            end
+          end
+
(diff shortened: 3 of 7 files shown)


pirj:
It seems that over time, all the three progressed. I was still believing that the implicit way didn't allow specifying `factory:` as it was in 5.0. Good to know FB improved so much on this front.

Personally, I strongly prefer the inline definition, for three reasons:
 - usage of `instance` (could be done before 6.0 with some hack, too) to refer to the model from association definitions
 - refer to (transient) attributes
 - with one-to-many associations an approach with no need for `after(:create)` that implies `create` strategy, or multiple per-strategy hooks that tend to get out of sync

The downside is that it's the most verbose out of the three in most of the cases.


> it behaves differently from the above 2 styles

Is there anything else that I'm missing?

pirj:
The `implicit` style run with [`real-world-rspec`](https://github.com/rubocop/rubocop-rspec/pull/1414) yields some false positives:
```
/Users/pirj/source/real-world-rspec/gitlabhq/spec/factories/issues.rb:49:5: C: [Correctable] RSpec/FactoryBot/AssociationStyle: Use explicit style to define associations.
    factory :closed_issue, traits: [:closed]
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
/Users/pirj/source/real-world-rspec/gitlabhq/spec/factories/issues.rb:50:5: C: [Correctable] RSpec/FactoryBot/AssociationStyle: Use explicit style to define associations.
    factory :reopened_issue, traits: [:opened]
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

The `explicit` style run results in some errors:
```
An error occurred while RSpec/FactoryBot/AssociationStyle cop was inspecting /Users/pirj/source/real-world-rspec/cartodb/spec/factories/likes.rb:2:2.
undefined method `begin_type?' for nil:NilClass
/Users/pirj/source/rubocop-rspec/lib/rubocop/cop/rspec/factory_bot/association_style.rb:154:in `children_of_factory_block'
/Users/pirj/source/rubocop-rspec/lib/rubocop/cop/rspec/factory_bot/association_style.rb:142:in `bad_associations_in'
/Users/pirj/source/rubocop-rspec/lib/rubocop/cop/rspec/factory_bot/association_style.rb:45:in `on_send'
```

for e.g.:
```ruby
FactoryBot.define do
  factory :like, class: Carto::Like do
  end
end
```
```ruby
    factory 'valid site' do
      # after(:build) { |valid_site| valid_site.stubs(:valid?).returns(true) }
    end
```

pirj:
Implicit:
```
57497 files inspected, 191 offenses detected, 191 offenses autocorrectable
```
Explicit:
```
57497 files inspected, 911 offenses detected, 911 offenses autocorrectable
```

Going with the implicit as the default enforced style seems reasonable 👍

pirj:
For the reference, this (apart from probably `pending` -> `true`) was needed to run the cop against real-world-rspec:
image

spec/rubocop/cop/rspec/factory_bot/association_style_spec.rb
```diff
@@ -0,0 +1,133 @@
...
+  end
+
+  let(:cop_config) do
+    { 'EnforcedStyle' => enforced_style }
+  end
+
+  context 'when EnforcedStyle is :implicit' do
+    let(:enforced_style) do
+      :implicit
+    end
```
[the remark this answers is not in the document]
r7kamura: It seems to me that it would be better to have a cop that specifies the style of the block for `let`. Would it be possible to specify which style should be used in which case?

```ruby
let(:cop_config) do
  { 'EnforcedStyle' => enforced_style }
end

let(:enforced_style) { :implicit }
```

r7kamura:
> > it behaves differently from the above 2 styles
> 
> Is there anything else that I'm missing?

As noted in the [documentation](https://github.com/thoughtbot/factory_bot/blob/ec71611954d07419c9b668be03641332443dcd78/GETTING_STARTED.md#inline-definition), attributes defined by inline-style return `nil` when called with `attributes_for` strategy.

My personal preference is to use inline-style only when I need it and not otherwise. If I always use inline-style, it makes difficult to infer from the code whether I am doing so because inline-style is necessary in the case or not.

If there seems to be a certain number of people who want to use the inline-style all the time, it might be nice to have a style in this cop that always prefers the inline-style. The downside is that it is quite difficult to autocorrect from inline-style to the other styles for the reasons mentioned above.

lib/rubocop/cop/rspec/factory_bot/association_style.rb
```diff
@@ -0,0 +1,175 @@
+# frozen_string_literal: true
+
+module RuboCop
+  module Cop
+    module RSpec
+      module FactoryBot
+        # Use consistent style to define associations.
+        #
+        # @safety
+        #   This cop may cause false-positives in `EnforcedStyle: explicit`
```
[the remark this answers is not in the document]
r7kamura: Thanks, [I commented to the Issue](https://github.com/rubocop/rubocop/issues/11092#issuecomment-1287486294) about that 👍

lib/rubocop/cop/rspec/factory_bot/association_style.rb
```diff
@@ -0,0 +1,175 @@
...
+            corrector.replace(node, "association #{arguments.join(', ')}")
+          end
+
+          # @param corrector [RuboCop::Cop::Corrector]
+          # @param node [RuboCop::AST::SendNode]
+          # @return [void]
+          def autocorrect_to_implicit_style(
+            corrector,
+            node
+          )
```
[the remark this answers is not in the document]
r7kamura: Oops, thank you.
As a matter of fact, this cop was originally written at [my hobby rubocop-extension gem](https://github.com/r7kamura/sevencop), and this style was [recommended](https://github.com/r7kamura/sevencop/blob/1882039b73a3f07506861c454de4f3307f53374b/lib/rubocop/cop/sevencop/method_definition_arguments_multiline.rb) there. Anyway multi-line arguments style is not common, so I fixed it 👍

spec/rubocop/cop/rspec/factory_bot/association_style_spec.rb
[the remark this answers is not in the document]
pirj: "Use your best judgement and your common sense" is the rule we're using now.

pirj (commented):
A couple of cosmetic suggestions

lib/rubocop/cop/rspec/factory_bot/association_style.rb
```diff
@@ -0,0 +1,167 @@
...
+                association,
+                message: "Use #{style} style to define associations."
+              ) do |corrector|
+                autocorrect(corrector, association)
+              end
+            end
+          end
+
+          private
+
```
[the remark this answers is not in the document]
r7kamura: Thanks, but I'll simply delete the unused `MSG` constant and leave the rest as is. This is because RuboCop insists that [`#message` should not be overridden](https://github.com/rubocop/rubocop/blob/722d1c1d70222306c8d63627ec81777025f5b29d/lib/rubocop/cop/base.rb#L100).

pirj:
## `implicit`

```
57497 files inspected, 206 offenses detected, 206 offenses autocorrectable
```

### Concerns

```
/Users/pirj/source/real-world-rspec/hound/spec/factories.rb:77:5: C: [Correctable] RSpec/FactoryBot/AssociationStyle: Use implicit style to define associations.
    association :repo, :active, :private
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```
```
/Users/pirj/source/real-world-rspec/gitlabhq/spec/factories/clusters/kubernetes_namespaces.rb:5:5: C: [Correctable] RSpec/FactoryBot/AssociationStyle: Use implicit style to define
associations.
    association :cluster, :project, :provided_by_gcp
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```
Would the following be a valid FactoryBot syntax?
```ruby
repo :active, :private
cluster :project, :provided_by_gcp
```
[Using traits with associations](https://github.com/thoughtbot/factory_bot/blob/5c01f05/GETTING_STARTED.md#with-associations-1) only has examples with the explicit syntax.

## `explicit`

```
57497 files inspected, 1026 offenses detected, 1026 offenses autocorrectable
```

### False positives

#### Static attributes

Static attributes [were deprecated in FactoryBot 4.11 and removed in 5.0](https://thoughtbot.com/blog/deprecating-static-attributes-in-factory_bot-4-11).
4.11 is nearly 5 years old, but I don't think it is a reasonable decision to break support for it for several major repos.

```
/Users/pirj/source/real-world-rspec/sharetribe/spec/factories.rb:535:5: C: [Correctable] RSpec/FactoryBot/AssociationStyle: Use explicit style to define associations.
    version           '1'
    ^^^^^^^^^^^^^^^^^^^^^
/Users/pirj/source/real-world-rspec/sharetribe/spec/factories.rb:536:5: C: [Correctable] RSpec/FactoryBot/AssociationStyle: Use explicit style to define associations.
    content           CustomLandingPage::ExampleData::DATA_STR
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
/Users/pirj/source/real-world-rspec/sharetribe/spec/factories.rb:541:5: C: [Correctable] RSpec/FactoryBot/AssociationStyle: Use explicit style to define associations.
    enabled           true
    ^^^^^^^^^^^^^^^^^^^^^^
/Users/pirj/source/real-world-rspec/sharetribe/spec/factories.rb:545:5: C: [Correctable] RSpec/FactoryBot/AssociationStyle: Use explicit style to define associations.
    listing_id 123
    ^^^^^^^^^^^^^^
/Users/pirj/source/real-world-rspec/sharetribe/spec/factories.rb:546:5: C: [Correctable] RSpec/FactoryBot/AssociationStyle: Use explicit style to define associations.
    blocked_at Date.current
    ^^^^^^^^^^^^^^^^^^^^^^^
```

What do you think if the cop should only register an offence if the first argument is a hash?
Another option would be to introduce `TargetFactoryBotVersion`. Wondering if this is something we can do in a minor release, @bquorning @Darhazer ?

### Blockpass

```
/Users/pirj/source/real-world-rspec/diaspora/spec/factories.rb:69:5: C: [Correctable] RSpec/FactoryBot/AssociationStyle: Use explicit style to define associations.
    password_confirmation(&:password)
```

### `build_association`

```
/Users/pirj/source/real-world-rspec/sharetribe/spec/factories.rb:499:5: C: [Correctable] RSpec/FactoryBot/AssociationStyle: Use explicit style to define associations.
    build_association(:paypal_account)
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

### `sequence`

```
/Users/pirj/source/real-world-rspec/gitlabhq/spec/factories/terraform/state_version.rb:9:5: C: [Correctable] RSpec/FactoryBot/AssociationStyle: Use explicit style to define associations.
    sequence(:version)
    ^^^^^^^^^^^^^^^^^^

/Users/pirj/source/real-world-rspec/chatwoot/spec/factories/channel/channel_widget.rb:6:5: C: [Correctable] RSpec/FactoryBot/AssociationStyle: Use explicit style to define associat
ions.
    sequence(:widget_color, &:to_s)
```

### `factory`

```
/Users/pirj/source/real-world-rspec/gitlabhq/spec/factories/keys.rb:22:5: C: [Correctable] RSpec/FactoryBot/AssociationStyle: Use explicit style to define associations.
    factory :deploy_key, class: 'DeployKey'
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
/Users/pirj/source/real-world-rspec/gitlabhq/spec/factories/issues.rb:48:5: C: [Correctable] RSpec/FactoryBot/AssociationStyle: Use explicit style to define associations.
    factory :closed_issue, traits: [:closed]
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

### `to_create`

```
/Users/pirj/source/real-world-rspec/cartodb/spec/factories/geocodings.rb:4:5: C: [Correctable] RSpec/FactoryBot/AssociationStyle: Use explicit style to define associations.
    to_create(&:save)
    ^^^^^^^^^^^^^^^^^
```


❓ Side note about using traits

This is ambiguous:
```ruby
factory :word do
  letters { "bar" }
  type { "noun" }

  trait :article do
    type { :article }
  end

  factory :the do
    article
    letters { "the" }
  end
```
Unless we parse traits in context, it's hard to understand if `article` is an association or a trait.
With `factory :verb, parent: :word` placed in a separate file, this is getting impossible.

An alternative syntax makes it unambiguous:
```ruby
  factory :the, traits: [:article] do
    letters { "the" }
  end
```

Might be an idea for a cop. Still, we won't be able to rely on that cop to be turned on and all traits to be defined in a `factory` call.

pirj:
I can also think of `traits_for_enum` for completeness. Wondering if there's anything else available to use/override in the context of a factory that we have to consider ignoring.

r7kamura:
> Would the following be a valid FactoryBot syntax?
> 
> ```ruby
> repo :active, :private
> cluster :project, :provided_by_gcp
> ```

It seems that the implicit style does not allow such trait usage, so I need to fix `#autocorrect_to_implicit_style` for this.

```ruby
# frozen_string_literal: true

require 'bundler/inline'

gemfile do
  source 'https://rubygems.org'

  gem 'factory_bot'
end

FactoryBot.define do
  factory :article do
    trait :active do
      active { true }
    end
  end

  factory :comment do
    article :active
  end
end
```

```
$ ruby wrong_association.rb 
/home/r7kamura/.rbenv/versions/3.1.2/lib/ruby/gems/3.1.0/gems/factory_bot-6.2.1/lib/factory_bot/definition_proxy.rb:99:in `method_missing': undefined method 'article' in 'comment' factory (NoMethodError)
Did you mean? 'article { :active }'


        raise NoMethodError.new(<<~MSG)
        ^^^^^
        from wrong_association.rb:19:in `block (2 levels) in <main>'
        from /home/r7kamura/.rbenv/versions/3.1.2/lib/ruby/gems/3.1.0/gems/factory_bot-6.2.1/lib/factory_bot/syntax/default.rb:18:in `instance_eval'
        from /home/r7kamura/.rbenv/versions/3.1.2/lib/ruby/gems/3.1.0/gems/factory_bot-6.2.1/lib/factory_bot/syntax/default.rb:18:in `factory'
        from wrong_association.rb:18:in `block in <main>'
        from /home/r7kamura/.rbenv/versions/3.1.2/lib/ruby/gems/3.1.0/gems/factory_bot-6.2.1/lib/factory_bot/syntax/default.rb:37:in `instance_eval'
        from /home/r7kamura/.rbenv/versions/3.1.2/lib/ruby/gems/3.1.0/gems/factory_bot-6.2.1/lib/factory_bot/syntax/default.rb:37:in `run'
        from /home/r7kamura/.rbenv/versions/3.1.2/lib/ruby/gems/3.1.0/gems/factory_bot-6.2.1/lib/factory_bot/syntax/default.rb:7:in `define'
        from wrong_association.rb:11:in `<main>'
```

Instead, it can be written by using `factory` option:

```ruby
# frozen_string_literal: true

require 'bundler/inline'

gemfile do
  source 'https://rubygems.org'

  gem 'activerecord', '~> 7.0.0'
  gem 'factory_bot'
  gem 'sqlite3'
end

require 'active_record'
require 'logger'
require 'minitest/autorun'

ActiveRecord::Base.establish_connection(
  adapter: 'sqlite3',
  database: ':memory:'
)

ActiveRecord::Schema.define do
  create_table :articles, force: true do |t|
    t.boolean :active, default: false
  end

  create_table :comments, force: true do |t|
    t.references :article
  end
end

class Article < ActiveRecord::Base
end

class Comment < ActiveRecord::Base
  belongs_to :article
end

FactoryBot.define do
  factory :article do
    trait :active do
      active { true }
    end
  end

  factory :comment do
    article factory: [:article, :active]
  end
end

class FactoryTest < Minitest::Test
  def test_association
    assert ::FactoryBot.create(:comment).article.active
  end
end
```

```
$ ruby example.rb 
-- create_table(:articles, {:force=>true})
   -> 0.0024s
-- create_table(:comments, {:force=>true})
   -> 0.0003s
Run options: --seed 55593

# Running:

.

Finished in 0.010602s, 94.3227 runs/s, 94.3227 assertions/s.

1 runs, 1 assertions, 0 failures, 0 errors, 0 skips
```

pirj:
I might be talking nonsense, but as an alternative to
```ruby
article factory: [:article, :active]
```
would it be possible to write it as
```ruby
article traits: [:active]
```
Which one would you prefer?

r7kamura:
I feel you. I actually tried that example before posting the previous comment too, but I couldn't use `:traits` option on implicit style. In that example, I don't want to write "article" twice, and I too think that ideally `:traits` should be available here.

docs/modules/ROOT/pages/cops_rspec_factorybot.adoc
```diff
@@ -1,5 +1,78 @@
...
+skip_create
+----
+
+=== Configurable attributes
+
+|===
+| Name | Default value | Configurable values
+
+| inherit_mode
+| `{"merge"=>["NonImplicitAssociationMethodNames"]}`
```
pirj, line 56: Unrelated to this PR. Seems like a doc generation problem, this shouldn't be here.

config/default.yml
```diff
@@ -923,6 +923,28 @@ RSpec/FactoryBot:
...
+    - features/support/factories/**/*.rb
+  VersionAdded: "<<next>>"
+  EnforcedStyle: implicit
+  SupportedStyles:
+    - explicit
+    - implicit
+  NonImplicitAssociationMethodNames:
+    - association
+    - skip_create
+    - traits_for_enum
```
pirj, line 945: Does this have to be user configurable?

r7kamura: Certainly there is no need to make these default values removable. The main reason why I have this config value is for [Global sequences](https://github.com/thoughtbot/factory_bot/blob/5c01f05/GETTING_STARTED.md#global-sequences).

```ruby
Factory.define do
  sequence(:email) { |n| "person#{n}@example.com" }

  factory :user do
    email
  end
end
```

lib/rubocop/cop/rspec/factory_bot/association_style.rb
```diff
@@ -0,0 +1,254 @@
...
+        #
+        # @example EnforcedStyle: explicit
+        #   # bad
+        #   user
+        #
+        #   # good
+        #   association :user
+        #
+        #   # good (defined in NonImplicitAssociationMethodNames)
+        #   skip_create
```
pirj, line 29: What do you think of using something that looks like an global sequence attribute call here, e.g. `email`?
```suggestion
        #   # good (defined in NonImplicitAssociationMethodNames)
        #   email
```

lib/rubocop/cop/rspec/factory_bot/association_style.rb
```diff
@@ -0,0 +1,254 @@
...
+                autocorrect(corrector, association)
+              end
+            end
+          end
+
+          private
+
+          # @!method explicit_association?(node)
+          #   @param node [RuboCop::AST::SendNode]
+          #   @return [Boolean]
```
pirj, line 61: I think it returns a `Node`

lib/rubocop/cop/rspec/factory_bot/association_style.rb
r7kamura: Looks good 👍 
Besides, now that `skip_create` is set by constant instead of config, it is strange that this value is used as an example here.

lib/rubocop/cop/rspec/factory_bot/association_style.rb
r7kamura: In fact, `@return [true, nil]` seems to be the exact value. Given a pattern with no captures, `def_node_matcher` generates a method which will return `true` or `nil` (when called without block).

- https://github.com/rubocop/rubocop-ast/blob/e487124ee4a5160fd85d2a47ee4adde6ebcd0775/lib/rubocop/ast/node_pattern/method_definer.rb#L12-L15

The real example would look like this:

```ruby
def explicit_association?(param0 = self)
  node = param0
  
  return unless node.is_a?(::RuboCop::AST::Node) && node.children.size  + 1 >= 4 &&
node.send_type? &&
node.children[0].nil?() &&
:association === node.children[1] &&
node.children[2].is_a?(::RuboCop::AST::Node) && node.children[2].sym_type? &&
true
block_given? ? yield() : (return true)

end
```

And since this method is intended to be used to return a boolean, I think `@return [Boolean]` would be acceptable.
Or perhaps the policy could be to always forbid `@param` and `@return` to be written in YARD, as is taken in rubocop-rails.

pirj:

r7kamura:
Recreated on rubocop-factory_bot repo:

- https://github.com/rubocop/rubocop-factory_bot/pull/33

PR #1419, opened 2022-10-22
Add `RSpec/MemoizedHelperBlockDelimiter` cop

r7kamura:
In response to the review comment at https://github.com/rubocop/rubocop-rspec/pull/1414#discussion_r1001107034, I made a cop to be consistent in the way memoized helpers such as `let` and `subject` are written.

My personal opinion is that since these helpers are method definitions, they should be written in a style similar to that of writing `def`. I don't see many cases where `def` is written in one-liner style. For that reason, I thought that I should prefer to use a consistent style regardless of whether the content of the block would be short or long, and that `do_end` should be the default.

There is another reason for wanting the default to be this way: the braces style may conflict with `Style/BlockDelimiters` cop in some cases, so it may not be possible to autocorrect to always use braces. I would also like to avoid the put another style to use different delimiters for different situations as possible because this increases the cognitive load.

I think this cop is somewhat opinionated, so I would like to hear your opinions first.

---

* [x] Added tests.
* [x] Updated documentation.


lib/rubocop/cop/rspec/memoized_helper_block_delimiter.rb
```diff
@@ -0,0 +1,135 @@
...
+        # @return [Boolean]
+        def bad?(node)
+          memoized_helper_method?(node) &&
+            !preferred_block_delimiter?(node)
+        end
+
+        # @param node [RuboCop::AST::BlockNode]
+        # @return [Boolean]
+        def memoized_helper_method?(node)
+          let?(node) || subject?(node)
```
pirj, line 99: This will also match blockpass, should it?

pirj, line 99: I suggest turning this into a node pattern:
```
        {
          #{block_pattern('#Helpers.all')}
          #{block_pattern('#Subjects.all')}
        }
```

lib/rubocop/cop/rspec/memoized_helper_block_delimiter.rb
```diff
@@ -0,0 +1,135 @@
...
+          end
+          corrector.replace(node.location.begin, 'do')
+          corrector.replace(node.location.end, 'end')
+          wrap_in_newlines(corrector, node) if node.single_line?
+        end
+
+        # @param node [RuboCop::AST::BlockNode]
+        # @return [Boolean]
+        def bad?(node)
+          memoized_helper_method?(node) &&
```
pirj, line 92: This line doesn't semantically belong to "bad", it detects if it's a match.
I suggest moving it to `on_block`:
```ruby
memoized_helper_method(node) do |memoized_helper|
  next if preferred_block_delimiter?(memoized_helper)
  add_offense(...)
```

config/default.yml
```diff
@@ -558,6 +558,16 @@ RSpec/LetSetup:
   VersionAdded: '1.7'
   Reference: https://www.rubydoc.info/gems/rubocop-rspec/RuboCop/Cop/RSpec/LetSetup
 
+RSpec/MemoizedHelperBlockDelimiter:
+  Description: Use consistent style block delimiters for memoized helpers.
+  Enabled: pending
+  EnforcedStyle: do_end
```
pirj, line 564: I disagree with the choice of the default style.

`do_end` doesn't seem to be a frequent choice in the wild, according to a run against `real-world-rspec`:
```
57497 files inspected, 104170 offenses detected, 104170 offenses autocorrectable
```
as opposed to `braces`:
```
57497 files inspected, 17121 offenses detected
```

And, I suspect some part of do/end usages (offences to `braces` style) are due to the line length limitation:
```ruby
    let!(:conference) do
      WimbaConference.create!(title: "my conference", user: @user, duration: 60, context: course_factory)
    end
```

or because of being multi-line e.g. `/Users/pirj/source/real-world-rspec/spree/emails/spec/mailers/spree/order_mailer_spec.rb:25:22:`:
```ruby
  let(:second_order) do
    order = create(:completed_order_with_totals, email: 'test2@example.com')
    product = create(:product, name: %{The "BESTEST" product})
    variant = create(:variant, product: product)
    price = create(:price, variant: variant, amount: 15.00)
    store = second_store
    line_item = create(:line_item, variant: variant, order: order, quantity: 1, price: 4.99)
    allow(product).to receive_messages(default_variant: variant)
    allow(variant).to receive_messages(default_price: price)
    allow(order).to receive_messages(line_items: [line_item])
    allow(order).to receive_messages(store: store)
    order
  end
```

pirj (commented):
To be useful, I believe the cop should work together with `Layout/LineLength` somehow. If the style is `braces`, the helper should be inlined unless this would break `Layout/LineLength`.

The default style should be `braces`, and it shouldn't even raise an offence for do/end usages if this would break the line length constraint set by `Layout/LineLength`.

I believe it is possible to offer auto-correction for `braces`, too, as:
1. It's not necessary that `BlockDelimiters` cop is enabled in the user configuration
2. It is possible to inline (except when it breaks line length)

r7kamura:
Sorry for the late reply.
It is true that the default should be braces, considering the usage ratio of existing code, and `Layout/LineLength` should be taken into account. I don't know if I can implement this well, but I will give it a try 👌

r7kamura:
While I was able to implement the do-end style well, I could not come up with a good implementation or appropriate rules for the braces style. I'll close this Pull Request for now 🙇

PR #1371, opened 2022-08-26
Add mdformat to the workflow in GitHub Actions

ydah:
We would like to use mdformat as markdown checking.

I would propose that we use ydah/mdformat-action.
- https://github.com/ydah/mdformat-action

Examples of failure:
https://github.com/rubocop/rubocop-rspec/actions/runs/3019219851

---

Before submitting the PR make sure the following are checked:

* [x] Feature branch is up-to-date with `master` (if not - rebase it).
* [x] Squashed related commits together.
* [-] Added tests.
* [-] Updated documentation.
* [-] Added an entry to the `CHANGELOG.md` if the new code introduces user-observable changes.
* [x] The build (`bundle exec rake`) passes (be sure to run this locally, since it may produce updated documentation that you will need to commit).

If you have created a new cop:

* [-] Added the new cop to `config/default.yml`.
* [-] The cop is configured as `Enabled: pending` in `config/default.yml`.
* [-] The cop is configured as `Enabled: true` in `.rubocop.yml`.
* [-] The cop documents examples of good and bad code.
* [-] The tests assert both that bad code is reported and that good code is not reported.
* [-] Set `VersionAdded` in `default/config.yml` to the next minor version.

If you have modified an existing cop's configuration options:

* [-] Set `VersionChanged` in `config/default.yml` to the next major version.

PR #1420, opened 2022-10-22
Fix and re-enable a Changelog spec

bquorning:
Enabling the spec we marked as pending in #1371, where I didn’t know how to fix the spec. But [today I learned](https://ruby-doc.org/core-3.1.2/Regexp.html#class-Regexp-label-Options) about "subexpression modifiers" like `(?i)` and `(?-i)` to toggle case insensitivity on parts of a regular expression.✨

- [x] Added tests.
- [ ] Updated documentation.


bquorning:
cc @ydah

ydah:
Thank you for the follow up! looks good.

pirj:
Thank you!

PR #1425, opened 2022-10-23
Tell git to merge CHANGELOG.md by combining changes

pirj:
https://stackoverflow.com/questions/22109586/what-is-the-difference-between-union-and-text/22109606#22109606

[Discussion](https://github.com/rubocop/rubocop-rspec/discussions/1417)

- [-] Added tests.
- [-] Updated documentation.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,1 +1,1 @@
-CHANGELOG.md merge=union
+CHANGELOG.md text merge=union
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,3 @@
-# Changelog
-
-## Master (Unreleased)
-
 - Add `require_implicit` style to `RSpec/ImplicitSubject`. ([@r7kamura])
 - Fix a false positive for `RSpec/Capybara/SpecificMatcher` when `have_css("a")` without attribute. ([@ydah])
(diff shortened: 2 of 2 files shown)


bquorning (changes requested):
I need some time to experiment more with this.

On my machine, `git` knows full well how to merge two branches each adding "the same" line to the changelog. The question is: Why doesn’t GitHub know how to do this?

pirj:
> Why doesn’t GitHub know how to do this?

Local git config, @bquorning ?

It's quite unfortunate, but there's no way of committing git config for a repo so that GitHub and other developers can benefit from that. But it's possible to commit `.gitattributes`.

> When deciding what attributes are assigned to a path, Git consults $GIT_DIR/info/attributes file (which has the highest precedence), .gitattributes file in the same directory as the path in question, and its parent directories up to the toplevel of the work tree

Maybe you have `.gitattributes` in your `~`?

There's even a global machine setting that can affect that basing on the OS and distribution:
> Attributes for all users on a system should be placed in the $(prefix)/etc/gitattributes file.

bquorning:
More details in https://github.com/rubocop/rubocop-rspec/discussions/1417#discussioncomment-3944872

PR #1428, opened 2022-10-24
Ignore commits in the blame view 4874a5a

ydah:
This PR is ignore commits in the blame view.
Ignored formatting changes in markdowns: 4874a5a4a2a58e76d343aaa02279cd93b16f5a30

Ref: https://docs.github.com/en/repositories/working-with-files/using-files/viewing-a-file#ignore-commits-in-the-blame-view

- [-] Added tests.
- [-] Updated documentation.


bquorning (approved):
We can probably add a lot more commits to this list (when enabling various new RuboCop cops etc.) but this is a great start.

Thank you.

Issue #1436, opened 2022-10-24
RSpec/Rails/InferredSpecType causes cascading errors for the default structure of request specs


shepmaster:
To reproduce:

```
rvm use 3.1.2
rails new repro
cd repro
bundle add rspec rspec-rails rubocop rubocop-rspec
rails generate rspec:install
rails generate rspec:scaffold cow
cat > .rubocop.yml <<EOF
require:
  - rubocop-rspec

AllCops:
  NewCops: enable
EOF
rubocop spec/routing -A
rubocop spec/routing
```

The final run of `rubocop` will have this warning:

```
spec/routing/cows_routing_spec.rb:5:1: C: RSpec/FilePath: Spec path should end with cows_controller*_spec.rb.
RSpec.describe CowsController do
^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

It seems odd that the cop and the generated body of the request spec are at odds with each other.

---

- rspec (3.11.0)
- rspec-rails (6.0.1)
- rubocop (1.37.1)
- rubocop-rspec (2.14.1)

NOTE: cop introduced in #1365

shepmaster:
This _may_ be what y'all expected, and if so, I'd love to hear suggestions about what should be changed — the argument to  `describe` or the file path.

pirj:
I suggest making an exception and keeping `type: :routing` in `RSpec/InferredSpecType`. Just becase RSpec Rails' [controller](https://github.com/rspec/rspec-rails/blob/59a1d9a9f1fa947b00ab3e635455743b1f0956a7/lib/generators/rspec/controller/controller_generator.rb#L47) and [scaffold](https://github.com/rspec/rspec-rails/blob/59a1d9a9f1fa947b00ab3e635455743b1f0956a7/lib/generators/rspec/scaffold/scaffold_generator.rb#L64) have special file naming schema.

@bquorning @Darhazer @ydah @r7kamura Would you agree?

ydah:
I agree with you.
https://relishapp.com/rspec/rspec-rails/v/6-0/docs/routing-specs/named-routes

ydah:
I can work on it right after this, so if no one has started this yet, I will.

pirj:
My apologies, I did, hold on for a PR.

ydah:
I see, that's good to know :) Thank you very much.

r7kamura:
[...]

PR #1439, opened 2022-10-24
Skip removing routing type metadata

pirj:
fixes #1436

We have an exception in RSpec/FilePath for routing metadata, to allow such a name for specs with type: :routing.

RSpec Rails' [controller](https://github.com/rspec/rspec-rails/blob/59a1d9a9f1fa947b00ab3e635455743b1f0956a7/lib/generators/rspec/controller/controller_generator.rb#L47) and [scaffold](https://github.com/rspec/rspec-rails/blob/59a1d9a9f1fa947b00ab3e635455743b1f0956a7/lib/generators/rspec/scaffold/scaffold_generator.rb#L64) have special file naming schema.

**Replace this text with a summary of the changes in your PR. The more detailed you are, the better.**

______________________________________________________________________

Before submitting the PR make sure the following are checked:

- [x] Feature branch is up-to-date with `master` (if not - rebase it).
- [x] Squashed related commits together.
- [x] Added tests.
- [x] Updated documentation.
- [x] Added an entry to the `CHANGELOG.md` if the new code introduces user-observable changes.
- [x] The build (`bundle exec rake`) passes (be sure to run this locally, since it may produce updated documentation that you will need to commit).

PR #1441, opened 2022-10-24
Change `RSpec/FilePath` so that it only checks suffix when path is under spec/routing or type is defined as routing

r7kamura:
Fixes https://github.com/rubocop/rubocop-rspec/issues/1436.

Before the change, it was recognized as a routing spec if `type: :routing` was used.
After the change, it is recognized as a routing spec if `type: routing` is used OR the file path contains `spec/routing/`.

By the way, another check thay may have been missing will be also fixed in this pull request. Even if it is a routing spec, it is now also be checked if the file path ends with `_spec.rb`.

- [x] Added tests.
- [ ] Updated documentation.

### DIFF
--- a/lib/rubocop/cop/rspec/file_path.rb
+++ b/lib/rubocop/cop/rspec/file_path.rb
@@ -76,8 +76,6 @@ module RuboCop
           return unless top_level_groups.one?
 
           example_group(node) do |send_node, example_group, arguments|
-            next if routing_spec?(arguments)
-
             ensure_correct_file_path(send_node, example_group, arguments)
           end
         end
@@ -85,7 +83,7 @@ module RuboCop
         private
 
         def ensure_correct_file_path(send_node, example_group, arguments)
-          pattern = pattern_for(example_group, arguments.first)
+          pattern = pattern_for(example_group, arguments)
           return if filename_ends_with?(pattern)
 
           # For the suffix shown in the offense message, modify the regular
@@ -97,11 +95,13 @@ module RuboCop
         end
 
         def routing_spec?(args)
-          args.any?(&method(:routing_metadata?))
+          args.any?(&method(:routing_metadata?)) || routing_spec_path?
         end
 
-        def pattern_for(example_group, method_name)
-          if spec_suffix_only? || !example_group.const_type?
+        def pattern_for(example_group, arguments)
+          method_name = arguments.first
+          if spec_suffix_only? || !example_group.const_type? ||
+              routing_spec?(arguments)
             return pattern_for_spec_suffix_only
           end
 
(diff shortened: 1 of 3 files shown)


r7kamura:
> By the way, another check thay may have been missing will be also fixed in this pull request. Even if it is a routing spec, it is now also be checked if the file path ends with `_spec.rb`.

Perhaps there may be an argument that this suffix change and the spec/routing change should be separated into separate Pull Requests, in which case I will do it 🙆‍♂️

lib/rubocop/cop/rspec/file_path.rb
```diff
@@ -160,6 +159,14 @@ def relevant_rubocop_rspec_file?(_file)
         def spec_suffix_only?
           cop_config['SpecSuffixOnly']
         end
+
+        def routing_spec_path?
+          expanded_file_path.include?('spec/routing/')
+        end
```
r7kamura, line 165: I believe this is enough in most cases, but perhaps it should be `'/spec/routing/'` or more strict check.

pirj (approved):
I strongly prefer this to #1439 👍 

I wonder if `routing_metadata?` is really needed. I can experiment by removing it and running the cop on a corpus of repos.

bquorning:
I cannot help but feeling that `routing_spec?(arguments)` is a very Rails-specific method in the middle of a non-Rails specific class. It would probably be a bit cleaner if could find a more generic way to e.g. configure the cop to be more “loose” for certain paths.

lib/rubocop/cop/rspec/file_path.rb
```diff
@@ -97,11 +95,13 @@ def ensure_correct_file_path(send_node, example_group, arguments)
         end
 
         def routing_spec?(args)
-          args.any?(&method(:routing_metadata?))
+          args.any?(&method(:routing_metadata?)) || routing_spec_path?
```
pirj, line 98: I've removed this, and there were no differences between master/PR/PR with removed `routing_metadata?` runs against real-world-rspec.

```suggestion
          routing_spec_path?
```

r7kamura: If there is a test file like spec/routes/foo_spec.rb including `type: :routing` to indicate that it is a routing spec, it would fail by this change. But you don't think there are any such tests in real application, so that means it should be ok to remove this check, right?

pirj:
@bquorning @Darhazer One last fix, and it's all we need for 2.14.2, and we can slow down for a while?

bquorning:
Is this PR ok to merge or should we resolve the conversation above @pirj?

bquorning (approved):

pirj:
The PR is fine as is, we can clean this up later, or even keep for those who opt out from removing type metadata.

pirj:
Thank you, @r7kamura !

Issue #1442, opened 2022-10-25
Autocorrect FactoryBot/ConsistentParenthesesStyle with omit_parentheses produces syntax error from multi-line instances


DMA57361:
To match the style generally used throughout our code I've switched the new `RSpec/FactoryBot/ConsistentParenthesesStyle` cop to `EnforcedStyle: omit_parentheses`. We've some cases of FactoryBot.create calls spread across multiple lines (due to many arguments, large values, whatever) and if autocorrect is run the brackets are just removed... producing invalid code.

Given this contrived example file `spec/models/user_spec.rb`:
```
# frozen_string_literal: true

require 'rails_helper'

RSpec.describe User, type: :model do
  let(:user) do
    FactoryBot.create(
      :user,
      name:     'John Smith',
      username: 'jsmith',
      email:    'jsmith@example.org'
    )
  end
  specify { expect(user.username).to eq 'jsmith' }
end

```

Run `rubocop -a spec/models/user_spec.rb`:

```
$ rubocop -a spec/models/user_spec.rb
Inspecting 1 file
E

Offenses:

spec/models/user_spec.rb:7:16: C: [Corrected] RSpec/FactoryBot/ConsistentParenthesesStyle: Prefer method call without parentheses
    FactoryBot.create(
               ^^^^^^
spec/models/user_spec.rb:8:12: E: Lint/Syntax: unexpected token tCOMMA
(Using Ruby 2.7 parser; configure using TargetRubyVersion parameter, under AllCops)
      :user,
           ^
spec/models/user_spec.rb:13:3: C: [Corrected] RSpec/EmptyLineAfterFinalLet: Add an empty line after the last let.
  end
  ^^^

1 file inspected, 3 offenses detected, 2 offenses corrected
```

Presumably the first offense makes a change that causes the Lint/Syntax error?
The end result for the user_spec.rb is this, which is a syntax error:

```
# frozen_string_literal: true

require 'rails_helper'

RSpec.describe User, type: :model do
  let(:user) do
    FactoryBot.create 
      :user,
      name:     'John Smith',
      username: 'jsmith',
      email:    'jsmith@example.org'
    
  end

  specify { expect(user.username).to eq 'jsmith' }
end
```

Ideally multi-line instances need to be ignored even if the configuration is for `EnforcedStyle: omit_parentheses`

## RuboCop version

```
$ rubocop -V
[...]

PR #1443, opened 2022-10-25
Fix an incorrect autocorrect for `FactoryBot/ConsistentParenthesesStyle` with `omit_parentheses` option when method name and first argument are not on same line

ydah:
Fix: #1442

This PR is following cases where the method name and the first argument are different line are not considered offense.
```ruby
create(
  :user
)
build(
  :user,
  name: 'foo'
)
```

This is because the absence of parentheses will result in a syntax error.
```ruby
create
  :user

build
  :user,
  name: 'foo'
```

In that case, the following modification would be necessary, and we have chosen not to make it an offense here, since we would have to suppress either offense if there is a length limit in Layout/LineLength, etc.

```ruby
create :user

build :user,
  name: 'foo'
```

- [x] Added tests.
- [x] Updated documentation.

### DIFF
--- a/spec/rubocop/cop/rspec/factory_bot/consistent_parentheses_style_spec.rb
+++ b/spec/rubocop/cop/rspec/factory_bot/consistent_parentheses_style_spec.rb
@@ -275,5 +275,43 @@ RSpec.describe RuboCop::Cop::RSpec::FactoryBot::ConsistentParenthesesStyle do
         end
       RUBY
     end
+
+    context 'when create and first argument are on same line' do
+      it 'register an offense' do
+        expect_offense(<<~RUBY)
+          create(:user,
+          ^^^^^^ Prefer method call without parentheses
+            name: 'foo'
+          )
+        RUBY
+
+        expect_correction(<<~RUBY)
+          create :user,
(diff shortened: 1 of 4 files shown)


spec/rubocop/cop/rspec/factory_bot/consistent_parentheses_style_spec.rb
```diff
@@ -275,5 +275,25 @@
...
+          )
+        RUBY
+      end
+    end
+
+    context 'when build and first argument are not on same line' do
+      it 'does not register an offense' do
+        expect_no_offenses(<<~RUBY)
+          build(
+            :user
```
pirj, line 293: I think `build` is sufficiently covered. Instead, do you mind adding:
 - an example with two arguments (and a comma that is causing invalid syntax)
 - an example where the first argument and its comma are on the same line with the opening parenthesis

bquorning:
I think this is fine for a bugfix. We can consider the `IgnoreMultiline` in a separate PR.

pirj:
@bquorning I take your last comment as an approval.

@ydah Thank you!

PR #1446, opened 2022-10-29
Ignore calls without the first positional argument

### LINKED ISSUE #1445: `RSpec/FactoryBot/ConsistentParenthesesStyle` cop triggers not only on FactoryBot methods
The [`RSpec/FactoryBot/ConsistentParenthesesStyle`](https://github.com/rubocop/rubocop-rspec/blob/d065a8d0cfcc216a497b85027f060462f681cf45/lib/rubocop/cop/rspec/factory_bot/consistent_parentheses_style.rb#L33) cop introduced in [v2.14.0](https://github.com/rubocop/rubocop-rspec/releases/tag/v2.14.0) is triggered on methods with the same names as the ones listed in `RuboCop::RSpec::FactoryBot::Language::METHODS`, even though they're not from FactoryBot.

https://github.com/rubocop/rubocop-rspec/blob/d065a8d0cfcc216a497b85027f060462f681cf45/lib/rubocop/rspec/factory_bot/language.rb#L10-L28

For example, we have our own `create` method (which doesn't take any arguments) that gets affected by this cop.

```
spec/spec_helpers/test_database.rb:28:5: C: [Corrected] RSpec/FactoryBot/ConsistentParenthesesStyle: Prefer method call with parentheses
    create
    ^^^^^^
```

We could just disable `RSpec/FactoryBot`, but it would be helpful if its scope were narrower.

### Reproduce

The only thing that is needed to reproduce the behaviour is to create a file in `spec/` with a method definition and call like this:

```ruby
# spec/test.rb

module Test
  def self.create
    nil
  end

  create
end
```

pirj:
fixes #1445

- [x] Added tests.
- [-] Updated documentation.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## Master (Unreleased)
 
 - Add `named_only` style to `RSpec/NamedSubject`. ([@kuahyeow])
+- Fix `RSpec/FactoryBot/ConsistentParenthesesStyle` to ignore calls without the first positional argument. ([@pirj])
 
 ## 2.14.2 (2022-10-25)
 
--- a/lib/rubocop/cop/rspec/factory_bot/consistent_parentheses_style.rb
+++ b/lib/rubocop/cop/rspec/factory_bot/consistent_parentheses_style.rb
@@ -57,9 +57,10 @@ module RuboCop
 
           # @!method factory_call(node)
           def_node_matcher :factory_call, <<-PATTERN
-              (send
-                ${#factory_bot? nil?} %FACTORY_CALLS
-              $...)
+            (send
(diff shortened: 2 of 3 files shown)


bquorning (approved):

bquorning:
Do we need a v2.14.3, or can this wait for a v2.15.0?

pirj:
I'd say this can wait, as the reporter mentioned they don't use FactoryBot, and it would make sense for them to disable the whole department.
Let's wait if someone else complains to see if we need a patch release.

ydah:
@pirj @bquorning 

The issue that this Pull Request resolves has been reported:
- https://github.com/rubocop/rubocop-rspec/issues/1458

Issue #1360, opened 2022-08-15
Feature Request: `IgnoredGroups` configuration in `RSpec/NestedGroups`


jtannas:
TL;DR: Please add `IgnoredGroups` configuration to `RSpec/NestedGroups` so that we can stop configured groups from contributing to the nesting level.

### Context

We use the [`RSwag`](https://github.com/rswag/rswag) gem in our Rails request specs. A hiccup we're running into is it uses a lot of custom example groups for boilerplate descriptions of the API format and it's causing problems with the `RSpec/NestedGroups` cop.

eg.
```ruby
RSpec.describe '/users', type: :request do # <-- Nesting group 1
  path '/users' do # <-- Nesting group 2
    post 'creates a new user' do # <-- Nesting group 3
      response '201', 'Created' do # <-- Nesting group 4
        # tests go here
      end

      response '401', 'Unauthorized' do
        # tests go here
      end
      # etc...
```

We're stuck with either setting a very high level of allowed nesting or disabling the cop.

### Requested Solution
The feature I'd like is to tell Rubocop not to count configured example groups against the nesting level of the cop

```yml
RSpec/NestedGroups:
  IgnoreGroups:
    - path
    - get
    - post
    - delete
    - put
    - response
```
```ruby
RSpec.describe '/users', type: :request do # <-- Cop nesting level += 1
  path '/users' do # <-- Cop nesting level += 0
    post 'creates a new user' do # <-- Cop nesting level += 0
      response '201', 'Created' do # <-- Cop nesting level += 0
        # tests go here
      end
```

Thanks for any feedback and help you can provide :bow: 

---
Relevant Cop Docs: https://www.rubydoc.info/gems/rubocop-rspec/1.7.0/RuboCop/Cop/RSpec/NestedGroups
Relevant Cop Code: https://github.com/rubocop/rubocop-rspec/blob/master/lib/rubocop/cop/rspec/nested_groups.rb

PR #1453, opened 2022-10-31
Fix nested groups allowed groups

samrjenkins:
I discovered this bug while trying to configure `RSpec/NestedGroups` with `AllowedGroups: ['describe']`. The cop was not ignoring the `describe` nesting levels.

After some investigation two things became apparent:
### First 
The example provided in this issue (https://github.com/rubocop/rubocop-rspec/issues/1360) does not seem to log an offence. In the first commit of this PR I remove the cop_config from the `AllowsGroups` tests and they still pass. This is because rubocop-rspec does not consider
```rb
path do
  # some examples
end
```
to be an RSpec "example group". As a result, `RSpec/NestedGroups` does not count `path` groups when calculating the level of nesting.

This cop only needs to be configured to ignore RSpec groups which we recognise as example groups (`describe`, `context`, `feature`, `example_group`). Other method names (eg. `path`) will not contribute towards the nesting count

### Second
The reason I could not configure the cop to allow `describe` groups was because the cop is implemented to compared the `AllowedGroups` values (strings) with the AST node's method name (symbol). The cop cannot find the `:describe` symbol in the `AllowedGroups` array of strings `['describe']`.

This PR fixes the second of these issues. The first seems to be a non-issue

- [x] Added tests.
- [x] Updated documentation.


pirj:
> The example provided in this issue (https://github.com/rubocop/rubocop-rspec/issues/1360) does not seem to log an offence. In the first commit of this PR I remove the `cop_config` from the `AllowsGroups` tests and they still pass

Nice catch. I believe it's sufficient to adjust the test to add `path` to example groups DSL:
```
      before do
        other_cops.dig('RSpec', 'Language', 'ExampleGroups', 'Regular')
          .push('path')
      end
```

Would you add this small fix to this PR?

pirj (approved):
Fantastic, thank you!

CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 
 - Add `named_only` style to `RSpec/NamedSubject`. ([@kuahyeow])
 - Fix `RSpec/FactoryBot/ConsistentParenthesesStyle` to ignore calls without the first positional argument. ([@pirj])
+- Fix `RSpec/NestedGroups` to correctly use `AllowedGroups` config. ([@samrjenkins])
```
pirj, line 7: To fix the Markdown check, you have to add yourself in a section listing contributors at the end of this file, maintaining the alphabetic order.

bquorning: This used to be pointed out by one of our changelog specs, but it seems to be not working since we change the Markdown format recently. I’ll go fix it right away.

samrjenkins: Ah! Thanks 🚀. I have addresses

pirj:
> what you think the advantage is of your approach

I prefer fixes that don’t need to change existing specs.
If the spec was deceiving like this one, let’s fix it, make it red and then fix the code to make it green.

And after all, this config option was introduced to support rswag-rspec with its ‘path’ alias.

I would generally not recommend adding built-in example group aliases to this ignore, as this undermines the very purpose of the cop. If the default level of nesting is not sufficient for your style - increase ‘Max’.

ydah:
Thank you for fixing it. We think it would be good if you could squash the commits.

lib/rubocop/cop/rspec/nested_groups.rb
```diff
@@ -83,9 +83,9 @@ module RSpec
       #     end
       #   end
       #
-      # @example `AllowedGroups: [path]`
-      #   describe Foo do # <-- nested groups 1
-      #     path '/foo' do # <-- nested groups 1 (not counted)
+      # @example `AllowedGroups: [describe]`
+      #   describe Foo do # <-- nested groups 0 (not counted)
+      #     context 'foo' do # <-- nested groups 1
```
pirj, line 88:
```suggestion
      # @example `AllowedGroups: [path]`
      #   describe Foo do # <-- nested groups 1
      #     path '/foo' do # <-- nested groups 1 (not counted)
```

docs/modules/ROOT/pages/cops_rspec.adoc
```diff
@@ -3531,12 +3531,12 @@ describe Foo do # <-- nested groups 1
 end
 ----
 
-==== `AllowedGroups: [path]`
+==== `AllowedGroups: [describe]`
```
pirj, line 3534:
```suggestion
==== `AllowedGroups: [path]`
```

docs/modules/ROOT/pages/cops_rspec.adoc
```diff
@@ -3531,12 +3531,12 @@ describe Foo do # <-- nested groups 1
...
 
 [source,ruby]
 ----
-describe Foo do # <-- nested groups 1
-  path '/foo' do # <-- nested groups 1 (not counted)
+describe Foo do # <-- nested groups 0 (not counted)
+  context 'foo' do # <-- nested groups 1
```
pirj, line 3539:
```suggestion
describe Foo do # <-- nested groups 1
  path '/foo' do # <-- nested groups 1 (not counted)
```

Darhazer (approved):

pirj:
Thank you, @samrjenkins !

PR #1455, opened 2022-10-31
Fix broken changelog spec

bquorning:
The changelog spec was no longer checking the "implicit links" after we change Markdown format a few weeks back (`implicit_link_names` was empty).


pirj (approved):
Thank you!

PR #1467, opened 2022-11-06
Check same factory calls in array on `RSpec/FactoryBot/CreateList`

r7kamura:
It would be nice to have a feature in `RSpec/FactoryBot/CreateList` that does the following correction. What do you think?

```ruby
# bad
[create(:user)]

# good
create_list(:user, 1)

# bad
[create(:user), create(:user)]

# good
create_list(:user, 2)
```

FYI: The real example in my project that makes me feel to add this feature is as follows:

```ruby
RSpec.describe 'GET /categories' do
  let!(:categories) do
    [
      create(:category),
      create(:category)
    ]
  end

  it 'returns categories' do
    get '/categories'
    expect(response).to have_http_status(200)
    expect(response.parsed_body).to match(
      'categories' => categories.map do |category|
        hash_including('id' => category.id)
      end
    )
  end
end
```

- [x] Added tests.
- [x] Updated documentation.

### DIFF
--- a/docs/modules/ROOT/pages/cops_rspec_factorybot.adoc
+++ b/docs/modules/ROOT/pages/cops_rspec_factorybot.adoc
@@ -146,6 +146,7 @@ This cop can be configured using the `EnforcedStyle` option
 ----
 # bad
 3.times { create :user }
+[create(:user), create(:user), create(:user)]
 
 # good
 create_list :user, 3
@@ -166,6 +167,7 @@ create_list :user, 3
 ----
 # bad
 create_list :user, 3
+[create(:user), create(:user), create(:user)]
 
 # good
 3.times { create :user }
--- a/spec/rubocop/cop/rspec/factory_bot/create_list_spec.rb
+++ b/spec/rubocop/cop/rspec/factory_bot/create_list_spec.rb
@@ -189,6 +189,61 @@ RSpec.describe RuboCop::Cop::RSpec::FactoryBot::CreateList do
         create_list(:user, 3) { |user| create(:account, user: user) }
       RUBY
     end
+
+    context 'with empty array' do
+      it 'does not register an offense' do
+        expect_no_offenses(<<~RUBY)
+          []
+        RUBY
+      end
+    end
+
+    context 'with different `create` nodes in array' do
+      it 'does not register an offense' do
+        expect_no_offenses(<<~RUBY)
+          [create(:user), create(:user, age: 18)]
+        RUBY
+      end
+    end
+
+    context 'with one `create` node in array' do
+      it 'registers and corrects an offense' do
+        expect_offense(<<~RUBY)
+          [create(:user)]
+          ^^^^^^^^^^^^^^^ Prefer create_list.
+        RUBY
+
+        expect_correction(<<~RUBY)
+          create_list(:user, 1)
+        RUBY
+      end
+    end
+
+    context 'with same `create` nodes in array' do
+      it 'registers and corrects an offense' do
+        expect_offense(<<~RUBY)
+          [create(:user), create(:user)]
+          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ Prefer create_list.
+        RUBY
+
+        expect_correction(<<~RUBY)
+          create_list(:user, 2)
+        RUBY
+      end
+    end
+
+    context 'with same `create` nodes in array with method calls' do
+      it 'registers and corrects an offense' do
+        expect_offense(<<~RUBY)
+          [create(:user, point: rand), create(:user, point: rand)]
+          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ Prefer 2.times.
+        RUBY
+
+        expect_correction(<<~RUBY)
+          2.times { create(:user, point: rand) }
+        RUBY
+      end
+    end
   end
 
   context 'when EnforcedStyle is :n_times' do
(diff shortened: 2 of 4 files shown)


r7kamura:
This is a bit off topic, but since this project does not have the Gemfile.lock under Git control, I see that sometimes older versions of the rubocop gem are used, in which case the `bundle exec rake` may fail by the recent docs format change. I was a little confused by this issue.

pirj:
> sometimes older versions of the rubocop gem are used

Yes, this gets in the way sometimes with various symptoms. What do you think would be a suitable solution to this?

spec/rubocop/cop/rspec/factory_bot/create_list_spec.rb
```diff
@@ -189,6 +189,48 @@
...
+        expect_correction(<<~RUBY)
+          create_list(:user, 1)
+        RUBY
+      end
+    end
+
+    context 'with same `create` nodes in array' do
+      it 'registers and corrects an offense' do
+        expect_offense(<<~RUBY)
+          [create(:user), create(:user)]
```
pirj, line 225: I'm curious how the cop would behave with more complicated cases, e.g.

```ruby
[create(:user, age: 18), create(:user, age: 18)]
```
and especially with
```ruby
[create(:user, age: rand(18..32)), create(:user, age: rand(18..32))]
```

r7kamura:
> What do you think would be a suitable solution to this?

I think it's better to put Gemfile.lock under Git's control. This is because this practice is the default for the templates generated by the `bundle gem` and Bundler recommends doing so as well.

- https://github.com/rubygems/bundler/pull/5822
- https://github.com/rubygems/rubygems/issues/3372
- https://bundler.io/v1.15/guides/creating_gem.html#testing-our-gem

spec/rubocop/cop/rspec/factory_bot/create_list_spec.rb
r7kamura: Good point. I'll add this kind of test cases. Actually, this will always be replaced with n_times style, and it will not be replaced with create_list style because it contains a method call.

```ruby
2.times { create(:user, age: rand(18..32) }
```

The following points must be considered in this case:

1. Should it really be replaced when `EnforcedStyle: create_list`?
    - I believe it would be good to replace in this case. This is fine as currently implemented.
2. What message should be displayed in the offense?
    - The current implementation simply shows a message corresponding to the EnforcedStyle, which is at odds with the final autocorrection result, so it would be better to calculate whether the method call is included inside and display the appropriate message.

docs/modules/ROOT/pages/cops_rspec_factorybot.adoc
```diff
@@ -151,6 +151,7 @@ This cop can be configured using the `EnforcedStyle` option
 ----
 # bad
 3.times { create :user }
+[create(:user), create(:user), create(:user)]
```
pirj, line 154: I wonder if this style is acceptable in factories, though, as [FactoryBot lists this as an example](https://github.com/thoughtbot/factory_bot/blob/master/GETTING_STARTED.md#has_and_belongs_to_many-associations):
```ruby
factory :profile_with_languages do
  transient do
    languages_count { 2 }
  end

  languages do
    Array.new(languages_count) do
      association(:language, profiles: [instance])
    end
  end
end
```
And we [inspect factory definitions, too](https://github.com/rubocop/rubocop-rspec/blob/a0ef995bb02f604fbe45a56973046d21cb75c7b8/config/default.yml#L938).

`Array.new(count) { ... }` is more like `count.times.map { ... }`.

docs/modules/ROOT/pages/cops_rspec_factorybot.adoc
```diff
@@ -171,6 +172,7 @@ create_list :user, 3
 ----
 # bad
 create_list :user, 3
+[create(:user), create(:user), create(:user)]
 
 # good
 3.times { create :user }
```
pirj, line 178: Now when I look at this, it appears that `3.times { ... }` is not a valid replacement for e.g.:
```diff
- users = create_list(:user, 3)
+ users = 3.times { create(:user) }
```
but `user = 3.times.map { create(:user) }` is.

What do you think?

r7kamura: Oh, I had not paid attention to this before, but you are right.

Even without this Pull Request, it might be better to always use `times.map` instead for safety when replacing with n_times style. In that case, it would be better to have a rule that `3.times.map { create(:user) }` is replaced with `create_list(:user, 3)` on create_list style side.

r7kamura: Even if we change so as in the Pull Request above, the autocorrection from `3.times { create(:user) }` to `create_list(:user, 3)` changes the return value as you commented, so we may want to mark this cop as `SafeAutoCorrect: false`.

pirj: Fair enough.

For `n_times` style, those cases are unsafe to correct ([origin](https://github.com/rubocop/rubocop-rspec/pull/1469/files#r1016897969)):
If it's inside a let:
```ruby
let(:three_models) { 3.times { create(...) }
```
or is assigned somewhere:
```ruby
three_models = 3.times { create(...) }
```
or used as an argument:
```ruby
create(:user, articles: 3.times { create(:book) })
```
`.map` is certainly essential.

There's an uncertainty for the [dynamic `SafeAutoCorrect` in RuboCop](https://github.com/rubocop/rubocop/issues/11092).

Wondering if your other PR with adding `.map` could reliably fix those cases.

pirj:

r7kamura:
Recreated on rubocop-factory_bot repo:

- https://github.com/rubocop/rubocop-factory_bot/pull/31

Issue #1216, opened 2021-11-26
How to extend `RSpec/FactoryBot/CreateList` ?


leoarnold:
FactoryBot actually has a few more `list` methods than just `create_list`:

```ruby
3.0.2 :006 > pp FactoryBot::Syntax::Methods.instance_methods.sort
[:attributes_for,
 :attributes_for_list,
 :attributes_for_pair,
 :build,
 :build_list,
 :build_pair,
 :build_stubbed,
 :build_stubbed_list,
 :build_stubbed_pair,
 :create,
 :create_list,
 :create_pair,
 :generate,
 :generate_list,
 :null,
 :null_list,
 :null_pair]
```

I would like to add these, but then the cop name `CreateList` does not fit any more. Maybe `MultipleInstances` would be more fitting.

Can I do that? How do I inform RuboCop about a renamed cop?

pirj:
If you consider a pair to be a list of two records, it's not necessary to rename the cop.

> How do I inform RuboCop about a renamed cop?

Example: https://github.com/rubocop/rubocop/blob/master/config/obsoletion.yml

The enforced style is `create_list`. Unless you plan to also extend the cop to suggest/correct to `_pair`, this can be left intact, too.

In any case - a PR is welcome!

leoarnold:
@pirj Thanks. PR to follow.

Darhazer:
The `CreateList` checks for `n.times { }` usage. `build_list` was ignored on purpose, as nobody would really do `n.times { build(...) }` since this code has no side effect. Same got for `attributes_for` and `build_stubbed`. So if you are to extend the cop for other constructs, I guess it should be extended also for other patterns (like methods being used inside `.each` / `.map` block)

pirj:
Ping

PR #1469, opened 2022-11-07
Change `RSpec/FactoryBot/CreateList` so that it prefers `times.map` to `times`

r7kamura:
In response to the following discussion,

- https://github.com/rubocop/rubocop-rspec/pull/1467#discussion_r1015036219

I will change the autocorrection in `n_times` style to use `times.map` instead of `times`, and in addition, change `times.map` to be the target of detection in `create_list` style.

- [x] Added tests.
- [x] Updated documentation.


spec/rubocop/cop/rspec/factory_bot/create_list_spec.rb
```diff
@@ -197,55 +208,55 @@
...
     it 'flags usage of FactoryGirl.create_list with a block' do
       expect_offense(<<~RUBY)
         FactoryGirl.create_list(:user, 3) { |user| user.points = rand(1000) }
-                    ^^^^^^^^^^^ Prefer 3.times.
+                    ^^^^^^^^^^^ Prefer 3.times.map.
       RUBY
 
       expect_correction(<<~RUBY)
-        3.times { FactoryGirl.create(:user) } { |user| user.points = rand(1000) }
+        3.times.map { FactoryGirl.create(:user) } { |user| user.points = rand(1000) }
```
Darhazer, line 259: Is this even valid? Perhaps you've uncovered a bug in the original autocorrection :sweat_smile:

docs/modules/ROOT/pages/cops_rspec_factorybot.adoc
```diff
@@ -173,7 +175,7 @@ create_list :user, 3
 create_list :user, 3
 
 # good
-3.times { create :user }
+3.times.map { create :user }
```
pirj, line 178: I'm not entirely convinced we should always recommend `.map`.
If the result is a standalone statement, `3.times { create(...) }` is fine.
But if it's inside a `let`:
```ruby
let(:three_models) { 3.times { create(...) }
```
or is assigned somewhere:
```ruby
three_models = 3.times { create(...) }
```
or used as an argument:
```ruby
create(:user, articles: 3.times { create(:book) })
```
`.map` is certainly essential.

In other cases - probably not.

spec/rubocop/cop/rspec/factory_bot/create_list_spec.rb
r7kamura: Oh, thanks, I hadn't noticed that until you pointed it out. I created another Pull Request for this 👍 

- https://github.com/rubocop/rubocop-rspec/pull/1472

docs/modules/ROOT/pages/cops_rspec_factorybot.adoc
r7kamura: Yes, there will be cases where `.map` is not needed, but that seems something that must be given up for safer replacement.

I suppose I could add a conditional branch to reduce someunnecessary `.map` like `Lint/Void` Cop does:

1. `times` if it is on void context
2. `times.map` otherwise

However, it seems a bit too complicated, and it still does not work well for some common cases, such as the following:

```ruby
context 'when there are 3 users' do
  before do
    create_list(:user, 3)
  end

  # ...
end
```

```ruby
context 'when there are 3 users' do
  before do
    3.times.map { create(:user) }
  end

  # ...
end
```

Strictly speaking, the opposite pattern in create_list style can also cause problems:

```ruby
# before
count = 3.times { create(:user) }
puts "Created #{count} users"

# after (wrong correction)
count = create_list(:user, 3)
puts "Created #{count} users"
```

r7kamura:
The change to always use `.map` would also affect what is pointed out in the comment below (about the relationship between `build_list` and `times`):

- https://github.com/rubocop/rubocop-rspec/issues/1216#issuecomment-980105233

pirj:

PR #1470, opened 2022-11-08
Improved processing speed for `RSpec/Be`, `RSpec/ExpectActual`, `RSpec/ImplicitExpect`, `RSpec/MessageSpies`, `RSpec/PredicateMatcher`, `RSpec/StubbedMock` and `RSpec/Rails/HaveHttpStatus`

ydah:
Defined `RESTRICT_ON_SEND` for cop where `on_send` is defined but `RESTRICT_ON_SEND` is not defined.
For those not defined, the following Custom Cop was used to check.

- https://github.com/ydah/workitcop/blob/main/lib/rubocop/cop/workit/restrict_on_send.rb

- [-] Added tests.
- [-] Updated documentation.

### DIFF
--- a/lib/rubocop/cop/rspec/rails/have_http_status.rb
+++ b/lib/rubocop/cop/rspec/rails/have_http_status.rb
@@ -20,6 +20,8 @@ module RuboCop
             'Prefer `expect(response).%<to>s have_http_status(%<status>i)` ' \
             'over `expect(response.status).%<to>s %<match>s`.'
 
+          RESTRICT_ON_SEND = Runners.all
+
           # @!method match_status(node)
           def_node_matcher :match_status, <<-PATTERN
             (send
--- a/lib/rubocop/rspec/language.rb
+++ b/lib/rubocop/rspec/language.rb
@@ -160,8 +160,12 @@ module RuboCop
 
       module Runners # :nodoc:
         ALL = %i[to to_not not_to].freeze
-        def self.all(element)
-          ALL.include?(element)
+        class << self
+          def all(element = nil)
+            return ALL if element.nil?
+
+            ALL.include?(element)
+          end
         end
       end
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## Master (Unreleased)
 
 - Add new `RSpec/FactoryBot/FactoryNameStyle` cop. ([@ydah])
+- Improved processing speed for `RSpec/Be`, `RSpec/ExpectActual`, `RSpec/ImplicitExpect`, `RSpec/MessageSpies`, `RSpec/PredicateMatcher` and `RSpec/Rails/HaveHttpStatus`. ([@ydah])
 - Fix wrong autocorrection in `n_times` style on `RSpec/FactoryBot/CreateList`. ([@r7kamura])
 
 ## 2.15.0 (2022-11-03)
--- a/lib/rubocop/cop/rspec/be.rb
+++ b/lib/rubocop/cop/rspec/be.rb
@@ -21,6 +21,8 @@ module RuboCop
       class Be < Base
         MSG = "Don't use `be` without an argument."
 
+        RESTRICT_ON_SEND = Runners.all
+
         # @!method be_without_args(node)
         def_node_matcher :be_without_args, <<-PATTERN
           (send _ #Runners.all $(send nil? :be))
--- a/lib/rubocop/cop/rspec/expect_actual.rb
+++ b/lib/rubocop/cop/rspec/expect_actual.rb
@@ -26,6 +26,8 @@ module RuboCop
 
         MSG = 'Provide the actual you are testing to `expect(...)`.'
 
+        RESTRICT_ON_SEND = Runners.all
+
         SIMPLE_LITERALS = %i[
           true
           false
--- a/lib/rubocop/cop/rspec/implicit_expect.rb
+++ b/lib/rubocop/cop/rspec/implicit_expect.rb
@@ -28,6 +28,8 @@ module RuboCop
 
         MSG = 'Prefer `%<good>s` over `%<bad>s`.'
 
+        RESTRICT_ON_SEND = Runners.all + %i[should should_not]
+
         # @!method implicit_expect(node)
         def_node_matcher :implicit_expect, <<-PATTERN
           {
--- a/lib/rubocop/cop/rspec/message_spies.rb
+++ b/lib/rubocop/cop/rspec/message_spies.rb
@@ -41,6 +41,8 @@ module RuboCop
 
         SUPPORTED_STYLES = %w[have_received receive].freeze
 
+        RESTRICT_ON_SEND = Runners.all
+
         # @!method message_expectation(node)
         def_node_matcher :message_expectation, %(
           (send (send nil? :expect $_) #Runners.all ...)
--- a/lib/rubocop/cop/rspec/predicate_matcher.rb
+++ b/lib/rubocop/cop/rspec/predicate_matcher.rb
@@ -283,6 +283,8 @@ module RuboCop
         include InflectedHelper
         include ExplicitHelper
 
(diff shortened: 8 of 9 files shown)


ydah:
This time We created the `to_a` method in `RuboCop::RSpec::Language::Runners`, but eventually I would like to create it in other modules as well.

However, what we need to consider here is that it changes dynamically depending on config as follows.

https://github.com/rubocop/rubocop-rspec/blob/3e499a73e892ea786be81eb9b39f71e68e3d5874/lib/rubocop/rspec/language.rb#L119-L123

If you simply implement and configure `to_a` as follows, the `Language.config` will be nil.

```ruby
# lib/rubocop/rspec/language.rb
module Expectations
  def self.to_a
    Language.config['Expectations']
  end
end

# lib/rubocop/cop/foo.rb
RESTRICT_ON_SEND = Expectations.to_a
```

lib/rubocop/rspec/language.rb
```diff
@@ -159,9 +159,14 @@ def context(element)
       end
 
       module Runners # :nodoc:
-        ALL = %i[to to_not not_to].freeze
-        def self.all(element)
-          ALL.include?(element)
+        class << self
+          def to_a
+            %i[to to_not not_to].freeze
+          end
```
pirj, line 165:
```suggestion
        ALL = %i[to to_not not_to].freeze
        class << self
          def to_a
            ALL
          end
```

This was a constant initially to save a penny on creating a 3-element array each time we access it.

lib/rubocop/cop/rspec/rails/have_http_status.rb
```diff
@@ -20,6 +20,8 @@ class HaveHttpStatus < Base
             'Prefer `expect(response).%<to>s have_http_status(%<status>i)` ' \
             'over `expect(response.status).%<to>s %<match>s`.'
 
+          RESTRICT_ON_SEND = Runners.to_a
```
pirj, line 23: `.to_a` looks weird, as `RESTRICT_ON_SEND` can be assigned a `Set`.

I don't have an idea for a better name.
`ALL` is a private API, implementation detail. And using `ALL` would make it inconsistent with other cops that would use non-static DSL configuraton  e.g. `Language.config['Expectations']` as you say above.

pirj (approved):
Thank you!

Nice cop :heart: Do you plan to submit it to RuboCop's internal investigations department?

lib/rubocop/rspec/language.rb
```diff
@@ -159,9 +159,14 @@ def context(element)
...
+
+          def all(element)
+            to_a.include?(element)
+          end
```
Darhazer, line 169: This will create a new array each time. Why not just defining `to_a = ALL`

lib/rubocop/cop/rspec/rails/have_http_status.rb
ydah: This may not be the best, but how about `configurations`?

lib/rubocop/rspec/language.rb
ydah: Indeed it is. I updated it. Thank you so much.

ydah:
> Nice cop ❤️ Do you plan to submit it to RuboCop's internal investigations department?

I missed it, sorry for the late reply.
I suggested this to rubocop/rubocop#10783 but it is now closed.
The reason is that there was a judgment that there may be more noise than detection efficiency.

Darhazer (approved):

pirj:
A heads-up:
```ruby
RESTRICT_ON_SEND = Expectations.to_a
```
won't work as expected, as `RESTRICT_ON_SEND` will only be initialized once with the return value of `.to_a`.

A way around this is to override [this method](https://github.com/rubocop/rubocop/blob/8cc65d1c693d72afdc4f6b19cf1685d1ea9f3002/lib/rubocop/cop/base.rb#L308):

```ruby
private_class_method def self.restrict_on_send
```

so that it fetches the config dynamically.

lib/rubocop/cop/rspec/rails/have_http_status.rb
pirj: Maybe `each`?

It's possible to double purpose `.all`:
```ruby
def self.all(element = NO_ARGUMENTS)
  return ALL if element == NO_ARGUMENTS
  ALL.include?(element)
end
```

I have no strong preference.

ydah:
Thank you for review. I updated this PR.
The following cop has also been optimized

```
lib/rubocop/cop/rspec/dialect.rb:45:7: C: Workit/RestrictOnSend: Consider defined RESTRICT_ON_SEND for optimization.
      class Dialect < Base ...
      ^^^^^^^^^^^^^^^^^^^^
lib/rubocop/cop/rspec/excessive_docstring_spacing.rb:26:7: C: Workit/RestrictOnSend: Consider defined RESTRICT_ON_SEND for optimization.
      class ExcessiveDocstringSpacing < Base ...
      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
lib/rubocop/cop/rspec/focus.rb:40:7: C: Workit/RestrictOnSend: Consider defined RESTRICT_ON_SEND for optimization.
      class Focus < Base ...
      ^^^^^^^^^^^^^^^^^^
lib/rubocop/cop/rspec/pending.rb:35:7: C: Workit/RestrictOnSend: Consider defined RESTRICT_ON_SEND for optimization.
      class Pending < Base ...
      ^^^^^^^^^^^^^^^^^^^^
lib/rubocop/cop/rspec/shared_examples.rb:23:7: C: Workit/RestrictOnSend: Consider defined RESTRICT_ON_SEND for optimization.
      class SharedExamples < Base ...
      ^^^^^^^^^^^^^^^^^^^^^^^^^^^
lib/rubocop/cop/rspec/stubbed_mock.rb:16:7: C: Workit/RestrictOnSend: Consider defined RESTRICT_ON_SEND for optimization.
      class StubbedMock < Base ...
      ^^^^^^^^^^^^^^^^^^^^^^^^
lib/rubocop/cop/rspec/subject_declaration.rb:22:7: C: Workit/RestrictOnSend: Consider defined RESTRICT_ON_SEND for optimization.
      class SubjectDeclaration < Base ...
lib/rubocop/cop/rspec/variable_definition.rb:26:7: C: Workit/RestrictOnSend: Consider defined RESTRICT_ON_SEND for optimization.
      class VariableDefinition < Base ...
lib/rubocop/cop/rspec/variable_name.rb:41:7: C: Workit/RestrictOnSend: Consider defined RESTRICT_ON_SEND for optimization.
      class VariableName < Base ...
      ^^^^^^^^^^^^^^^^^^^^^^^^^
```

lib/rubocop/cop/rspec/dialect.rb
```diff
@@ -51,6 +51,12 @@ class Dialect < Base
         # @!method rspec_method?(node)
         def_node_matcher :rspec_method?, send_pattern('#ALL.all')
 
+        private_class_method def self.restrict_on_send
+          return [] unless RuboCop::RSpec::Language.config
+
```
pirj, line 56: This is where it gets really interesting.

`restrict_on_send` implies that it would return a static value, while with dynamic, per-directory RSpec DSL conf, it won't.

`rubocop` runs cops in parallel, one per processor core.

Two issues here, one is somewhat suggested by the need fir this ﻿check

1 Is RuboCop caching this kist [here]https://github.com/rubocop/rubocop/blob/bc4ae3ef798c8db4f7427369777deb69be8f4d63/lib/rubocop/cop/commissioner.rb#L145)? If yes, supplying it dynamically is pointless

2 Is it safe to cache it on the class level in a multi-thread run?

ydah: All of these were perspectives I had not considered, and I appreciate them very much. Thank you very much.
In this Pull Request, let's improve where we can optimize by setting `RESTRICT_ON_SEND`!

pirj:
Thank you, @ydah !

PR #1473, opened 2022-11-09
Declare autocorrect as unsafe for `RSpec/Capybara/SpecificFinders`

### LINKED ISSUE #1468: `RSpec/Capybara/SpecificFinders` autocorrection shoud remove escape
### Reproducion steps
Run `RSpec/Capybara/SpecificFinders` autocorrection on:
```ruby
find('#domain-deadbeef\.org-delete')
```
This returns a tag `<tag id="domain-deadbeef.org-delete">`, which id contains `.`.
The `.` shoud be escaped because bare `.` is a class selector in CSS.

### Expected result
```ruby
find_by_id('domain-deadbeef.org-delete')
```
CSS selector escape should be removed.

### Actual result
```ruby
find_by_id('domain-deadbeef\.org-delete')
```

### References
Per HTML Standard, `id` attribute can contain any character except whitespace.
https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/id

Tietew:
Fixes #1468

- [ ] Added tests.
- [x] Updated documentation.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Master (Unreleased)
 
+- Declare autocorrect as unsafe for `RSpec/Capybara/SpecificFinders`. ([@Tietew])
+
 ## 2.15.0 (2022-11-03)
 
 - Fix a false positive for `RSpec/RepeatedDescription` when different its block expectations are used. ([@ydah])
--- a/config/default.yml
+++ b/config/default.yml
@@ -890,7 +890,9 @@ RSpec/Capybara/SpecificActions:
 RSpec/Capybara/SpecificFinders:
   Description: Checks if there is a more specific finder offered by Capybara.
   Enabled: pending
+  SafeAutoCorrect: false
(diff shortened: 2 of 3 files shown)


pirj:
Is it in the cop’s design that the autocorrection is inherently unsafe, and it is impossible to fix the cop?

Or are there deficiencies due to some edge cases that we’ve missed previously, but are well aware of presently?

Tietew:
@pirj Thanks for reply.
See my comments on #1468 for edge cases. It may be possible to fix, but not easy.

ydah:
I am planning to work on edge case support. What do you think?
We do not consider this cop is Unsafe, as it may be difficult to accommodate, but not impossible.

pirj:
We’re not practicing marking cops that are not unsafe by design as such. If a cop has bugs, or uncovered edge cases, we encourage affected users to exclude problematic files from inspection using .rubocop_todo.unless or by local rubocop:disable statements.

Thank you for understanding.

PR #1476, opened 2022-11-09
Proof of concept separation of departments

pirj:
See [discussion](https://github.com/rubocop/rubocop-rspec/discussions/1440)

🔎 better viewed [without indentation changes](https://github.com/rubocop/rubocop-rspec/pull/1476/files?diff=split&w=1)

I went too far, as renaming cops is a breaking change:
```
.rubocop_todo.yml: RSpec/FactoryBot/ConsistentParenthesesStyle has the wrong namespace - should be FactoryBot
.rubocop_todo.yml: RSpec/FactoryBot/CreateList has the wrong namespace - should be FactoryBot
.rubocop_todo.yml: RSpec/FactoryBot/FactoryClassName has the wrong namespace - should be FactoryBot
Error: Ambiguous cop name `RSpec/Rails/HttpStatus` used in /Users/pirj/source/test/.rubocop_todo.yml needs department qualifier. Did you mean Rails/HttpStatus or RSpecRails/HttpStatus?
```
This way it would be a breaking change. And we could only afford this in 3.0.
But then, when we'll decide to actually extract them, we'd have to release 4.0.
And when RuboCop 2.0 is finally released, and if we decide not to sync our 3.0 and 2.0 with it, we'll have to release 5.0. Just because they will enable their pending cops.

What do you think of such a series of major updates?

### Not done

1. `RuboCop::RSpec::FactoryBot::Language` is used by both the extracted (`lib/rubocop/cop/factory_bot/create_list.rb`, `lib/rubocop/cop/factory_bot/consistent_parentheses_style.rb`) and the remaining `lib/rubocop/cop/rspec/factory_bot/syntax_methods.rb` cops.

2. Too lazy to fix one last spec

- [+] Feature branch is up-to-date with `master` (if not - rebase it).
- [-] Squashed related commits together.
- [-] Added tests.
- [-] Updated documentation.
- [+/-] The build (`bundle exec rake`) passes (be sure to run this locally, since it may produce updated documentation that you will need to commit).


config/obsoletion.yml
```diff
@@ -12,3 +12,20 @@ changed_parameters:
     parameters: IgnoredPatterns
     alternative: AllowedPatterns
     severity: warning
+
+renamed:
```
pirj, line 16: [Explanation why this is a breaking change](https://github.com/rubocop/rubocop/pull/10725#issuecomment-1163565021)

lib/rubocop-rspec.rb
```diff
@@ -37,7 +35,10 @@
```
pirj, line 35:
```
require_relative 'rubocop/rspec/factory_bot'
```
above is needed by both the extracted (RSpec-agnostic) and the remaining (RSpec-specific) FactoryBot cops.

lib/rubocop/rspec/config_formatter.rb
```diff
@@ -47,7 +48,7 @@ def replace_nil(config)
       end
 
       def reference(cop)
-        COP_DOC_BASE_URL + cop.sub('RSpec/', '')
```
pirj, line 50: Any idea why this was needed?

lib/rubocop/rspec/description_extractor.rb
```diff
@@ -32,10 +32,7 @@ def initialize(yardoc)
         #
         # @return [Boolean]
         def rspec_cop?
-          class_documentation? &&
```
pirj, line 35: Replaced with an argument `YARD::Registry.all(:class)`.

lib/rubocop/rspec/description_extractor.rb
```diff
@@ -32,10 +32,7 @@ def initialize(yardoc)
...
-            rspec_cop_namespace? &&
```
pirj, line 36: We only load our classes using a glob, why would we check if it's really ours?

spec/project/default_config_spec.rb
```diff
@@ -15,7 +15,8 @@
   end
 
   let(:cop_names) do
-    glob = SpecHelper::ROOT.join('lib', 'rubocop', 'cop', 'rspec',
+    glob = SpecHelper::ROOT.join('lib', 'rubocop', 'cop',
+                                 '{capybara,factory_bot,rspec,rspec-rails}',
                                  '{,capybara,factory_bot,rails}', '*.rb')
```
pirj, line 20: Not perfect. Could probably be `'{capybara,factory_bot,rspec,rspec/capybara,rspec/factory_bot,rspec/rails,rspec-rails}`, but it's not platform-independent. Do we care about other platforms (that one with `\` file separator!) running our tests?

pirj:
I'm closing this one and will make another iteration going further, with putting all the prospective gems in separate directories, with own gemspecs, lib, spec etc.

PR #1477, opened 2022-11-11
Use `module_function` for `RSpec/Capybara/SpecificActions`

ydah:
Follow up: https://github.com/rubocop/rubocop-rspec/pull/1402#discussion_r995904977

- [x] Added tests.
- [x] Updated documentation.


pirj (approved):
Thank you!
Does it make sense to replace `module_function` there now with `class << self` to prevent the use with `include`?

PR #1478, opened 2022-11-12
Add `RSpec/Rails/ResponseParsedBody` cop

r7kamura:
Since Rails 5, there is an often overlooked feature: `response.parsed_body`.

- https://github.com/rails/rails/issues/23594

As a typical pattern, many Rails applications have `JSON.parse(response.body)` in their controlelr-specs and request-specs, but most of these could be replaced by `response.parsed_body`. So it would be nice for rubocop-rails to have a Cop that recommends `response.parsed_body`.

I think there are several advantages to using `response.parsed_body`.

1. the fact that Rails provides this by default means that Rails would probably encourage the use of it
2. it avoids direct references to `JSON`
3. shorter code

What do you think?

- [x] Added tests.
- [x] Updated documentation.

- [x] Set `VersionAdded` in `default/config.yml` to the next minor version.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -144,3 +144,5 @@ RSpec/Rails/HaveHttpStatus:
   Enabled: true
 RSpec/Rails/InferredSpecType:
   Enabled: true
+RSpec/Rails/ResponseParsedBody:
+  Enabled: true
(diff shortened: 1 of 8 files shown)


pirj:
Just glanced over. Would this cop better suit rubocop-rails, as it is would also be useful for inspecting Minitest tests?

TIL ‘parsed_body’ :+1: (how could I have missed such a nicety that exists since Rails 5.0!?)

koic:
Yeah, I think this cop can be put into RuboCop Rails. Because it's not RSpec specific API, so RuboCop Rails would be a better suit.

r7kamura:
Certainly rspec-rails is more appropriate. I'll close this one and create another pull request to rspec-rails 👍

PR #1486, opened 2022-11-15
Add `RSpec/DuplicatedMetadata` cop

r7kamura:
Looking at RSpec code in some libraries and applications, I see several duplicate metadata. So I think it would be nice to have a Cop that detects such code. What do you think?

For example, the following is the result from rubocop gem:

```
$ bundle exec rubocop spec
Inspecting 646 files
.................................................................................................................................................................................................................................................................................................................................................................................C....................................................................................................................................................................................................................................................................................

Offenses:

spec/rubocop/cop/style/conditional_assignment_assign_to_condition_spec.rb:3:69: C: [Correctable] RSpec/DuplicatedMetadata: Avoid duplicated metadata.
RSpec.describe RuboCop::Cop::Style::ConditionalAssignment, :config, :config, :config do
                                                                    ^^^^^^^
spec/rubocop/cop/style/conditional_assignment_assign_to_condition_spec.rb:3:78: C: [Correctable] RSpec/DuplicatedMetadata: Avoid duplicated metadata.
RSpec.describe RuboCop::Cop::Style::ConditionalAssignment, :config, :config, :config do
                                                                             ^^^^^^^

646 files inspected, 2 offenses detected, 2 offenses autocorrectable
```

And the following is from rubocop-rails gem:

```
$ bundle exec rubocop spec
Inspecting 133 files
................CCC..................................C...............................................................................

Offenses:

spec/rubocop/cop/rails/application_job_spec.rb:3:62: C: [Correctable] RSpec/DuplicatedMetadata: Avoid duplicated metadata.
RSpec.describe RuboCop::Cop::Rails::ApplicationJob, :config, :config do
                                                             ^^^^^^^
spec/rubocop/cop/rails/application_mailer_spec.rb:3:65: C: [Correctable] RSpec/DuplicatedMetadata: Avoid duplicated metadata.
RSpec.describe RuboCop::Cop::Rails::ApplicationMailer, :config, :config do
                                                                ^^^^^^^
spec/rubocop/cop/rails/application_record_spec.rb:3:65: C: [Correctable] RSpec/DuplicatedMetadata: Avoid duplicated metadata.
RSpec.describe RuboCop::Cop::Rails::ApplicationRecord, :config, :config do
                                                                ^^^^^^^
spec/rubocop/cop/rails/http_positional_arguments_spec.rb:3:71: C: [Correctable] RSpec/DuplicatedMetadata: Avoid duplicated metadata.
RSpec.describe RuboCop::Cop::Rails::HttpPositionalArguments, :config, :config do
                                                                      ^^^^^^^

133 files inspected, 4 offenses detected, 4 offenses autocorrectable
```

- [x] Added tests.
- [x] Updated documentation.

- [x] Set `VersionAdded` in `default/config.yml` to the next minor version.


spec/rubocop/cop/rspec/duplicated_metadata_spec.rb
```diff
@@ -0,0 +1,100 @@
...
+        RSpec.configure do |configuration|
+          configuration.before(:each, :a, :a) do
+                                          ^^ Avoid duplicated metadata.
+          end
+        end
+      RUBY
+
+      expect_correction(<<~RUBY)
+        RSpec.configure do |configuration|
+          configuration.before(:each, :a) do
```
pirj, line 94: Is `:each` the default? 
Would it still work? https://github.com/rspec/rspec-core/blob/6dede8ae1b9cce6e1de4baa912e77c97b587b446/lib/rspec/core/hooks.rb#L584
```ruby
configuration.before(:a) do
```

`#Hooks.all _ ${send` expects a scope, right?

r7kamura: You are right that the default scope is `:each` (`:example` to be exact), but the code `configuration.before(:a)` is not allowed because you must explicitly pass the scope as a first argument when passing metadata by Symbol to hook methods.

- https://github.com/rspec/rspec-core/blob/6dede8ae1b9cce6e1de4baa912e77c97b587b446/lib/rspec/core/hooks.rb#L591-L603

pirj: Fair enough 👍 
What about hash metadata, e.g.:
```ruby
configuration.before(a: true, a: true) do
```
Probably not a big deal, as another cop would tell about a duplicate hash element, and even if not, Ruby would collapse it.

pirj: I was going to make a point that this not only relates to this cop, but I've realized that this was a previously existing code, so I take this argument back.

pirj (approved):
Looks good, thank you!
Did a run on `real-world-rspec`, no errors.

Darhazer (commented):
Not a requirement but just wondering - can it also flag :config, config: true duplication?

I’ll do a full review tomorrow

r7kamura:
> Not a requirement but just wondering - can it also flag :config, config: true duplication?

It would be nice to support that case as well 👍 
It might be good to do that in this Cop, but it is also a little hard to decide which one should be removed in case of such duplicates. How about we achieve that by having a separate Cop that converts `config: true` to `:config`? (Is there already one?)

pirj:
> a separate Cop that converts config: true to :config? (Is there already one?)

There was [a PR](https://github.com/rubocop/rubocop-rspec/pull/1325) that initially implemented that, but this part disappeared after a force-push. It remains [here](https://github.com/rubocop/rubocop-rspec/compare/52c205d301bfadf6e0cfac812c3ed62dbab5ca80..dfc40a173f0617f05b10d52259acaa6404d2c660#diff-03485f6708ca5005e3c495a4227ff63883a0c6477ab4bff129ae51491b078366L187) in `lib/rubocop/cop/rspec/sort_tags.rb` around line 187.

Darhazer:
I agree that :config vs config: true should be separate cop, and that it partially will resolve the question. (`:config, config: false` will remain an edge case)
Also I was wondering whether a hash metadata should be handled, but since Ruby itself will warn about the overwritten key, I'd say it's a no-goal for a cop.

bquorning (approved):

Darhazer:
@r7kamura can you please rebase to fix the changelog entry :bow:

PR #1482, opened 2022-11-15
Fix a false positive for `RSpec/FactoryBot/ConsistentParenthesesStyle` when using `generate` with multiple arguments

ydah:
Fix: https://github.com/rubocop/rubocop-rspec/issues/1480

______________________________________________________________________

Before submitting the PR make sure the following are checked:

- [x] Feature branch is up-to-date with `master` (if not - rebase it).
- [x] Squashed related commits together.
- [x] Added tests.
- [-] Updated documentation.
- [x] Added an entry to the `CHANGELOG.md` if the new code introduces user-observable changes.
- [x] The build (`bundle exec rake`) passes (be sure to run this locally, since it may produce updated documentation that you will need to commit).

PR #1487, opened 2022-11-17
Add `ExplicitOnly` configuration option to `RSpec/FactoryBot/ConsistentParenthesesStyle`, `RSpec/FactoryBot/CreateList` and `RSpec/FactoryBot/FactoryNameStyle`

ydah:
Follow up: https://github.com/rubocop/rubocop-rspec/pull/1482#issuecomment-1315496663
Resolve: https://github.com/rubocop/rubocop-factory_bot/issues/5

- [x] Added tests.
- [x] Updated documentation.

### DIFF
--- a/config/default.yml
+++ b/config/default.yml
@@ -988,11 +988,12 @@ RSpec/FactoryBot/FactoryClassName:
 RSpec/FactoryBot/FactoryNameStyle:
   Description: Checks for name style for argument of FactoryBot::Syntax::Methods.
   Enabled: pending
-  VersionAdded: '2.16'
   EnforcedStyle: symbol
   SupportedStyles:
     - symbol
     - string
+  ExplicitOnly: false
+  VersionAdded: '2.16'
   Reference: https://www.rubydoc.info/gems/rubocop-rspec/RuboCop/Cop/RSpec/FactoryBot/FactoryNameStyle
 
 RSpec/FactoryBot/SyntaxMethods:
--- a/config/default.yml
+++ b/config/default.yml
@@ -951,7 +951,9 @@ RSpec/FactoryBot/ConsistentParenthesesStyle:
   SupportedStyles:
     - require_parentheses
     - omit_parentheses
+  ExplicitOnly: false
   VersionAdded: '2.14'
+  VersionChanged: "<<next>>"
   Reference: https://www.rubydoc.info/gems/rubocop-rspec/RuboCop/Cop/RSpec/FactoryBot/ConsistentParenthesesStyle
 
 RSpec/FactoryBot/CreateList:
--- a/config/default.yml
+++ b/config/default.yml
@@ -969,8 +969,9 @@ RSpec/FactoryBot/CreateList:
   SupportedStyles:
     - create_list
     - n_times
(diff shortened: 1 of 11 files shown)


config/default.yml
```diff
@@ -929,7 +929,9 @@ RSpec/FactoryBot/ConsistentParenthesesStyle:
   SupportedStyles:
     - require_parentheses
     - omit_parentheses
+  ExplicitOnly: false
```
Darhazer, line 932: I was thinking whether this could be applied to `RSpec/FactoryBot` (the whole department) and affect the matcher on the receiver (whether to include nil receiver or not). @pirj WDYT?

lib/rubocop/cop/rspec/factory_bot/consistent_parentheses_style.rb
```diff
@@ -40,6 +40,28 @@ module FactoryBot
         #     name: 'foo'
         #   )
         #
+        # @example `AssignmentOnly: false` (default)
```
pirj, line 43: I'm confused by the `ExplicitOnly` vs `AssignmentOnly` difference. Is this correct?

ydah: Oops, that is wrong. I corrected it. Thank you.

pirj:
We have `FactoryBot/SyntaxMethods` already that enforces `FactoryBot.create` vs just `create`, right?
Is there a situation where non-explicit and explicit would be used sparingly?
I'm hinting that those two cops together can enforce any given style combination, and this makes the option redundant.
Can you please help me understand the need for this option?

Darhazer:
Suppose you want to use FactoryBot only with explicit receiver - as you are using methods like `create` either from another gem or defined yourself. So you disable the SyntaxMethods cop. The idea is to prevent the other FactoryBot cops of raising false positives.

ydah:
My understanding is that `FactoryBot/SyntaxMethods` can explicitly force `FactoryBot` to be a receiver or not.
Therefore, methods such as `create` that do not explicitly have a receiver with `FactoryBot` can be considered as not belonging to FactoryBot, so setting this option to true will help exclude methods with the same name other than FactoryBot, which would be a false positive. We believe that this option will help to exclude methods with the same name other than FactoryBot that are False positive.

config/default.yml
ydah: Indeed it is, looks good. So we cut out `ExplicitOnly` into a common class to make it easier to handle, and added the configuration values to the other cop. WDYT?

ydah:
It was conflicted, so I rebased it to the latest master branch.

ydah:
I rebased this PR.

ydah:
This PR resolves the following Issue.
- https://github.com/rubocop/rubocop-rspec/issues/1523

ydah:

pirj:
I'd hold this off until `rubocop-factroy_bot` extraction.

ydah:
I've opened a PR in the rubocop-factory_bot repository and this one is closed
- https://github.com/rubocop/rubocop-factory_bot/pull/18

PR #1490, opened 2022-11-17
Mark `RSpec/BeEq` as `Safe: false`

r7kamura:
I tried autocorrecting `expect(value).to eq true` to `expect(value).to be true` in my Rails app and the tests result in failure. When I looked into the cause of this, I noticed that this `Value#==` has a slightly special implementation.

~~So I think it is better to mark this as an unsafe autocorrection, since it could change the test result.~~
After discussion, I thought that it is appropriate to set `Safe: false`.

What do you think?

- [x] Added tests.
- [x] Updated documentation.

    - wait... I certainly changed the existing cop's configuration, but should I really set this to the next major version?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 - Add new `RSpec/FactoryBot/FactoryNameStyle` cop. ([@ydah])
 - Fix wrong autocorrection in `n_times` style on `RSpec/FactoryBot/CreateList`. ([@r7kamura])
 - Fix a false positive for `RSpec/FactoryBot/ConsistentParenthesesStyle` when using `generate` with multiple arguments. ([@ydah])
+- Mark `RSpec/BeEq` as `Safe: false` ([@r7kamura])
 
 ## 2.15.0 (2022-11-03)
 
--- a/config/default.yml
+++ b/config/default.yml
@@ -147,7 +147,9 @@ RSpec/Be:
 RSpec/BeEq:
   Description: Check for expectations where `be(...)` can replace `eq(...)`.
   Enabled: pending
+  Safe: false
   VersionAdded: 2.9.0
+  VersionChanged: "<<next>>"
   Reference: https://www.rubydoc.info/gems/rubocop-rspec/RuboCop/Cop/RSpec/BeEq
 
 RSpec/BeEql:
--- a/docs/modules/ROOT/pages/cops_rspec.adoc
+++ b/docs/modules/ROOT/pages/cops_rspec.adoc
@@ -192,10 +192,10 @@ expect(foo).to be(true)
 | Enabled by default | Safe | Supports autocorrection | Version Added | Version Changed
 
 | Pending
-| Yes
-| Yes
+| No
+| Yes (Unsafe)
 | 2.9.0
-| -
+| <<next>>
 |===
 
 Check for expectations where `be(...)` can replace `eq(...)`.
@@ -203,6 +203,10 @@ Check for expectations where `be(...)` can replace `eq(...)`.
 The `be` matcher compares by identity while the `eq` matcher compares
 using `==`. Booleans and nil can be compared by identity and therefore
 the `be` matcher is preferable as it is a more strict test.
+
+=== Safety
+
+This cop is unsafe because it changes how values are compared.
 
 === Examples
 
--- a/lib/rubocop/cop/rspec/be_eq.rb
+++ b/lib/rubocop/cop/rspec/be_eq.rb
@@ -9,6 +9,9 @@ module RuboCop
       # using `==`. Booleans and nil can be compared by identity and therefore
       # the `be` matcher is preferable as it is a more strict test.
       #
+      # @safety
+      #   This cop is unsafe because it changes how values are compared.
+      #
       # @example
       #   # bad
       #   expect(foo).to eq(true)


pirj:
Mostly out of curiosity, can you please provide a sneak peek of what `Value#==` does, @r7kamura ?
I'm aware that [`==`](https://ruby-doc.org/core-2.7.1/BasicObject.html#method-i-3D-3D) is
> overridden in descendant classes to provide class-specific meaning

However, it's still quite interesting why an instance of `Value` would identify itself as `== true`.

pirj:
> The definition of safety is that the cop doesn’t generate false positives.

Can we say that the cop itself is not unsafe, not just its autocorrection?

Darhazer:
It's not a false positive per se (it finds the right offence), but the correction is not safe.

Darhazer (approved):

pirj:
```ruby
class Value
  def initialize(value)
    @value = value
  end

  def ==(other)
    @value.positive? == other
  end
end

Value.new(1) == true # => true
Value.new(-1) == true # => false

expect(value).to eq true
```

If the expectation is safely detected as offensive, and the autocorrection is not, what would be a safe correction in this specific case?

r7kamura:
The problem I encountered occurred with `Enumerize::Value`.

- https://github.com/brainspec/enumerize/blob/d92c9e24d7b996e95369e0ed4e5ee92770323742/lib/enumerize/value.rb#L33-L35

To simplify, my code is as follows:

```ruby
class User < ActiveRecord::Base
  extend ::Enumerize

  enumerize(
    :status,
    in: {
      active: true,
      inactive: false
    }
  )
end
```

```ruby
user = User.create!(status: 'active')
expect(user.status).to eq(true)
```

A reproducible code would look like this:

```ruby
# frozen_string_literal: true

require 'bundler/inline'

gemfile do
  source 'https://rubygems.org'

  gem 'activerecord', '>= 7'
  gem 'enumerize'
  gem 'rspec'
  gem 'sqlite3'
end

require 'active_record'
require 'enumerize'
require 'minitest/autorun'

ActiveRecord::Base.establish_connection(
  adapter: 'sqlite3',
  database: ':memory:'
)

ActiveRecord::Schema.define do
  create_table :users, force: true do |t|
    t.boolean :status, null: false
  end
end

class User < ActiveRecord::Base
  extend ::Enumerize

  enumerize(
    :status,
    in: {
      active: true,
      inactive: false
    }
  )
end

RSpec.describe User do
  describe '#status' do
    let(:user) do
      described_class.new(status: 'active')
    end

    it do
      expect(user.status).to eq(true)
    end
  end
end
```

With `eq(true)`, it succeeds as follows:

```
$ bundle exec rspec user_spec.rb
-- create_table(:users, {:force=>true})
   -> 0.0024s

User
  #status
    is expected to eq true

Finished in 0.00224 seconds (files took 1.53 seconds to load)
1 example, 0 failures

Run options: --seed 53852

# Running:



Finished in 0.001421s, 0.0000 runs/s, 0.0000 assertions/s.

0 runs, 0 assertions, 0 failures, 0 errors, 0 skips
```

And if I autocorrect this test code  to `be(true)` as this Cop says,

image

it fails as follows:

```
 bundle exec rspec user_spec.rb
-- create_table(:users, {:force=>true})
   -> 0.0025s

User
  #status
    is expected to equal true (FAILED - 1)

Failures:

  1) User#status is expected to equal true
     Failure/Error: expect(user.status).to be(true)
     
       expected true
            got #<Enumerize::Value:5800> => "active"
     # ./user_spec.rb:48:in `block (3 levels) in <top (required)>'

Finished in 0.00668 seconds (files took 1.6 seconds to load)
1 example, 1 failure

Failed examples:

rspec ./user_spec.rb:47 # User#status is expected to equal true
```

r7kamura:
> Can we say that the cop itself is not unsafe, not just its autocorrection?

I am more interested in this topic because just the other day there was some disagreement between myself and the maintainer as to whether Safe or SafeAutoCorrect should be used. (cc: @koic; sorry to bother you, but I thought you might be interested in this topic too)

- https://github.com/rubocop/rubocop-rails/pull/863#discussion_r1024697250

I suspect that this configuration is determined by the rule like:
If the purpose of this cop is to avoid **any** `eq(true)` at all times, there is nothing wrong with marking this as an offense, even if it causes the test to fail, so it would be good to set `SafeAutoCorrect: false`. On the other hand, if the purpose is to avoid **some** bad `eq(true)`, it would be good to set `Safe: false` since it can cause a mistake in judging whether it is bad or not.

Considering the magnitude of the impact, I first weakly suggested `SafeAutoCorrect: false`, but in the sense described above, now I feel that it would be better to set `Safe: false`. This Cop is not saying that all `eq(true)` is bad, but only that we should replace it whenever possible. What do you think?

pirj:
> The be matcher compares by identity while the `eq` matcher compares using `==`. Booleans and `nil` can be compared by identity and therefore the be matcher is preferable as it is a more strict test.

This seems incorrect as `equal?` and `==` are non-commutative, and we can't assume that the "actual" used in `expect` is a boolean or a `nil`. And we have no means of detecting this.

According to the [safety definition](https://docs.rubocop.org/rubocop/usage/autocorrect.html#safe-autocorrect):
> Safe (true/false) - indicates whether the cop can yield false positives (by design) or not.

from my perspective, this makes the `RSpec/BeEq` cop unsafe by design, and `RSpec/BeEql` as well.


I strongly agree with your impact note, as in a normal scenario, marking a cop is unsafe has the same effect as just removing it.

I assume that those false positives would be rare in practice, but we strive to provide safety, and in this case, I side with you to mark it (or both cops) as `Safe: false`.

r7kamura:
In response to the comments, I have reconsidered that it is appropriate for this Cop to be `Safe: false` and have updated it 👌

pirj (approved):

pirj:
Thank you, @r7kamura !

Do you agree that the same argument applies to `RSpec/BeEql`?

r7kamura:
> Do you agree that the same argument applies to `RSpec/BeEql`?

Yes. I have not investigated in detail what use-cases actually cause trouble, but I think `RSpec/BeEql` should be `Safe: false` as well.

PR #1489, opened 2022-11-17
Add new `RSpec/Rails/MinitestAssertions` cop

ydah:
Partially addresses #1485
______________________________________________________________________

Before submitting the PR make sure the following are checked:

- [x] Feature branch is up-to-date with `master` (if not - rebase it).
- [x] Squashed related commits together.
- [x] Added tests.
- [x] Updated documentation.
- [x] Added an entry to the `CHANGELOG.md` if the new code introduces user-observable changes.
- [x] The build (`bundle exec rake`) passes (be sure to run this locally, since it may produce updated documentation that you will need to commit).

If you have created a new cop:

- [x] Added the new cop to `config/default.yml`.
- [x] The cop is configured as `Enabled: pending` in `config/default.yml`.
- [x] The cop is configured as `Enabled: true` in `.rubocop.yml`.
- [x] The cop documents examples of good and bad code.
- [x] The tests assert both that bad code is reported and that good code is not reported.
- [x] Set `VersionAdded` in `default/config.yml` to the next minor version.

PR #1491, opened 2022-11-18
Change the `rake newcop` task to output `<<next>>` instead of the next minor version for config/default.yml

ydah:
Ref: https://github.com/rubocop/rubocop-rspec/pull/1489#discussion_r1025488243

- [-] Added tests.
- [-] Updated documentation.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -74,14 +74,7 @@ task :new_cop, [:cop] do |_task, args|
   generator.write_source
   generator.write_spec
   generator.inject_require(root_file_path: 'lib/rubocop/cop/rspec_cops.rb')
-  generator.inject_config(config_file_path: 'config/default.yml',
-                          version_added: bump_minor_version)
+  generator.inject_config
 
   puts generator.todo
-end
-
-def bump_minor_version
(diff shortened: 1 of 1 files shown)


Rakefile
```diff
@@ -74,14 +74,7 @@ task :new_cop, [:cop] do |_task, args|
   generator.write_source
   generator.write_spec
   generator.inject_require(root_file_path: 'lib/rubocop/cop/rspec_cops.rb')
-  generator.inject_config(config_file_path: 'config/default.yml',
-                          version_added: bump_minor_version)
```
ydah, line 78: The requirements can be satisfied by default arguments as follows:

https://github.com/rubocop/rubocop/blob/bc4ae3ef798c8db4f7427369777deb69be8f4d63/lib/rubocop/cop/generator.rb#L133-L134
```ruby
      def inject_config(config_file_path: 'config/default.yml',
                        version_added: '<<next>>')
```

pirj (approved):
Thank you!

Darhazer (approved):

bquorning (approved):

PR #1456, opened 2022-11-01
Add new `RSpec/Capybara/MatchStyle` cop

ydah:
This PR is add new cop.
This cop considers the use of methods associated with deprecated styles to be an offense.

Check the use of the following methods:

- Capybara::Node::Matchers#has_style?
  - https://rubydoc.info/github/jnicklas/capybara/master/Capybara%2FNode%2FMatchers:has_style%3F
- Capybara::Node::Matchers#assert_style
  - https://rubydoc.info/github/jnicklas/capybara/master/Capybara%2FNode%2FMatchers:assert_style
- Capybara::RSpecMatchers#have_style
  - https://rubydoc.info/github/jnicklas/capybara/master/Capybara%2FRSpecMatchers:have_style

______________________________________________________________________

Before submitting the PR make sure the following are checked:

- [x] Feature branch is up-to-date with `master` (if not - rebase it).
- [x] Squashed related commits together.
- [x] Added tests.
- [x] Updated documentation.
- [x] Added an entry to the `CHANGELOG.md` if the new code introduces user-observable changes.
- [x] The build (`bundle exec rake`) passes (be sure to run this locally, since it may produce updated documentation that you will need to commit).

If you have created a new cop:

- [x] Added the new cop to `config/default.yml`.
- [x] The cop is configured as `Enabled: pending` in `config/default.yml`.
- [x] The cop is configured as `Enabled: true` in `.rubocop.yml`.
- [x] The cop documents examples of good and bad code.
- [x] The tests assert both that bad code is reported and that good code is not reported.
- [x] Set `VersionAdded` in `default/config.yml` to the next minor version.

PR #1495, opened 2022-11-21
Add new `RSpec/Capybara/PredicateMatcher` cop

ydah:
Follow up: https://github.com/rubocop/rubocop-rspec/pull/1456/files#r1026096951

- [x] Added tests.
- [-] Updated documentation.


spec/rubocop/cop/rspec/predicate_matcher_spec.rb
```diff
@@ -64,6 +66,7 @@
           expect(foo).to include(something)
           expect(foo).to respond_to(:bar)
           expect(foo).to be_something()
+          expect(foo).to match_style(bar: 'baz')
```
pirj, line 69: Would this work with any method?
`rspec-expectations` magically provides `have_` and `be_` dynamic predicate matchers, but I suspect not `match_` ones.
In this case, it's an existing matcher defined by capybara. What if such a matcher doesn't exist, e.g.:

```ruby
expect(inventory.matches_or_a_lighter?).to be_truthy
```

ydah: I remembered the plan that from now on Capybara related cops will be separated from RuboCop RSpec, so I thought again in the direction of creating a new cop in `RSpec/Capybara/*`. What do you think?

pirj: Makes sense :+1:

pirj (approved):
Nice refactoring and a cool cop, thank you!

ydah:
It was conflicted, so I rebased it to the latest master branch.

bquorning (commented):
Won’t this PR make it harder to split e.g. the Capybara cops into a separate gem? See https://github.com/rubocop/rubocop-rspec/discussions/1440 for context.

ydah:
I updated this PR for easy division with reference to https://github.com/rubocop/rubocop-rspec/pull/1511 .

lib/rubocop/cop/rspec/capybara/predicate_matcher.rb
```diff
@@ -0,0 +1,85 @@
...
+        #   # good
+        #   expect(foo).to match_css(bar: 'baz')
+        #   expect(foo).to match_selector(bar: 'baz')
+        #   expect(foo).to match_style(bar: 'baz')
+        #   expect(foo).to match_xpath(bar: 'baz')
+        #
+        #   # also good - It checks "true" strictly.
+        #   expect(foo.matches_style?(bar: 'baz')).to be(true)
+        #
+        # @example Strict: false, EnforcedStyle: inflected
```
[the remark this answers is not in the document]
ydah: The default is to have the same behavior as `RSpec/PredicateMatcher`.
https://github.com/rubocop/rubocop-rspec/blob/e0ff0a0f45c3d53d430378d2336cde2a4260bedf/config/default.yml#L683-L687

However, I personally do not have strong feelings that things should remain as they are.

ydah:
It was conflicted, so I rebased it to the latest master branch.

pirj:
Please don't merge this just yet, I'm extracting `rspec-capybara`.

pirj:
Wondering if `rubocop-rspec` or `rubocop-capybara` is the best fit for this cop 🤔

ydah:
It is indeed difficult to find this cop 🤔 .
RSpec, Minitest and more framework when managed in rubocop-capybara and each testing framework must be able to handle it.
But, I think it is worth to consider a good way to support each testing framework. WDYT?

pirj:
I suggest re-targeting this PR to `rspec-capybara`. We planned to have support for Minitest there, too, but we have some RSpec-only cops now there too.

We only kept `FeatureMethods`, which I would actually move to the root `RSpec` department renaming it to something like `UnnecessaryExampleGroupAlias`.

ydah:
I have opened the following PR and will close this PR
- https://github.com/rubocop/rubocop-capybara/pull/55

PR #1496, opened 2022-11-22
Mark `RSpec/BeEql` as `Safe: false`

r7kamura:
I believe `RSpec/BeEql` should be marked as `Safe: false` for the same reason as `RSpec/Be`.

- https://github.com/rubocop/rubocop-rspec/pull/1490

- [ ] Added tests.
- [x] Updated documentation.

### DIFF
--- a/docs/modules/ROOT/pages/cops_rspec.adoc
+++ b/docs/modules/ROOT/pages/cops_rspec.adoc
@@ -233,10 +233,10 @@ expect(foo).to be(nil)
 | Enabled by default | Safe | Supports autocorrection | Version Added | Version Changed
 
 | Enabled
-| Yes
-| Yes
+| No
+| Yes (Unsafe)
 | 1.7
-| -
+| <<next>>
 |===
 
 Check for expectations where `be(...)` can replace `eql(...)`.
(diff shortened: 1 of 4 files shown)


docs/modules/ROOT/pages/cops_rspec.adoc
```diff
@@ -253,6 +253,10 @@ than `!equal?`. We also do not try to flag `eq` because if
 necessarily the same type as `b` since the `#==` operator can
 coerce objects for comparison.
 
+=== Safety
+
+This cop is unsafe because it changes how values are compared.
```
pirj, line 258: A notable example is:
```ruby
Breed.new('Shepard').eql? Breed.new('Shepard') # => true
Breed.new('Shepard').equal? Breed.new('Shepard') # => false
```

pirj (approved):
Thank you!

docs/modules/ROOT/pages/cops_rspec.adoc
pirj: And the reasoning for making it unsafe is

> Integers, floats, booleans, symbols, and nil can be compared by identity and therefore the `be` matcher is preferable as it is a more strict test.

that we can't imply that the "actual" is one of those types that are safe to compare with `equal?`.

bquorning (approved):

PR #1497, opened 2022-11-22
Add `RSpec/MetadataStyle` and `RSpec/EmptyMetadata` cops

r7kamura:
This Pull Request implements the new Cop discussed in the following Pull Request:

- https://github.com/rubocop/rubocop-rspec/pull/1486#issuecomment-1319258896

Besides just the benefit of a consistent look and feel of code, this Cop would allow us to autocorrect the following code duplication:

```ruby
# 1
describe 'Something', :config, config: true

# 2 (autocorrected by RSpec/MetadataStyle)
describe 'Something', :config, :config

# 3 (autocorrected by RSpec/DuplicatedMetadata)
describe 'Something', :config
```

- [x] Added tests.
- [x] Updated documentation.


lib/rubocop/cop/rspec/metadata_style.rb
```diff
@@ -0,0 +1,197 @@
...
+
+module RuboCop
+  module Cop
+    module RSpec
+      # Use consistent metadata style.
+      #
+      # @safety
+      #   This cop's autocorrection is unsafe in the case of
+      #   `EnforcedStyle: hash` because it cannot determine whether the trailing
+      #   non-literal argument is a Hash or not.
```
pirj, line 11: What is the case exactly?

spec/rubocop/cop/rspec/excessive_docstring_spacing_spec.rb
```diff
@@ -1,6 +1,6 @@
 # frozen_string_literal: true
 
-RSpec.describe RuboCop::Cop::RSpec::ExcessiveDocstringSpacing, only: true do
+RSpec.describe RuboCop::Cop::RSpec::ExcessiveDocstringSpacing, :only do
```
pirj, line 3: Wondering what `:only` means here. Looks like a leftover.

lib/rubocop/cop/rspec/metadata_style.rb
r7kamura: If you try to autocorrect code like this:

```ruby
hash = { config: true }
describe 'Something', :a, hash
```

you will probably get the following result:

```ruby
hash = { config: true }
describe 'Something', hash, a: true
```

---

That's what I thought at first, but now that I've written the test, I see that `RSpec::Cop::RSpec::Metadata#on_block` is not working as intended 🤔 

```ruby
context 'with symbol metadata with another existing non-literal metadata' do
  it 'registers offense' do
    expect_offense(<<~RUBY)
      hash = { config: true }
      describe 'Something', :a, hash
                            ^^ Use hash style for metadata.
    RUBY

    expect_correction(<<~RUBY)
      hash = { config: true }
      describe 'Something', hash, a: true
    RUBY
  end
end
```

```
$ bundle exec rspec spec/rubocop/cop/rspec/metadata_style_spec.rb:213
Run options: include {:locations=>{"./spec/rubocop/cop/rspec/metadata_style_spec.rb"=>[213]}}

Randomized with seed 34162

RuboCop::Cop::RSpec::MetadataStyle
  with `EnforcedStyle: hash`
    with symbol metadata with another existing non-literal metadata
      registers offense (FAILED - 1)

Failures:

  1) RuboCop::Cop::RSpec::MetadataStyle with `EnforcedStyle: hash` with symbol metadata with another existing non-literal metadata registers offense
     Failure/Error: super
     
       Diff:
       @@ -1,4 +1,3 @@
        hash = { config: true }
        describe 'Something', :a, hash
       -                      ^^ Use hash style for metadata.
       
     # ./spec/support/expect_offense.rb:17:in `expect_offense'
     # ./spec/rubocop/cop/rspec/metadata_style_spec.rb:215:in `block (4 levels) in <top (required)>'

Finished in 0.02466 seconds (files took 0.60568 seconds to load)
1 example, 1 failure

Failed examples:

rspec ./spec/rubocop/cop/rspec/metadata_style_spec.rb:214 # RuboCop::Cop::RSpec::MetadataStyle with `EnforcedStyle: hash` with symbol metadata with another existing non-literal metadata registers offense

Randomized with seed 34162
```

We might want to improve this matcher:

https://github.com/rubocop/rubocop-rspec/blob/2aea1be327407a656766432f467beb9dbaecbff6/lib/rubocop/cop/rspec/mixin/metadata.rb#L26-L28

pirj: Indeed, `Metadata` only works with literal metadata.
Maybe we should keep it this way?
Then, we can make this cop autocorrect safe.
I'd add a test that this case with a var is ignored.

r7kamura: The reasons why I think this matcher can be improved are:

- We can consider any non-trailing metadata as Symbol metadata even if it is not literal
- Not really relevant to this issue, but there is no need to support `str` node here

How about marking the autocorrection as safe by saying that it is not supported if the tailing argument is neither a Hash nor Symbol literal?

That way, users can at least know that there is an offense in that case like this:

```ruby
describe 'Something', :a, b do
                      ^^ Use hash style for metadata.
end
```

I know that this kind of case is quite rare, but one might argue that the use of Symbol metadata should be allowed in this case. Personally, I prefer to consistently assume that this is an offense as it is simpler and easier to understand.

lib/rubocop/cop/rspec/metadata_style.rb
```diff
@@ -0,0 +1,211 @@
...
+      class MetadataStyle < Base # rubocop:disable Metrics/ClassLength
+        extend AutoCorrector
+
+        include ConfigurableEnforcedStyle
+        include Metadata
+        include RangeHelp
+
+        # @!method extract_metadata_hash(node)
+        def_node_matcher :extract_metadata_hash, <<~PATTERN
+          (send _ _ _ ... $hash)
```
Darhazer, line 34: wouldn't this be just `last_argument` of the node?

lib/rubocop/cop/rspec/mixin/metadata.rb
```diff
@@ -24,25 +24,40 @@ module Metadata
...
         def on_metadata(_symbols, _pairs)
           raise ::NotImplementedError
         end
+
+        private
+
+        def on_matadata_arguments(metadata_arguments)
+          symbols = metadata_arguments[0..-2]
+          pairs = []
+          last = metadata_arguments.last
```
Darhazer, line 52:
```suggestion
          *symbols, last = metadata_arguments
          pairs = []
```

spec/rubocop/cop/rspec/metadata_style_spec.rb
```diff
@@ -0,0 +1,234 @@
...
+          describe 'Something', :a do
+          end
+        RUBY
+      end
+    end
+
+    context 'with hash metadata with preceding symbol metadata' do
+      it 'registers offense' do
+        expect_offense(<<~RUBY)
+          describe 'Something', :b, a: true do
```
Darhazer, line 36: aren't we missing spec that mixes both preceding symbol metadata, and having other hash metadata?
What about having 2 hash keys with values true?

spec/rubocop/cop/rspec/metadata_style_spec.rb
```diff
@@ -0,0 +1,234 @@
...
+        RUBY
+
+        expect_correction(<<~RUBY)
+          describe 'Something', { a: true } do
+          end
+        RUBY
+      end
+    end
+
+    context 'with symbol metadata with another existing non-literal metadata' do
```
Darhazer, line 222: also would be nice to have test with :a, :b

Darhazer (commented):
Once again, sorry for the delay in the reviews.

Overall this looks good. Perhaps you can bring it up to date and enrich the specs 🙏  🙇

lib/rubocop/cop/rspec/metadata_style.rb
r7kamura: This cannot simply be replaced by `last_argument`, since it also checks whether the last argument node is `hash` type or not.

r7kamura:
Sorry for the delay. I was adding tests and found a bug around autocorrection that left unnecessary commas and empty braces, so I fixed a lot of things.

In addition, since yesterday, CI seemed to be failing for other reasons, so I created the following pull request.

- https://github.com/rubocop/rubocop-rspec/pull/1586

I'm having trouble with CI failing due to decreased coverage, but it doesn't look like I've added any new untested code 🤔

pirj:
Force-pushed to make ruby-head green.

pirj:
> CI failing due to decreased coverage, but it doesn't look like I've added any new untested code 🤔

The only reasonable complaint is that those lines are not covered: https://github.com/rubocop/rubocop-rspec/pull/1497/files#diff-cbd3200503c5c2c8e9b3c2cdc44ee465010609fdbbe49908cdd278da67c9ad1aR84-R85

can it be the culprit?

r7kamura:
Sorry, I was mistaken about the coverage report. Indeed, a condition was written in `if` that would always be true, and the `else` side had not been tested!

```diff
diff --git a/lib/rubocop/cop/rspec/metadata_style.rb b/lib/rubocop/cop/rspec/metadata_style.rb
index 768ea074..84b31b14 100644
--- a/lib/rubocop/cop/rspec/metadata_style.rb
+++ b/lib/rubocop/cop/rspec/metadata_style.rb
@@ -79,11 +79,7 @@ module RuboCop
         end
 
         def format_symbol_to_pair_source(node)
-          if node.sym_type?
-            "#{node.value}: true"
-          else
-            "#{node.source} => true"
-          end
+          "#{node.value}: true"
         end
 
         def insert_pair(corrector, node)
```

lib/rubocop/cop/rspec/metadata_style.rb
```diff
@@ -0,0 +1,217 @@
...
+        def_node_matcher :match_ambiguous_trailing_metadata?, <<~PATTERN
+          (send _ _ _ ... !{hash sym})
+        PATTERN
+
+        def on_metadata(symbols, hash)
+          on_empty_metadata_hash(hash) if hash&.pairs&.empty?
+
+          hash&.pairs&.each do |pair|
+            on_metadata_pair(pair)
+          end
```
Darhazer, line 52: nitpick:
```suggestion
          pairs = hash&.pairs || []
          on_empty_metadata_hash(hash) if pairs.empty?

          pairs.each do |pair|
            on_metadata_pair(pair)
          end
```

lib/rubocop/cop/rspec/metadata_style.rb
```diff
@@ -0,0 +1,217 @@
...
+                node.source_range,
+                side: :left
+              ),
+              :left
+            )
+          )
+        end
+
+        def remove_pair(corrector, node)
+          if !node.parent.source.start_with?('{') || node.left_siblings.any?
```
Darhazer, line 170: the hash node has a `braces?` method

Darhazer (commented):
@r7kamura can you bring this up to date 🙏 
I'll do my best for a timely review this time

lib/rubocop/cop/rspec/metadata_style.rb
r7kamura: This rewrite does not seem equivalent, since it would call `on_empty_metadata_hash(nil)` when `hash` is `nil`.

spec/rubocop/cop/rspec/metadata_style_spec.rb
```diff
@@ -0,0 +1,306 @@
...
+      it 'registers offense' do
+        expect_offense(<<~RUBY)
+          describe 'Something', :a, :b do
+                                    ^^ Use hash style for metadata.
+                                ^^ Use hash style for metadata.
+          end
+        RUBY
+
+        expect_correction(<<~RUBY)
+          describe 'Something', b: true, a: true do
```
Darhazer, line 228: That's a curious side effect. But we can live with that

Darhazer (approved):
Great work. Thank you 🙇

r7kamura:
Thanks, I added test cases for `a: false` and `"a" => true` 👍

Darhazer (approved):

ydah (approved):
Looks good. Thank you!

pirj:
I have limited capacity for reviews now. I trust your judgement guys, please proceed at your discretion.
@r7kamura Thank you for another contribution!

.rubocop.yml
```diff
@@ -151,6 +151,8 @@ RSpec/IdenticalEqualityAssertion:
   Enabled: true
 RSpec/IndexedLet:
   Enabled: true
+RSpec/MetadataStyle:
+  Enabled: true
 RSpec/MatchArray:
   Enabled: true
```
koic, line 157:
```suggestion
RSpec/MatchArray:
  Enabled: true
RSpec/MetadataStyle:
  Enabled: true
```

lib/rubocop/cop/rspec/metadata_style.rb
```diff
@@ -0,0 +1,217 @@
...
+        def on_metadata(symbols, hash)
+          on_empty_metadata_hash(hash) if hash&.pairs&.empty?
+
+          hash&.pairs&.each do |pair|
+            on_metadata_pair(pair)
+          end
+
+          symbols.each do |symbol|
+            on_metadata_symbol(symbol)
+          end
```
koic, line 56: Redundant navigation can be eliminated by changing and adjusting the processing flow.

```suggestion
          symbols.each do |symbol|
            on_metadata_symbol(symbol)
          end

          return unless hash

          on_empty_metadata_hash(hash) if hash.pairs.empty?

          hash.pairs.each do |pair|
            on_metadata_pair(pair)
          end
```

          on_empty_metadata_hash(hash) if hash.pairs.empty?

          hash.pairs.each do |pair|
            on_metadata_pair(pair)
          end
```

File: lib/rubocop/cop/rspec/metadata_style.rb [hunk 12]
```diff
@@ -0,0 +1,219 @@
...
+          )
+        end
+
+        def on_empty_metadata_hash(node)
+          return if style == :hash
+
+          add_offense(node, message: 'Remove empty metadata.') do |corrector|
+            remove_empty_metadata(corrector, node)
+          end
+        end
```
ydah, line 141: ask) I guess we could cut out this to another cop with a name like `RSpec/EmptyMetadata`. What do you think?
If we can keep it small, I want to keep it small, because the cost of maintenance goes up a bit when one cop work increases.

Darhazer: I'm fine either way. @rubocop/rubocop-rspec WDYT ?

pirj: Sounds good to have two cops. Easier than splitting later.

bquorning: Agree on splitting now.

r7kamura: Okay, so I'm going to split it into 2 cops 👌

lib/rubocop/cop/rspec/metadata_style.rb
```diff
@@ -0,0 +1,197 @@
...
+      #   # good
+      #   describe 'Something', :a
+      #
+      # @example EnforcedStyle: hash
+      #   # bad
+      #   describe 'Something', :a
+      #
+      #   # good
+      #   describe 'Something', a: true
+      class MetadataStyle < Base # rubocop:disable Metrics/ClassLength
```
bquorning, line 25: The `Metrics/ClassLength` is spot on, though. This class is so long that I find it hard to read. But extracting logic into classes (e.g. `remove_pair(corrector, node)` → `RemovePair.new(corrector, node).call`) is quite hard.

bquorning (approved):

spec/rubocop/cop/rspec/metadata_style_spec.rb
```diff
@@ -0,0 +1,309 @@
...
+    context 'with 1 boolean hash metadata' do
+      it 'registers offense' do
+        expect_offense(<<~RUBY)
+          describe 'Something', { a: true } do
+                                  ^^^^^^^ Use symbol style for metadata.
+          end
+        RUBY
+
+        expect_correction(<<~RUBY)
+          describe 'Something', :a, {  } do
```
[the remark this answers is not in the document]
r7kamura: Yes, this is why I am trying to add `RSpec/EmptyMetadata` this time.

spec/rubocop/cop/rspec/metadata_style_spec.rb
```diff
@@ -0,0 +1,309 @@
...
+
+    context 'with symbol metadata with another existing non-literal metadata' do
+      it 'registers offense' do
+        expect_offense(<<~RUBY)
+          describe 'Something', :a, b do
+                                ^^ Use hash style for metadata.
+          end
+        RUBY
+
+        expect_no_corrections
```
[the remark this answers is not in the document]
r7kamura: The trailing argument (`b` in this case) may be a Hash, in which case autocorrection cannot be performed. I added this `expect_no_corrections` here to show that specification. If this line is missing, it is not possible to determine it from test code.

ydah (approved):
Looks good! Thank you!

lib/rubocop/cop/rspec/metadata_style.rb
Darhazer: Perhaps the corrector code can be extract to a private AutoCorrector class but I'm happy for this to be an unrelated refactoring - when and if someone wants to undertake it

Darhazer (approved):

Darhazer:
@r7kamura thank you 🙇 🎉

PR #1498, opened 2022-11-23
Remove unnecessary `only: true`

r7kamura:
Perhaps this `only: true` is unnecessary and has just got mixed in, so removed it.

- https://github.com/rubocop/rubocop-rspec/pull/1497#pullrequestreview-1192132522

- [ ] Added tests.
- [x] Updated documentation.


pirj:
Thank you!
Let the origin of this artefact remains a mystery 😄

PR #1500, opened 2022-11-30
Fix false positive ExampleLength offense when skipping/pending a group example

morissetcl:
The `ExampleLength` raise a false positive offense when using `pending` or `skip` on example group.

Example:

```
# Cop config => { "Max": 4 }

pending "#total_count" do
  let(:store) { create(:parkstoreing) }

  it "returns the number of eligible car" do
    create(:car, :qualified)
    create(:car, :disqualified)
    create(:car, :test)
    expect(subject.total_count).to eq(1)
  end
 end
 
 => RSpec/ExampleLength: Example has too many lines. [7/4]
  pending "#total_count" do ...
  ^^^^^^^^^^^^^^^^^^^^^^^^^
```

Whereas using regular example group (eg: `describe`) does not raise an offense.

This PR is a proposal to fix this behaviour by excluding node example group from the `check_code_length` method.

Let me know!
Thanks.

- [X] Added tests.
- [x] Updated documentation.

### DIFF
--- a/spec/rubocop/cop/rspec/example_length_spec.rb
+++ b/spec/rubocop/cop/rspec/example_length_spec.rb
@@ -56,6 +56,36 @@ RSpec.describe RuboCop::Cop::RSpec::ExampleLength do
     end
   end
 
+  context 'given skipped nested examples' do
+    it 'does not flag the example' do
+      expect_no_offenses(<<-RUBY)
+        skip do
+          line 1
+          it do
+            line 1
+            line 2
+            line 3
+          end
+        end
+      RUBY
+    end
+  end
+
+  context 'given pending nested examples' do
+    it 'does not flag the example' do
+      expect_no_offenses(<<-RUBY)
+        pending do
+          line 1
+          it do
+            line 1
+            line 2
+            line 3
+          end
+        end
+      RUBY
+    end
+  end
+
   context 'with CountComments enabled' do
     let(:cop_config) do
       { 'Max' => 3, 'CountComments' => true }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Master (Unreleased)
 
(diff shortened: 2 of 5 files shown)


spec/rubocop/cop/rspec/example_length_spec.rb
```diff
@@ -56,6 +56,36 @@
...
+    it 'does not flag the example' do
+      expect_no_offenses(<<-RUBY)
+        pending do
+          line 1
+          it do
+            line 1
+            line 2
+            line 3
+          end
+        end
```
pirj, line 84: If you remove `line 1`:
```
     # No reason given
     Failure/Error:
       it do
         line 1
         line 2
         line 3
       end

       `it` is not available from within an example (e.g. an `it` block) or from constructs that run in the scope of an example (e.g. `before`, `let`, etc). It is only available on an example group (e.g. a `describe` or `context` block).
```

`pending`/`skip` are aliases of `example`, not `example_group`.
And they are not supposed to have nested examples.

pirj:
`skip` works coincidentally, as the block contents isn't executed.
`pending` fails with an obvious error. [RSpec `pending` doc](https://relishapp.com/rspec/rspec-core/v/3-9/docs/pending-and-skipped-examples/pending-examples#%60pending%60-for-an-example-that-is-currently-passing). `pending` is not purposed to include examples, as it is an example itself.

Use `xdescribe` for skipping whole example groups.

spec/rubocop/cop/rspec/example_length_spec.rb
morissetcl: Thanks @pirj  for your answer. 
I was not aware about that, feel like RSpec is really permissive with the usage of `pending`/`skip` with `example_group`,  and can easily lead to this misunderstanding. 

> And they are not supposed to have nested examples.

Seems like it's a good flag for a missing cop, does it make sense to add cop to flag an `example_group` using these methods?

PR #1501, opened 2022-11-30
Added dependabot for GitHub Actions

ydah:
How about using dependabot in this way?
When this works, a PullRequest is created as follows:
- https://github.com/ruby/bigdecimal/pull/242

It was created with reference to the following:
- https://github.com/ruby/csv/pull/265

- [-] Added tests.
- [^] Updated documentation.


pirj:
Thank you!

PR #1502, opened 2022-12-01
Add autocorrect support for `RSpec/ScatteredSetup`

r7kamura:
I think autocorrection on this cop could be achieved by merging the contents of the two `before` (or `after`) with a blank line.

```ruby
# bad
describe Foo do
  before { setup1 }
  before { setup2 }
end

# good
describe Foo do
  before do
    setup1
    setup2
  end
end
```

Since this Cop often reports offenses when introducing rubocop-rspec, it would be nice to support autocorrection on this cop, even though it is unsafe.

Note that in supporting autocorrection, I have also made changes to the way offenses are reported. When multiple same hooks are written, all of them used to be considered as offenses. However, when implementing autocorrection this was inconvenient, so only the second and later defined hooks are now considered offenses. Given the way other similar Cop offenses are reported, I believe this is an acceptable change.

```ruby
# before
before { bar }
^^^^^^^^^^^^^^ Do not define...
before { baz }
^^^^^^^^^^^^^^ Do not define...
```

```ruby
# after
before { bar }
before { baz }
^^^^^^^^^^^^^^ Do not define...
```

This autocorrection gives little consideration to line breaks and spaces because it's not this cop's business, but I believe that this will not be a problem since they will be autocorrected at the same time by other Layout cops.

- [x] Added tests.
- [x] Updated documentation.

### DIFF
--- a/lib/rubocop/cop/rspec/scattered_setup.rb
+++ b/lib/rubocop/cop/rspec/scattered_setup.rb
@@ -7,6 +7,14 @@ module RuboCop
       #
       # Unify `before`, `after`, and `around` hooks when possible.
       #
+      # Autocorrection is not supported when `around` is used because it can
+      # be difficult to unify them.
+      #
+      # @safety
+      #   This cop's autocorrection is unsafe because it could change the order
+      #   of processing if there is another hook between them.
+      #   (e.g. `let!` between `before`s)
+      #
       # @example
       #   # bad
       #   describe Foo do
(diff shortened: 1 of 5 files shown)


lib/rubocop/cop/rspec/scattered_setup.rb
```diff
@@ -7,6 +7,14 @@ module RSpec
       #
       # Unify `before`, `after`, and `around` hooks when possible.
       #
+      # Autocorrection is not supported when `around` is used because it can
+      # be difficult to unify them.
+      #
+      # @safety
+      #   This cop's autocorrection is unsafe because it could change the order
```
pirj, line 14: Does it also mean that the inspection can raise offences in an unsafe way?

Maybe we can fix that by checking if there are `let!`s between hooks?

r7kamura: > Maybe we can fix that by checking if there are `let!`s between hooks?

`let!` is just one example. In fact, the existence of other method calls that may define the same kind of hook in between, such as `around`, `include_context`, `include`, or something may change the order of execution. In this case, autocorrect may succeed or fail. (Success here means that the test succeeds even if autocorrected)

We can choose not to autocorrect if there are such method calls in between, but I am afraid that if we do so, the number of cases where autocorrection can be applied will be considerably reduced. I think it is rare that the order could cause a problem in fact, and in many cases, the users would want it to be autocorrected even if it has a risk. This is because the users can easily check if it is wrong by executing the autocorrected test cases.

pirj: Would it be safe to combine `before` hooks that follow one another with nothing in between?

Something like `ScatteredLet` could take care of `before` statements intermingled with other statements, and would work together quite nicely.

r7kamura: I see. So if we prepare another `SafeAutoCorrect: false` cop (e.g. `RSpec/ScatteredHook`) that arranges the order of hooks so that they are next to each other, we can make `RSpec/ScatteredSetup`'s autocorrection safe.

```ruby
# bad
before { before1 }
foo
before { before2 }

# good
before { before1 }
before { before2 }
foo
```

In that case, would it be natural to change `RSpec/ScatteredSetup` so that only adjacent hooks are offensed?

```ruby
# good
before { before1 }
foo
before { before2 }
```

pirj: > change RSpec/ScatteredSetup so that only adjacent hooks are offensed

I believe so.
@bquorning @Darhazer WDYT?

pirj: > another `SafeAutoCorrect: false` cop (e.g. `RSpec/ScatteredHook`) that arranges the order of hooks so that they are next to each other

What I find the most appealing is that this cop doesn't have to be `Safe: false`, `SafeAutocorrect: false` is sufficient, as it suggests the right thing. And order-dependency of `let!` and hooks is evil.

pirj:
Fixed in #1616

PR #1503, opened 2022-12-02
Add `RSpec/Rails/TravelAround` and `RSpec/RedundantAround` cops

r7kamura:
Since Rails 5.2, `travel_back` is called automatically, so you can easily do that by just calling it in `before` instead of using `around` with nested blocks.

- https://github.com/rails/rails/pull/29860

In response to this change, I thought it would be a good idea to add a cop recommending the easier way when possible.

```ruby
# bad
around do |example|
  freeze_time do
    example.run
  end
end

# good
before { freeze_time }
```

Unlike a typical Pull Request, the exception here is the addition of two cops.

`RSpec/Rails/TravelAround` applies the following autocorrection:

```ruby
# before
around do |example|
  freeze_time do
    example.run
  end
end
```

```ruby
# after
before { freeze_time }

around do |example|
  example.run
end
```

This may result in a meaningless `around`. As a measure to remove this, I have also added `RSpec/RedundantAround` cop to remove this kind of redundant `around`. I thought it would be more orthogonal to leave it to an independentcCop rather than have each cop take care of such a redundant `around`. This is why the both two cops are included in this one Pull Request.

For further reference, another possible idea related to this Rails change would be to remove the redundant `travel_back` as shown below:

```ruby
# bad
after { travel_back }

# good

# bad
around do |example|
  foo
  example.run
  bar
  travel_back
end

# good
around do |example|
  foo
  example.run
  bar
end
```

In some cases, there may be travel-dependent hooks after this, so this would be an unsafe autocorrection.

- [x] Added tests.
- [x] Updated documentation.


pirj:
If I recall it correctly, `travel_back` is called from `teardown`, which is only called when RSpec's `MinitestLifecycleAdapter` is included. This is the case for Rails example groups, but not in general. Can you please confirm or refute this?
References https://github.com/rspec/rspec-rails/issues/2263 https://github.com/rubocop/rubocop-rails/pull/38

r7kamura:
Oh, it's true. There are cases where `#after_teardown` is not called. Specifically, if the ExampleGroup does not have one of the following `metadata[:type]`, then this Cop is not applicable because of this reason.

- `:controller`
- `:feature`
- `:generator`
- `:helper`
- `:job`
- `:mailbox`
- `:mailer`
- `:model`
- `:request`
- `:request`
- `:request`
- `:routing`
- `:system`
- `:view`

So, would it be better to add the above explanation and set `Safe: false`?
Or, instead, maybe we could limit `Include` to `spec/{controllers,features,...}/**/*.rb`, expecting `config.infer_spec_type_from_file_location!` to be enabled.

pirj:
If the cop is focused to specs with Rails metadata it might be fine.
Just one scenario that is risky:
1. a spec has Rails metadata, but doesn't have really, like a `type: :model` for MyService spec
2. we suggest removing `travel_back`
3. they clean up metadata as redundant when it turns out the spec works without it
4. `travel_back` is now not called automatically on teardown :boom:

r7kamura:
I see. Based on the background above, I agree to set `Safe: false`, as false-positives can happen, even if they are a bit rare.

config/default.yml
```diff
@@ -700,6 +700,13 @@ RSpec/ReceiveNever:
   VersionAdded: '1.28'
   Reference: https://www.rubydoc.info/gems/rubocop-rspec/RuboCop/Cop/RSpec/ReceiveNever
 
+RSpec/RedundantAround:
+  Description: Remove redundant `around` hook.
+  Enabled: pending
+  SafeAutoCorrect: false
```
pirj, line 706: Why?

lib/rubocop/cop/rspec/rails/travel_around.rb
```diff
@@ -0,0 +1,83 @@
...
+
+          MSG = 'Prefer to travel in `before` rather than `around`.'
+
+          TRAVEL_METHOD_NAMES = Set.new(
+            %i[
+              freeze_time
+              travel
+              travel_to
+            ]
+          )
```
pirj, line 38:
```suggestion
          TRAVEL_METHOD_NAMES = %i[freeze_time travel travel_to].to_set
```

config/default.yml
r7kamura: Oh, this is a mistake 🙇

pirj (approved):
Looks great, thank you!

No errors, and offences look legit:
```
/Users/pirj/source/real-world-rspec/gitlabhq/spec/spec_helper.rb:258:3: C: [Correctable] RSpec/RedundantAround: Remove redundant around hook.
  config.around do |example| ...
  ^^^^^^^^^^^^^^^^^^^^^^^^^^

/Users/pirj/source/real-world-rspec/solidus/core/spec/models/spree/return_item/eligibility_validator/time_since_purchase_spec.rb:23:7: C: [Correctable] RSpec/Rails/TravelAround: Prefer to travel in before rather than around.
      travel_to(Time.current) { e.run }

57497 files inspected, 90 offenses detected, 90 offenses autocorrectable
```

lib/rubocop/cop/rspec/redundant_around.rb
```diff
@@ -0,0 +1,69 @@
...
+            (send _ :around ...)
+            (args _)
+            (send _ :run)
+          )
+        PATTERN
+
+        # @!method match_redundant_around_hook_send?(node)
+        def_node_matcher :match_redundant_around_hook_send?, <<~PATTERN
+          (send
+            _
```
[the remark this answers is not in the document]
r7kamura: I set this `_` with the intention of the following code.

```ruby
RSpec.configure do |config|
  config.around do |example|
    example.run
  end
end
```

I'll add those test cases as well 👌

Darhazer:
What happens if the example is using freeze_time, is travel_back still called automatically? In other words, can we now avoid:
```
it 'does something' do
  freeze_time do
    # whole body of the test
  end
end
```
and just call freeze_time in the beginning? Note that this is valid only if the block is the last thing in the example, e.g. we don't rely on time being restored for next statements. 
Also, in the above case, perhaps someone would prefer to still use the block, instead of just `freeze_time` statement

spec/rubocop/cop/rspec/rails/travel_around_spec.rb
```diff
@@ -0,0 +1,144 @@
...
+          example.run
+        end
+      RUBY
+    end
+  end
+
+  context 'with `freeze_time` and another node in `around`' do
+    it 'registers offense' do
+      expect_offense(<<~RUBY)
+        around do |example|
```
Darhazer, line 81: What if we have
```
around do |example|
  freeze_time do
     do_some_preparation
     example.run
  end
end
```
So the do_some_preparation also relies on the time freeze?

spec/rubocop/cop/rspec/redundant_around_spec.rb
```diff
@@ -0,0 +1,83 @@
...
+      expect_correction(<<~RUBY)
+
+      RUBY
+    end
+  end
+
+  context 'with redundant block-pass `around`' do
+    it 'registers offense' do
+      expect_offense(<<~RUBY)
+        around(&:run)
```
Darhazer, line 44: You're missing a test for the numblock syntax, e.g.
```
around { _1.run }
```

Darhazer (commented):
Overall looks good.

spec/rubocop/cop/rspec/rails/travel_around_spec.rb
r7kamura: The above example is ignored, since it is only violated if there is only `example.run` in `freeze_time { ... }` block 👌

Darhazer: @r7kamura  Can you please add a test case for this 👍  I have no other remarks and would be happy to merge. Sorry for the great delay in reviewing this

r7kamura: Thanks, I've added that test case 👌

Darhazer (approved):

Darhazer:
@r7kamura I'm sorry, there were few releases around the capybara extraction.
Would you mind bringing the Changelog up to date and I'll merge the PR

pirj:
Thank you, @r7kamura !

PR #1506, opened 2022-12-06
Add `RSpec/PendingWithoutReason` cop

r7kamura:
## Motivation

I think it is a bad practice to use `pending` or `skip` without giving any reason. So, how about discouraging such practices by default?

## About naming

Since the convention of calling "pending or skip" as "pending" seems to already exist as `RSpec/Pending`, I decided to name it `RSpec/PendingWithoutReason` here as well for consistency. Incidentally, other conventions of calling "pending or skip" as "skip" also seem to exist in `RSpec::Cop::RSpec::SkipOrPending#skipped_in_metadata?`.

Also, `RuboCop::Cop::RSpec::Pending` says "pending or skip", while this `RSpec::Cop::RSpec::SkipOrPending` says "skip or pending". I decided to use "pending or skip" in this case in preference to `RSpec/Pending`.

- [x] Added tests.
- [x] Updated documentation.


r7kamura:
Regardless of whether this Cop is added or not, the other three commits are changes that should be put in independently of each other, and this Cop is dependent on those changes. That is why I am adding 4 commits this time. If you have a policy that each should be separated into a Pull Request, please let me know so I can try to separate them.

config/default.yml
```diff
@@ -182,6 +182,61 @@ RSpec/BeforeAfterAll:
   StyleGuide: https://rspec.rubystyle.guide/#avoid-hooks-with-context-scope
   Reference: https://www.rubydoc.info/gems/rubocop-rspec/RuboCop/Cop/RSpec/BeforeAfterAll
 
+RSpec/Capybara:
```
pirj, line 185: The reason for such ordering was to put departments below main cops

lib/rubocop/cop/rspec/pending_without_reason.rb
```diff
@@ -0,0 +1,123 @@
...
+      #   xdescribe 'something' do
+      #   end
+      #
+      #   # bad
+      #   skip 'does something' do
+      #   end
+      #
+      #   # bad
+      #   it 'does something', :skip do
+      #   end
```
pirj, line 32: How about `it 'works on newer Ruby versions', skip: RUBY_VERSION < 3.0` do?

spec/project/default_config_spec.rb
```diff
@@ -47,7 +47,7 @@ def cop_configuration(config_key)
       .to match_array([*config_keys, 'Metrics/BlockLength'])
   end
 
-  it 'sorts configuration keys alphabetically', :pending do
+  it 'sorts configuration keys alphabetically' do
```
pirj, line 50:
```suggestion
  it 'sorts configuration keys alphabetically', pending: 'sort putting sub-departments below' do
```

spec/rubocop/cop/rspec/pending_without_reason_spec.rb
```diff
@@ -0,0 +1,143 @@
+# frozen_string_literal: true
+
+RSpec.describe RuboCop::Cop::RSpec::PendingWithoutReason do
+  context 'when pending by pending step with reason' do
+    it 'registers no offense' do
```
pirj, line 5: ❤️ for putting good examples above

pirj (approved):
Looks great, thank you!
I'd revert the reordering of cops unless there's a good reason.

spec/project/default_config_spec.rb
bquorning: I tried rewriting this spec in #1507. That PR could replace this commit.

lib/rubocop/cop/rspec/pending_without_reason.rb
```diff
@@ -0,0 +1,123 @@
...
+            #ExampleGroups.skipped
+          PATTERN
+        )
+
+        # @!method skipped_by_metadata_without_reason?(node)
+        def_node_matcher :skipped_by_metadata_without_reason?, <<~PATTERN
+          (send #rspec? {#ExampleGroups.all #Examples.all} ... {<(sym :skip) ...> (hash <(pair (sym :skip) true) ...>)})
+        PATTERN
+
+        def on_send(node)
```
bquorning, line 90: Can we use `RESTRICT_ON_SEND` to improve performance?

r7kamura: This cop needs to inspect user-defined methods for Example and ExampleGroup, so it is not possible to filter them with pre-defined `RESTRICT_ON_SEND` in advance.

lib/rubocop/cop/rspec/pending_without_reason.rb
r7kamura: Ideally, we would like to make it an offense, but in this case the offense is not detectable. In the current implementation, it is only considered an offense when the metadata is specified in Symbol or Boolean literal.

bquorning (approved):
Do we need to cover edge cases like `skip: false` / `skip: nil` / `pending: false` / `pending: nil`? I think perhaps we don’t.

pirj:
Thank you, @r7kamura !

PR #1507, opened 2022-12-06
Fix 'sorts configuration keys alphabetically' spec

bquorning:
The spec now checks that our config file 'sorts configuration keys alphabetically with nested namespaces last'.


pirj (approved):
Thank you!

PR #1509, opened 2022-12-08
Enable `Style/RequireOrder` cop

r7kamura:
By enabling the `Style/RequireOrder` cop added in rubocop 1.40.0, you no longer need to check it yourself in spec/project/project_requires_spec.rb. In addition, it makes the order consistent in other files as well.

Note that if you run it with older than rubocop 1.40.0, it will fail because this Cop does not yet exist.
______________________________________________________________________

Before submitting the PR make sure the following are checked:

- [x] Feature branch is up-to-date with `master` (if not - rebase it).
- [x] Squashed related commits together.
- [ ] Added tests.
- [ ] Updated documentation.
- [ ] Added an entry to the `CHANGELOG.md` if the new code introduces user-observable changes.
- [ ] The build (`bundle exec rake`) passes (be sure to run this locally, since it may produce updated documentation that you will need to commit).

PR #1510, opened 2022-12-08
Fix test failure on rubocop 1.40.0

r7kamura:
Perhaps the following change is the cause.

- https://github.com/rubocop/rubocop/pull/11247

I ran into this failure at the following Pull Request:

- https://github.com/rubocop/rubocop-rspec/pull/1509
- https://github.com/rubocop/rubocop-rspec/actions/runs/3652710709/jobs/6171400421

- [ ] Added tests.
- [ ] Updated documentation.

### DIFF
--- a/spec/rubocop/cop/rspec/file_path_spec.rb
+++ b/spec/rubocop/cop/rspec/file_path_spec.rb
@@ -232,6 +232,8 @@ RSpec.describe RuboCop::Cop::RSpec::FilePath do
   end
 
   it 'does not register offense for absolute file path' do
+    # RuboCop's Commissioner is calling it, too
+    allow(File).to receive(:expand_path).and_call_original
     allow(File).to receive(:expand_path).with('my_class_spec.rb').and_return(
       '/home/foo/spec/very/long/namespace/my_class_spec.rb'
     )


spec/rubocop/cop/rspec/file_path_spec.rb
```diff
@@ -232,7 +232,7 @@ module Medium
   end
 
   it 'does not register offense for absolute file path' do
-    allow(File).to receive(:expand_path).with('my_class_spec.rb').and_return(
+    allow(cop).to receive(:expanded_file_path).and_return(
```
pirj, line 235: Stubject detected!

Wondering why this works, while the other one doesn't :thinking:  as [`expanded_file_path` just delegates to `File.expand_path`](https://github.com/rubocop/rubocop-rspec/blob/a46bc5e9688d60b8c0e4317cdd09af634aa85a22/lib/rubocop/cop/rspec/file_path.rb#L167).

r7kamura: Since rubocop 1.40.0, `File.expand_path` is always called from `RuboCop::Cop::Base#file_name_matches_any?`, so stubbing this in a test case will fail because a different argument is passed there.

- https://github.com/rubocop/rubocop/commit/54bd547fb9b2b457023e9cbb3b3cde185a62a427#diff-7c7ad2faad81954938dc1293695ac6c6a8e6f0f87a87526f411f6c6f0fbdd767L415-R416

On the other hand, stubbing `cop.expanded_file_path` does not pose a problem here.

pirj, line 235: RSpec suggests a way of fixing this:
```
       #<File (class)> received :expand_path with unexpected arguments
         expected: ("my_class_spec.rb")
              got: ("/Users/pirj/source/rubocop-rspec")
        Please stub a default value first if message might be received with other args as well.
```

```suggestion
    allow(File).to receive(:expand_path).and_call_original # RuboCop's Commissioner is calling it, too
    allow(File).to receive(:expand_path).with('my_class_spec.rb').and_return(
```

pirj:
Thank you!

PR #1511, opened 2022-12-11
Use non-RSpec Base cop where possible

pirj:
This is the first, non-breaking, step towards #1440, extraction of Capybara, Rails and FactoryBot cops to their own departments, outside RSpec/.

This change is extracted from a more radical, breaking proof of concept, #1476.

- [-] Added tests.
- [-] Updated documentation.

### DIFF
--- a/bin/build_config
+++ b/bin/build_config
@@ -20,7 +20,8 @@ rspec_cop_path = File.join('lib', 'rubocop', 'cop', 'rspec', 'base.rb')
 YARD::Tags::Library.define_tag('Cop Safety Information', :safety)
 YARD.parse(Dir[glob].prepend(rspec_cop_path), [])
 
-descriptions = RuboCop::RSpec::DescriptionExtractor.new(YARD::Registry.all).to_h
+descriptions =
+  RuboCop::RSpec::DescriptionExtractor.new(YARD::Registry.all(:class)).to_h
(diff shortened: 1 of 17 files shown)


bquorning (approved):
Looks good so far. I was thinking, depending on how long the process is for extracting the “compartment cops”, perhaps we should use a custom cop to check e.g. that the superclass cop is the correct one, etc. What do you think?

pirj:
> depending on how long the process is for extracting the “compartment cops”

Can't tell for sure. It's certainly a time investment, but I already have green specs for `rubocop-capybara` that includes all cops excluding the RSpec-specific `RSpec/Cabybara/FeatureMethods`.
With my recent [suggestion](https://github.com/rubocop/rubocop-rspec/discussions/1440#discussioncomment-4377833) I think we can iteratively extract those departments in a way completely transparent to our users.

I'll do my best to check that all the recent changes to the extracted department cops are merged into the extracted repository, too.

What I'm slightly concerned of is that I don't have sufficient permissions to create a repository for those extracted cops, and it would be rather inappropriate to keep e.g. `rubocop-capybara` under my name.
image
Wondering if creating it under `rubocop-hq` would work, since it's an alias to `rubocop` :smile: 
Do you happen to have permission to create new repos under `rubocop`, @bquorning @Darhazer ?

Issue #1514, opened 2022-12-13
False positive: PendingWithoutReason and factories


tagliala:
Hi,

I don't know if this should be taken into consideration, however this newly introduced cop may fire false positives with factories

### versions

```
1.40.0 (using Parser 3.1.3.0, rubocop-ast 1.24.0, running on ruby 3.0.4) [x86_64-darwin22]
  - rubocop-performance 1.15.1
  - rubocop-rails 2.17.3
  - rubocop-rspec 2.16.0
```

### Reproducible test case

```rb
FactoryBot.define do
  factory :task do
    pending
  end
end
```

`pending` is taking into advantage the `traits_for_enum` functionality of FactoryBot, which automatically create traits for existing enums in the model

### Output

```
Offenses:

spec/factories/tasks.rb:3:5: C: RSpec/PendingWithoutReason: Give the reason for pending.
    pending
```

### Workaround

in `rubocop.yml`

```yml
RSpec/PendingWithoutReason:
  Exclude:
    - 'spec/factories/**/*'
```

mikdiet:
I also have false positive, when tested method named `pending`, as in `expect(statement.pending).to eq 0.0`

ydah:
Thank you for the feedback! We are attempting to correct this below.

- https://github.com/rubocop/rubocop-rspec/pull/1515
- https://github.com/rubocop/rubocop-rspec/pull/1516

PR #1515, opened 2022-12-13
Fix a false positive for `RSpec/PendingWithoutReason` when not inside example

ydah:
Fix: #1514

- [x] Added tests.
- [-] Updated documentation.

### DIFF
--- a/lib/rubocop/cop/rspec/pending_without_reason.rb
+++ b/lib/rubocop/cop/rspec/pending_without_reason.rb
@@ -98,7 +98,10 @@ module RuboCop
         private
 
         def pending_by_pending_step_without_reason?(node)
-          node.method?(:pending) && node.first_argument.nil?
+          node.method?(:pending) &&
+            node.first_argument.nil? &&
+            inside_example?(node) &&
+            !node.receiver
         end
 
         def pending_without_reason?(node)
@@ -108,7 +111,10 @@ module RuboCop
         end
 
         def skipped_by_skip_step_without_reason?(node)
-          node.method?(:skip) && node.first_argument.nil?
+          node.method?(:skip) &&
+            node.first_argument.nil? &&
+            inside_example?(node) &&
+            !node.receiver
         end
 
         def skipped_without_reason?(node)
@@ -116,6 +122,10 @@ module RuboCop
             skipped_by_example_method?(node.block_node) ||
             skipped_by_metadata_without_reason?(node) ||
             skipped_by_skip_step_without_reason?(node)
+        end
+
+        def inside_example?(node)
+          node.ancestors.find { |parent| example?(parent) }
         end
       end
     end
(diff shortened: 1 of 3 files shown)


lib/rubocop/cop/rspec/pending_without_reason.rb
```diff
@@ -98,7 +98,9 @@ def on_send(node)
         private
 
         def pending_by_pending_step_without_reason?(node)
-          node.method?(:pending) && node.first_argument.nil?
+          node.method?(:pending) &&
+            node.first_argument.nil? &&
+            inside_example?(node)
```
Darhazer, line 103: shouldn't we check if we are inside example in the on_send? instead
E.g. having `return unless inside_example?(node)` there

Also do we have a check that it has a `nil` receiver, e.g. that Foo.pending won't produce false positive?

ydah: >Also do we have a check that it has a nil receiver, e.g. that Foo.pending won't produce false positive?

Indeed it is. I updated this PR.

lib/rubocop/cop/rspec/pending_without_reason.rb
```diff
@@ -98,7 +98,10 @@ def on_send(node)
         private
 
         def pending_by_pending_step_without_reason?(node)
-          node.method?(:pending) && node.first_argument.nil?
+          node.method?(:pending) &&
+            node.first_argument.nil? &&
+            inside_example?(node) &&
+            !node.receiver
```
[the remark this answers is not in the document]
ydah: We took the first step in refactoring. What do you think?

lib/rubocop/cop/rspec/pending_without_reason.rb
```diff
@@ -87,35 +87,38 @@ class PendingWithoutReason < Base
...
+          (send nil? ${:pending :skip})
+        PATTERN
+
         def on_send(node)
           if pending_without_reason?(node)
             add_offense(node, message: 'Give the reason for pending.')
           elsif skipped_without_reason?(node)
             add_offense(node, message: 'Give the reason for skip.')
+          elsif inside_example?(node)
+            without_reason?(node) do |method|
```
Darhazer, line 101: Performance wise we may want to first check if it's a pending or skip method, e.g. 
```
elsif %I[pending skip].include?(node.method_name) && inside_example?(node)

File: lib/rubocop/cop/rspec/pending_without_reason.rb (hunk 2 above)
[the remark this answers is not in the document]
Comment (Darhazer) [reply lib/rubocop/cop/rspec/pending_without_reason.rb]: Looks very well 🙇

**Review by Darhazer — APPROVED** [review]

File: lib/rubocop/cop/rspec/pending_without_reason.rb (hunk 3 above)
Comment (Darhazer) [reply lib/rubocop/cop/rspec/pending_without_reason.rb:101]: Actually you don't need the without_reason?(node) to return the method name, as it's just node.method_name
So it could be
```
elsif without_reason?(node) && inside_example?(node)
  add_offense(node, message: "Give the reason for #{node.method_name}.")
```

ydah: Thank you very much. That certainly looks good. I updated this PR.

Darhazer (approved):

pirj (approved):

PR #1516, opened 2022-12-13
FIx a false positive for `RSpec/PendingWithoutReason` when pending/skip is argument of methods

ydah:
This PR is fix a false positive for `RSpec/PendingWithoutReason` when pending/skip is argument of methods. 
The following cases:
```ruby
it 'does something' do
  foo pending
  foo(bar, skip)
  foo(bar, pending: pending)
  is_expected.to match_array [foo, pending, skip, bar]
end
```

- [x] Added tests.
- [-] Updated documentation.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Master (Unreleased)
 
+- Fix a false positive for `RSpec/PendingWithoutReason` when pending/skip is argument of methods. ([@ydah])
 - Add new `RSpec/Capybara/MatchStyle` cop. ([@ydah])
 - Add new `RSpec/Rails/MinitestAssertions` cop. ([@ydah])
 - Fix a false positive for `RSpec/PendingWithoutReason` when not inside example. ([@ydah])
--- a/lib/rubocop/cop/rspec/pending_without_reason.rb
+++ b/lib/rubocop/cop/rspec/pending_without_reason.rb
@@ -97,7 +97,7 @@ module RuboCop
             add_offense(node, message: 'Give the reason for pending.')
           elsif skipped_without_reason?(node)
             add_offense(node, message: 'Give the reason for skip.')
-          elsif without_reason?(node) && inside_example?(node)
+          elsif without_reason?(node) && example?(node.parent)
             add_offense(node,
                         message: "Give the reason for #{node.method_name}.")
           end
@@ -114,10 +114,6 @@ module RuboCop
           skipped_by_example_group_method?(node.block_node) ||
             skipped_by_example_method?(node.block_node) ||
             skipped_by_metadata_without_reason?(node)
-        end
-
-        def inside_example?(node)
-          node.ancestors.find { |parent| example?(parent) }
         end
       end
     end
--- a/spec/rubocop/cop/rspec/pending_without_reason_spec.rb
+++ b/spec/rubocop/cop/rspec/pending_without_reason_spec.rb
@@ -85,6 +85,43 @@ RSpec.describe RuboCop::Cop::RSpec::PendingWithoutReason do
     end
   end
 
+  context 'when pending is argument of methods' do
+    it 'registers no offense' do
+      expect_no_offenses(<<~RUBY)
+        it 'does something' do
+          expect('foo').to eq pending
+          foo(bar, pending)
+          foo(bar, pending: pending)
+          is_expected.to match_array [foo, pending, bar]
+        end
+      RUBY
+    end
+  end
+
+  context 'when skip is argument of methods' do
+    it 'registers no offense' do
+      expect_no_offenses(<<~RUBY)
+        it 'does something' do
+          expect('foo').to eq skip
+          foo(bar, skip)
+          foo(bar, skip: skip)
+          is_expected.to match_array [foo, skip, bar]
+        end
+      RUBY
+    end
+  end
+
+  context 'when pending/skip is argument of methods' do
+    it 'registers no offense' do
+      expect_no_offenses(<<~RUBY)
+        it 'does something' do
+          list = [skip, pending]
+          foo(list)
+        end
+      RUBY
+    end
+  end
+
   context 'when pending by example method' do
     it 'registers offense' do
       expect_offense(<<~RUBY)
@@ -131,6 +168,19 @@ RSpec.describe RuboCop::Cop::RSpec::PendingWithoutReason do
         it 'does something' do
           pending
           ^^^^^^^ Give the reason for pending.
+        end
+      RUBY
+    end
+  end
+
+  context 'when pending/skip inside conditional' do
+    it 'registers no offense' do
+      expect_no_offenses(<<~RUBY)
+        it 'does something' do
+          pending if RUBY_VERSION < '3.0'
+          if RUBY_VERSION < '3.0'
+            skip
+          end
         end
       RUBY
     end


ydah:
If they are arguments to a method call, the following cases are assumed Certainly, with `foo pending`, there will be a case where it is a pending. Should this be checked again in another cop?

``ruby
let(:pending) {...}
# ...
expect('foo').to eq pending.
```

**Comment by Darhazer** [comment]
So you're saying that if `pending` is used as a method argument, there's a high chance it is defined as something else. I tend to agree with this inference

File: lib/rubocop/cop/rspec/pending_without_reason.rb [hunk 1]
```diff
@@ -97,6 +99,14 @@ def on_send(node)
 
         private
 
+        def method_argument?(node)
+          return false unless (send_node = node.ancestors.find(&:send_type?))
```
Darhazer, line 103: If `pending` is used as an argument, its parent node will be of a send (or csend) type. What is the scenario, in which we should go up more?

Darhazer: Ok, I answered myself from the specs: if it's a value in a kwargs / hash

ydah: Yes, and in the case of array as well.

```console
> ruby-parse -e "is_expected.to match_array [foo, skip, bar]"
(send
  (send nil :is_expected) :to
  (send nil :match_array
    (array
      (send nil :foo)
      (send nil :skip)
      (send nil :bar))))
```

pirj:
Are those cases real? Should we really account for that?
I'd imagine a cop preventing from overriding RSpec DSL methods with memoized helpers, or even RSpec doing that.

ydah:
I created the following test code to try it out:

```ruby
# frozen_string_literal: true

RSpec.describe 'a specification' do
  let(:pending) { 'pending' }

  it 'does something' do
    expect(pending).to eq 'pending'
  end
end
```

```
bundle exec rubocop -A spec/rubocop/cop/rspec/test_spec.rb
Inspecting 1 file
C

Offenses:

spec/rubocop/cop/rspec/test_spec.rb:3:16: C: RSpec/DescribeClass: The first argument to describe should be the class or module being tested.
RSpec.describe 'a specification' do
               ^^^^^^^^^^^^^^^^^
spec/rubocop/cop/rspec/test_spec.rb:7:12: C: RSpec/PendingWithoutReason: Give the reason for pending.
    expect(pending).to eq 'pending'
           ^^^^^^^

1 file inspected, 2 offenses detected
```

ydah:
There didn't seem to be any cop that prevents you from overriding RSpec's DSL methods with memoization helpers, or that RSpec does so.
It seems to me that we could leave these to other new cops, but when I looked at this cop by itself, the cases exemplified in the text of these PRs seemed to be as false positives.

lib/rubocop/cop/rspec/pending_without_reason.rb
pirj: Would it match:
```ruby
list = [skip, pending]
foo(list)
```

Is it possible to only trigger the offence if it's directly called?

Would it then detect `pending` inside a conditional?
```ruby
pending if RUBY_VERSION < '3.0'
if RUBY_VERSION < '3.0'
  skip
end
```

pirj (approved):
Makes sense, thank you for the detailed explanation.

ricardopacheco:
@ydah this happen in spec/factories for example. I recommend ignore directories like this, `pending` could be a status for almost everything abstract concepts. In my case, i add my factories which contains this string in exclude.

ydah:
@ricardopacheco Thanks for the suggestion. But I'm sorry to say that it's not happening just in spec/factory. Therefore, we do not believe that the directory exclusion setting is appropriate.

Are there any use cases where this modification would be a problem? If there are, I would like to hear about them.

ricardopacheco:
@ydah i understand, but, in my humble opinion, this cop only make sense in describe/context scenarios. Should be restricted for scenario blocks only.

pirj:
> this cop only make sense in describe/context scenarios

@ricardopacheco This [already done this way](https://github.com/rubocop/rubocop-rspec/blob/495e175da98b86be2c13fcd15cf3808f707094ca/lib/rubocop/cop/rspec/pending_without_reason.rb#L119).
The fix is [not released in a version](https://github.com/rubocop/rubocop-rspec/blob/master/CHANGELOG.md#master-unreleased) though.

Do you mind checking if the cop in edge version of `rubocop-rspec` works for you as intended?

ricardopacheco:
@pirj Well done! 🎉

CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 ## Master (Unreleased)
 
+- FIx a false positive for `RSpec/PendingWithoutReason` when pending/skip is argument of methods. ([@ydah])
```
jdufresne, line 5: Small typo:

```suggestion
- Fix a false positive for `RSpec/PendingWithoutReason` when pending/skip is argument of methods. ([@ydah])
```

pirj:
I dared pushing a slight simplification that is approaching the problem from another side. Instead of excluding certain cases, it only detects certain cases. Specifically, the direct usage, and usage in conditional branches.
The reasoning is that those `pending`s can still be in some other language constructs, and we'll keep having false positives for those cases.
Do you think we're missing some legitimate offences with this change?

lib/rubocop/cop/rspec/pending_without_reason.rb
```diff
@@ -116,8 +116,9 @@ def skipped_without_reason?(node)
             skipped_by_metadata_without_reason?(node)
         end
 
-        def inside_example?(node)
-          node.ancestors.find { |parent| example?(parent) }
+        def directly_inside_example?(node)
+          example?(node.parent) ||
+            (node.parent.if_type? && example?(node.parent.parent.parent))
```
pirj, line 121:
```
> node.parent.parent.parent
=> s(:block,
  s(:send, nil, :it,
    s(:str, "does something")),
  s(:args),
  s(:begin,
    s(:if,
      s(:send,
        s(:const, nil, :RUBY_VERSION), :<,
        s(:str, "3.0")),
      s(:send, nil, :pending), nil),
    s(:if,
      s(:send,
        s(:const, nil, :RUBY_VERSION), :<,
        s(:str, "3.0")),
      s(:send, nil, :skip), nil)))
```

I believe `begin` will always be there for a multi-statement block. Unless the real meat of the test is inside the `else`:
```ruby
it do
  if RUBY_VERSION < '3.0'
    skip
  else
    expect(...).to ...
  end
end
```
But it looks quite weird and doesn't make much sense.

CHANGELOG.md
ydah: Nice catch! Thank you.

ydah:
@pirj Thanks for keeping it simple! I ran this code on my project and it didn't seem to have any error or problems.

lib/rubocop/cop/rspec/pending_without_reason.rb
ydah: I tried to take on a local variable, but I am a little unsure if this variable name is optimal.

spec/rubocop/cop/rspec/pending_without_reason_spec.rb
```diff
@@ -136,6 +173,21 @@
     end
   end
 
+  context 'when pending/skip inside conditional' do
+    it 'registers offense' do
+      expect_offense(<<~RUBY)
+        it 'does something' do
+          pending if RUBY_VERSION < '3.0'
+          ^^^^^^^ Give the reason for pending.
```
koic, line 181: This is a point I have concerns involving existing behavior. I think this cop can allow it when `pending` (and `skip`) in a condition. For example, `Minitest/SkipWithoutReason` cop allows this case. What do you think?
https://github.com/rubocop/rubocop-minitest/issues/198

bquorning: I would say that in cases like `pending if RUBY_VERSION < '3.0'` or `pending if $testing_new_activerecord_version` or `pending if new_version?` it’s pretty obvious why the spec is pending. I think we shouldn’t need to enforce adding a message in these cases.

ydah: Indeed, it is. We have modified the if/unless case so that it is not an offenses.

pirj:
Merging right away since I believe we've come to a conclusion how it should work, to align it with a similar Minitest cop.

Thanks, @ydah !

PR #1519, opened 2022-12-15
Extract Capybara cops

pirj:
https://github.com/rubocop/rubocop-rspec/discussions/1440

Extracted to https://github.com/rubocop-hq/rubocop-capybara

To test that cop renaming isn't a breaking change, add some Capybara cop configuration to some other project, and:

1. Observe a cop renamed warning
```
.rubocop.yml: RSpec/Capybara/SpecificActions has the wrong namespace - should be Capybara
```

2. No errors!

### TODO

- [x] once the gem is published, remove the temp commit that adds `rubocop-capybara` to `Gemfile` 

- [-] Added tests.
- [-] Updated documentation.


Gemfile
```diff
@@ -8,6 +8,7 @@ gem 'bump'
 gem 'rack'
 gem 'rake'
 gem 'rspec', '~> 3.11'
+gem 'rubocop-capybara', github: 'rubocop-hq/rubocop-capybara'
```
ydah, line 11: It looks like the rubocop-capybara organization is rubocop-hq instead of rubocop.
Sorry if this is the correct procedure for adding a new extension gem.

pirj: I don't have permissions to create new repos in `rubocop` 🤷

ydah: Oh that's right. My apologies 🙇‍♂

koic:
> Extracted to https://github.com/rubocop-hq/rubocop-capybara

@pirj I've created https://github.com/rubocop/rubocop-capybara repository and granting permissions to the RuboCop RSpec team.

JFY, It's probably best to have a commit history. Once upon a time when I extracted Performance and Rails from the RuboCop core, I used `git cherry-pick`, not `cp`:
https://speakerdeck.com/koic/road-to-rubocop-1-dot-0?slide=26

However, there may be a better way :-)

bquorning:
> It's probably best to have commit history.

I agree, we should try hard to keep the commit history.

@pirj When you move the code to the new repo, could you do it in another branch than master/main, and open a pull request for us to review and merge?

pirj:
>  I've created https://github.com/rubocop/rubocop-capybara repository and granting permissions to the RuboCop RSpec team.

Thanks a lot, @koic ! :heart:

pirj:
The problem with no history I believe is down to this:
image

Git doesn't seem to be able to detect file renames of significantly changed files.
I intend to split this into two commits, one for file rename, and another to change the indentation.

pirj:
[Rename/remove](https://github.com/rubocop-hq/rubocop-capybara/pull/1/commits/9f185c341039371f5a3dab7612b8b72cc32f0eff).
[The rest](https://github.com/rubocop-hq/rubocop-capybara/pull/1/commits/4c806786e1d822cd8ec6e7e5b9f54cf5955fec41) - reviews are welcome.

Gemfile
ydah, line 11: [note] This place should eventually need to refer to the following
https://github.com/rubocop/rubocop-capybara

pirj: Ideally, this should not point to GitHub, but rather to a version published to RubyGems.

pirj:
@koic I'm trying to `git push --set-upstream origin master`
```
ERROR: Permission to rubocop/rubocop-capybara.git denied to pirj.
fatal: Could not read from remote repository.
```
May I kindly ask you to check permissions there?

pirj:
@koic @bbatsov May I kindly ask you to check permissions in https://github.com/rubocop/rubocop-capybara?
I'd suggest adding @ydah as a maintainer there, as he had contributed a significant part of cops there.
I'll make sure to set gem publishing rights to @bquorning @Darhazer @ydah and optionally myself.

Merry X-Mas! 🎄

koic:
@pirj @ydah 

> May I kindly ask you to check permissions in https://github.com/rubocop/rubocop-capybara?

Sorry for the late reply. I just checked and I had insufficient write permissions. So updated the permissions.

> I'd suggest adding @ydah as a maintainer there, as he had contributed a significant part of cops there.

I've invited as a collaborator for RuboCop Capybara.

Merry Christmas! 🎄

koic:
@pirj JFYI, I think @bquorning can also manage repositories and collaborators related to the RuboCop RSpec. Thank you.

ydah:
@pirj Thanks for the recommendation! I'm very honored.
@koic Thank you for the invitation. I accepted and successfully became a collaborator.

Today is a very wonderful day for me. I really appreciate all of you.
Merry X-Mas! 🎄

pirj:
Looks green. Removed the hack to use `rubocop-capybara` from GitHub source, now it should pick the one from RubyGems.

bquorning (approved):
To be _really_ nitpicking, the commit messages and PR description are a bit out of date after the recent publishing of the rubocop-capybara gem.

pirj:
@bquorning Fair enough 👍. Updated the PR description and the first commit's message.

pirj:
I really appreciate if you find some time and test it on some real project that is uses `rubocop-rspec` and preferably extracted cops to make sure those cops are still on duty :shrek-cat-eyes:

pirj:
I'll dump this here for history.

A method to keep the history for a limited number of files in the repo:
```sh
git clone --no-local rubocop-rspec rubocop-capybara-hist
cd rubocop-capybara-hist
git-filter-repo $( (cat ../rubocop-capybara/rename | xargs -I_ echo '--path-rename _ ') )
git-filter-repo $( (cd ../rubocop-capybara && git ls-files) | xargs -I_ echo '--path _ ')
git remote add origin git@github.com:rubocop/rubocop-capybara.git
git push --set-upstream origin master
git checkout -b capybara-only
cd ../rubocop-capybara
cp -r * ../rubocop-capybara-hist
cp -r $( git ls-files | grep '^\.' | grep -v '\.github' ) ../rubocop-capybara-hist
cp -r .github ../rubocop-capybara-hist
cd -
git commit --all --message 'RSpec -> Capybara'
rvm use --create 3.0.4@rubocop-capybara
bundle
rake
git push origin capybara-only
hub pull-request --assign pirj
```

`rename`:
```
docs/modules/ROOT/pages/cops_rspec_capybara.adoc:docs/modules/ROOT/pages/cops_capybara.adoc
docs/modules/ROOT/pages/development.adoc:docs/modules/ROOT/pages/development.adoc
docs/modules/ROOT/pages/index.adoc:docs/modules/ROOT/pages/index.adoc
docs/modules/ROOT/pages/installation.adoc:docs/modules/ROOT/pages/installation.adoc
docs/modules/ROOT/pages/usage.adoc:docs/modules/ROOT/pages/usage.adoc
lib/rubocop/rspec/config_formatter.rb:lib/rubocop/capybara/config_formatter.rb
lib/rubocop/rspec/description_extractor.rb:lib/rubocop/capybara/description_extractor.rb
lib/rubocop/rspec/version.rb:lib/rubocop/capybara/version.rb
lib/rubocop/cop/rspec/mixin/capybara_help.rb:lib/rubocop/cop/capybara/mixin/capybara_help.rb
lib/rubocop/cop/rspec/mixin/css_selector.rb:lib/rubocop/cop/capybara/mixin/css_selector.rb
lib/rubocop/cop/rspec/capybara/current_path_expectation.rb:lib/rubocop/cop/capybara/current_path_expectation.rb
lib/rubocop/cop/rspec/capybara/match_style.rb:lib/rubocop/cop/capybara/match_style.rb
lib/rubocop/cop/rspec/capybara/negation_matcher.rb:lib/rubocop/cop/capybara/negation_matcher.rb
lib/rubocop/cop/rspec/capybara/specific_actions.rb:lib/rubocop/cop/capybara/specific_actions.rb
lib/rubocop/cop/rspec/capybara/specific_finders.rb:lib/rubocop/cop/capybara/specific_finders.rb
lib/rubocop/cop/rspec/capybara/specific_matcher.rb:lib/rubocop/cop/capybara/specific_matcher.rb
lib/rubocop/cop/rspec/capybara/visibility_matcher.rb:lib/rubocop/cop/capybara/visibility_matcher.rb
lib/rubocop/cop/rspec_cops.rb:lib/rubocop/cop/capybara_cops.rb
lib/rubocop-rspec.rb:lib/rubocop-capybara.rb
rubocop-rspec.gemspec:rubocop-capybara.gemspec
spec/project/changelog_spec.rb:spec/project/changelog_spec.rb
spec/project/default_config_spec.rb:spec/project/default_config_spec.rb
spec/rubocop/rspec/config_formatter_spec.rb:spec/rubocop/capybara/config_formatter_spec.rb
spec/rubocop/rspec/description_extractor_spec.rb:spec/rubocop/capybara/description_extractor_spec.rb
spec/rubocop/cop/rspec/capybara/current_path_expectation_spec.rb:spec/rubocop/cop/capybara/current_path_expectation_spec.rb
spec/rubocop/cop/rspec/capybara/match_style_spec.rb:spec/rubocop/cop/capybara/match_style_spec.rb
spec/rubocop/cop/rspec/capybara/negation_matcher_spec.rb:spec/rubocop/cop/capybara/negation_matcher_spec.rb
spec/rubocop/cop/rspec/capybara/specific_actions_spec.rb:spec/rubocop/cop/capybara/specific_actions_spec.rb
spec/rubocop/cop/rspec/capybara/specific_finders_spec.rb:spec/rubocop/cop/capybara/specific_finders_spec.rb
spec/rubocop/cop/rspec/capybara/specific_matcher_spec.rb:spec/rubocop/cop/capybara/specific_matcher_spec.rb
spec/rubocop/cop/rspec/capybara/visibility_matcher_spec.rb:spec/rubocop/cop/capybara/visibility_matcher_spec.rb
```

checklist:
 - push `master` and a branch with an identical PR to `rubocop/rubocop-capybara` as soon as I have proper permissions to push there
 - publish the gem, and share publish rights with maintainers
 - send a PR to https://github.com/rubocop/docs.rubocop.org/ to add the new extension

bquorning:
Two questions:

1. Now that the cops exist in the rubocop-capybara gem, and we have obsoletion code in place – shouldn’t we delete the classes from rubocop-rspec? 
2. What happens to the FeatureMethods cop? Should we move it into RSpec namespace?

I am sorry if these questions have already been asked and answered elsewhere.

bquorning:
I ran `rubocop --regenerate-todo` on a project, and it finished with this error:

> Error: The `RSpec/Capybara/SpecificMatcher` cop has been moved to `Capybara/SpecificMatcher`.
(obsolete configuration found in .rubocop_todo.yml, please update it)

Now the TODO file contains both a section for both `Capybara/SpecificMatcher` _and_ `RSpec/Capybara/SpecificMatcher`. And a lot of lines are missing from the “regenerated” TODO file, so it seems the error above causes the generation to end prematurely.

pirj:
1. If we remove classes, it would be effectively be cop renaming (e.g. `RSpec/capybara/SpecificFinder` -> `capybara/SpecificFinder`). The config loader won't automatically pick up the configuration from the user file for the old name and map it to the new cop name using our `config/obsoletion.yml`, see [this message](https://github.com/rubocop/rubocop/pull/10725#issuecomment-1163565021).
2. We potentially have another capybara-related cop in `rubocop-rspec`, see #1495. So it might make sense to keep the sub-department. I don't have a strong opinion on this one.

bquorning:
I tried in my local checkout of this PR to remove the extracted cops from our config/default.yml, lib/rubocop/cop/rspec_cops.rb and lib/rubocop/cop/rspec/capybara/, and now the TODO generation works fine.

> If we remove classes, it would be effectively be cop renaming (e.g. RSpec/capybara/SpecificFinder -> capybara/SpecificFinder). The config loader won't automatically pick up the configuration from the user file for the old name and map it to the new cop name using our config/obsoletion.yml, see https://github.com/rubocop/rubocop/pull/10725#issuecomment-1163565021.

Users will have to update their configuration. But otherwise, I don’t see the problem. Do you think it qualifies as a breaking change → major version release?

pirj:
Now I see that the behaviour in the PR description is different. I see an error running `rubocop` in a third-party repo pointing to this PR's local copy.

A quick solution for both running `rubocop` and `rubocop --regenerate-todo` is to remove the `renamed:` section from `config/obsoletion.yml`. This way, there's no "has the wrong namespace" warning, but there's no error either.

So the question is - do we need a warning that suggests adjusting user's config files before we go 3.0?

bquorning:
Weird, even if I remove the `renamed:` section from `config/obsoletion.yml`, I still see “Error: The `RSpec/Capybara/SpecificMatcher` cop has been moved to `Capybara/SpecificMatcher`.” when running `rubocop --regenerate-todo` (with no changes otherwise to my RuboCop config).

pirj:
Wondering what is the difference. Just in case:
```
$ rubocop -V
1.41.1 (using Parser 3.1.3.0, rubocop-ast 1.24.1, running on ruby 2.7.7) [x86_64-darwin21]
  - rubocop-performance 1.15.0
  - rubocop-rails 2.17.2
  - rubocop-rspec 2.16.0
```

Which makes me wonder why `rubocop-capybara` isn't listed.

I'll test if offences for both the old `RSpec/`-namespaced cops and new cop names are both detected if enabled in `.rubocop.yml`.

bquorning:
Did you add this to .rubocop.yml?
```yml
require:
  - rubocop-capybara
```

pirj:
> Did you add this to .rubocop.yml?
> 
> require:
>   - rubocop-capybara
This shouldn’t be necessary at this point?
Wondering if we can add this to our config/default.yml for new cops to be loaded

pirj:
So, I've:
1. made a project
2. added `rubocop-rspec`
3. pointed to this branch
4. added a config for both the new and the old cop names:
```yml
Capybara/CurrentPathExpectation:
  Enabled: true

RSpec/Capybara/SpecificActions:
  Enabled: true
```
5. removed the `renamed` section from `obsoletion.yml`
6. set the project's `NewCops: enabled`
7. added an offensive spec

```
$ rubocop spec/cap_spec.rb
Inspecting 1 file
C

Offenses:

spec/cap_spec.rb:3:5: C: Capybara/SpecificActions: Prefer click_link over find('a').click.
    find('a[href]').click
    ^^^^^^^^^^^^^^^^^^^^^
spec/cap_spec.rb:3:5: C: RSpec/Capybara/SpecificActions: Prefer click_link over find('a').click.
    find('a[href]').click
    ^^^^^^^^^^^^^^^^^^^^^
spec/cap_spec.rb:4:5: C: [Correctable] Capybara/CurrentPathExpectation: Do not set an RSpec expectation on current_path in Capybara feature specs - instead, use the have_current_path matcher on page
    expect(current_path).to eq('/callback')
    ^^^^^^
spec/cap_spec.rb:4:5: C: [Correctable] RSpec/Capybara/CurrentPathExpectation: Do not set an RSpec expectation on current_path in Capybara feature specs - instead, use the have_current_path matcher on page
    expect(current_path).to eq('/callback')
    ^^^^^^

1 file inspected, 4 offenses detected, 2 offenses autocorrectable
```

It doesn't affect the result if I add `require: rubocop-capybara` in `rubocop-rspec`'s `config/default.yml` or the project's `.rubocop.yml`.

It doesn't affe﻿﻿﻿ct the result if I add or remove cops config from the project's `.rubocop.yml`.

If I turn disable new cops, only offences for **both** `CurrentPathExpectation` remain, as it was introduced in version 1.18.

I couldn't reproduce the failure you're experiencing, @bquorning . Is there a way for me to reproduce it?

pirj:
I've also noticed [this](https://github.com/rubocop/rubocop/blob/6d37149b7d373e63f2fbb665df4891fffd22380f/config/obsoletion.yml#L104) is possible:
```yml
# Cops that were extracted into a different gem
extracted:
  Performance/*: rubocop-performance
  Rails/*: rubocop-rails
```

But I'm uncertain if we need this before the major version update.
It might be somehow useful during the major version update. But I guess only if cops were configured in the user's `.rubocop.yml` files.

pirj:
Following [this advice](https://github.com/rubocop/rubocop/pull/10725#issuecomment-1366414078) and the original implementationI've tried changing:
```diff
-        class MatchStyle < ::RuboCop::Cop::Capybara::MatchStyle
-        end
+        MatchStyle = ::RuboCop::Cop::Capybara::MatchStyle
```
and it does remove duplicated offences.

It leaves the one with the new name, not the old one:
```
spec/cap_spec.rb:3:5: C: Capybara/SpecificActions: Prefer click_link over find('a').click.
    find('a[href]').click
    ^^^^^^^^^^^^^^^^^^^^^
spec/cap_spec.rb:4:5: C: [Correctable] Capybara/CurrentPathExpectation: Do not set an RSpec expectation on current_path in Capybara feature specs - instead, use the have_current_path matcher on page
    expect(current_path).to eq('/callback')
    ^^^^^^

1 file inspected, 2 offenses detected, 1 offense autocorrectable
```

🎉 What's even more surprising, this method survives the presence of `renamed` section!
And the renaming warning is in place!
```
.rubocop.yml: RSpec/Capybara/SpecificActions has the wrong namespace - should be Capybara
Inspecting 1 file
C

Offenses:

spec/cap_spec.rb:3:5: C: Capybara/SpecificActions: Prefer click_link over find('a').click.
    find('a[href]').click
    ^^^^^^^^^^^^^^^^^^^^^
spec/cap_spec.rb:4:5: C: [Correctable] Capybara/CurrentPathExpectation: Do not set an RSpec expectation on current_path in Capybara feature specs - instead, use the have_current_path matcher on page
    expect(current_path).to eq('/callback')
    ^^^^^^

1 file inspected, 2 offenses detected, 1 offense autocorrectable
```

Thanks for bringing this up, @koic !

This works like this because of the `Base.inherited` hook.

But at the same time it breaks cops docs generation, including adding it to the registry, its badge etc.
I've played a bit with it, and it feels too hacky:
```ruby
  RuboCop::Cop::Registry.with_temporary_global do
    global = RuboCop::Cop::Registry.global
    # We set a constant for the old cop class, and this skips the
    # `Base.inherited` hook that enlists those old cops.
    %w(
      RuboCop::Cop::RSpec::Capybara::CurrentPathExpectation
      RuboCop::Cop::RSpec::Capybara::MatchStyle
      RuboCop::Cop::RSpec::Capybara::NegationMatcher
      RuboCop::Cop::RSpec::Capybara::SpecificActions
      RuboCop::Cop::RSpec::Capybara::SpecificFinders
      RuboCop::Cop::RSpec::Capybara::SpecificMatcher
      RuboCop::Cop::RSpec::Capybara::VisibilityMatcher
    ).each do |extracted_cop|
      cop = Class.const_get(extracted_cop)
      class << cop
        def badge
          RuboCop::Cop::Badge.for(name)
        end
        def name
          super.sub('::Capybara::', '::RSpec::Capybara::')
        end
        def department
          :'RSpec/Capybara'
        end
      end
      global.enlist(cop)
      global.cops # force enlist to be able to use non-overridden to fetch YARD class docs
      cop.singleton_class.remove_method(:name)
      cop.singleton_class.remove_method(:department)
      cop.singleton_class.remove_method(:badge)
    end

    generator = CopsDocumentationGenerator.new(
      departments: %w[RSpec/Capybara RSpec/FactoryBot RSpec/Rails RSpec]
    )
    generator.call
  end
```

Any suggestions?

pirj:
Fixed doc generation. Appreciate another round of review.
I expect it not to break with obsoletion errors in any circumstances now.

lib/rubocop/cop/rspec/capybara/current_path_expectation.rb
```diff
@@ -4,121 +4,35 @@ module RuboCop
...
-            expectation_last_child = expectation_node.children.last
-            return if %i[regexp str].include?(expectation_last_child.type)
-
-            corrector.insert_after(
-              expectation_last_child,
-              ', ignore_query: true'
-            )
-          end
-        end
+        # @!parse
```
pirj, line 7: This hack is needed to:
1. Only enlist the cop once to avoid obsoletion **errors** (obsoletion warning will be shown)
2. Parse YARD cop docs for the old cop name for cop docs generation

tasks/cops_documentation.rake
```diff
@@ -12,10 +12,41 @@ end
...
+
+        def name
+          super.sub('::Capybara::', '::RSpec::Capybara::')
+        end
+
+        def department
+          :'RSpec/Capybara'
+        end
+      end
+      global.enlist(cop)
```
pirj, line 42: Enlist cops manually, since we don't inherit from `Base`, and inject-enlisting won't work.

tasks/cops_documentation.rake
```diff
@@ -12,10 +12,41 @@ end
...
+          RuboCop::Cop::Badge.for(name)
+        end
+
+        def name
+          super.sub('::Capybara::', '::RSpec::Capybara::')
+        end
+
+        def department
+          :'RSpec/Capybara'
+        end
```
pirj, line 40: A constant, e.g. `Rubocop::cop::RSpec::capybara::SpecificAction` would point to a real class not namespaced with `RSpec`, and its department, name and badge would not include `RSpec`, too.

bquorning:
> I couldn't reproduce the failure you're experiencing, @bquorning . Is there a way for me to reproduce it?

I just tried again, after your latest changes (on SHA 1f3cc3d216d82a037e1dd2246f4423d1ad480db) and it seems to just work now.

And if I run RuboCop before updating the todo file, I get this output:

```
.rubocop_todo.yml: RSpec/Capybara/SpecificMatcher has the wrong namespace - should be Capybara
```

In other words, it seems to work perfectly. 🎉

pirj:
@Darhazer @ydah , does this change look good to you?
I intend to merge and proceed with the extraction of `rubocop-factory_bot` and `rubocop-rspec-rails`.

bquorning:
Is the intention to extract all 3 new gems in the same RuboCop-RSpec release? I keep going back and forth on whether that would be a good idea or not.

pirj:
Good question!
I can't think of any benefit of extracting all three at once. There are certain risks something goes wrong.
The only far-fetched downside of extracting iteratively would be different starting versions.

ydah (approved):
Looks good😀 Great work 🎉 

As for the future extraction schedule, it seems less risky and better to complete the extraction of Capybara and confirm that there seems to be no problem after the release, and then proceed with the next extraction after we are confident of success.

Note that I am opening the following PR for FactoryBot, but I don't think it will be a stop for extraction, so please don't worry about it and proceed.
Even if the extraction is done first, the PR is modified for the post-extraction repository and sent again 👍
- https://github.com/rubocop/rubocop-rspec/pull/1487

bquorning (approved):
🎉

koic:
@pirj Please let me have some feedback.

> No errors!

I get the following error. My local changes are below:

```diff
% git diff 
diff --git a/.rubocop.yml b/.rubocop.yml
index 9223aa07..c07d140c 100644
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -16,6 +16,9 @@ AllCops:
     - 'tmp/**/*'
     - 'spec/smoke_tests/**/*.rb'

+RSpec/Capybara/SpecificActions:
+  Enabled: true
+
 Layout/HashAlignment:
   EnforcedHashRocketStyle:
     - key
diff --git a/Gemfile b/Gemfile
index ad31d642..8e2c8cad 100644
--- a/Gemfile
+++ b/Gemfile
@@ -8,7 +8,7 @@ gem 'bump'
 gem 'rack'
 gem 'rake'
 gem 'rspec', '~> 3.11'
-gem 'rubocop-capybara', github: 'rubocop-hq/rubocop-capybara'
+gem 'rubocop-capybara', github: 'rubocop/rubocop-capybara'
 gem 'rubocop-performance', '~> 1.7'
 gem 'rubocop-rake', '~> 0.6'
 gem 'yard'
```

It's a reproduction:

```
% bundle exec rubocop --only RSpec/Capybara/SpecificActions
Error: The `RSpec/Capybara/SpecificActions` cop has been moved to `Capybara/SpecificActions`.
(obsolete configuration found in .rubocop.yml, please update it)
```

> This works like this because of the `Base.inherited` hook.

I'm sorry, I forgot this existed. In the current implementation, it may be simpler to use inheritance as the first your suggestion.

As an example, I updated https://github.com/rubocop/rubocop/pull/11111 to prevent breaking change. The points are below:

- Inherit deprecated cop
- Update deprecated cop docs
- Add `VersionRemoved` to deprecated config/config.yml (Actually, `VersionRemoved` has never been used, so I'm not sure if it's appropriate)
- Comment out config/obsoletion.yml. If not commented out, an error will occur when a deprecated name is specified in .rubocop.yml

So, it saves the documentation generation hacks and prevents the error.

koic:
> I get the following error. My local changes are below:

Oops, I was using old resources. The warning prints as expected:

```console
% bundle exec rubocop --only RSpec/Capybara/SpecificActions
.rubocop.yml: RSpec/Capybara/SpecificActions has the wrong namespace - should be Capybara
--only option: RSpec/Capybara/SpecificActions has the wrong namespace - should be Capybara
(snip)
```

Please don’t worry about it. 🙇

bquorning:
Yes, as I understand this extraction, we require our users to change any explicit `RSpec/Capybara/SpecificActions` in their config into `Capybara/SpecificActions`.

pirj:
>  we require our users to change any explicit RSpec/Capybara/SpecificActions in their config into Capybara/SpecificActions.

Not really. We just warn them that they may want to change it.

bquorning:
@pirj If you consider this PR done, you should feel free to merge at will. We can then release this as v2.18.0 without any other changes. Or would you prefer a v2.18.0.pre?

pirj:
> would you prefer a v2.18.0.pre?

Good question. Does it require a `--pre` Bundler flag to be installed? I can't recall if I've seen this used in practice.

bquorning:
Yea, exactly that. But perhaps it’s not necessary?

By the way, there were a few patches merged into master yesterday. I’ll see if I can release them separately from this change even though they’re both merged.

Darhazer:
Sorry I've missed your plan for release and merged some patches

bquorning:
No worries. It wasn’t well communicated to start with. 😅

Drenmi:
@pirj How come [RSpec/Capybara/FeatureMethods](https://www.rubydoc.info/gems/rubocop-rspec/RuboCop/Cop/RSpec/Capybara/FeatureMethods) was not included in the extraction? Curious if this was intentional or it was missed out. If it was missed I can help lift it over. 🙂

Drenmi:
> What happens to the FeatureMethods cop? Should we move it into RSpec namespace?

It seems to me this should also go in rubocop-capybara, since it concerns the Capybara DSL. I'll be happy to lift and shift this unless there are any objections. Maybe @pirj had a reason for omitting it?

pirj:
@Drenmi It’s intentional. Even though it would only detect Captbara aliases, it is specific to RSpec aliases. It might become a more generic cop, covering more aliases to make specs less magic-y, and more RSpec’y, like e.g. https://actionpolicy.evilmartians.io/#/./testing?id=rspec-dsl
We can make it configurable in those other gems (I hope config parsing would work with loose dependency).

PR #1522, opened 2022-12-19
Inject config in a new way

pirj:
What Inject was doing will be a part of RuboCop in the next release.
https://github.com/rubocop/rubocop/pull/11265#event-8072854570

### TODO

 - [x] wait for the release (1.41.0?)
 - [x] change the minimum dependency in gemspec
 - [x] propose other well-known extensions to use the same method

- [-] Added tests.
- [-] Updated documentation.
- [?] The build (`bundle exec rake`) passes (be sure to run this locally, since it may produce updated documentation that you will need to commit).


pirj:
@koic - I'm uncertain of your policy of updating the dependency of `rubocop` in `rubocop-rails`, and if you think it's worth changing before this API is stable in `rubocop` 2.0.

koic:
If rubocop/rubocop#11265 would be a new stable API, we wouldn't have to wait for RuboCop 2.0.

Currently, rubocop/rubocop#11265 is marked `@private` and no doc, it doesn't appear to be a stable published API. If the API is unstable, RuboCop upgrades may break third-party extensions and and users can get sad.
So, I think whether rubocop/rubocop#11265 is a stable API is more important than RuboCop 2.0.

Of cause, it is RuboCop Rails policy only and does not prevent what this PR is trying to lead :-)

lib/rubocop-rspec.rb
```diff
@@ -40,10 +39,13 @@
 require_relative 'rubocop/rspec/factory_bot'
 require_relative 'rubocop/rspec/hook'
 
-RuboCop::RSpec::Inject.defaults!
-
 require_relative 'rubocop/cop/rspec_cops'
 
+project_root = File.join(__dir__, '..')
+RuboCop::ConfigLoader.inject_defaults!(project_root)
```
bquorning, line 45: Why don’t we pass in the constant `RuboCop::RSpec::PROJECT_ROOT` here? Although, it’s marked as a private constant, so I don’t know if it’s accessible here.

lib/rubocop-rspec.rb
```diff
@@ -40,10 +39,13 @@
...
+obsoletion = File.join(project_root, 'config', 'obsoletion.yml')
+RuboCop::ConfigObsoletion.files << obsoletion
```
bquorning, line 47: We do this already in https://github.com/rubocop/rubocop-rspec/blob/495e175da98b86be2c13fcd15cf3808f707094ca/lib/rubocop/rspec.rb#L11-L12. Is the plan to remove that file entirely?

bquorning (commented):
Nice cleanup! But I have a few questions.

pirj:
> Before, we did the configuration right before loading out cops, and now we do it right after loading the cops. Is there a specific reason for that?

Yes, it's an intentional change. If cops are not loaded, the `RuboCop::ConfigLoader` complains:
```
rake aborted!
RuboCop::ValidationError: unrecognized cop or department RSpec found in config/default.yml
unrecognized cop or department RSpec/AlignLeftLetBrace found in config/default.yml
...
```

bquorning (approved):
Looks good to me. I understand the concern about whether this private API is going to be stable.

Perhaps we wait with merging until similar PRs for other RuboCop extensions have been reviewed?

pirj:
I'd say I see no reason why this method would be suddenly removed, and if it would, it would be a breaking change for us indeed.
As for a better, documented and a public API to achieve the same - I'm not aware of the plans.

For us - it's a slight cleanup. And considering that we're splitting departments, this benefits all three new gems.

nijikon (approved):

bquorning (approved):

pirj:
It appears that ﻿﻿the extra﻿ction affected this PR, and now the doc generation leaves for aliased cops is broken. I'll take a closer look.

PR #1525, opened 2022-12-22
Update document for RuboCop 1.41.1

ydah:
We are updating the documentation as versions are released that include the following
- https://github.com/rubocop/rubocop/pull/11317

- [-] Added tests.
- [x] Updated documentation.

### DIFF
--- a/lib/rubocop/cop/rspec/context_wording.rb
+++ b/lib/rubocop/cop/rspec/context_wording.rb
@@ -10,7 +10,6 @@ module RuboCop
       # include `if`, `unless`, `for`, `before`, `after`, or `during`.
       # They may consist of multiple words if desired.
       #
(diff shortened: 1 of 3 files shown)


lib/rubocop/cop/rspec/context_wording.rb
```diff
@@ -10,7 +10,6 @@ module RSpec
       # include `if`, `unless`, `for`, `before`, `after`, or `during`.
       # They may consist of multiple words if desired.
       #
-      # @see https://rspec.rubystyle.guide/#context-descriptions
```
ydah, line 13: [memo] The link to StyleGuide has been removed since it is already included in the Reference.

pirj:
So much better now to have it published in our docs, not only code, thank you!

Issue #1148, opened 2021-05-06
Update RSpec/Rails/HttpStatus to support be_* syntax


bkuhlmann:
## Overview

When writing Rails request specs, it'd be handy to auto-correct/enforce the "be_" syntax since it's shorter and easier to read instead of the existing `have_http_status` syntax.

## Desired Behavior

I'd like to propose we update this cop (or make a new one if that makes sense) so we can promote the following:

``` ruby
# No
it "answers unauthenticated status" do
  get api_users_path
  expect(response).to have_http_status(:unauthorized)
end

# Yes
it "answers unauthenticated status" do
  get api_users_path
  expect(response).to be_unauthorized
end
```

## Environment

``` 
1.13.0 (using Parser 3.0.1.1, rubocop-ast 1.5.0, running on ruby 3.0.1 arm64-darwin20)
  - rubocop-performance 1.11.1
  - rubocop-rake 0.5.1
  - rubocop-rspec 2.3.0
```

pirj:
Sounds reasonable. Would you like to hack on it?

bkuhlmann:
Yeah, it's been a while since I've had to write one of these but probably time I get back into this space. Yeah, might take me a while to get ramped up -- on top of everything else I'm doing -- but will see what I can do. :wink:

tejasbubane:
Would it be better to add another `EnforcedStyle` - `be_status` - and maybe make it a default? This will allow users to use old style if they wish to.

sl4vr:
It seems that `be_*` matchers work with predicate methods however `ActionDispatch::TestResponse` doesn't have predicates for [all the statuses](https://github.com/rack/rack/blob/master/lib/rack/utils.rb#L520-L584), but only for most common ones.

pirj:
```diff
- expect(response).to have_http_status(200)
- expect(response).to have_http_status(:ok)
+ expect(response).to be_ok
```

Research:
```
$ rg --no-ignore --hidden --count-matches 'have_http_status.200' | wc -l
126 # files, some file have more than one match

$ rg --no-ignore --hidden --count-matches 'have_http_status.:ok' | wc -l
91 # files, some file have more than one match

$ rg --no-ignore --hidden --count-matches 'be_ok' | wc -l
33 # files, some file have more than one match. Just a few are unrelated to responses

$ rg --no-ignore --hidden --count-matches 'be_successful' | wc -l
365 # files, some file have more than one match. All are related to responses
```

[...]

PR #1527, opened 2022-12-23
Add support `be_status` style for `RSpec/Rails/HttpStatus`

ydah:
Fix: #1148

- [x] Added tests.
- [x] Updated documentation.

### DIFF
--- a/lib/rubocop/cop/rspec/rails/http_status.rb
+++ b/lib/rubocop/cop/rspec/rails/http_status.rb
@@ -8,6 +8,11 @@ module RuboCop
       module Rails
         # Enforces use of symbolic or numeric value to describe HTTP status.
         #
+        # This cop inspects only `have_http_status` calls.
+        # So, this cop does not check if a method starting with `be_*` is used
+        # when setting for `EnforcedStyle: symbolic` or
+        # `EnforcedStyle: numeric`.
+        #
         # @example `EnforcedStyle: symbolic` (default)
         #   # bad
         #   it { is_expected.to have_http_status 200 }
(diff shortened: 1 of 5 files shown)


ydah:
I updated this PR.

pirj:
I'd hold this off until `rubocop-rspec-rails` extraction.

lib/rubocop/cop/rspec/rails/http_status.rb
```diff
@@ -41,12 +54,13 @@ class HttpStatus < Base
           PATTERN
 
           def on_send(node)
```
pirj, line 56: It seems this cop has a bias towards the new style, as it inspects `have_http_status` calls, but doesn't care if `be_*` are used when configured to use `numeric` or `symbolic` styles.

It's not a big problem, since otherwise it would detect false positives in e.g. non-controller/request specs.

Maybe it makes sense to reflect this in our docs somehow.

pirj (commented):
Wondering what are usage stats for various styles in `real-world-rspec` repos.

pirj:
`EnforcedStyle: symbolic`:
```
57497 files inspected, 413 offenses detected
```

`EnforcedStyle: numeric`:
```
57497 files inspected, 933 offenses detected
```

`EnforcedStyle: be_status`:
```
57497 files inspected, 1346 offenses detected
```

ydah:
@pirj Sorry for the late reply. And thanks for checking usage stats for various styles in `real-world-rspec` repos.

lib/rubocop/cop/rspec/rails/http_status.rb
```diff
@@ -8,6 +8,10 @@ module RSpec
       module Rails
         # Enforces use of symbolic or numeric value to describe HTTP status.
         #
+        # This cop inspects only `have_http_status` calls.
+        # So, this cop do not check if a method starting with `be_*` is used
+        # when setting for `EnforcedStyle: symbolic` or `EnforcedStyle: numeric`.
```
ydah, line 13: memo) Resolve: https://github.com/rubocop/rubocop-rspec/pull/1527/files#r1084928242

lib/rubocop/cop/rspec/rails/http_status.rb
```diff
@@ -8,6 +8,11 @@ module RSpec
       module Rails
         # Enforces use of symbolic or numeric value to describe HTTP status.
         #
+        # This cop inspects only `have_http_status` calls.
+        # So, this cop do not check if a method starting with `be_*` is used
```
bquorning, line 12:
```suggestion
        # So, this cop does not check if a method starting with `be_*` is used
```

bquorning (approved):

PR #1532, opened 2022-12-26
Fix a false negative for `RSpec/PredicateMatcher` when using `include` and `respond_to`

ydah:
This PR is fixed false negative in the following code:

```ruby
expect(foo).to include('bar')
expect(foo).to respond_to(:method)
```

I noticed this when I was looking at the coverage of the test code generated by simplecov.

Before
before

After
after

- [x] Added tests.
- [-] Updated documentation.


Darhazer (approved):
Thank you

pirj:
Nice find, thank you!

PR #1533, opened 2022-12-26
Add spec for `RSpec/AlignLeftLetBrace` and `RSpec/AlignRightLetBrace`with `Layout/ExtraSpacing`

ydah:
This PR is add spec for `RSpec/AlignLeftLetBrace` and `RSpec/AlignRightLetBrace` with `Layout/ExtraSpacing`

I found this while looking at the coverage of the test code generated by simplecov.

Before
Before

After
After

- [x] Added tests.
- [-] Updated documentation.

### DIFF
--- a/spec/rubocop/cli/autocorrect_spec.rb
+++ b/spec/rubocop/cli/autocorrect_spec.rb
@@ -48,4 +48,46 @@ RSpec.describe 'RuboCop::CLI --autocorrect' do # rubocop:disable RSpec/DescribeC
       RUBY
     end
   end
+
+  context 'when corrects `RSpec/AlignLeftLetBrace` and ' \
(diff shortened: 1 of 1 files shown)


pirj (approved):
Fantastic!

Wondering, would it be possible to write a spec that would cover `AlignRightLetBrace.autocorrect_incompatible_with`? If I remove this method, this spec remains green (as opposed to it becoming red if I remove the `RuboCop::Cop::Layout::ExtraSpacing.singleton_class.prepend`).

Thank you!

bquorning (approved):

ydah:
@pirj Perhaps, but I think it is because it is a method chain.
If we write this separately, it would look like this, but I don't think it's a good idea to change our code to check for code coverage, so I'm leaving it as it is.

screenshot

pirj:
Thank you, @ydah !

PR #1534, opened 2022-12-27
Add simplecov to measure test coverage of our code

ydah:
This PR is add [simplecov](https://github.com/simplecov-ruby/simplecov) to measure test coverage of our code.

- https://github.com/rubocop/rubocop-rspec/discussions/1530

Branch coverage is enabled, but [minimum coverage by file](https://github.com/simplecov-ruby/simplecov#minimum-coverage-by-file) is enabled.

Below the lowest of the current coverage, RSpec will fail as shown below.
This keeps the test coverage of our code high.

fail

- [-] Added tests.
- [-] Updated documentation.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -40,7 +40,18 @@ jobs:
         with:
           ruby-version: "${{ matrix.ruby }}"
           bundler-cache: true
-      - run: bundle exec rake ${{ matrix.task }}
+      - run: NO_COVERAGE=true bundle exec rake ${{ matrix.task }}
+
+  coverage:
+    runs-on: ubuntu-latest
+    name: "Test coverage"
+    steps:
+      - uses: actions/checkout@v3
+      - uses: ruby/setup-ruby@v1
+        with:
+          ruby-version: "3.2"
+          bundler-cache: true
+      - run: bundle exec rake spec
 
   edge-rubocop:
     runs-on: ubuntu-latest
--- a/.gitignore
+++ b/.gitignore
@@ -17,3 +17,6 @@ pkg
 
 .ruby-gemset
 .ruby-version
+
+# simplecov generated
+coverage
--- a/.simplecov
+++ b/.simplecov
@@ -1,0 +1,8 @@
+# frozen_string_literal: true
+
+SimpleCov.start do
+  enable_coverage :branch
+  minimum_coverage line: 99.86, branch: 95.33
+  add_filter '/spec/'
+  add_filter '/vendor/bundle/'
+end
--- a/Gemfile
+++ b/Gemfile
@@ -10,6 +10,7 @@ gem 'rake'
 gem 'rspec', '~> 3.11'
 gem 'rubocop-performance', '~> 1.7'
 gem 'rubocop-rake', '~> 0.6'
+gem 'simplecov', '>= 0.19'
 gem 'yard'
 
 local_gemfile = 'Gemfile.local'
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -3,6 +3,8 @@
 require 'rubocop'
 require 'rubocop/rspec/support'
 
+require 'simplecov' unless ENV['NO_COVERAGE']
+
 module SpecHelper
   ROOT = Pathname.new(__dir__).parent.freeze
 end


.gitignore
```diff
@@ -17,3 +17,6 @@ pkg
 
 .ruby-gemset
 .ruby-version
+
+# simplecov generated
+coverage
```
ydah, line 22: [memo] Avoid git management of generated files.

spec/spec_helper.rb
```diff
@@ -1,5 +1,12 @@
 # frozen_string_literal: true
 
+require 'simplecov'
+
+SimpleCov.start do
+  enable_coverage :branch
+  add_filter '/spec/'
+end
+
```
ydah, line 9: [memo] Currently, rspec always measures coverage when it is run.

Gemfile
```diff
@@ -10,6 +10,7 @@ gem 'rake'
 gem 'rspec', '~> 3.11'
 gem 'rubocop-performance', '~> 1.7'
 gem 'rubocop-rake', '~> 0.6'
+gem 'simplecov'
```
pirj, line 13: While we're at it, have you tried https://github.com/grosser/single_cov ?
I have no experience with it, and have no idea if it provides such a pretty report as Simplecov.

pirj (approved):
Thank you!

Gemfile
nijikon: @ydah can you add minimal coverage?

nijikon:
@ydah can you add minimal coverage?

pirj:
The coverage report seems to be generated on ci, but there's no way to access it. Does it make sense to make it add its output to all RSpec jobs? Or doing one run for one Ruby version would be sufficient? 🤔 

Found those:
 - https://github.com/marketplace/actions/simplecov-reporter
 - https://github.com/marketplace/actions/simplecov-check
 - https://github.com/marketplace/actions/simplecov-action

nijikon:
What I meant is to add `SimpleCov.minimum_coverage(<current value>)`. The `rspec` command will fail if the coverage is not on at least the current level.

nijikon:
> The coverage report seems to be generated on ci, but there's no way to access it. Does it make sense to make it add its output to all RSpec jobs? Or doing one run for one Ruby version would be sufficient? 🤔
> 
> Found those:
> 
> * https://github.com/marketplace/actions/simplecov-reporter
> * https://github.com/marketplace/actions/simplecov-check
> * https://github.com/marketplace/actions/simplecov-action

Why would we need coverage files on CI?

.simplecov
```diff
@@ -0,0 +1,8 @@
+# frozen_string_literal: true
+
+SimpleCov.start do
+  enable_coverage :branch
+  minimum_coverage line: 99.86, branch: 95.33
```
ydah, line 5: [memo] Current coverage is specified.

Gemfile
ydah: The following was going to need to be added to all specs.

```ruby
require 'spec_helper'
SingleCov.covered!
```

If there are areas not covered by the test as shown below, the following output is likely to be present.

```console
lib/rubocop/cop/rspec/align_left_let_brace.rb new uncovered lines introduced (1 current vs 0 configured)
Lines missing coverage:
lib/rubocop/cop/rspec/align_left_let_brace.rb:25
```

spec/spec_helper.rb
```diff
@@ -3,6 +3,8 @@
 require 'rubocop'
 require 'rubocop/rspec/support'
 
+require 'simplecov' unless ENV['NO_COVERAGE']
```
ydah, line 6: [memo] Turned on/off by environment variable, but usually enabled coverage

.github/workflows/main.yml
```diff
@@ -34,6 +34,23 @@ jobs:
           - internal_investigation
           - spec
     name: "Ruby ${{ matrix.ruby }}: ${{ matrix.task }}"
+    steps:
+      - uses: actions/checkout@v3
+      - uses: ruby/setup-ruby@v1
+        with:
+          ruby-version: "${{ matrix.ruby }}"
+          bundler-cache: true
+      - run: NO_COVERAGE=true bundle exec rake ${{ matrix.task }}
```
ydah, line 43: [memo] Since there is no need to measure coverage for each version, it is disabled here.

.github/workflows/main.yml
```diff
@@ -34,6 +34,23 @@ jobs:
...
+
+  coverage:
```
ydah, line 45: [memo] New coverage job is being created only check coverage.

ydah:
@nijikon I added minimal coverage to match the current coverage. What do you think?

nijikon:
@ydah looks good on my end, let's merge and see how this will work

nijikon (approved):

ydah:
>The coverage report seems to be generated on ci, but there's no way to access it. Does it make sense to make it add its output to all RSpec jobs? Or doing one run for one Ruby version would be sufficient? 🤔

@pirj RSpec will fail if the coverage is below the minimum. So I don't seem to need to do anything on the GitHub Actions side, what do you think?

Gemfile
pirj, line 13: I had some old version of `simplecov` installed, 
```
Warning: Error occurred while trying to load /Users/pirj/source/rubocop-rspec/.simplecov. Error message: undefined method `enable_coverage' for main:Object
```

Updating it resolved the issue:
```
Fetching simplecov 0.22.0 (was 0.17.1)
```

pirj (approved):
Thank you!

Gemfile
ydah: I tried to specify the version. What do you think of this?
https://github.com/rubocop/rubocop-rspec/blob/3bd319323f5959a29617941f89fa1550f70c18ec/Gemfile#L13

Darhazer (approved):
Great :+1:
I'll leave the final decision to @bquorning

bquorning (approved):
Thank you

PR #1539, opened 2023-01-05
Use concurrency for GitHub Actions workflow

ydah:
This PR ensures that only the last execution per branch always works, even when pushed multiple times in a row, for example. The older one is canceled.

I tried it on this branch and sure enough, all but the most recent workflow is cancelled.
https://github.com/rubocop/rubocop-rspec/actions/runs/3848207691
https://github.com/rubocop/rubocop-rspec/actions/runs/3848207692
https://github.com/rubocop/rubocop-rspec/actions/runs/3848207694

Refs:
https://docs.github.com/en/actions/using-jobs/using-concurrency https://github.blog/changelog/2021-04-19-github-actions-limit-workflow-run-or-job-concurrency/

- [-] Added tests.
- [-] Updated documentation.

### DIFF
--- a/.github/workflows/codespell.yml
+++ b/.github/workflows/codespell.yml
@@ -1,6 +1,9 @@
 name: CodeSpell
 on:
   - pull_request
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
 jobs:
   codespell:
     name: CodeSpell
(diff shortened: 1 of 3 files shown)


pirj:
This is so good, thank you! I can't believe this feature is nearly 2 years old now, I've completely missed it from the docs.

PR #1542, opened 2023-01-08
Add rake tasks for alternative way to specify Changelog entries

ydah:
Follow: https://github.com/rubocop/rubocop/pull/8930 and https://github.com/rubocop/rubocop-rails/pull/496

- [ ] Added tests.
- [-] Updated documentation.

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -18,10 +18,25 @@ If you encounter problems or have ideas for improvements or new features, please
 2. Create a feature branch.
 3. Make sure to add tests.
 4. Make sure the test suite passes (run `rake`).
-5. Add a [changelog](https://github.com/rubocop/rubocop-rspec/blob/master/CHANGELOG.md) entry.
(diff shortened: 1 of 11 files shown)


pirj:
I feel severe reluctance to this approach. We don't have that many updates, and rebasing and fixing the changelog doesn't feel to be a significant overhead. Contrary to a strict suggested format.

The changelog file in between the releases is also now scattered across several files.

I hope that with splitting the repo, the surface will reduce as well.

During a few recent minor releases, we were doing a good job of communicating on PRs and merged them quickly.

ydah:
- https://github.com/rubocop/rubocop-rspec/discussions/1417
It was spoken about here, so we worked on it. However, it is true that we don't get that many PRs sent to us on a daily basis now, so it may just be overhead.
Basically this approach is the same as rubocop and rubocop-rspec, so we can always come back to this.

pirj:
My apologies. I refreshed my memory by reading #1417 and now recall I gave in, provided all the arguments, but kept my previous opinion.
Thanks for this PR.
I would personally postpone the decision until the need arises.

ydah:
We can resume this again, so let's Close this PR.

PR #1543, opened 2023-01-10
Change the start label of a heredocument to be enclosed in single quotes

ydah:
This PR is change as follows

```ruby
# bad - This would require escaping inside heredoc. 
expect_offense(<<-RUBY)
  it 'should do something ' \\ <---here
  ^^^^^^^^^^^^^^^^^^^^^^^ Do not use should when describing your tests.
    'and correctly fix' do
  end
RUBY

# good - This makes the heredoc readable with no need for escaping inside it.
expect_offense(<<-'RUBY')
  it 'should do something ' \
  ^^^^^^^^^^^^^^^^^^^^^^^ Do not use should when describing your tests.
    'and correctly fix' do
  end
RUBY
```

- [x] Added tests.
- [-] Updated documentation.

### DIFF
--- a/spec/rubocop/cop/rspec/context_wording_spec.rb
+++ b/spec/rubocop/cop/rspec/context_wording_spec.rb
@@ -13,17 +13,17 @@ RSpec.describe RuboCop::Cop::RSpec::ContextWording do
   end
 
   it 'finds context without `when` at the beginning' do
-    expect_offense(<<-RUBY)
+    expect_offense(<<-'RUBY')
       context 'the display name not present' do
-              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ Context description should match /^when\\b/, /^with\\b/, or /^without\\b/.
+              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ Context description should match /^when\b/, /^with\b/, or /^without\b/.
       end
     RUBY
   end
 
   it 'finds shared_context without `when` at the beginning' do
-    expect_offense(<<-RUBY)
+    expect_offense(<<-'RUBY')
       shared_context 'the display name not present' do
-                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ Context description should match /^when\\b/, /^with\\b/, or /^without\\b/.
+                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ Context description should match /^when\b/, /^with\b/, or /^without\b/.
       end
     RUBY
   end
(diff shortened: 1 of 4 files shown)


spec/rubocop/cop/rspec/context_wording_spec.rb
```diff
@@ -13,17 +13,17 @@
   end
 
   it 'finds context without `when` at the beginning' do
-    expect_offense(<<-RUBY)
+    expect_offense(<<-'RUBY')
       context 'the display name not present' do
-              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ Context description should match /^when\\b/, /^with\\b/, or /^without\\b/.
+              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ Context description should match /^when\b/, /^with\b/, or /^without\b/.
```
pirj, line 18: There is a difference at a glance. Wondering how it turns out that both work :thinking: 
```
<<~A
/^when\b/
A
=> "/^when\b/\n"
<<~'A'
/^when\b/
A
=> "/^when\\b/\n"
```
Maybe there's some escaping logic in `expect_offense`? I can't find any code in `expect_offense` and there are no specs for this method or `AnnotatedSource` in `rubocop` specs.

spec/rubocop/cop/rspec/excessive_docstring_spacing_spec.rb
```diff
@@ -120,17 +120,17 @@
 
     it 'skips \-separated multiline strings whose trailing whitespace ' \
        'makes sense' do
-      expect_no_offenses(<<-RUBY)
-        describe '#mymethod ' \\
+      expect_no_offenses(<<-'RUBY')
+        describe '#mymethod ' \
             '(is cool)' do
         end
```
pirj, line 126:
```
<<-A
a \
b
A
=> "a b\n"
```
The ` \` part is lost, and that undermines the very existence of this example which is purposed to test "\-separated multiline strings".

spec/rubocop/cop/rspec/context_wording_spec.rb
ydah: The argument of `expect_offense` is a heap document, so a heap document enclosed in single quotes is completely a string as written. So the following would be the same.

```ruby
irb(main):001:0" <<~A
irb(main):002:0" /^when\\b/
irb(main):003:0> A
=> "/^when\\b/\n"
irb(main):004:0' <<~'A'
irb(main):005:0' /^when\b/
irb(main):006:0> A
=> "/^when\b/\n"
```

Am I missing something?

spec/rubocop/cop/rspec/excessive_docstring_spacing_spec.rb
ydah: The results seem to be the same here as well.

```ruby
irb(main):001:0" <<-A
irb(main):002:0" a \\
irb(main):003:0" b
irb(main):004:0> A
=> "a \\\nb\n"
irb(main):005:0' <<-'A'
irb(main):006:0' a \
irb(main):007:0' b
irb(main):008:0> A
=> "a \\\nb\n"
```

ydah:
The title of the PR was incorrect and has been corrected.

spec/rubocop/cop/rspec/context_wording_spec.rb
pirj: Oh, please accept my apologies, I've missed the point that you've changed the contents of the string and removed unnecessary escaping.
Looks much better this way indeed!

pirj (approved):
Much easier to read now, thank you!

spec/rubocop/cop/rspec/context_wording_spec.rb
bquorning: Please add to the PR description what the benefit/downside is of either syntax. I always forget which is which.

Issue #1518, opened 2022-12-15
RSpec/StubbedMock false positive


ngouy:
I have a case where I want to play more extensively with the arguments that are passed to a call
I don't stubb the return value at all
Yet it is suggested I use `allow` over `expect` which really is out of question given I absolutely want to know if the method  was called, with specific args

```rb
# I want to test that .some_method is called on MyClass with some arguments
expect(MyClass).to receive(:some_method).with("a", "b") # rubocop is fine

expect(MyClass).to receive(:some_method) do |arg_1, arg_2| do
  # I want to do some complicated stuff. (It's exactly what I do, the point is I cannot use argument matchers)
  row = Table.find_by(id: arg_1)
  expect(arg_1).to_not be nil
end # rubocop tell me I should use `allow(MyClass)`, which is absolutely not the point
```

ngouy:
@pirj Thanks for confirming! (so fast 🏎️ 💨 )

Maybe we could ignore as long as we don't call `and_return` maybe?

Also I could try, it won't be before 2023. things are getting busy for OEY

pirj:
Even though we have[`RSpec/ReturnFromStub`](https://docs.rubocop.org/rubocop-rspec/cops_rspec.html#rspecreturnfromstub) that suggests using `and_return` by default, it has an alternative `block` style that would fix all stubs to use the block form. This would effectively disable `RSpec/StubbedMock` if we would just ignore this form.

On the other hand, if the block form captures arguments, we can assume that the block is used to verify arguments, not to stub the response.

ngouy:
So what is your directive?
Seems to me like the 2 options have pro and cons
And I am not sure to be knowledgable enough to pick one over the other myself

pirj:
The AST difference is:
```ruby
receive(:foo) { bar }
```
```
(block
  (send nil :receive
    (sym :foo))
  (args)
  (send nil :bar))
```

```ruby
receive(:foo) { |x| bar }
```
```
(block
  (send nil :receive
    (sym :foo))
  (args
    (procarg0
      (arg :x)))
  (send nil :bar))
```

The [node matcher](https://github.com/rubocop/rubocop-rspec/blob/495e175da98b86be2c13fcd15cf3808f707094ca/lib/rubocop/cop/rspec/stubbed_mock.rb#L93) is:
```
(block #message_expectation? args _)  # receive(:foo) { 'bar' }
```

[...]

PR #1546, opened 2023-01-11
Fix a false positive for `RSpec/StubbedMock` when stubbed message expectation with a block and block parameter

ydah:
Fix: https://github.com/rubocop/rubocop-rspec/issues/1518

For the context, we don't want to conside `expect.to receive(:foo) { |arg, arg2| ... }` an offence, since most likely the block contains expectations on arguments.
If it's an `allow`, those argument expectations might be missed of `foo` is not called.

- [x] Added tests.
- [-] Updated documentation.


pirj (approved):
Awesome, thank you!

bquorning (approved):

PR #1548, opened 2023-01-11
Add a test case for `RSpec/ContextMethod` when context without `.` or `#` at the beginning

ydah:
This PR is add a test case for `RSpec/ContextMethod` when context without `.` or `#` at the beginning.

The test case was added because there was no case where the following method was false.

https://github.com/rubocop/rubocop-rspec/blob/a1ece2cad9d18fdfe0bdd481f35d7cc471f1e5d4/lib/rubocop/cop/rspec/context_method.rb#L47-L49

- [x] Added tests.
- [-] Updated documentation.

### DIFF
--- a/spec/rubocop/cop/rspec/context_method_spec.rb
+++ b/spec/rubocop/cop/rspec/context_method_spec.rb
@@ -13,7 +13,7 @@ RSpec.describe RuboCop::Cop::RSpec::ContextMethod do
 
   it 'ignores context without `.` or `#` at the beginning' do
     expect_no_offenses(<<-RUBY)
-      context '_foo_bar' do
+      context "when it's sunny" do
       end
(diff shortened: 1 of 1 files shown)


spec/rubocop/cop/rspec/context_method_spec.rb
```diff
@@ -11,6 +11,13 @@
     RUBY
   end
 
+  it 'ignores context without `.` or `#` at the beginning' do
+    expect_no_offenses(<<-RUBY)
+      context '_foo_bar' do
```
pirj, line 16:
```suggestion
      context "when it's sunny" do
```
I think this would be closer to real-life usages

pirj (approved):
Looks good, thank you!

Issue #1549, opened 2023-01-12
False positive for RSpec/PendingWithoutReason in 2.16.0 when object has a method pending


sihu:
Hi rubocop-rspec team,

I noticed that we had a false positive in our project (at least i think it's a false positive): There is a object with an enum `state` with the value `pending`. In our test we are querying all pending transations. I would expect that the spec does not consider this as an issue.

```
    it 'approves the transaction' do
      expect do
        request
      end.to change(Transaction.pending, :count).from(1).to(0)
```

(Ruby 2.7.7, rubocop 1.43.0, rubocop-ast 1.24.1, rubocop-rspec 2.16.0)

ydah:
Thank you for your report!
This issue has been fixed in https://github.com/rubocop/rubocop-rspec/pull/1515 and https://github.com/rubocop/rubocop-rspec/pull/1516, but not yet released. Please wait for the next release.

pirj:
My apologies for a temporary inconvenience.

@bquorning does it make sense to release bug fixes in a patch release before 2.17 with Capybara extraction?

PR #1551, opened 2023-01-13
Bump version to v2.17.0

bquorning:
@pirj You suggested in https://github.com/rubocop/rubocop-rspec/issues/1549#issuecomment-1381360466 that we release a v2.16.1 now, then release the Capybara extraction in a v2.17.0. But seeing that there are new cops in the changelog, I would prefer to release a v2.17.0 now, and then a v2.18.0 with the Capybara extraction. Would that be ok, or does that mess up the version numbers of rubocop-capybara?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Master (Unreleased)
 
+## 2.17.0 (2023-01-13)
+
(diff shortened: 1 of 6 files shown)


pirj (approved):
I think it's completely fine. Let's postpone capybara extraction to the next one, probably soon after this one.

Thank you!

PR #1554, opened 2023-01-13
Fix a false positive for `RSpec/PredicateMatcher` when multiple arguments and not replaceable method

ydah:
This PR is fix a false positive for `RSpec/PredicateMatcher` when multiple arguments and not replaceable method.

following code:
```ruby
expect(foo).to include(foo, bar)
```

`RSpec/PredicateMatcher` autocorrects it as follows:
```ruby
expect(foo.include?(foo, bar)).to be(true)
```

However, this is a SyntaxError.
```
ArgumentError:
       wrong number of arguments (given 2, expected 1)
```

- [x] Added tests.
- [-] Updated documentation.

### DIFF
--- a/lib/rubocop/cop/rspec/predicate_matcher.rb
+++ b/lib/rubocop/cop/rspec/predicate_matcher.rb
@@ -149,11 +149,22 @@ module RuboCop
           return if part_of_ignored_node?(node)
 
           predicate_matcher?(node) do |actual, matcher|
+            next unless replaceable_matcher?(matcher)
+
             add_offense(node, message: message_explicit(matcher)) do |corrector|
               next if uncorrectable_matcher?(node, matcher)
 
               corrector_explicit(corrector, node, actual, matcher, matcher)
             end
+          end
+        end
+
+        def replaceable_matcher?(matcher)
+          case matcher.method_name.to_s
+          when 'include'
+            matcher.arguments.one?
+          else
+            true
           end
         end
 
--- a/spec/rubocop/cop/rspec/predicate_matcher_spec.rb
+++ b/spec/rubocop/cop/rspec/predicate_matcher_spec.rb
@@ -95,14 +95,14 @@ RSpec.describe RuboCop::Cop::RSpec::PredicateMatcher do
 
       it 'registers an offense for a predicate method with heredoc' do
         expect_offense(<<~RUBY)
-          expect(foo.include?(<<~TEXT)).to be_truthy
-          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ Prefer using `include` matcher over `include?`.
+          expect(foo.something?(<<~TEXT)).to be_truthy
+          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ Prefer using `be_something` matcher over `something?`.
             bar
           TEXT
         RUBY
 
         expect_correction(<<~RUBY)
-          expect(foo).to include(<<~TEXT)
+          expect(foo).to be_something(<<~TEXT)
             bar
           TEXT
         RUBY
@@ -346,14 +346,14 @@ RSpec.describe RuboCop::Cop::RSpec::PredicateMatcher do
          'heredoc and multiline expect' do
         expect_offense(<<~RUBY)
           expect(foo)
-          ^^^^^^^^^^^ Prefer using `include?` over `include` matcher.
-            .to include(<<~TEXT)
+          ^^^^^^^^^^^ Prefer using `something?` over `be_something` matcher.
+            .to be_something(<<~TEXT)
               bar
             TEXT
 
           expect(foo)
-          ^^^^^^^^^^^ Prefer using `include?` over `include` matcher.
-            .to include(bar, <<~TEXT, 'baz')
+          ^^^^^^^^^^^ Prefer using `something?` over `be_something` matcher.
+            .to be_something(bar, <<~TEXT, 'baz')
               bar
             TEXT
         RUBY
@@ -365,14 +365,14 @@ RSpec.describe RuboCop::Cop::RSpec::PredicateMatcher do
          'heredoc include #{} and multiline expect' do
         expect_offense(<<~'RUBY')
           expect(foo)
-          ^^^^^^^^^^^ Prefer using `include?` over `include` matcher.
-            .to include(<<~TEXT)
+          ^^^^^^^^^^^ Prefer using `something?` over `be_something` matcher.
+            .to be_something(<<~TEXT)
               #{bar}
             TEXT
 
           expect(foo)
-          ^^^^^^^^^^^ Prefer using `include?` over `include` matcher.
-            .to include(bar, <<~TEXT, 'baz')
+          ^^^^^^^^^^^ Prefer using `something?` over `be_something` matcher.
+            .to be_something(bar, <<~TEXT, 'baz')
               #{bar}
             TEXT
         RUBY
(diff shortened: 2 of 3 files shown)


lib/rubocop/cop/rspec/predicate_matcher.rb
```diff
@@ -25,6 +27,16 @@ def check_inflected(node)
           end
         end
 
+        def replaceable_predicate?(predicate)
+          case predicate.method_name.to_s
+          when 'is_a?', 'instance_of?', 'include?',
+            'respond_to?', 'exist?', 'exists?', /^has_/
```
koic, line 33: Match to multiline is unexpected, right?

```suggestion
            'respond_to?', 'exist?', 'exists?', /\Ahas_/
```

lib/rubocop/cop/rspec/predicate_matcher.rb
```diff
@@ -149,12 +161,25 @@ def check_explicit(node) # rubocop:disable Metrics/MethodLength
...
               corrector_explicit(corrector, node, actual, matcher, matcher)
             end
           end
         end
 
+        def replaceable_matcher?(matcher)
+          case matcher.method_name.to_s
+          when 'be_a', 'be_an', 'be_a_kind_of', 'a_kind_of', 'be_kind_of',
+            'be_an_instance_of', 'be_instance_of', 'an_instance_of',
+            'include', 'respond_to', /^have_(.+)/
```
koic, line 176: Ditto:

```suggestion
            'include', 'respond_to', /\Ahave_(.+)/
```

tiagotex:
I've just downloaded the latest version and am using the following config:

```yml
RSpec/PredicateMatcher:
  Strict: true
  EnforcedStyle: explicit
```

It now corrects all occurrences of predicate matcher `include` to be similar to `expect(foo.include?("bar")).to be(true)`, but in the case of `include` we would want to skip the checks for this cop. Would you be open to a PR allowing users to define ignored predicates?

lib/rubocop/cop/rspec/predicate_matcher.rb
bquorning: It looks like the capture (`(.+)`) is not used for anything. So I think it could be replaced by a dot:
```suggestion
            'include', 'respond_to', /\Ahave_./
```

lib/rubocop/cop/rspec/predicate_matcher.rb
ydah: Indeed it is. Thank you so much.

lib/rubocop/cop/rspec/predicate_matcher.rb
ydah: Thank you so much. I updated this PR.

bquorning (approved):

spec/rubocop/cop/rspec/predicate_matcher_spec.rb
```diff
@@ -312,6 +326,23 @@
...
+          expect(foo).to be_an foo, bar
+          expect(foo).to be_a_kind_of(foo, bar, baz)
+          expect(foo).to a_kind_of foo, bar, baz
+          expect(foo).to be_kind_of(foo, bar)
+          expect(foo).to be_an_instance_of foo, bar
+          expect(foo).to be_instance_of(foo, bar, baz)
+          expect(foo).to an_instance_of foo, bar, baz
+          expect(foo).to include(foo, bar)
+          expect(foo).to respond_to foo, bar
+          expect(foo).to have_key(foo, bar, baz)
```
pirj, line 342: Is this even possible?

I suggest to only target those predicate matchers in this fix that can reasonably accept more than one argument.

`include` is a special case, as it's behaves differently depending on the `actual`'s type. And it [accepts varargs](https://github.com/rspec/rspec-expectations/blob/4ef5b0e21c2273a14ff46b2382ab921e7b100d72/lib/rspec/matchers/built_in/include.rb#L15).

Are all other listed predicate matchers like that?

ydah: I thought there might be a case where it would be a false positive if the method was overridden.
But is it excessive to care so much?

bquorning:
FYI I intend to do a v2.17.1 release when this and #1555 is merged. Followed by a v2.18.0 release with the extraction of Capybara cops (#1519).

spec/rubocop/cop/rspec/predicate_matcher_spec.rb
pirj: If it's overridden - it won't result in an ArgumentError, would it?

ydah: It is likely that the method to be replaced must be overridden in the same way. So, I am trying to avoid unnecessary autocorrects and offense here. 

So I think it would be better not to make it an offense here except for methods that can exist without override.

pirj: `be_a` [won't accept two arguments](https://github.com/rspec/rspec-expectations/blob/4ef5b0e21c2273a14ff46b2382ab921e7b100d72/lib/rspec/matchers.rb#L355):
```
> be_a(Integer, 1).matches?(1)
ArgumentError: wrong number of arguments (given 2, expected 1)
from /Users/pirj/.rvm/gems/ruby-3.0.3@rubocop-rspec/gems/rspec-expectations-3.12.1/lib/rspec/matchers.rb:355:in `be_a'
```

How would the original `be_a(foo, bar)` code even work?

ydah: > be_a [won't accept two arguments](https://github.com/rspec/rspec-expectations/blob/4ef5b0e21c2273a14ff46b2382ab921e7b100d72/lib/rspec/matchers.rb#L355):

Yes. I understand about that.

>How would the original be_a(foo, bar) code even work?

In the case of the original, I believe it would be an ArgumentError as you say.

So I didn't know if it was a good thing to give it as a violation since it could be a custom matcher, etc., redefined by the user.
The reason is that the methods to which they are substituted must also be overridden and extended to multiple arguments.

However, I don't have a use case to extend these either, so I will not consider these. Thank you for taking this up.

bquorning:
@pirj, @Darhazer Is this PR ok to merge?

lib/rubocop/cop/rspec/predicate_matcher.rb
```diff
@@ -149,12 +149,23 @@ def check_explicit(node) # rubocop:disable Metrics/MethodLength
...
             add_offense(node, message: message_explicit(matcher)) do |corrector|
               corrector_explicit(corrector, node, actual, matcher, matcher)
             end
           end
         end
 
+        def replaceable_matcher?(matcher)
+          case matcher.method_name.to_s
+          when 'include'
+            matcher.arguments.count <= 1
```
Darhazer, line 163: We don't have a test case with zero arguments, do we?

bquorning: I thought the same thing. Also, which implementation of include takes zero arguments?

pirj:
```suggestion
            matcher.arguments.one?
```
May be just this if there are no cases where it's used with no arguments. And `include` does not accept a block, so `expect(foo).to include` is nonsense and is probably worth a cop if anyone have seen this in practice, or there are offenders in real-world apps open source code.

pirj (approved):
Thank you, @ydah !

bquorning:
Actually this PR conflicts with 3 spec examples from #1555: https://github.com/rubocop/rubocop-rspec/blob/3a31ab146dc410ac600e36982f8290c89cc105ba/spec/rubocop/cop/rspec/predicate_matcher_spec.rb#L354-L358, https://github.com/rubocop/rubocop-rspec/blob/3a31ab146dc410ac600e36982f8290c89cc105ba/spec/rubocop/cop/rspec/predicate_matcher_spec.rb#L373-L377 and https://github.com/rubocop/rubocop-rspec/blob/3a31ab146dc410ac600e36982f8290c89cc105ba/spec/rubocop/cop/rspec/predicate_matcher_spec.rb#L392-L396

ydah:
Sorry for the late reply. I updated this PR.

bquorning (approved):

PR #1555, opened 2023-01-14
Fix an incorrect autocorrect for `RSpec/PredicateMatcher` when multiline expect and predicate method with heredoc

ydah:
This PR is fix an incorrect autocorrect for `RSpec/PredicateMatcher` when multiline expect and predicate method with heredoc.

The following code:
```ruby
expect(foo)
  .to be_something(<<~TEXT)
    bar
  TEXT
```

This would be corrected as follows:
```ruby
expect(foo.something?(<<~TEXT))
  .to be(true) <<----------- this would be included in heredoc
    bar
TEXT
```

- [x] Added tests.
- [x] Updated documentation.

### DIFF
--- a/lib/rubocop/cop/rspec/predicate_matcher.rb
+++ b/lib/rubocop/cop/rspec/predicate_matcher.rb
@@ -118,7 +118,7 @@ module RuboCop
       end
 
       # A helper for `explicit` style
-      module ExplicitHelper
+      module ExplicitHelper # rubocop:disable Metrics/ModuleLength
         include RuboCop::RSpec::Language
         extend NodePattern::Macros
 
@@ -150,9 +150,21 @@ module RuboCop
 
           predicate_matcher?(node) do |actual, matcher|
             add_offense(node, message: message_explicit(matcher)) do |corrector|
+              next if uncorrectable_matcher?(node, matcher)
+
               corrector_explicit(corrector, node, actual, matcher, matcher)
             end
           end
+        end
+
+        def uncorrectable_matcher?(node, matcher)
+          heredoc_argument?(matcher) && !same_line?(node, matcher)
+        end
+
+        def heredoc_argument?(matcher)
+          matcher.arguments.select do |arg|
+            %i[str dstr xstr].include?(arg.type)
+          end.any?(&:heredoc?)
         end
 
         # @!method predicate_matcher?(node)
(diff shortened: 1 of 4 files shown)


bquorning (approved):

lib/rubocop/cop/rspec/predicate_matcher.rb
```diff
@@ -150,11 +150,20 @@ def check_explicit(node) # rubocop:disable Metrics/MethodLength
...
             add_offense(node, message: message_explicit(matcher)) do |corrector|
+              next if uncorrectable_matcher?(node, matcher)
+
               corrector_explicit(corrector, node, actual, matcher, matcher)
             end
           end
         end
 
+        def uncorrectable_matcher?(node, matcher)
+          return false unless (arg = matcher.first_argument)
```
pirj, line 161: Does it have to be the first argument?
What about
```ruby
expect(foo.include?('one', <<~TWO))
  ...
TWO
```

ydah: Thank you! I updated this PR.

bquorning:
FYI I intend to do a v2.17.1 release when this and #1554 is merged. Followed by a v2.18.0 release with the extraction of Capybara cops (#1519).

lib/rubocop/cop/rspec/predicate_matcher.rb
```diff
@@ -150,11 +150,18 @@ def check_explicit(node) # rubocop:disable Metrics/MethodLength
...
+              next if uncorrectable_matcher?(node, matcher)
+
               corrector_explicit(corrector, node, actual, matcher, matcher)
             end
           end
         end
 
+        def uncorrectable_matcher?(node, matcher)
+          str_args = matcher.arguments.select(&:str_type?)
+          str_args.any?(&:heredoc?) && !same_line?(node, matcher)
```
Darhazer, line 162: what if the heredoc contains variables? wouldn't it be dstr_type and still contain heredoc nodes? I suggest adding dstr and xstr here

ydah: Thank you for catching this. I updated this PR.

pirj (approved):
Thank you!

bquorning:
@Darhazer Is this PR ok to merge?

Darhazer (approved):

PR #1556, opened 2023-01-14
Extract Location module as mixin

ydah:
This PR is extract Location module as mixin.

Generally usable Location-related methods are extracted as modules. This PR is a first step. Other extractable methods can be placed in this module.

We are also writing mixin tests for the first time in RuboCop RSpec. We believe this will be even better if we can check the coverage and add other mixins as well.

- [x] Added tests.
- [-] Updated documentation.


lib/rubocop/cop/rspec/mixin/location.rb
```diff
@@ -0,0 +1,37 @@
+# frozen_string_literal: true
+
+module RuboCop
```
pirj, line 3: What if we put this under `lib/rubocop/rspec` next to `Language` etc?
Then, tricks for RSpec/FilePath and `:cop` metadata inclusion won't be necessary.

lib/rubocop/cop/rspec/mixin/location.rb
```diff
@@ -0,0 +1,37 @@
...
+  module Cop
+    module RSpec
+      # Helper methods to location.
+      module Location
+        module_function
```
pirj, line 8: How is it purposed to be used? As an includable module, by calling its methods, or both ways?

Does it make sense to name it `LocationHelp` to match `RangeHelp`/`commentsHelp`?

spec/rubocop/cop/rspec/mixin/location_spec.rb
```diff
@@ -0,0 +1,76 @@
+# frozen_string_literal: true
+
+RSpec.describe RuboCop::Cop::RSpec::Location do
+  describe 'Location.arguments_with_whitespace' do
+    subject(:arguments_with_whitespace) do
+      described_class.arguments_with_whitespace(source_ast)
+    end
+
+    context 'when `a b`' do
+      let(:source_ast) { parse_source('a b').ast }
```
pirj, line 10:
```suggestion
    
    let(:source_ast) { parse_source(source).ast }

    context 'when `a b`' do
      let(:source) { 'a b' }
```

pirj (commented):
Awesome!

lib/rubocop/cop/rspec/mixin/location.rb
ydah: I propose that you do not allow Include to do so. 
We believe that this will make it easier to follow the code, and we also use it at https://github.com/rubocop/rubocop-rspec/pull/1402#discussion_r995904977 .

So I changed it to define it as a private, singular method.

Also, it seems like a good to use `LocationHelp`. Thank you! I updated this PR.

lib/rubocop/cop/rspec/mixin/location.rb
ydah: We do not want to use include modules, so we would like to keep mixin directory as much as possible. We want to keep it as a mixin as much as possible, because we want to call it `LocationHelp.some_method`.

If we move it to `lib/rubocop/rspec` We will need to call it `RuboCop::RSpec::LocationHelp.some_method`.
If I have missed something, please let me know.

pirj (approved):
Thank you!

ydah:
@bquorning @Darhazer I hope you will take a look at this as well when you have time.

lib/rubocop/cop/rspec/mixin/location_help.rb
```diff
@@ -0,0 +1,35 @@
...
+  module Cop
+    module RSpec
+      # Helper methods to location.
+      module LocationHelp
+        # @param node [RuboCop::AST::SendNode]
+        # @return [Parser::Source::Range]
+        # @example
+        #   foo 1, 2
+        #      ^^^^^
+        def self.arguments_with_whitespace(node)
```
bquorning, line 13: I think we/RuboCop normally use `module_function` on modules that aren’t meant to be included/extended anywhere:
```suggestion
        module_function

        # @param node [RuboCop::AST::SendNode]
        # @return [Parser::Source::Range]
        # @example
        #   foo 1, 2
        #      ^^^^^
        def arguments_with_whitespace(node)
```

ydah: I see, if that is the way it is with us/RuboCop, it seemed like a good to follow it.
I updated this PR.

.rubocop.yml
```diff
@@ -106,6 +106,10 @@ Style/FormatStringToken:
 Style/RequireOrder:
   Enabled: true
 
+RSpec/FilePath:
+  Exclude:
+    - spec/rubocop/cop/rspec/mixin/**/*.rb
```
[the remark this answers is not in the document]
ydah: I agree. BTW, RuboCop seems to disable `RSpec/FilePath` in the disable comment.
Which is better, namespace change or 🤔?

https://github.com/rubocop/rubocop/blob/cf2d98b89c366886f4f2bfd88efabf6810fe9cc1/spec/rubocop/cop/mixin/enforce_superclass_spec.rb#L3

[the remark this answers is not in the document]
bquorning: I think disable for now, then namespace change in a later PR.

spec/spec_helper.rb
```diff
@@ -18,7 +18,7 @@ module SpecHelper
 RSpec.configure do |config|
   # Set metadata so smoke tests are run on all cop specs
   config.define_derived_metadata(file_path: %r{/spec/rubocop/cop/}) do |meta|
```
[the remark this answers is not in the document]
ydah: Thank you. I updated this PR.

.rubocop.yml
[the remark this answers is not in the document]
ydah: That sounds like good. Thank you. So we would like to leave it disabled in this PR.

lib/rubocop/cop/rspec/mixin/location.rb
bquorning: Oh, so you originally had `module_function` before @pirj convinced you to use `def self.foo`, after which I said what we should use `module_function` 😅

The downside of `module_function` and `extend(self)` is that the module’s methods can still be included into a class (as private vs public methods). At least I think that’s a downside.

I see that the Ruby source uses `module_function` a lot, and `extend self` quite little. It’s hard to tell how often they define a module only with class methods.

bquorning (approved):

lib/rubocop/cop/rspec/mixin/location.rb
ydah: >Oh, so you originally had module_function before @pirj convinced you to use def self.foo, after which I said what we should use module_function 😅

Yes, it is. But I thought if we as RuboCop have a flow that we basically use `module_function` we had better choose to do that. @pirj I would like to get your opinion.

pirj: My rule of thumb is:
1. Use `module_function` for double-purpose modules that may be included or used as a standalone helper, and when included should not expose their methods
2. Use `extend self` for double-purpose modules that may be included or used as a standalone helper and would expose their methods when included
3. Use `class << self` for modules purposed to be used as helper that, or that are complex enough to have private methods not intended to be exposed in the module's API

In our case, I see no reason why we'd want `extend self` or `module_function` over `class << self`, but this is only a purist nitpick, and I don't insist on that at all.

pirj: More on this https://github.com/rubocop/ruby-style-guide/issues/556

bquorning: > a purist nitpick

That could be the slogan of our club 😆 

I like your rules of thumb. It seems silly that RuboCop offers to enforce one style over the others when they clearly are not interchangeable.

PR #1557, opened 2023-01-16
Fix a false positive for `RSpec/ContextWording` when context is interpolated string literal or execute string

ydah:
This PR is fix a false positive for `RSpec/ContextWording` when context is interpolated string literal or execute string.

- [x] Added tests.
- [-] Updated documentation.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Master (Unreleased)
 
+- Fix a false positive for `RSpec/ContextWording` when context is interpolated string literal or execute string. ([@ydah])
+
 ## 2.18.1 (2023-01-19)
 
 - Add `rubocop-capybara` version constraint to prevent sudden cop enabling when it hits 3.0. ([@pirj])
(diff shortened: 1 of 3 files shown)


pirj (approved):
Thank you!

bquorning (approved):
Since this PR does not fix a problem introduced in 2.17.0, would it ok to wait with merging until 2.17.1 and 2.18.0 have been released?

ydah:

AlexWayfer:
FYI: This change helped to catch `context "has #{} something"`, but makes unable to write something like `context "#{variable ? 'with' : 'without'} something"`.

bquorning:
> `context "#{variable ? 'with' : 'without'} something"`

I think a `rubocop:disable` is appropriate for this specific case.

AlexWayfer:
> > `context "#{variable ? 'with' : 'without'} something"`
> 
> I think a `rubocop:disable` is appropriate for this specific case.

Yeah, I thought about it. Thanks. I've changed the string completely. So, it was as "FYI", not an issue. You can support such things or you can leave it.

bquorning:
I think we could only support it by `eval`ing the interpolated string, or by matching it against some regex. Either way, it would be a mess, and would probably still not work properly 😅

AlexWayfer:
> I think we could only support it by `eval`ing the interpolated string, or by matching it against some regex. Either way, it would be a mess, and would probably still not work properly sweat_smile

I believe so, so it's not even a purpose, just a note, maybe for others who will find this, or for contributors who like challenges or have great ideas. :sweat_smile:

PR #1558, opened 2023-01-16
Fix a false positive for `RSpec/DescribeMethod` when multi-line describe without `#` or `.` at the beginning

ydah:
This PR is fix a false positive for `RSpec/DescribeMethod` when multi-line describe without `#` and `.` at the beginning.

- [x] Added tests.
- [-] Updated documentation.

### DIFF
--- a/lib/rubocop/cop/rspec/describe_method.rb
+++ b/lib/rubocop/cop/rspec/describe_method.rb
@@ -23,27 +23,27 @@ module RuboCop
         MSG = 'The second argument to describe should be the method ' \
               "being tested. '#instance' or '.class'."
 
-        # @!method second_argument(node)
-        def_node_matcher :second_argument, <<~PATTERN
+        # @!method second_string_literal_argument(node)
+        def_node_matcher :second_string_literal_argument, <<~PATTERN
           (block
-            (send #rspec? :describe _first_argument $_ ...)
+            (send #rspec? :describe _first_argument ${str dstr} ...)
           ...)
         PATTERN
 
-        # @!method not_method_name?(node)
-        def_node_matcher :not_method_name?, <<~PATTERN
-          {(str !#method_name?) (dstr (str !#method_name?) ...)}
+        # @!method method_name?(node)
+        def_node_matcher :method_name?, <<~PATTERN
+          {(str #method_name_prefix?) (dstr (str #method_name_prefix?) ...)}
         PATTERN
 
         def on_top_level_group(node)
-          second_argument(node) do |argument|
-            add_offense(argument) if not_method_name?(argument)
+          second_string_literal_argument(node) do |argument|
+            add_offense(argument) unless method_name?(argument)
           end
         end
 
         private
 
-        def method_name?(description)
+        def method_name_prefix?(description)
           description.start_with?('.', '#')
         end
       end
--- a/lib/rubocop/cop/rspec/describe_method.rb
+++ b/lib/rubocop/cop/rspec/describe_method.rb
@@ -26,17 +26,25 @@ module RuboCop
         # @!method second_argument(node)
         def_node_matcher :second_argument, <<~PATTERN
           (block
-            (send #rspec? :describe _first_argument $(str _) ...) ...
-          )
+            (send #rspec? :describe _first_argument $_ ...)
+          ...)
+        PATTERN
+
+        # @!method not_method_name?(node)
+        def_node_matcher :not_method_name?, <<~PATTERN
+          {(str !#method_name?) (dstr (str !#method_name?) ...)}
         PATTERN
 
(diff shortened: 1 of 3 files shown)


pirj (approved):
Fantastic, thank you!
Does it make sense to add an example where a multi-line docstring that begins with a method name is ignored?

ydah:
I updated this PR.

ydah:

pirj:
There's still a conflict, probably due to the recent 2.18.1 changelog change.

lib/rubocop/cop/rspec/describe_method.rb
```diff
@@ -23,20 +23,25 @@ class DescribeMethod < Base
...
+            add_offense(describe)
+          end
+        end
 
-          return unless second_argument
-          return if second_argument.str_content.start_with?('#', '.')
+        private
 
-          add_offense(second_argument)
+        def method_name?(description)
```
pirj, line 43: This seems to do the opposite, like "not_a_method_name".

lib/rubocop/cop/rspec/describe_method.rb
```diff
@@ -23,20 +23,25 @@ class DescribeMethod < Base
         MSG = 'The second argument to describe should be the method ' \
               "being tested. '#instance' or '.class'."
 
-        # @!method second_argument(node)
-        def_node_matcher :second_argument, <<~PATTERN
+        # @!method describe_method(node)
+        def_node_matcher :describe_method, <<~PATTERN
```
pirj, line 27: `describe_method` seems to be more like "describe with a second argument that does not resemble a method name", quite the opposite of its name.

lib/rubocop/cop/rspec/describe_method.rb
ydah: Oops, thanks for catching this! I refactored and organized it. What do you think?

lib/rubocop/cop/rspec/describe_method.rb
ydah: Thanks for catching this one too. I refactored it.

lib/rubocop/cop/rspec/describe_method.rb
```diff
@@ -26,17 +26,25 @@ class DescribeMethod < Base
...
           (block
-            (send #rspec? :describe _first_argument $(str _) ...) ...
-          )
+            (send #rspec? :describe _first_argument $_ ...)
+          ...)
+        PATTERN
+
+        # @!method not_method_name?(node)
+        def_node_matcher :not_method_name?, <<~PATTERN
+          {(str !#method_name?) (dstr (str !#method_name?) ...)}
```
pirj: I usually find it hard to understand boolean logic with negations.
Here it's like "if it's not a string prefixed with # OR not an interpolated string prefixed with #".
And it makes you think - what if it's neither? Like a literal `1`.
And just by looking at the code it's not obvious that we tolerate everything that is not a str/dstr.

I dared pushing a commit making this explicit and removes negation from boolean logic involved. Leaving it up to you if you like it or not.

ydah: Certainly what you have suggested would be better. Let's go with that. Thank you very much.

ydah: @pirj I've added you to the CHANGELOG as you have refactored many parts of this PR. Thank you!

lib/rubocop/cop/rspec/describe_method.rb
```diff
@@ -23,20 +23,28 @@ class DescribeMethod < Base
...
+        def_node_matcher :second_string_literal_argument, <<~PATTERN
           (block
-            (send #rspec? :describe _first_argument $(str _) ...) ...
-          )
+            (send #rspec? :describe _first_argument ${str dstr} ...)
+          ...)
+        PATTERN
+
+        # @!method method_name?(node)
+        def_node_matcher :method_name?, <<~PATTERN
```
bquorning, line 34: What advantages does this node matcher have over a regular method ? E.g.
```
def method_name?(str)
  str.start_with?('.', '#')
end
```

pirj: The argument is a node, and if it’s a dstr, I’m uncertain if it has a ‘strvalue’ similar to the str type. But it’s worth checking out.

pirj: This worked:
```ruby
        def method_name?(node)
          if node.str_type?
            method_name_prefix?(node.str_content)
          else # dstr
            method_name_prefix?(node.children.first.str_content)
          end
        end
```
Node pattern seems more concise.

bquorning (approved):

Issue #1563, opened 2023-01-17
Idea: disallow passing a block to `skip`/`pending` inside examples


pirj:
See #1562 
```ruby
it "works" do
  # ...

  skip 'not yet implemented' do
    # do some checks
  end
end
```

PR #1564, opened 2023-01-17
Add new `RSpec/PendingBlockInsideExample` cop

ydah:
Resolve: https://github.com/rubocop/rubocop-rspec/issues/1563

- [x] Added tests.
- [x] Updated documentation.


docs/modules/ROOT/pages/cops_rspec.adoc
```diff
@@ -3839,6 +3839,63 @@ end
...
+
+=== Examples
+
+==== EnforcedStyle: bar (default)
+
+[source,ruby]
+----
+# Description of the `bar` style.
+
+# bad
```
pirj, line 3864:
```suggestion
# bad
```

lib/rubocop/cop/rspec/pending_block_inside_example.rb
```diff
@@ -0,0 +1,65 @@
...
+
+module RuboCop
+  module Cop
+    module RSpec
+      # Checks for passing a block by `pending` or `skip` within examples.
+      #
+      # @example EnforcedStyle: bar (default)
+      #   # Description of the `bar` style.
+      #
+      #   # bad
```
pirj, line 11:
```suggestion
      #   # bad
```

lib/rubocop/cop/rspec/pending_block_inside_example.rb
```diff
@@ -0,0 +1,65 @@
...
+      #     skip 'not yet implemented'
+      #     do_something
+      #   end
+      #
+      #   # good - when outside example
+      #   pending 'not yet implemented' do
+      #   end
+      #
+      class PendingBlockInsideExample < Base
+        MSG = 'Remove passing a block by `pending` or `skip` within examples.'
```
pirj, line 42:
```suggestion
        MSG = "Don't pass a block to `pending` or `skip` inside examples."
```

pirj (approved):
That was blazing fast, @ydah ! Thank you

lib/rubocop/cop/rspec/pending_block_inside_example.rb
```diff
@@ -0,0 +1,63 @@
+# frozen_string_literal: true
+
+module RuboCop
+  module Cop
+    module RSpec
+      # Checks for passing a block by `pending` or `skip` within examples.
+      #
+      # @example EnforcedStyle: bar (default)
```
pirj, line 8:
```suggestion
      # @example
```

lib/rubocop/cop/rspec/pending_block_inside_example.rb
```diff
@@ -0,0 +1,63 @@
...
+
+        def inside_example?(node)
+          Examples.all(first_left_sibling(node)&.method_name)
+        end
+
+        def first_left_sibling(node)
+          return unless (block_node = node.block_node)
+
+          block_node.left_siblings.first
+        end
```
pirj, line 59:
```suggestion
        def on_block(node)
          return unless inside_example?(node)

          add_offense(node) if %i[pending skip].include?(node.method_name)
        end

        private

        def inside_example?(node)
          node.each_ancestor(:block).any? { |ancestor| example?(ancestor) }
        end
```

Suggesting a bigger change to:
 - reduce some complexity (I couldn't quickly grasp the left sibling method)
 - only check block nodes

A viable alternative is to keep the send check restricted to `skip`/`pending` names, and use `node.parent.each_ancestor` in `inside_example?`, should also work. Leaving the decision up to you.

ydah: I think your suggestion is certainly very good! 
Let's go with your suggestion. Thank you very much.

ydah:

lib/rubocop/cop/rspec/pending_block_inside_example.rb
```diff
@@ -0,0 +1,56 @@
...
+      #   pending 'not yet implemented' do
+      #   end
+      #
+      class PendingBlockInsideExample < Base
+        MSG = "Don't pass a block to `pending` or `skip` inside examples."
+
+        def on_block(node) # rubocop:disable InternalAffairs/NumblockHandler
+          return unless inside_example?(node)
+
+          add_offense(node) if %i[pending skip].include?(node.method_name)
```
pirj, line 45: On an afterthought, this may be more performant:
```suggestion
          return unless %i[pending skip].include?(node.method_name)
          return unless inside_example?(node)

          add_offense(node)
```

ydah: Thank you! I updated this PR.

pirj:
I'm puzzled why:
```
    it 'registers an offense when using `pending` with a numblock' do
      expect_offense(<<~RUBY)
        it "does something" do
          pending 'not yet implemented' do
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ Don't pass a block to `pending` or `skip` inside examples.
            _1
          end
        end
      RUBY
    end
```
doesn't parse `pending` as a `numblock` 🤔 
```
> node
=> s(:block,
  s(:send, nil, :pending,
    s(:str, "not yet implemented")),
  s(:args),
  s(:send, nil, :_1))
> puts node.source
pending 'not yet implemented' do
    _1
  end
```

Updated to the latest `rubocop` 1.43, `rubocop-ast` 1.24.1, `parser` 3.2.0.0.

```
$ ruby-parse -e 'r(2) do; _1 ; end'
(numblock
  (send nil :r
    (int 2)) 1
  (lvar :_1))
```
Am I missing something?

pirj:
Filed https://github.com/rubocop/rubocop/issues/11479

I suggest holding this off, as we disable the offence that insists on adding `on_numblock` handler, and can't really test our code against numblocks with the above issue.

pirj:
Alright, it slipped my mind that we have to add `:ruby27` for examples that use numblocks.

bquorning (approved):

Issue #1562, opened 2023-01-17
False positive for RSpec/PendingWithoutReason with block syntax


tomchipchase:
We've got one last false positive with this new cop:

When the skip has a block, it gets incorrectly flagged:
```
it "works" do
  # do some stuff

  skip 'not yet implemented' do
    # do some checks
  end
end
```

This works fine though (its not quite the same thing if there was some other stuff after the skip block, but fine for our usecase).
```
it "works" do
  # do some stuff

  skip 'not yet implemented'
  # do some checks
end
```

pirj:
Passing a block to `skip`/`pending` seems to be a misuse, as [they don't accept blocks](https://github.com/rspec/rspec-core/blob/79ec0fd5b5ebd13031a10960f293b8a4a43cb734/lib/rspec/core/pending.rb#L91) ([docs](https://relishapp.com/rspec/rspec-core/v/3-9/docs/pending-and-skipped-examples/skip-examples)).

> its not quite the same thing if there was some other stuff after the skip block

Can you give an example when it's not the same thing?

`skip` stops example execution immediately. The presence of code after the `skip` statement doesn't matter.
`pending` doesn't stop, but its purpose is to make sure that the example fails, so it would be weird to skip anything in a block that never gets called.

I'd prefer to keep flagging this case, even though it's not this cop's responsibility, but rather an idea for some other cop.

tomchipchase:
Fair enough! We were misusing this feature then.

pirj:
Provided there is a cop that is watching over in-example ‘pending’/‘skip’ with blocks, I’m fine with fixing this false positive in ‘PendingWiyhoutReason’ that additionally avoids duplicate offences, and to completely split responsibilities.

pirj:
Fixed in #1566

PR #1572, opened 2023-02-05
Rename and narrow to inspect `skip` only

pirj:
We are not expected to flag `pending`, as it would blow up specs:

 - https://github.com/rubocop/rubocop-rspec/pull/1566#discussion_r1088923352
 - https://github.com/rubocop/rubocop-rspec/pull/1566#discussion_r1089015507
 - https://github.com/rspec/rspec-core/blob/79ec0fd5b5ebd13031a10960f293b8a4a43cb734/lib/rspec/core/pending.rb#L66

See also #1562 #1564

- [x] Added tests.
- [x] Updated documentation.


bquorning:
Just to clarify: [RSpec will raise an ArgumentError when you call `pending` with a block](https://github.com/rspec/rspec-core/blob/79ec0fd5b5ebd13031a10960f293b8a4a43cb734/lib/rspec/core/pending.rb#L66) – so there is no point is us having a cop that checks if you are calling `pending` with a block.

bquorning (approved):

PR #1575, opened 2023-02-07
Refactor and adding a common matcher in SkipOrPending module

ydah:
This PR is refactor and adding a common matcher in SkipOrPending module.

- [-] Added tests.
- [-] Updated documentation.

### DIFF
--- a/lib/rubocop/cop/rspec/mixin/skip_or_pending.rb
+++ b/lib/rubocop/cop/rspec/mixin/skip_or_pending.rb
@@ -10,13 +10,29 @@ module RuboCop
         # @!method skipped_in_metadata?(node)
         def_node_matcher :skipped_in_metadata?, <<-PATTERN
           {
-            (send _ _ <#skip_or_pending? ...>)
-            (send _ _ ... (hash <(pair #skip_or_pending? { true str dstr }) ...>))
+            (send _ _ <(sym {:skip :pending}) ...>)
+            (send _ _ ... (hash <(pair (sym {:skip :pending}) { true str dstr }) ...>))
           }
         PATTERN
 
-        # @!method skip_or_pending?(node)
-        def_node_matcher :skip_or_pending?, '{(sym :skip) (sym :pending)}'
+        # @!method skip_or_pending_inside_block?(node)
+        #   Match skip/pending statements inside a block (e.g. `context`)
+        #
+        #   @example source that matches
+        #     context 'when color is blue' do
+        #       skip 'not implemented yet'
+        #       pending 'not implemented yet'
+        #     end
+        #
+        #   @example source that does not match
+        #     skip 'not implemented yet'
+        #     pending 'not implemented yet'
+        #
+        #   @param node [RuboCop::AST::Node]
+        #   @return [Array<RuboCop::AST::Node>] matching nodes
+        def_node_matcher :skip_or_pending_inside_block?, <<-PATTERN
+          (block <(send nil? {:skip :pending} ...) ...>)
+        PATTERN
       end
     end
   end
--- a/lib/rubocop/cop/rspec/repeated_example_group_body.rb
+++ b/lib/rubocop/cop/rspec/repeated_example_group_body.rb
@@ -43,6 +43,8 @@ module RuboCop
       #   end
       #
       class RepeatedExampleGroupBody < Base
+        include SkipOrPending
+
         MSG = 'Repeated %<group>s block body on line(s) %<loc>s'
 
         # @!method several_example_groups?(node)
@@ -59,11 +61,6 @@ module RuboCop
         # @!method const_arg(node)
         def_node_matcher :const_arg, '(block (send _ _ $const ...) ...)'
 
-        # @!method skip_or_pending?(node)
-        def_node_matcher :skip_or_pending?, <<-PATTERN
-          (block <(send nil? {:skip :pending} ...) ...>)
-        PATTERN
-
         def on_begin(node)
           return unless several_example_groups?(node)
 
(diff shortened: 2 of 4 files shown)


lib/rubocop/cop/rspec/mixin/skip_or_pending.rb
```diff
@@ -10,13 +10,18 @@ module SkipOrPending
...
+            (send _ _ ... (hash <(pair #skip_or_pending_sym? { true str dstr }) ...>))
           }
         PATTERN
 
-        # @!method skip_or_pending?(node)
-        def_node_matcher :skip_or_pending?, '{(sym :skip) (sym :pending)}'
+        # @!method skip_or_pending_sym?(node)
+        def_node_matcher :skip_or_pending_sym?, '{(sym :skip) (sym :pending)}'
+
+        # @!method skip_or_pending_block?(node)
```
pirj, line 21: Nice! It's probably worth mentioning in the method docs that it matches exa﻿﻿ctly, as this is non-trivial.

Example:
```ruby
        # @!method examples_inside_block?(node)
        #   Match examples defined inside a block which is not a hook
        #
        #   @example source that matches
        #     %w(r g b).each do |color|
        #       it { is_expected.to have_color(color) }
        #     end
        #
        #   @example source that does not match
        #     before do
        #       it { is_expected.to fall_into_oblivion }
        #     end
        #
        #   @param node [RuboCop::AST::Node]
        def_node_matcher :examples_inside_block?, <<~PATTERN
          (block !#{send_pattern('#Hooks.all')} _ #examples?)
        PATTERN
```

Basing on just the name I have some questions:
1. Would it match a block form of `skip` call? Probably not, right?
```
$ ruby-parse -e 'it { skip { } }'
(block
  (send nil :it)
  (args)
  (block
    (send nil :skip)
    (args) nil))
```
2. Would it really match multi-statement blocks with a `begin`? Seems unlikely, as `<` syntax only matches elements following `block`, not nested
```
$ ruby-parse -e 'd { 1; skip }'
(block
  (send nil :d)
  (args)
  (begin
    (int 1)
    (send nil :skip)))
```

pirj (approved):
👍 and thank you for the refactoring, but now when I look at the `skip_or_pending?` node matcher, I struggle to understand what it does exactly.

lib/rubocop/cop/rspec/mixin/skip_or_pending.rb
ydah: Sorry for the late. I updated this PR.

lib/rubocop/cop/rspec/mixin/skip_or_pending.rb
```diff
@@ -10,13 +10,33 @@ module SkipOrPending
...
           }
         PATTERN
 
-        # @!method skip_or_pending?(node)
-        def_node_matcher :skip_or_pending?, '{(sym :skip) (sym :pending)}'
+        # @!method skip_or_pending_sym?(node)
+        def_node_matcher :skip_or_pending_sym?, '{(sym :skip) (sym :pending)}'
+
+        # @!method skip_or_pending_inside_block?(node)
+        #   Match examples defined inside a block which is not a hook
```
pirj, line 22: > examples defined inside a block

> which is not a hook

Are those correct?
It appears to me that this matcher is more like "Match skip/pending statements inside a block".

lib/rubocop/cop/rspec/mixin/skip_or_pending.rb
```diff
@@ -10,13 +10,33 @@ module SkipOrPending
...
+        #
+        #   @example source that matches
+        #     context 'when color is blue' do
+        #       skip 'not implemented yet'
+        #     end
+        #
+        #   @example source that does not match
+        #     pending 'not implemented yet'
+        #     it 'does not match' do
+        #       skip { }
```
pirj, line 32: Why do we ignore this? Even though we have `RSpec/SkipBlockInsideExample` cop, it may be turned off and this syntax won't cause RSpec to blow up.

Unrelated to this refactoring, an idea for a follow-up improvement.

lib/rubocop/cop/rspec/mixin/skip_or_pending.rb
```diff
@@ -10,13 +10,33 @@ module SkipOrPending
         # @!method skipped_in_metadata?(node)
         def_node_matcher :skipped_in_metadata?, <<-PATTERN
           {
-            (send _ _ <#skip_or_pending? ...>)
-            (send _ _ ... (hash <(pair #skip_or_pending? { true str dstr }) ...>))
+            (send _ _ <#skip_or_pending_sym? ...>)
+            (send _ _ ... (hash <(pair #skip_or_pending_sym? { true str dstr }) ...>))
```
pirj, line 14:
```suggestion
            (send _ _ <(sym {:skip :pending}) ...>)
            (send _ _ ... (hash <(pair (sym {:skip :pending}) { true str dstr }) ...>))
```

lib/rubocop/cop/rspec/mixin/skip_or_pending.rb
```diff
@@ -10,13 +10,33 @@ module SkipOrPending
...
           }
         PATTERN
 
-        # @!method skip_or_pending?(node)
-        def_node_matcher :skip_or_pending?, '{(sym :skip) (sym :pending)}'
+        # @!method skip_or_pending_sym?(node)
+        def_node_matcher :skip_or_pending_sym?, '{(sym :skip) (sym :pending)}'
+
```
pirj, line 20:
```suggestion
```

lib/rubocop/cop/rspec/mixin/skip_or_pending.rb
ydah: Surely that is a wrong. I updated it.

ydah:
Sorry for the delay. I have updated this PR.
And I have also add a test for mixin.

pirj (approved):
Looks great, thank you!

bquorning (approved):

PR #1576, opened 2023-02-07
Replace where send_pattern and block_pattern can be used

ydah:
This PR is replace where send_pattern and block_pattern can be used.

- [-] Added tests.
- [-] Updated documentation.

### DIFF
--- a/lib/rubocop/cop/rspec/capybara/feature_methods.rb
+++ b/lib/rubocop/cop/rspec/capybara/feature_methods.rb
@@ -63,11 +63,8 @@ module RuboCop
           PATTERN
 
           # @!method feature_method(node)
-          def_node_matcher :feature_method, <<-PATTERN
-            (block
-              $(send #rspec? $#capybara_speak ...)
-            ...)
-          PATTERN
+          def_node_matcher :feature_method,
+                           "(block $#{send_pattern('$#capybara_speak')} ...)"
 
           def on_block(node) # rubocop:disable InternalAffairs/NumblockHandler
             return unless inside_example_group?(node)
--- a/lib/rubocop/cop/rspec/context_method.rb
+++ b/lib/rubocop/cop/rspec/context_method.rb
@@ -30,12 +30,8 @@ module RuboCop
         MSG = 'Use `describe` for testing methods.'
 
         # @!method context_method(node)
-        def_node_matcher :context_method, <<-PATTERN
-          (block
-            (send #rspec? :context
-              ${(str #method_name?) (dstr (str #method_name?) ...)}
-            ...)
-          ...)
+        def_node_matcher :context_method, block_pattern(<<-PATTERN)
+          :context ${(str #method_name?) (dstr (str #method_name?) ...)}
         PATTERN
 
         def on_block(node)  # rubocop:disable InternalAffairs/NumblockHandler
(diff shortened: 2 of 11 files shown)


pirj:
If I recall it correctly, we've introduced `*_pattern` methods as a replacement to the old way to programmatically define send/block for dynamic DSL configuration:
```diff
- def_node_matcher :shared_context, SharedGroups::CONTEXT.block_pattern
+ def_node_matcher :shared_context, block_pattern('#SharedGroups.context')
- def_node_search :examples?, (Includes::EXAMPLES + Examples::ALL).send_pattern
+ def_node_search :examples?, send_pattern('{#Includes.examples #Examples.all}')
- def_node_search :find_rspec_blocks, ExampleGroups::ALL.block_pattern
+ def_node_search :find_rspec_blocks, block_pattern('#ExampleGroups.all')
```

There is just no other way to match the dynamic DSL configuration, and we've introduced it knowing the tradeoff that it adds complexity to how our node patterns look like.

Just looking at our current codebase, we don't seem to use them for anything more complicated than `block_pattern('{#ExampleGroups.all #Examples.all}')`.

Adding another layer of abstraction to node pattern, keeping in mind that its syntax is already quite hard to grasp is not what we want in general.

Even though it helps to dry node patterns, at the same time one reading them would have to pass a mental exercise of expanding those `_pattern` calls to be able to visually match with AST.

There's also a huge change it may be mis-used, as the `#rspec?` part. The receiver is defined as `{#explicit_rspec? nil?}`. E.g. `send_pattern(':foo')` can be used to match any `foo` call, while it's purposed to match `RSpec.foo { }`/`foo { }` where RSpec is providing this `foo` as part of its DSL.


With the above criteria in mind of exclusively using dynamic DSL config helpers, I would have no objections against this change:
```diff
- (send #rspec? #Helpers.all _ block_pass)
+ #{send_pattern('#Helpers.all _ block_pass')}
```
other than it doesn't help with readability of the node pattern.

I'm open to discuss this, @bquorning @Darhazer if you think otherwise.

ydah:
Thank you for giving me some background on the implementation of these methods. I understand.

How about reopening it with only the following changes?

```diff
- (send #rspec? #Helpers.all _ block_pass)
+ #{send_pattern('#Helpers.all _ block_pass')}
```

pirj:
Please accept my apologies for your time, @ydah. We never came to describing when exactly those `*_pattern` should be used, and even if we ever thought of our plans for them beyond how we used them initially.

I'd keep the small change you mention out, too, just because it's a bit longer, has additional level of indirection in the form of the string interpolation. I agree with the consistency argument, but frankly I'm on the fence even with the line above that uses the `block_pattern` over a more explicit version:
```diff
- #{block_pattern('#Helpers.all')}
+ (block (send #rspec? #Helpers.all ...) ...)
```
Especially since this implies it would match a weird `RSpec.let(:foo) { 'bar' }`, so this should better be `nil?`, not `#rspec?`.

ydah:
It is good to know this through this PR. @pirj Thank you for spending your valuable time to explain this to us, it is very helpful.

>Especially since this implies it would match a weird RSpec.let(:foo) { 'bar' }, so this should better be nil?, not #rspec?.

Ah! That's true😇

PR #1577, opened 2023-02-07
Use `line_with_comment?` instead of bad API

ydah:
`RuboCop::Cop::Util#comment_line?` is flagged by https://github.com/rubocop/rubocop/pull/8405 as a bad API. So I propose to fix not use this method.

- [-] Added tests.
- [-] Updated documentation.


bquorning (approved):
Do you know if `processed_source.line_with_comment?` is available only in recent versions of rubocop / rubocop_ast?

PR #1578, opened 2023-02-08
Change to ignore `example?`, `hook?`, `let?`, `include?` and `subject?` matchers with `RSpec` receivers

ydah:
Follow up: https://github.com/rubocop/rubocop-rspec/pull/1576#issuecomment-1421499150

- [-] Added tests.
- [-] Updated documentation.

### DIFF
--- a/lib/rubocop/rspec/language.rb
+++ b/lib/rubocop/rspec/language.rb
@@ -44,30 +44,34 @@ module RuboCop
       PATTERN
 
       # @!method example?(node)
-      def_node_matcher :example?, block_pattern('#Examples.all')
+      def_node_matcher :example?, '(block (send nil? #Examples.all ...) ...)'
 
       # @!method hook?(node)
-      def_node_matcher :hook?,
-                       block_or_numblock_pattern('#Hooks.all')
+      def_node_matcher :hook?, <<-PATTERN
+        {
+          (numblock (send nil? #Hooks.all ...) ...)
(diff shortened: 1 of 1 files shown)


pirj:
I think the same applies to `hook?`/`example?`/`subject?`/`include?`. Would you like to take care of all that in one go?

ydah:
> I think the same applies to `hook?`/`example?`/`subject?`/`include?`. Would you like to take care of all that in one go?

Indeed it is. I updated this PR. Thank you.

lib/rubocop/rspec/language.rb
```diff
@@ -44,30 +44,34 @@ class << self
...
+      def_node_matcher :example?, '(block (send nil? #Examples.all ...) ...)'
 
       # @!method hook?(node)
-      def_node_matcher :hook?,
-                       block_or_numblock_pattern('#Hooks.all')
+      def_node_matcher :hook?, <<-PATTERN
+      {
+        (numblock (send nil? #Hooks.all ...) ...)
+        (block (send nil? #Hooks.all ...) ...)
+      }
```
pirj, line 54:
```suggestion
        {
          (numblock (send nil? #Hooks.all ...) ...)
          (block (send nil? #Hooks.all ...) ...)
        }
```

pirj (approved):
Thank you! Personally, those node patterns look less cryptic to me now 🤩

PR #1579, opened 2023-02-08
Enable some new cops

ydah:
I would like to enable these cop following https://github.com/rubocop/rubocop-rspec/pull/1410 .
One cop is correcting offenses and it looks good. What do you think?

- [-] Added tests.
- [-] Updated documentation.


pirj (approved):
Thanks!

PR #1584, opened 2023-02-16
Mark `@api private` to internal APIs

ydah:
Follow up: https://github.com/rubocop/rubocop-minitest/commit/4b5c9d641047da7feef8098e5f3f4d82a30b8c17

- [-] Added tests.
- [-] Updated documentation.

### DIFF
--- a/lib/rubocop/cop/rspec/mixin/top_level_group.rb
+++ b/lib/rubocop/cop/rspec/mixin/top_level_group.rb
@@ -4,6 +4,7 @@ module RuboCop
   module Cop
     module RSpec
       # Helper methods for top level example group cops
+      # @api private
       module TopLevelGroup
         extend RuboCop::NodePattern::Macros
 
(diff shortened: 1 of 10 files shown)


lib/rubocop/cop/rspec/mixin/top_level_group.rb
```diff
@@ -4,6 +4,7 @@ module RuboCop
   module Cop
     module RSpec
       # Helper methods for top level example group cops
+      # @api private
       module TopLevelGroup
```
pirj, line 8: I can argue that this one is not private - it's just not documented. We mention it [here](https://github.com/rubocop/rubocop-rspec/blob/808ce138ca1a69e13fb33f0eff097c5a614b1b98/docs/modules/ROOT/pages/upgrade_to_version_2.adoc#replace-topleveldescribe) to be used on custom RSpec cops.

I can't tell which other mixins/helpers are really private, or can be used by other projects with no fear of them being removed in a minor release. But I'm on the fence of marking them all as private.

ydah: Oh, thank you. I had overlooked that document. And it is true that these mixins may not as private.

ydah:
I have found it difficult to proceed with this and would like to close this PR. @pirj Thank you for taking your valuable time to review this.

PR #1592, opened 2023-03-07
Fix a false positive for `RSpec/MatchArray` when `match_array` with an empty array

ydah:
Fix: #1591

- [x] Added tests.
- [-] Updated documentation.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Master (Unreleased)
 
+- Fix a false positive for `RSpec/MatchArray` when `match_array` with an empty array. ([@ydah])
+
 ## 2.19.0 (2023-03-06)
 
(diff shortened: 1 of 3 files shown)


bquorning:
I wonder if we should change `match_array([])` into `eq([])`, of if that is too much of a change?

ydah:
> I wonder if we should change `match_array([])` into `eq([])`, of if that is too much of a change?

I would suggest letting the next cop make that change. What do you think?
- https://github.com/rubocop/rubocop-rspec/pull/1594

CKolkey:
For what it's worth, when accounting for this cop yesterday, the change I made was from `match_array([])` to `eq([])`.

ydah:
We have determined that the https://github.com/rubocop/rubocop-rspec/pull/1597 way is best and will close this.

PR #1593, opened 2023-03-07
Add support for `RSpec/ContainExactly` when calling `match_array` with no arguments

ydah:
When calling `contain_exactly` with no arguments, you might as well use the simpler eq matcher with a [] argument.

Same fixing: https://github.com/rubocop/rubocop-rspec/pull/1597

- [x] Added tests.
- [-] Updated documentation.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## Master (Unreleased)
 
 - Fix a false positive for `RSpec/PendingWithoutReason` when pending/skip has a reason inside an example group. ([@ydah])
+- Add support for `RSpec/ContainExactly` when calling `match_array` with no arguments. ([@ydah])
 
 ## 2.19.0 (2023-03-06)
 
--- a/config/default.yml
+++ b/config/default.yml
@@ -202,7 +202,7 @@ RSpec/ClassCheck:
   Reference: https://www.rubydoc.info/gems/rubocop-rspec/RuboCop/Cop/RSpec/ClassCheck
 
 RSpec/ContainExactly:
-  Description: Prefer `match_array` when matching array values.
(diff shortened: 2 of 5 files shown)


ydah:
@bquorning 
I updated this PR. This PR follows https://github.com/rubocop/rubocop-rspec/pull/1597 to support the contain_exactly case.

bquorning:
In `match_array([])` we are certain about the Array type. But I wonder if `contain_exactly()` may also match e.g. `nil` or `{}`?

ydah:
Oh, that's right. It only needs to be a collection, so `expect({}).to contain_exactly` is possible.
If `nil`, an Error is returned.

```
Failure/Error: expect(nil).to contain_exactly
       expected a collection that can be converted to an array with `#to_ary` or `#to_a`, but got nil
```

bquorning:
Do you think we should autocorrect to `be_empty` instead?

ydah:
I'm not sure if it is best to do an autocorrection to `be_empty` as it would reduce the information.
As discussed here: https://github.com/rubocop/rubocop-rspec/issues/244#issuecomment-522991027

bquorning (approved):

bquorning:
> In any case, it's quite odd that we suggest `be_empty` in one case, and `eq([])` in other as a replacement for mutually interchangeable `contains_exactly`/`match_array`.

The thinking was that for `contains_exactly()` we really have no idea about the type being matched, but we know that it is an empty collection. for `match_array([])` we do know the type, so we can use the equality matcher.

But thinking the whole thing over again, we probably should go with #1594 instead, and just skip auto-correction of no argument in RSpec/ContainExactly and empty array in RSpec/MatchArray.

bquorning:
@pirj and @ydah, please see #1604.

PR #1594, opened 2023-03-07
Add new `RSpec/BeEmpty` cop

ydah:
This pull request add a new cop should use the `be_empty` matcher when checking for an empty array.

### @example
```ruby
# bad
expect(array.empty?).to be true
expect(array.empty?).to be_truthy
expect(array.size).to eq(0)
expect(array.length).to eq(0)
expect(array.count).to eq(0)
expect(array).to eql([])
expect(array).to contain_exactly
expect(array).to match_array([])

# good
expect(array).to be_empty
```

- [x] Added tests.
- [x] Updated documentation.


bquorning:
I suggested something similar back in https://github.com/rubocop/rubocop-rspec/issues/244, and among others @pirj had comments in https://github.com/rubocop/rubocop-rspec/issues/244#issuecomment-522991027:

> `be_empty` is less specific than `eq({})`/`eq("")`/`eq([])`, and some prefer the examples to be more indicative and strict.

ydah:
I had missed that comment. Thank you.
I see. That is indeed true. How about targeting only the following two?

```ruby
# bad
expect(array).to contain_exactly
expect(array).to match_array([])

# good
expect(array).to eq([])
expect(array).to eql([])
```

ydah:
We have determined that the https://github.com/rubocop/rubocop-rspec/pull/1597 method is best and will close this.

Issue #1596, opened 2023-03-11
`PendingWithoutReason` should not require be inside `it`


AlexWayfer:
After #1566 I've got the offense from this cop for such code:

```ruby
RSpec.describe Faraday::PKCS7::Middleware do
  pending 'write some tests specific to the middleware'
end
```

I know, it's not the best code and specs, but for me it seems legit.

I got warnings for this pending, and there is definitely a reason.

But now `PendingWithoutReason` makes me put `pending` inside `it`:

![image](https://user-images.githubusercontent.com/6510020/224493632-c4e73893-5928-468e-bfc2-7dbe1347c5f9.png)

But it works in another way: instead of warnings tests fail:

https://github.com/rubocop/rubocop-rspec/pull/1566#issuecomment-1464933825

I'd like to see calibrated `PendingWithoutReason`, allowing `pending 'some reason'` outside of `it`.

/cc @bquorning @ydah

PR #1598, opened 2023-03-12
Fix a false positive for `RSpec/PendingWithoutReason` when pending/skip has a reason inside an example group

ydah:
Fix: https://github.com/rubocop/rubocop-rspec/issues/1596

- [x] Added tests.
- [-] Updated documentation.

### DIFF
--- a/lib/rubocop/cop/rspec/pending_without_reason.rb
+++ b/lib/rubocop/cop/rspec/pending_without_reason.rb
@@ -70,7 +70,12 @@ module RuboCop
 
         # @!method skipped_by_example_method?(node)
         def_node_matcher :skipped_by_example_method?, <<~PATTERN
-          (send nil? ${#Examples.skipped #Examples.pending} ...)
+          (send nil? ${#Examples.skipped #Examples.pending})
+        PATTERN
+
+        # @!method skipped_by_example_method_with_block?(node)
+        def_node_matcher :skipped_by_example_method_with_block?, <<~PATTERN
+          ({block numblock} (send nil? ${#Examples.skipped #Examples.pending} ...) ...)
         PATTERN
 
         # @!method metadata_without_reason?(node)
@@ -135,6 +140,10 @@ module RuboCop
 
         def on_skipped_by_example_method(node)
           skipped_by_example_method?(node) do |pending|
+            add_offense(node, message: "Give the reason for #{pending}.")
+          end
+
+          skipped_by_example_method_with_block?(node.parent) do |pending|
             add_offense(node, message: "Give the reason for #{pending}.")
           end
         end
--- a/lib/rubocop/cop/rspec/pending_without_reason.rb
+++ b/lib/rubocop/cop/rspec/pending_without_reason.rb
@@ -107,7 +107,7 @@ module RuboCop
             on_skipped_by_example_method(node)
             on_skipped_by_example_group_method(node)
           elsif example?(parent)
-            on_skipped_by_in_example_method(node, parent)
+            on_skipped_by_in_example_method(node)
           end
         end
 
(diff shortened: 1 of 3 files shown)


ydah:
cc/ @AlexWayfer @bquorning  @pirj

bquorning (approved):

AlexWayfer (approved):

lib/rubocop/cop/rspec/pending_without_reason.rb
```diff
@@ -70,7 +70,12 @@ class PendingWithoutReason < Base
 
         # @!method skipped_by_example_method?(node)
         def_node_matcher :skipped_by_example_method?, <<~PATTERN
-          (send nil? ${#Examples.skipped #Examples.pending} ...)
+          (send nil? ${#Examples.skipped #Examples.pending})
+        PATTERN
+
+        # @!method skipped_by_example_method_with_block?(node)
+        def_node_matcher :skipped_by_example_method_with_block?, <<~PATTERN
+          ({block numblock} (send nil? ${#Examples.skipped #Examples.pending} ...) ...)
```
pirj, line 78: There is a difference here.
It manifests this way:
```diff
        RSpec.describe Foo do
          skip 'does something'
          pending 'does something'
          skip 'does something' do
       +  ^^^^^^^^^^^^^^^^^^^^^ Give the reason for skip.
            1
          end
```

Should it be like that?

pirj: If you look at [the implementation](https://github.com/rspec/rspec-core/blob/cb1c8f1b5b12193221729971fbe5b390a4be877b/lib/rspec/core/example_group.rb#L186) of skipped/pending/xit/..., you see that those set a bogus reason/pending message.

The goal of this cop is to enforce having a reason:
> Checks for pending or skipped examples without reason.

And the output clearly doesn't give any hint to why those were skipped/are pending:
```
  1) Foo does something
     # No reason given
     # ./spec/a_spec.rb:2

  2) Foo does something
     # Not yet implemented
     # ./spec/a_spec.rb:3
```

pirj: Even though putting `pending` inside an example changes the behaviour as stated in the ticket, I'd say that a better way would be:
```ruby
pending "does something", pending: "not yet implemented"
```
Is it too verbose to your liking, as it just duplicates the default?
Or is it a case of explicit vs implicit?

AlexWayfer: `pending "does something", reason: "not yet implemented"` would be better.

pirj: Agree, but it's an RSpec DSL, we don't have much control over it here. pending...pending sounds like a tautology indeed.

ydah: I apologize for my mistake.
The reason for the mistake was that we were checking with test code like this.

code
````ruby
pending 'This cop is not implemented yet. Feel free to submit a pull request.'
```

result
```
Pending: (Failures listed here are expected and do not affect your suite's status)

  1) RuboCop::Cop::Style::RedundantLineContinuations This cop is not implemented yet. Feel free to submit a pull request. <---- It looked to me like the output was coming from here.
     # Not yet implemented <---- But the correct one is here.
     # ./spec/foo_spec.rb:4
```

bquorning: I actually do find `pending "does something", pending: "not yet implemented"pending "does something", pending: "not yet implemented"` too verbose.

I don’t think I have ever used the form `pending "<string>"` for any other reason than it being a reminder that I need to implement a certain example (as described by `<string>`). Having to spell out `pending: "not yet implemented"` or similar is unnecessary.

At the least, we should allow users to configure if the `:pending` metadata should be enforced when calling `pending` without a block.

AlexWayfer: In my experience, I usually have such code, as in issue example:

```ruby
describe 'does something' do
  pending 'implement later'
end
```

I like nesting `describe` and `context`. Reason? What should I write there? "was too lazy", "had no enough time"? Something reasonable, like "we have a blocker", is pretty rare. And can be in code comment, with links.

PR #1601, opened 2023-03-14
Fix a false positive for RSpec/FactoryBot/ConsistentParenthesesStyle inside &&, || and :? when omit_parentheses is on

DmitryTsepelev:
Now cop does not suggest changes in following situations (these suggestions end up in invalid syntax):

```
can_create_user? && create(:user)
can_create_user? || create(:user)
can_create_user? ? create(:user) : nil
```

- [X] Added tests.
- [X] Updated documentation.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Master (Unreleased)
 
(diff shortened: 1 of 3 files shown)


pirj (approved):
Thank you!

bquorning (approved):

PR #1600, opened 2023-03-13
Bump version to v2.20.0

bquorning:
Should we push a ~bugfix~ minor release?

______________________________________________________________________

Before submitting the PR make sure the following are checked:

- [x] Feature branch is up-to-date with `master` (if not - rebase it).
- [x] Squashed related commits together.
- [ ] Added tests.
- [x] Updated documentation.
- [x] Added an entry to the `CHANGELOG.md` if the new code introduces user-observable changes.
- [x] The build (`bundle exec rake`) passes (be sure to run this locally, since it may produce updated documentation that you will need to commit).

PR #1603, opened 2023-03-16
Add `Strict` configuration option to `RSpec/PendingWithoutReason`

ydah:
This PR is add `Strict` configuration option to `RSpec/PendingWithoutReason`.
The following PR was found to be an incorrect fix, so I have reverted and added support for this option.
- https://github.com/rubocop/rubocop-rspec/pull/1598

This option works as before by default, but setting it to `Strict: false` will make it a violation only in the following cases
- when a `pending` or `skip` method is called with no arguments
- When a pending/skip is done as metadata for example or example group, but there is no reason for it

- [x] Added tests.
- [x] Updated documentation.

### DIFF
--- a/lib/rubocop/cop/rspec/pending_without_reason.rb
+++ b/lib/rubocop/cop/rspec/pending_without_reason.rb
@@ -5,57 +5,54 @@ module RuboCop
     module RSpec
       # Checks for pending or skipped examples without reason.
       #
-      # @example
-      #   # bad
-      #   pending 'does something' do
-      #   end
-      #
-      #   # bad
-      #   it 'does something', :pending do
-      #   end
-      #
+      # @example Strict: false (default)
       #   # bad
       #   it 'does something' do
       #     pending
       #   end
       #
       #   # bad
-      #   xdescribe 'something' do
-      #   end
-      #
-      #   # bad
-      #   skip 'does something' do
-      #   end
-      #
-      #   # bad
       #   it 'does something', :skip do
       #   end
       #
-      #   # bad
-      #   it 'does something' do
-      #     skip
+      #   # good
+      #   it 'does something', skip: 'reason' do
       #   end
-      #
-      #   # bad
-      #   it 'does something'
       #
       #   # good
       #   it 'does something' do
       #     pending 'reason'
       #   end
       #
-      #   # good
-      #   it 'does something' do
-      #     skip 'reason'
+      #   # also good - when pending/skip block with string argument
+      #   pending 'does something' do
       #   end
       #
-      #   # good
-      #   it 'does something', pending: 'reason' do
+      #   # also good - when xdescribe block with string argument
+      #   xdescribe 'something' do
       #   end
       #
-      #   # good
-      #   it 'does something', skip: 'reason' do
+      #   # also good - when pending/skip block with string argument
+      #   #             but not inside an example
+      #   RSpec.describe 'something' do
+      #     pending 'does something'
       #   end
+      #
+      # @example Strict: true
+      #   # bad - when pending/skip block with string argument
+      #   pending 'does something' do
+      #   end
+      #
+      #   # bad - when xdescribe block with string argument
+      #   xdescribe 'something' do
+      #   end
+      #
+      #   # bad - when pending/skip block with string argument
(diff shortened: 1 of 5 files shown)


bquorning:
I wonder is `Strict` should be "true" or "false" by default. I would say "false", but that may just be my personal preference.

ydah:
It is very difficult to decide which is the default.
I would somewhat prefer to leave `Strict` set to `true`, but I guess that is also my personal preference.

lib/rubocop/cop/rspec/pending_without_reason.rb
```diff
@@ -23,39 +23,61 @@ module RSpec
...
+      #   FactoryBot.define do
+      #     factory :task do
+      #       pending
+      #     end
+      #   end
+      #
+      # @example Strict: false
       #   # bad
       #   it 'does something' do
-      #     skip
```
pirj, line 36: Why the `skip` -> `pending` change?

spec/rubocop/cop/rspec/pending_without_reason_spec.rb
```diff
@@ -1,295 +1,689 @@
...
+              pending 'reason'
+              pending "#{reason}"
+              pending `echo #{reason}`
+              skip 'reason'
+              skip "#{reason}"
+              skip `echo #{reason}`
+            end
+          end
+        RUBY
+      end
```
pirj, line 21: If this is common between the two styles, does it make sense to use a shared example?

It's kind of hard to understand the difference between the two.

lib/rubocop/cop/rspec/pending_without_reason.rb
```diff
@@ -23,39 +23,61 @@ module RSpec
...
+      #   it 'does something', skip: 'reason' do
       #   end
       #
+      #   # good - when pending/skip not inside an example
+      #   FactoryBot.define do
+      #     factory :task do
+      #       pending
+      #     end
+      #   end
+      #
```
pirj, line 47:
```suggestion
```
I'd say it's not worth mentioning that this cop only inspects specs, and not factories, it's the way it should work with no doubt.

pirj (commented):
We should just ignore this
```ruby
RSpec.describe 'something' do
  pending 'does something'
end
```
as it's a way to remind self to write a spec for something, or to indicate that something is yet to be implemented along with a proper spec. Previously discussed [here](https://github.com/rubocop/rubocop-rspec/pull/1598#discussion_r1135508040).

Is there a good reason to be strict about this?

lib/rubocop/cop/rspec/pending_without_reason.rb
ydah: I simply reverted the change since it was just an attempt to use skip/pending more consistently.

spec/rubocop/cop/rspec/pending_without_reason_spec.rb
ydah: The behaviorally different specs are moved to the top of each context, and the good/bad cases have also been reviewed. How does that look?

ydah:
> Is there a good reason to be strict about this?

The motivation for this PR was not to make it more strict than before, but rather to make it possible to avoid violations for the following cases depending on the configuration options. 

```ruby
RSpec.describe 'something' do
  pending 'does something'
end
```

The goal was not to ignore them entirely, but to allow control through configuration options. 
As mentioned in https://github.com/rubocop/rubocop-rspec/pull/1598#discussion_r1134657996, it is a good reason to be able to consider those without reasons as offenses in a strict sense.

ydah:
However, it may be valid to simply ignore offenses even without using a configuration option. In that case, this PR could be closed and we could proceed with the release, unless I have missed something.

pirj:
I'd go with just ignoring this case.

ydah:
> I'd go with just ignoring this case.

I agree. So let's leave this case on ignore.

```ruby
RSpec.describe 'something' do
  pending 'does something'
end
```
So is it correct to say that the following concerns have been addressed and the blocker in the release is gone? Please point out if I have missed something.
- https://github.com/rubocop/rubocop-rspec/pull/1600#issuecomment-1467024286

PR #1604, opened 2023-03-16
Add new `RSpec/BeEmpty` cop

bquorning:
This PR is almost a copy of #1594, but with [this feedback](https://github.com/rubocop/rubocop-rspec/pull/1593#issuecomment-1467063104) incorporated.

This pull request add a new cop should use the `be_empty` matcher when using contain_exactly or match_array to check for an empty array.

```ruby
# bad
expect(array).to contain_exactly
expect(array).to match_array([])

# good
expect(array).to be_empty
```

Similar but unpublished behavior is removed from `RSpec/ContainExactly` and `RSpec/MatchArray` cops.

- [x] Added tests.
- [x] Updated documentation.


spec/rubocop/cop/rspec/contain_exactly_spec.rb
```diff
@@ -45,17 +45,14 @@
     RUBY
   end
 
-  it 'flags `contain_exactly` with no arguments' do
+  it 'does not autocorrect `contain_exactly` with no arguments' do
     expect_offense(<<-RUBY)
       it { is_expected.to contain_exactly }
-                          ^^^^^^^^^^^^^^^ Prefer `be_empty` when matching an empty collection.
+                          ^^^^^^^^^^^^^^^ Prefer `match_array` when matching array values.
```
ydah, line 51: I was wondering if modifying as per this offense would seem to be an offense of `RSpec/MatchArray` and might confuse people, especially those who are just starting to use RSpec?

lib/rubocop/cop/rspec/be_empty.rb
```diff
@@ -0,0 +1,44 @@
...
+        MSG = 'Use `be_empty` matchers for checking an empty array.'
+        RESTRICT_ON_SEND = %i[contain_exactly match_array].freeze
+
+        # @!method expect_array_matcher?(node)
+        def_node_matcher :expect_array_matcher?, <<~PATTERN
+          (send
+            (send nil? :expect _)
+            #Runners.all
+            ${
+              (send nil? {:eq :eql :match_array} (array))
```
ydah, line 28: It looks like it would be an offense for `eq([])` and `eql([])` as well, but what do you think?

```suggestion
              (send nil? :match_array (array))
```

spec/rubocop/cop/rspec/contain_exactly_spec.rb
bquorning: Me too. Would we just not register any offense then?

ydah: Well, perhaps it would be better to leave that violation to the 'RSpec/BeEmpty' rather than to those cops, or to change the offense message.

```
Prefer `eq` or `be_empty` when matching empty collection.
```

lib/rubocop/cop/rspec/be_empty.rb
bquorning: Fixed

spec/rubocop/cop/rspec/contain_exactly_spec.rb
bquorning: I chose the first, “leave that offense to the 'RSpec/BeEmpty' cop”

ydah (approved):
Thank you so much!

bquorning:
cc @pirj if you’re back 🙏🏼

pirj (approved):
Just perfect, thank you!

PR #1608, opened 2023-03-20
Improve the configuration options of the yamllint-github-action

ydah:
Make the following changes:
- Change to strict mode so that even warnings cause the CI to fail.
- Change the comment format to make it easier to read.

- [-] Added tests.
- [-] Updated documentation.

### DIFF
--- a/.github/workflows/linting.yml
+++ b/.github/workflows/linting.yml
@@ -13,6 +13,8 @@ jobs:
       - name: Yamllint
         uses: karancode/yamllint-github-action@master
         with:
+          yamllint_strict: true
+          yamllint_format: parsable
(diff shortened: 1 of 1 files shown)


.github/workflows/linting.yml
```diff
@@ -13,6 +13,8 @@ jobs:
       - name: Yamllint
         uses: karancode/yamllint-github-action@master
         with:
+          yamllint_strict: true
+          yamllint_format: parsable
```
ydah, line 17: memo) The comments will be easier to read as follows.

### Before
before

### After
after

pirj (approved):
Thank you!

PR #1610, opened 2023-04-03
Add support for `RSpec/HaveHttpStatus` when using `response.code`

### LINKED ISSUE #1609: RSpec/HaveHttpStatus is missing `response.code`
The cop is missing the following:
```ruby
expect(response.code).to eq("200")
```
which is just a variation of an offence that the cop flags:
```ruby
expect(response.status).to be(200)
```

According to [Rails docs](https://api.rubyonrails.org/classes/ActionDispatch/Response.html#method-i-code):
> Returns a string to ensure compatibility with Net::HTTPResponse.
```ruby
def code
  @status.to_s
end
```

ydah:
Fixes #1609
This PR adds support for `RSpec/HaveHttpStatus` when using `response.code`.

- [x] Added tests.
- [x] Updated documentation.


pirj (approved):
Thank you for the quick fix!

ydah:
This PR is conflicted, so I rebased this PR.

bquorning (approved):

PR #1615, opened 2023-04-04
Add RSpec/IndexedLet cop

DmitryTsepelev:
Adds the cop that tries to find indexed let statements (`item_1`, `item_2`, ...):

```ruby
# bad
let(:item_1) { create(:item) }
let(:item_2) { create(:item) }

let(:item_1) { create(:item, visible: true) }
let(:item_2) { create(:item, visible: false) }

before do
  item_1 = create(:item)
  item_2 = create(:item)
end

# good
let(:items) { create_list(:item, 2) }

let(:visible_item) { create(:item, visible: true) }
let(:invisible_item) { create(:item, visible: false) }

before do
  create_list(:item, 2)
end
```

- [X] Added tests.
- [X] Updated documentation.

- [X] Added the new cop to `config/default.yml`.
- [X] The cop is configured as `Enabled: pending` in `config/default.yml`.
- [X] The cop is configured as `Enabled: true` in `.rubocop.yml`.
- [X] The cop documents examples of good and bad code.
- [X] The tests assert both that bad code is reported and that good code is not reported.
- [X] Set `VersionAdded: "<<next>>"` in `default/config.yml`.

- [X] Set `VersionChanged: "<<next>>"` in `config/default.yml`.


DmitryTsepelev:
Not sure why that check fails, works fine locally

lib/rubocop/cop/rspec/indexed_let.rb
```diff
@@ -0,0 +1,96 @@
...
+        # @!method describe_or_context(node)
+        def_node_matcher :describe_or_context, <<~PATTERN
+          (block (send nil? {:describe | :context} ...) _ (begin $...))
+        PATTERN
+
+        # @!method any_let(node)
+        def_node_matcher :any_let, <<~PATTERN
+          (block (send nil? {:let | :let! | :let_it_be} (sym $_)) _ ...)
+        PATTERN
+
```
pirj, line 48:
```suggestion
```
We have `let?` node matcher.

lib/rubocop/cop/rspec/indexed_let.rb
```diff
@@ -0,0 +1,96 @@
...
+            .group_by(&:itself)
+            .filter { |_, lets| lets.count > cop_config['MaxRepeats'] }
+            .map(&:first)
+        end
+
+        def register_offense(node, duplicated_indexed_names)
+          duplicates = duplicated_indexed_names.map { "`#{_1}_x`" }.join(', ')
+          message = format(MSG, duplicates: duplicates)
+
+          add_offense(node, message: message)
```
pirj, line 91: I'd suggest adding separate offences for all definitions that exceed the maximum allowed limit.

lib/rubocop/cop/rspec/indexed_let.rb
```diff
@@ -0,0 +1,96 @@
...
+    module RSpec
+      # Do not set up test data using indexes (e.g., `item_1`, `item_2`).
+      #
+      # It makers reading the test harder because it's not clear what exactly
+      # is tested by this particular example.
+      #
+      # @example
+      #   # bad
+      #   let(:item_1) { create(:item) }
+      #   let(:item_2) { create(:item) }
```
pirj, line 14: This doesn't reflect the fact that the cop tolerates 3 indexed definitions.

lib/rubocop/cop/rspec/indexed_let.rb
```diff
@@ -0,0 +1,96 @@
+# frozen_string_literal: true
+
+module RuboCop
+  module Cop
+    module RSpec
+      # Do not set up test data using indexes (e.g., `item_1`, `item_2`).
+      #
+      # It makers reading the test harder because it's not clear what exactly
+      # is tested by this particular example.
```
pirj, line 9: It makes sense to mention the configurable `Max` here.

config/default.yml
```diff
@@ -503,6 +503,16 @@ RSpec/ImplicitSubject:
...
 
+RSpec/IndexedLet:
+  Description: Do not set up test data using indexes (e.g., `item_1`, `item_2`).
+  Enabled: pending
+  VersionAdded: "<<next>>"
+  Reference: https://www.rubydoc.info/gems/rubocop-rspec/RuboCop/Cop/RSpec/IndexedLet
+  MaxRepeats: 3
+  LetTypes:
+    - let
+    - let!
```
pirj, line 514: We have `Helpers` [here](https://github.com/rubocop/rubocop-rspec/blob/39a0d090991e9bc96b297655dd808a41a02e1788/config/default.yml#L71), and it can be configured by users and [third-party gems](https://docs.rubocop.org/rubocop-rspec/third_party_rspec_syntax_extensions.html), e.g. [for `let_it_be`](https://github.com/test-prof/test-prof/pull/199). Wouldn't it be sufficient?

spec/rubocop/cop/rspec/indexed_let_spec.rb
```diff
@@ -0,0 +1,105 @@
...
+      end
+    end
+
+    context 'when file contains maximum indexed repeats' do
+      specify do
+        expect_offense(<<~RUBY)
+          describe SomeService do
+          ^^^^^^^^^^^^^^^^^^^^^^^ This block declares indexed `let` statements: `item_x`. Please give them meaningful names, use create_list or move creation to the `before` block.
+            let(:item_1) { create(:item) }
+            let(:item_2) { create(:item) }
```
pirj, line 28: Would it catch indexes not prefixed with an underscore, e.g. `item1`/`item2`? Why not?

spec/rubocop/cop/rspec/indexed_let_spec.rb
```diff
@@ -0,0 +1,105 @@
...
+  describe 'before' do
+    context 'when file contains maximum indexed repeats' do
+      specify do
+        expect_offense(<<~RUBY)
+          context SomeService do
+            before do
+            ^^^^^^^^^ This block declares indexed `let` statements: `item_x`. Please give them meaningful names, use create_list or move creation to the `before` block.
+              item_1 = create(:item)
+              item_2 = create(:item)
+              item_1 = create(:item)
```
pirj, line 98: `item3`?

config/default.yml
```diff
@@ -503,6 +503,16 @@ RSpec/ImplicitSubject:
   VersionChanged: '2.13'
   Reference: https://www.rubydoc.info/gems/rubocop-rspec/RuboCop/Cop/RSpec/ImplicitSubject
 
+RSpec/IndexedLet:
+  Description: Do not set up test data using indexes (e.g., `item_1`, `item_2`).
+  Enabled: pending
+  VersionAdded: "<<next>>"
+  Reference: https://www.rubydoc.info/gems/rubocop-rspec/RuboCop/Cop/RSpec/IndexedLet
+  MaxRepeats: 3
```
pirj, line 511: Personally, I'd vote for no indexed definitions. If some would like to tolerate 3 indexed definitions, let's keep this configurable, but I would set the default to 1.

lib/rubocop/cop/rspec/indexed_let.rb
pirj: That would also match blockpass. It probably makes sense to add a spec for blockpass just in case.

bquorning:
I see `create_list` and think this probably should be a FactoryBot cop: `RSpec/FactoryBot/IndexedLet`?

pirj:
Not necessarily, `create` can be replaced with `fixture(:...)`, and the cop would still inspect the same way.

docs/modules/ROOT/pages/cops_rspec.adoc
```diff
@@ -2433,6 +2433,68 @@ it { expect(named_subject).to be_truthy }
...
+# bad
+let(:item_1) { create(:item) }
+let(:item_2) { create(:item) }
+
+let(:item_1) { create(:item, visible: true) }
+let(:item_2) { create(:item, visible: false) }
+
+before do
+  item_1 = create(:item)
+  item_2 = create(:item)
```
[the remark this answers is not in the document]
DmitryTsepelev: dropped

lib/rubocop/cop/rspec/indexed_let.rb
```diff
@@ -0,0 +1,96 @@
...
+          register_offense(node, duplicated_names)
+        end
+
+        alias on_numblock on_block
+
+        private
+
+        def find_duplicate_indexed_names(candidates)
+          candidates
+            .filter_map { yield _1 }
```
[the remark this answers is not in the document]
DmitryTsepelev: Yeah, changed to good old args and it works 🙂

lib/rubocop/cop/rspec/indexed_let.rb
```diff
@@ -0,0 +1,96 @@
...
+      #
+      #   let(:visible_item) { create(:item, visible: true) }
+      #   let(:invisible_item) { create(:item, visible: false) }
+      #
+      #   before do
+      #     create_list(:item, 2)
+      #   end
+      #
+      class IndexedLet < Base
+        MSG = 'This block declares indexed `let` statements: ' \
```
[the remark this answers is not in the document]
DmitryTsepelev: Not applied because of another suggestion to have one offence per offending node

lib/rubocop/cop/rspec/indexed_let.rb
```diff
@@ -0,0 +1,96 @@
...
+              '%<duplicates>s. Please give them meaningful names,' \
+              ' use create_list or move creation to the `before` block.'
+
+        # @!method describe_or_context(node)
+        def_node_matcher :describe_or_context, <<~PATTERN
+          (block (send nil? {:describe | :context} ...) _ (begin $...))
```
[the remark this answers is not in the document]
DmitryTsepelev: Woo hoo, that was easier than I thought

lib/rubocop/cop/rspec/indexed_let.rb
DmitryTsepelev: Fixed

lib/rubocop/cop/rspec/indexed_let.rb
DmitryTsepelev: Now we have one, so example is valid again

spec/rubocop/cop/rspec/indexed_let_spec.rb
DmitryTsepelev: Because I forgot it 😅now this check is in place too

config/default.yml
DmitryTsepelev: agreed, now it's one

DmitryTsepelev:
@pirj Addressed everything, ready for another round 🙂

lib/rubocop/cop/rspec/indexed_let.rb
```diff
@@ -0,0 +1,74 @@
...
+  module Cop
+    module RSpec
+      # Do not set up test data using indexes (e.g., `item_1`, `item_2`).
+      #
+      # It makers reading the test harder because it's not clear what exactly
+      # is tested by this particular example.
+      #
+      # This cop is configurable using the `MaxRepeats` option.
+      #
+      # @example
```
ydah, line 13: I think it would be easier to understand and better if we specify the change with the `MaxRepeats` setting value in the example.

``ruby
# @example `MaxRepeats: 1 (default)`
# good
: :
# bad

# @example `MaxRepeats: 2` # good : # bad
# good
good : : # bad
# bad
```

File: spec/rubocop/cop/rspec/indexed_let_spec.rb [hunk 14]
```diff
@@ -0,0 +1,107 @@
+# frozen_string_literal: true
+
+RSpec.describe RuboCop::Cop::RSpec::IndexedLet do
+  let(:max_repeats) { 1 }
+  let(:cop_config) { { 'MaxRepeats' => max_repeats } }
+
+  specify do
```
ydah, line 7: It sometimes outputs like this, so it seems better to pass a description to it or specify like other specs.
```
RuboCop::Cop::RSpec::IndexedLet
  is expected to eq describe SomeService do
  let(:item_1) { create(:item) }
  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ This `let` ...ame. Please give it a meaningful names, use create_list or move creation to the `before` block.
end
```

lib/rubocop/cop/rspec/indexed_let.rb
```diff
@@ -0,0 +1,74 @@
...
+
+          children = node.body&.child_nodes
+          return unless children
+
+          filter_indexed_lets(children).each do |let_node|
+            add_offense(let_node)
+          end
+        end
+
+        alias on_numblock on_block
```
ydah, line 57: This may not be necessary. Instead, it may be better to suppress it with a disable comment. For reference, `RSpec/LetBeforeExamples` and `RSpec/LetSetup` are suppressed.

https://github.com/rubocop/rubocop-rspec/blob/712eb8571dfdf1d0582bb9b2edf563a3060a17f7/lib/rubocop/cop/rspec/let_setup.rb#L50
https://github.com/rubocop/rubocop-rspec/blob/712eb8571dfdf1d0582bb9b2edf563a3060a17f7/lib/rubocop/cop/rspec/let_before_examples.rb#L54

lib/rubocop/cop/rspec/indexed_let.rb
```diff
@@ -0,0 +1,74 @@
...
+
+        private
+
+        INDEX_REGEX = /_?\d+/.freeze
+
+        def filter_indexed_lets(candidates)
+          candidates
+            .filter { |candidate| let?(candidate) }
+            .group_by { |node| let_name(node).to_s.gsub(INDEX_REGEX, '') }
```
ydah, line 66: According to this logic, I think the following code will be considered an offense.

```
describe SomeService do
  let(:item) { create(:item) }
  let(:item_1) { create(:item) }
end
```

If this is the intended behavior, it is recommended to add documentation and test code.

lib/rubocop/cop/rspec/indexed_let.rb
ydah: If numblock should also be supported, it would be good to have test code related to numblock.

ydah:
I think it would be good to have the following as test cases as well.

```ruby
describe SomeService do
  let(:item_1) { create(:item) }
  let(:foo_item_1) { create(:item) }
end

describe SomeService do
  let(:item_1) { create(:item) }
  let("item_2") { create(:item) }
end
```

lib/rubocop/cop/rspec/indexed_let.rb
DmitryTsepelev: I think it should not be intented, looks like `item` is pretty legal. Made a tweak to count this as 1

lib/rubocop/cop/rspec/indexed_let.rb
```diff
@@ -0,0 +1,84 @@
+# frozen_string_literal: true
+
+module RuboCop
+  module Cop
+    module RSpec
+      # Do not set up test data using indexes (e.g., `item_1`, `item_2`).
+      #
+      # It makers reading the test harder because it's not clear what exactly
```
bquorning, line 8:
```suggestion
      # It makes reading the test harder because it's not clear what exactly
```

lib/rubocop/cop/rspec/indexed_let.rb
```diff
@@ -0,0 +1,84 @@
...
+      #   let(:item_3) { create(:item) }
+      #
+      #   # good
+      #   let(:item_1) { create(:item) }
+      #   let(:item_2) { create(:item) }
+      #
+      class IndexedLet < Base
+        MSG = 'This `let` statement uses index in its name. Please give it ' \
+              'a meaningful names, use create_list or move creation ' \
+              'to the `before` block.'
```
bquorning, line 44:
```suggestion
              'a meaningful name, use create_list or move creation ' \
              'to a `before` block.'
```

lib/rubocop/cop/rspec/indexed_let.rb
```diff
@@ -0,0 +1,84 @@
...
+        private
+
+        INDEX_REGEX = /_?\d+/.freeze
+
+        def filter_indexed_lets(candidates)
+          candidates
+            .filter { |node| indexed_let?(node) }
+            .group_by { |node| let_name(node).to_s.gsub(INDEX_REGEX, '') }
+            .values
+            .filter { |lets| lets.count > cop_config['MaxRepeats'] }
```
bquorning, line 74: I think `#size` or `#length` is preferable when not passing a block to `#count` anyway:
```suggestion
            .filter { |lets| lets.length > cop_config['MaxRepeats'] }
```

lib/rubocop/cop/rspec/indexed_let.rb
```diff
@@ -0,0 +1,84 @@
...
+            .flatten
+        end
+
+        def indexed_let?(node)
+          let?(node) && let_name(node).to_s.match(INDEX_REGEX)
```
bquorning, line 79: Two things: We don’t need to create `MatchData` for the regex result, we just need a boolean – so we can use `#match?` instead of `#match`. And to avoid the `#to_s` to because there is no `NilClass#match` method, we can use `Regexp#match?` instead:
```suggestion
          let?(node) && INDEX_REGEX.match?(let_name(node))
```

DmitryTsepelev:
Applied all suggesstions, updated specs and docs, thanks 🙂

bquorning (approved):

ydah (approved):
Thank you for doing a great work ❤️

pirj (approved):
Thank you!
A few nitpicks below.

lib/rubocop/cop/rspec/indexed_let.rb
```diff
@@ -0,0 +1,84 @@
...
+      #   let(:item_1) { create(:item) }
+      #   let(:item_2) { create(:item) }
+      #   let(:item_3) { create(:item) }
+      #
+      #   # good
+      #   let(:item_1) { create(:item) }
+      #   let(:item_2) { create(:item) }
+      #
+      class IndexedLet < Base
+        MSG = 'This `let` statement uses index in its name. Please give it ' \
```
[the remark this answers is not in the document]
DmitryTsepelev: Dropped the second part of the message — now there are no suggestions regarding possible workarounds

DmitryTsepelev:
Updated 🙂

pirj:
Thank you, @DmitryTsepelev !

bquorning:
Oops, it looks like we forgot to squash commits before merging 😅

PR #1617, opened 2023-04-06
Add a Danger check to prevent review approval for cops that have become `Enabled: true`

ydah:
I suggest adding checks with [Danger](https://danger.systems/ruby/) as a preventive measure against recurrence.
Resolve: https://github.com/rubocop/rubocop-rspec/pull/1613#issuecomment-1497696573

If there is a cop that became `Enabled: true` due to the changes, it leaves a comment on the Pull Request as follows.
We only comment to keep track of when the cop is intentionally set to `Enabled: true`.

screenshot

- [-] Added tests.
- [-] Updated documentation.

### DIFF
--- a/Dangerfile
+++ b/Dangerfile
@@ -1,0 +1,9 @@
+# frozen_string_literal: true
+
+diff = git.diff_for_file('config/default.yml')
+if diff && diff.patch =~ /^\+\s*Enabled: true$/
+  warn(<<~MESSAGE)
+    There is a cop that became `Enabled: true` due to this pull request.
+    Please review the diff and make sure there are no issues.
+  MESSAGE
+end
--- a/.github/workflows/danger.yml
+++ b/.github/workflows/danger.yml
@@ -1,0 +1,22 @@
+name: Danger
+on:
+  - pull_request
+concurrency:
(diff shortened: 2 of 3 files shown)


pirj (approved):
Looks great! Thank you

Darhazer (approved):

Darhazer:
@bquorning objections?

Dangerfile
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
+diff = git.diff_for_file('config/default.yml')
+if diff && diff.patch =~ /^\+\s*Enabled: true$/
+  warn(<<~MESSAGE)
+    There is a cop that became `Enabled: true`` due to this pull request.
```
bquorning, line 6: Looks like a typo:
```suggestion
    There is a cop that became `Enabled: true` due to this pull request.
```

bquorning (approved):
Sorry for overlooking this PR.

Dangerfile
Darhazer: Eagle 👁️

bquorning (approved):
❤️

Issue #1485, opened 2022-11-15
Cop idea: discourage the use of `assert_`/`refute_` Minitest matchers


pirj:
[Suggested by @Darhazer](https://github.com/rubocop/rubocop-rspec/pull/1456#issuecomment-1315150349)

RSpec Expectations [allows using](https://github.com/rspec/rspec-expectations/blob/a85a42cb4c4c700b7591e4dc2cb27089bdf425c4/lib/rspec/expectations/minitest_integration.rb) [`assert_`/`refute_` Minitest matchers](http://docs.seattlerb.org/minitest/Minitest/Assertions.html)
However, RSpec comes with its matchers that pretty much cover the same:
```diff
- assert_equal a, b
+ expect(a).to eq(b)
```

A notable difference is that Minitest allows for an optional failure message.
It's a lesser known RSpec feature, [but it allows it, too](https://github.com/rspec/rspec-expectations/blob/6ebb53150c1b1a629e45a4649fa7965e6a95e39a/lib/rspec/expectations/expectation_target.rb#L63):
```diff
- assert_equal a, b, "must be equal"
+ expect(a).to(eq(b), "must be equal")
```

bquorning:
Do we know _why_ rspec-expectations provide this feature? Is it going to be deprecated?

One of the [non-goals of rubocop-rspec](https://docs.rubocop.org/rubocop-rspec/#non-goals-of-rubocop-rspec) is

> Enforcing should vs. expect syntax

and I wonder if the suggested cop doesn’t fall under this category.

pirj:
Fair enough.
RSpeс has a [сonfiguration option for that](https://relishapp.com/rspec/rspec-core/v/3-9/docs/expectation-framework-integration/configure-expectation-framework#configure-minitest-assertions):
```ruby
RSpec.configure do |config|
  config.expect_with :minitest
end
```
which is [exclusively using `rspec-expectations` in the default project `spec/spec_helper.rb` initializer](https://github.com/rspec/rspec-core/blob/6dede8ae1b9cce6e1de4baa912e77c97b587b446/lib/rspec/core/project_initializer/spec/spec_helper.rb#L20) (and most of the projects, I assume):
```ruby
config.expect_with :rspec do |expectations|
```

If with this it's still possible to use Minitest's assertions, there's probably something wrong.

[...]

PR #1618, opened 2023-04-16
Fix order of arguments in `RSpec/Rails/MinitestAssertions` correction

mvz:
The order of arguments in `assert_equals` is expected, then actual, while in `expect` syntax, the actual comes first.

In the existing correction, the actual value was incorrectly placed as argument for `eq`, while the expected value was placed as the argument for `expect`. This changes fixes the order so the resulting expression matches the intent of the original assertion.

[Mentioned here](https://github.com/rubocop/rubocop-rspec/issues/1485#issuecomment-1509276914).

- [x] Added tests.
- [x] Updated documentation.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## Master (Unreleased)
 
 - Add new `RSpec/IndexedLet` cop. ([@dmitrytsepelev])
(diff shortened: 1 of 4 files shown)


pirj (approved):
Thank you!

mvz:
I'll rebase to fix the conflicts.

bquorning (approved):

PR #1620, opened 2023-04-18
Fix a false negative for `RSpec/RedundantAround` when redundant numblock `around`

ydah:
This PR fixes a false negative for `RSpec/RedundantAround` when redundant numblock `around`.

______________________________________________________________________

Before submitting the PR make sure the following are checked:

- [x] Feature branch is up-to-date with `master` (if not - rebase it).
- [x] Squashed related commits together.
- [x] Added tests.
- [-] Updated documentation.
- [x] Added an entry to the `CHANGELOG.md` if the new code introduces user-observable changes.
- [x] The build (`bundle exec rake`) passes (be sure to run this locally, since it may produce updated documentation that you will need to commit).

PR #1621, opened 2023-04-18
Refactor with integration into `match_redundant_around_hook_block`

ydah:
Follow up: https://github.com/rubocop/rubocop-rspec/pull/1620#pullrequestreview-1389351397

- [-] Added tests.
- [-] Updated documentation.

### DIFF
--- a/lib/rubocop/cop/rspec/redundant_around.rb
+++ b/lib/rubocop/cop/rspec/redundant_around.rb
@@ -27,14 +27,7 @@ module RuboCop
             autocorrect(corrector, node)
           end
         end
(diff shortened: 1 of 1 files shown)


pirj (approved):
Thank you!
I think the original intention behind `(args _?)` was to make sure `|example|` is present as a block argument. But the whole hook wouldn't make sense without it anyway.

bquorning (approved):
Should we put this into the v2.20.0 release as well?

ydah:
@bquorning Please include, if possible 😀

Issue #1622, opened 2023-04-18
RSpec/IndexedLet should allow identifiers defined in Naming/VariableNumber: AllowedIdentifiers


pjg:
With the following configuration:

```yaml
Naming/VariableNumber:
  AllowedIdentifiers:
    - "address_line_1"
    - "address_line_2"
```

rubocop-rspec will still report `RSpec/IndexedLet` offence with this code:

```ruby
let!(:address_line_1) { 'line1' }
let!(:address_line_2) { 'line2' }
```

It would be better if the `RSpec/IndexedLet` cop would read the configuration for `AllowedIdentifiers` in the `Naming/VariableNumber` cop and ignored those.

pjg:
@bquorning I did. From the end-user (developer) perspective, I believe it to be better if the `RSpec/IndexedLet` cop would read the configuration for `AllowedIdentifiers` in the `Naming/VariableNumber` cop and ignored those. It's essentially the same rule, just that one is for application code, while the other for specs. Although adding `AllowedPatterns` is still better than not having an exclusion list.

pirj:
https://github.com/rubocop/rubocop-rspec/pull/1624/files#diff-45cfe6684f59265ea9bdff7bd9f6be687a10f30e9ec0613e2c8a58b9b6e5af7bR16 It does
I see a slight difference in option names. Is it Patterns or Identifiers ?

PR #1624, opened 2023-04-18
Add `AllowedIdentifiers` and `AllowedPatterns` configuration option to `RSpec/IndexedLet`

ydah:
Fix: https://github.com/rubocop/rubocop-rspec/issues/1622

- [x] Added tests.
- [x] Updated documentation.

### DIFF
--- a/docs/modules/ROOT/pages/cops_rspec.adoc
+++ b/docs/modules/ROOT/pages/cops_rspec.adoc
@@ -2472,6 +2472,9 @@ Do not set up test data using indexes (e.g., `item_1`, `item_2`).
 It makes reading the test harder because it's not clear what exactly
 is tested by this particular example.
 
+The configurable options `AllowedIdentifiers` and `AllowedPatterns`
+will also read those set in `Naming/VariableNumber`.
+
 === Examples
 
 ==== `Max: 1 (default)`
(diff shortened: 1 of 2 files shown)


pirj (approved):
Thank you!

I suppose if it's needed to share `AllowedIdentifiers` between this ﻿﻿﻿cop and `Naming/VariableNumber`, it can be done with YAML references.

bquorning:
I would rather that we didn’t ask our users to duplicate configuration. But reading configuration from `Naming/VariableNumber` is an anti-pattern, right? (I thought that we read a RuboCop cop’s config from one or two places in this gem, but I cannot find any now.)

ydah:
Oh, there was one thing I overlooked.
[The `AllowedIdentifiers` in `Naming/VariableNumber` must include the suffix.](https://docs.rubocop.org/rubocop/cops_naming.html#allowedidentifiers-capture3) So it may be `AllowedPatterns` that you want as a setting for this cop.

pirj (approved):
Looks flawless, thank you!

ydah:
Also added `AllowedIdentifiers`.

bquorning (approved):
Should we mention somewhere that config is read from Naming/VariableNumber also?

ydah:
Indeed it is. I have added documentation👍

Issue #1623, opened 2023-04-18
`RSpec/Rails/HaveHttpStatus` error


ccutrer:
```
invalid value for Integer(): "404 Not Found"
/Users/cody/.gem/ruby/3.1.3/gems/rubocop-rspec-2.20.0/lib/rubocop/cop/rspec/rails/have_http_status.rb:40:in `format'
/Users/cody/.gem/ruby/3.1.3/gems/rubocop-rspec-2.20.0/lib/rubocop/cop/rspec/rails/have_http_status.rb:40:in `block in on_send'
/Users/cody/.gem/ruby/3.1.3/gems/rubocop-rspec-2.20.0/lib/rubocop/cop/rspec/rails/have_http_status.rb:52:in `match_status'
/Users/cody/.gem/ruby/3.1.3/gems/rubocop-rspec-2.20.0/lib/rubocop/cop/rspec/rails/have_http_status.rb:39:in `on_send'
```

the offending line of code:

```
expect(response.status).to eql "404 Not Found"
```

Yes, this test probably fails (it's inside a conditional that I suspect never gets hit), but the cop shouldn't error on it.

PR #1625, opened 2023-04-18
Fix an error for `RSpec/Rails/HaveHttpStatus` with comparison with strings containing non-numeric characters

ydah:
Fix: https://github.com/rubocop/rubocop-rspec/issues/1623

- [x] Added tests.
- [-] Updated documentation.

### DIFF
--- a/lib/rubocop/cop/rspec/rails/have_http_status.rb
+++ b/lib/rubocop/cop/rspec/rails/have_http_status.rb
@@ -18,7 +18,7 @@ module RuboCop
           extend AutoCorrector
 
           MSG =
-            'Prefer `expect(response).%<to>s have_http_status(%<status>i)` ' \
+            'Prefer `expect(response).%<to>s have_http_status(%<status>s)` ' \
             'over `%<bad_code>s`.'
 
           RUNNERS = %i[to to_not not_to].to_set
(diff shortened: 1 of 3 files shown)


pirj:
The original reported issue is incorrect. I wouldn't bend our code to tolerate that.
Still, we could work around the formatting issue by using `have_http_status(%{status})` instead of `have_http_status(%<status>i)`, see https://github.com/rubocop/rubocop/issues/8827.

lib/rubocop/cop/rspec/rails/have_http_status.rb
```diff
@@ -37,12 +37,14 @@ class HaveHttpStatus < ::RuboCop::Cop::Base
 
           def on_send(node)
             match_status(node) do |response_status, to, match, status|
+              return unless status.to_s.match?(/\A\d+\z/)
```
pirj, line 40: Nitpick: “404 “ with a trailing space would be ignored. Should it?

pirj (approved):
Thank you!

lib/rubocop/cop/rspec/rails/have_http_status.rb
bquorning: I think that is an acceptable edge case.

bquorning (approved):

PR #1627, opened 2023-04-18
Fix a false positive with suffixes after index-like numbers

### LINKED ISSUE #1626: `RSpec/IndexedLet` is not really checking for an index
These are not indexed let's but still trigger the cop

```ruby
let(:user_90_day_avg) { 5000 }
let(:user_7_day_avg) { 3500 }
```

pirj:
Fixes #1626

- [x] Added tests.
- [-] Updated documentation.

### DIFF
--- a/lib/rubocop/cop/rspec/indexed_let.rb
+++ b/lib/rubocop/cop/rspec/indexed_let.rb
@@ -56,19 +56,24 @@ module RuboCop
 
         private
 
-        INDEX_REGEX = /_?\d+/.freeze
+        SUFFIX_INDEX_REGEX = /_?\d+$/.freeze
+        INDEX_REGEX = /\d+/.freeze
 
         def filter_indexed_lets(candidates)
           candidates
             .filter { |node| indexed_let?(node) }
-            .group_by { |node| let_name(node).to_s.gsub(INDEX_REGEX, '') }
+            .group_by { |node| let_name_stripped_index(node) }
             .values
             .filter { |lets| lets.length > cop_config['Max'] }
             .flatten
         end
 
         def indexed_let?(node)
-          let?(node) && INDEX_REGEX.match?(let_name(node))
+          let?(node) && SUFFIX_INDEX_REGEX.match?(let_name(node))
+        end
+
+        def let_name_stripped_index(node)
+          let_name(node).to_s.gsub(INDEX_REGEX, '')
         end
       end
     end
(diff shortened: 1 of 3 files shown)


pirj:
﻿﻿cc @ydah @DmitryTsepelev

lib/rubocop/cop/rspec/indexed_let.rb
```diff
@@ -56,19 +56,24 @@ def on_block(node) # rubocop:disable InternalAffairs/NumblockHandler
...
             .flatten
         end
 
         def indexed_let?(node)
-          let?(node) && INDEX_REGEX.match?(let_name(node))
+          let?(node) && SUFFIX_INDEX_REGEX.match?(let_name(node))
+        end
+
+        def let_name_stripped_index(node)
+          let_name(node).to_s.gsub(INDEX_REGEX, '')
```
bquorning, line 76: How do we treat a let like `foo_1_bar_2`? Should we rather do e.g.

```suggestion
          let_name(node).to_s.sub(SUFFIX_INDEX_REGEX m, '')
```

pirj: Do you think some case would behave differently after this change?
We have [this spec](https://github.com/rubocop/rubocop-rspec/pull/1627/files#diff-fff401968bfa5dae29968a6893ef87c57850c142ca18637268d6ee7991611853R63):
```ruby
  it 'flags names with two numbers' do
    expect_offense(<<~RUBY)
      context SomeService do
        let(:user_1_item_1) { create(:item) }
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ This `let` statement uses index in its name. Please give it a meaningful name.
        let(:user_1_item_2) { create(:item) }
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ This `let` statement uses index in its name. Please give it a meaningful name.
        let(:user_2_item_1) { create(:item) }
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ This `let` statement uses index in its name. Please give it a meaningful name.
      end
    RUBY
  end
```

bquorning: Ahh, I was mistakenly thinking we used `#let_name_stripped_index` for auto-correcting, but we don’t. And I see now that we _do_ want to group `user_1_item_2 and `user_2_item_1` together as linked offenses.

bquorning (approved):

ydah (approved):
Thank you 🙌🏻

PR #1630, opened 2023-04-23
Add support `a_block_changing` and `changing` for `RSpec/ChangeByZero`

ydah:
This PR is add support `a_block_changing` and `changing` for `RSpec/ChangeByZero`.

- [x] Added tests.
- [x] Updated documentation.


spec/rubocop/cop/rspec/change_by_zero_spec.rb
```diff
@@ -25,6 +25,56 @@
...
+    expect_offense(<<-RUBY)
+      it do
+        expect { foo }.to a_block_changing(Foo, :bar).by(0)
+                          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ Prefer `not_to a_block_changing` over `to a_block_changing.by(0)`.
+        expect { foo }.to a_block_changing(::Foo, :bar).by(0)
+                          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ Prefer `not_to a_block_changing` over `to a_block_changing.by(0)`.
+        expect { foo }.to a_block_changing { Foo.bar }.by(0)
+                          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ Prefer `not_to a_block_changing` over `to a_block_changing.by(0)`.
+        expect { foo }.to a_block_changing(Foo, :bar).by 0
+                          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ Prefer `not_to a_block_changing` over `to a_block_changing.by(0)`.
```
pirj, line 39: Does it make sense to cover all of that?

pirj (approved):
Nice catch, thank you!
There's also `RSpec/Expectchange` that is unaware of those aliases.

spec/rubocop/cop/rspec/change_by_zero_spec.rb
bquorning: `a_block_changing` is just an alias of `change` and `changing`, right? I am not sure in which context it is meant to be used. Grammatically, I think only `change` could be used in this situation
```
expect { foo }.not_to change(Foo, :bar)
```

`a_block_changing` was added to RSpec in https://github.com/rspec/rspec-expectations/commit/9cf0fea4100ba347f75da014046e628edfac5de2, but it is unclear what it is used for. Perhaps composable matchers?

pirj: Are you hinting that `a_block_changing` is not used in practice, @bquorning ? I can confirm that by scanning real-world-rspec, the only usages I could find are from rspec-expectations. It's hard to say if `changing` is used or not, as it's a commonly used wording.

bquorning: Not hinting anything, except that I have never used it myself 😅

lib/rubocop/cop/rspec/change_by_zero.rb
```diff
@@ -29,6 +31,8 @@ module RSpec
       #   # good
       #   expect { run }.not_to change(Foo, :bar)
       #   expect { run }.not_to change { Foo.bar }
+      #   expect { run }.not_to a_block_changing(Foo, :bar)
+      #   expect { run }.not_to changing(Foo, :bar)
```
bquorning, line 35: I would omit these two lines from the documentation.

lib/rubocop/cop/rspec/change_by_zero.rb
```diff
@@ -17,6 +17,8 @@ module RSpec
       #   # bad
       #   expect { run }.to change(Foo, :bar).by(0)
       #   expect { run }.to change { Foo.bar }.by(0)
+      #   expect { run }.to a_block_changing(Foo, :bar).by(0)
+      #   expect { run }.to changing(Foo, :bar).by(0)
```
bquorning, line 21: I would remove these two lines too.

spec/rubocop/cop/rspec/change_by_zero_spec.rb
ydah: We have changed the autocorrect to the following format. WDYT?

```ruby
expect { run }.not_to change(Foo, :bar)
```

lib/rubocop/cop/rspec/change_by_zero.rb
ydah: I see. I removed it 👍🏻

bquorning (approved):
❤️ Thank you, I think it is better this way. What do you say @pirj?

pirj:
Looks good. I wish we had some feedback on the offences end users detect to focus our efforts.
Thank you, @ydah !

PR #1635, opened 2023-04-27
Require RuboCop 1.33 or higher and smaller than 2.0

ydah:
This PR requires RuboCop 1.33 or higher and smaller than 2.0. By doing this, we would only need to be concerned with the lower limit of the version, making management easier. WDYT?

- [-] Added tests.
- [-] Updated documentation.

### DIFF
--- a/rubocop-rspec.gemspec
+++ b/rubocop-rspec.gemspec
@@ -37,6 +37,6 @@ Gem::Specification.new do |spec|
     'rubygems_mfa_required' => 'true'
   }
 
-  spec.add_runtime_dependency 'rubocop', '~> 1.33'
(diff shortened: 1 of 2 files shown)


rubocop-rspec.gemspec
```diff
@@ -37,6 +37,6 @@ Gem::Specification.new do |spec|
     'rubygems_mfa_required' => 'true'
   }
 
-  spec.add_runtime_dependency 'rubocop', '~> 1.33'
+  spec.add_runtime_dependency 'rubocop', '>= 1.33', '< 2.0'
```
koic, line 40: This means exactly the same thing:

- https://guides.rubygems.org/patterns/#declaring-dependencies
- https://thoughtbot.com/blog/rubys-pessimistic-operator

So there is no difference in behavior.

PR #1642, opened 2023-05-07
Change to use common ASCII characters

ydah:
- Use U+002d instead of U+2013
- Use U+2019 instead of U+0060

- [-] Added tests.
- [x] Updated documentation.

### DIFF
--- a/docs/modules/ROOT/pages/index.adoc
+++ b/docs/modules/ROOT/pages/index.adoc
@@ -5,7 +5,7 @@ https://github.com/rubocop/rubocop[RuboCop].
 
 RuboCop RSpec follows the https://docs.rubocop.org/rubocop/versioning.html[RuboCop versioning guide].
 In a nutshell, between major versions new cops are introduced in a special `pending` status.
-That means that they won’t be run unless explicitly told otherwise.
+That means that they won't be run unless explicitly told otherwise.
 RuboCop will warn on start that certain cops are neither explicitly enabled and disabled.
 On a major version release, all `pending` cops are enabled.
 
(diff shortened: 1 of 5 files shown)


pirj (approved):
Thanks!

pirj:
> Use U+2019 instead of U+0060

Isn't it the other way around?

bquorning:
Is this just because it’s simpler to use ASCII, or is there another reason? I kinda wish that we could use proper typography in our documentation, but I also know that it would probably be a larger project that I’m willing to fight for 😅

pirj:
My main problem with typography in this context was Antora's search feature, e.g. `won't` wouldn't be found. But Algolia seems to have fixed this. Or I'm making this up due to my bad memory.
And I have no objections if it was displayed with typographic symbols, but the source is ASCII, and the search just works either way.

bquorning (approved):

pirj:
Thank you!

docs/modules/ROOT/pages/index.adoc
```diff
@@ -32,7 +32,7 @@ over
 calculator.compute(line_item).should == 5
 ----
 
-is a feature of RSpec itself - you can read about it in the https://relishapp.com/rspec/rspec-expectations/docs/syntax-configuration#disable-should-syntax[RSpec Documentation].
+is a feature of RSpec itself – you can read about it in the "Disable should syntax" section of https://rspec.info/features/3-12/rspec-expectations/syntax-configuration[RSpec Documentation].
```
pirj: Oh, I think I've messed up with this merge :facepalm:

PR #1651, opened 2023-05-27
Use `\A` instead of `^` in `RSpec/PredicateMatcher`

ydah:
It never multiline matches against method name.

- [-] Added tests.
- [-] Updated documentation.

### DIFF
--- a/lib/rubocop/cop/rspec/predicate_matcher.rb
+++ b/lib/rubocop/cop/rspec/predicate_matcher.rb
@@ -74,7 +74,7 @@ module RuboCop
             name[0..-2]
           when 'exist?', 'exists?'
             'exist'
-          when /^has_/
+          when /\Ahas_/
             name.sub('has_', 'have_')[0..-2]
           else
             "be_#{name[0..-2]}"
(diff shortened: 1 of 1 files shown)


pirj:
Why?

ydah:
Follow up this  commit: https://github.com/rubocop/rubocop/commit/8c01e7239ab45b0c7ffc7883de376914eff13136

ydah:
The part where this PR made this correction is looking for the method name in a regular expression, but what I am looking for in the regular expression here is the beginning of the method name, not a multi-line match.
So I think it would be better to use `\A` rather than `^` here. WDYT?

lib/rubocop/cop/rspec/predicate_matcher.rb
```diff
@@ -74,7 +74,7 @@ def to_predicate_matcher(name)
             name[0..-2]
           when 'exist?', 'exists?'
             'exist'
-          when /^has_/
+          when /\Ahas_/
```
ydah, line 77: For example, this is looking for method names starting with `has_`. Rather than using `^`, which is a multi-line match, `\A` seems more appropriate.

bquorning (approved):
While it doesn’t matter much in this case (as @pirj commented, is it even possible to have multi-line method names?), I think that using `\A` and `\z` by default instead of `^` and `$` is a good habit.

ydah:
I don't know how to write method names on multiple lines. I thought it was not possible. So I thought it would be a good to use `\A` .

koic:
For example `define_method` can define it, TBH it's not practical:

```ruby
> define_method("foo\non_bar\nbaz") { puts 'hello' } #=> :"foo\non_bar\nbaz"
> methods.grep(/^on_/)                               #=> [:"foo\non_bar\nbaz"]
> public_send(:"foo\non_bar\nbaz")                   #=> hello
```

Also, this does not mean the start of method name. So `\A` would be suitable.

PR #1654, opened 2023-05-29
Fix a false negative for `RSpec/ExcessiveDocstringSpacing` when finds description with em space

ydah:
This PR is fix a false negative for `RSpec/ExcessiveDocstringSpacing` when finds description with em space.

The motivation for this PR is to be able to detect and autocorrect the em space that are sometimes left behind because they cannot be detected by the current `RSpec/ExcessiveDocstringSpacing`.

```ruby
describe '　　#mymethod' do
       # ^^ em space
end

describe '#mymethod　　' do
                 # ^^ em space
end
```

- [x] Added tests.
- [-] Updated documentation.

### DIFF
--- a/lib/rubocop/cop/rspec/excessive_docstring_spacing.rb
+++ b/lib/rubocop/cop/rspec/excessive_docstring_spacing.rb
@@ -52,14 +52,21 @@ module RuboCop
 
         # @param text [String]
         def excessive_whitespace?(text)
-          return true if text.start_with?(' ') || text.end_with?(' ')
-
-          text.match?(/[^\n ]  +[^ ]/)
+          text.match?(/
+            # Leading space
+            \A[[:blank:]]
+            |
+            # Trailing space
+            [[:blank:]]\z
+            |
+            # Two or more consecutive spaces, except if they are leading spaces
+            [^[[:space:]]][[:blank:]]{2,}[^[[:blank:]]]
+          /x)
         end
 
         # @param text [String]
         def strip_excessive_whitespace(text)
-          text.strip.gsub(/  +/, ' ')
+          text.strip.gsub(/[[:blank:]]{2,}/, ' ')
         end
 
         # @param node [RuboCop::AST::Node]
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Master (Unreleased)
 
+- Fix a false negative for `RSpec/ExcessiveDocstringSpacing` when finds description with em space. ([@ydah])
+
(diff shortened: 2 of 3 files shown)


lib/rubocop/cop/rspec/excessive_docstring_spacing.rb
```diff
@@ -52,14 +52,14 @@ def on_send(node)
...
-          return true if text.start_with?(' ') || text.end_with?(' ')
+          return true if text.start_with?(' ', '　') || text.end_with?(' ', '　')
 
           text.match?(/[^\n ]  +[^ ]/)
         end
 
         # @param text [String]
         def strip_excessive_whitespace(text)
-          text.strip.gsub(/  +/, ' ')
+          text.strip.gsub(/(\s|　)(\s|　)+/, ' ')
```
pirj, line 62: I wonder if we could check if ‘strip’ped docstring matches raw one. And if ‘strip’ strips out em space. We could have taken care of more weird cases.

lib/rubocop/cop/rspec/excessive_docstring_spacing.rb
```diff
@@ -52,14 +52,14 @@ def on_send(node)
 
         # @param text [String]
         def excessive_whitespace?(text)
-          return true if text.start_with?(' ') || text.end_with?(' ')
+          return true if text.start_with?(' ', '　') || text.end_with?(' ', '　')
 
           text.match?(/[^\n ]  +[^ ]/)
```
bquorning, line 57: Instead of explicitly listing multiple different whitespace characters, perhaps we should match with `/[[:blank:]]/` instead?

```suggestion
          text.match?(/
            # Leading space
            \A[[:blank:]]
            |
            # Trailing space
            [[:blank:]]\z
            |
            # Two or more consecutive spaces, except if they are leading spaces
            [^[[:space:]]][[:blank:]]{2,}[^[[:blank:]]]
          /x)
```

lib/rubocop/cop/rspec/excessive_docstring_spacing.rb
bquorning, line 62: Like above, let’s not list all different whitespace characters.
```suggestion
          text.strip.gsub(/[[:blank:]]{2,}/, ' ')
```

bquorning (approved):

lib/rubocop/cop/rspec/excessive_docstring_spacing.rb
ydah: It seems that `strip` does not strip em spaces. I updated this PR. What do you think?

Issue #1254, opened 2022-03-25
Cop idea: suggest receive_messages when multiple messages are stubbed


Darhazer:
When dealing with APIs it's often required to stub multiple messages. Instead of having:
```
before do
  allow(Service).to receive(:retrieve).and_return(resource)
  allow(Service).to receive(:create).and_return(resource)
  allow(Service).to receive(:update).and_return(updated_resource)
end
```

We could have a spec like:
```
  allow(Service).to receive_messages(retrieve: resource, create: resource, update: updated_resource)
```

It could be configured with minimum number of messages to be triggered, e.g. allow 2 messages on the same stub, but trigger or 3 or more

PR #1655, opened 2023-06-05
Add new `RSpec/ReceiveMessages` cop

ydah:
Resolve: https://github.com/rubocop/rubocop-rspec/issues/1254

- [x] Added tests.
- [x] Updated documentation.


lib/rubocop/cop/rspec/receive_messages.rb
```diff
@@ -0,0 +1,107 @@
...
+      #   # good
+      #   before do
+      #     allow(Service).to receive_messages(foo: bar, baz: qux)
+      #   end
+      #
+      class ReceiveMessages < Base
+        extend AutoCorrector
+        include RangeHelp
+
+        MSG = 'Use `receive_messages` instead of multiple stub on line(s) ' \
```
pirj, line 24: Shouldn't "lines" be always plural?

lib/rubocop/cop/rspec/receive_messages.rb
```diff
@@ -0,0 +1,107 @@
...
+        PATTERN
+
+        # @!method receive_node(node)
+        def_node_search :receive_node, <<-PATTERN
+          $(send (send nil? :receive ...) ...)
+        PATTERN
+
+        # @!method receive_and_return_argument(node)
+        def_node_matcher :receive_and_return_argument, <<-PATTERN
+          (send (send nil? :allow ...) :to (send (send nil? :receive (sym $_)) :and_return (send nil? $_)))
```
pirj, line 44: What if the argument to `and_return` is a literal, and not a `send`?

lib/rubocop/cop/rspec/receive_messages.rb
```diff
@@ -0,0 +1,107 @@
...
+              '%<loc>s.'
+
+        # @!method allow_receive_message?(node)
+        def_node_matcher :allow_receive_message?, <<-PATTERN
+          (send (send nil? :allow ...) :to (send (send nil? :receive ...) ...) ...)
+        PATTERN
+
+        # @!method allow_argument(node)
+        def_node_matcher :allow_argument, <<-PATTERN
+          (send (send nil? :allow (const nil? $_) ...) ...)
```
pirj, line 34: Why only a const receiver?

lib/rubocop/cop/rspec/receive_messages.rb
```diff
@@ -0,0 +1,107 @@
...
+        PATTERN
+
+        # @!method receive_node(node)
+        def_node_search :receive_node, <<-PATTERN
+          $(send (send nil? :receive ...) ...)
+        PATTERN
+
+        # @!method receive_and_return_argument(node)
+        def_node_matcher :receive_and_return_argument, <<-PATTERN
```
pirj, line 43:
```suggestion
        def_node_matcher :receive_and_return_argument, <<~PATTERN
```

spec/rubocop/cop/rspec/receive_messages_spec.rb
```diff
@@ -0,0 +1,54 @@
+# frozen_string_literal: true
+
+RSpec.describe RuboCop::Cop::RSpec::ReceiveMessages, :config do
+  it 'registers an offense when multiple messeages stubbed on the ' \
+     'same object' do
+    expect_offense(<<~RUBY)
+      before do
+        allow(Service).to receive(:foo).and_return(baz)
+        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ Use `receive_messages` instead of multiple stub on line(s) [3, 4].
+        allow(Service).to receive(:bar).and_return(bar)
```
pirj, line 10: What if both those allow to receive the same `foo` message?

lib/rubocop/cop/rspec/receive_messages.rb
ydah: Indeed it is, I updated to `lines`

lib/rubocop/cop/rspec/receive_messages.rb
ydah: It should be searchable as well. I updated this PR.

spec/rubocop/cop/rspec/receive_messages_spec.rb
ydah: It would be better to create a separate cop to check for duplicate messages and leave it to them. As for this cop, I will try to ignore duplicate messages. What do you think?

lib/rubocop/cop/rspec/receive_messages.rb
```diff
@@ -0,0 +1,125 @@
...
+            register_offense(item, repeated_lines, args)
+          end
+        end
+
+        private
+
+        def repeated_receive_message(node)
+          node
+            .children
+            .select { |child| allow_receive_message?(child) }
```
bquorning, line 69: I’m a little concerned about the performance here.

spec/rubocop/cop/rspec/receive_messages_spec.rb
pirj: There’s an ‘ordered’ qualifier that makes sense with duplicates, along with ‘exactly(count)’ and ‘once’/‘twice’.

lib/rubocop/cop/rspec/receive_messages.rb
```diff
@@ -0,0 +1,125 @@
...
+      #   before do
+      #     allow(Service).to receive(:foo).and_return(bar)
+      #     allow(Service).to receive(:foo).and_return(qux)
+      #   end
+      #
+      class ReceiveMessages < Base
+        extend AutoCorrector
+        include RangeHelp
+
+        MSG = 'Use `receive_messages` instead of multiple stub on lines ' \
```
bquorning, line 30:
```suggestion
        MSG = 'Use `receive_messages` instead of multiple stubs on lines ' \
```

bquorning (commented):
I wonder if we should add a spec for code like e.g.
```ruby
before do
  allow(Service).to receive(:foo).and_return(1)
  calling_some_method
  calling_another_method
  allow(Service).to receive(:bar).and_return(2)
end
```

lib/rubocop/cop/rspec/receive_messages.rb
```diff
@@ -0,0 +1,125 @@
...
+          items.map do |item|
+            receive_and_return_argument(item) do |receive_value, return_arg|
+              "#{receive_value}: #{return_arg.source}"
+            end
+          end
+        end
+
+        def register_offense(item, repeated_lines, args)
+          add_offense(item, message: message(repeated_lines)) do |corrector|
+            if item.loc.line < repeated_lines.min
```
bquorning, line 100: I am running into an `comparison of Integer with nil failed` error here. I’ll try to come up with a reproduction later.

ydah:
I updated this PR.

lib/rubocop/cop/rspec/receive_messages.rb
ydah: Thank you! I updated this PR.

bquorning (approved):
@pirj or @Darhazer , do you have any comments on this new addition?

spec/rubocop/cop/rspec/receive_messages_spec.rb
```diff
@@ -0,0 +1,130 @@
+# frozen_string_literal: true
+
+RSpec.describe RuboCop::Cop::RSpec::ReceiveMessages, :config do
```
Darhazer, line 3: For spec completeness, can we add:
* allow with block syntax
`allow(Service).to receive(:foo) { baz }`
* allow with `with`
`allow(Service).to receive(:foo).with(:bar).and_return(:baz)`
* different return value types: array, hash and especially heredoc. What if we have two messages returning heredocs?

Darhazer (changes requested):
I received an error trying to correct:
```
allow(Foo).to receive(:bar).and_call_original
allow(Foo).to receive(:baz).and_call_original
```
corrected to
```
allow(Foo).to receive_messages(, )
```

Darhazer:
Another error I've received is when multiple return values are provided:
```
 allow(Foo).to receive(:completed?).and_return(false, false, true)
 allow(Foo).to receive(:bar).and_return(:baz)
```
Corrected to:
```
 allow(Foo).to receive_messages(, bar: :baz)
```

ydah:
@Darhazer Thank you for catching so many errors! I updated this PR.

spec/rubocop/cop/rspec/receive_messages_spec.rb
```diff
@@ -0,0 +1,219 @@
...
+      end
+    RUBY
+  end
+
+  it 'does not register an offense when multiple messeages stubbed on the ' \
+     'same object and `.with` method`' do
+    expect_no_offenses(<<~RUBY)
+      before do
+        allow(Service).to receive(:foo).with(1).and_return(baz)
+        allow(Service).to receive(:foo).with(2).and_return(bar)
```
Darhazer, line 169: nitpick: 
```suggestion
        allow(Service).to receive(:bar).with(2).and_return(bar)
```
Otherwise it falls under different category (stubbing on the same method)

Darhazer (approved):
```
1316 files inspected, 71 offenses detected, 71 offenses corrected
```

No false positives or wrong corrections 🎉

bquorning:
I am seeing a few wrong auto-corrections, of code like this:
```rb
allow(foo).to receive(:bar).and_return("baz" => qux)
allow(foo).to receive(:bar=).and_return(baz)
```

pirj (changes requested):
The cop should ignore [`.once` and other counts](http://rspec.info/features/3-12/rspec-mocks/setting-constraints/receive-counts/), and [`ordered`](http://rspec.info/features/3-12/rspec-mocks/setting-constraints/message-order/) just like it ignores message expectations with `with`.

pirj:
```
allow(foo).to receive(:bar).and_return("baz" => qux)
allow(foo).to receive(:bar=).and_return(baz)
```
This looks important. Please add a spec.

ydah:
@bquorning @pirj Thank you for review and catching! I updated this PR.

pirj (approved):
Looks great, thanks a lot.
I’ll keep in mind to run against teal-world-rspec before we release it.

pirj:
> 57497 files inspected, 2952 offenses detected, 2952 offenses autocorrectable

💯

PR #1662, opened 2023-06-15
Change default.yml path to use `**/spec/*` instead of `spec/*`

ydah:
This PR is change default.yml path to use `**/spec/*` instead of `spec/*`.

- [-] Added tests.
- [x] Updated documentation.


ydah:
memo)The overall default is as follows

```yaml
  Include: &1
    - "**/*_spec.rb"
    - "**/spec/**/*"
```

pirj (approved):
Thank you!

Issue #1660, opened 2023-06-10
Cop idea: Negative be_valid matcher


ydakuka:
## Actual behavior

I have the code:

```ruby
RSpec.describe AdvertForm, type: :model do
  subject(:form) { described_class.new }

  specify do
    expect(form).not_to be_valid
  end
end
```

I run rubocop:

```
ydakuka@yauhenid:~/Work/project$ bin/rails_docker rubocop spec/forms/advert_form_spec.rb
Inspecting 1 file
.

1 file inspected, no offenses detected
```

## Expected behavior

I expected to receive the following code:

```ruby
RSpec.describe AdvertForm, type: :model do
  subject(:form) { described_class.new }

  specify do
    expect(form).to be_invalid
  end
end
```

PR #1665, opened 2023-06-18
Add new `RSpec/Rails/NegationBeValid` cop

ydah:
Fix: https://github.com/rubocop/rubocop-rspec/issues/1660

- [x] Added tests.
- [x] Updated documentation.

- [-] The cop is configured as `Enabled: true` in `.rubocop.yml`.

### DIFF
--- a/lib/rubocop/cop/rspec/rails/negation_be_valid.rb
+++ b/lib/rubocop/cop/rspec/rails/negation_be_valid.rb
@@ -1,0 +1,92 @@
+# frozen_string_literal: true
+
+module RuboCop
+  module Cop
+    module RSpec
+      module Rails
+        # Enforces use of `be_invalid` or `not_to` for negated be_valid.
+        #
+        # @example EnforcedStyle: not_to (default)
+        #   # bad
+        #   expect(foo).to be_invalid
+        #
+        #   # good
+        #   expect(foo).not_to be_valid
+        #
+        # @example EnforcedStyle: be_invalid
+        #   # bad
+        #   expect(foo).not_to be_valid
+        #
+        #   # good
+        #   expect(foo).to be_invalid
+        #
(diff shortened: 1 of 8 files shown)


lib/rubocop/cop/rspec/rails/negation_be_valid.rb
```diff
@@ -0,0 +1,88 @@
...
+              corrector.replace(node.parent.loc.selector, replaced_runner)
+              corrector.replace(node.loc.selector, replaced_matcher)
+            end
+          end
+
+          private
+
+          def offense?(node)
+            (style == :not_to && be_invalid?(node)) ||
+              (style == :be_invalid && not_to?(node))
```
bquorning, line 54: Let’s make it more similar to `#replaced_runner` and `#replaced_matcher`:
```suggestion
            case
            when :not_to
              be_invalid?(node)
            when :be_invalid
              not_to?(node)
            end
```

bquorning (approved):

pirj (approved):
Nice, thank you!

ojab:
TBH quite a strange cop. As mentioned above, `#[in]valid?` is a pretty common method name for non-AR/AM objects, so it could braek something. 
Also scope of the cop is just too small, AFAIU it's mostly about style of assertions where two methods exist and return opposite booleans, in that sense `#valid?` is no different from `#odd?`, i. e. 
```ruby
it { expect(1).to be_odd }
it { expect(1).not_to be_even }
```

Also using `.not_to` by default introducing inconsistencies, multiple expectations with `.not_to`, e. g.
```ruby
expect(foo).not_to be_valid.and be_odd
```
are not supported and must be written as `expect(foo).to be_invalid.and be_odd`, so IMHO `.to be_invalid` is a more appropriate choice.

PR #1680, opened 2023-08-02
Use offense instead of violation

ydah:
This PR change to use offense instead of violation. And this PR set also `Naming/InclusiveLanguage`.

- [-] Added tests.
- [-] Updated documentation.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -70,6 +70,9 @@ Naming/InclusiveLanguage:
     auto_correct:
       Suggestions:
         - autocorrect
+    ' a violation':
+      Suggestions:
+        - an offense
     behaviour:
       Suggestions:
         - behavior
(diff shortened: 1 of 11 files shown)


.rubocop.yml
```diff
@@ -51,7 +51,7 @@ Metrics/BlockLength:
   Exclude:
     - rubocop-rspec.gemspec
     - Rakefile
-    - '**/*.rake'
+    - "**/*.rake"
```
bquorning, line 54: I am not sure about this line, but the other changes look fine.

ydah: Oops, looks like prettier has made that change. I reverted it back.

bquorning (approved):
Fine by me. What do you say, @pirj and @Darhazer?

Darhazer (approved):
This is the accepted term we use in RuboCop so 👍 for keeping the consistency

Issue #1675, opened 2023-07-28
Cop idea: 


ydakuka:

pirj:
‘be’ supports many (https://github.com/rspec/rspec-expectations/blob/9838a8f353d620a60c5c92840160f72cb28dca82/lib/rspec/matchers/built_in/be.rb#L120) other operators, including ‘===‘ and ‘=~’. Why restrict just one?

ydakuka:
For the `be ===`, I would get offences:

```ruby
# frozen_string_literal: true

require 'rails_helper'

RSpec.describe ApplicationHelper do
  specify do
    expect(foo).to be === true
  end

  specify do
    expect(bar).to be === 1
  end

  def foo
    true
  end

  def bar
    1
  end
end
```

```bash
ydakuka@yauhenid:~/Work/main_app$ brd rubocop spec/helpers/application_helper_spec.rb
Inspecting 1 file
C

Offenses:

spec/helpers/application_helper_spec.rb:7:23: C: Style/CaseEquality: Avoid the use of the case equality operator ===.
    expect(foo).to be === true
                      ^^^
spec/helpers/application_helper_spec.rb:11:23: C: Style/CaseEquality: Avoid the use of the case equality operator ===.
    expect(bar).to be === 1
                      ^^^

1 file inspected, 2 offenses detected
```

ydakuka:
As far as I know, the `=~` matcher is mainly used for regexp.

Also, the expect syntax does not support the `=~` matcher (https://github.com/rspec/rspec-expectations#regular-expressions), and rubocop has detected it:

```ruby
# frozen_string_literal: true

require 'rails_helper'

RSpec.describe ApplicationHelper do
  specify do
    expect(foo).to be =~ /string/
  end

  def foo
    'string'
  end
end
```

```bash
ydakuka@yauhenid:~/Work/main_app$ brd rubocop spec/helpers/application_helper_spec.rb
Inspecting 1 file
C

Offenses:

spec/helpers/application_helper_spec.rb:7:20: C: [Correctable] Performance/StringInclude: Use String#include? instead of a regex match with literal-only pattern.
    expect(foo).to be =~ /string/
                   ^^^^^^^^^^^^^^

1 file inspected, 1 offense detected, 1 offense autocorrectable
```

PR #1682, opened 2023-08-02
Add new `RSpec/Eq` cop

ydah:
Fix: https://github.com/rubocop/rubocop-rspec/issues/1675

- [x] Added tests.
- [x] Updated documentation.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 - Fix an infinite loop error when `RSpec/ExcessiveDocstringSpacing` finds a description with non-ASCII leading/trailing whitespace. ([@bcgraham])
 - Add new `RSpec/Eq` cop. ([@ydah])
+- Fix an incorrect autocorrect for `RSpec/ReceiveMessages` when return values declared between stubs. ([@marocchino])
 
 ## 2.23.2 (2023-08-09)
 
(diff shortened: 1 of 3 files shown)


lib/rubocop/cop/rspec/eq.rb
```diff
@@ -0,0 +1,50 @@
...
+
+module RuboCop
+  module Cop
+    module RSpec
+      # Use `eq` instead of `be ==` to compare objects.
+      #
+      # @example
+      #   # bad
+      #   expect(foo).to be == true
+      #   expect(bar).not_to be === 1
```
pirj, line 11: === is different from eq/eql/equal/==. I regret bringing it up

ydah: Ah that’s right 😅 Thank you. I updated this PR.

lib/rubocop/cop/rspec/eq.rb
```diff
@@ -0,0 +1,47 @@
+# frozen_string_literal: true
+
+module RuboCop
+  module Cop
+    module RSpec
+      # Use `eq` instead of `be ==` to compare objects.
+      #
+      # @example
+      #   # bad
+      #   expect(foo).to be == true
```
pirj, line 10: I suggest using some other value rather than ‘true’ to avoid any confusion with be_truthy etc.

bquorning: Fixed in https://github.com/rubocop/rubocop-rspec/pull/1706

ydah: Oh, I had overlooked this comment🙇‍♂ Thanks for catching this.

Issue #1679, opened 2023-08-02
Incorrect autocorrect for `RSpec/VerifiedDoubleReference`


ydah:
We have following RSpec code:
```ruby
let(:foo_mock) { instance_double('Foo bar') }
```

Run `bundle exec rubocop -A --only RSpec/VerifiedDoubleReference`.
Then it's autocorrected as follows:
```ruby
# before
let(:foo_mock) { instance_double('Foo bar') }

# after
let(:foo_mock) { instance_double(Foo bar) }
```

What should we do if the argument string of instance_double contains spaces?

pirj:
My apologies for the late opinion here, @r7kamura @ydah 

This is a duplicate of https://github.com/rubocop/rubocop-rspec/issues/1262 and had been rejected in the past.

Issue #1262, opened 2022-04-20
`RSpec/VerifiedDoubleReference` is not safe to auto-correct


jrmhaig:
It is possible to use `instance_double` with a string that is not a valid class name;

```ruby
instance_double('not a valid class', **arbitrary_attribute_list)
```

and auto-correct is not safe in this case. I believe that it is correct for this cop to have an auto-correct option but it should be marked as unsafe.

pirj:
Depending on the enforced style, the cop autocorrects either a literal class to a literal string, or the other way around.
The cop assumes that valid arguments to `instance_double` are [a literal class, a literal string](https://github.com/rubocop/rubocop-rspec/blob/a1a3351a0c9ebfe9d9e45e28c42fb3f9764fa33c/lib/rubocop/cop/rspec/verified_double_reference.rb#L15), or a variable referencing an object. The cop ignores variables.

Let's check with RSpec's documentation what are acceptable arguments to `instance_double`.
["Using an instance double" suggests](https://relishapp.com/rspec/rspec-mocks/v/3-9/docs/verifying-doubles/using-an-instance-double):
> It takes a class name or object as its first argument

You may argue that the class passed as an argument might not exist, in the cases when TDD is used, or specs are run in complete isolation. It's a valid point according to ["Verifying doubles" doc](https://relishapp.com/rspec/rspec-mocks/v/3-9/docs/verifying-doubles)
> No checking will happen if the underlying object or class is not defined

This can be confirmed by [RSpec's own specs for `instance_double`](https://github.com/rspec/rspec-mocks/blob/1611752449b5beefa1939dd2a20beb0175d46110/spec/rspec/mocks/verifying_doubles/instance_double_with_class_not_loaded_spec.rb#L20).
This, however, disprove the fact that the argument should still be a literal string that would make a valid class name.

Apparently, you are not using `instance_double` as intended to be used, and this, I suppose, is enforced by another cop, `RSpec/VerifiedDoubles`. If you would, the autocorrection would be completely safe.

On a side note, the RSpec core team recommends using literal strings to avoid excessive class loading when running specs selectively.

titusfortner:
@pirj fwiw, this auto-correct broke my legacy tests; no failures with `VerifiedDoubles` either

pirj:
@titusfortner Do you mind sharing an example?

titusfortner:
after running rubocop with auto-correct, this line causes the test to fail:
[...]

PR #1683, opened 2023-08-02
Ignore if non-constant string is passed to double on `RSpec/VerifiedDoubleReference`

r7kamura:
In this case, I believe it is better not to make it a violation in this Cop, since it cannot be replaced by constant.

- Fix https://github.com/rubocop/rubocop-rspec/issues/1679

- [x] Added tests.
- [ ] Updated documentation.

~~If you have created a new cop:~~

~~If you have modified an existing cop's configuration options:~~

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Master (Unreleased)
 
+- Ignore if non-constant string is passed to double on `RSpec/VerifiedDoubleReference`. ([@r7kamura])
+
 ## 2.23.0 (2023-07-30)
 
 - Add new `RSpec/Rails/NegationBeValid` cop. ([@ydah])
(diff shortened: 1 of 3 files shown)


r7kamura:
Of course I don't think `instance_double("foo bar")` is good code. That is why I did not add the `# good` example for this in the document. This Cop is for style consistency, so if we want to detect such codes, I think it would be better to add a separate Cop for that purpose.
What do you think?

ydah (approved):
Looks good! Thank you.

>  I think it would be better to add a separate Cop for that purpose.

I agree with you 👍

bquorning:
I think this behavior exist, by configuring the cop with `{ 'EnforcedStyle' => 'string' }`. I assume that is why @pirj closed the PR?

ydah:
memo) Just to be sure, I'll transcribe the reason for closing this PR as well. As mentioned in #1679#issuecomment-1664487516 discussion, "This is a duplicate of #1262 and had been rejected in the past."

Issue #1677, opened 2023-07-31
`RSpec/ReceiveMessages` is marked as a safe autocorrect, but is not


john-h-k:
Aside from the fact that `receive_messages` and similar could be overridden (which in my mind means it should never be considered a safe cop), it is unsafe:

```rb
allow(Foo).to receive(:bar).and_return(1)
baz_double = instance_double(Baz)
allow(Foo).to receive(:cat).and_return(baz_double)
```

autocorrects to

```rb
allow(Foo).to receive_messages(bar: 1, cat: baz_double)
baz_double = instance_double(Baz)
# errors as the double is created after its use
```

john-h-k:
I should be able to confidently run safe rubocop autocorrects without even checking them, and this doesn’t pass that test. Even with the change to reorder the lines better, this is still an inherently unsafe fix

pirj:
> I should be able to confidently run safe rubocop autocorrects without even checking them

> THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND

I don’t have much to add to that.

john-h-k:
I don't mean in a legal sense. I mean that is the attitude taken by the rest of rubocop. Other "safe" autocorrects are ones which have zero effect on the actual behaviour of the program and instead just visual. This affects the behaviour

ydah:
Consecutive `allow(Foo).to receive(:bar).and_return(baz)` limiting offenses to only should eliminate cases of unintended violations and automatic modifications, so it may be a good to change it that way. WDYT?

ccutrer:
I was thinking that for a bit, until I just ran into one of this form:

```
allow(object).to receive(:message1).and_return([{ some_complicated_config}])
allow(object).to receive(:message2).and_return(object.message1.first)
```

so... consecutive expectations _and_ none of those expectations' return value makes a call to the receiving object? but what about if it calls some other method, which would then call the receiving object? I think it should just be marked as unsafe, with some of these examples in the docs for why it's unsafe.

pirj:
I don’t mean in a legal way. I mean bugs and deficiencies take place, and “should confidently without checking” is an aspiration, but we almost never can make sure.
We do our best by running autocorrection on new/modified cops against our current project’s code and check if it works as expected.
But we don’t have the capacity to ensure it is so for e.g. doing this on all real-world-rspec repos.

[...]

PR #1687, opened 2023-08-06
Declare autocorrect as unsafe for `RSpec/ReceiveMessages`

bquorning:
As suggested in #1677.

I am unsure if I should change `VersionChanged` for the cop too.

- [ ] Added tests.
- [ ] Updated documentation.

### DIFF
--- a/config/default.yml
+++ b/config/default.yml
@@ -733,7 +733,9 @@ RSpec/ReceiveCounts:
 RSpec/ReceiveMessages:
   Description: Checks for multiple messages stubbed on the same object.
   Enabled: pending
+  SafeAutoCorrect: false
   VersionAdded: '2.23'
+  VersionChanged: "<<next>>"
   Reference: https://www.rubydoc.info/gems/rubocop-rspec/RuboCop/Cop/RSpec/ReceiveMessages
 
 RSpec/ReceiveNever:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## Master (Unreleased)
 
 - Mark to `Safe: false` for `RSpec/Rails/NegationBeValid`  cop. ([@ydah])
+- Declare autocorrect as unsafe for `RSpec/ReceiveMessages`. ([@bquorning])
 
 ## 2.23.0 (2023-07-30)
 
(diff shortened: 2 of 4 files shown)


john-h-k (approved):
LGTM

config/default.yml
```diff
@@ -733,6 +733,7 @@ RSpec/ReceiveCounts:
 RSpec/ReceiveMessages:
   Description: Checks for multiple messages stubbed on the same object.
   Enabled: pending
+  SafeAutoCorrect: false
   VersionAdded: '2.23'
```
ydah, line 737:
```suggestion
  VersionAdded: '2.23'
  VersionChanged: "<<next>>"
```

ydah (approved):
Thank you!

ydah:
We can also follow up after this PR to make autocorrect as safe as possible 👍🏻

pirj:
Historically I was always against marking cops as unsafe if they were buggy.
The only case when cops should be marked as unsafe is when it is unsafe by design.
Not temporarily. Not in a non-real far fetched scenario.

ydah:
>Historically I was always against marking cops as unsafe if they were buggy.
The only case when cops should be marked as unsafe is when it is unsafe by design.
Not temporarily. Not in a non-real far fetched scenario.

I agree with you. However, I believe that the case mentioned in the next comment may be difficult to design.
https://github.com/rubocop/rubocop-rspec/issues/1677#issuecomment-1659274540
However, it may be "a non-realistic far fetched scenario."

bquorning:
Ah sorry @pirj, I only just saw your comment now. I had just merged #1688 and was about to push this change as a bugfix release (but I’ll wait, pending this discussion).

> Not temporarily.

My idea was exactly to _temporarily_ mark the cop as unsafe, and then have a potential subsequent PR make the cop safe again. Can you elaborate on why you prefer not having temporarily unsafe cops?

pirj:
There’s nothing more permanent than temporary 😅

bquorning:
Ah yes, that is true. However in this case we have released a bug (claiming the autocorrect is safe) and the fastest way to fix it is to stop claiming the autocorrect is safe. Issue #1677 is still open, so we (someone) can pick it up and implement safe autocorrect.

Would it be ok if I proceed with the v2.23.1 release?

pirj:
Please go ahead with the release.

Issue #1689, opened 2023-08-07
Cop idea: Replace `be_exist` and `be_exists` matchers with the `exist` built-in matcher


ydakuka:
## Actual behavior

If I have the following code:

```ruby
# frozen_string_literal: true

require 'rails_helper'

RSpec.describe ApplicationHelper do
  specify do
    expect(Dir.exist?(Dir.pwd)).to be_truthy
  end

  specify do
    expect(Dir.exists?(Dir.pwd)).to be_truthy
  end
end
```

I will run rubocop with the autocorrect and get the following code:

```ruby
# frozen_string_literal: true

require 'rails_helper'

RSpec.describe ApplicationHelper do
  specify do
    expect(Dir).to exist(Dir.pwd)
  end
end
```

but If I have the following code:

```ruby
# frozen_string_literal: true

require 'rails_helper'

RSpec.describe ApplicationHelper do
  specify do
    expect(Dir).to be_exist(Dir.pwd)
  end

  specify do
    expect(Dir).to be_exists(Dir.pwd)
  end
end
```

I will run rubocop and not get any offences:

```bash
ydakuka@yauhenid:~/Work/project$ bin/rails_docker rubocop spec/helpers/application_helper_spec.rb 
Inspecting 1 file
.

1 file inspected, no offenses detected
```

## Rubocop

```
ydakuka@yauhenid:~/Work/project$ bin/rails_docker rubocop -V
1.55.1 (using Parser 3.2.2.3, rubocop-ast 1.29.0, running on ruby 2.7.8) [x86_64-linux]
  - rubocop-capybara 2.18.0
  - rubocop-factory_bot 2.23.1
  - rubocop-performance 1.18.0
  - rubocop-rails 2.20.2
  - rubocop-rake 0.6.0
  - rubocop-rspec 2.23.0
  - rubocop-thread_safety 0.5.1
```

pirj:
I guess it’s not the only one like this? I can recall ‘include’ at least.

ydah:
Maybe, `have_key` is not a built-in matcher, but a predicate matcher.
So, `be_has_key` and the rest should be handled by different cop.

ydakuka:
> Maybe, have_key is not a built-in matcher, but a predicate matcher.

Yep, https://github.com/rspec/rspec-expectations/blob/main/lib/rspec/matchers.rb#L49

PR #1694, opened 2023-08-12
Add new `RSpec/RedundantPredicateMatcher` cop

ydah:
Fix: https://github.com/rubocop/rubocop-rspec/issues/1689

- [x] Added tests.
- [x] Updated documentation.


spec/rubocop/cop/rspec/builtin_matcher_spec.rb
```diff
@@ -0,0 +1,132 @@
...
+
+RSpec.describe RuboCop::Cop::RSpec::BuiltinMatcher do
+  it 'registers an offense when using `be_all`' do
+    expect_offense(<<~RUBY)
+      expect(foo).to be_all(bar)
+                     ^^^^^^^^^^^ Use `all` instead of `be_all`.
+    RUBY
+
+    expect_correction(<<~RUBY)
+      expect(foo).to all(bar)
```
pirj, line 11: Is this correct?

spec/rubocop/cop/rspec/builtin_matcher_spec.rb
```diff
@@ -0,0 +1,132 @@
...
+  end
+
+  it 'registers an offense when using `be_exists`' do
+    expect_offense(<<~RUBY)
+      expect(foo).to be_exists(bar)
+                     ^^^^^^^^^^^^^^ Use `exist` instead of `be_exists`.
+    RUBY
+
+    expect_correction(<<~RUBY)
+      expect(foo).to exist(bar)
```
pirj, line 77: Does it accept an argument?

spec/rubocop/cop/rspec/builtin_matcher_spec.rb
```diff
@@ -0,0 +1,132 @@
...
+    RUBY
+
+    expect_correction(<<~RUBY)
+      expect(foo).to start_with(bar)
+    RUBY
+  end
+
+  it 'does not register an offense when using built-in matcher' do
+    expect_no_offenses(<<~RUBY)
+      expect(foo).to all(bar)
```
pirj, line 127: All again

lib/rubocop/cop/rspec/builtin_matcher.rb
```diff
@@ -0,0 +1,52 @@
...
+      # @example
+      #   # bad
+      #   expect(foo).to be_exist(bar)
+      #   expect(foo).not_to be_include(bar)
+      #
+      #   # good
+      #   expect(foo).to exist(bar)
+      #   expect(foo).not_to include(bar)
+      #
+      class BuiltinMatcher < Base
```
pirj, line 17: There’s a wip cop naming guideline saying yhat the name should indicate the offence. I’d suggest RedundantPredicateMatcherUse

spec/rubocop/cop/rspec/builtin_matcher_spec.rb
ydah: I have changed to the following correct code:
```ruby
expect(foo).to be_all be_odd
```

spec/rubocop/cop/rspec/builtin_matcher_spec.rb
ydah: It seems accept one argument:
https://rubydoc.info/github/rspec/rspec-expectations/RSpec/Matchers#exist-instance_method

```ruby
expect(File).to exist("path/to/file")
```

spec/rubocop/cop/rspec/builtin_matcher_spec.rb
ydah: https://github.com/rubocop/rubocop-rspec/pull/1694#discussion_r1293057164

Ditto.

lib/rubocop/cop/rspec/builtin_matcher.rb
ydah: I see. How about `RedundantPredicateMatcher`?

spec/rubocop/cop/rspec/redundant_predicate_matcher_spec.rb
```diff
@@ -0,0 +1,132 @@
...
+  it 'registers an offense when using `be_all`' do
+    expect_offense(<<~RUBY)
+      expect(foo).to be_all be_odd
+                     ^^^^^^^^^^^^^ Use `all` instead of `be_all`.
+    RUBY
+
+    expect_correction(<<~RUBY)
+      expect(foo).to all be_odd
+    RUBY
+  end
```
pirj, line 13: Is there an example with `all`?
```ruby
            expect(units).to be_all { |unit| unit.line_item == line_item_2 }
                             ^^^^^^
```
Would the conversion be correct?

pirj, line 13: Yet another
```ruby
        expect(deployment_refs_exist?).to be_all(false)
                                          ^^^^^^^^^^^^^
```

pirj, line 13: This may deserve its own cop:
```
/Users/pirj/source/real-world-rspec/gitlabhq/spec/services/deployments/archive_in_project_service_spec.rb:29:49: C: [Correctable] RSpec/RedundantPredicateMatcher: Use all instead of be_all.
        expect(deployments.map(&:archived?)).to be_all(false)
                                                ^^^^^^^^^^^^^
```
It could be `expect(deployments).to all be_archived(false)`. Looks bad with such an informal negation, though.

ydah: Since the auto-correction is not correct, would it be better not to make this an offense. So I change not offense for this case.

ydah: Ditto. I change to not offense for this case.

ydah: This looks like a good jurisdiction for other cops.

lib/rubocop/cop/rspec/redundant_predicate_matcher.rb
```diff
@@ -0,0 +1,62 @@
...
+
+        private
+
+        def message(bad_method, good_method)
+          format(MSG, bad: bad_method, good: good_method)
+        end
+
+        def replacable_arguments?(node)
+          if node.method?(:be_all)
+            node.first_argument.send_type?
```
pirj, line 45: Wondering if all send nodes are fine?
```ruby
expect(foo).to be_all be_odd
```
is. But
```ruby
let(:expected_foo_element_value) { 0 }
it do
  expect(foo).to be_all(expected_foo_element_value)
```
? It's also a send node, but can't be used as
```ruby
  expect(foo).to all expected_foo_element_value
```

It not only has to be a send, it also has to be a matcher. But we can't know it for sure in our static code checks if it is.

ydah: Ah, I see. That's for sure. Thanks for catching this. I wonder if I should exclude `all` from the scope, since it's hard to check if it's a matcher and I think it's likely to generate a lot of false positives. What do you think?

pirj: Yes, that’s something I’ve been thinking, too. Maybe ‘be_all’ can be ignored for autocorrection when there’s a send node? We already ignore the block form. I don’t think ‘be_all(true)’ eg is problematic.

ydah: We make it so that it is ignored in the autocorrection. 
Maybe we might as well make it not an offense in the first place.

pirj (approved):
Fantastic! Thank you!

pirj:
Broke on:
```ruby
      expect(json[:submissions]).to be_all do |ss|
        ss.key?("submission_history") && ss["submission_history"].empty?
      end
```
with:
```
undefined method `send_type?' for nil:NilClass
/Users/pirj/source/rubocop-rspec/lib/rubocop/cop/rspec/redundant_predicate_matcher.rb:47:in `replacable_arguments?'
/Users/pirj/source/rubocop-rspec/lib/rubocop/cop/rspec/redundant_predicate_matcher.rb:27:in `on_send'
```
Wondering why `if node.parent.block_type?` didn't work 🤔

pirj:
Otherwise:
```
57497 files inspected, 883 offenses detected, 742 offenses autocorrectable
```
🎉 

False positive? [`match` takes a mandatory argument](https://github.com/rspec/rspec-expectations/blob/562d327cb951575d13db2a43b232c9ca1456ea6a/lib/rspec/matchers.rb#L697).
```
/Users/pirj/source/real-world-rspec/gitlabhq/spec/lib/gitlab/zoom_link_extractor_spec.rb:30:58: C: [Correctable] RSpec/RedundantPredicateMatcher: Use match instead of be_match.
        expect(described_class.new('issue text')).not_to be_match
                                                         ^^^^^^^^
```
```ruby
module Gitlab
  class ZoomLinkExtractor
    # ...
    def match?
      ZOOM_REGEXP.match?(@text)
    end
```

Notes for future cops:
```ruby
expect(json.count).to be_equal 1
expect(json[0]["group_submissions"]).to be_equal nil
expect(effective_paths.size).to be_eql(1)
```

ydah:
Sorry for the delay. I updated this PR with the following corrections:
https://github.com/rubocop/rubocop-rspec/pull/1694#issuecomment-1745654090
https://github.com/rubocop/rubocop-rspec/pull/1694#issuecomment-1745680669

CHANGELOG.md
```diff
@@ -21,6 +23,7 @@
 - Fix an infinite loop error when `RSpec/ExcessiveDocstringSpacing` finds a description with non-ASCII leading/trailing whitespace. ([@bcgraham])
 - Fix an incorrect autocorrect for `RSpec/ReceiveMessages` when return values declared between stubs. ([@marocchino])
 - Fix a false positive `RSpec/Focus` when chained method call and inside define method. ([@ydah])
+- Split `RSpec/FilePath` into `RSpec/SpecFilePathSuffix` and `RSpec/SpecFilePathFormat`. `RSpec/FilePath` cop is enabled by default, the two new cups are pending and need to be enabled explicitly. ([@ydah])
```
pirj, line 26: Unrelated?

pirj (approved):
Thank you!

CHANGELOG.md
ydah: Thank you! I updated this PR.

PR #1696, opened 2023-08-14
Fix infinite loop in ExcessiveDocstringSpacing

bcgraham:
#### TL;DR: Reproduction
Leading/trailing non-ASCII whitespace triggers an infinite loop error.
```sh
ruby -e 'File.open("spec/file_spec.rb", "w+") { |f| f.puts "describe \'\u00a0foo\' do; end" }'
bundle exec rubocop --only RSpec/ExcessiveDocstringSpacing -a spec/file_spec.rb
```
#### Slightly longer
The test predicate for RSpec/ExcessiveDocstringSpacing now flags leading and trailing text matching [`[[:blank:]]`](https://github.com/rubocop/rubocop-rspec/blob/e09a7017ce918dd46db0f39c4a206a1fae7ad0c0/lib/rubocop/cop/rspec/excessive_docstring_spacing.rb#L56-L60), but the corrector still calls [`strip`](https://github.com/rubocop/rubocop-rspec/blob/e09a7017ce918dd46db0f39c4a206a1fae7ad0c0/lib/rubocop/cop/rspec/excessive_docstring_spacing.rb#L69)[^1].  [POSIX bracket expressions match non-ASCII characters](https://docs.ruby-lang.org/en/3.2/Regexp.html#class-Regexp-label-Character+Classes), but [strip only removes ASCII whitespace](https://docs.ruby-lang.org/en/3.2/String.html#class-String-label-Whitespace+in+Strings). The correction does not remove the offense. As a result, leading/trailing non-ASCII whitespace triggers an infinite correction loop, since the corrector doesn't remove the offending pattern.
#### Details

In 408e69be903f1114e44e2f1ea8ef70d8c02cf279, the test predicate for excessive whitespace was [changed](https://github.com/rubocop/rubocop-rspec/commit/408e69be903f1114e44e2f1ea8ef70d8c02cf279#diff-75148f9ae42c55870d87e3888b51b81f6f75e780890c95302fc22540f1ff5f5aR54-R70) to use `[[:blank:]]`:
```diff
         def excessive_whitespace?(text)
-          return true if text.start_with?(' ') || text.end_with?(' ')
-
-          text.match?(/[^\n ]  +[^ ]/)
+          text.match?(/
+            # Leading space
+            \A[[:blank:]]
+            |
+            # Trailing space
+            [[:blank:]]\z
+            |
+            # Two or more consecutive spaces, except if they are leading spaces
+            [^[[:space:]]][[:blank:]]{2,}[^[[:blank:]]]
+          /x)
         end
```
But the corrector method continued to use `strip` for leading/trailing whitespace, only using `[[:blank:]]` for internal whitespace:
```diff
         # @param text [String]
         def strip_excessive_whitespace(text)
-          text.strip.gsub(/  +/, ' ')
+          text.strip.gsub(/[[:blank:]]{2,}/, ' ')
         end
```
The test predicate now flags leading and trailing text matching `[[:blank:]]`, but the corrector still calls `strip`. [POSIX bracket expressions like `[[:blank:]]` match non-ASCII characters](https://docs.ruby-lang.org/en/3.2/Regexp.html#class-Regexp-label-Character+Classes), but [ `strip` only removes ASCII whitespace](https://docs.ruby-lang.org/en/3.2/String.html#class-String-label-Whitespace+in+Strings). Calling the corrector does not correct the problem. As a result, leading/trailing non-ASCII whitespace triggers an infinite correction loop.

[^1]: I've hard-coded the commit SHAs to the current `master` referent to keep it working into the future, but today, August 14, 2023, those point to today's `master`, 408e69be903f1114e44e2f1ea8ef70d8c02cf279.)

- [X] Added tests.
- [X] Updated documentation.

### DIFF
--- a/lib/rubocop/cop/rspec/excessive_docstring_spacing.rb
+++ b/lib/rubocop/cop/rspec/excessive_docstring_spacing.rb
@@ -66,7 +66,9 @@ module RuboCop
 
         # @param text [String]
         def strip_excessive_whitespace(text)
-          text.strip.gsub(/[[:blank:]]{2,}/, ' ')
+          text
+            .gsub(/[[:blank:]]{2,}/, ' ')
+            .gsub(/\A[[:blank:]]|[[:blank:]]\z/, '')
         end
 
         # @param node [RuboCop::AST::Node]
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Master (Unreleased)
 
+- Fix an infinite loop error when `RSpec/ExcessiveDocstringSpacing` finds a description with non-ASCII leading/trailing whitespace. ([@bcgraham])
+
 ## 2.23.2 (2023-08-09)
 
 - Fix an incorrect autocorrect for `RSpec/ReceiveMessages` when method is only non-word character. ([@marocchino])
@@ -797,6 +799,7 @@ Compatibility release so users can upgrade RuboCop to 0.51.0. No new features.
 [@aried3r]: https://github.com/aried3r
 [@baberthal]: https://github.com/baberthal
 [@backus]: https://github.com/backus
+[@bcgraham]: https://github.com/bcgraham
 [@biinari]: https://github.com/biinari
 [@bmorrall]: https://github.com/bmorrall
 [@bquorning]: https://github.com/bquorning
--- a/spec/rubocop/cop/rspec/excessive_docstring_spacing_spec.rb
+++ b/spec/rubocop/cop/rspec/excessive_docstring_spacing_spec.rb
@@ -28,8 +28,21 @@ RSpec.describe RuboCop::Cop::RSpec::ExcessiveDocstringSpacing do
 
     it 'finds description with leading em space' do
       expect_offense(<<-RUBY)
-        describe '　　#mymethod' do
+        describe '\u3000\u3000#mymethod' do
                   ^^^^^^^^^^^ Excessive whitespace.
+        end
+      RUBY
+
+      expect_correction(<<-RUBY)
+        describe '#mymethod' do
+        end
+      RUBY
+    end
+
+    it 'finds description with leading non-breaking space' do
+      expect_offense(<<-RUBY)
+        describe '\u00a0#mymethod' do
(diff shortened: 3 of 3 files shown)


bcgraham:
I left the actual character in the test to match existing convention, but since the file now has several whitespace characters that can't be visually differentiated, it might make sense to switch out the em space for `"\u3000"` and the non-breaking space for `"\u00a0"`. Let me know if you'd like me to make that change.

ydah:
> it might make sense to switch out the em space for "\u3000" and the non-breaking space for "\u00a0". Let me know if you'd like me to make that change.

Certainly, that's right. Can you make the changes? It's fine to squash the commits into one.

lib/rubocop/cop/rspec/excessive_docstring_spacing.rb
```diff
@@ -66,7 +66,9 @@ def excessive_whitespace?(text)
 
         # @param text [String]
         def strip_excessive_whitespace(text)
-          text.strip.gsub(/[[:blank:]]{2,}/, ' ')
+          text
+            .gsub(/[[:blank:]]{2,}/, ' ')
+            .sub(/\A[[:blank:]]|[[:blank:]]\z/, '')
```
Darhazer, line 71: question: what happens if it has both leading and trailing whitespace? will this remove them both or do we need to `.sub` calls to achieve that?

bcgraham: They're both removed, but it takes more than one iteration. There's not really a good reason to keep this `sub` instead of `gsub`, so I'll switch this to a `gsub`, so it only takes one iteration.

When writing this, I was also curious about this, and wrote a quick test and saw it pass. I removed it to be consistent with the current file (since there weren't any other combined leading/trailing whitespace tests). Since I was curious about this when I wrote it, and you were curious about it when you read it, it probably makes sense to keep those in. Locally, each of those tests takes about 0.004 seconds, so total increase of about a hundredth of a second to the test suite seems okay—let me know if you disagree.

ydah (approved):
Thank you!

bquorning (approved):
Thank you

PR #1697, opened 2023-08-16
Fix autocorrect in `RSpec/ReceiveMessages`

marocchino:
If you replace the first one, you can reference the object before initializing any objects that may be in between. So I changed it to replace it in the last position.

- [x] Added tests.
- ~~[ ] Updated documentation.~~

Related #1677


ydah (commented):
Maybe, cases mentioned in the following should also be addressed.
- https://github.com/rubocop/rubocop-rspec/issues/1677#issuecomment-1658917945

marocchino:
Sorry to disappoint you, but I read it and it seemed difficult so I just removed the fixes. 😅

ydah (approved):
Looks good! Thank you!

Issue #1069, opened 2020-11-02
RSpec/FilePath cop is completely skipped on routing specs, or specs that don't describe a constant


tom-lord:
See: https://github.com/rubocop-hq/rubocop-rspec/blob/137dc91ba422d701eed88a121275dbdec182a180/lib/rubocop/cop/rspec/file_path.rb#L75-L79

For example, consider the following:

```
# spec/requests/bad_filename_format.rb

RSpec.describe "Some really cool tests" do
  # ...
end
```

Because a _constant_ was not defined (i.e. the test description is just a `String`), the cop is skipped (passes).

However, it still would have been helpful to raise a cop offense here because **the filename does not end with `_spec.rb`**.

Likewise, the same thing happens for routing specs - rather than merely _relaxing_ the filename rules for such tests, this cop is completely skipped.

pirj:
Good catch, thanks for bringing this up. 👍 

I guess the cop should be split into the one that suggests a better file name basing on the constant name, and that reports all other file name oddities, not in regards to the constant described.

Do you think you could tackle that?
There's a 2.0 release that we're preparing, and now is the best moment to split those two cops. Otherwise, we'll have to split and keep all three cops around, two new `pending` ones and one old deprecated, up until 3.0.

ydah:
Fix by https://github.com/rubocop/rubocop-rspec/pull/1441 .

ydah:
However, dividing these into two cops seems like a good and I will try to work on it.

PR #1698, opened 2023-08-16
Split `RSpec/FilePath` into `RSpec/SpecFilePathSuffix` and `RSpec/SpecFilePathFormat`

ydah:
Fix: https://github.com/rubocop/rubocop-rspec/issues/1069

- [x] Added tests.
- [x] Updated documentation.


lib/rubocop/cop/rspec/spec_file_path_format.rb
```diff
@@ -0,0 +1,131 @@
...
+        PATTERN
+
+        # @!method metadata_key_value(node)
+        def_node_search :metadata_key_value, '(pair (sym $_key) (sym $_value))'
+
+        def on_top_level_example_group(node)
+          return unless top_level_groups.one?
+
+          example_group_arguments(node) do |class_name, arguments|
+            next if !class_name.const_type? || ignore_metadata?(arguments)
```
pirj, line 54: Unrelated to the PR. It seems we’re completely ignoring class names in the string format.
RSpec.describe ‘MyClass’ do
Also, Would ‘const_type?’ return true for cbase?

lib/rubocop/cop/rspec/spec_file_path_format.rb
```diff
@@ -0,0 +1,131 @@
...
+          return if filename_ends_with?(pattern)
+
+          # For the suffix shown in the offense message, modify the regular
+          # expression pattern to resemble a glob pattern for clearer error
+          # messages.
+          suffix = pattern.gsub('.*', '*').sub('[^/]', '').sub('\.', '.')
+          add_global_offense(format(MSG, suffix: suffix))
+        end
+
+        def ignore_metadata?(arguments)
```
pirj, line 73: Unrelated to the PR.
I think it might be possible to use a trmplated node pattern instead of the two nested loops.

lib/rubocop/cop/rspec/spec_file_path_format.rb
```diff
@@ -0,0 +1,131 @@
...
+          arguments.any? do |argument|
+            metadata_key_value(argument).any? do |key, value|
+              ignore_metadata.values_at(key.to_s).include?(value.to_s)
+            end
+          end
+        end
+
+        def correct_path_pattern(example_group, arguments)
```
pirj, line 81: Why is it called ‘example_group’ while ‘class_name’ is passed as an argument?

CHANGELOG.md
```diff
@@ -3,6 +3,9 @@
 ## Master (Unreleased)
 
 - Fix an infinite loop error when `RSpec/ExcessiveDocstringSpacing` finds a description with non-ASCII leading/trailing whitespace. ([@bcgraham])
+- Add `RSpec/SpecFilePathSuffix` cop. ([@ydah])
+- Add `RSpec/SpecFilePathFormat` cop. ([@ydah])
+- Deprecate `RSpec/FilePath` cop. Split `RSpec/FilePath` into `RSpec/SpecFilePathSuffix` and `RSpec/SpecFilePathFormat`. ([@ydah])
```
pirj, line 8: Those three lines are excessive. Just the “Split” sentence is fine.
My huge respect to your efforts is not measured by the number of lines in the changelist, @ydah !

pirj (approved):
Perfect, thank you so much!
Just one cosmetic note regarding parameter naming, and a few notes that we can turn into issues to work on in the future.

CHANGELOG.md
ydah: I had separated each commit in stages, so I also separated the changelog, but three lines is certainly an excessive. I updated it 👍

lib/rubocop/cop/rspec/spec_file_path_format.rb
ydah: Oh, it remained a pre-extraction practice. I changed it to `class_name`.

pirj (approved):
Perfect, thank you!

lib/rubocop/cop/rspec/spec_file_path_format.rb
```diff
@@ -0,0 +1,131 @@
...
+            metadata_key_value(argument).any? do |key, value|
+              ignore_metadata.values_at(key.to_s).include?(value.to_s)
+            end
+          end
+        end
+
+        def correct_path_pattern(class_name, arguments)
+          [
+            expected_path(class_name),
+            name_pattern(arguments.first),
```
bquorning, line 84: Suggestion: Let’s move ".*" in here, and make `#name_pattern` a bit simpler.
```suggestion
            ".*",
            name_pattern(arguments.first),
```

lib/rubocop/cop/rspec/spec_file_path_format.rb
```diff
@@ -0,0 +1,131 @@
...
+            '[^/]*_spec\.rb'
+          ].join
+        end
+
+        def name_pattern(method_name)
+          return unless method_name&.str_type?
+          return if ignore_methods?
+
+          ".*#{method_name.str_content.gsub(/\s/, '_').gsub(/\W/, '')}"
```
bquorning, line 93:
```suggestion
          method_name.str_content.gsub(/\s/, '_').gsub(/\W/, '')
```

CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 ## Master (Unreleased)
 
 - Fix an infinite loop error when `RSpec/ExcessiveDocstringSpacing` finds a description with non-ASCII leading/trailing whitespace. ([@bcgraham])
+- Split `RSpec/FilePath` into `RSpec/SpecFilePathSuffix` and `RSpec/SpecFilePathFormat`. ([@ydah])
```
bquorning, line 6: Should we remind people here that while the old `RSpec/FilePath` cop was enabled by default, the two new cops are `pending` and need to be explicitly enabled?

ydah:
@bquorning Sorry for the delay. I updated this PR.

bquorning (approved):
Looks good. Thank you.

config/default.yml
```diff
@@ -442,7 +442,7 @@ RSpec/ExpectOutput:
 
 RSpec/FilePath:
   Description: Checks that spec file paths are consistent and well-formed.
-  Enabled: true
+  Enabled: false
```
bquorning, line 445: When using `inherit_gem` I get the following error:

    Error: The `RSpec/FilePath` cop has been split into `RSpec/SpecFilePathFormat` and `RSpec/SpecFilePathSuffix`.
    (obsolete configuration found in /Users/bquorning/.gem/ruby/2.7.8/gems/rubocop-rspec-2.24.0/config/default.yml, please update it)

```yaml
require:
  - rubocop-rspec

inherit_gem:
  rubocop-rspec: config/default.yml
```

pirj: @bquorning Shouldn’t you also inherit the obsoletion config if you inherit the default config? I’m looking at this doc https://docs.rubocop.org/rubocop/configuration.html#inheriting-configuration-from-a-dependency-gem and it states that it is possible to inherit from several config files:
‘’’
inherit_gem:
  my-shared-gem:
    - default.yml
    - strict.yml
‘’’
Or am I missing something? What is your use cas?

pirj: Please disregard my comment above, it seems that the default config already loaded the obsoletion config.
Why would you inherit from the default config? Isn’t it the same as ‘require: rubocop-rspec’?

bquorning: This is a project where I use StandardRb’s configuration of RuboCop, along with other plugins such as rubocop-rspec. [Their documentation](https://github.com/standardrb/standard#running-standards-rules-via-rubocop) gives an example of how to do this setup:

```
require:
  - standard
  - standard-custom
  - standard-performance
  - rubocop-performance

inherit_gem:
  standard: config/base.yml
  standard-custom: config/base.yml
  standard-performance: config/base.yml
```

I guess I got carried away when adding lines to `inherit_gem`. When I remove `rubocop-rspec` from that section, it works perfectly 👍🏼

splattael, line 445: Question: The changelog reads that `RSpec/FilePath` is "enabled by default" but it seems it isn't anymore.

Currently, users are forced to replace `RSpec/FilePath` with their alternatives when upgrading a minor release from 2.23.0 to 2.24.0. Is this expected or do we consider this a breaking change? :thinking: 

Refs https://gitlab.com/gitlab-org/ruby/gems/gitlab-styles/-/merge_requests/195 

/cc @ydah @bquorning

splattael: I can open an issue to continue the discussion if needed :heart:

bquorning: Yes, please open an issue. I am not sure what is the preferred course of action – should we add the old cop back?

splattael: @ydah @bquorning I've moved the discussion to https://github.com/rubocop/rubocop-rspec/issues/1717.

PR #1704, opened 2023-09-04
Fix a false positive `RSpec/Focus` when chained method call and inside define method

ydah:
Fix: https://github.com/rubocop/rubocop-rspec/issues/398

- [x] Added tests.
- [x] Updated documentation.

### DIFF
--- a/spec/rubocop/cop/rspec/focus_spec.rb
+++ b/spec/rubocop/cop/rspec/focus_spec.rb
@@ -220,4 +220,43 @@ RSpec.describe RuboCop::Cop::RSpec::Focus do
       describe 'test', js: true do; end
     RUBY
   end
+
+  it 'ignores with chained method calls' do
+    expect_no_offenses(<<-RUBY)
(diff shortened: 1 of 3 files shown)


spec/rubocop/cop/rspec/focus_spec.rb
```diff
@@ -220,4 +220,43 @@
...
+          fscenario 'test'
+          fexample 'test'
+          fspecify 'test'
+          focus 'test'
+        end
+      end
+    RUBY
+  end
+
+  it 'ignores when inside define singletom method' do
```
bquorning, line 248:
```suggestion
  it 'ignores when inside define singleton method' do
```

bquorning (approved):

spec/rubocop/cop/rspec/focus_spec.rb
ydah: Thank you! I update this PR.

bquorning (approved):
Thanks @ydah

PR #1710, opened 2023-09-11
Fix an error when using `inherit_gem: rubocop-rspec: config/default.yml`

ydah:
Fix: https://github.com/rubocop/rubocop-rspec/pull/1698#discussion_r1319868935

Once we delete it from config/obsoletion.yml, because it is an error if it is a split.

How about proceeding as follows?
- Remove the configuration from config/obsoletion.yml once, since split would result in an error. (Patch version release)
- Release a version that warns if `RSpec/FilePath` is used (Next and subsequent releases)
- Remove `RSpec/FilePath` in the next major version update

- [-] Added tests.
- [x] Updated documentation.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Master (Unreleased)
 
+- Fix an error when using `inherit_gem: rubocop-rspec: config/default.yml`. ([@ydah])
+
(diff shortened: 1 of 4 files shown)


pirj:
See the discussion in the comment thread.

PR #1716, opened 2023-09-11
Add configuration option `CustomTransformPatterns` to `RSpec/SpecFilePathFormat`

ydah:
## Motivation

For example, if you have the following files:
```ruby
# rspec_foo_spec.rb
RSpec.describe RSpecFoo do
end

# rspec_bar_spec.rb
RSpec.describe RSpecBar do
end

# rspec_baz_spec.rb
RSpec.describe RSpecBaz do
end
```

Suppose you have the following .rubocop.yml:
```yaml
RSpec/SpecFilePathFormat:
  CustomTransform:
    RSpec: rspec
```

But this would be offense:
```
❯ bundle exec rubocop rspec* --only RSpec/SpecFilePathFormat
Inspecting 3 files
CCC

Offenses:

rspec_bar_spec.rb:1:1: C: RSpec/SpecFilePathFormat: Spec path should end with r_spec_bar*_spec.rb.
RSpec.describe RSpecBar do

rspec_baz_spec.rb:1:1: C: RSpec/SpecFilePathFormat: Spec path should end with r_spec_baz*_spec.rb.
RSpec.describe RSpecBaz do

rspec_foo_spec.rb:1:1: C: RSpec/SpecFilePathFormat: Spec path should end with r_spec_foo*_spec.rb.
RSpec.describe RSpecFoo do


3 files inspected, 3 offenses detected
```

So how about something that can be specified in a pattern?

- [x] Added tests.
- [x] Updated documentation.


pirj (approved):
Wr had this option in FilePath? How comes we’ve missed copying it over to this cop whennsplitting ?
What about yhe other cop, does it need the option?
There is no way to declare this in the obsoletion conf, is there?

Thank you!

ydah:
@pirj We didn't have this option in FilePath and had the same issue 👍🏻 

```
❯ bundle exec rubocop rspec* --only RSpec/FilePath          
Inspecting 3 files
CCC

Offenses:

rspec_bar_spec.rb:1:1: C: RSpec/FilePath: Spec path should end with r_spec_bar*_spec.rb.
RSpec.describe RSpecBar do
^^^^^^^^^^^^^^^^^^^^^^^
rspec_baz_spec.rb:1:1: C: RSpec/FilePath: Spec path should end with r_spec_baz*_spec.rb.
RSpec.describe RSpecBaz do
^^^^^^^^^^^^^^^^^^^^^^^
rspec_foo_spec.rb:1:1: C: RSpec/FilePath: Spec path should end with r_spec_foo*_spec.rb.
RSpec.describe RSpecFoo do
^^^^^^^^^^^^^^^^^^^^^^^

3 files inspected, 3 offenses detected
```

pirj:
We probably had the same issue because ‘rspec’ was not in the defaults?
I see this option in the docs https://docs.rubocop.org/rubocop-rspec/cops_rspec.html#rspecfilepath

ydah:
This is occurring though `RSpec: rspec` is set by default.
https://docs.rubocop.org/rubocop-rspec/cops_rspec.html#configurable-attributes-16

It also occurs when explicitly specified as follows
```
RSpec/FilePath:.
  Enabled: true
  CustomTransform: true
    RSpec: rspec
```

pirj:
That is weird. Wondering, why?
We even have a spec for that https://github.com/rubocop/rubocop-rspec/blob/b99bf786d15d5e8b2b950307be8742980f1d09ed/spec/rubocop/cop/rspec/file_path_spec.rb#L288

Is there a guarantee that specs from this PR won’t conceal the problem ?

lib/rubocop/cop/rspec/spec_file_path_format.rb
```diff
@@ -100,6 +110,10 @@ def expected_path(constant)
 
           File.join(
             constants.map do |name|
+              k, v = custom_transform_patterns.each.find do |k, _|
```
pirj, line 113: ‘each’ seems redundant

pirj, line 113: Would it make sense to give proper names like ‘pattern’ and ‘transformed’?

ydah: Indeed it is, I updated this PR.

ydah: Indeed it is, Thank you. Change to ‘pattern’ and ‘transformed’

ydah:
> So this is an additional option. Wondering if we could just treat existing transforms as patterns?

I'm wondering if adding this feature as a separate additional option might be better in order to ensure backward compatibility when there are already custom settings in CustomTransform. What do you think?

spec/rubocop/cop/rspec/spec_file_path_format_spec.rb
```diff
@@ -299,12 +299,40 @@ class Foo
...
+       'module name transformation' do
+      expect_no_global_offenses(<<-RUBY, 'rspec/some/foo/bar_spec.rb')
+        describe RSpec::Some::Foo, '#bar' do; end
+      RUBY
+    end
+  end
+
+  context 'when configured with `CustomTransformPatterns: ' \
+          '{ "RSpec" => "rspec" }`' do
+    let(:cop_config) { { 'CustomTransformPatterns' => { 'RSpec' => 'rspec' } } }
```
pirj, line 323: What i meant with reusing the option was to allow ‘“*RSpec*” => “rspec”’ config in CustomTransform.
That would still be backwards-compatible.
I’m not against the new option, but I experience some cognitive resistance due to the lack of real examples.
Would the following make good examples of partial trasgormations?
API -> api
DSL -> dsl

ydah:
Sorry for the delay. I have updated this PR. How do you like it?

lib/rubocop/cop/rspec/spec_file_path_format.rb
```diff
@@ -15,11 +15,21 @@ module RSpec
       #   my_class_method_spec.rb  # describe MyClass, '#method'
       #   my_class/method_spec.rb  # describe MyClass, '#method'
       #
-      # @example `CustomTransform: {RuboCop=>rubocop, RSpec=>rspec}` (default)
+      # @example `CustomTransform: {}` (default)
       #   # good
```
pirj, line 19: Why is this good?

lib/rubocop/cop/rspec/spec_file_path_format.rb
```diff
@@ -15,11 +15,21 @@ module RSpec
...
       #   rubocop_spec.rb          # describe RuboCop
       #   rspec_spec.rb            # describe RSpec
       #
+      # @example `CustomTransformPatterns: {RuboCop=>rubocop, RSpec=>rspec}` (default)
+      #   # bad
+      #   rspec_spec.rb            # describe RSpec
```
pirj, line 25: Why is this bad?

spec/rubocop/cop/rspec/spec_file_path_format_spec.rb
pirj: I’m still not convinced with RSpecClass.

lib/rubocop/cop/rspec/spec_file_path_format.rb
ydah: Ah sorry. I updated it.

lib/rubocop/cop/rspec/spec_file_path_format.rb
ydah: Sorry my bad. I updated it.

ydah:
I updated this PR.

spec/rubocop/cop/rspec/spec_file_path_format_spec.rb
```diff
@@ -233,19 +233,66 @@ class Foo
...
+      {
+        'CustomTransform' => { 'RSpec' => 'rspec' },
+        'CustomTransformPatterns' => {}
+      }
+    end
+
+    it 'registers an offense when not an exact match to custom ' \
+       'module name transformation' do
+      expect_offense(<<-RUBY, 'rspec_foo_spec.rb')
+        RSpec.describe RSpecFoo do; end
```
pirj, line 247: The best way to demo the difference between those two option would be `RSpec.describe CustomDSL` and `DSL` => `dsl` mapping. One would insist on `custom_d_s_l_*spec.rb`, the other on `custom_dsl_*spec.rb`, right?

Namespaces and `RSpecFoo` don't help understand it.

I keep losing grip about this option.

ydah: This does not appear to be an offense when followed by only one uppercase letter, as in DSL, so this is not the case that can be resolved with this option. This option may not be a common use case.

pirj:
Please pardon my continuous inability to understand what problem this PR solves.
Is it some capitalized words with untypical conversion to snake case like DBase, RSpec, VSync, macOS that are used as **part** of multiple classes (and specs for them)? Like MacOSFileSystem MacOSKeyChain etc?
Is it widespread? How does e.g Zeitwerk deal with classes themselves?

bquorning:
I’ll try to argue that `CustomTransform` solves this problem already. Yes, in a few cases it will require a long list, but at least that long list will be easy to understand. I find `CustomTransformPatterns` more difficult to understand.

ydah:
That may indeed be true. Thank you. I will close this PR.

Issue #1717, opened 2023-09-12
`RSpec/FilePath` split is disabled by default


splattael:
The changelog reads that `RSpec/FilePath` is "enabled by default" but it seems it isn't anymore.

Currently, users are forced to replace `RSpec/FilePath` with their alternatives when upgrading a minor release from 2.23.0 to 2.24.0. Is this expected or do we consider this a breaking change? :thinking: 

Refs https://gitlab.com/gitlab-org/ruby/gems/gitlab-styles/-/merge_requests/195 

/cc @ydah @bquorning

_Originally posted by @splattael in https://github.com/rubocop/rubocop-rspec/pull/1698#discussion_r1322578845_

splattael:
From https://github.com/rubocop/rubocop-rspec/pull/1698#discussion_r1322601331:

@[bquorning](https://github.com/bquorning) [Sep 12, 2023](https://github.com/rubocop/rubocop-rspec/pull/1698#discussion_r1322601331)

> [...] I am not sure what is the preferred course of action – should we add the old cop back?

Good question! I was hoping we could turn the violation into a warning but unfortunately it's [supported yet](https://github.com/rubocop/rubocop/blob/1b7da8189e0aaebc15b906a41c46ed9f46aa15f4/lib/rubocop/config_obsoletion/cop_rule.rb#L23).

I see 2 options:
1. Fix the changelog and highlight that this is a breaking change
2. Enable `FilePath` by default and undo the split (in the `config/obsoletion.yml`) and highlight the tension of the cops in the changelog. 

Later (or in the  meanwhile):
- Allow split rules to emit warnings (via `severity: warning` in "cop rule", see above) in RuboCop repository
- Re-add the split configuration with `severity: warning`
- Make it a violation (instead of warning) in the next major release

Depending on what we consider a "breaking change" both options are possible :sweat_smile: 

WDYT?

ydah:
As mentioned in the following PR
- https://github.com/rubocop/rubocop-rspec/pull/1710

How about proceeding as follows?
- Remove the configuration from config/obsoletion.yml once, since split would result in an error. (Patch version release)
- Release a version that warns if `RSpec/FilePath` is used (Next and subsequent releases)
- Remove `RSpec/FilePath` in the next major version update

splattael:
@ydah :wave: 

> How about proceeding as follows?

This sounds great!

> I think it looks good. Perhaps it would be better to implement it before step2.

[...]

PR #1718, opened 2023-09-12
Fix an error when using `RSpec/FilePath` and revert to enabled by default

ydah:
Fix: https://github.com/rubocop/rubocop-rspec/issues/1717

- [ ] Added tests.
- [x] Updated documentation.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,9 +2,11 @@
 
 ## Master (Unreleased)
 
+- Fix an error when using `RSpec/FilePath` and revert to enabled by default. If you have already moved to `RSpec/SpecFilePathSuffix` and `RSpec/SpecFilePathFormat`, disable `RSpec/FilePath` explicitly as `Enabled: false`. The `RSpec/FilePath` before migration and the `RSpec/SpecFilePathSuffix` and `RSpec/SpecFilePathFormat` as the target are available respectively. ([@ydah])
+
 ## 2.24.0 (2023-09-08)
 
-- Split `RSpec/FilePath` into `RSpec/SpecFilePathSuffix` and `RSpec/SpecFilePathFormat`. `RSpec/FilePath` cop is enabled by default, the two new cops are pending and need to be enabled explicitly. ([@ydah])
+- Split `RSpec/FilePath` into `RSpec/SpecFilePathSuffix` and `RSpec/SpecFilePathFormat`. `RSpec/FilePath` cop is disabled by default and the two new cops are pending and need to be enabled explicitly. ([@ydah])
 - Add new `RSpec/Eq` cop. ([@ydah])
 - Add `RSpec/MetadataStyle` and `RSpec/EmptyMetadata` cops. ([@r7kamura])
 - Add support `RSpec/Rails/HttpStatus` when `have_http_status` with string argument. ([@ydah])
--- a/config/default.yml
+++ b/config/default.yml
@@ -448,7 +448,7 @@ RSpec/ExpectOutput:
 
 RSpec/FilePath:
   Description: Checks that spec file paths are consistent and well-formed.
-  Enabled: false
+  Enabled: true
   Include:
     - "**/*_spec*rb*"
     - "**/spec/**/*"
--- a/config/obsoletion.yml
+++ b/config/obsoletion.yml
@@ -27,9 +27,3 @@ renamed:
   RSpec/FactoryBot/FactoryClassName: FactoryBot/FactoryClassName
   RSpec/FactoryBot/FactoryNameStyle: FactoryBot/FactoryNameStyle
   RSpec/FactoryBot/SyntaxMethods: FactoryBot/SyntaxMethods
-
-split:
-  RSpec/FilePath:
-    alternatives:
-      - RSpec/SpecFilePathFormat
-      - RSpec/SpecFilePathSuffix
--- a/docs/modules/ROOT/pages/cops_rspec.adoc
+++ b/docs/modules/ROOT/pages/cops_rspec.adoc
@@ -2003,7 +2003,7 @@ expect { my_app.print_report }.to output('Hello World').to_stdout
 |===
 | Enabled by default | Safe | Supports autocorrection | Version Added | Version Changed
 
-| Disabled
+| Enabled
 | Yes
 | No
 | 1.2
@@ -2011,6 +2011,12 @@ expect { my_app.print_report }.to output('Hello World').to_stdout
 |===
 
 Checks that spec file paths are consistent and well-formed.
+
+This cop is deprecated.
+We plan to remove it in the next major version update to 3.0.
+The migration targets are `RSpec/SpecFilePathSuffix`
+and `RSpec/SpecFilePathFormat`.
+If you are using this cop, please plan for migration.
 
 By default, this checks that spec file paths are consistent with the
 test subject and enforces that it reflects the described
--- a/lib/rubocop/cop/rspec/file_path.rb
+++ b/lib/rubocop/cop/rspec/file_path.rb
@@ -5,6 +5,12 @@ module RuboCop
     module RSpec
       # Checks that spec file paths are consistent and well-formed.
       #
+      # This cop is deprecated.
+      # We plan to remove it in the next major version update to 3.0.
+      # The migration targets are `RSpec/SpecFilePathSuffix`
+      # and `RSpec/SpecFilePathFormat`.
+      # If you are using this cop, please plan for migration.
+      #
       # By default, this checks that spec file paths are consistent with the
       # test subject and enforces that it reflects the described
       # class/module and its optionally called out method.
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -98,6 +98,9 @@ RSpec/ExampleLength:
     - heredoc
   Max: 11
 
+RSpec/FilePath:
+  Enabled: false
+
 RSpec/DescribeClass:
   Exclude:
     - spec/project/**/*.rb


ydah:
It appears that Danger is failing because GITHUB_TOKEN does not have write permission.
I don't have permission to see the Settings in this repository, so could someone please check if it is set as follows?

https://github.com/rubocop/rubocop-rspec/settings/actions
スクリーンショット 2023-09-12 18 30 32

config/obsoletion.yml
```diff
@@ -27,9 +27,3 @@ renamed:
   RSpec/FactoryBot/FactoryClassName: FactoryBot/FactoryClassName
   RSpec/FactoryBot/FactoryNameStyle: FactoryBot/FactoryNameStyle
   RSpec/FactoryBot/SyntaxMethods: FactoryBot/SyntaxMethods
-
-split:
```
pirj, line 31: Maybe keep this commented out?
Is there a better way to remember to do it on a major release?

docs/modules/ROOT/pages/cops_rspec.adoc
```diff
@@ -2012,6 +2012,12 @@ expect { my_app.print_report }.to output('Hello World').to_stdout
 
 Checks that spec file paths are consistent and well-formed.
 
+This cop is deprecated.
+We plan to remove it in the next major version update to 3.0.
+The migration destinations are `RSpec/SpecFilePathSuffix`
```
pirj, line 2017: Targets?

pirj (approved):
Thank you!

config/obsoletion.yml
bquorning: We can set up a milestone in GitHub and add an issue to it.

bquorning:
> It appears that Danger is failing because GITHUB_TOKEN does not have write permission.

On the surface, that seems a bit dangerous. Isn’t is possible to use a token with fewer permissions.

pirj:
Is Danger trying to tell us that we’re introducing a cop in Enabled status?

wondering if there’s an option to tell danger not to post PR comments but rather just dump their warnings to stdout for inspection

CHANGELOG.md
```diff
@@ -2,9 +2,11 @@
 
 ## Master (Unreleased)
 
+- Fix an error when using `RSpec/FilePath` and revert to enabled by default. If you have already migrated `RSpec/SpecFilePathSuffix` and `RSpec/SpecFilePathFormat`, no problem. The `RSpec/FilePath` before migration and the `RSpec/SpecFilePathSuffix` and `RSpec/SpecFilePathFormat` at the destination are available respectively. ([@ydah])
+
 ## 2.24.0 (2023-09-08)
 
-- Split `RSpec/FilePath` into `RSpec/SpecFilePathSuffix` and `RSpec/SpecFilePathFormat`. `RSpec/FilePath` cop is enabled by default, the two new cops are pending and need to be enabled explicitly. ([@ydah])
+- **(Breaking)** Split `RSpec/FilePath` into `RSpec/SpecFilePathSuffix` and `RSpec/SpecFilePathFormat`. `RSpec/FilePath` cop is disabled by default and the two new cops are pending and need to be enabled explicitly. ([@ydah])
```
splattael, line 9: I am confused. Isn't `RSpec/FilePath` now enabled by default again? :thinking: If so, is this really a breaking change then? :nerd_face: 

```suggestion
- Split `RSpec/FilePath` into `RSpec/SpecFilePathSuffix` and `RSpec/SpecFilePathFormat`. `RSpec/FilePath` cop is enabled by default and the two new cops are pending and need to be enabled explicitly. ([@ydah])
```

ydah: Oh, sure, we'll enable it back, but isn't it is breaking change for those who use v2.24.0?

bquorning: Perhaps we’re asking “does two breaking changes equal zero breaking changes”?

ydah: Ah, I understand. Indeed it is.

ydah:
> Is Danger trying to tell us that we’re introducing a cop in Enabled status?
> 
> wondering if there’s an option to tell danger not to post PR comments but rather just dump their warnings to stdout for inspection

https://danger.systems/ruby/
No functionality seems to be provided for output to stdout.

ydah:
@rubocop/rubocop-rspec Danger replacement should be able to be done as a follow-up, so it would be nice to be able to merge and do a patch version release if there are no problems with this first. How about it?

CHANGELOG.md
ydah: I updated this PR. How about it?

pirj (approved):
Yhank you!

CHANGELOG.md
```diff
@@ -2,9 +2,11 @@
 
 ## Master (Unreleased)
 
+- Fix an error when using `RSpec/FilePath` and revert to enabled by default. If you have already migrated `RSpec/SpecFilePathSuffix` and `RSpec/SpecFilePathFormat`, no problem. The `RSpec/FilePath` before migration and the `RSpec/SpecFilePathSuffix` and `RSpec/SpecFilePathFormat` at the destination are available respectively. ([@ydah])
```
bquorning, line 5: Isn’t it a problem for people had custom configuration for `RSpec/FilePath`, and changed it when upgrading to v2.24.0?

config/obsoletion.yml
bquorning: I added [the milestone](https://github.com/rubocop/rubocop-rspec/milestone/2), but didn’t open a issue.

CHANGELOG.md
ydah: Indeed it is. Explicitly disabled `RSpec/FilePath`. How about this?

CHANGELOG.md
```diff
@@ -2,9 +2,11 @@
 
 ## Master (Unreleased)
 
+- Fix an error when using `RSpec/FilePath`. If you have already migrated `RSpec/SpecFilePathSuffix` and `RSpec/SpecFilePathFormat`, no problem. The `RSpec/FilePath` before migration and the `RSpec/SpecFilePathSuffix` and `RSpec/SpecFilePathFormat` at the destination are available respectively. If you use `RSpec/FilePath` as is, add `RSpec/FilePath: Enabled: true` to your .rubocop.yml. ([@ydah])
```
pirj, line 5: Destination -> target?

ydah: 😱 Oh, my apologize.

pirj (commented):
Now when I think of this - why have we disabled the cop? Should we enable it again for those who don’t follow the brave ‘NewCops: true’ path? Can we split and remove the old cop in a major version, and meanwhile just keep those two new pending cops?
New cops would duplicate the function of the old one, but this is not a problem, is it?

ydah:
I had misunderstood the comment I received at https://github.com/rubocop/rubocop-rspec/pull/1718#discussion_r1327976868 . I changed it to `Enabled: false` and changed the migration plan for CHANGELOG. How about this?

pirj:
Just for the reference https://github.com/rubocop/rubocop-rspec/issues/1717#issuecomment-1715201355

i don’t think we need to disable FilePath until v3 (where we remove it).

config/default.yml
```diff
@@ -448,7 +448,7 @@ RSpec/ExpectOutput:
 
 RSpec/FilePath:
   Description: Checks that spec file paths are consistent and well-formed.
-  Enabled: false
+  Enabled: true
```
ydah, line 451: I updated this.

CHANGELOG.md
```diff
@@ -2,9 +2,11 @@
 
 ## Master (Unreleased)
 
+- Fix an error when using `RSpec/FilePath` and revert to enabled by default. If you have already moved to `RSpec/SpecFilePathSuffix` and `RSpec/SpecFilePathFormat`, disable `RSpec/FilePath` explicitly as `Enabled: false`. The `RSpec/FilePath` before migration and the `RSpec/SpecFilePathSuffix` and `RSpec/SpecFilePathFormat` at the target are available respectively. ([@ydah])
```
pirj, line 5:
```suggestion
- Fix an error when using `RSpec/FilePath` and revert to enabled by default. If you have already moved to `RSpec/SpecFilePathSuffix` and `RSpec/SpecFilePathFormat`, disable `RSpec/FilePath` explicitly as `Enabled: false`. The `RSpec/FilePath` before migration and the `RSpec/SpecFilePathSuffix` and `RSpec/SpecFilePathFormat` as the target are available respectively. ([@ydah])
```

pirj (approved):
Thank you!

bquorning (approved):

CHANGELOG.md
ydah: I updated this PR.

pirj (approved):

timon:
I think I see regression with this and custom inflections in Zeitwerk.
rubocop-rspec 2.24.1 fails CI for our project using customer inflections:
```ruby
# config/initializers/zeitwerk.rb
Rails.autoloaders.each do |autoloader|
  autoloader.inflector = Zeitwerk::Inflector.new
  autoloader.inflector.inflect(
    'bvd' => 'BvD'
  )
end
```

Every spec file that covers `BvD` module now registers an offense like this:
```
spec/services/bvd/adapters/fetch_data_spec.rb:5:1: C: RSpec/FilePath: Spec path should end with bv_d/adapters/fetch_data*_spec.rb.
describe BvD::Adapters::FetchData do
```

It was not an issue neither for `2.24.0`, nor for `2.23.2` (although I'm not sure if it was skipped)

AlexWayfer:
@timon I see no info that `rubocop-rspec` can support Zeitwerk inflections for it's own configuration.

I can suggest to you to add some configuration.

You've got the warning from `RSpec/FilePath`. Let's see its documentation: https://docs.rubocop.org/rubocop-rspec/cops_rspec.html#rspecfilepath

`CustomTransform` option looks exactly what do you need.

CHANGELOG.md
domon-envato: Feedback: Our builds were actually broken twice by this two breaking changes.

lib/rubocop/cop/rspec/file_path.rb
```diff
@@ -5,6 +5,12 @@ module Cop
     module RSpec
       # Checks that spec file paths are consistent and well-formed.
       #
+      # This cop is deprecated.
+      # We plan to remove it in the next major version update to 3.0.
+      # The migration targets are `RSpec/SpecFilePathSuffix`
+      # and `RSpec/SpecFilePathFormat`.
+      # If you are using this cop, please plan for migration.
```
domon-envato, line 12: @ydah What is the suggested way to migrate?

I've updated my config like the following for `0.24.0`.

* Removed `RSpec/FilePath`
* Disabled `RSpec/SpecFilePathFormat`
* Enabled `RSpec/SpecFilePathSuffix`

But this new change in `0.24.1` is forcing me to update my config again like the following.

* Add `RSpec/FilePath` back and disable it

Does that mean my config will be broken when `RSpec/FilePath` is removed in the next major version update to 3.0?

pirj: I understand your frustration with some reconfiguration required for a minor/patch level version dependency updates.

May I kindly ask you to elaborate what you mean by "forcing"?

You meant 2.24.0.

> my config will be broken ... in the next major version update to 3.0?

Major updates often come with breaking changes, so most probably yes.
But worry not, we'll provide a detailed update checklist like we did with [1.x -> 2.x update](https://docs.rubocop.org/rubocop-rspec/upgrade_to_version_2.html).

domon-envato: Thanks for responding to my question 🙏 

> May I kindly ask you to elaborate what you mean by "forcing"?
> You meant 2.24.0.

* Before `2.24.0`: I had `RSpec/FilePath` explicitly disabled in my config.
* At `2.24.0`: I had to remove `RSpec/FilePath` completely from my config or the rubocop command would not exit with a successful status.
* At `2.24.1`: Now I had to disable `RSpec/FilePath` again or it would be enabled by default.

After updated the config twice in each of these two versions, I was a bit alerted by the words **"plan for migration"** and curious about what that actually means.

But on second thought, I think `2.24.1` just moved the state back to what we were at before `2.24.0`.

If I interpret what you said correctly, there will be no breaking changes like this until 3.0. But we probably will need to update the config when upgrading to 3.x, which is quite normal. Is my understanding correct?

pirj: I see, thanks for the explanation. And please accept my apologies for introducing breaking changes.

Your understanding is correct.

PR #1722, opened 2023-09-19
Add new RSpec/RepeatedSubjectCall cop

drcapulet:
This adds a cop for catching uses of `subject` whereby you forget that it is memoized, often when trying to test idempotence:

```ruby
it do
  expect { subject }.to change(Webhook.count).by(1)

  expect { subject }.to not_change(Webhook.count)
end
```

Had a hard time coming up with a succinct, descriptive name so open to suggestions.

- [x] Added tests.
- [x] Updated documentation.


lib/rubocop/cop/rspec/repeated_subject_call.rb
```diff
@@ -0,0 +1,60 @@
...
+      #     expect { my_method }.to change { A.count }
+      #     expect { my_method }.to not_change { A.count }
+      #   end
+      #
+      class RepeatedSubjectCall < Base
+        BLOCK_MSG = 'Calls to subject are memoized, this block is misleading'
+
+        # @!method subject_calls(node)
+        def_node_search :subject_calls, <<~PATTERN
+          (send nil? :subject)
```
pirj, line 31: I think we need to detect named subjects and memoized helpers in general, too.

lib/rubocop/cop/rspec/repeated_subject_call.rb
```diff
@@ -0,0 +1,60 @@
...
+      #   it do
+      #     expect { my_method }.to change { A.count }
+      #     expect { my_method }.to not_change { A.count }
+      #   end
+      #
+      class RepeatedSubjectCall < Base
+        BLOCK_MSG = 'Calls to subject are memoized, this block is misleading'
+
+        # @!method subject_calls(node)
+        def_node_search :subject_calls, <<~PATTERN
```
pirj, line 30: Node search is usually a suspect for non-linear time complexity.

lib/rubocop/cop/rspec/repeated_subject_call.rb
```diff
@@ -0,0 +1,60 @@
...
+          (send nil? :subject)
+        PATTERN
+
+        def on_send(node)
+          return unless node.method?(:it)
```
pirj, line 35: We have helpers to detect if a node is an example, ‘example?(node)’ to my best memory. ‘it’ can be a ‘specify’ etc

pirj (commented):
Thank you so much! This cop, indeed, is very helpful, especially with the ‘not to change’ case!
Let’s shape it up a bit.

lib/rubocop/cop/rspec/repeated_subject_call.rb
drcapulet: Agreed - wasn't sure on the best way to achieve that especially given the complex searching we already do below with just `subject`.

lib/rubocop/cop/rspec/repeated_subject_call.rb
drcapulet: Certainly not ideal, but any other suggestions on how to detect the 2nd+ `subject` call(s)?

lib/rubocop/cop/rspec/repeated_subject_call.rb
pirj: [This](https://github.com/rubocop/rubocop-rspec/blob/e609b2a655ed9609ec8cc9287a5c9ad877c6aaf4/lib/rubocop/cop/rspec/subject_stub.rb#L71) should work.

lib/rubocop/cop/rspec/repeated_subject_call.rb
pirj: Fair enough 👍

pirj (approved):
Looks good already.
Leaving it up to you if you want to also cover named subjects, or leave it for a follow-up.
Thank you!

pirj:
Please rebase

pirj:
I’m sorry, I broke something. I’m from my mobile. Can you please check?

pirj:
It's strange, I've added:
```
  RSpec.describe RuboCop::Cop::RSpec::AlignRightLetBrace do
    # rubocop:disable RSpec/ExampleLength
    it 'registers offense for unaligned braces' do
+     expect { subject }.to change(:a)
+     expect { subject }.to change(:a)
```
and the cop didn't register that. 🤔

drcapulet:
@prij Was a little trickier, but updated to support custom subjects!

lib/rubocop/cop/rspec/repeated_subject_call.rb
```diff
@@ -0,0 +1,121 @@
...
+
+        def detect_offense(example_node, subject_node)
+          walker = subject_node
+
+          while walker.parent? && walker.parent != example_node.body
+            walker = walker.parent
+
+            if walker.block_type? && walker.method?(:expect)
+              add_offense(walker, message: BLOCK_MSG)
+              return
```
ydah, line 75: ask) Why is this `return` necessary?

lib/rubocop/cop/rspec/repeated_subject_call.rb
```diff
@@ -0,0 +1,121 @@
...
+      #   # good
+      #   it do
+      #     expect { my_method }.to change { A.count }
+      #     expect { my_method }.to not_change { A.count }
+      #   end
+      #
+      class RepeatedSubjectCall < Base
+        include TopLevelGroup
+
+        BLOCK_MSG = 'Calls to subject are memoized, this block is misleading'
```
ydah, line 29: By setting this variable name to MSG, it is not necessary to pass it as an argument to `add_offense`.

```suggestion
        MSG = 'Calls to subject are memoized, this block is misleading'
```

lib/rubocop/cop/rspec/repeated_subject_call.rb
drcapulet: This is primarily a performance consideration since we're walking down the AST and then back up from a `subject` call to the `expect` block. Without it, we'd walk out of the example - it's unlikely we'd potentially flag something incorrectly but I haven't been able to concoct an example to check that.

lib/rubocop/cop/rspec/repeated_subject_call.rb
drcapulet: Updated, thanks!

lib/rubocop/cop/rspec/repeated_subject_call.rb
ydah: Thanks for the explanation. I also think this `return' makes sense!

ydah (approved):
Looks good! Thank you great work!

spec/rubocop/cop/rspec/repeated_subject_call_spec.rb
```diff
@@ -0,0 +1,91 @@
...
+
+  it 'registers no offenses for block first' do
+    expect_no_offenses(<<~RUBY)
+      RSpec.describe Foo do
+        it do
+          expect { subject }.to change { Foo.count }
+          expect(subject.b).to eq(4)
+        end
+      end
+    RUBY
```
bquorning, line 75: I would be interested in seeing a similar example, but with two `it` blocks:
```rb
RSpec.describe Foo do
  it do
    expect { subject }.to change { Foo.count }
  end

  it do
    expect(subject.b).to eq(4)
  end
end
```
In this case, the first `subject` block _should_ cause an offense, right?

lib/rubocop/cop/rspec/repeated_subject_call.rb
```diff
@@ -0,0 +1,121 @@
...
+            else
+              subjects_used[call.method_name] = true
+            end
+          end
+        end
+
+        def detect_subjects_in_scope(node)
+          node.each_descendant(:block).with_object({}) do |child, h|
+            name = subject?(child)
+            next unless name
```
bquorning, line 109: Returning other than a boolean from a `foo?` is a bit confusing 😅 How about we use the yielded value instead?
```suggestion
            subject?(child) do |name|
```
(Not that yielding from a `foo?` method is less confusing, really 😄)

spec/rubocop/cop/rspec/repeated_subject_call_spec.rb
drcapulet: Added a spec for this - no offense should be raised since those are unique example blocks and `subject` is evaluated once for each.

bquorning: Huh, I had somehow got the impression that `let` and `subject` memoized calls not per example, but per example group. I was wrong 😬

I think the added example still has value though.

bquorning (approved):

ydah (approved):

timon:
This cop seems to bring in a lot of false positives, at least for me.

Here's the failing case:

```ruby

describe LockManager do
  subject(:lock_manager) { described_class }

  describe '.lock!' do
    it 'requires a lock name and TTL in seconds' do # No offense raised
      expect { lock_manager.lock! }.to raise_error(ArgumentError, %r{wrong number of arguments.*required keyword:.*ttl})
    end

    it 'yields to a block if locking is successful' do # No offense raised
      expect { |block| lock_manager.lock!('some_lock', ttl: 1, &block) }.to yield_control
    end


    it 'raises an exception if locking is not successful', :aggregate_failures do
      lock_manager.lock!('some_lock', ttl: 3) do |_lock_held| # No offense raised
        expect do
          expect { |block| lock_manager.lock!('some_lock', ttl: 1, &block) }.not_to yield_control
          # C: RSpec/RepeatedSubjectCall: Calls to subject are memoized, this block is misleading
        end.to raise_error(Redlock::LockError, %r{failed to acquire.*some_lock})
      end
    end
  end
end
```

I find the actual error misleading:

- there is no call to `subject` here
- memoization is actually expected to happen

pirj:
Doesn’t this PR fix this case for you, @timon ?

timon:
No, dependabot PR with 2.27.0 just failed CI for me

No offence was registered for `expect { |block| lock_manager.lock!(…) }` line in previous version

pirj:
You would have to use edge to get the fix from this PR, @timon

timon:
Changelog says it was released today
https://github.com/rubocop/rubocop-rspec/blob/master/CHANGELOG.md#2270-2024-03-01

bquorning:
I think @pirj meant to ask if PR #1822 (and not _this_ PR) fixes the issue. It was merged just about half an hour ago.

timon:
Yes, I added `github: 'rubocop/rubocop-rspec'` to my Gemfile and no offenses were registered

PR #1725, opened 2023-09-21
Add support single quoted string and percent string and heredoc for `RSpec/Rails/HttpStatus`

ydah:
This PR add support single quoted string and percent string and heredoc for `RSpec/Rails/HttpStatus`.

reproduction code
```rb
it { is_expected.to have_http_status '200' }
it { is_expected.to have_http_status %[200] }
it { is_expected.to have_http_status <<~HTTP }
  200
HTTP
```

expected
```rb
it { is_expected.to have_http_status :ok }
it { is_expected.to have_http_status :ok }
it { is_expected.to have_http_status <<~HTTP } # not offense
  200
HTTP
```

actual
```rb
it { is_expected.to have_http_status }
it { is_expected.to have_http_status }
it { is_expected.to have_http_status }
  200
HTTP
```

### DIFF
--- a/lib/rubocop/cop/rspec/rails/http_status.rb
+++ b/lib/rubocop/cop/rspec/rails/http_status.rb
@@ -66,6 +66,8 @@ module RuboCop
 
           def on_send(node)
             http_status(node) do |arg|
+              return if arg.str_type? && arg.heredoc?
+
               checker = checker_class.new(arg)
               return unless checker.offensive?
 
@@ -105,6 +107,10 @@ module RuboCop
               format(MSG, prefer: prefer, current: current)
             end
 
+            def current
+              offense_range.source
+            end
+
             def offense_range
               node
             end
@@ -129,10 +135,6 @@ module RuboCop
               symbol.inspect
             end
 
-            def current
-              node.value.inspect
-            end
-
             private
 
             def symbol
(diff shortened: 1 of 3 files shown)


lib/rubocop/cop/rspec/rails/http_status.rb
```diff
@@ -205,7 +205,7 @@ def number
             end
 
             def normalize_str
-              normalized = node.source.delete('"')
+              normalized = node.source.gsub(/["']/, '')
```
pirj, line 208: One can still get away with ‘%[200]’, heredoc etc? Is there a common approach to get the contents of a string?

pirj: I meant the ‘value’ method instead of regexps on ‘source’.
Example https://github.com/rubocop/rubocop-rspec/blob/865ea173b61cbb0a902cf478e0b24bca7c5751e9/lib/rubocop/cop/rspec/context_wording.rb#L94

Do you think the same approach will be usable here?

ydah: Ah, I see. Thank you very much. How about this?

pirj (approved):
Thank you!

bquorning (approved):

Issue #1720, opened 2023-09-14
Bug in `RSpec/SpecFilePathFormat` inline disables?


dgollahon:
Hi. I tried to upgrade to the latest version and convert my inline disables for `RSpec/FilePath` to `RSpec/SpecFilePathFormat` but they don't seem to work either on blanket disables early in the file or when attached to the end of a `RSpec.describe` line. Either way worked for `RSpec/FilePath`.

bquorning:
@ydah, is this something you’d want to take a look at?

ydah:
Ah, I see, because I changed it to add_global_offense. I'm wondering if this is a file name issue, and `Exclude` in rubocop.yml seems more appropriate for suppression than suppression in inline comments.

ydah:
I thought it would be incongruous to comment on the class name to be tested and invalidate it when it is a file name issue. What do you think?

ydah:
However, I don't feel strongly about it because I think compatibility is important.

hjheath:
I've just hit this. I agree `Exclude` in rubocop.yml is more appropriate for file name issues.

However I'm hitting an issue where the rubocop.yml exclude configuration seems to be ignored for this new cop. Is this expected? Is this the same issue?

dgollahon:
I prefer inline disables in all circumstances because if the file changes it is automatically reflected correctly or removed. Disabling via `Exclude` can easily go stale (it does not validate that the paths exist or still violate the cop). Since this used to work I would prefer it continue to work. I can move to an `Exclude` workflow but I think it's more confusing because most disables do not work this way. If it can only be disabled via `Exclude` it should probably tell you that in the output of the cop.

PR #1726, opened 2023-09-23
Change to be inline disable for `RSpec/SpecFilePathFormat` like `RSpec/FilePath`

ydah:
Fix: https://github.com/rubocop/rubocop-rspec/issues/1720

- [x] Added tests.
- [-] Updated documentation.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## Master (Unreleased)
 
 - Add support single quoted string and percent string and heredoc for `RSpec/Rails/HttpStatus`. ([@ydah])
+- Change to be inline disable for `RSpec/SpecFilePathFormat` like `RSpec/FilePath`. ([@ydah])
 
 ## 2.24.1 (2023-09-23)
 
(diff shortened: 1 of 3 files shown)


ydah:
@dgollahon How about this? Changed to match the behavior of `RSpec/FilePath`.

pirj (approved):
Thank you!

bquorning:
I don’t see any inline disables in the spec. Am I missing something?

pirj:
The problem was that with ‘add_global_offense’ it wasn’t possible to use inline disabling.

Though now when you ask this, I feel that it’s weird to inline disable this cop.

> blanket disables early in the file or when attached to the end of a RSpec.describe line. Either way worked for RSpec/FilePath.

But for the sake of continuity, it feels we are doing the right thing.

Would you guys agree to revert this and use ‘add_global_offense’ in 3.0.0?

bquorning:
@dgollahon actually makes a good case for inline disables in https://github.com/rubocop/rubocop-rspec/issues/1720#issuecomment-1728521808

pirj:
It makes sense to me.
There are still cases for config file disables, e.g. a directory, so one doesn’t have to inline disable each newly-added file.

The approach provided by this PR would allow both inline and config-based, so our users could use it to their preference.

Agree to keep it this way 👍

bquorning (approved):

PR #1731, opened 2023-10-17
RSpec/MetadataStyle: Fix false positive for multiple strings preceding metadata

### LINKED ISSUE #1714: RSpec/MetadataStyle: False positives and syntax-breaking autocorrects
Regression from new `RSpec/MetadataStyle` cop introduced in #1497.

### Reproduction example

Correct syntax as allowed by the rswag-specs gem:

```ruby
      response "404", "unauthorized access", api_docs: false do
        let(:authorization) { another_user.token }

        run_test!
      end
```

This probably triggers because we also have this in our Rubocop config:

```yaml
RSpec:
  Language:
    ExampleGroups:
      Regular:
        - response
```

False positive:

```
spec/requests/api/v0/foobar_spec.rb:30:23: C: [Correctable] RSpec/MetadataStyle: Use hash style for metadata.
      response "404", "unauthorized access", api_docs: false do
                      ^^^^^^^^^^^^^^^^^^^^^
```

Proposed auto-correct:

```ruby
      response "404", api_docs: false, unauthorized access: true do
        let(:authorization) { another_user.token }

        run_test!
      end
```

Notice how the "unauthorized access" string is treated like a symbol.

### Possible fix

Considering that the new cop supports the two styles "symbol" and "hash", it should probably just ignore strings? If you agree this is the right approach, I'm happy to attempt a fix.

franzliedke:
This fixes a false positive in an example group like the following:

```ruby
      response "404", "unauthorized access", api_docs: false do
        let(:authorization) { another_user.token }

        run_test!
      end
```

(This uses DSL from the `rswag` gem.)

Fixes #1714.

- [x] Added tests.
- [ ] Updated documentation.

### DIFF
--- a/lib/rubocop/cop/rspec/metadata_style.rb
+++ b/lib/rubocop/cop/rspec/metadata_style.rb
@@ -45,6 +45,11 @@ module RuboCop
         PATTERN
 
         def on_metadata(symbols, hash)
+          # RSpec example groups accept two string arguments. In such a case,
+          # the rspec_metadata matcher will interpret the second string
+          # argument as a metadata symbol.
+          symbols.shift if symbols.first&.str_type?
+
           symbols.each do |symbol|
             on_metadata_symbol(symbol)
           end
--- a/lib/rubocop/cop/rspec/mixin/metadata.rb
+++ b/lib/rubocop/cop/rspec/mixin/metadata.rb
@@ -30,12 +30,12 @@ module RuboCop
         def on_block(node)
           rspec_configure(node) do |block_var|
             metadata_in_block(node, block_var) do |metadata_arguments|
-              on_matadata_arguments(metadata_arguments)
+              on_metadata_arguments(metadata_arguments)
             end
           end
 
           rspec_metadata(node) do |metadata_arguments|
-            on_matadata_arguments(metadata_arguments)
+            on_metadata_arguments(metadata_arguments)
           end
         end
         alias on_numblock on_block
@@ -46,7 +46,7 @@ module RuboCop
 
         private
 
-        def on_matadata_arguments(metadata_arguments)
+        def on_metadata_arguments(metadata_arguments)
           *symbols, last = metadata_arguments
           hash = nil
           case last&.type
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 - Add support single quoted string and percent string and heredoc for `RSpec/Rails/HttpStatus`. ([@ydah])
 - Change to be inline disable for `RSpec/SpecFilePathFormat` like `RSpec/FilePath`. ([@ydah])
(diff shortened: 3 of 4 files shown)


lib/rubocop/cop/rspec/mixin/metadata.rb
```diff
@@ -46,7 +46,7 @@ def on_metadata(_symbols, _hash)
 
         private
 
-        def on_matadata_arguments(metadata_arguments)
+        def on_metadata_arguments(metadata_arguments)
```
franzliedke, line 49: Drive-by typo fix 😇

lib/rubocop/cop/rspec/metadata_style.rb
```diff
@@ -45,6 +45,11 @@ class MetadataStyle < Base # rubocop:disable Metrics/ClassLength
         PATTERN
 
         def on_metadata(symbols, hash)
+          # RSpec example groups accept two string arguments. In such a case,
+          # the rspec_metadata matcher will interpret the second string
+          # argument as a metadata symbol.
+          symbols.shift if symbols.first&.str_type?
```
franzliedke, line 51: I first fixed this in the `rspec_metadata` matcher in `lib/rubocop/cop/rspec/mixin/metadata.rb` (see individual commits), but this made one test for the `RSpec/SortMetadata` cop fail. Looking at that cop, it seems to respect many more interesting cases, such as dynamic arguments, which could probably create more false positives in the `MetadataStyle` cop.

Example:

```ruby
context "something dynamic based on", variable do
  # ...
end
```

I decided not to dive into these cases and leave it to the AST matcher experts. (They may also be outside the scope of this PR, as this cop didn't worry about them so far.) Feel free to amend this PR if you see an obvious improvement.

franzliedke:
This should fix the tests.

lib/rubocop/cop/rspec/metadata_style.rb
bquorning: I suspect the real solution in mixin/metadata.rb would be to separate which metadata can be passed to `#Examples.all` vs `#ExampleGroups.all`, but I’m not sure I want to go down that rabbit hole 😅 The suggested fix works for me, too.

bquorning (approved):

franzliedke:
@bquorning Thanks so much for the review, I appreciate it! Do I need to do something else (like squashing the commits), or just be patient? 😇

pirj (approved):
Thank you!
I had some ideas how to improve node patterns, but will attempt separately.

bquorning:
> Do I need to do something else (like squashing the commits), or just be patient? 😇

Valid question 😅 If you squash the last 3 commits together, I’ll merge it.

franzliedke:
@bquorning Consider it done. 👍🏼

bquorning (approved):
Thank you 👍🏼

PR #1739, opened 2023-11-14
Refactor for `RSpec/BeforeAfterAll`

ydah:
- Use `Set[...]` instead of  %i[...]
- Use RESTRICT_ON_SEND in def_node_matcher

- [-] Added tests.
- [-] Updated documentation.

### DIFF
--- a/lib/rubocop/cop/rspec/before_after_all.rb
+++ b/lib/rubocop/cop/rspec/before_after_all.rb
@@ -30,11 +30,11 @@ module RuboCop
               '`use_transactional_fixtures` is enabled, then records created ' \
               'in `%<hook>s` are not automatically rolled back.'
 
-        RESTRICT_ON_SEND = %i[before after].freeze
+        RESTRICT_ON_SEND = Set[:before, :after].freeze
 
         # @!method before_or_after_all(node)
         def_node_matcher :before_or_after_all, <<-PATTERN
-          $(send _ {:before :after} (sym {:all :context}))
+          $(send _ RESTRICT_ON_SEND (sym {:all :context}))
         PATTERN
 
         def on_send(node)


lib/rubocop/cop/rspec/before_after_all.rb
```diff
@@ -30,11 +30,11 @@ class BeforeAfterAll < Base
               '`use_transactional_fixtures` is enabled, then records created ' \
               'in `%<hook>s` are not automatically rolled back.'
 
-        RESTRICT_ON_SEND = %i[before after].freeze
+        RESTRICT_ON_SEND = Set[:before, :after].freeze
```
Darhazer, line 33: what's the reason to use a set?
Seems that for the majority of cops we use %i. If we prefer set:
* we need to update the rest of the cops
* we can provide a check in the internal investigations that RESTRICT_ON_SEND is defined as Set

lib/rubocop/cop/rspec/before_after_all.rb
```diff
@@ -30,11 +30,11 @@ class BeforeAfterAll < Base
...
 
         # @!method before_or_after_all(node)
         def_node_matcher :before_or_after_all, <<-PATTERN
-          $(send _ {:before :after} (sym {:all :context}))
+          $(send _ RESTRICT_ON_SEND (sym {:all :context}))
```
Darhazer, line 37: makes sense to reuse 👍

lib/rubocop/cop/rspec/before_after_all.rb
ydah: This was to make it easier to reuse in `def_node_matcher`. If we use `%i`, it would have to be something like `{:#{RESTRICT_ON_SEND.join(' :')}}`, wouldn't it?

Darhazer: Ah I see. Thank you 👍

pirj:
Thank you!

Issue #1756, opened 2023-12-27
`PendingWithoutReason` does not flag violations inside `shared_examples` and similar blocks


robinaugh:
We recently enabled `PendingWithoutReason` but it appears to miss usages of `xcontext` or `xit` inside certain blocks, i.e.:

```ruby
RSpec.describe SomeClass do
  xcontext 'this is a violation' do
    # ...
  end

  shared_examples 'description here' do
    xcontext 'this is not a violation' do
      # ...
    end
  end

  # helper block from papertrail gem
  with_versioning do
    xit 'this is not a violation either' do
      # ...
    end
  end
end
```

The output of `rubocop -V` is below:

```
╰─⠠⠵ rubocop -V
1.59.0 (using Parser 3.2.2.4, rubocop-ast 1.30.0, running on ruby 3.2.2) [arm64-darwin22]
  - rubocop-capybara 2.19.0
  - rubocop-factory_bot 2.24.0
  - rubocop-rails 2.23.1
  - rubocop-rspec 2.25.0
```

pirj:
Thanks for reporting.
Regarding shared groups, we miss a check of the parent [here](https://github.com/rubocop/rubocop-rspec/blob/fb85f8819a94e6d7e70313d89bf2500366642415/lib/rubocop/cop/rspec/pending_without_reason.rb#L106).

Speaking of third-party helpers, see this doc on how third-party gems can [augment our RSpec DSL conf](https://docs.rubocop.org/rubocop-rspec/third_party_rspec_syntax_extensions.html), and [this doc](https://docs.rubocop.org/rubocop-rspec/usage.html#rspec-dsl-configuration) on how to configure it for your project.

Would you like to send a PR to fix the shared example false negative?

robinaugh:
> Thanks for reporting. Regarding shared groups, we miss a check of the parent [here](https://github.com/rubocop/rubocop-rspec/blob/fb85f8819a94e6d7e70313d89bf2500366642415/lib/rubocop/cop/rspec/pending_without_reason.rb#L106).
> 
> Speaking of third-party helpers, see this doc on how third-party gems can [augment our RSpec DSL conf](https://docs.rubocop.org/rubocop-rspec/third_party_rspec_syntax_extensions.html), and [this doc](https://docs.rubocop.org/rubocop-rspec/usage.html#rspec-dsl-configuration) on how to configure it for your project.
> 
> Would you like to send a PR to fix the shared example false negative?

[...]

PR #1757, opened 2023-12-27
Ensure PendingWithoutReason can detect violations inside shared groups

robinaugh:
This fixes https://github.com/rubocop/rubocop-rspec/issues/1756 by allowing `PendingWithoutReason` to flag violations that are inside shared groups (i.e. `shared_examples`).

I've confirmed that this change can also support the other edge case mentioned in #1756 (using Papertrail's `with_versioning` helper and other arbitrary blocks) if my project configuration is updated to map that helper as if it is a shared group, i.e.

```yml
RSpec:
  Language:
    SharedGroups:
      Examples:
        - with_versioning
```

- [x] Added tests.
- [x] Updated documentation.

### DIFF
--- a/lib/rubocop/cop/rspec/pending_without_reason.rb
+++ b/lib/rubocop/cop/rspec/pending_without_reason.rb
@@ -103,7 +103,7 @@ module RuboCop
           on_pending_by_metadata(node)
           return unless (parent = parent_node(node))
 
-          if example_group?(parent) || block_node_example_group?(node)
+          if spec_group?(parent) || block_node_example_group?(node)
             on_skipped_by_example_method(node)
             on_skipped_by_example_group_method(node)
           elsif example?(parent)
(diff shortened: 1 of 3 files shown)


lib/rubocop/cop/rspec/pending_without_reason.rb
```diff
@@ -103,7 +103,7 @@ def on_send(node)
           on_pending_by_metadata(node)
           return unless (parent = parent_node(node))
 
-          if example_group?(parent) || block_node_example_group?(node)
+          if shared_parent?(parent) || block_node_example_group?(node)
```
pirj, line 106: You can use [spec_group](https://github.com/rubocop/rubocop-rspec/blob/fb85f8819a94e6d7e70313d89bf2500366642415/lib/rubocop/rspec/language.rb#L38) here

pirj (approved):
Please add a changeling entry and this is good to go. Thank you!

lib/rubocop/cop/rspec/pending_without_reason.rb
robinaugh: Updated!

bquorning (approved):

ydah (approved):
Thank you!

PR #1758, opened 2023-12-30
Add Ruby 3.3 to CI matrix

m-nakamura145:
This PR adds Ruby 3.3 to CI matrix. It runs green on local.

- [x] Added tests.
- [x] Updated documentation.


pirj (approved):
Thank you!

pirj:
Wdyt of dropping 2.7 as it’s EOL, and also JRuby? @bquorning @Darhazer @ydah

bquorning:
We usually only drop support for a Ruby version when RuboCop itself does, and I think we should continue doing it that way.

I’m fine with dropping the CI run for JRuby. I cannot remember one single time where specs failed for JRuby but passed for C Ruby. Of course we would keep _supporting_ JRuby, but just treat problems via bug reports.

PR #1761, opened 2024-01-03
Use squiggly heredocs wherever possible

bquorning:
Most new code uses the squiggly heredoc, but in a lot of places we were still using the "old" `<<-` heredoc.

I have seen pull requests changing the heredoc style from `<<-` to `<<~` and I don't like that :-) I prefer having style changes separate from  "real" code changes.

I have not checked if any of the specs were actually relying on the leading spaces in heredoc content, so reviewers may need to take a quick glance at that.

- [ ] Added tests.
- [ ] Updated documentation.


ydah (approved):
Thank you!

PR #1764, opened 2024-01-04
Fix a false positive for RSpec/ExpectActual with rspec-rails

### LINKED ISSUE #759: RSpec/ExpectActual false positive against routing specs (when using a negated matcher)
On our project we have a test as follows:
```ruby
expect(get: :show, params: { id: 'null', format: :json }).not_to be_routable
```

Which is in turn raising a cop:
```ruby
C: RSpec/ExpectActual: Provide the actual you are testing to expect(...).
          expect(get: :show, params: { id: 'null', format: :json }).not_to be_routable
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

I understand that this is very similar to #198. Our `route_to` tests work fine though.

Useful info:
```
$ bundle exec rubocop -v
0.64.0
$ bundle exec rails -v
Rails 5.2.3
$ ruby -v
ruby 2.6.2p47 (2019-03-13 revision 67232) [x86_64-darwin17]
```

naveg:
rspec-rails comes with routing matchers that use literals in the `expect()`.

Fixes #759

- [x] Added tests.
- [x] Updated documentation.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## Master (Unreleased)
 
 - Support correcting `assert_nil` and `refute_nil` to `RSpec/Rails/MinitestAssertions`. ([@G-Rath])
+- Fix a false positive for `RSpec/ExpectActual` when used with rspec-rails routing matchers. ([@naveg])
 
 ## 2.26.1 (2024-01-05)
 
@@ -894,6 +895,7 @@ Compatibility release so users can upgrade RuboCop to 0.51.0. No new features.
 [@mockdeep]: https://github.com/mockdeep
 [@mothonmars]: https://github.com/MothOnMars
 [@mvz]: https://github.com/mvz
+[@naveg]: https://github.com/naveg
 [@nc-holodakg]: https://github.com/nc-holodakg
 [@nevir]: https://github.com/nevir
 [@ngouy]: https://github.com/ngouy
--- a/lib/rubocop/cop/rspec/expect_actual.rb
+++ b/lib/rubocop/cop/rspec/expect_actual.rb
@@ -50,7 +50,8 @@ module RuboCop
           regexp
         ].freeze
 
-        SUPPORTED_MATCHERS = %i[eq eql equal be].freeze
+        SKIPPED_MATCHERS = %i[route_to be_routable].freeze
+        CORRECTABLE_MATCHERS = %i[eq eql equal be].freeze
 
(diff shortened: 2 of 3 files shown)


pirj (approved):
Perfect, thank you!

lib/rubocop/cop/rspec/expect_actual.rb
```diff
@@ -66,6 +67,8 @@ class ExpectActual < Base
 
         def on_send(node)
           expect_literal(node) do |actual, matcher, expected|
+            next if SKIPPED_MATCHERS.include?(matcher)
+
             add_offense(actual.source_range) do |corrector|
               next unless SUPPORTED_MATCHERS.include?(matcher)
```
bquorning, line 73: I was a bit confused that we’d need both “skipped matchers” and “supported matchers”. Perhaps the supported matchers should be renamed “correctable matchers”? 😅

lib/rubocop/cop/rspec/expect_actual.rb
```diff
@@ -50,6 +50,7 @@ class ExpectActual < Base
           regexp
         ].freeze
 
+        SKIPPED_MATCHERS = %i[route_to be_routable].freeze
```
bquorning, line 53: Since we are planning to extract the rspec-rails specific parts from rubocop-rspec, I don’t think we should hardcode these matchers names. Can we either have (the non-existing) rubocop-rspec-rails inject them, or maybe just add a `SkippedMatchers` to the configuration?

naveg: Could we wait to see how the rails extraction takes shape before doing that? It's true that this is rails-specific, but it's also highly unlikely to matter, since we'd only miss adding offenses for non-rails matchers that happened to be named "route_to" or "be_routable".

bquorning (approved):

CHANGELOG.md
```diff
@@ -5,6 +5,7 @@
 ## 2.26.1 (2024-01-05)
 
 - Fix an error for `RSpec/SharedExamples` when using examples without argument. ([@ydah])
+- Fix a false positive for `RSpec/ExpectActual` when used with rspec-rails routing matchers. ([@naveg])
```
bquorning, line 8: Please move this line to the `Unreleased` section :-)

naveg: woops, fixed

Issue #1765, opened 2024-01-05
v2.26.0 regression: `undefined method 'sym_type?' for nil:NilClass`


jtannas:
On the newly release v2.26.0 version, rubocop-rspec encounters an error whenever inspecting the command `run_test!` in my code

```ruby
        context 'when the user does not agree to the terms of use' do
          let(:body) { valid_body.merge(tosAcceptedAt: false) }

          run_test!
        end
```

Environment information: `1.59.0 (using Parser 3.2.2.4, rubocop-ast 1.30.0, running on ruby 3.1.2) [x86_64-linux]`

<details><summary>Stack Trace</summary>
<p>

```
An error occurred while RSpec/SharedExamples cop was inspecting /home/jtannas/foo-project/backend/spec/requests/api/v1/users/url_spec.rb:87:10.
undefined method `sym_type?' for nil:NilClass

            ast_node.sym_type?
                    ^^^^^^^^^^
/home/jtannas/.rbenv/versions/3.1.2/lib/ruby/gems/3.1.0/gems/rubocop-rspec-2.26.0/lib/rubocop/cop/rspec/shared_examples.rb:72:in `offense?'
/home/jtannas/.rbenv/versions/3.1.2/lib/ruby/gems/3.1.0/gems/rubocop-rspec-2.26.0/lib/rubocop/cop/rspec/shared_examples.rb:57:in `block in on_send'
/home/jtannas/.rbenv/versions/3.1.2/lib/ruby/gems/3.1.0/gems/rubocop-rspec-2.26.0/lib/rubocop/cop/rspec/shared_examples.rb:59:in `shared_examples'
/home/jtannas/.rbenv/versions/3.1.2/lib/ruby/gems/3.1.0/gems/rubocop-rspec-2.26.0/lib/rubocop/cop/rspec/shared_examples.rb:55:in `on_send'
/home/jtannas/.rbenv/versions/3.1.2/lib/ruby/gems/3.1.0/gems/rubocop-1.59.0/lib/rubocop/cop/commissioner.rb:107:in `public_send'
/home/jtannas/.rbenv/versions/3.1.2/lib/ruby/gems/3.1.0/gems/rubocop-1.59.0/lib/rubocop/cop/commissioner.rb:107:in `block (2 levels) in trigger_responding_cops'
[...]

PR #1768, opened 2024-01-05
Fix an error for `RSpec/SharedExamples` when using examples without argument

ydah:
Fix: https://github.com/rubocop/rubocop-rspec/issues/1765

- [x] Added tests.
- [-] Updated documentation.

### DIFF
--- a/lib/rubocop/cop/rspec/shared_examples.rb
+++ b/lib/rubocop/cop/rspec/shared_examples.rb
@@ -53,7 +53,7 @@ module RuboCop
 
         def on_send(node)
           shared_examples(node) do
-            ast_node = node.first_argument
+            next unless (ast_node = node.first_argument)
             next unless offense?(ast_node)
 
             checker = new_checker(ast_node)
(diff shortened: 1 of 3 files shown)


Tabby:
Thanks @ydah, you beat me to it :)

I can confirm that this fixes the issue in the repo where I was seeing it

bquorning (approved):
Thank you

lib/rubocop/cop/rspec/shared_examples.rb
```diff
@@ -53,7 +53,7 @@ class SharedExamples < Base
 
         def on_send(node)
           shared_examples(node) do
-            ast_node = node.first_argument
+            next unless (ast_node = node.first_argument)
```
Darhazer, line 56: question? Would we instead add argument capturing in the node_matches and not match on shared groups / includes with no argument?

bquorning: Good idea. Sorry I merged too fast 😅

PR #1770, opened 2024-01-05
Remove unnecessary constants

ydah:
Perhaps, but it appears to be a constant that is not used now.

- [-] Added tests.
- [-] Updated documentation.

### DIFF
--- a/lib/rubocop/cop/rspec/message_expectation.rb
+++ b/lib/rubocop/cop/rspec/message_expectation.rb
@@ -29,7 +29,6 @@ module RuboCop
 
         MSG = 'Prefer `%<style>s` for setting message expectations.'
 
-        SUPPORTED_STYLES = %w[allow expect].freeze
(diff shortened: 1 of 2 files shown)


ydah:
```
❯ find . -type f -print | xargs grep 'SUPPORTED_STYLES' 
./lib/rubocop/cop/rspec/message_expectation.rb:        SUPPORTED_STYLES = %w[allow expect].freeze
./lib/rubocop/cop/rspec/message_spies.rb:        SUPPORTED_STYLES = %w[have_received receive].freeze
```

Darhazer (approved):

bquorning:
Where do they come from? I found the same constant mentioned in rubocop, but didn’t find out _why_.

bquorning (approved):
I found SUPPORTED_STYLES mentioned in https://github.com/rubocop/rubocop/pull/693, but only in specs 🤷🏼

PR #1772, opened 2024-01-05
Mark `RSpec/RedundantPredicateMatcher` as `Safe: false`

koic:
`RSpec/RedundantPredicateMatcher` cop is marked as unsafe because false positives occur when `be_start_with` is used as a legitimate predicate matcher for `start_with?` method.

e.g., https://github.com/rubocop/rubocop-ast/commit/f9e6682

- [ ] Added tests.
- [x] Updated documentation.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Master (Unreleased)
 
+- Mark `RSpec/RedundantPredicateMatcher` as `Safe: false`. ([@koic])
+
 ## 2.26.1 (2024-01-05)
 
 - Fix an error for `RSpec/SharedExamples` when using examples without argument. ([@ydah])
--- a/config/default.yml
+++ b/config/default.yml
@@ -774,6 +774,7 @@ RSpec/RedundantAround:
 RSpec/RedundantPredicateMatcher:
   Description: Checks for redundant predicate matcher.
   Enabled: pending
+  Safe: false
   VersionAdded: '2.26'
   Reference: https://www.rubydoc.info/gems/rubocop-rspec/RuboCop/Cop/RSpec/RedundantPredicateMatcher
 
--- a/docs/modules/ROOT/pages/cops_rspec.adoc
+++ b/docs/modules/ROOT/pages/cops_rspec.adoc
@@ -4500,13 +4500,19 @@ end
 | Enabled by default | Safe | Supports autocorrection | Version Added | Version Changed
 
 | Pending
-| Yes
-| Yes
+| No
+| Yes (Unsafe)
 | 2.26
 | -
 |===
 
 Checks for redundant predicate matcher.
+
+=== Safety
+
+This cop is marked as unsafe because false positives occur
+when `be_start_with` is used as a legitimate predicate matcher
+for `start_with?` method.
 
 === Examples
 
--- a/lib/rubocop/cop/rspec/redundant_predicate_matcher.rb
+++ b/lib/rubocop/cop/rspec/redundant_predicate_matcher.rb
@@ -5,6 +5,11 @@ module RuboCop
     module RSpec
       # Checks for redundant predicate matcher.
       #
+      # @safety
+      #   This cop is marked as unsafe because false positives occur
+      #   when `be_start_with` is used as a legitimate predicate matcher
+      #   for `start_with?` method.
+      #
       # @example
       #   # bad
       #   expect(foo).to be_exist(bar)


pirj:
We’ve had this discussion recently with @bquorning and I still think that we should not mark cops as unsafe for any period of time for a reason of a bug causing incorrect inspection or autocorrection. See #1687. I don’t think we’ll ever fix those issues by muting them.

bquorning:
Maybe I misunderstood the issue. But we have [a spec](https://github.com/rubocop/rubocop-rspec/blob/297a3a2e48c7c1b68d6a1535f1d4927a2cfb27d1/spec/rubocop/cop/rspec/redundant_predicate_matcher_spec.rb#L141-L150) for correcting `be_start_with` to `start_with`.

pirj:
The problem as I understand it is that rspec-expectations’ [start_with/end_with](https://github.com/rspec/rspec-expectations/blob/main/lib/rspec/matchers/built_in/start_or_end_with.rb) matcher won’t work for ProcessedSource custom implementation of start_with?/end_with?
I suggest to still flag those as potential offences, but not autocorrect those two matchers.
However, let’s stretch our imagination as far as overriding respond_to? in a way that the [respond_to](https://github.com/rspec/rspec-expectations/blob/main/lib/rspec/matchers/built_in/respond_to.rb) matcher won’t work. Should we consider this?

Why doesn’t the ProcessedSource’s implementation work with the regular matcher?

bquorning:
The regular matcher seems to _not_ call `#start_with?` but rather `#[]` with two arguments (start and length). Here’s the source for those methods in ProcessedSource:
```rb
      def [](*args)
        lines[*args]
      end

      def start_with?(string)
        return false if self[0].nil?

        self[0].start_with?(string)
      end
```
It seems that [`#[]`](https://github.com/rubocop/rubocop-ast/blob/f9e6682c49e70a8efaea4b339c85d554d9e32d56/lib/rubocop/ast/processed_source.rb#L63-L65) will return lines, and [`#start_with?`](https://github.com/rubocop/rubocop-ast/blob/f9e6682c49e70a8efaea4b339c85d554d9e32d56/lib/rubocop/ast/processed_source.rb#L141-L145) will compare characters. This is not compatible with the `String` implementation of those two methods that rspec-expectations seem to expect.

And yes, we should probably consider if other of the `RedundantPredicateMatcher` corrections are unsafe too.

I would suggest changing the specs to e.g. `expect(processed_source.start_with?('start')).to be(false)`, I don’t think the predicate matchers does anything good for readability in this case.

pirj:
Speaking in defence of [the RSpec matchers in wuestion](https://github.com/rspec/rspec-expectations/blob/main/spec/rspec/matchers/built_in/start_and_end_with_spec.rb), they support types beyond just strings.

The decision to (manually) delegate the start_with? to the first line is justified. Even though it may be surprising if used with a multi-line string/regexp.

There’s yet another thing that makes me suspicious regarding `starts_with?`, and it’s [this processing](https://github.com/rubocop/rubocop-ast/blob/f9e6682c49e70a8efaea4b339c85d554d9e32d56/lib/rubocop/ast/processed_source.rb#L49). Is this a potential performance optimization?

tdeo:
@pirj could you please elaborate a bit on the decision to close this while there seem to be known unsafe cases? This one seems definitely unsafe by design and not because of a bug or improper handling of edge-cases.

pirj:
We can trick cops into suggesting contextually undafe corrections by using the near-infinite flexibility of Ruby.
I disagree that we should defensively mark cops as unsafe for all such situations.

I admit that I am biased against the example in question, and the code looks sub-par. Also marking cops unsafe is my pet peeve.
You may convince me in this specific situation with more examples, but at the moment I have a gut feeling that the current state of affairs is how it should remain.

pirj:
I admit I have better things to apply my time than to start a RuboCop-wide discussion on the unsafeness topic and to refactor rubocop-ast. I wish someone did that for me. In the meanwhile I’ll try to defend the RSpec department from being disabled one by one. I hope noone will think I’m challenging them :smile:

tdeo:
> We can trick cops into suggesting contextually undafe corrections by using the near-infinite flexibility of Ruby.

Agreed that you can probably make every cop unsafe if you try to figure out the specific counter-example that will break things. 

> I disagree that we should defensively mark cops as unsafe for all such situations.

But we're several people having ran into that without trying at all

> I admit that I am biased against the example in question, and the code looks sub-par. Also marking cops unsafe is my pet peeve. You may convince me in this specific situation with more examples, but at the moment I have a gut feeling that the current state of affairs is how it should remain.
> I admit I have better things to apply my time than to start a RuboCop-wide discussion on the unsafeness topic and to refactor rubocop-ast. I wish someone did that for me. In the meanwhile I’ll try to defend the RSpec department from being disabled one by one. I hope noone will think I’m challenging them 😄

I gladly admin that the advantage with invalid autocorrections of RSpec cops is that the tests break instead of production code 😅

pirj:
> ran into that without trying at all

This happens with existing code. I would prefer fixing the implementation of ProcessedSource.
For the code created in the future, the cop will cue towards an implementation that is more compatible to String/Hash etc.

> tests break instead of production code

Good point. Sometime tests may become pass-all deceptive, though.

Issue #1517, opened 2022-12-13
Support for more than similar methods `RSpec/Rails/MinitestAssertions`


ydah:
Support for more than similar methods `RSpec/Rails/MinitestAssertions`

Suggested by @pirj [here](https://github.com/rubocop/rubocop-rspec/pull/1489#pullrequestreview-1184752297).

>I dared to change "resolves" to "partially addresses" as I had more than just assert_equal in mind:
assert_empty - to be_empty
assert_in_delta/assert_in_epsilon - to be_within(...).of(...)
and more (Minitest, RSpec)

ydah:
@PIJI Oh, sorry. I menshioned you by mistake.

ydah:
Thank you for re-open 👍

PR #1773, opened 2024-01-05
feat: support correcting `assert_nil` and `refute_nil` to `RSpec/Rails/MinitestAssertions`

G-Rath:
I'm not sure how far folks want to take this (if at all), but I got the impression from #1485 & #1517 that it was desirable to support at least some more assertions, and this felt like a pretty straightforward one to land.

I've deliberately gone the "boring" route of just effectively duplicating everything twice rather than try to use a single set of methods because lines of code are cheap and if more matchers are supported there might be better ways to deduplicate that are not compatible with one or both of these matchers.

- [x] Added tests.
- [x] Updated documentation.

### DIFF
--- a/config/default.yml
+++ b/config/default.yml
@@ -1169,6 +1169,7 @@ RSpec/Rails/MinitestAssertions:
   Description: Check if using Minitest matchers.
   Enabled: pending
   VersionAdded: '2.17'
+  VersionChanged: "<<next>>"
   Reference: https://www.rubydoc.info/gems/rubocop-rspec/RuboCop/Cop/RSpec/Rails/MinitestAssertions
 
 RSpec/Rails/NegationBeValid:
--- a/spec/rubocop/cop/rspec/rails/minitest_assertions_spec.rb
+++ b/spec/rubocop/cop/rspec/rails/minitest_assertions_spec.rb
@@ -1,73 +1,146 @@
 # frozen_string_literal: true
 
 RSpec.describe RuboCop::Cop::RSpec::Rails::MinitestAssertions do
-  it 'registers an offense when using `assert_equal`' do
-    expect_offense(<<~RUBY)
-      assert_equal(a, b)
-      ^^^^^^^^^^^^^^^^^^ Use `expect(b).to eq(a)`.
-    RUBY
+  context 'with equal assertions' do
+    it 'registers an offense when using `assert_equal`' do
+      expect_offense(<<~RUBY)
+        assert_equal(a, b)
+        ^^^^^^^^^^^^^^^^^^ Use `expect(b).to eq(a)`.
+      RUBY
 
-    expect_correction(<<~RUBY)
-      expect(b).to eq(a)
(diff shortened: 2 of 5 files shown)


G-Rath:
Initially I thought it would be best to correct to `be_nil` since that's what other cops already do for `to eq(nil)` but I realised that's why it actually shouldn't because Rubocop will automatically apply that fix on the same pass and then people who have turned off the default cops for `eq(nil) -> be(nil) -> be_nil` will end up with their preferred syntax too.

G-Rath:
oh yeah good point! I've been trailing minitest recently and so had completely forgotten we've got `context` here 🙈

bquorning:
@pirj, could I get a second opinion on this PR, please?

spec/rubocop/cop/rspec/rails/minitest_assertions_spec.rb
```diff
@@ -1,73 +1,146 @@
...
-    RUBY
-
-    expect_correction(<<~RUBY)
-      expect(b).to eq(a)
-    RUBY
-  end
+  context 'with equal assertions' do
+    it 'registers an offense when using `assert_equal`' do
+      expect_offense(<<~RUBY)
+        assert_equal(a, b)
```
pirj, line 7: It was a miss in the original implementation, it’s confusing that `a` is “expected, and `b` is “actual”.
Up to you if you want to fix this along the way or leave out.

pirj (approved):
Fantastic, thank you!

spec/rubocop/cop/rspec/rails/minitest_assertions_spec.rb
G-Rath: Good point - I'll leave it for now since this diff is already pretty big, but happy to do a follow up addressing it

pirj:
Thanks again!

config/default.yml
```diff
@@ -1169,6 +1169,7 @@ RSpec/Rails/MinitestAssertions:
   Description: Check if using Minitest matchers.
   Enabled: pending
   VersionAdded: '2.17'
+  VersionChanged: "<<next>>"
```
ydah, line 1172: This change should not have been made. This should only be changed when the default settings are changed. For example, when you add a configurable option or change the default for a configurable option.

ydah: fix: https://github.com/rubocop/rubocop-rspec/pull/1777

bquorning: Oops, my bad. I should have caught that in review.

Thanks for discovering, and fixing.

G-Rath: for what it's worth I feel like that was required by `bundle exec rake`, but it might have been I just messed something up or changed something else I shouldn't have locally to make it think this needed updating 🤷 

I'll keep this in mind for future contributions and let you know if I do find the tasks behaving wrongly

ydah: Never mind! Thank you for your work.

PR #1776, opened 2024-01-10
feat: support `assert_not_equal` and `assert_not_nil` in `RSpec/Rails/MinitestAssertions`

G-Rath:
I forgot about these when doing #1773 - I assume they're ok to include even though they're not technically part of native `minitest`.

I've also very slightly tweaked the wording of the changelog entry for #1773.

- [x] Added tests.
- [x] Updated documentation.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,8 @@
 
 ## Master (Unreleased)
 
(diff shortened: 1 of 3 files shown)


bquorning (approved):

ydah (approved):
Thank you!

PR #1778, opened 2024-01-11
Add support `assert_empty`, `assert_not_empty` and `refute_empty` for `RSpec/Rails/MinitestAssertions`

ydah:
This PR is add support `assert_empty`, `assert_not_empty` and `refute_empty` for `RSpec/Rails/MinitestAssertions`.
Related to #1485 
______________________________________________________________________

Before submitting the PR make sure the following are checked:

- [x] Feature branch is up-to-date with `master` (if not - rebase it).
- [x] Squashed related commits together.
- [x] Added tests.
- [x] Updated documentation.
- [x] Added an entry to the `CHANGELOG.md` if the new code introduces user-observable changes.
- [x] The build (`bundle exec rake`) passes (be sure to run this locally, since it may produce updated documentation that you will need to commit).

PR #1779, opened 2024-01-11
feat: support `*_includes` assertions in `RSpec/Rails/MinitestAssertions`

G-Rath:
~Branched off #1778~

This adds support for the `includes` assertions

related to #1485 

- [x] Added tests.
- [x] Updated documentation.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,8 +3,8 @@
 ## Master (Unreleased)
 
 - Add support for `assert_empty`, `assert_not_empty` and `refute_empty` to `RSpec/Rails/MinitestAssertions`. ([@ydah])
-- Support correcting `assert_nil` and `refute_nil` in `RSpec/Rails/MinitestAssertions`. ([@G-Rath])
-- Support correcting `assert_not_equal` and `assert_not_equal` in `RSpec/Rails/MinitestAssertions`. ([@G-Rath])
+- Support correcting `*_includes` assertions in `RSpec/Rails/MinitestAssertions`. ([@G-Rath])
+- Support correcting `assert_not_equal` and `assert_not_nil` in `RSpec/Rails/MinitestAssertions`. ([@G-Rath])
 - Fix a false positive for `RSpec/ExpectActual` when used with rspec-rails routing matchers. ([@naveg])
 - Add new `RSpec/RepeatedSubjectCall` cop. ([@drcapulet])
 
--- a/lib/rubocop/cop/rspec/rails/minitest_assertions.rb
+++ b/lib/rubocop/cop/rspec/rails/minitest_assertions.rb
@@ -10,6 +10,7 @@ module RuboCop
         #   # bad
         #   assert_equal(a, b)
         #   assert_equal a, b, "must be equal"
+        #   assert_not_includes a, b
         #   refute_equal(a, b)
         #   assert_nil a
         #   refute_empty(b)
@@ -17,6 +18,7 @@ module RuboCop
         #   # good
         #   expect(b).to eq(a)
         #   expect(b).to(eq(a), "must be equal")
+        #   expect(a).not_to include(b)
         #   expect(b).not_to eq(a)
         #   expect(a).to eq(nil)
         #   expect(a).not_to be_empty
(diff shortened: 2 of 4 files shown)


lib/rubocop/cop/rspec/rails/minitest_assertions.rb
```diff
@@ -30,66 +30,146 @@ class MinitestAssertions < Base
...
-            minitest_equal_assertion(node) do |expected, actual, fail_message|
-              prefer = replace_equal_assertion(node, expected, actual,
-                                               fail_message.first)
-              add_an_offense(node, prefer)
+          # @!method minitest_empty(node)
+          def_node_matcher :minitest_empty, <<~PATTERN
+            (send nil? {:assert_empty :assert_not_empty :refute_empty} $_ $_?)
+          PATTERN
+
+          def on_send(node) # rubocop:disable Metrics/MethodLength, Metrics/AbcSize
```
G-Rath, line 65: I've just disabled `AbcSize` for now because I don't want to go too far from #1778 until it's landed, but I think a good next step for improving this could be to move the node matchers /  block into the assertion classes too, so it becomes something like:

```
EqualAssertion.check(node)
IncludesAssertion.check(node)
NilAssertion.check(node)
EmptyAssertion.check(node)
# ...
```

(which you could probably take further by having an ASSERTION_CHECKERS array, and then the rest of this rule just iterates over that in a few places to pull through node matchers, restrict sends, `check`, etc)

CHANGELOG.md
```diff
@@ -2,7 +2,8 @@
 
 ## Master (Unreleased)
 
-- Support correcting `assert_nil` and `refute_nil` in `RSpec/Rails/MinitestAssertions`. ([@G-Rath])
+- Add support `assert_empty`, `assert_not_empty` and `refute_empty` for `RSpec/Rails/MinitestAssertions`. ([@ydah])
+- Support correcting `*_includes` assertions in `RSpec/Rails/MinitestAssertions`. ([@G-Rath])
 - Support correcting `assert_not_equal` and `assert_not_equal` in `RSpec/Rails/MinitestAssertions`. ([@G-Rath])
```
pirj, line 7: I’ve just noticed a duplicate matcher name here

lib/rubocop/cop/rspec/rails/minitest_assertions.rb
```diff
@@ -30,66 +30,146 @@ class MinitestAssertions < Base
           RESTRICT_ON_SEND = %i[
             assert_equal
             assert_not_equal
-            refute_equal
+            assert_includes
+            assert_not_includes
```
pirj, line 34: Does it exist?

lib/rubocop/cop/rspec/rails/minitest_assertions.rb
```diff
@@ -28,11 +30,14 @@ class MinitestAssertions < Base
           RESTRICT_ON_SEND = %i[
             assert_equal
             assert_not_equal
+            assert_includes
+            assert_not_includes
             assert_nil
             assert_not_nil
```
pirj, line 36: There’s an “a-ha” moment, since all those assert_no_* are from test-unit, not minitest.

But I’d say we don’t care much if there’s a corresponding native RSpec expectation/matcher we can suggest using instead.

test-unit assertions https://www.rubydoc.info/gems/test-unit/2.3.2/Test/Unit/Assertions
Minitest assertions https://ruby-doc.org/stdlib-3.0.0/libdoc/minitest/rdoc/Minitest/Assertions.html

pirj (commented):
Looks good, thank you!

lib/rubocop/cop/rspec/rails/minitest_assertions.rb
G-Rath: It's defined by Rails: https://github.com/rails/rails/blob/main/activesupport/lib/active_support/test_case.rb#L206

lib/rubocop/cop/rspec/rails/minitest_assertions.rb
G-Rath: Yea though also they're defined by Rails which is actually why I was including them: https://github.com/rails/rails/blob/main/activesupport/lib/active_support/test_case.rb#L206

CHANGELOG.md
G-Rath: 🤦 I'll just fix that in this if you're ok with that

ydah (approved):
Looks good. Thank you

PR #1780, opened 2024-01-13
feat: support `*_instance_of` in `RSpec/Rails/MinitestAssertions`

G-Rath:
Related to #1485

I've _not_ included this in the cops documentation as I figure ultimately documenting every single form of matcher translation could end up being pretty big and this is a straightforward one - happy to add it though if folks disagree/don't mind the length.

I've also refactored the assertion classes so they extend from a basic parent that handles the generation of the actual matchers - I've kept this as a separate commit for easier review and in case folks want that landed in a different PR (or not at all).

- [x] Added tests.
- [x] Updated documentation.

### DIFF
--- a/lib/rubocop/cop/rspec/rails/minitest_assertions.rb
+++ b/lib/rubocop/cop/rspec/rails/minitest_assertions.rb
@@ -82,17 +82,17 @@ module RuboCop
             end
 
             minitest_includes(node) do |collection, expected, failure_message|
-              on_assertion(node, IncludesAssertion.new(collection, expected,
+              on_assertion(node, IncludesAssertion.new(expected, collection,
                                                        failure_message.first))
             end
 
             minitest_nil(node) do |actual, failure_message|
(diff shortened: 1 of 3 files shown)


lib/rubocop/cop/rspec/rails/minitest_assertions.rb
```diff
@@ -96,79 +109,75 @@ def message(preferred)
...
-                  " #{@fail_message.source})"
-              end
+          # :nodoc:
+          class InstanceOfAssertion < BasicAssertion
+            def negated?(node)
+              !node.method?(:assert_instance_of)
+            end
+
+            def assertion
+              "be_an_instance_of(#{@expected})"
```
G-Rath, line 147: I went with this form since it's the grammatically correct version and the "original" (`be_instance_of` is an alias)

pirj (approved):
Nice refactoring and addition! Thank you

ydah (approved):
Thank you!

PR #1784, opened 2024-01-19
fix: support asserts with messages in `Rspec/BeEmpty`

G-Rath:
Adds support for when `contain_exactly` and `match_array are used with a message

- [x] Added tests.
- [x] Updated documentation.

### DIFF
--- a/lib/rubocop/cop/rspec/be_empty.rb
+++ b/lib/rubocop/cop/rspec/be_empty.rb
@@ -28,6 +28,7 @@ module RuboCop
               (send nil? :match_array (array))
               (send nil? :contain_exactly)
             }
+            _?
           )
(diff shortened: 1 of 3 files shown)


lib/rubocop/cop/rspec/be_empty.rb
```diff
@@ -28,6 +28,7 @@ class BeEmpty < Base
               (send nil? :match_array (array))
               (send nil? :contain_exactly)
             }
+            $_?
```
pirj, line 31: $ seems unnecessary

bquorning (approved):
Not sure what the failing CI check is about.

ydah (approved):
Thank you!

PR #1786, opened 2024-01-20
Add support for `assert_true` and `assert_false` to `RSpec/Rails/MinitestAssertions`

ydah:
related to #1485 

This PR add support for `assert_true` and `assert_false` to `RSpec/Rails/MinitestAssertions`.

- [x] Added tests.
- [x] Updated documentation.

### DIFF
--- a/config/default.yml
+++ b/config/default.yml
@@ -1172,7 +1172,7 @@ RSpec/Rails/InferredSpecType:
     views: view
 
 RSpec/Rails/MinitestAssertions:
-  Description: Check if using Minitest matchers.
+  Description: Check if using Minitest-like matchers.
   Enabled: pending
   VersionAdded: '2.17'
   Reference: https://www.rubydoc.info/gems/rubocop-rspec/RuboCop/Cop/RSpec/Rails/MinitestAssertions
--- a/docs/modules/ROOT/pages/cops_rspec_rails.adoc
+++ b/docs/modules/ROOT/pages/cops_rspec_rails.adoc
@@ -252,7 +252,10 @@ end
 | -
 |===
 
-Check if using Minitest matchers.
+Check if using Minitest-like matchers.
+
+Check the use of minitest-like matchers
+starting with `assert_` or `refute_`.
 
 === Examples
 
@@ -265,6 +268,8 @@ assert_not_includes a, b
 refute_equal(a, b)
 assert_nil a
 refute_empty(b)
+assert_true(a)
+assert_false(a)
 
 # good
 expect(b).to eq(a)
@@ -273,6 +278,8 @@ expect(a).not_to include(b)
 expect(b).not_to eq(a)
 expect(a).to eq(nil)
 expect(a).not_to be_empty
+expect(a).to be(true)
+expect(a).to be(false)
 ----
 
 === References
--- a/lib/rubocop/cop/rspec/rails/minitest_assertions.rb
+++ b/lib/rubocop/cop/rspec/rails/minitest_assertions.rb
@@ -4,7 +4,10 @@ module RuboCop
   module Cop
     module RSpec
       module Rails
-        # Check if using Minitest matchers.
+        # Check if using Minitest-like matchers.
+        #
+        # Check the use of minitest-like matchers
+        # starting with `assert_` or `refute_`.
         #
         # @example
         #   # bad
@@ -14,6 +17,8 @@ module RuboCop
         #   refute_equal(a, b)
         #   assert_nil a
         #   refute_empty(b)
+        #   assert_true(a)
+        #   assert_false(a)
         #
         #   # good
         #   expect(b).to eq(a)
@@ -22,6 +27,8 @@ module RuboCop
         #   expect(b).not_to eq(a)
         #   expect(a).to eq(nil)
         #   expect(a).not_to be_empty
+        #   expect(a).to be(true)
+        #   expect(a).to be(false)
         #
         class MinitestAssertions < Base
           extend AutoCorrector
@@ -212,6 +219,46 @@ module RuboCop
 
             def assertion
               'be_empty'
+            end
+          end
+
+          # :nodoc:
+          class TrueAssertion < BasicAssertion
+            MATCHERS = %i[
+              assert_true
+            ].freeze
+
+            # @!method self.minitest_assertion(node)
+            def_node_matcher 'self.minitest_assertion', <<~PATTERN # rubocop:disable InternalAffairs/NodeMatcherDirective
+              (send nil? {:assert_true} $_ $_?)
+            PATTERN
+
+            def self.match(actual, failure_message)
+              new(nil, actual, failure_message.first)
+            end
+
+            def assertion
+              'be(true)'
+            end
+          end
+
+          # :nodoc:
+          class FalseAssertion < BasicAssertion
+            MATCHERS = %i[
+              assert_false
+            ].freeze
+
+            # @!method self.minitest_assertion(node)
+            def_node_matcher 'self.minitest_assertion', <<~PATTERN # rubocop:disable InternalAffairs/NodeMatcherDirective
+              (send nil? {:assert_false} $_ $_?)
+            PATTERN
+
+            def self.match(actual, failure_message)
+              new(nil, actual, failure_message.first)
+            end
+
+            def assertion
+              'be(false)'
             end
           end
 
--- a/spec/rubocop/cop/rspec/rails/minitest_assertions_spec.rb
+++ b/spec/rubocop/cop/rspec/rails/minitest_assertions_spec.rb
@@ -508,6 +508,100 @@ RSpec.describe RuboCop::Cop::RSpec::Rails::MinitestAssertions do
     end
   end
 
+  context 'with boolean assertions' do
+    it 'registers an offense when using `assert_true`' do
+      expect_offense(<<~RUBY)
+        assert_true(a)
+        ^^^^^^^^^^^^^^ Use `expect(a).to be(true)`.
+      RUBY
+
+      expect_correction(<<~RUBY)
+        expect(a).to be(true)
+      RUBY
+    end
+
+    it 'registers an offense when using `assert_true` with no parentheses' do
+      expect_offense(<<~RUBY)
+        assert_true a
+        ^^^^^^^^^^^^^ Use `expect(a).to be(true)`.
+      RUBY
+
+      expect_correction(<<~RUBY)
+        expect(a).to be(true)
+      RUBY
+    end
+
+    it 'registers an offense when using `assert_true` with failure message' do
+      expect_offense(<<~RUBY)
+        assert_true a, "must be true"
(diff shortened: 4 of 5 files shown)


lib/rubocop/cop/rspec/rails/minitest_assertions.rb
```diff
@@ -22,30 +24,23 @@ module Rails
         #   expect(b).not_to eq(a)
         #   expect(a).to eq(nil)
         #   expect(a).not_to be_empty
+        #   expect(a).to be_truthy
+        #   expect(b).to be_falsey
```
koic, line 28: `assert_true` and `assert_false` are methods in Test::Unit and don't exist in Minitest (and ActiveSupport::TestCase):

- Minitest ... https://docs.seattlerb.org/minitest/Minitest/Assertions.html
- Test::Unit ... https://test-unit.github.io/test-unit/ja/Test/Unit/Assertions.html

 Should these be supported in `RSpec/Rails/Minitest`?

If so, `assert_true` should be contrasted with `be(true)`, not `be_truthy`. And `assert_false` should be contrasted with `be(false)`, not `be_falsey`.

docs/modules/ROOT/pages/cops_rspec_rails.adoc
```diff
@@ -265,6 +265,8 @@ assert_not_includes a, b
 refute_equal(a, b)
 assert_nil a
 refute_empty(b)
+assert_true(a)
```
pirj, line 268: The a/b naming for expected/actual, especially when used interchangeably, us getting confusing. Can we use just one for “actual”?

lib/rubocop/cop/rspec/rails/minitest_assertions.rb
```diff
@@ -22,30 +24,23 @@ module Rails
...
-            refute_includes
-            refute_nil
-            refute_empty
-            refute_match
+          ASSERT_SUFFIXES = %w[
+            equal not_equal instance_of not_instance_of includes not_includes
+            match nil not_nil empty not_empty true false
           ].freeze
+          REFUTE_SUFFIXES = %w[
+            equal instance_of includes nil empty match
```
pirj, line 39: This decreases discoverability.
Are we hitting the line limit for this cop? What about grouping assertions per sub-cop (eg one line for “equal” etc)?

lib/rubocop/cop/rspec/rails/minitest_assertions.rb
```diff
@@ -14,6 +14,8 @@ module Rails
         #   refute_equal(a, b)
         #   assert_nil a
         #   refute_empty(b)
+        #   assert_true(a)
+        #   assert_false b
```
pirj, line 18: Does it make sense to also detect plain `assert`/`refute` from Minitest?

pirj (commented):
Looks mostly good.
I agree with @koic that we should auto-correct to stricter be(true)/be(false).

lib/rubocop/cop/rspec/rails/minitest_assertions.rb
pirj: The relationship between test::unit and minitest assertions is not trivial https://github.com/rspec/rspec-core/blob/f8c8880dabd8f0544a6f91d8d4c857c1bd8df903/lib/rspec/core/test_unit_assertions_adapter.rb#L9

The cop, despite the name, is taking care of all of those assertions.

lib/rubocop/cop/rspec/rails/minitest_assertions.rb
koic: This is a technical note. Minitest's `assert` is compatible with `be_truthy`, and `refute` is compatible with `be_falsey`. This means that the behavior of Test::Unit's `assert_true` / `assert_false` and Minitest's `assert` / `refute` are different.

- Test::Unit ... https://github.com/test-unit/test-unit/blob/3.6.1/lib/test/unit/assertions.rb#L1379
- Minitest ... https://github.com/minitest/minitest/blob/v5.21.2/lib/minitest/assertions.rb#L180

pirj: Thanks for bringing this up!

lib/rubocop/cop/rspec/rails/minitest_assertions.rb
G-Rath: I've opened #1790 which goes in another approach to this that should retain the readability

lib/rubocop/cop/rspec/rails/minitest_assertions.rb
```diff
@@ -14,6 +14,8 @@ module Rails
         #   refute_equal(a, b)
         #   assert_nil a
         #   refute_empty(b)
+        #   assert_true(a)
+        #   assert_false(a)
```
koic, line 18: I have already commented in detail, these are Test::Unit APIs, not Minitest APIs. It seems to contradict the description of "`Check if using Minitest matchers`".

spec/rubocop/cop/rspec/rails/minitest_assertions_spec.rb
```diff
@@ -506,6 +506,98 @@
         expect(a).not_to be_empty
       RUBY
     end
+
+    it 'registers an offense when using `assert_true`' do
```
G-Rath, line 510: this is currently within the context for `*_empty` - ideally each assertion class should have its own `context`

lib/rubocop/cop/rspec/rails/minitest_assertions.rb
pirj: We don’t plan to introduce separate cops for teat::unit or rails-specific assertions. It handles all minitest-like assertions.

ydah:
Ah I missed some comments. Hmm. I think we should distinguish between Minitest and Test:: Unit.
I thought it best not to support `assert_true` and `assert_false` in this Cop. It would be better to discuss the extent of this support first. For example, it looks like it could be transferred to a Cop called `RSpec/TestUnitAssertions`.

lib/rubocop/cop/rspec/rails/minitest_assertions.rb
koic: I see, it would probably help to prevent confusion by mentioning such distinctions in the documentation.

G-Rath:
fwiw I currently vote for "have a single cop" because these all seem to be using the same structure and as a downstream user my focus is on "use `expect` consistently" - that doesn't matter if the assertions come from Minitest, Test::Unit, or Rails.

Put another way: I don't think I've yet seen a technical reason for having different cops? i.e. the assertions so far have followed the same logic and structure regardless of where they're actually from.

If people want to rename it to something like `RSpec/PreferExpectInsteadOfAssert`, that'd be fine with me but I think having two cops based on "Minitest vs TestUnit" when they have the same behaviours is just adding more work for ourselves (and to take that a step further: should we have a third cop dedicated to Rails' extension assertions?)

ydah:
@G-Rath Thank you for expressing your opinion.

> Put another way: I don't think I've yet seen a technical reason for having different cops?

In many cases, I think there is no problem to combine them into one cop. However, as the devil's advocate, I would like to deepen this discussion.

For example, the following comments seem to be points where different actions are required. The compatibility of `assert` and `refute` is different between Minitest and Test:: Unit.
- https://github.com/rubocop/rubocop-rspec/pull/1786#discussion_r1460740373

Also, the adapter to Minitest is defined in rspec-rails, which is appropriate for the RSpec/Rails department.
https://github.com/rspec/rspec-rails/blob/main/lib/rspec/rails/adapters.rb#L140

Note, however, that Unit:: Test exists in rspec-core.
https://github.com/rspec/rspec-core/blob/f8c8880dabd8f0544a6f91d8d4c857c1bd8df903/lib/rspec/core/test_unit_assertions_adapter.rb#L9

In other words, it seems odd to enable Cop for the RSpec/Rails department if you are using Unit:: Test-style assertions and not rspec-rails.

G-Rath:
> However, as the devil's advocate, I would like to deepen this discussion.

Yup fully in support of that, and in doing so sooner rather than later.

>In other words, it seems odd to enable Cop for the RSpec/Rails department if you are using Unit:: Test-style assertions and not rspec-rails

Just want to check that I'm understanding what you mean correctly, which is effectively: it's weird for someone who's using just Unit::Test but not Rails, to enable a _Rails_ cop.

If so then yes I agree with that - I'd wondered why this was considered a Rails cop though you've pointed out it's `rspec-rails` that has the Minitest adapter so I'm guessing that is (or was...) the why.

pirj:
> odd to enable Cop for the RSpec/Rails department if you are using Unit:: Test-style assertions and not rspec-rails.

Fair enough, but in the future when we extract rubocop-rspec_rails, we could also extract Rails-specific assertion detectors from this cop, and make the cop open for extension.
We’ll depend on rubocop-rspec from rubocop-rspec_rails, right?

spec/rubocop/cop/rspec/rails/minitest_assertions_spec.rb
pirj:
```suggestion
  end

  context 'with boolean assertions' do
    it 'registers an offense when using `assert_true`' do
```

pirj (approved):
Thank you!

G-Rath:
Probably best to take this to the issue, but there's also Capybara assertions we could be translating, and I'd like to also include [`aggregate_assertions`](https://github.com/tjwp/aggregate_assertions) as it seems very solid and 1:1 with `aggregate_failures`

ydah:
> We’ll depend on rubocop-rspec from rubocop-rspec_rails, right?
Oh, I see. If we needs to extract to rubocop-rspec_rails, maybe we shouldn't add support for this to the RSpec/Rails department's Cop for now?

ydah:
I update this PR.

config/default.yml
```diff
@@ -1172,7 +1172,7 @@ RSpec/Rails/InferredSpecType:
     views: view
 
 RSpec/Rails/MinitestAssertions:
-  Description: Check if using Minitest matchers.
+  Description: Check if using Minitest like matchers.
```
pirj, line 1175:
```suggestion
  Description: Check if using Minitest-like matchers.
```

pirj (approved):
Thank you!

lib/rubocop/cop/rspec/rails/minitest_assertions.rb
```diff
@@ -215,6 +222,46 @@ def assertion
...
+            def_node_matcher 'self.minitest_assertion', <<~PATTERN # rubocop:disable InternalAffairs/NodeMatcherDirective
+              (send nil? {:assert_true} $_ $_?)
+            PATTERN
+
+            def self.match(actual, failure_message)
+              new(nil, actual, failure_message.first)
+            end
+
+            def assertion
+              'eq(true)'
```
bquorning, line 241: Just curious – why not `'be(true)'`?

ydah: Oops, sorry😅 I updated this PR.

bquorning (approved):
Thank you

PR #1789, opened 2024-01-25
feat: support some "predicate" assertions in `RSpec/Rails/MinitestAssertions`

G-Rath:
This implements some support for `assert_predicate` and co, focused on transforming usage with predicate methods (e.g. `valid?`) into their `be_` equivalents.

Predicates that are not symbols or that don't end with `?` are ignored - a follow up PR can expand support for other uses of `assert_predicate`.

I think technically this makes the autofix unsafe because `assert_predicate` uses `send` whereas RSpec uses `send_public` so you could assert a private method which would break when fixed, but I feel like that would generally be considered bad practice anyway and an easy mistake so arguably this could be useful?

- [x] Added tests.
- [x] Updated documentation.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 - Support asserts with messages in `Rspec/BeEmpty`. ([@G-Rath])
 - Add support for `assert_empty`, `assert_not_empty` and `refute_empty` to `RSpec/Rails/MinitestAssertions`. ([@ydah])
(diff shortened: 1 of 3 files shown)


pirj (approved):
Thank you!
I agree that even though we deliberately suggest a potentially unsafe correction, it's for the greater good.

You're on fire with those assertion PRs 🔥

bquorning (approved):

PR #1790, opened 2024-01-26
refactor: encapsulate minitest assertion matchers

G-Rath:
This aims to move all of the specific code for each minitest assertion matcher into the assertion class, so that the outer cop becomes responsible for calling a specific assertion matcher class and reporting, and the inner classes become responsible for handling the actual matching.

Opening as a draft for now to get confirmation that people are ok with the direction of this - I'll squash and do any remaining cleanup afterwards.

- [x] ~Added tests.~
- [x] ~Updated documentation.~
- [x] ~Added an entry to the `CHANGELOG.md` if the new code introduces user-observable changes.~

### DIFF
--- a/lib/rubocop/cop/rspec/rails/minitest_assertions.rb
+++ b/lib/rubocop/cop/rspec/rails/minitest_assertions.rb
@@ -26,115 +26,6 @@ module RuboCop
         class MinitestAssertions < Base
           extend AutoCorrector
 
-          MSG = 'Use `%<prefer>s`.'
-          RESTRICT_ON_SEND = %i[
-            assert_equal
-            assert_not_equal
-            assert_instance_of
-            assert_not_instance_of
-            assert_includes
-            assert_not_includes
-            assert_predicate
-            assert_not_predicate
-            assert_match
-            assert_nil
-            assert_not_nil
-            assert_empty
-            assert_not_empty
-            refute_equal
-            refute_instance_of
-            refute_includes
-            refute_predicate
-            refute_nil
-            refute_empty
-            refute_match
-          ].freeze
-
-          # @!method minitest_equal(node)
-          def_node_matcher :minitest_equal, <<~PATTERN
-            (send nil? {:assert_equal :assert_not_equal :refute_equal} $_ $_ $_?)
-          PATTERN
-
-          # @!method minitest_instance_of(node)
-          def_node_matcher :minitest_instance_of, <<~PATTERN
-            (send nil? {:assert_instance_of :assert_not_instance_of :refute_instance_of} $_ $_ $_?)
-          PATTERN
-
-          # @!method minitest_includes(node)
-          def_node_matcher :minitest_includes, <<~PATTERN
-            (send nil? {:assert_includes :assert_not_includes :refute_includes} $_ $_ $_?)
-          PATTERN
-
-          # @!method minitest_predicate(node)
-          def_node_matcher :minitest_predicate, <<~PATTERN
-            (send nil? {:assert_predicate :assert_not_predicate :refute_predicate} $_ ${sym} $_?)
-          PATTERN
-
-          # @!method minitest_match(node)
-          def_node_matcher :minitest_match, <<~PATTERN
-            (send nil? {:assert_match :refute_match} $_ $_ $_?)
-          PATTERN
-
-          # @!method minitest_nil(node)
-          def_node_matcher :minitest_nil, <<~PATTERN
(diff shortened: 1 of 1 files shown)


lib/rubocop/cop/rspec/rails/minitest_assertions.rb
```diff
@@ -26,115 +26,6 @@ module Rails
...
-              corrector.replace(node, preferred)
-            end
-          end
-
-          def message(preferred)
-            format(MSG, prefer: preferred)
-          end
-
           # :nodoc:
           class BasicAssertion
```
bquorning, line 30: Should we consider adding empty methods to `BasicAssertion` to indicate the interface?
```rb
def negated?
  raise NotImplementedError
end

def assertion
  raise NotImplementedError
end
```

lib/rubocop/cop/rspec/rails/minitest_assertions.rb
```diff
@@ -229,6 +219,48 @@ def assertion
...
+
+            const if const.is_a?(Class) && const.superclass == BasicAssertion
+          end
+
+          RESTRICT_ON_SEND = ASSERTION_MATCHERS.flat_map { |m| m::MATCHERS }
+
+          ASSERTION_MATCHERS.each do |m|
+            name = m.name.split('::').last
+
+            def_node_matcher "minitest_#{name}".to_sym, m::NODE_MATCHER_PATTERN
```
bquorning, line 237: If you add `extend NodePattern::Macros` to `BasicAssertion`, it should be possible to add the node matchers in the `BasicAssertion` subclasses themselves. I’m not 100% sure if it works, but we’d get rid of this loop defining the node patterns, and `#on_send` would also become a lot simpler, e.g.
```rb
def on_send(node)
  ASSERTION_MATCHERS.each do |matcher|
    assertion = matcher.new(node).match
    return if assertion.nil?

    on_assertion(node, assertion)
  end
end
```

G-Rath: ah cool - @pirj made a similar comment over in #1791 but I'd not explored further since I thought it'd require a more complicated change to the constructor, but maybe not.

I'll give that a try but it if turns out too complicated I would prefer to tackle it in a follow-up PR if that's ok

lib/rubocop/cop/rspec/rails/minitest_assertions.rb
```diff
@@ -26,10 +26,47 @@ module Rails
...
+          # :nodoc:
+          class EqualAssertion < BasicAssertion
+            MATCHERS = %i[
+              assert_equal
+              assert_not_equal
+              refute_equal
+            ].freeze
+
+            NODE_MATCHER = <<~PATTERN
+              (send nil? {:assert_equal :assert_not_equal :refute_equal} $_ $_ $_?)
```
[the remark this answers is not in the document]
G-Rath: Yeah I had the same hope but no dice 😞

lib/rubocop/cop/rspec/rails/minitest_assertions.rb
G-Rath: hmm yeah this is turning ugly a bit too ugly for my taste: since `BasicAssertion` is handling a lot of the instance variables, we effectively want an artificial second constructor, and the node matcher method takes a block which makes it messier to do a return (unless I've missed something).

Would you mind sticking with this current layout and I can explore this further as a follow up?

bquorning: Sure, we can stick with the current layout.

I tried finding a suitable implementation, and it’s tricky. I ended up with something like this in the subclasses:
```rb
            def initialize(node)
              pattern(node) do |actual, failure_message|
                super(nil, actual, failure_message.first)
              end
            end
```
The “sometimes calling `super`” smells a bit, and I had to add this method to the `BasicAssertion`:
```
            def match?
              !!@expected || !!@actual
            end
```
to be called instead of `next if assertion.nil?`.

bquorning (approved):

lib/rubocop/cop/rspec/rails/minitest_assertions.rb
G-Rath: Yeah it's annoying how ugly it gets and I'm pretty sure it's all because of `node_matcher` not being on the class... 😅

I've got an idea or two on how to approach it though which I'll explore after this is landed.

pirj (approved):
Thanks!
I hope it will turn out to be possible to eliminate the accumulated meta-programming.

ydah (approved):
Looks great!

PR #1791, opened 2024-01-26
feat: support correcting `*_in_delta` assertions in `RSpec/Rails/MinitestAssertions`

G-Rath:
This is branched off #1790, showing how easy it is to now add a new assertion matcher - until the prior PR is merged, the last commit of this PR should be looked at for review.

- [x] Added tests.
- [x] Updated documentation.

### DIFF
--- a/lib/rubocop/cop/rspec/rails/minitest_assertions.rb
+++ b/lib/rubocop/cop/rspec/rails/minitest_assertions.rb
@@ -127,6 +127,36 @@ module RuboCop
           end
 
           # :nodoc:
+          class InDeltaAssertion < BasicAssertion
+            MATCHERS = %i[
+              assert_in_delta
+              assert_not_in_delta
+              refute_in_delta
+            ].freeze
+
+            # @!method self.minitest_assertion(node)
+            def_node_matcher 'self.minitest_assertion', <<~PATTERN # rubocop:disable InternalAffairs/NodeMatcherDirective
+              (send nil? {:assert_in_delta :assert_not_in_delta :refute_in_delta} $_ $_ $!{sym str}? $_?)
+            PATTERN
+
+            def self.match(expected, actual, delta, failure_message)
+              return nil if delta.empty? && !failure_message.empty?
+
+              new(expected, actual, delta.first, failure_message.first)
+            end
+
+            def initialize(expected, actual, delta, fail_message)
+              super(expected, actual, fail_message)
+
+              @delta = delta&.source || '0.001'
+            end
+
+            def assertion
+              "be_within(#{@delta}).of(#{expected})"
+            end
+          end
+
+          # :nodoc:
           class PredicateAssertion < BasicAssertion
             MATCHERS = %i[
               assert_predicate
--- a/lib/rubocop/cop/rspec/rails/minitest_assertions.rb
+++ b/lib/rubocop/cop/rspec/rails/minitest_assertions.rb
@@ -136,12 +136,10 @@ module RuboCop
 
             # @!method self.minitest_assertion(node)
             def_node_matcher 'self.minitest_assertion', <<~PATTERN # rubocop:disable InternalAffairs/NodeMatcherDirective
-              (send nil? {:assert_in_delta :assert_not_in_delta :refute_in_delta} $_ $_ $!{sym str}? $_?)
+              (send nil? {:assert_in_delta :assert_not_in_delta :refute_in_delta} $_ $_ $_? $_?)
             PATTERN
 
             def self.match(expected, actual, delta, failure_message)
-              return nil if delta.empty? && !failure_message.empty?
-
               new(expected, actual, delta.first, failure_message.first)
             end
 
--- a/spec/rubocop/cop/rspec/rails/minitest_assertions_spec.rb
+++ b/spec/rubocop/cop/rspec/rails/minitest_assertions_spec.rb
@@ -271,6 +271,127 @@ RSpec.describe RuboCop::Cop::RSpec::Rails::MinitestAssertions do
     end
   end
 
+  context 'with in_delta assertions' do
+    it 'registers an offense when using `assert_in_delta`' do
+      expect_offense(<<~RUBY)
+        assert_in_delta(a, b)
+        ^^^^^^^^^^^^^^^^^^^^^ Use `expect(b).to be_within(0.001).of(a)`.
+      RUBY
+
+      expect_correction(<<~RUBY)
+        expect(b).to be_within(0.001).of(a)
+      RUBY
+    end
+
+    it 'registers an offense when using `assert_in_delta` with ' \
+       'no parentheses' do
+      expect_offense(<<~RUBY)
+        assert_in_delta a, b
+        ^^^^^^^^^^^^^^^^^^^^ Use `expect(b).to be_within(0.001).of(a)`.
+      RUBY
+
+      expect_correction(<<~RUBY)
+        expect(b).to be_within(0.001).of(a)
+      RUBY
+    end
+
+    it 'registers an offense when using `assert_in_delta` with ' \
+       'a custom delta' do
+      expect_offense(<<~RUBY)
+        assert_in_delta a, b, 1
+        ^^^^^^^^^^^^^^^^^^^^^^^ Use `expect(b).to be_within(1).of(a)`.
+      RUBY
+
+      expect_correction(<<~RUBY)
+        expect(b).to be_within(1).of(a)
+      RUBY
+    end
+
+    it 'registers an offense when using `assert_in_delta` with ' \
+       'a custom delta from a variable' do
+      expect_offense(<<~RUBY)
+        delta = 1
+
+        assert_in_delta a, b, delta
+        ^^^^^^^^^^^^^^^^^^^^^^^^^^^ Use `expect(b).to be_within(delta).of(a)`.
+      RUBY
+
+      expect_correction(<<~RUBY)
+        delta = 1
+
+        expect(b).to be_within(delta).of(a)
+      RUBY
+    end
+
+    it 'registers an offense when using `assert_in_delta` with ' \
+       'a custom delta and a failure message' do
+      expect_offense(<<~RUBY)
+        assert_in_delta a, b, 1, "must be within delta"
+        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ Use `expect(b).to(be_within(1).of(a), "must be within delta")`.
+      RUBY
+
+      expect_correction(<<~RUBY)
+        expect(b).to(be_within(1).of(a), "must be within delta")
+      RUBY
+    end
+
+    it 'registers an offense when using `assert_in_delta` with ' \
+       'multi-line arguments' do
+      expect_offense(<<~RUBY)
+        assert_in_delta(a,
+        ^^^^^^^^^^^^^^^^^^ Use `expect(b).to be_within(1).of(a)`.
+                      b,
+                      1)
+      RUBY
+
+      expect_correction(<<~RUBY)
+        expect(b).to be_within(1).of(a)
+      RUBY
+    end
+
+    it 'registers an offense when using `assert_not_in_delta`' do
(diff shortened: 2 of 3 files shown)


lib/rubocop/cop/rspec/rails/minitest_assertions.rb
```diff
@@ -229,6 +252,45 @@ def assertion
               'be_empty'
             end
           end
+
+          MSG = 'Use `%<prefer>s`.'
+
+          ASSERTION_MATCHERS = constants(false).filter_map do |c|
```
pirj, line 258: BasicAssertion.subclasses ? Or is it a railsism?

lib/rubocop/cop/rspec/rails/minitest_assertions.rb
```diff
@@ -229,6 +252,45 @@ def assertion
...
+
+            const if const.is_a?(Class) && const.superclass == BasicAssertion
+          end
+
+          RESTRICT_ON_SEND = ASSERTION_MATCHERS.flat_map { |m| m::MATCHERS }
+
+          ASSERTION_MATCHERS.each do |m|
+            name = m.name.split('::').last
+
+            def_node_matcher "minitest_#{name}".to_sym, m::NODE_MATCHER_PATTERN
```
pirj, line 269: Wondering if it’s possible to define node matchers in assertion classes

lib/rubocop/cop/rspec/rails/minitest_assertions.rb
```diff
@@ -229,6 +252,45 @@ def assertion
...
+              end
+            end
+          end
+
+          def on_assertion(node, assertion)
+            return if assertion.nil?
+
+            preferred = assertion.replaced(node)
+            add_offense(node, message: message(preferred)) do |corrector|
+              corrector.replace(node, preferred)
```
pirj, line 287: And just pass the corrector there, too

lib/rubocop/cop/rspec/rails/minitest_assertions.rb
```diff
@@ -229,6 +252,45 @@ def assertion
...
+              name = m.name.split('::').last
+
+              public_send("minitest_#{name}".to_sym, node) do |*args|
+                on_assertion(node, m.match(*args))
+              end
+            end
+          end
+
+          def on_assertion(node, assertion)
+            return if assertion.nil?
```
pirj, line 283: Can it be nil? ‘match` delegates to new?

G-Rath: it delegates in _most_ cases, but this check allows more advanced bail outs that cannot be expressed with node patterns alone - i.e. for checking `_predicate` we bail if the method does not end with `?` and for `in_delta` we bail if the `delta` is missing but the `failure_message` is present (which indicates an `assert_in_delta x, y, "message"` type call).

lib/rubocop/cop/rspec/rails/minitest_assertions.rb
G-Rath: It would mean that they'd have to extend from Rubocop itself in some way but that might not be the end of the world - however it would also mean a greater refactor because the node matcher gets defined on the instance rather than as a static method, so the whole constructor would need to change.

If it's ok with you, I'd prefer to stick with the current refactor since that is working and (in my opinion) is pretty optimal - I'm happy to explore a further refactor otherwise though.

lib/rubocop/cop/rspec/rails/minitest_assertions.rb
G-Rath: No that's a ruby 3.1ism - I've added a note over in #1790 about switching when 3.1 becomes the min.

lib/rubocop/cop/rspec/rails/minitest_assertions.rb
pirj: Aha, got it. Still, I feel it’s a bit off.
Would you still feel it’s good if we checked if it’s nil or not before passing it to on_assertion a few rows above?

G-Rath: Yup I agree - the only reason I had it there earlier is I was hoping I might be able to pass in the method as the block, but seems that can't be done with class methods.

Done in https://github.com/rubocop/rubocop-rspec/pull/1790/commits/4fe5c518b3a2039027cddd5923faf600b7c64636

lib/rubocop/cop/rspec/rails/minitest_assertions.rb
```diff
@@ -186,8 +119,57 @@ def assertion
...
+          # :nodoc:
+          class InDeltaAssertion < BasicAssertion
+            MATCHERS = %i[
+              assert_in_delta
+              assert_not_in_delta
+              refute_in_delta
+            ].freeze
+
+            NODE_MATCHER_PATTERN = <<~PATTERN
+              (send nil? {:assert_in_delta :assert_not_in_delta :refute_in_delta} $_ $_ $!{sym str}? $_?)
```
pirj, line 131: Message can be a proc:
```ruby

# File lib/minitest/unit.rb, line 196
    def assert test, msg = nil
      msg ||= "Failed assertion, no message given."
      self._assertions += 1
      unless test then
        msg = msg.call if Proc === msg
        raise MiniTest::Assertion, msg
      end
      true
    end
    ```

File: lib/rubocop/cop/rspec/rails/minitest_assertions.rb [hunk 6]
```diff
@@ -186,8 +119,57 @@ def assertion
...
+            PATTERN
+
+            def self.match(expected, actual, delta, failure_message)
+              return nil if delta.empty? && !failure_message.empty?
```
pirj, line 135: Do you think it is possible to pass a message without a delta to this assertion? I have doubts by just looking at the signature

lib/rubocop/cop/rspec/rails/minitest_assertions.rb
G-Rath: That's not unique to this though, and none of our code depends on that - we're just copying the source code as is (you might be mentioning this because we're doing `!failure_message.empty?` for this assertion, but that is checking against an _array_ which contains the matched nodes).

We should have tests for this but since it applies for every assertion we currently have I'll do it in a dedicated PR

lib/rubocop/cop/rspec/rails/minitest_assertions.rb
G-Rath: No, I wondered that myself but I don't see how it's possible since it doesn't use keywords

pirj: Should we remove this guard statement then?

G-Rath: No this is needed because Rubocop ends up matching even if delta isn't present - I think it's a combo of how optional captures and specific captures work; I think it does something like "this doesn't match `${sym str}?` but it does match `$_` after that".

i.e. this test will break without this guard: https://github.com/rubocop/rubocop-rspec/pull/1791/files#diff-48b7271a8a66b1c5d7538b2fd3dbdb450bf9b78a6852855ddfbcb796260bdc2dR388-R392

spec/rubocop/cop/rspec/rails/minitest_assertions_spec.rb
```diff
@@ -271,6 +271,127 @@
...
+    it 'does not register an offense when ' \
+       'using `expect(b).not_to be_within(1).of(a)`' do
+      expect_no_offenses(<<~RUBY)
+        expect(b).not_to be_within(1).of(a)
+      RUBY
+    end
+
+    it 'does not register an offense when the delta is missing' do
+      expect_no_offenses(<<~RUBY)
+        assert_in_delta a, b, "whoops, we forgot about the actual delta!"
```
pirj, line 390: If this is illegal for Minitest, I suggest removing this example altogether

G-Rath: It's illegal at runtime but that doesn't mean people can't write it and then run rubocop which'll "safely" "fix" this by transforming it into RSpec which might not react the same way.

(I'm also pretty sure the cop itself will error without accounting for this, if memory serves)

lib/rubocop/cop/rspec/rails/minitest_assertions.rb
pirj: I’m most concerned (see my earlier comments below), is that we assume that the message (str/sym) can be passed as the third argument, and in this case, skip matching. But this would be incorrect for Minitest.
My point is to disregard the type, and always consider the third argument to be the delta (optional), and the fourth - the message (optional, no type constraints.

I agree that we’d better investigate proc messages in a separate pr. Wondering if RSpec provides a lazily evaluated failure message and we could autocorrect those.
If it can’t - we need to skip autocorrection.

pirj: Good news that [RSpec does seem to support proc failure messages.](https://github.com/rspec/rspec-expectations/blob/96a3e5ad8a739318378a0d226a5bff307272cbd8/lib/rspec/expectations/handler.rb#L6). I would still double check this.

spec/rubocop/cop/rspec/rails/minitest_assertions_spec.rb
pirj: How far should we go into supporting invalid code? I’m more inclined to only support what would run in RSpec without errors.
We have a [weird test smoke test](https://github.com/rubocop/rubocop-rspec/blob/master/spec/smoke_tests/weird_rspec_spec.rb), but it would also run with no errors to my best memory.

G-Rath: Generally my rule of thumb has been to account for "reasonable valid syntax" where it's not that much effort to do so, but I'm not fussed - I've pushed up a commit to invert this behaviour.

lib/rubocop/cop/rspec/rails/minitest_assertions.rb
```diff
@@ -115,6 +115,33 @@ def assertion
...
 
+          # :nodoc:
+          class InDeltaAssertion < BasicAssertion
+            MATCHERS = %i[
+              assert_in_delta
+              assert_not_in_delta
+              refute_in_delta
+            ].freeze
+
+            NODE_MATCHER_PATTERN = <<~PATTERN
```
pirj, line 126: We [do it differently now](https://github.com/rubocop/rubocop-rspec/blob/f062f923ce0b246e62ee9d61064ee1fedb54a298/lib/rubocop/cop/rspec/rails/minitest_assertions.rb#L72), right? This needs an update?

spec/rubocop/cop/rspec/rails/minitest_assertions_spec.rb
```diff
@@ -271,6 +271,133 @@
...
+        assert_in_delta a, b, 1, "must be within delta"
+        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ Use `expect(b).to(be_within(1).of(a), "must be within delta")`.
+      RUBY
+
+      expect_correction(<<~RUBY)
+        expect(b).to(be_within(1).of(a), "must be within delta")
+      RUBY
+    end
+
+    it 'registers an offense when using `assert_in_delta` with ' \
```
pirj, line 338: I'm still against including this example. That spec would fail the hard way. Once it's fixed - we'll detect it. We don't have a spec that would show that we ignore `assert_in_delta` with more than four arguments, but we do ignore it (hopefully 😄).

G-Rath: Personally I believe that it's good to capture real world situations like these when the cost is very cheap to highlight what happens without requiring a lot of code reading - in contrast, "handling more than four arguments" is something that is very easily understood by just looking at the node pattern.

However this isn't a hill worth fighting over so I've removed the spec 🙂

pirj (approved):
Thank you! What a streak!

bquorning (approved):

PR #1792, opened 2024-01-30
Implement #negated? in superclass

bquorning:
All current matchers follow the pattern of negative assertions matching `assert_not_*` and `refute_*`. If we add a matcher where the negative assertions do not match this pattern, the method can still be overwritten in that subclass.

- [ ] ~Added tests.~


bquorning:
cc @G-Rath

ydah (approved):
Looks good!

G-Rath (approved):
Nice!

PR #1793, opened 2024-01-30
refactor: define node matchers on assertion classes

G-Rath:
I figured this would be enough but didn't want to overload #1790.

It does send the fixer for `InternalAffairs/NodeMatcherDirective` into an inf. loop though - it looks like one part of the cop sees just the "self" whereas the other sees the whole definition, so it reports "use self.match_assertion instead of self" and then the fixer just keeps adding `.match_assertion` to the end of the string until it errors; for now I've just disabled it, but I might try and look into it at some point.

- [x] Added tests.
- [x] ~Updated documentation.~
- [x] ~Added an entry to the `CHANGELOG.md` if the new code introduces user-observable changes.~

### DIFF
--- a/lib/rubocop/cop/rspec/rails/minitest_assertions.rb
+++ b/lib/rubocop/cop/rspec/rails/minitest_assertions.rb
@@ -28,6 +28,12 @@ module RuboCop
 
           # :nodoc:
           class BasicAssertion
+            extend NodePattern::Macros
+
+            def self.minitest_assertion
+              raise NotImplementedError
+            end
+
             def initialize(expected, actual, fail_message)
               @expected = expected&.source
               @actual = actual.source
@@ -60,7 +66,8 @@ module RuboCop
               refute_equal
             ].freeze
 
-            NODE_MATCHER_PATTERN = <<~PATTERN
+            # @!method self.minitest_assertion(node)
+            def_node_matcher 'self.minitest_assertion', <<~PATTERN # rubocop:disable InternalAffairs/NodeMatcherDirective
               (send nil? {:assert_equal :assert_not_equal :refute_equal} $_ $_ $_?)
             PATTERN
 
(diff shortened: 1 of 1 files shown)


lib/rubocop/cop/rspec/rails/minitest_assertions.rb
```diff
@@ -211,17 +224,9 @@ def assertion
...
-
-            def_node_matcher "minitest_#{name}".to_sym, m::NODE_MATCHER_PATTERN
-          end
-
           def on_send(node)
             ASSERTION_MATCHERS.each do |m|
-              name = m.name.split('::').last
-
-              public_send("minitest_#{name}".to_sym, node) do |*args|
+              m.match_assertion(node) do |*args|
```
G-Rath, line 229: at this point the names are getting a little silly - I'll next explore if we can nicely have the classes handle calling their `match` logic and just return a result

bquorning:
@G-Rath I pushed an alternative implementation at the same time as you opened your PR: #1794. I hadn’t considered using class methods for the node matchers, that’s a neat idea.

lib/rubocop/cop/rspec/rails/minitest_assertions.rb
```diff
@@ -60,7 +66,8 @@ class EqualAssertion < BasicAssertion
               refute_equal
             ].freeze
 
-            NODE_MATCHER_PATTERN = <<~PATTERN
+            # @!method self.minitest_assertion(node)
```
bquorning, line 69: I’m actually not sure if the correct YARD is your suggestion or
```suggestion
            # @!method minitest_assertion(node)
            # @!scope class
```

bquorning (approved):

lib/rubocop/cop/rspec/rails/minitest_assertions.rb
G-Rath: I think you're right, but it still annoys `InternalAffairs/NodeMatcherDirective` 😞

bquorning: I opened an issue on RuboCop: https://github.com/rubocop/rubocop/issues/12665

For now, should we consider your PR ready to merge?

G-Rath: yup works for me

PR #1794, opened 2024-01-30
Refactor assertions: Move node patterns into subclasses

bquorning:
This way we can remove the metaprogramming of `def_node_matcher` and `public_send`, making the cop simpler.

- [ ] Added tests.
- [ ] Updated documentation.


pirj:
Does this still apply?

Issue #1754, opened 2023-12-21
Cop Idea: Only xstr in description


ydah:
If only `xstr` is passed to `describe`, `context` or `it`, the typo of `"` or `'` is suspect in many cases.
Also, if only `xstr` is, it may not be a good description of `describe` or `context` or `it`. What do you think?

```ruby
# bad
describe `time` do
end

# bad
context `time` do
end

# bad
it `time` do
end
```

pirj:
Is this an issue from the real code? Why would anyone want to use that?

ydah:
> Is this an issue from the real code? Why would anyone want to use that?

Yes, it is. Strictly speaking I saw the `it` case I encountered.

pirj:
If those are typos, surely let’s add a cop to fix those

r7kamura:
I guess someone who uses JavaScript template literals a lot wrote it by mistake. I've encountered this misuse too, but since this project was a Rails app, where `` Kernel#` `` is not used very often, I decided to enforce the `%x` literal.

```yaml
Style/CommandLiteral:
  EnforcedStyle: percent_x
```

PR #1796, opened 2024-02-01
Add new `RSpec/UndescriptiveLiteralsDescription` cop

ydah:
Fix: https://github.com/rubocop/rubocop-rspec/issues/1754

- [x] Added tests.
- [x] Updated documentation.

- [x]  The cop is configured as `Enabled: true` in `.rubocop.yml`.


ydah:
> we only allowed str and dstr nodes instead?

For example, description specifies the class to be tested, so I don't think it is just string and dstr. Regex and Integer have added support and renamed Cop.

config/default.yml
```diff
@@ -294,6 +294,12 @@ RSpec/DescribedClassModuleWrapping:
   VersionAdded: '1.37'
   Reference: https://www.rubydoc.info/gems/rubocop-rspec/RuboCop/Cop/RSpec/DescribedClassModuleWrapping
 
+RSpec/DescriptiveDescription:
```
pirj, line 297: I suggest to follow the wip guideline for cop naming and make it reflect what it detects, `UndescriptiveDescription` or ‘UnclearDescription’.

ydah: I updated this PR.

.vscode/settings.json
```diff
@@ -0,0 +1 @@
+{}
```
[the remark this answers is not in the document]
ydah: Oops, I remove it. and create this PR: https://github.com/rubocop/rubocop-rspec/pull/1804

lib/rubocop/cop/rspec/unclear_description.rb
```diff
@@ -0,0 +1,73 @@
...
+      #   context "when #{foo} is bar" do
+      #     # ...
+      #   end
+      #
+      #   # good
+      #   it 'does something' do
+      #     # ...
+      #   end
+      #
+      class UnclearDescription < Base
```
koic, line 47: The name `RSpec/UnclearDescription` might seem not clear enough. How about exploring other options like `RSpec/UndescriptiveLiterals`, `RSpec/UndescriptiveLiteralArguments`, or similar names?

ydah:
I changed this cop name from UnclearDescription to UndescriptiveLiteralsDescription.

spec/rubocop/cop/rspec/undescriptive_literals_description_spec.rb
```diff
@@ -0,0 +1,97 @@
...
+      describe <<~DESC do
+        foo
+      DESC
+      end
+    RUBY
+  end
+
+  it 'does not register an offense when using `describe` with a string and ' \
+     'execute string' do
+    expect_no_offenses(<<~RUBY)
```
pirj, line 92: What if the doc is a method call or a local var?

ydah: Added tests for method call and local var. Perhaps these should not be considered offenses. Since we do not know what is in these values or returns value, there will be false positives if they are considered offense.

lib/rubocop/cop/rspec/undescriptive_literals_description.rb
```diff
@@ -0,0 +1,73 @@
+# frozen_string_literal: true
+
+module RuboCop
+  module Cop
+    module RSpec
+      # Description should be descriptive.
+      #
+      # If `example group` or `example` contains only `execute string`, numbers,
```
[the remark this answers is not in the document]
ydah: That's true! Thank you!

lib/rubocop/cop/rspec/undescriptive_literals_description.rb
```diff
@@ -0,0 +1,73 @@
...
+        def on_block(node) # rubocop:disable InternalAffairs/NumblockHandler
+          example_groups_or_example?(node) do |arg|
+            add_offense(arg) if offense?(arg)
+          end
+        end
+
+        private
+
+        def offense?(node)
+          if node.send_type? && node.method?(:+)
```
pirj, line 64: You’ve added a test for foo + time, but only when all? is false.

If your experience suggests that we need to dive as deep as handling +, I think that if any of parts is an xstr, the whole thing is non-deterministic.
Is sometimh like “‘time’ + ‘whoami’” real? Or just as mistyped quotes like “‘when it is’ + time”?

Again, no objections on merging.

pirj (approved):

lib/rubocop/cop/rspec/undescriptive_literals_description.rb
ydah: Indeed it is! If xstr is included when concatenated with `+`, it's an empty string, so it might as well be offense. In the case of int and regexp, there are cases of false positives, as in the case of the added tests, so I thought it would be a good idea to accept them uniformly. What do you think?

```
❯ irb
irb(main):001> `time` + `time`
=> ""
irb(main):002> `time` + 1
(irb):2:in `+': no implicit conversion of Integer into String (TypeError)
        from (irb):2:in `<main>'
        from <internal:kernel>:187:in `loop'
        from /ydah/.rbenv/versions/3.3.0/lib/ruby/gems/3.3.0/gems/irb-1.11.2/exe/irb:9:in `<top (required)>'
        from /ydah/.rbenv/versions/3.3.0/bin/irb:25:in `load'
        from /ydah/.rbenv/versions/3.3.0/bin/irb:25:in `<main>'
irb(main):003> `time` + 'foo'
=> "foo"
irb(main):004> 'when' + `time`
=> "when"
```

ydah: Or it might be better to always allow for concatenation by `+`. Because it seems difficult to understand the violation in case of violation, and it seems to be outside the scope of this cop.

```ruby
  it 'registers an offense when using `describe` with `time`+"foo"' do
    expect_offense(<<~RUBY)
      describe `time` + 'foo' do
               ^^^^^^^^^^^^^^ Description should be descriptive.
      end
    RUBY
  end

  it 'registers an offense when using `describe` with "foo"+`time`' do
    expect_offense(<<~RUBY)
      describe 'foo' + `time` do
               ^^^^^^^^^^^^^^ Description should be descriptive.
      end
    RUBY
  end

  it 'registers an offense when using `describe` with "foo+(`time`+"bar")' do
    expect_offense(<<~RUBY)
      describe 'foo' + (`time` + 'bar') do
               ^^^^^^^^^^^^^^^^^^^^^^^^ Description should be descriptive.
      end
    RUBY
  end
```

ydah:
Looks fine:
```
Offenses:

/ydah/real-world-rspec/gitlabhq/spec/services/merge_requests/base_service_spec.rb:83:12: C: RSpec/UndescriptiveLiteralsDescription: Description should be descriptive.
  describe `#create_pipeline_for` do
           ^^^^^^^^^^^^^^^^^^^^^^

22772 files inspected, 1 offense detected
```

ydah:
@pirj Do you have any concerns?

bquorning (approved):

pirj:
@ydah The code looks good.
As usual, I’d prefer to run new cops against real-world-rspec, as a preventive measure to user-reported errors. But if you lack capacity to do so, too, I’m fine with releasing this.
My main concern with user-reported errors is that the first thing they suggest is to mark cops as unsafe. And we end up with a bunch of effectively retired cops.

pirj:
@ydah my apologies, this slipped my attention.
Thanks a lot for the cop, it looks great!

PR #1798, opened 2024-02-05
LeakyConstantDeclaration - allow definitions on example group

naveg:
Constant, classes, and modules that are defined on the example group do
not leak into the global namespace. While doing this may not be
a particularly good practice, it should ideally not result in
an offense.

The practice is described here: https://makandracards.com/makandra/47189-rspec-how-to-define-classes-for-specs#section-1-defining-the-constant-on-the-example-class

- [ ] Added tests.
- [ ] Updated documentation.


pirj:
Sorry, I didn’t read the article, but the description of this is misleading.
With self:: (or anything:: really), you explicitly specify the owner of the constant.
The problem that this cop solves is auto-detection of the owner that defaults to the top-level namespace, which is unintuitive and confusing.

I’m fine to accept this if we change the scope to ignore all const definitions with explicitly specified owner.
Or is it problematic for examples?

pirj:
The problem with the approach described in the article is that examples may litter class, and those changes would persist between examples.

The cop was initially introduced to avoid side effects and surprises. I’d love it to keep what it’s doing. https://fili.pp.ru/leaky-constants.html

I’m more inclined to close the PR. WDYT @rubocop/rubocop-rspec?

naveg:
> The problem with the approach described in the article is that examples may litter class, and those changes would persist between examples.

This is true of any class defined in any location. The issue flagged by the cop is that it's quite easy to define classes in the global namespace without realizing that is happening. Using `self::` (or any other namespace, as you point out) resolves that.

pirj:
> This is true of any class defined in any location. 

let(:klass) { Class.new { } } won’t suffer from this.
But in general - yes, it’s possible to use regular variables in example groups like `klass = “hello”`, and it would be possible to alter it, too. This is why we have `let`s.

> Allowing any explicitly specified namespace would make sense to me - even ::. If you do that intentionally, the cop should respect that.

Agree.

naveg:
@pirj I have finally returned to this one and updated it to allow all constants defined with explicit namespaces. Do you think this is a reasonable balance between avoiding the global namespace footgun, and providing an escape hatch for code authors?

I'm not sure how to resolve the code coverage failure - I _think_ that I have tests for everything

spec/rubocop/cop/rspec/leaky_constant_declaration_spec.rb
```diff
@@ -31,6 +31,30 @@
...
+        describe SomeClass do
+          Foo::CONSTANT = "Accessible as Foo::CONSTANT".freeze
+        end
+      RUBY
+    end
+
+    it 'ignores classes defined explicitly in the global namespace' do
+      expect_no_offenses(<<~RUBY)
+        describe SomeClass do
+          ::CONSTANT = "Accessible as ::CONSTANT".freeze
```
pirj, line 53: Also accessible as `CONSTANT` aka “I know what I am doing”.

pirj (approved):
Looks reasonable, thanks for bringing this up.

pirj:
> resolve the code coverage failure 

What’s in coverage/index.html if you run specs locally?

lib/rubocop/cop/rspec/leaky_constant_declaration.rb
```diff
@@ -121,6 +124,16 @@ def on_module(node)
         def inside_describe_block?(node)
           node.each_ancestor(:block).any? { |ancestor| spec_group?(ancestor) }
         end
+
+        def defined_on_explicit_namespace?(node)
+          if node.is_a?(RuboCop::AST::ClassNode) ||
+              node.is_a?(RuboCop::AST::ModuleNode)
+
+            node.loc.name.source.include?('::')
+          elsif node.is_a?(RuboCop::AST::CasgnNode)
```
pirj: If the namespace is a cbase, doesn’t it guarantee that there’s `::` in the const name?

lib/rubocop/cop/rspec/leaky_constant_declaration.rb
```diff
@@ -121,6 +124,16 @@ def on_module(node)
         def inside_describe_block?(node)
           node.each_ancestor(:block).any? { |ancestor| spec_group?(ancestor) }
         end
+
+        def defined_on_explicit_namespace?(node)
+          if node.is_a?(RuboCop::AST::ClassNode) ||
```
pirj: Can this be just `node.class_type?` etc?

lib/rubocop/cop/rspec/leaky_constant_declaration.rb
```diff
@@ -121,6 +124,16 @@ def on_module(node)
...
+              node.is_a?(RuboCop::AST::ModuleNode)
+
+            node.loc.name.source.include?('::')
```
pirj: What would be node.namespace for cases without a namespace, cbase (::Const) and some explicit parent module?
I’d rather prefer  a check against namespace rather than the text source.

bquorning:
> What’s in coverage/index.html if you run specs locally?

It’s the subtle “not covering the implicit `else`” that we’ve seen a few times before. This time in https://github.com/rubocop/rubocop-rspec/blob/cdbfcce97c21edab12f145df8304dba20232ebd3/lib/rubocop/cop/rspec/leaky_constant_declaration.rb#L133

Since the conditional covers the only 3 types of nodes that can reach this method, I think you should add the following lines to the method:
```rb
else
  # :nocov:
  :noop
  # :nocov:
```

lib/rubocop/cop/rspec/leaky_constant_declaration.rb
bquorning: Since relatively recently, it can also read 
```rb
if node.type?(:class, :module)
  # ...
elsif node.type?(:casgn)
  # ...
```

lib/rubocop/cop/rspec/leaky_constant_declaration.rb
naveg: ~const_name won't include the leading `::` in this case, that's why we check for cbase explicitly~

on second thought, if there is any namespace at all, we are good. That's exactly what we're checking for

lib/rubocop/cop/rspec/leaky_constant_declaration.rb
naveg: I think that `!node.identifier.namespace.nil?` does what we want here, and agree it is nicer than looking at the text source

lib/rubocop/cop/rspec/leaky_constant_declaration.rb
```diff
@@ -121,6 +124,19 @@ def on_module(node)
         def inside_describe_block?(node)
           node.each_ancestor(:block).any? { |ancestor| spec_group?(ancestor) }
         end
+
+        def defined_on_explicit_namespace?(node)
```
pirj: Cosmetic: would this become cleaner if we separate this method for class/module and casgn?
Or name it just ‘explicit_namespace?’, made it accept a namespace argument, and only keep the `!namespace.nil?` check inside?

lib/rubocop/cop/rspec/leaky_constant_declaration.rb
```diff
@@ -121,6 +124,19 @@ def on_module(node)
...
+          namespace = if node.type?(:class, :module)
+                        node.identifier.namespace
+                      elsif node.type?(:casgn)
+                        node.namespace
+                      else
+                        # :nocov:
+                        :noop
+                        # :novoc:
```
bquorning:
```suggestion
                        # :nocov:
```

lib/rubocop/cop/rspec/leaky_constant_declaration.rb
bquorning: Perhaps it would be clearer if we invert the method?
```rb
def defined_on_implicit_namespace?(node)
  if node.type?(:class, :module)
    node.identifier.namespace.nil?
  elsif node.type?(:casgn)
    node.namespace.nil?
  else
    :nocov
    true
    :nocov
  end
end
```

pirj: The type check inside the method is effectively the check of who the caller was. This is confusing and redundant.

naveg: Passing the namespace as an argument is much better. Thanks!

bquorning (approved):

spec/rubocop/cop/rspec/leaky_constant_declaration_spec.rb
```diff
@@ -60,15 +84,32 @@ def method
...
+          class Foo::DummyClass
+          end
+        end
+      RUBY
+    end
+
+    it 'ignores classes defined explicitly in the global namespace' do
+      expect_no_offenses(<<~RUBY)
+        describe SomeClass do
+          module ::DummyClass
```
pirj:
```suggestion
          class ::DummyClass
```

bquorning:
Could you please squash the commits into one. Otherwise, I think it’s ready to merge.

naveg:
@bquorning done, thanks!

Issue #1799, opened 2024-02-06
`RSpec/ExampleWording` unintentionally removes escapes


r7kamura:
I noticed that it autocorrects as follows, which causes syntax error:

```ruby
# before
it 'should return foo\'s bar' do
end

# after
it 'returns foo's bar' do
end
```

This can also be confirmed by the following test code.

```diff
diff --git a/spec/rubocop/cop/rspec/example_wording_spec.rb b/spec/rubocop/cop/rspec/example_wording_spec.rb
index 100ae719..ae83c6ac 100644
--- a/spec/rubocop/cop/rspec/example_wording_spec.rb
+++ b/spec/rubocop/cop/rspec/example_wording_spec.rb
@@ -351,5 +351,18 @@ RSpec.describe RuboCop::Cop::RSpec::ExampleWording do
         end
       RUBY
     end
+
+    it 'keeps escaped single-quote after autocorrection' do
+      expect_offense(<<~'RUBY')
+        it 'should return foo\'s bar'
+            ^^^^^^^^^^^^^^^^^^^^^^^^ Do not use should when describing your tests.
+        end
+      RUBY
+
+      expect_correction(<<~'RUBY')
+        it 'returns foo\'s bar' do
+        end
+      RUBY
+    end
   end
 end
```

```
Failures:

  1) RuboCop::Cop::RSpec::ExampleWording when `DisallowedExamples: Workz` keeps escaped single-quote after autocorrection
     Failure/Error:
       expect_correction(<<~'RUBY')
         it 'returns foo\'s bar' do
         end
       RUBY
     
       expected: "it 'returns foo\\'s bar' do\nend\n"
            got: "it 'returns foo's bar' do\nend\n"
     
       (compared using ==)
     
       Diff:
       @@ -1 +1 @@
       -it 'returns foo\'s bar' do
       +it 'returns foo's bar' do
       
     # ./spec/rubocop/cop/rspec/example_wording_spec.rb:362:in `block (3 levels) in <top (required)>'
```

Since this cop supports not only `str` but also `dstr`, it would be difficult to simply fix it with `String#inspect` 🤔

pirj:
We’ve been discussing this in #1189 and closed it as won’t fix.

> this cop supports not only str but also dstr, it would be difficult to simply fix it with String#inspect

Would it be possible to at least fix the str cases? Would you want to give the fix a try, @r7kamura ?

r7kamura:
[...]

PR #1800, opened 2024-02-07
Fix `RSpec/ExampleWording` autocorrection to correctly escape quotes on str node case

r7kamura:
This change resolves some of the following Issue:

- https://github.com/rubocop/rubocop-rspec/issues/1799

It still does not handle quotes well for dstr node, but at least it will handle them well for str node.

- [x] Added tests.
- [ ] Updated documentation.


r7kamura:
Not related to this issue, but if string literals using `%q`, `%`, etc. are used, this cop may not be able to handle that well. It bothered me a little while reading the code. I haven't addressed it directly, but I have taken care to use `node.loc.begin` and `node.loc.end` insetad of `'` or `"` so it will be somewhat better when we try to address it in the future.

ydah (commented):
Looks good! Thank you.

ydah (approved):

lib/rubocop/cop/rspec/example_wording.rb
```diff
@@ -106,6 +110,11 @@ def docstring(node)
           )
         end
 
+        def needs_escape?(node)
+          node.value.include?(node.loc.begin.source) ||
```
pirj, line 114: For quoted strings, both begin and end are the same. Do we need to check twice?

pirj (approved):
Thank you!

I’d love to extensively test this, but only the quoting autocorrection part. Wondering how to do that.

lib/rubocop/cop/rspec/example_wording.rb
r7kamura: This is for `%q(...)`, `%Q(...)`, or `%<...>` etc.

```
irb(#<RuboCop::Cop::RSpec::Exampl...):001> node.loc.begin.source
=> "%q("
irb(#<RuboCop::Cop::RSpec::Exampl...):002> node.loc.end.source
=> ")"
```

But on second thought, we must check against `node.loc.begin.source[-1]`, not `node.loc.begin.source`. I'll try to fix that.

r7kamura: hmm, since escaping should only be necessary if begin and end are different (e.g. `%!...!! ` may require escaping, while `%(...) ` does not), it seems that only one of them needs to be checked after all.

r7kamura:
I've added a few more improvements:

- changed `#docstring` for more accurate `%` string literals handling
- removed unnecessary double-checking from `#needs_escape?`
- added test cases about `%(...)` and `%!...!`

Since the handling of `%` string literals has been improved and the scope of change is no longer only autocorrection, the title and changelog have been slightly changed accordingly.

spec/rubocop/cop/rspec/example_wording_spec.rb
```diff
@@ -351,5 +351,57 @@
...
+
+      expect_correction(<<~'RUBY')
+        it "returns \"foo\"" do
+        end
+      RUBY
+    end
+
+    it 'corrects %(...) quote' do
+      expect_offense(<<~RUBY)
+        it %q(should return foo (bar)) do
```
pirj, line 383: What if there’s a single double quote inside?

r7kamura: In that case, the `should ` part is simply removed. Since the `node.loc.end` (`"` ) is not included, `String#inspect` is not used, and the behavior is the same as before.

I added a test case for that 👌

r7kamura: Good point. Whether or not `"` is included, it is autocorrected with `String#inspect` if `%q(...)` includes `)`.
I added more test cases for that 👌 

```
when message includes `)` in `%q(...)` string literal
  corrects message with `String#inspect`
when message includes `"` in `%q(...)` string literal
  corrects message with direct substring replacement
when message includes `"` and `)` in `%q(...)` string literal
  corrects message with `String#inspect`
```

spec/rubocop/cop/rspec/example_wording_spec.rb
```diff
@@ -351,5 +351,95 @@
         end
       RUBY
     end
+
+    context "when message includes `'` in `'...'`" do
```
bquorning, line 355: I think these examples are in the wrong context (`'when `DisallowedExamples: Workz`'`).

spec/rubocop/cop/rspec/example_wording_spec.rb
```diff
@@ -351,5 +351,95 @@
...
+      it 'corrects message with `String#inspect`' do
+        expect_offense(<<~'RUBY')
+          it 'should return foo\'s bar' do
+              ^^^^^^^^^^^^^^^^^^^^^^^^ Do not use should when describing your tests.
+          end
+        RUBY
+
+        expect_correction(<<~RUBY)
+          it "returns foo's bar" do
```
bquorning, line 364: Maybe it’s a silly question, but why are we changing the type of quotes from `'` to `"` here? If we kept the same quote type, we wouldn’t need to concern ourselves with escaping, would we? (unless _we_ insert an apostrophe/single quote)

r7kamura: This is for implementation convenience. If there is an implementation like `string.inspect(quote_type: 'single_quote')`, I would like to use it. If there is a way to rewrite only substrings while preserving quoting and escaping, that would be great, but do you have any other ideas?

r7kamura: This is a just idea, but, if the offense range had been like this, we might not have to go through all this trouble?

```
it "should return foo's bar"
    ^^^^^^^^^^^^^
```

pirj: Fantastic idea!

PR #1807, opened 2024-02-16
various: fix miscellaneous typos

p-linnane:
This PR aims to fix various minor typos across the codebase. I noticed some of these when doing the same review on our Homebrew/brew repo, since we use rubocop-rspec. I've checked off what I could below given the minor nature of this change.

- [ ] Added tests.
- [ ] Updated documentation.


ydah (approved):
Thank you!

PR #1817, opened 2024-02-26
CI: Fix the RSpec 4 job

bquorning:
I think rspec/rspec was finally removed recently. We should use rspec/rspec-metagem instead.


pirj:
All good, thank you!

Issue #1821, opened 2024-03-01
RSpec/RepeatedSubjectCall flags a false positive


luizkowalski:
Hey 👋🏻 

I have a test like this:

```ruby
subject { described_class.new(url) }
...
context "remaining is exhausted" do
  let(:remaining) { 0 }

  it "fails on the next call" do
    expect { subject.call }.not_to raise_error
    expect { subject.call }.to raise_error(RateLimiter::LimitedError)
  end
end
...
```

from what I understand, `subject` is memorized; in this case, `described_class.new(url)` but the `.call` is not memoized. I know that because this test is currently passing. 

I don't think this cop should flag such cases (when the subject invokes another method).

franzliedke:
Ha, I just came here to report the same problem.

Another example with false positives:

```ruby
expect { subject.method1 }.to raise_error(SomeException)
expect { subject.method2 }.to raise_error(SomeException)
expect { subject.method3 }.to raise_error(SomeException)
```

tobiasamft:
Hello 👋 , I just wanted to add another example which could be worth mentioning:

```ruby
class Something
  def work
    private_method
  end

  private

  def private_method
    # some call that may raise an error
  end
end
```

```ruby
describe Something do
  subject { described_class.new }

  describe '#work' do
    it 'raises' do
      allow(subject).to receive(:private_method).and_raise(
        StandardError,
        'error message',
      )
      expect { subject.work }.to raise_error(StandardError, 'error message')
    end
  end
end
```

sgessa:
Not sure if this case is already covered but seems a false positive to me.
The following example is called from a shared example.

```ruby
it 'raises an error if country GB' do
  if country_id == 'GB'
    expect { action }.to raise_error(ArgumentError)
  else
    expect { action }.not_to raise_error
  end
end
```

pirj:
The expectation inside the "else" clause doesn’t match the docstring. It deserves to be extracted to a separate co text or example.
Thus, no need for an `if` and no false positive.

PR #1822, opened 2024-03-01
Fix a false positive for `RSpec/RepeatedSubjectCall`

ydah:
Fix: #1821

- [x] Added tests.
- [x] Updated documentation.

### DIFF
--- a/lib/rubocop/cop/rspec/repeated_subject_call.rb
+++ b/lib/rubocop/cop/rspec/repeated_subject_call.rb
@@ -23,6 +23,12 @@ module RuboCop
       #     expect { my_method }.to not_change { A.count }
       #   end
       #
+      #   # also good
+      #   it do
+      #     expect { subject.a }.to change { A.count }
+      #     expect { subject.b }.to not_change { A.count }
+      #   end
+      #
       class RepeatedSubjectCall < Base
         include TopLevelGroup
 
@@ -64,17 +70,15 @@ module RuboCop
 
         private
 
-        def detect_offense(example_node, subject_node)
-          walker = subject_node
+        def detect_offense(subject_node)
+          return if subject_node.chained?
+          return unless (block_node = expect_block(subject_node))
 
-          while walker.parent? && walker.parent != example_node.body
-            walker = walker.parent
+          add_offense(block_node)
+        end
 
-            if walker.block_type? && walker.method?(:expect)
-              add_offense(walker)
-              return
-            end
-          end
+        def expect_block(node)
+          node.each_ancestor(:block).find { |block| block.method?(:expect) }
         end
 
         def detect_offenses_in_block(node, subject_names = [])
@@ -96,7 +100,7 @@ module RuboCop
 
           subject_calls(node.body, Set[*subject_names, :subject]).each do |call|
             if subjects_used[call.method_name]
-              detect_offense(node, call)
+              detect_offense(call)
             else
               subjects_used[call.method_name] = true
             end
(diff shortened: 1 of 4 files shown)


bquorning (approved):

lib/rubocop/cop/rspec/repeated_subject_call.rb
```diff
@@ -64,16 +70,11 @@ def on_top_level_group(node)
 
         private
 
-        def detect_offense(example_node, subject_node)
-          walker = subject_node
+        def detect_offense(subject_node)
+          subject_node.each_ancestor(:block) do |block|
+            next if subject_node.chained?
```
pirj, line 75: I have never seen this method. Is it old enough for us to be sure it will be there with a minimum required rubocop and rubocop-ast versions we require?

pirj (approved):
A nice simplification and a quick fix. Thank you!

lib/rubocop/cop/rspec/repeated_subject_call.rb
Darhazer: It was there like forever. I just wonder if we should not just skip the whole each_ancestor in this case, instead of skipping each iteration 😅

ydah: > we should not just skip the whole each_ancestor in this case, instead of skipping each iteration 😅

Indeed! Thank you catching this.

lib/rubocop/cop/rspec/repeated_subject_call.rb
```diff
@@ -64,16 +70,11 @@ def on_top_level_group(node)
 
         private
 
-        def detect_offense(example_node, subject_node)
-          walker = subject_node
+        def detect_offense(subject_node)
+          subject_node.each_ancestor(:block) do |block|
+            break if subject_node.chained?
```
Darhazer, line 75:
```suggestion
          return if subject_node.chained?

          subject_node.each_ancestor(:block) do |block|
```

Darhazer: Even better if the method is not called at all if the node is chained

lib/rubocop/cop/rspec/repeated_subject_call.rb
```diff
@@ -64,16 +70,12 @@ def on_top_level_group(node)
...
 
-          while walker.parent? && walker.parent != example_node.body
-            walker = walker.parent
+          subject_node.each_ancestor(:block) do |block|
 
-            if walker.block_type? && walker.method?(:expect)
-              add_offense(walker)
-              return
-            end
+            add_offense(block) if block.method?(:expect)
```
pirj, line 78: One last optional suggestion for cases with nested ‘except’ - do we need to mark both, or the inner is fine? We could probably ‘each_ancestor.find’?
Completely optional.
All is good, mege at will and thanks once again!

ydah: How about this? Let me know if I'm wrong.

Darhazer (approved):
I have no objections against this change.

The original spec however is light on use cases. What happens for instance if you access subject in a hook, let's say configuring a stub like `allow(Foo).to receive(:bar).and_return(subject)` (which might also use block synax `allow(Foo).to receive(:bar) { subject }`. It's hard to tell from reading the code alone

ydah:
>The original spec however is light on use cases. What happens for instance if you access subject in a hook, let's say configuring a stub like allow(Foo).to receive(:bar).and_return(subject) (which might also use block synax allow(Foo).to receive(:bar) { subject }. It's hard to tell from reading the code alone

Indeed. I'll make a follow-up PR.

Issue #1741, opened 2023-11-16
Improve `RSpec/DescribedClass` cop


ydakuka:
## Describe the solution you'd like

```ruby
RSpec.describe Site do
  subject(:site) { create(:site, value: value) }

  # bad
  let(:value) { Site::MAX_VALUE }
  # good
  let(:value) { described_class::MAX_VALUE }

  specify do
    expect(site).not_to be_nil
  end
end
```

## Rubocop

```console
ydakuka@yauhenid:~/Work/project$ bin/rails_docker rubocop -V
1.57.2 (using Parser 3.2.2.4, rubocop-ast 1.29.0, running on ruby 2.7.8) [x86_64-linux]
  - rubocop-capybara 2.19.0
  - rubocop-factory_bot 2.24.0
  - rubocop-performance 1.19.1
  - rubocop-rails 2.22.0
  - rubocop-rake 0.6.0
  - rubocop-rspec 2.25.0
  - rubocop-thread_safety 0.5.1
```

ydah:
We don't know if it is a class name or a constant name? For example, if we use Foo::A or Foo::HTML, we don't know whether A or HTML is really a class name or a constant name.

pirj:
Do we care if the `A` const is assigned a class or a value? We only care about `Foo` here, and want to detect cases where it is used as a namespace like `Foo::A`, right?

ydah:
Ah, I see. I was mistaken.

timon:
This change makes a lot of our CI pipelines fail, with typical example being a gem autogenerated by `bundle`:

```ruby
 RSpec.describe MyNewGem do 
   it 'has a version number' do 
     expect(MyNewGem::VERSION).not_to be_nil 
   end 
 end 
```

For things like these I would prefer full name for verbosity and explicitness.
Is there a way to configure this cop to still accept old style of referencing namespaced constants, besides having an engineer to go through this and manually putting a disable comment? (Yes, I'm aware of autocorrect — no, I don't want to autocorrect these)

bquorning:
There’s the `EnforcedStyle: explicit` configuration option, as documented in https://docs.rubocop.org/rubocop-rspec/cops_rspec.html#rspecdescribedclass. There is not currently an option to enforce different styles for “stand-alone constants” (e.g. `MyNewGem`) vs namespaces (e.g. `MyNewGem::VERSION`).

timon:
[...]

PR #1825, opened 2024-03-01
Add configuration option `OnlyStaticConstants` to `RSpec/DescribedClass`

ydah:
Fix: https://github.com/rubocop/rubocop-rspec/issues/1741#issuecomment-1973221319
Fix: https://github.com/rubocop/rubocop-rspec/issues/1741#issuecomment-1973604382

- [x] Added tests.
- [x] Updated documentation.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## Master (Unreleased)
 
 - Fix a false positive for `RSpec/RepeatedSubjectCall` when `subject.method_call`. ([@ydah])
+- Add configuration option `OnlyStaticConstants` to `RSpec/DescribedClass`. ([@ydah])
 
 ## 2.27.0 (2024-03-01)
 
--- a/config/default.yml
+++ b/config/default.yml
@@ -283,9 +283,10 @@ RSpec/DescribedClass:
   SupportedStyles:
     - described_class
     - explicit
+  OnlyStaticConstants: true
   SafeAutoCorrect: false
   VersionAdded: '1.0'
-  VersionChanged: '1.11'
+  VersionChanged: "<<next>>"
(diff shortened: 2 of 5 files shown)


bquorning (approved):

bquorning:
@pirj, I would love to hear your opinion on this PR.

pirj:
I see it as a code smell when constants of the class under test are referenced from its spec, at least in a way that #1741 describes.

In general, this option probably makes sense to avoid offences in examples that are less coupled with the class under test.

I don’t have experience turning on Sorbet for specs, so I can’t suggest anything on this front.

I don’t take the “our CI pipelines fail” [argument](https://github.com/rubocop/rubocop-rspec/issues/1741#issuecomment-1972784112) given that auto-correction doesn’t break the spec.

Personally, I don’t see why ‘described_class::MyException' is any different from ‘described_class::VERSION’, but I understand that some people may prefer to use different styles for them.

Again, personally I think ‘described_class’ was a mistake, and we it’s good to quote class names in the top-level example group’s description, and to use the full class name elsewhere in the spec, with no indirection.

bquorning:
> Again, personally I think ‘described_class’ was a mistake, and we it’s good to quote class names in the top-level example group’s description, and to use the full class name elsewhere in the spec, with no indirection.

I don’t remember us having this discussion before. I’m not sure I find `described_class` to be _that_ useful either. Do you know why it’s even a feature in RSpec?

pirj:
> why it’s even a feature in RSpec?

This I can’t remember, but there’s a funny story about the RSpec 3 release and `described_class`, also a nice compilation of quotes and opinions in a discussion below the description https://github.com/rubocop/rspec-style-guide/issues/104. Since most likely the `described_class` will stay in RSpec 4, the cop should remain in some form, too.

If it’s not a big maintenance overhead for us to provide our users with an optional … option to style their code the way they like, why not?

bquorning (approved):

Issue #1827, opened 2024-03-01
Trusted publishing


bquorning:
I’m looking into using [RubyGems’ trusted publishing](https://guides.rubygems.org/trusted-publishing/adding-a-publisher/) to push a gem from a GitHub action ([example](https://github.com/segiddins/rubygems-await/blob/main/.github/workflows/push_gem.yml)). @rubocop/rubocop-rspec Do you see any downsides to this approach?

ydah:
I think there doesn't seem to be any particular downsides.

PR #1829, opened 2024-03-04
Use Rubygems Trusted Publishers to publish

ydah:
Fix: https://github.com/rubocop/rubocop-rspec/issues/1827

- [-] Added tests.
- [-] Updated documentation.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -1,0 +1,28 @@
+name: Publish
+on:
+  push:
+    branches: master
+    paths: lib/rubocop/rspec/version.rb
+jobs:
+  publish:
+    name: Publish to RubyGems
+    runs-on: ubuntu-latest
+    permissions:
+      contents: write
+      id-token: write
+    steps:
+      - uses: actions/checkout@v4
+      - name: Set up Ruby
+        uses: ruby/setup-ruby@v1
+        with:
+          bundler-cache: true
+          ruby-version: ruby
+      - uses: rubygems/release-gem@v1
+      - name: Replace version in Antora config
+        run: |
+          sed -i 's/version:.*$/version: ~/' docs/antora.yml
+          git config user.name "${GITHUB_ACTOR}"
+          git config user.email "${GITHUB_ACTOR}@users.noreply.github.com"
+          git add docs/antora.yml
+          git commit -m "Switch docs version back"
+          git push


.github/workflows/publish.yml
```diff
@@ -0,0 +1,20 @@
+name: Publish
+on:
+  release:
+    types:
+      - published
```
ydah, line 5: memo) https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#release

bquorning: If we trigger on `release`, we’d have to manually create a release in the GitHub UI, right?

Another option is to run the workflow when a change to version.rb has been merged into `master` branch, like this:
```yml
on:
  push:
    branches: master
    paths: lib/rubocop/rspec/version.rb
```

.github/workflows/publish.yml
```diff
@@ -0,0 +1,20 @@
...
+    permissions:
+      contents: write
+      id-token: write
+    steps:
+      - uses: actions/checkout@v4
+      - name: Set up Ruby
+        uses: ruby/setup-ruby@v1
+        with:
+          bundler-cache: true
+          ruby-version: '2.7'
```
bquorning, line 19: The example workflow at https://github.com/segiddins/rubygems-await/blob/main/.github/workflows/push_gem.yml has `ruby-version: ruby`, which is a value I cannot find documentation for.

bquorning (commented):
I was hoping we could have this workflow reset docs/antora.yml (like e.g. 447fa4e3d63e8998c8039ff7b9a2b216f61aa35c) after each release.

bquorning:
Ooh, maybe this workflow could also create a new GH Release after publishing a new version.

.github/workflows/publish.yml
ydah: Indeed I updated this PR. Thank you .

ydah:
I updated this PR. How about this?

.github/workflows/publish.yml
```diff
@@ -0,0 +1,30 @@
...
+      - uses: actions/checkout@v4
+      - name: Set up Ruby
+        uses: ruby/setup-ruby@v1
+        with:
+          bundler-cache: true
+          ruby-version: ruby
+      - uses: rubygems/release-gem@v1
+      - name: Replace version in README
+        run: |
+          sed -i "s/version:.*$/version: ~/g" docs/antora.yml
```
bquorning, line 23: Let’s remove the `g` option and use single quotes:
```suggestion
          sed -i 's/version:.*$/version: ~/' docs/antora.yml
```

.github/workflows/publish.yml
```diff
@@ -0,0 +1,30 @@
...
+      id-token: write
+    steps:
+      - uses: actions/checkout@v4
+      - name: Set up Ruby
+        uses: ruby/setup-ruby@v1
+        with:
+          bundler-cache: true
+          ruby-version: ruby
+      - uses: rubygems/release-gem@v1
+      - name: Replace version in README
```
bquorning, line 21:
```suggestion
      - name: Replace version in Antora config
```

.github/workflows/publish.yml
```diff
@@ -0,0 +1,30 @@
...
+  push:
+    branches: master
+    paths: lib/rubocop/rspec/version.rb
+jobs:
+  publish:
+    name: Publish to RubyGems
+    runs-on: ubuntu-latest
+    permissions:
+      contents: write
+      id-token: write
```
bquorning, line 12: Are these permissions actually necessary?

ydah: It is added because it was used in `example`. Is it unnecessary?
https://github.com/rubocop/rubocop-rspec/issues/1827#issue-2164094559
https://github.com/segiddins/rubygems-await/blob/136ee45fac859c913d7f4b745c63c9ecdc920e14/.github/workflows/push_gem.yml#L13-L15

bquorning: I ran a few tests and you are absolutely correct, they are both needed. I believe `id-token` is necessary in order to fetch the temporary auth token, and `contents` is necessary for pushing the git tag.

pirj:
This is incredible!
I’ve just been reading a story about xz-utils today, what they pushed as a release was different from the source code. I hope we reduce the possibility of this happening with such trusted publishing?

.github/workflows/publish.yml
```diff
@@ -0,0 +1,30 @@
...
+        uses: ruby/setup-ruby@v1
+        with:
+          bundler-cache: true
+          ruby-version: ruby
+      - uses: rubygems/release-gem@v1
+      - name: Replace version in Antora config
+        run: |
+          sed -i 's/version:.*$/version: ~/' docs/antora.yml
+      - name: Commit version change
+        run: |
```
bquorning, line 25: In my opinion, these two lines could be removed.
```suggestion
```

ydah:
@bquorning Thank you 😄! Let's merge #1859. Then we will close this PR👍

bquorning (approved):

bquorning:
@ydah If you squash the two commits, we can merge this PR and I’ll rebase #1859.

bquorning:
The final `git push` failed because of our branch protection rules: https://github.com/rubocop/rubocop-rspec/actions/runs/8561970997/job/23464311900

I don’t know if we can somehow override that.

ydah:
Maybe, I think the solution is as follows.
* creating a service account user that has read-write access to repo
* create a pat for that service account
* add the service account to the bypass protection rules in the branch
* create a repo secret that contains the PAT token
* push using the created service account

Issue #1734, opened 2023-10-24
Extract rspec-rails cops


bquorning:
Discussed in https://github.com/rubocop/rubocop-rspec/discussions/1440.

bquorning:
@ydah In #1819 you hinted as wanting to do this extraction before the next minor release. Do you have some work in progress? 😄

ydah:
@bquorning It's still WIP, but I'm creating a PR below.
https://github.com/rubocop/rubocop-rspec/pull/1830

PR #1830, opened 2024-03-04
Extract RSpec Rails cops

ydah:
Fix: https://github.com/rubocop/rubocop-rspec/issues/1734

### TODO
- [x] once the gem is published, adds rubocop-rspec_rails to gemspec and remove from Gemfile
- [x] set the version constraint in gemspec for `rspec-rspec_rails`
- [x] Creating a Repository for https://github.com/rubocop/rubocop-rspec_rails
- [x] Extracted to https://github.com/rubocop/rubocop-rspec_rails

Here's where to temporary extract it:.
- https://github.com/ydah/rubocop-rspec_rails/pull/1

- [ ] Added tests.
- [ ] Updated documentation.


pirj:
Should we do rubocop-rspec_rails with underscore lije we do for rubocop-factory_bot?
As a repo name and in the code especially: we’ll use RuboCop::RSpecRails as the namespace, right?

bquorning:
Based on https://guides.rubygems.org/name-your-gem/ and e.g. https://github.com/ydah/rubocop-rspec-rails/blob/207fc3483e75605b12276b90e5bb3a7ebcdc223f/lib/rubocop/cop/rspec_rails/avoid_setup_hook.rb defining `RuboCop::Cop::RSpecRails` and not `RuboCop::Cop::RSpec::Rails`, I think the gem should be named rubocop-rspec_rails.

ydah:
Thank you. I changed rubocop-rspec_rails instead of rubocop-rspec-rails.

bquorning:
I created https://github.com/rubocop/rubocop-rspec_rails with @rubocop/rubocop-core given write-access. Ping me if anyone needs admin access.

ydah:
If the modifications on the rubocop-rspec_rais side are correct, we will apply them to the repository you created. Please review https://github.com/ydah/rubocop-rspec_rails/pull/1 .

config/obsoletion.yml
```diff
@@ -27,3 +27,10 @@ renamed:
   RSpec/FactoryBot/FactoryClassName: FactoryBot/FactoryClassName
   RSpec/FactoryBot/FactoryNameStyle: FactoryBot/FactoryNameStyle
   RSpec/FactoryBot/SyntaxMethods: FactoryBot/SyntaxMethods
+  RSpec/Rails/AvoidSetupHook: RSpecRails/AvoidSetupHook
```
pirj, line 30: The rest should be this way, too

docs/modules/ROOT/pages/departments.adoc
```diff
@@ -35,7 +35,7 @@ There are plans for extracting the following three departments before RuboCop RS
 |Extracted in https://github.com/rubocop/rubocop-rspec/releases/tag/v2.22.0[v2.22.0]
 
 |RSpec/Rails
-|rubocop-rspec-rails
+|rubocop-rspec_rails
 |Not yet extracted
```
pirj, line 39: Extracted! 🥳
```suggestion
|Extracted in https://github.com/rubocop/rubocop-rspec/releases/tag/v2.28.0[v2.28.0]
```

lib/rubocop/cop/rspec/rails/http_status.rb
```diff
@@ -1,213 +1,60 @@
 # frozen_string_literal: true
 
-require 'rack/utils'
```
pirj, line 3: We can now remove rack-utils from the gemspec?

pirj (approved):
Looks great, thank you!
Were there any modules that are needed to to run the newly-extracted extension without rubocop-rspec? Can we remove them from rubocop-rspec? I recall we had that for capybara/factory_bot.

pirj:
I can’t spot any rails-specific modules in our code at a glance, so all good.

lib/rubocop/cop/rspec/rails/http_status.rb
ydah: Yes! Thank you. I remove it.

config/obsoletion.yml
ydah: Ah, sorry. I fixed it.

pirj (approved):
Lgtm! 😍

ydah:
I overlooked something serious. Gems cannot depend on each other.

```
❯ bundle install
Your bundle requires gems that depend on each other, creating an infinite loop. Please remove either gem 'rubocop-rspec' or gem 'rubocop-rspec_rails' and try again.
```

pirj:
I guess we’ll have to remove the dependency from rubocop-rspec_rails and make a note there that it implicitly depends on rubocop-rspec.
In 3.0, we’ll remove the dependency on rubocop-rspec_rails from rubocop-rspec, and will restore the dependency declaration in rubocop-rspec_rails.
This might be a mild surprise to those who would want just rubocop-rspec_rails, but I suppose it’s a very rare case.

ydah:
@pirj Should rubocop-rspec_rails be a patch upgrade or a minor upgrade?

pirj:
I’d yank the 2.28.0 and released a bugfix patch in 2.28.1.
@bquorning what do you think?

bquorning:
>I’d yank the 2.28.0 and released a bugfix patch in 2.28.1.
Yes that sounds like a good idea. And I assume nobody is using the gem yet, so the yank won’t affect users.

ydah:
I update this PR. And All Green ✅

PR #1837, opened 2024-03-18
Fix AssignmentOnly saying false instead of true in documentation

Skipants:
This option on the documentation was accidentally flipped from what it should have been. The example following is what happens when the "AssignmentOnly" configuration is set to "true", not "false"

- [x] Updated documentation.

(Rest of prerequisites not applicable to this change)


ydah (approved):
Thank you! Looks good.

PR #1854, opened 2024-03-30
Add new RSpec/EmptyOutput cop

bquorning:
Instead of calling `output` with an empty string, you should use the inverse runner and call `output` without an argument. E.g. instead of

    expect { foo }.to output('').to_stdout

you should call

    expect { foo }.not_to output.to_stdout

- [x] Added tests.
- [x] Updated documentation.


CHANGELOG.md
```diff
@@ -6,6 +6,8 @@
 
 - Extract RSpec Rails cops to a separate repository, [`rubocop-rspec_rails`](https://github.com/rubocop/rubocop-rspec_rails). The `rubocop-rspec_rails` repository is a dependency of `rubocop-rspec` and the cops related to rspec-rails are aliased (`RSpec/Rails/Foo` == `RSpecRails/Foo`) until v3.0 is released, so the change will be invisible to users until then. ([@ydah])
 
+- Add new `RSpec/EmptyOutput` cop. ([@bquorning])
```
ydah, line 9: Could you move to `Master (Unreleased) `?

bquorning:
cc @pirj

pirj:
One exception I can think of is a compund expectation like `output(‘’).to_stdout.and output(‘’).to_stderr`.
We had a cop that suggested using a custom negation matcher for compound expectations. Was that `not_change`?

Otherwise - looks good, thanks!

bquorning:
I actually did look into the compound marchers, and I think`to_stderr` and `to_stdout` both return `true` so they cannot be chained. I may be wrong of course.

pirj:
At a glance, it returns [self](https://github.com/rspec/rspec-expectations/blob/e859d358fff504f9132cb9eaa9b33765e73fc187/lib/rspec/matchers/built_in/output.rb#L33), and the vase class does include Composable that defines and/or.

Not a big deal, just an edge case. But since `to_std…` is [mandatory](https://github.com/rspec/rspec-expectations/blob/e859d358fff504f9132cb9eaa9b33765e73fc187/lib/rspec/matchers/built_in/output.rb#L134), it’s not unrealistic.

bquorning:
I’ll have a look.

Actually what should we recommend if someone matches e.g.

```rb
expect {}.to output("foo").to_stdout.and output("").to_stderr
```
I don’t think there’s a way to do e.g.
```rb
expect {}.to output("foo").to_stdout.and not output.to_stderr
# or
expect {}.to_not output.to_stderr.and output("foo").to_stdout
```

The last code tells me that
> `expect(...).not_to matcher.and matcher` is not supported, since it creates a bit of an ambiguity. Instead, define negated versions of whatever matchers you wish to negate with `RSpec::Matchers.define_negated_matcher` and use `expect(...).to matcher.and matcher`.

pirj (approved):
Looks good, thank you!

pirj:
For the other cop, we [register an offence](https://github.com/rubocop/rubocop-rspec/blob/ef302ee7802cf5d0becd33a7be5946433393fb25/spec/rubocop/cop/rspec/change_by_zero_spec.rb#L61), but don’t autocorrect, since it’s not guaranteed that the negated matcher (not_change) is defined in the user’s codebase.

```ruby
expect { :nop }
  .to not_output.to_stdout
  .and not_output.to_stderr
```
should work, just like a combination of output.to_stderr with eg not_output.to_stdout

bquorning:
Yeah, I think it’s too much if we ask the user to implement a `#not_output` method when `output("")` works fine.

bquorning:
Just to be sure, the `all` matcher cannot be used together with `output`, right? E.g. `expect([proc_1, proc_2]).to all(output("").to_stderr)`

ydah:
I checked with real-world-rspec. There seems to be no errors or false positives ✅

```
Offenses:

/Users/yudai.takada/ydah/real-world-rspec/rubocop/spec/rubocop/config_loader_spec.rb:1016:74: C: [Correctable] RSpec/EmptyOutput: Use not_to instead of matching on an empty output.
        expect { expect(configuration_from_file.to_h).to eq(config) }.to output('').to_stderr
                                                                         ^^^^^^^^^^
/Users/yudai.takada/ydah/real-world-rspec/rubocop/spec/rubocop/config_loader_spec.rb:1213:18: C: [Correctable] RSpec/EmptyOutput: Use not_to instead of matching on an empty output.
          end.to output('').to_stderr
                 ^^^^^^^^^^
/Users/yudai.takada/ydah/real-world-rspec/rubocop/spec/rubocop/config_loader_spec.rb:1249:18: C: [Correctable] RSpec/EmptyOutput: Use not_to instead of matching on an empty output.
          end.to output('').to_stderr
                 ^^^^^^^^^^
/Users/yudai.takada/ydah/real-world-rspec/rubocop/spec/rubocop/config_loader_spec.rb:1317:47: C: [Correctable] RSpec/EmptyOutput: Use not_to instead of matching on an empty output.
        expect { configuration_from_file }.to output('').to_stderr
                                              ^^^^^^^^^^

22407 files inspected, 4 offenses detected, 4 offenses autocorrectable
```

ydah (approved):
Looks good!

PR #1856, opened 2024-03-31
Fix autocorrect error for RSpec/ExpectActual

### LINKED ISSUE #1834: `RSpec/ExpectActual` autocorrects unsafely given a hash and a matcher without brackets
Given this input, for example:

```rb
expect({ a: 1 }).to eq something_else
```

`RSpec/ExpectActual` autocorrects to this:

```rb
expect(something_else).to eq { a: 1 }
```

This isn't interpreted correctly by Ruby, which returns:

```
syntax error, unexpected ':', expecting '}' (SyntaxError)
```

Safe output for this cop would be:

```rb
expect(something_else).to eq({ a: 1 })
```

bquorning:
When the expect value is not wrapped in parentheses, the result of autocorrecting would sometimes be invalid Ruby code.

Fixes #1834

- [x] Added tests.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Master (Unreleased)
 
+- Fix an autocorrect error for `RSpec/ExpectActual`. ([@bquorning])
+
 ## 2.28.0 (2024-03-30)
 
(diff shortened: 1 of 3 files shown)


ydah (approved):
Looks good!

PR #1859, opened 2024-04-02
Automatically create release notes

bquorning:
Adds a new rake task, `rake create_release_notes` which generates a relnotes.md file.

When CI creates and publishes a new gem version, it will now also create a release on GitHub, with the content of relnotes.md being used as release notes.

- [ ] Added tests.
- [ ] Updated documentation.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -18,6 +18,12 @@ jobs:
           bundler-cache: true
           ruby-version: ruby
       - uses: rubygems/release-gem@v1
+      - name: Create a GitHub release
+        env:
+          GH_TOKEN: ${{ github.token }}
+        run: |
+          bundle exec rake create_release_notes
+          gh release create $(git tag --points-at @) --notes-file relnotes.md
       - name: Replace version in Antora config
         run: |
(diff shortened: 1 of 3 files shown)


.github/workflows/publish.yml
```diff
@@ -18,6 +18,12 @@ jobs:
           bundler-cache: true
           ruby-version: ruby
       - uses: rubygems/release-gem@v1
+      - name: Create a GitHub release
+        env:
+          GH_TOKEN: ${{ github.token }}
+        run: |
+          bundle exec rake create_release_notes
+          gh release create $(git tag --points-at @) --notes-file relnotes.md
```
ydah, line 26: ask) Should we need to create a release version tag before this?

bquorning: Yes. It’s created and pushed by line 20. And also, it should run _before_ the `git push` below which would cause `git tag --points-at @` to no longer return the current tag 😅

ydah: Thank ypu for the explanation! I understand😀

ydah (approved):
Looks good!

PR #1869, opened 2024-04-09
Improve release workflow

ydah:
This PR improves the release workflow.
It turns out that being able to push directly to the default branch is quite difficult. That's because if GitHub actions can push to a protected branch, all you have to do is code a workflow that creates a branch and pushes it to other branches, and any collaborator in the repository can push whatever code they want to any branch.
https://github.com/orgs/community/discussions/25305#discussioncomment-3247401

That last one step might be a pain, but I think it's better to automate it to the point where you create a PR that switches back versions.
After the release workflow, a PR is created that switches back, so we should just merge it.

Note that this requires the following settings to be checked on:

- [-] Added tests.
- [-] Updated documentation.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -10,6 +10,7 @@ jobs:
     permissions:
       contents: write
       id-token: write
+      pull-requests: write
     steps:
       - uses: actions/checkout@v4
       - name: Set up Ruby
@@ -32,7 +33,9 @@ jobs:
           if ! git diff --exit-code docs/antora.yml; then
             git config user.name "${GITHUB_ACTOR}"
             git config user.email "${GITHUB_ACTOR}@users.noreply.github.com"
+            git checkout -b switch-docs-version
             git add docs/antora.yml
             git commit -m "Switch docs version back"
-            git push
+            git push -u origin switch-docs-version
+            gh pr create --fill --head switch-docs-version
           fi


.github/workflows/publish.yml
```diff
@@ -32,7 +33,10 @@ jobs:
           if ! git diff --exit-code docs/antora.yml; then
             git config user.name "${GITHUB_ACTOR}"
             git config user.email "${GITHUB_ACTOR}@users.noreply.github.com"
+            git checkout -b switch-docs-version
             git add docs/antora.yml
             git commit -m "Switch docs version back"
-            git push
+            git push -u origin switch-docs-version
+            gh pr create --title "Switch docs version back" --body "" \
```
bquorning, line 40: There’s a shortcut to fill the PR details from the commit message:

```suggestion
            gh pr create --fill \
```

.github/workflows/publish.yml
```diff
@@ -32,7 +33,10 @@ jobs:
...
+              --head master -H switch-docs-version
```
bquorning, line 41: `-H` and `--head` are aliases for the same option. I think you meant
```suggestion
              --head switch-docs-version
```

.github/workflows/publish.yml
bquorning: Should we add assignee, so it’s less likely we’ll overlook the PR?
```sh
              --assignee @rubocop/rubocop-rspec \
```

ydah: I think CODEOWNERS will automatically add it to the assigny. Am I missing something?

bquorning: No, I think you are right. I sometimes overlook _issues_, but I think get notifications for all PRs due to CODEOWNERS. Thanks for pointing it out.

ydah:
@bquorning I don't have permission, so please check if the following settings are checked ON.

bquorning:
Yep, the setting is ON.

Issue #1852, opened 2024-03-30
Remove RSpec/Capybara/FeatureMethods


bquorning:
See if it can’t be replaced with `RSpec/Dialect` configuration. ([source of this idea](https://github.com/rubocop/rubocop-rspec/pull/1848#discussion_r1545140340))

PR #1876, opened 2024-05-03
Remove RSpec/Capybara/FeatureMethods

ydah:
Fix: https://github.com/rubocop/rubocop-rspec/issues/1852

- [x] Added tests.
- [x] Updated documentation.


config/obsoletion.yml
```diff
@@ -34,3 +34,8 @@ renamed:
   RSpec/Rails/MinitestAssertions: RSpecRails/MinitestAssertions
   RSpec/Rails/NegationBeValid: RSpecRails/NegationBeValid
   RSpec/Rails/TravelAround: RSpecRails/TravelAround
+
+removed:
+  RSpec/Capybara/FeatureMethods:
+    reason: >
+      this cop has migrated to `RSpec/Dialect`. Please use `RSpec/Dialect` instead.
```
bquorning, line 41: Should we explain a bit more _how_ to use RSpec/Dialect instead?

bquorning (approved):
What does @pirj think?

config/default.yml
```diff
@@ -298,8 +298,15 @@ RSpec/DescribedClassModuleWrapping:
 RSpec/Dialect:
   Description: Enforces custom RSpec dialects.
   Enabled: false
-  PreferredMethods: {}
+  PreferredMethods:
```
pirj, line 301: It doesn’t feel right to change the config by default.
We can’t imply that those methods are always for capybara specs.
I can’t quickly find it, but I recall someone mentioned that ‘describe’ isn’t a safe correction for ‘feature’ that has a certain metadata attached to it by default.

I suggest instead adding what needs to be done to this message below:
```
in .rubocop.yml file, enable `RSpec/Dialect`.
```

pirj:
Nice! A small note on config, otherwise looks good. Thank you!

I’m sorry if I’ve missed that in some other ticket/pr/discussion, is there anything that’s stopping us from releasing 3.0 with those breaking changes as our next release? I’ve seen that we have two branches now.

bquorning:
> I’m sorry if I’ve missed that in some other ticket/pr/discussion, is there anything that’s stopping us from releasing 3.0 with those breaking changes as our next release? I’ve seen that we have two branches now.

No, I don’t think there’s a whole lot more to be done before releasing a v3.0 ✨ I started a discussion at #1880.

config/default.yml
bquorning: I agree. One thing we could do it to link from the obsoletion.yml message to https://docs.rubocop.org/rubocop-rspec/cops_rspec.html#rspecdialect, and add to the documentation in https://github.com/rubocop/rubocop-rspec/blob/6f340256dd1502cf23f20f6c6b7410e05de0c923/lib/rubocop/cop/rspec/dialect.rb#L42-L47


```diff
diff --git a/lib/rubocop/cop/rspec/dialect.rb b/lib/rubocop/cop/rspec/dialect.rb
index 7140541c..95d75161 100644
--- a/lib/rubocop/cop/rspec/dialect.rb
+++ b/lib/rubocop/cop/rspec/dialect.rb
@@ -42,6 +42,19 @@ module RuboCop
       #     # ...
       #   end
       #
+      # If you were previously using the `RSpec/Capybara/FeatureMethods` cop and
+      # want to keep disabling all Capybara-specific methods that have the same
+      # native RSpec method (e.g. are just aliases), use the following config:
+      #
+      #   RSpec/Dialect:
+      #     PreferredMethods:
+      #       background: :before
+      #       scenario:   :it
+      #       xscenario:  :xit
+      #       given:      :let
+      #       given!:     :let!
+      #       feature:    :describe
+      #
       class Dialect < Base
         extend AutoCorrector
         include MethodPreference
```

bquorning:
@ydah Could you please rebase once again?

Related: Should I enable the “Always suggest updating pull request branches” feature for the repo? 

> Whenever there are new changes available in the base branch, present an “update branch” option in the pull request.

That would mean other maintainers could easily rebase a PR (or merge origin/master in), but perhaps at the cost of (many) more commits, and a more confusing workflow for the contributor whose local branch no longer matches the remote.

ydah:
@bquorning I did it!

> That would mean other maintainers could easily rebase a PR (or merge origin/master in), but perhaps at the cost of (many) more commits, and a more confusing workflow for the contributor whose local branch no longer matches the remote.

I think it's a very good point of view. IMO, but we don't want to rush the merge. And confuse PR authors. So I think it's fine the way it is.

config/obsoletion.yml
```diff
@@ -40,3 +40,9 @@ split:
     alternatives:
       - RSpec/SpecFilePathFormat
       - RSpec/SpecFilePathSuffix
+
+removed:
+  RSpec/Capybara/FeatureMethods:
+    reason: >
+      this cop has migrated to `RSpec/Dialect`. Please use `RSpec/Dialect` instead.
+      in .rubocop.yml file, enable and `RSpec/Dialect`.
```
bquorning, line 48: Looking at [the style used in rubocop/rubocop](https://github.com/rubocop/rubocop/blob/08a7244081df110c8513584c156ff31950218757/config/obsoletion.yml#L64-L81), I think this should be edited just a little bit:
```suggestion
      this cop has migrated to `RSpec/Dialect`. Please use `RSpec/Dialect` instead
```

bquorning (approved):

bquorning (approved):
Perfect ❤️

pirj (approved):
Incredible! Thank you!

One last smoke check - does rubocop-capybara’s spec suite still pass with this?

bquorning:
> does rubocop-capybara’s spec suite still pass with this?

I tried it locally, and the specs and most of the `rubocop` check passes.

bquorning:
I ran rubocop with this branch of rubocop-rspec on a rather large project at work, and the only unexpected issue I encountered was the problem with rubocop-rspec_rails described in #1884 (which is easily fixed _after_ our next release). This PR is ready to merge!

PR #1882, opened 2024-05-19
Add CI task testing with the oldest RuboCop version allowed by gemspec

bquorning:
Fixes: https://github.com/rubocop/rubocop-rspec/pull/1879#issuecomment-2107631095

> There is one thing we might be missing - it’s to test our code against our [minimum specified rubocop version](https://github.com/rubocop/rubocop-rspec/blob/6f340256dd1502cf23f20f6c6b7410e05de0c923/rubocop-rspec.gemspec#L40).

- [ ] Added tests.
- [ ] Updated documentation.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -70,6 +70,27 @@ jobs:
       - name: Use latest RuboCop from `master`
         run: |
           echo "gem 'rubocop', github: 'rubocop/rubocop'" > Gemfile.local
+          cat Gemfile.local
+      - uses: ruby/setup-ruby@v1
+        with:
+          ruby-version: "3.3"
+          bundler-cache: true
+      - run: NO_COVERAGE=true bundle exec rake ${{ matrix.task }}
+
+  oldest-rubocop:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        task:
+          - spec
+    name: "Oldest RuboCop: ${{ matrix.task }}"
+    steps:
+      - uses: actions/checkout@v4
+      - name: Use oldest RuboCop allowed by gemspec
+        run: |
+          sed -nr "s/spec.add_dependency 'rubocop', '~> ([0-9\.]+)'/gem 'rubocop', '= \1'/p" \
+            rubocop-rspec.gemspec > Gemfile.local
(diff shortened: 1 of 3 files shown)


pirj:
Nice!

can the failure be caused by some default TargetRubyVersion setting for that older RuboCop version?
I guess we may have new syntax used in our specs, and conditionally enable those examples, but they won’t parse if the TargetRubyVersion won’t match the RUBY_VERSION.
We have no need to test combinations of those two, right? We can just set the TRV to the RV.

bquorning:
I guess we _may_ be using a newer RUBY_VERSION than RuboCop can handle. E.g. if we use Ruby 3.3 and RuboCop 1.40 can only handle up to 3.1 or 3.2… Should we then rather test with the _oldest_ Ruby that we support or the most recent Ruby that RuboCop supports?

.github/workflows/main.yml
```diff
@@ -70,6 +70,27 @@ jobs:
...
+        task:
+          - spec
+    name: "Oldest RuboCop: ${{ matrix.task }}"
+    steps:
+      - uses: actions/checkout@v4
+      - name: Use oldest RuboCop allowed by gemspec
+        run: |
+          sed -nr "s/spec.add_dependency 'rubocop', '~> ([0-9\.]+)'/gem 'rubocop', '= \1'/p" \
+            rubocop-rspec.gemspec > Gemfile.local
+          cat Gemfile.local
```
pirj, line 93: Do you want to keep this line?

bquorning:
The problem is fixed; PR is ready for review.

pirj (approved):
Fantastic!
Frankly I was afraid that something’s broken and it’s just a coincidence that people update rubocop and rubocop-rspec simultaneously 😅

.github/workflows/main.yml
bquorning: Yes actually, I find it useful to have some readable output from most CI steps.

I’m not sure if you noticed, but I added a similar `cat` to the edge-rubocop job above. It may not be necessary there, but here in line 91–92 I think it may be hard to understand exactly _what_ is being written into Gemfile.local.

ydah (approved):
Looks good!

Issue #1868, opened 2024-04-09
Cop idea: Prevent using expect in let.


marocchino:
I've noticed that some people use let as an escape from the before hook to avoid `RSpec/ExpectInHook`.

so..

```rb
  let(:expect_let) do
    expect(Module).to receive(:call)
  end

  before do
    expect_let
  end
```
will not trigger `RSpec/ExpectInHook`.

There is no good reason to put expect inside let in the first place, so I think it would be better to disallow it.

marocchino:
https://github.com/betterspecs/betterspecs/issues/233 Maybe related this issue.

marocchino:
I will make pr part of `RSpec/ExpectInHook` as a let containing expect is called in before.

pirj:
If those people deliberately put expectations inside let’s to escape a cop, what would prevent them from defining a method like ‘def expect_foo’ and using it inside a ‘before’ hook?

I don’t think that our goal is to follow and attempt to flag all those nonsensical usages. We’re more like to enforce the style guide plus help enforce a given styling if several equally acceptable styles exist plus help people avoiding common mistakes.
Making an effort against deliberate abuse seems overkill.

pirj:
BetterSpecs is dead, long live rspec-style-guide that absorbed all good things from betterspecs.

PR #1885, opened 2024-05-21
Add new `RSpec/ExpectInLet` cop

yasu551:
Fix: https://github.com/rubocop/rubocop-rspec/issues/1868

- [x] Added tests.
- [x] Updated documentation.


lib/rubocop/cop/rspec/expect_in_let.rb
```diff
@@ -0,0 +1,44 @@
...
+        # @!method expectation(node)
+        def_node_search :expectation, '(send nil? #Expectations.all ...)'
+
+        def on_block(node)
+          return unless let?(node)
+          return if node.body.nil?
+
+          expectation(node.body) do |expect|
+            add_offense(expect.loc.selector, message: message(expect))
+          end
```
bquorning, line 31: `#let?` is a node matcher, so you can use the block form instead of a guard clause:
```suggestion
          return if node.body.nil?
          let?(node) do
            expectation(node.body) do |expect|
              add_offense(expect.loc.selector, message: message(expect))
            end
          end
```

bquorning (approved):
❤️

lib/rubocop/cop/rspec/expect_in_let.rb
pirj: I think it’s how it is done in ExpectInHook.
I’d keep for conformity.

spec/rubocop/cop/rspec/expect_in_let_spec.rb
```diff
@@ -0,0 +1,32 @@
...
+        ^^^^^^^^^^^^^^^^^^^^^^ Do not use `expect_any_instance_of` in let
+      end
+    RUBY
+  end
+
+  it 'accepts an empty let' do
+    expect_no_offenses(<<~RUBY)
+      let(:foo) {}
+    RUBY
+  end
```
pirj, line 21: Can you please add an example where ‘expect’ is the only statement in the ‘let’ block?

pirj (approved):
Thank you!

Even though I was opposed to handling the case that is no doubt a sabotage of a spec suite, but it isn’t a noticeable burden for us to support such a concise cop.

lib/rubocop/cop/rspec/expect_in_let.rb
bquorning: Ah yes, I didn’t notice. Let’s keep it this way.

spec/rubocop/cop/rspec/expect_in_let_spec.rb
bquorning: Probably just split the example in line 4 into three separate examples?

pirj:
https://www.githubstatus.com/incidents/nj47vccwm2zj

spec/rubocop/cop/rspec/expect_in_let_spec.rb
yasu551: Thank you for reviewing!
Should I keep the original source example that I split?

pirj: The point is that the AST and the node pattern for single-statement blocks and multi-statement blocks are different.
I’m being extra cautious here and have examples (where offences are added) for both cases.

Another case is when there are multiple offences (for multiple statements) all offensive statements are marked as such.

Basing on this, I suggest to structure this as:
- several examples with one statement each (you have those)
- One example with a combination of two statements (please add)
- One example with no offences and an empty ‘let’ block (you have this)
- One example with ‘it’ and no offences (you have this)

I’m happy to explain it further. Please correct me if I’m talking nonsense - it’s an evening of a dense day

And thanks for the persistence!

yasu551: > The point is that the AST and the node pattern for single-statement blocks and multi-statement blocks are different.

I see.  Thank you.

> One example with a combination of two statements (please add)

I did that.

pirj (approved):
Thank you! 🙌

pirj:
Last passengers, board the train, next station 3.0.0

ydah:
@yasu551 FYI, A version containing the cop you created has been released. Thank you for your contribution!
https://github.com/rubocop/rubocop-rspec/releases/tag/v2.30.0

PR #1889, opened 2024-05-22
Remove unnecessary `on_numblock` method

yasu551:
Because numbered parameters are rarely used with let.


ydah:
memo) It is true that the numbered parameter is not used for `let`, so it is disabled:
https://github.com/rubocop/rubocop-rspec/blob/306061220137110d325d1eac4a6a328750aab700/lib/rubocop/cop/rspec/empty_line_after_final_let.rb#L26

ydah (approved):
Looks good!

PR #1894, opened 2024-06-02
Backport `RSpec/ExpectInLet` cop to v2 branch

bquorning:
Backport the addition of new cop `RSpec/ExpectInLet` added in #1885 from the `main` branch to the `2-x-stable` branch.

We want to get as much feedback as possible on new cops before (eventually) enabling them when v3 is released.

______________________________________________________________________

Before submitting the PR make sure the following are checked:

- [x] Feature branch is up-to-date with ~master~ `2-x-stable` (if not - rebase it).
- [x] Squashed related commits together.
- [x] Added tests.
- [x] Updated documentation.
- [x] Added an entry to the `CHANGELOG.md` if the new code introduces user-observable changes.
- [x] The build (`bundle exec rake`) passes (be sure to run this locally, since it may produce updated documentation that you will need to commit).

If you have created a new cop:

- [x] Added the new cop to `config/default.yml`.
- [x] The cop is configured as `Enabled: pending` in `config/default.yml`.
- [x] The cop is configured as `Enabled: true` in `.rubocop.yml`.
- [x] The cop documents examples of good and bad code.
- [x] The tests assert both that bad code is reported and that good code is not reported.
- [x] Set `VersionAdded: "<<next>>"` in `default/config.yml`.

PR #1895, opened 2024-06-02
Run actions on 2-x-stable branch

bquorning:
Our (CI and gem-publish) workflows on the `master` branch are configured to run when changes happen to on `master` branch. For the same workflows to run on changes to the `2-x-stable` branch, we need to reconfigure the workflow triggers. (Note that the workflows are unchanged on `master`, so they will now trigger on _both_ branches).

### DIFF
--- a/.github/workflows/codespell.yml
+++ b/.github/workflows/codespell.yml
@@ -1,7 +1,6 @@
 name: CodeSpell
 on:
-  pull_request:
-    branches: 2-x-stable
+  - pull_request
 concurrency:
   group: ${{ github.workflow }}-${{ github.head_ref || github.run_id }}
   cancel-in-progress: true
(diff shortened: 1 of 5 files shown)


pirj:
I’m sorry, I’m lost. What is the 2-x-stable for?

bquorning:
> I’m sorry, I’m lost. What is the 2-x-stable for?

I saw your comment https://github.com/rubocop/rubocop-rspec/pull/1894#issuecomment-2144024083, but just in case anyone else are lost:

My apologies, I probably have never properly explained it. My idea with having a `2-x-stable` branch was to keep making v2.x releases of rubocop-rspec while we are using `master` branch to prepare the v3 release. Some of the changes that go into `master` can be backported to `2-x-stable`, and I think especially new cops should be backported. This ensures that we can get feedback on their use _before_ we release v3. If we don’t solicit such feedback, we can hardly make those new cops `enabled`, but would keep them as `pending` until v4 some time in the future.

ydah (approved):
Looks good

bquorning:
🤦🏼 I forgot to squash before merging. Sorry.

PR #1898, opened 2024-06-03
CI: Add GH_TOKEN to publish workflow

bquorning:
When trying to release a minor version, [GitHub Actions said](https://github.com/rubocop/rubocop-rspec/actions/runs/9347594123/job/25724913002#step:6:42):

```
gh: To use GitHub CLI in a GitHub Actions workflow, set the GH_TOKEN environment variable. Example:
  env:
    GH_TOKEN: ${{ github.token }}
```
we didn't notice it before, because this code doesn't run on patch releases.


ydah (approved):
Thank you!

Issue #1883, opened 2024-05-19
We also need support `AutoCorrect: contextual` option for LSP


ydah:
- https://github.com/rubocop/rubocop/pull/12657

For example, if `RSpec/Focus` is enabled, we will always have `focus` removed by the LSP.
https://docs.rubocop.org/rubocop-rspec/cops_rspec.html#rspecfocus

bquorning:
As far as I can tell, this is the current state of autocorrection for the different cops, sorted by “Always”, “Always (unsafe)”, “No”:

Always

- [ ] `RSpec/AlignLeftLetBrace`
- [ ] `RSpec/AlignRightLetBrace`
- [ ] `RSpec/BeEmpty`
- [ ] `RSpec/BeNil`
- [ ] `RSpec/ChangeByZero`
- [ ] `RSpec/ClassCheck`
- [ ] `RSpec/ContainExactly`
- [ ] `RSpec/ContextMethod`
- [ ] `RSpec/Dialect`
- [ ] `RSpec/DuplicatedMetadata`
- [ ] `RSpec/EmptyHook`
- [ ] `RSpec/EmptyLineAfterExample`
- [ ] `RSpec/EmptyLineAfterExampleGroup`
- [ ] `RSpec/EmptyLineAfterFinalLet`
- [ ] `RSpec/EmptyLineAfterHook`
- [ ] `RSpec/EmptyLineAfterSubject`
- [ ] `RSpec/EmptyMetadata`
- [ ] `RSpec/EmptyOutput`
- [ ] `RSpec/Eq`
- [ ] `RSpec/ExampleWording`
- [ ] `RSpec/ExcessiveDocstringSpacing`
- [ ] `RSpec/ExpectActual`
- [ ] `RSpec/Focus`
- [ ] `RSpec/HookArgument`
- [ ] `RSpec/HooksBeforeExamples`
- [ ] `RSpec/ImplicitExpect`
- [ ] `RSpec/ImplicitSubject`
- [ ] `RSpec/InstanceSpy`
- [ ] `RSpec/IsExpectedSpecify`
- [ ] `RSpec/ItBehavesLike`
- [ ] `RSpec/LeadingSubject`
- [ ] `RSpec/LetBeforeExamples`
- [ ] `RSpec/MatchArray`
- [ ] `RSpec/MetadataStyle`
- [ ] `RSpec/MultipleSubjects`
- [ ] `RSpec/NotToNot`
[...]

PR #1899, opened 2024-06-03
Support `AutoCorrect: contextual` option for LSP

ydah:
Fix: https://github.com/rubocop/rubocop-rspec/issues/1883

- [-] Added tests.
- [x] Updated documentation.


bquorning (approved):
I think it’s hard to tell if this is the correct list of cops or not, but it’s a good start. We can adjust the list based on real-world use and the resulting bug reports.

PR #1901, opened 2024-06-03
`AutoCorrect: contextual` needs rubocop v1.61

ydah:
https://github.com/rubocop/rubocop/releases/tag/v1.61.0

- [-] Added tests.
- [-] Updated documentation.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Master (Unreleased)
 
+- Bump RuboCop requirement to +1.61. ([@ydah])
+
 ## 3.0.0 (2024-06-11)
 
 - Remove extracted cops in `Capybara`, `FactoryBot` and `Rails` departments. ([@ydah])
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Master (Unreleased)
 
+## 3.0.1 (2024-06-11)
+
 - Bump RuboCop requirement to +1.61. ([@ydah])
 
 ## 3.0.0 (2024-06-11)
--- a/lib/rubocop/rspec/version.rb
+++ b/lib/rubocop/rspec/version.rb
@@ -4,7 +4,7 @@ module RuboCop
   module RSpec
     # Version information for the RSpec RuboCop plugin.
     module Version
-      STRING = '3.0.0'
+      STRING = '3.0.1'
     end
   end
 end
--- a/rubocop-rspec.gemspec
+++ b/rubocop-rspec.gemspec
@@ -37,5 +37,5 @@ Gem::Specification.new do |spec|
     'rubygems_mfa_required' => 'true'
   }
 
-  spec.add_dependency 'rubocop', '~> 1.40'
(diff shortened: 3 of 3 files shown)


pirj:
The is is a big jump
Can we make it together with the 3.0 release, or in 3.x?

How did you catch this?
I see that https://github.com/rubocop/rubocop-rspec_rails/commit/ff9bf2a24a4157cd58c51c92f8e31a98571c69cc is :x:, but I don’t see why.
 Do we have this build job that checks with the lowest supported rubocop version?

bquorning:
This makes me a bit hesitant to backport our contextual autocorrection to 2-x-stable.

bquorning:
> Do we have this build job that checks with the lowest supported rubocop version?

That would be https://github.com/rubocop/rubocop-rspec/actions/runs/9355023880/job/25749240673, using `gem 'rubocop', '= 1.40'` on e49f221245f8e23fd1167677f6557623d9d702ce.

bquorning:
> I see that [rubocop/rubocop-rspec_rails@ff9bf2a](https://github.com/rubocop/rubocop-rspec_rails/commit/ff9bf2a24a4157cd58c51c92f8e31a98571c69cc) is ❌, but I don’t see why.

The failing CI is because https://github.com/rubocop/rubocop-rspec_rails/pull/30 is not (and cannot be) merged yet.

bquorning:
I’m trying to read https://github.com/rubocop/rubocop/pull/12657 and I _think_ if people are using an older version of RuboCop than v1.61, autocorrect will just keep working as it always has. But if they’re using v1.61+ then they get the new LSP feature too.

In conclusion, I don’t think we _need_ this PR. People who want to use LSP should use RuboCop v1.61+ anyway. But on the other hand, the target branch is `master` which is targeted at our v3.0, and we can decide to _not_ backport to `2-x-stable`, right?

ydah:
@bquorning Thanks for catching that. You're right, it is indeed. I'm going to close this PR since there is no need to necessarily update version it.

bquorning:
Yeah, maybe we should have merged this before v3.0.0.
Reopening the PR as maybe we want to do it as a bugfix in v3.0.1?

ydah:
Let's do that. @pirj WDYT?

bquorning:
I added one commit, so that merging this PR will release v3.0.1.

I feel like I have been a little too eager to merge and release lately, so I’ll wait for @pirj to review before merging.

pirj:
Let’s merge and release as 3.0.1, i see no big harm here.

The only problem would be people doing `bundle update —conservative rubocop-rspec`, which would resolve to 3.0.0 as 3.0.1 would depend on newer rubocop and won’t match.

I’d be most interested in why our CI “oldest rubocop” job didn’t catch this.

ydah:
In the following sections, if AutoCorrect was set to a value other than boolean, the corresponding error message was output, so it was necessary to update to v1.61. So, let's change it and release it.

https://github.com/rubocop/rubocop/blob/edb56c927d4d1112a9c02990da31767e68d2121c/lib/rubocop/config_validator.rb#L251-L268

PR #1906, opened 2024-06-06
Move node patterns into private scope

bquorning:
Private code should be private. I think this makes it easier to read and understand the code.

I am not sure if this is a code style we should enforce with a cop? We haven’t done so with the previous style, and yet it was uses consistently on all cops except two.

If we don’t have a cop to enforce this style, it’s something we need to be aware of when doing code review.

- [ ] Added tests.
- [ ] Updated documentation.

### DIFF
--- a/lib/rubocop/cop/rspec/be_empty.rb
+++ b/lib/rubocop/cop/rspec/be_empty.rb
@@ -19,6 +19,16 @@ module RuboCop
         MSG = 'Use `be_empty` matchers for checking an empty array.'
         RESTRICT_ON_SEND = %i[contain_exactly match_array].freeze
 
+        def on_send(node)
+          expect_array_matcher?(node.parent) do |expect|
+            add_offense(expect) do |corrector|
+              corrector.replace(expect, 'be_empty')
+            end
+          end
+        end
+
+        private
+
         # @!method expect_array_matcher?(node)
         def_node_matcher :expect_array_matcher?, <<~PATTERN
           (send
@@ -31,14 +41,6 @@ module RuboCop
             _?
           )
         PATTERN
-
-        def on_send(node)
-          expect_array_matcher?(node.parent) do |expect|
-            add_offense(expect) do |corrector|
-              corrector.replace(expect, 'be_empty')
-            end
-          end
-        end
       end
     end
   end
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -47,6 +47,11 @@ Lint/InterpolationCheck:
 Lint/RedundantCopDisableDirective:
   Enabled: false
 
+Lint/UselessAccessModifier:
+  MethodCreatingMethods:
+    - def_node_matcher
+    - def_node_search
+
 Metrics/BlockLength:
   Exclude:
     - rubocop-rspec.gemspec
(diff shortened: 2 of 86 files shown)


ydah (approved):
Looks good!

lib/rubocop/cop/rspec/be_empty.rb
```diff
@@ -19,6 +19,16 @@ class BeEmpty < Base
         MSG = 'Use `be_empty` matchers for checking an empty array.'
         RESTRICT_ON_SEND = %i[contain_exactly match_array].freeze
 
+        def on_send(node)
```
pirj, line 22: I think we can go even further, and make on_hooks private, too.
To my best memory, the ‘inspect’ method on a base class calls them.

But should we?

I would rather revert this change to reduce code churn.

Please tell me if there is a real issue with exposing “non-public” parts of cops that I’m missing.

cc @rubocop/rubocop-rspec

bquorning: The hooks are being called with `public_send`, so they should remain private: https://github.com/rubocop/rubocop/blob/89aa2aa2a32e76c1ddd8199d17d50693602d1d73/lib/rubocop/cop/commissioner.rb#L107

The main reason for this change is subjective (and I’m sorry for rushing to merge this PR instead of waiting for feedback): When reading pull requests on this repo, I always find myself having to scroll past the node patterns before reaching the callback methods. I generally find code easier to read if the “entry points” (the public API) is at the top, and (if possible) all method calls go “downwards”.

We can revert if you insist, but I think the upcoming v3 release is a good time _if_ we want to introduce code churn.

pirj: > code easier to read if the “entry points”

I strongly agree with this. However, this depends on what you consider an entry point.

For me, it’s either an on_block or on_send calling the very first node pattern, passing it a block with a call to add_offense and corrector inside.
So the entry is the first node pattern.
And by the node pattern it’s visible if it matches a send or a block, so even the hook name doesn’t matter much.

Also, node patterns can be documented. Even better than hooks. And good examples of that are both “method”-level docs showing what matches and what doesn’t; and also line-level documentation.
We don’t enforce this, but I recall writing those. [Example](https://github.com/rubocop/rubocop-rspec/blob/2b69a04dbb5a45a7df49f194669d56c0d0908f45/lib/rubocop/cop/rspec/stubbed_mock.rb#L34):
```ruby

        #     receive_message_chain(:foo, :bar)
        #
        #   @example source that matches
        #     receive(:foo).with('bar')
        #
        #   @param node [RuboCop::AST::Node]
        def_node_matcher :message_expectation?, <<~PATTERN
          {
            (send nil? { :receive :receive_message_chain } ...)  # receive(:foo)
            (send (send nil? :receive ...) :with ...)            # receive(:foo).with('bar')
          }
        PATTERN
```

Like in this cop,
```ruby
        def on_send(node)
          expect_array_matcher?(node.parent) do |expect|
            add_offense(expect) do |corrector|
              corrector.replace(expect, 'be_empty')
```
provides little information on what the cop does. I consider this boilerplate code.
Yes, autocorrection is there.

On the contrary, `expect_array_matcher?` tells what the cop matches on.

How are our parent project cops structured?
I don’t think any of them is using node pattern as intensively as our cops, though.

bquorning: Ah jeez, I think you’re right. Again.
Let me open a new PR in a bit where I partially revert this one, so only “non-entrylevel patterns” are made into private methods.

PR #1909, opened 2024-06-07
CI: When publishing, create a new branch per tag

bquorning:
When doing a release, there _may_ be a `switch-docs-version` still in the repo. We lower the risk of such collisions by adding the tag value to the branch name.

cc @ydah re. https://github.com/rubocop/rubocop-rspec/actions/runs/9412705266/job/25928071688

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -33,11 +33,12 @@ jobs:
         run: |
           sed -i 's/version:.*$/version: ~/' docs/antora.yml
           if ! git diff --exit-code docs/antora.yml; then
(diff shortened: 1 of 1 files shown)


.github/workflows/publish.yml
```diff
@@ -1,7 +1,7 @@
 name: Publish
 on:
   push:
-    branches: master
+    branches: 2-x-stable
```
bquorning, line 4: This is wrong, sorry. 
```suggestion
    branches: master
```

ydah:
@bquorning could you squash commits?

PR #1912, opened 2024-06-09
Fix an error for `RSpec/ScatteredSetup` when one of the hooks is an empty block

Earlopain:
The produced autocorrect is not the best but there are other cops that will push this into something more pleasing.

- [x] Added tests.
- [ ] Updated documentation.
- x ] Added an entry to the `CHANGELOG.md` if the new code introduces user-observable changes.


bquorning (approved):

ydah (approved):
Thank you!

PR #1918, opened 2024-06-10
[Fix #1913] Fix broken autocorrect for `RSpec/ScatteredSetup` when block contains heredoc

### LINKED ISSUE #1913: Wrong autocorrect for `RSpec/ScatteredSetup` when block contains heredoc
Here's a testcase and roughtly what I would expect:

```rb
it 'flags hooks that contain heredoc arguments and autocorrects correctly' do
  expect_offense(<<~RUBY)
    describe Foo do
      before { foo }
      ^^^^^^^^^^^^^^ Do not define multiple `before` hooks in the same example group (also defined on line 3).
      before do
      ^^^^^^^^^ Do not define multiple `before` hooks in the same example group (also defined on line 2).
        bar(<<~'TEXT')
          Hello World!
        TEXT
      end
    end
  RUBY

  expect_correction(<<~RUBY)
    describe Foo do
      before {
        foo
        bar(<<~'TEXT')
          Hello World!
        TEXT
      }
    end
  RUBY
end
```

The actual autocorrect looks like this, which is not valid syntax (among other things):

```rb
describe Foo do
  before { foo
bar(<<~TEXT) }

end
```

Earlopain:
Autocorrect previously looked like this which not correct syntax:
```rb
describe Foo do
  before { foo
bar(<<~TEXT) }

end
```

The block source doesn't contain the heredoc source: I initially thought this was a bug somewhere else but turns out this is just ruby. https://github.com/rubocop/rubocop-ast/issues/293 contains more context if you are confused as I was.

This shifts a few test results around but that seems fine. The cop doesn't need to produce beautiful results by itself. Just resizing the range by 1 wouldn't work since code like this is valid: `before{}`, handling that would just make it even more complicated.

- [x] Added tests.
- [ ] Updated documentation.


Earlopain:
Oh, I put the changelog entry at the wrong place. I can fix that in an hour or so

lib/rubocop/cop/rspec/scattered_setup.rb
```diff
@@ -75,8 +75,14 @@ def message(occurrences, occurrence)
         def autocorrect(corrector, first_occurrence, occurrence)
           return if first_occurrence == occurrence || !first_occurrence.body
 
+          # Heredocs are not part of the blocks source. Assume that
+          # the block consists only of code between block open/end,
+          # no funky heredoc business.
+          loc_after_open = occurrence.loc.begin.end
+          loc_after_close = occurrence.loc.end.begin
+          block_range = loc_after_open.join(loc_after_close)
```
bquorning, line 83: I’m ok with this implementation, but I cannot help but wonder if this isn’t a problem in other cops too. And if so, is this really a bug in rubocop-ast, or ast?

Darhazer: But that’s why we have the MoveNode helper (and the FinalEndLocation). Haven’t looked at the issue/PR but wouldn’t they solve the problem?

Earlopain: > I’m ok with this implementation, but I cannot help but wonder if this isn’t a problem in other cops too. And if so, is this really a bug in rubocop-ast, or ast?

I thought so too and opened https://github.com/rubocop/rubocop-ast/issues/293 which then contained this example:

```rb
<<~'FOO' + foo {<<~'TEXT'}
  This is FOO
FOO
    Hello World!
TEXT
```

What to me logically belongs to the block is not contiguous (I would hope this is fringe code and nobody actually writes this). After searching for cops with the same issue I couldn't actually find a single one so it didn't feel worth it to try and and something to `rubocop-ast` for this.

> But that’s why we have the MoveNode helper (and the FinalEndLocation). Haven’t looked at the issue/PR but wouldn’t they solve the problem?

I'll check that out tomorrow, would be nice if that hides this mess away. FinalEndLocation looks particularly relevant

Earlopain: I couldn't make it work with `MoveNode` but the method from `FinalEndLocation` works somewhat nicely.

bquorning (approved):
Thank you.

lib/rubocop/cop/rspec/scattered_setup.rb
Darhazer: Ah yes, `MoveNode` won't work here, because we are not moving the hook before or after another one, we are combining their bodies 👍

Darhazer (approved):
LGTM

PR #1930, opened 2024-06-19
Fix false negative for `RSpec/PredicateMatcher` when expectation contains custom failure message

### LINKED ISSUE #1929: False negative for `RSpec/PredicateMatcher` when the expectations includes a custom assertion message
Config:

```yml
RSpec/PredicateMatcher:
  Strict: false
```

I have the following code:

```rb
expect(cop_names_without_department.include?(cop_name)).to be(false), "Missing department for #{cop_name}."
```

which I would expect to correct to:

```rb
expect(cop_names_without_department).not_to include(cop_name), "Missing department for #{cop_name}."
```

This happens because of the custom failure message at the end, offense happens correctly when its not present.

Earlopain:
Closes #1929

This leaves `predicate_matcher_block?` since RSpec doesn't seem to handle that case.

- [x] Added tests.
- [ ] Updated documentation.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## Master (Unreleased)
 
 - Fix wrong autocorrect for `RSpec/ScatteredSetup` when hook contains heredoc. ([@earlopain])
+- Fix false negative for `RSpec/PredicateMatcher` when expectation contains custom failure message. ([@earlopain])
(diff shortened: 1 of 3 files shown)


bquorning (approved):

pirj (approved):
Thank you!

PR #1940, opened 2024-07-17
Fix false-negative for `UnspecifiedException` when matcher is chained

r7kamura:
I noticed that this cop could not detect the following cases, so fixed it.

```ruby
expect {
  foo
}.to raise_exception.and change { bar }
```

- [x] Added tests.
- [ ] Updated documentation.

### DIFF
--- a/lib/rubocop/cop/rspec/unspecified_exception.rb
+++ b/lib/rubocop/cop/rspec/unspecified_exception.rb
@@ -32,34 +32,39 @@ module RuboCop
       #
       class UnspecifiedException < Base
         MSG = 'Specify the exception being captured'
-        RESTRICT_ON_SEND = %i[to].freeze
 
-        # @!method empty_raise_error_or_exception(node)
-        def_node_matcher :empty_raise_error_or_exception, <<~PATTERN
-          (send
-            (block
-                (send nil? :expect) ...)
-            :to
-            (send nil? {:raise_error :raise_exception})
-          )
+        RESTRICT_ON_SEND = %i[
+          raise_exception
+          raise_error
+        ].freeze
+
+        # @!method expect_to?(node)
+        def_node_matcher :expect_to?, <<~PATTERN
+          (send (block (send nil? :expect) ...) :to ...)
         PATTERN
 
         def on_send(node)
           return unless empty_exception_matcher?(node)
 
-          add_offense(node.children.last)
+          add_offense(node)
         end
 
         private
 
         def empty_exception_matcher?(node)
-          empty_raise_error_or_exception(node) && !block_with_args?(node.parent)
+          return false if node.arguments? || node.block_literal?
+
+          expect_to = find_expect_to(node)
+          return false unless expect_to
+          return false if expect_to.block_node&.arguments?
+
+          true
         end
 
-        def block_with_args?(node)
-          return false unless node&.block_type?
-
-          node.arguments?
+        def find_expect_to(node)
+          node.each_ancestor(:send).find do |ancestor|
+            expect_to?(ancestor)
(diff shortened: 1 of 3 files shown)


spec/rubocop/cop/rspec/unspecified_exception_spec.rb
```diff
@@ -171,5 +171,14 @@
         }.to raise_exception(my_exception)
       RUBY
     end
+
+    it 'detects chained offenses' do
+      expect_offense(<<~RUBY)
+        expect {
+          foo
+        }.to raise_exception.and change { bar }
```
pirj, line 179: What if it’s `raise_error().and change {}.and change {}`?

r7kamura: In the end, I decided to radically change the implementation to support the cases I commented on.
Instead of searching from `to` node, I will change it to search from `raise_exception` or `raise_error` nodes. I think this will eliminate the possibility of false-negatives and make it easier to report error locations.

lib/rubocop/cop/rspec/unspecified_exception.rb
```diff
@@ -32,34 +32,39 @@ module RSpec
...
         def empty_exception_matcher?(node)
-          empty_raise_error_or_exception(node) && !block_with_args?(node.parent)
-        end
+          return false if node.arguments? || node.block_literal?
 
-        def block_with_args?(node)
-          return false unless node&.block_type?
+          expect_to = find_expect_to(node)
+          return false unless expect_to
+          return false if expect_to.block_node&.arguments?
```
pirj, line 59: Is this checking `expect { |args| … }.to raise_error`?
Why do we ignore if there are any, how do we care?
Do we have a an example that will fail if we remove this check?

lib/rubocop/cop/rspec/unspecified_exception.rb
```diff
@@ -32,34 +32,39 @@ module RSpec
...
         private
 
         def empty_exception_matcher?(node)
-          empty_raise_error_or_exception(node) && !block_with_args?(node.parent)
-        end
+          return false if node.arguments? || node.block_literal?
 
-        def block_with_args?(node)
-          return false unless node&.block_type?
+          expect_to = find_expect_to(node)
```
pirj, line 57: I was thinking we could simplify all the following to just `node.each_ancestor.any? { |ancestor| expect_to?(ancestor) }`

lib/rubocop/cop/rspec/unspecified_exception.rb
r7kamura: If you want to test the raised error in detail, `to` may be called with a block with an argument as follows:

```ruby
expect { foo }.to raise_error do |error|
  expect(error.bar).to eq("baz")
end
```

If this code is removed, the following test will fail:

https://github.com/rubocop/rubocop-rspec/blob/ce4cd77d2ac1ea03ff8cd136e913aeb9bf12b80a/spec/rubocop/cop/rspec/unspecified_exception_spec.rb#L70-L78

pirj: Understood.
Such syntax is confusing as hell. The block is not passed to `to`, but `raise_error` itself, isn’t it? 

 I recall there’s some poke-yoke warning in RSpec warning users when the block goes actually to the wrong one.

However, yes, I now recall that we decided that such syntax should be tolerated as it is expected that aome check agains what the exception is will be made inside that block.

pirj (approved):
Thank you!

ydah (approved):
Thank you!

PR #1947, opened 2024-08-08
Fix false-negative and error for `RSpec/MetadataStyle`

cbliard:
When RSpec/MetadataStyle cop is used with `EnforcedStyle: hash`, offenses must be registered only for metadata arguments being symbols.

This code:
```ruby
describe 'Something', a, b do
end
```

would previously report
```ruby
describe 'Something', a, b do
                      ^ Use hash style for metadata.
end
```

And this code:
```ruby
describe 'Something', a, :b do
end
```

would previously produce the following exception: "NoMethodError: undefined method `value' for an instance of RuboCop::AST::SendNode".

- [x] Added tests.
- [ ] Updated documentation.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Master (Unreleased)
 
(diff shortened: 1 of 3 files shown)


pirj (approved):
Thank you!

bquorning (approved):

PR #1948, opened 2024-08-12
Fix `RSpec/SortMetadata` cop to sort only trailing metadata

### LINKED ISSUE #1946: `RSpec/SortMetadata` autocorrect can break context groups having additional description
1. Create sample RSpec file like this:

```ruby
RSpec.describe "Something", "in this context", :a, :b do
  subject { true }

  it "works well" do
    expect(subject).not_to be false
  end
end
```

2. Run it with documentation formatter: `rspec --format documentation spec/test_spec.rb`

```
Something in this context
  works well

Finished in 0.00172 seconds (files took 0.07348 seconds to load)
1 example, 0 failures
```

3. Run autocorrect: `rubocop -A spec/test_spec.rb`

```
Inspecting 1 file
C

Offenses:

spec/test_spec.rb:1:29: C: [Corrected] RSpec/SortMetadata: Sort metadata alphabetically.
RSpec.describe "Something", "in this context", :a, :b do
                            ^^^^^^^^^^^^^^^^^^^^^^^^^

1 file inspected, 1 offense detected, 1 offense corrected
```

File updated to:
```ruby
RSpec.describe "Something", :a, :b, "in this context" do
  subject { true }

  it "works well" do
    expect(subject).not_to be false
  end
end
```

4. Run it again: `rspec --format documentation spec/test_spec.rb`

```
An error occurred while loading ./spec/test_spec.rb.
Failure/Error:
  RSpec.describe "Something", :a, :b, "in this context" do
    subject { true }

    it "works well" do
      expect(subject).not_to be false
    end
  end

ArgumentError:
  wrong number of arguments (given 4, expected 0..2)
# ./spec/test_spec.rb:1:in `<top (required)>'
No examples found.

Finished in 0.00001 seconds (files took 0.10503 seconds to load)
0 examples, 0 failures, 1 error occurred outside of examples
```

The number of arguments expected is from `#build_description_from(parent_description=nil, my_description=nil)` in `lib/rspec/core/metadata.rb` from rspec-core.

We sometimes use this ability to have an additional description parameter with `RSpec.describe` when writing spec for a specific feature flag or to split up a spec file in smaller parts when it becomes too big. Occasionally the `RSpec/SortMetadata` breaks it depending on the metadata we use by swapping the description and the symbols like above. We disable the cop for the line and that's ok, but it would be nice to have it fixed upstream.

It seems like it is somewhat an expected behavior though: https://github.com/rubocop/rubocop-rspec/blob/ff213aecd8579c638631d37a5de05468685a405a/spec/rubocop/cop/rspec/sort_metadata_spec.rb#L98-L111

cbliard:
Fixes #1946.

Symbols in metadata are processed by RSpec only when they are positioned last, meaning the other parameter types must be positioned before the symbols. RSpec `context`/`describe` accepts second non-symbol argument as an additional description.

Previously this cop was sorting all arguments, regardless of their type, which could lead to having the second string argument (the additional description) moved to another place (see #1946 for details). This PR fixes it.

Questions for reviewers:
- ~~is it ok to sort strings before variables? I have the feeling that this cop's responsibility should be to just take care of sorting symbols at the end of the arguments list.~~ As discussed during the review, only the last metadata arguments are considered (symbols and final hash argument).
- ~~the message "Sort metadata alphabetically." does not fully reflect what the cop is doing anymore, and can be confusing for code like `describe 'Something', :a, b, :c` because the metadata *is* sorted. I could not come up with a good message. Suggestions?~~ As only the trailing metadata is sorted now, the message seems ok.
- I noticed an autocorrect issue with `describe 'Something', :b, :a, { foo: :bar }`: it gets autocorrected to `describe 'Something', :b, :a, foo: :bar }`. Should it be fixed in the same PR?
- I updated the `RuboCop::Cop::RSpec::Metadata#on_metadata_arguments` method because it was skipping the last argument if it was not a hash. I also renamed `symbols` to `metadata_arguments` (or `args` in `RSpec/SortMetadata` cop). Should `symbols` be renamed to `metadata_arguments` or `args` in other cops relying on `on_metadata` too?

- [x] Added tests.
- [ ] Updated documentation.

### DIFF
--- a/spec/rubocop/cop/rspec/sort_metadata_spec.rb
+++ b/spec/rubocop/cop/rspec/sort_metadata_spec.rb
@@ -4,7 +4,7 @@ RSpec.describe RuboCop::Cop::RSpec::SortMetadata do
   it 'does not register an offense when using only symbol metadata ' \
      'in alphabetical order' do
     expect_no_offenses(<<~RUBY)
-      RSpec.describe 'Something', :a, :b do
+      describe 'Something', :a, :b do
       end
     RUBY
   end
@@ -12,13 +12,72 @@ RSpec.describe RuboCop::Cop::RSpec::SortMetadata do
   it 'registers an offense when using only symbol metadata, ' \
      'but not in alphabetical order' do
     expect_offense(<<~RUBY)
-      RSpec.describe 'Something', :b, :a do
-                                  ^^^^^^ Sort metadata alphabetically.
+      describe 'Something', :b, :a do
+                            ^^^^^^ Sort metadata alphabetically.
       end
     RUBY
 
     expect_correction(<<~RUBY)
-      RSpec.describe 'Something', :a, :b do
+      describe 'Something', :a, :b do
+      end
+    RUBY
+  end
+
+  it 'does not register an offense for a symbol metadata before a non-symbol ' \
+     'argument' do
+    expect_no_offenses(<<~RUBY)
+      describe 'Something', :z, :a, variable, foo: :bar do
+      end
+    RUBY
+  end
+
+  it 'registers an offense only for trailing symbol metadata not in ' \
+     'alphabetical order' do
+    expect_offense(<<~RUBY)
+      describe 'Something', :z, :a, variable, :c, :b do
+                                              ^^^^^^ Sort metadata alphabetically.
+      end
+    RUBY
+
+    expect_correction(<<~RUBY)
+      describe 'Something', :z, :a, variable, :b, :c do
+      end
+    RUBY
+  end
+
+  it 'registers an offense when a symbol metadata not in alphabetical order ' \
+     'is before a variable argument being the last argument ' \
+     'as it could be a hash' do
+    expect_offense(<<~RUBY)
+      describe 'Something', :z, :a, some_hash do
+                            ^^^^^^ Sort metadata alphabetically.
+      end
+    RUBY
+
+    expect_correction(<<~RUBY)
+      describe 'Something', :a, :z, some_hash do
+      end
+    RUBY
+  end
+
+  it 'does not register an offense when using a second level description ' \
+     'not in alphabetical order with symbol metadata' do
+    expect_no_offenses(<<~RUBY)
+      describe 'Something', 'second docstring', :a, :b do
+      end
+    RUBY
+  end
+
+  it 'registers an offense when using a second level description ' \
+     'and metadata not in alphabetical order' do
+    expect_offense(<<~RUBY)
+      describe 'Something', 'second docstring', :b, :a do
+                                                ^^^^^^ Sort metadata alphabetically.
+      end
+    RUBY
+
+    expect_correction(<<~RUBY)
+      describe 'Something', 'second docstring', :a, :b do
       end
     RUBY
   end
@@ -100,27 +159,27 @@ RSpec.describe RuboCop::Cop::RSpec::SortMetadata do
      'and the hash values are complex objects' do
     expect_offense(<<~RUBY)
       it 'Something', variable, 'B', :a, key => {}, foo: ->(x) { bar(x) }, Identifier.sample => true, baz: Snafu.new do
-                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ Sort metadata alphabetically.
+                                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ Sort metadata alphabetically.
       end
     RUBY
 
     expect_correction(<<~RUBY)
-      it 'Something', :a, 'B', variable, baz: Snafu.new, foo: ->(x) { bar(x) }, Identifier.sample => true, key => {} do
+      it 'Something', variable, 'B', :a, baz: Snafu.new, foo: ->(x) { bar(x) }, Identifier.sample => true, key => {} do
       end
     RUBY
   end
 
   it 'registers an offense only when example or group has a block' do
     expect_offense(<<~RUBY)
-      shared_examples 'a difficult situation', 'B', :a do |x, y|
-                                               ^^^^^^^ Sort metadata alphabetically.
+      shared_examples 'a difficult situation', 'B', :z, :a do |x, y|
+                                                    ^^^^^^ Sort metadata alphabetically.
       end
 
       include_examples 'a difficult situation', 'value', 'another value'
     RUBY
 
     expect_correction(<<~RUBY)
-      shared_examples 'a difficult situation', :a, 'B' do |x, y|
+      shared_examples 'a difficult situation', 'B', :a, :z do |x, y|
       end
 
       include_examples 'a difficult situation', 'value', 'another value'
@@ -129,14 +188,14 @@ RSpec.describe RuboCop::Cop::RSpec::SortMetadata do
 
   it 'registers an offense also when the metadata is not on one single line' do
     expect_offense(<<~RUBY)
-      RSpec.describe 'Something', :foo, :bar,
-                                  ^^^^^^^^^^^ Sort metadata alphabetically.
+      describe 'Something', :foo, :bar,
+                            ^^^^^^^^^^^ Sort metadata alphabetically.
                                    baz: 'goo' do
       end
     RUBY
 
     expect_correction(<<~RUBY)
-      RSpec.describe 'Something', :bar, :foo, baz: 'goo' do
+      describe 'Something', :bar, :foo, baz: 'goo' do
       end
     RUBY
   end
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Master (Unreleased)
 
+- Fix `RSpec/SortMetadata` cop to limit sorting to trailing metadata arguments. ([@cbliard])
+
 ## 3.3.0 (2024-12-12)
 
 - Deprecate `top_level_group?` method from `TopLevelGroup` mixin as all of its callers were intentionally removed from `Rubocop/RSpec`. ([@corsonknowles])
--- a/docs/modules/ROOT/pages/cops_rspec.adoc
+++ b/docs/modules/ROOT/pages/cops_rspec.adoc
@@ -5735,6 +5735,8 @@ end
 
 Sort RSpec metadata alphabetically.
 
+Only the trailing metadata is sorted.
+
 [#examples-rspecsortmetadata]
 === Examples
 
@@ -5749,6 +5751,10 @@ it 'works', :b, :a, foo: 'bar', baz: true
 describe 'Something', :a, :b
 context 'Something', baz: true, foo: 'bar'
 it 'works', :a, :b, baz: true, foo: 'bar'
+
+# good, trailing metadata is sorted
+describe 'Something', 'description', :a, :b, :z
+context 'Something', :z, variable, :a, :b
 ----
 
 [#references-rspecsortmetadata]
--- a/lib/rubocop/cop/rspec/mixin/metadata.rb
+++ b/lib/rubocop/cop/rspec/mixin/metadata.rb
@@ -47,15 +47,12 @@ module RuboCop
         private
 
         def on_metadata_arguments(metadata_arguments)
-          *symbols, last = metadata_arguments
-          hash = nil
-          case last&.type
-          when :hash
-            hash = last
-          when :sym
-            symbols << last
+          if metadata_arguments.last&.hash_type?
+            *metadata_arguments, hash = metadata_arguments
+            on_metadata(metadata_arguments, hash)
+          else
+            on_metadata(metadata_arguments, nil)
           end
-          on_metadata(symbols, hash)
         end
       end
     end
--- a/lib/rubocop/cop/rspec/sort_metadata.rb
+++ b/lib/rubocop/cop/rspec/sort_metadata.rb
@@ -5,6 +5,8 @@ module RuboCop
     module RSpec
       # Sort RSpec metadata alphabetically.
       #
+      # Only the trailing metadata is sorted.
+      #
       # @example
       #   # bad
       #   describe 'Something', :b, :a
@@ -16,6 +18,9 @@ module RuboCop
       #   context 'Something', baz: true, foo: 'bar'
       #   it 'works', :a, :b, baz: true, foo: 'bar'
       #
+      #   # good, trailing metadata is sorted
+      #   describe 'Something', 'description', :a, :b, :z
+      #   context 'Something', :z, variable, :a, :b
       class SortMetadata < Base
         extend AutoCorrector
         include Metadata
@@ -23,8 +28,14 @@ module RuboCop
 
         MSG = 'Sort metadata alphabetically.'
 
-        def on_metadata(symbols, hash)
+        # @!method match_ambiguous_trailing_metadata?(node)
+        def_node_matcher :match_ambiguous_trailing_metadata?, <<~PATTERN
+          (send _ _ _ ... !{hash sym str dstr xstr})
+        PATTERN
+
+        def on_metadata(args, hash)
           pairs = hash&.pairs || []
+          symbols = trailing_symbols(args)
           return if sorted?(symbols, pairs)
 
           crime_scene = crime_scene(symbols, pairs)
@@ -34,6 +45,15 @@ module RuboCop
         end
 
         private
+
+        def trailing_symbols(args)
+          args = args[...-1] if last_arg_could_be_a_hash?(args)
+          args.reverse.take_while(&:sym_type?).reverse
+        end
+
+        def last_arg_could_be_a_hash?(args)
+          args.last && match_ambiguous_trailing_metadata?(args.last.parent)
+        end
 
         def crime_scene(symbols, pairs)
           metadata = symbols + pairs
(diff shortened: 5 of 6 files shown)


spec/rubocop/cop/rspec/sort_metadata_spec.rb
```diff
@@ -23,6 +23,50 @@
     RUBY
   end
 
+  it 'registers an offense when a symbol metadata is before second docstring ' \
+     'argument' do
+    expect_offense(<<~RUBY)
+      RSpec.describe 'Something', :a, :z, 'second docstring' do
```
pirj, line 29: Invalid RSpec syntax:
```
ArgumentError:
  wrong number of arguments (given 4, expected 0..2)
```

spec/rubocop/cop/rspec/sort_metadata_spec.rb
```diff
@@ -23,6 +23,50 @@
...
+    expect_correction(<<~RUBY)
+      RSpec.describe 'Something', 'second docstring', :a, :z do
+      end
+    RUBY
+  end
+
+  it 'registers an offense when a symbol metadata is before a variable ' \
+     'argument' do
+    expect_offense(<<~RUBY)
+      RSpec.describe 'Something', :a, :z, variable, foo: :bar do
```
pirj, line 43: If the `variable` holds something that is not a symbol, this will fail with a similar ArgumentError

spec/rubocop/cop/rspec/sort_metadata_spec.rb
```diff
@@ -23,6 +23,50 @@
...
+    expect_correction(<<~RUBY)
+      RSpec.describe 'Something', variable, :a, :z, foo: :bar do
+      end
+    RUBY
+  end
+
+  it 'does not register an offense when a symbol metadata is before a ' \
+     'variable argument being the last argument as it could be a hash' do
+    expect_no_offenses(<<~RUBY)
+      RSpec.describe 'Something', :a, :z, some_hash do
```
pirj, line 57: What if it's a
```
RSpec.describe 'Something', :z, :a, some_hash do
```
?

spec/rubocop/cop/rspec/sort_metadata_spec.rb
```diff
@@ -23,6 +23,50 @@
...
+      end
+    RUBY
+  end
+
+  it 'does not register an offense when using a second level description ' \
+     'not in alphabetical order with symbol metadata' do
+    expect_no_offenses(<<~RUBY)
+      RSpec.describe 'Something', 'second docstring', :a, :b do
```
pirj, line 65: Invalid syntax

pirj:
I suggest extending the spec in accordance with our findings. I'm happy to help on this front. Please accept my apologies for the delay in review, it's a summer season over here.

spec/rubocop/cop/rspec/sort_metadata_spec.rb
cbliard: Then it would register an offense and would autocorrect it like this:

```ruby
RSpec.describe 'Something', :a, :z, some_hash do
```

This test was inspired from this `MetadataStyle` cop test:

https://github.com/rubocop/rubocop-rspec/blob/74d78371f4495099cf23bcf93b8af7ef711aec26/spec/rubocop/cop/rspec/metadata_style_spec.rb#L354-L364

It ensures that a variable is not moved at the start of the arguments list if this variable is at the end of the arguments list, because it could hold a hash.

Should I update the spec to swap `:a` and `:z` and expect the offense and correction as described above?

spec/rubocop/cop/rspec/sort_metadata_spec.rb
cbliard: Actually, this one is correct.

It's when there are more than 2 non-symbol arguments that the `ArgumentError: wrong number of arguments (given 4, expected 0..2)` is raised.

cbliard (commented):
I replied to your comments and I am open to amend this pull request with your suggestions.

Please accept my apologies for the delay as well. As you wrote, it's summer season.

cbliard:
Hi @pirj,

Is there anything I can do to help with this pull request?

pirj:
Sorry for the delayed response.
We had a fruitful discussion in the ticket about what cop should and shouldn’t do. I recall it as:
 - it should not attempt to handle code that will make RSpec fail anyway
 - it should handle realistic code
 - it should not autocorrect to code that would make RSpec to fail (was it what sparked this discussion initially?)
 - it should consider those subtle differences between examples and example groups when it comes to metadata arguments
 - it should ignore non-literal metadata arguments, and avoid risks automatically ordering them

Does this make sense? I’m writing this from my memory, and those criteria may be incomplete or incorrect.

Would it be a stretch to ask you to go over existing specs and make sure those criteria are met?
Last time I checked, I was under impression that a few cases in the very beginning of the changes to specs were off, and this repelled me from code reviewing further, as I thought it would require a lot commenting.

If you prefer, I can go through those specs, too.

And in any case - thank you for the dedication!
This is doable, let’s push the cop through the line where it can shine! 🌟

cbliard:
> We had a fruitful discussion in the ticket about what cop should and shouldn’t do. I recall it as:
> 
> * it should not attempt to handle code that will make RSpec fail anyway
> * it should handle realistic code
> * it should not autocorrect to code that would make RSpec to fail (was it what sparked this discussion initially?)
> * it should consider those subtle differences between examples and example groups when it comes to metadata arguments
> * it should ignore non-literal metadata arguments, and avoid risks automatically ordering them
> 
> Does this make sense? I’m writing this from my memory, and those criteria may be incomplete or incorrect.

Yes this makes sense. Not sure if that's exactly what we discussed but this makes sense.

Indeed "it should not autocorrect to code that would make RSpec to fail" is what sparked this discussion initially as rubocop-rspec autocorrects [this spec file](https://github.com/opf/openproject/blob/17dac48b2478a80fc91540c6033e10fc883a346b/spec/features/work_packages/progress_modal_pre_14_4_without_percent_complete_edition_spec.rb) and the resulting corrected file makes RSpec fail because the string parameter is moved from second to third place.

> Would it be a stretch to ask you to go over existing specs and make sure those criteria are met? Last time I checked, I was under impression that a few cases in the very beginning of the changes to specs were off, and this repelled me from code reviewing further, as I thought it would require a lot commenting.

Sure, so I should adapt these specs to ensure that it does not trigger at all, right?

I'll do that.

Thanks for the feedback.

cbliard:
@pirj I have adapted the spec and code as discussed. It is up for review.

It does not completely ignore code that _could_ make RSpec fail. Instead it limits sorting to the trailing args which are actual symbols. Any symbol arg before a string literal or a variable will be ignored.

spec/rubocop/cop/rspec/sort_metadata_spec.rb
```diff
@@ -23,6 +23,54 @@
...
+      RSpec.describe 'Something', :z, :a, 'string', :b, :c do
+      end
+    RUBY
+  end
+
+  it 'registers an offense when a symbol metadata not in alphabetical order ' \
+     'is before a variable argument being the last argument ' \
+     'as it could be a hash' do
+    expect_offense(<<~RUBY)
+      RSpec.describe 'Something', :z, :a, some_hash do
```
pirj, line 55: Now when I look at `some_hash`, did we mean a literal hash? How do we know `some_hash` is hash, and not a symbol?

I have no objections to such "chunky" reordering is in place.

Let's split this into two:

1. Example with a literal hash
```ruby
describe 'Something', :z, :a, { foo: 'bar' } do
```

2. Example with many chunks
```ruby
describe 'Chunks', :z, :a, var, :y, :b, another_var, :x, :c, last_var
```
The current implementation fixed this to:
```ruby
describe 'Chunks', :a, another_var, :b, :c, var, :x, :y, :z, last_var do
```
Which I don't entirely agree with. Variables should retain their positions. Only literal symbols should be rearranged.

Variables not remaining on their places result is a potentially breaking behaviour:
```ruby
RSpec.describe 'Chunks', second_docstring, :z, :a, var do
```
```ruby
RSpec.describe 'Chunks', :a, second_docstring, :z, var do 
```
:boom: "wrong number of arguments (given 3, expected 0..2)".

spec/rubocop/cop/rspec/sort_metadata_spec.rb
```diff
@@ -100,27 +148,27 @@
      'and the hash values are complex objects' do
     expect_offense(<<~RUBY)
       it 'Something', variable, 'B', :a, key => {}, foo: ->(x) { bar(x) }, Identifier.sample => true, baz: Snafu.new do
```
pirj, line 150: This works for "it". Confusing, but I recall that we've explored this diverged behaviour.

pirj (commented):
Two months later 🙈

spec/rubocop/cop/rspec/sort_metadata_spec.rb
```diff
@@ -23,6 +23,54 @@
     RUBY
   end
 
+  it 'does not register an offense for a symbol metadata before a non-symbol ' \
+     'argument' do
+    expect_no_offenses(<<~RUBY)
+      RSpec.describe 'Something', :z, :a, 'string' do
+      end
+
+      RSpec.describe 'Something', :z, :a, variable, foo: :bar do
```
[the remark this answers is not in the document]
cbliard: I feel neutral about removing them. They were present in the file before my modifications.
I will remove them.

spec/rubocop/cop/rspec/sort_metadata_spec.rb
cbliard: > Now when I look at some_hash, did we mean a literal hash? How do we know some_hash is hash, and not a symbol?

We don't know.

> 2. Example with many chunks
> 
> ```ruby
> describe 'Chunks', :z, :a, var, :y, :b, another_var, :x, :c, last_var
> ```
> 
> The current implementation fixed this to:
> 
> ```ruby
> describe 'Chunks', :a, another_var, :b, :c, var, :x, :y, :z, last_var do
> ```
> 
> Which I don't entirely agree with. Variables should retain their positions. Only literal symbols should be rearranged.
> 
> Variables not remaining on their places result is a potentially breaking behaviour:

Yep agreed.

With the implementation proposed in this pull request, it would fix this to:

```ruby
describe 'Chunks', :z, :a, var, :y, :b, another_var, :c, :x, last_var
```

only the last symbols part is reordered.

> Variables not remaining on their places result is a potentially breaking behaviour:
> 
> ```ruby
> RSpec.describe 'Chunks', second_docstring, :z, :a, var do
> ```
> 
> ```ruby
> RSpec.describe 'Chunks', :a, second_docstring, :z, var do 
> ```
> 
> 💥 "wrong number of arguments (given 3, expected 0..2)".

With the implementation proposed in this pull request, it would fix this to:

```ruby
RSpec.describe 'Chunks', second_docstring, :a, :z, var do 
```

That's what triggered me to open this PR.

spec/rubocop/cop/rspec/sort_metadata_spec.rb
cbliard: yes, that's a bit unsatisfying.

spec/rubocop/cop/rspec/sort_metadata_spec.rb
```diff
@@ -23,6 +23,54 @@
     RUBY
   end
 
+  it 'does not register an offense for a symbol metadata before a non-symbol ' \
+     'argument' do
+    expect_no_offenses(<<~RUBY)
+      RSpec.describe 'Something', :z, :a, 'string' do
+      end
+
```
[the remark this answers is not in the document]
cbliard: 👍 Removed as suggested.

spec/rubocop/cop/rspec/sort_metadata_spec.rb
```diff
@@ -23,6 +23,54 @@
...
+      RSpec.describe 'Something', :z, :a, variable, foo: :bar do
+      end
+    RUBY
+  end
+
+  it 'registers an offense only for trailing symbol metadata not in ' \
+     'alphabetical order' do
+    expect_offense(<<~RUBY)
+      RSpec.describe 'Something', :z, :a, 'string', :c, :b do
```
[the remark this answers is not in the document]
cbliard: 👍 Updated as suggested.

spec/rubocop/cop/rspec/sort_metadata_spec.rb
```diff
@@ -23,6 +23,54 @@
...
+                                                    ^^^^^^ Sort metadata alphabetically.
+      end
+    RUBY
+
+    expect_correction(<<~RUBY)
+      RSpec.describe 'Something', :z, :a, 'string', :b, :c do
```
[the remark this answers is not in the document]
cbliard: 👍 Updated as suggested.

cbliard:
Thanks @pirj for the review. I updated it following your suggestions and rebased it on master branch.

Is there anything else I should do?

spec/rubocop/cop/rspec/sort_metadata_spec.rb
pirj: > it would fix this to:

In a quick test (I've created a spec/a_spec.rb and ran `rubocop -a`), it fixed differently:
```diff
- describe 'Chunks', :z, :a, var, :y, :b, another_var, :c, :x, last_var do
+ describe 'Chunks', :a, another_var, :b, :c, var, :x, :y, :z, last_var do
```

Can you please check?

The rest seems to be 👍

cbliard: I'm sorry I did not get the same results as you. It left the file unchanged.

Here is how I tested it so you can test it on your end:

From my `code/rubocop-rspec` directory:

```shell
git switch sort_metadata_symbols_at_the_end_of_args_list
mkdir ../checking_cop
cd ../checking_cop
bundle init
bundle add rspec
bundle exec rspec --init

bundle add rubocop
echo 'gem "rubocop-rspec", path: "../rubocop-rspec"' >> Gemfile
bundle install
cat <<RUBOCOP_CONF > .rubocop.yml
require:
  - rubocop-rspec
AllCops:
  NewCops: enable
RUBOCOP_CONF

cat <<SPEC > spec/checking_cop_spec.rb
# frozen_string_literal: true

var = :var
another_var = :another_var
last_var = :last_var

describe 'Chunks', :z, :a, var, :y, :b, another_var, :c, :x, last_var do
  it 'is not an empty example group' do
    expect(true).to be_truthy
  end
end
SPEC

bundle exec rubocop --only RSpec/SortMetadata spec/checking_cop_spec.rb
```

And I get the output

```
Inspecting 1 file
.

```

I tested swapping the final parameters and running the cop again:

```shell
cat <<SPEC >> spec/checking_cop_spec.rb
describe 'Chunks 2', :z, :a, var, :y, :b, another_var, :x, :c, last_var do
  it 'is not an empty example group' do
    expect(true).to be_truthy
  end
end
SPEC

bundle exec rubocop --only RSpec/SortMetadata spec/checking_cop_spec.rb
```

got output:

```
Inspecting 1 file
C

Offenses:

spec/checking_cop_spec.rb:12:56: C: [Correctable] RSpec/SortMetadata: Sort metadata alphabetically.
describe 'Chunks 2', :z, :a, var, :y, :b, another_var, :x, :c, last_var do
                                                       ^^^^^^

1 file inspected, 1 offense detected, 1 offense autocorrectable
```

Can you please check?

pirj: Indeed, something stuck on my machine 🙈 . Please accept my apologies for that.

pirj (approved):
Thank you for the immense effort you've put into this!

Please accept my apologies for the daunting experience with those ages-long reviews. It is usually not like this in our repos 🙈

spec/rubocop/cop/rspec/sort_metadata_spec.rb
```diff
@@ -4,21 +4,80 @@
...
+
+    expect_correction(<<~RUBY)
+      describe 'Something', :z, :a, variable, :b, :c do
+      end
+    RUBY
+  end
+
+  it 'registers an offense when a symbol metadata not in alphabetical order ' \
+     'is before a variable argument being the last argument ' \
+     'as it could be a hash' do
```
bquorning, line 50: This seems a bit strange to me. If `some_hash` “could be a hash”, it could also be e.g. an array. And if we sort symbol metadata before an array variable, why don’t we _always_ sort symbol metadata before a variable?

bquorning (approved):

spec/rubocop/cop/rspec/sort_metadata_spec.rb
cbliard: Sure, no worries.

spec/rubocop/cop/rspec/sort_metadata_spec.rb
cbliard: That's a valid point. Indeed each group of symbols could be alphabetically sorted independently.

This PR was originally addressing #1946 and is now approved. If that makes sense, could your point be registered as a separate issue to be addressed later? Feel free to mention me and I can work on a PR for it.

cbliard:
Thank you @pirj for the review.

No worries for the time it took. I was not very available as well.

spec/rubocop/cop/rspec/sort_metadata_spec.rb
bquorning: My comment was not meant to block this PR; I’m find with merging this as-is.

cbliard:
Anything particular that can be done to make the "CI / Prism" check pass and get that PR merged?

ydah:
> Anything particular that can be done to make the "CI / Prism" check pass and get that PR merged?

@cbliard Could you try git rebase and import the latest master branch?

cbliard:
Thanks @ydah, that fix the Prism issue thanks to the example producing a syntax error you removed 2 weeks ago.

Then there was [this error](https://github.com/rubocop/rubocop-rspec/actions/runs/12692150402/job/35376820002?pr=1948) poping on "Edge RuboCop: internal_investigation" workflow:

```
lib/rubocop/rspec/description_extractor.rb:65:11: C: [Correctable] Style/MultipleComparison: Avoid comparing a variable with multiple items in a conditional, use Array#include? instead.
          yardoc.superclass.path == RSPEC_COP_CLASS_NAME || ...
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

I fixed it too and force pushed again.

pirj:
Thank you! 🙌🌟

PR #1950, opened 2024-08-16
Add RSpec-Cop unexpected requires

TheWudu:
We were facing issues with files required in spec-files only, which made the tests work, but at the staging or production environment the service was not able to start up due to missing requires.

This cop prevents adding requires to spec files, except for spec/rails_helper and except requires in spec/rails_helpers.


pirj:
Can you please run this cop against [real-world-rspec](https://github.com/pirj/real-world-rspec) repos to see the offences?

spec/rubocop/cop/rspec/unexpected_requires_spec.rb
```diff
@@ -0,0 +1,47 @@
+# frozen_string_literal: true
+
+RSpec.describe RuboCop::Cop::RSpec::UnexpectedRequires do
+  context 'when there is an require' do
+    it 'registers an offense' do
+      expect_offense(<<~RUBY)
+        require "base64"
```
pirj, line 7: I could find `require "rspec-parameterized"` and also "fast_spec_helper" in the same spec file. Both seem legitimate.

spec/rubocop/cop/rspec/unexpected_requires_spec.rb
```diff
@@ -0,0 +1,47 @@
...
+
+        describe "MyTest" do
+          do_something
+        end
+      RUBY
+    end
+
+    context 'with the file being spec_helper' do
+      it 'does not register an offense' do
+        expect_no_offenses(<<~RUBY, "spec_helper.rb")
```
bquorning, line 18: The file name should match one of the include patterns in https://github.com/rubocop/rubocop-rspec/blob/e02576f1a8c39d9dc5b3ca37fea4596f7e75f493/config/default.yml#L6-L8, or the example won’t reach your cop code at all.
```suggestion
        expect_no_offenses(<<~RUBY, "spec/spec_helper.rb")
```

spec/rubocop/cop/rspec/unexpected_requires_spec.rb
```diff
@@ -0,0 +1,47 @@
...
+
+          describe "MyTest" do
+          end
+        RUBY
+      end
+    end
+
+    context 'with the file being rails_helper' do
+      it 'does not register an offense' do
+        expect_no_offenses(<<~RUBY, 'rails_helper.rb')
```
bquorning, line 29: Same here:
```suggestion
        expect_no_offenses(<<~RUBY, 'spec/rails_helper.rb')
```

lib/rubocop/cop/rspec/unexpected_requires.rb
```diff
@@ -0,0 +1,60 @@
...
+
+        IGNORED_PATTERN = /(spec|rails)_helper/
+
+        def on_new_investigation
+          return if processed_source.path.match?(IGNORED_PATTERN)
+
+          unnecessary_requires.each do |token|
+            add_offense(token.pos, message: MSG)
+          end
+        end
```
bquorning, line 37: I think the whole implementation can be a lot simpler, e.g.
```rb
        def on_new_investigation
          @ignore_file = processed_source.path.match?(IGNORED_PATTERN)
        end

        def on_send(node)
          return if @ignore_file

          foo_node_matcher { add_offense(node) }
        end
```
(and then define the actual node matcher).

But we need to address @pirj’s question about legitimate requires first.

spec/rubocop/cop/rspec/unexpected_requires_spec.rb
TheWudu: shouldn't those requires then be inside spec_helper / rails_helper then?

TheWudu: I ran it against the real-world-specs, and if i did it right, there were no offences with this rubocop, but plenty of others ;-)

TheWudu: Is there an easy way to run rubocop against all repos in real-world-rspec repo?

lib/rubocop/cop/rspec/unexpected_requires.rb
TheWudu: require 'spec_helper' and require 'rails_helper' should be allowed, so a simple node matcher for require is not good enough.

Would then look like this. 
I think adding the exclusion of spec/rails helper to the regex would make it a) hard to write b) hard to read c) hard to understand ... 


```
        def_node_matcher :require_symbol?, <<~PATTERN
          (send #rspec? :require $_ ...)
        PATTERN
        
                def on_new_investigation
          @ignore_file = processed_source.path.match?(IGNORED_PATTERN)
        end

        def on_send(node)
          return if @ignore_file

          require_symbol?(node) do |match|
            next if match.value.match?(IGNORED_PATTERN)

            add_offense(node, message: MSG)
          end
        end
```

spec/rubocop/cop/rspec/unexpected_requires_spec.rb
pirj: Should a "fast_spec_helper" be required inside spec_helper? No, because supposedly the fast spec helper is a [subset of spec_helper](https://github.com/gitlabhq/gitlabhq/blob/ead25aaac3ab8e8883af13ec2e3b5308cf61ef70/spec/fast_spec_helper.rb#L4) that is purposed to run faster than spec_helper. I don't know the details, though.

Should a "rspec-parameterized" be required inside spec_helper? If it would, this spec alone will require spec_helper to load. And any spec requiring spec_helper would automatically require rspec-parameterized even if it's not used by that spec.

pirj: > an easy way to run rubocop against all repos in real-world-rspec repo?

I can't call it easy, but:
1. You clone it with submodules (this may require removing some of the dead ones or ones that changed the primary branch)
2. You recursively find .rubocop*.yml and remove it
3. You run the cop from rubocop-rspec with `rubocop --only <your-cop> ../real-world-rspec`

I kindly appreciate it if you could document this in https://docs.rubocop.org/rubocop/development.html#testing-your-cop-in-a-real-codebase

ydah:
Errors occur in the following cases:
```ruby
require Rails.root.join("db/migrate/20230910021213_update_category_setting_approval_values.rb")
```

Errors that occur are as follows:

```
bundle exec rubocop -d --only Rspec/UnexpectedRequires /ydah/real-world-rspec/discourse/spec/migrations/uppdate_category_setting_approval_values_spec.rb
For /ydah/rubocop-rspec: configuration from /ydah/rubocop-rspec/.rubocop.yml
configuration from /ydah/rubocop-rspec/config/default.yml
configuration from /ydah/rubocop-rspec/config/default.yml
Default configuration from /.rbenv/versions/3.3.0/lib/ruby/gems/3.3.0/gems/rubocop-1.65.1/config/default.yml
Use parallel by default.
Skipping parallel inspection: only a single file needs inspection
Inspecting 1 file
Scanning /ydah/real-world-rspec/discourse/spec/migrations/uppdate_category_setting_approval_values_spec.rb
For /ydah/real-world-rspec/discourse/spec/migrations: --only option: Rspec/UnexpectedRequires has the wrong namespace - replace it with RSpec/UnexpectedRequires
An error occurred while RSpec/UnexpectedRequires cop was inspecting /ydah/real-world-rspec/discourse/spec/migrations/uppdate_category_setting_approval_values_spec.rb:3:0.
undefined method `value' for an instance of RuboCop::AST::SendNode
/ydah/rubocop-rspec/lib/rubocop/cop/rspec/unexpected_requires.rb:44:in `block in on_send'
/ydah/rubocop-rspec/lib/rubocop/cop/rspec/unexpected_requires.rb:41:in `require_symbol?'
/ydah/rubocop-rspec/lib/rubocop/cop/rspec/unexpected_requires.rb:43:in `on_send'
/.rbenv/versions/3.3.0/lib/ruby/gems/3.3.0/gems/rubocop-1.65.1/lib/rubocop/cop/commissioner.rb:107:in `public_send'
/.rbenv/versions/3.3.0/lib/ruby/gems/3.3.0/gems/rubocop-1.65.1/lib/rubocop/cop/commissioner.rb:107:in `block (2 levels) in trigger_responding_cops'
/.rbenv/versions/3.3.0/lib/ruby/gems/3.3.0/gems/rubocop-1.65.1/lib/rubocop/cop/commissioner.rb:171:in `with_cop_error_handling'
/.rbenv/versions/3.3.0/lib/ruby/gems/3.3.0/gems/rubocop-1.65.1/lib/rubocop/cop/commissioner.rb:106:in `block in trigger_responding_cops'
/.rbenv/versions/3.3.0/lib/ruby/gems/3.3.0/gems/rubocop-1.65.1/lib/rubocop/cop/commissioner.rb:105:in `each'
/.rbenv/versions/3.3.0/lib/ruby/gems/3.3.0/gems/rubocop-1.65.1/lib/rubocop/cop/commissioner.rb:105:in `trigger_responding_cops'
/.rbenv/versions/3.3.0/lib/ruby/gems/3.3.0/gems/rubocop-1.65.1/lib/rubocop/cop/commissioner.rb:69:in `on_send'
/.rbenv/versions/3.3.0/lib/ruby/gems/3.3.0/gems/rubocop-ast-1.32.0/lib/rubocop/ast/traversal.rb:137:in `block in on_dstr'
/.rbenv/versions/3.3.0/lib/ruby/gems/3.3.0/gems/rubocop-ast-1.32.0/lib/rubocop/ast/traversal.rb:137:in `each'
/.rbenv/versions/3.3.0/lib/ruby/gems/3.3.0/gems/rubocop-ast-1.32.0/lib/rubocop/ast/traversal.rb:137:in `on_dstr'
/.rbenv/versions/3.3.0/lib/ruby/gems/3.3.0/gems/rubocop-1.65.1/lib/rubocop/cop/commissioner.rb:71:in `on_begin'
/.rbenv/versions/3.3.0/lib/ruby/gems/3.3.0/gems/rubocop-ast-1.32.0/lib/rubocop/ast/traversal.rb:20:in `walk'
/.rbenv/versions/3.3.0/lib/ruby/gems/3.3.0/gems/rubocop-1.65.1/lib/rubocop/cop/commissioner.rb:87:in `investigate'
/.rbenv/versions/3.3.0/lib/ruby/gems/3.3.0/gems/rubocop-1.65.1/lib/rubocop/cop/team.rb:164:in `investigate_partial'
/.rbenv/versions/3.3.0/lib/ruby/gems/3.3.0/gems/rubocop-1.65.1/lib/rubocop/cop/team.rb:102:in `investigate'
/.rbenv/versions/3.3.0/lib/ruby/gems/3.3.0/gems/rubocop-1.65.1/lib/rubocop/runner.rb:349:in `block in inspect_file'
/.rbenv/versions/3.3.0/lib/ruby/gems/3.3.0/gems/rubocop-1.65.1/lib/rubocop/runner.rb:348:in `each'
/.rbenv/versions/3.3.0/lib/ruby/gems/3.3.0/gems/rubocop-1.65.1/lib/rubocop/runner.rb:348:in `flat_map'
/.rbenv/versions/3.3.0/lib/ruby/gems/3.3.0/gems/rubocop-1.65.1/lib/rubocop/runner.rb:348:in `inspect_file'
/.rbenv/versions/3.3.0/lib/ruby/gems/3.3.0/gems/rubocop-1.65.1/lib/rubocop/runner.rb:291:in `block in do_inspection_loop'
/.rbenv/versions/3.3.0/lib/ruby/gems/3.3.0/gems/rubocop-1.65.1/lib/rubocop/runner.rb:325:in `block in iterate_until_no_changes'
<internal:kernel>:187:in `loop'
/.rbenv/versions/3.3.0/lib/ruby/gems/3.3.0/gems/rubocop-1.65.1/lib/rubocop/runner.rb:318:in `iterate_until_no_changes'
/.rbenv/versions/3.3.0/lib/ruby/gems/3.3.0/gems/rubocop-1.65.1/lib/rubocop/runner.rb:287:in `do_inspection_loop'
/.rbenv/versions/3.3.0/lib/ruby/gems/3.3.0/gems/rubocop-1.65.1/lib/rubocop/runner.rb:168:in `block in file_offenses'
/.rbenv/versions/3.3.0/lib/ruby/gems/3.3.0/gems/rubocop-1.65.1/lib/rubocop/runner.rb:193:in `file_offense_cache'
/.rbenv/versions/3.3.0/lib/ruby/gems/3.3.0/gems/rubocop-1.65.1/lib/rubocop/runner.rb:167:in `file_offenses'
/.rbenv/versions/3.3.0/lib/ruby/gems/3.3.0/gems/rubocop-1.65.1/lib/rubocop/runner.rb:158:in `process_file'
/.rbenv/versions/3.3.0/lib/ruby/gems/3.3.0/gems/rubocop-1.65.1/lib/rubocop/runner.rb:139:in `block in each_inspected_file'
/.rbenv/versions/3.3.0/lib/ruby/gems/3.3.0/gems/rubocop-1.65.1/lib/rubocop/runner.rb:138:in `each'
/.rbenv/versions/3.3.0/lib/ruby/gems/3.3.0/gems/rubocop-1.65.1/lib/rubocop/runner.rb:138:in `reduce'
/.rbenv/versions/3.3.0/lib/ruby/gems/3.3.0/gems/rubocop-1.65.1/lib/rubocop/runner.rb:138:in `each_inspected_file'
/.rbenv/versions/3.3.0/lib/ruby/gems/3.3.0/gems/rubocop-1.65.1/lib/rubocop/runner.rb:124:in `inspect_files'
/.rbenv/versions/3.3.0/lib/ruby/gems/3.3.0/gems/rubocop-1.65.1/lib/rubocop/runner.rb:77:in `run'
/.rbenv/versions/3.3.0/lib/ruby/gems/3.3.0/gems/rubocop-1.65.1/lib/rubocop/cli/command/execute_runner.rb:26:in `block in execute_runner'
/.rbenv/versions/3.3.0/lib/ruby/gems/3.3.0/gems/rubocop-1.65.1/lib/rubocop/cli/command/execute_runner.rb:52:in `with_redirect'
/.rbenv/versions/3.3.0/lib/ruby/gems/3.3.0/gems/rubocop-1.65.1/lib/rubocop/cli/command/execute_runner.rb:25:in `execute_runner'
/.rbenv/versions/3.3.0/lib/ruby/gems/3.3.0/gems/rubocop-1.65.1/lib/rubocop/cli/command/execute_runner.rb:17:in `run'
/.rbenv/versions/3.3.0/lib/ruby/gems/3.3.0/gems/rubocop-1.65.1/lib/rubocop/cli/command.rb:11:in `run'
/.rbenv/versions/3.3.0/lib/ruby/gems/3.3.0/gems/rubocop-1.65.1/lib/rubocop/cli/environment.rb:18:in `run'
/.rbenv/versions/3.3.0/lib/ruby/gems/3.3.0/gems/rubocop-1.65.1/lib/rubocop/cli.rb:122:in `run_command'
/.rbenv/versions/3.3.0/lib/ruby/gems/3.3.0/gems/rubocop-1.65.1/lib/rubocop/cli.rb:129:in `execute_runners'
/.rbenv/versions/3.3.0/lib/ruby/gems/3.3.0/gems/rubocop-1.65.1/lib/rubocop/cli.rb:51:in `block in run'
/.rbenv/versions/3.3.0/lib/ruby/gems/3.3.0/gems/rubocop-1.65.1/lib/rubocop/cli.rb:81:in `profile_if_needed'
/.rbenv/versions/3.3.0/lib/ruby/gems/3.3.0/gems/rubocop-1.65.1/lib/rubocop/cli.rb:43:in `run'
/.rbenv/versions/3.3.0/lib/ruby/gems/3.3.0/gems/rubocop-1.65.1/exe/rubocop:19:in `block in <top (required)>'
/.rbenv/versions/3.3.0/lib/ruby/3.3.0/benchmark.rb:313:in `realtime'
/.rbenv/versions/3.3.0/lib/ruby/gems/3.3.0/gems/rubocop-1.65.1/exe/rubocop:19:in `<top (required)>'
/.rbenv/versions/3.3.0/lib/ruby/gems/3.3.0/bin/rubocop:25:in `load'
/.rbenv/versions/3.3.0/lib/ruby/gems/3.3.0/bin/rubocop:25:in `<top (required)>'
/.local/share/gem/ruby/3.3.0/gems/bundler-2.5.7/lib/bundler/cli/exec.rb:58:in `load'
/.local/share/gem/ruby/3.3.0/gems/bundler-2.5.7/lib/bundler/cli/exec.rb:58:in `kernel_load'
/.local/share/gem/ruby/3.3.0/gems/bundler-2.5.7/lib/bundler/cli/exec.rb:23:in `run'
/.local/share/gem/ruby/3.3.0/gems/bundler-2.5.7/lib/bundler/cli.rb:451:in `exec'
/.local/share/gem/ruby/3.3.0/gems/bundler-2.5.7/lib/bundler/vendor/thor/lib/thor/command.rb:28:in `run'
/.local/share/gem/ruby/3.3.0/gems/bundler-2.5.7/lib/bundler/vendor/thor/lib/thor/invocation.rb:127:in `invoke_command'
/.local/share/gem/ruby/3.3.0/gems/bundler-2.5.7/lib/bundler/vendor/thor/lib/thor.rb:527:in `dispatch'
/.local/share/gem/ruby/3.3.0/gems/bundler-2.5.7/lib/bundler/cli.rb:34:in `dispatch'
/.local/share/gem/ruby/3.3.0/gems/bundler-2.5.7/lib/bundler/vendor/thor/lib/thor/base.rb:584:in `start'
/.local/share/gem/ruby/3.3.0/gems/bundler-2.5.7/lib/bundler/cli.rb:28:in `start'
/.local/share/gem/ruby/3.3.0/gems/bundler-2.5.7/exe/bundle:28:in `block in <top (required)>'
/.local/share/gem/ruby/3.3.0/gems/bundler-2.5.7/lib/bundler/friendly_errors.rb:117:in `with_friendly_errors'
/.local/share/gem/ruby/3.3.0/gems/bundler-2.5.7/exe/bundle:20:in `<top (required)>'
/.rbenv/versions/3.3.0/bin/bundle:25:in `load'
/.rbenv/versions/3.3.0/bin/bundle:25:in `<main>'
.

1 error occurred:
An error occurred while RSpec/UnexpectedRequires cop was inspecting /ydah/real-world-rspec/discourse/spec/migrations/uppdate_category_setting_approval_values_spec.rb:3:0.
```

TheWudu:
> I ran it through real-world-rspec:
> 
> Errors occur in the following cases:
> 
> ```ruby
> require Rails.root.join("db/migrate/20230910021213_update_category_setting_approval_values.rb")
> ```
> 
> Errors that occur are as follows:
> 
> ```
> bundle exec rubocop -d --only Rspec/UnexpectedRequires /ydah/real-world-rspec/discourse/spec/migrations/uppdate_category_setting_approval_values_spec.rb
> ...
> 
> 1 file inspected, no offenses detected
> 
> 1 error occurred:
> An error occurred while RSpec/UnexpectedRequires cop was inspecting /ydah/real-world-rspec/discourse/spec/migrations/uppdate_category_setting_approval_values_spec.rb:3:0.
> ```

I fixed this. The reason is that only "string" nodes have the value method, and this require Rails.root is a node (require) with a "nested" node "Rails.root"...

E.g.
```
(send nil :require
  (str "base64"))
```

vs 

```
(send nil :require
  (send
    (send
      (const nil :Rails) :root) :join
    (str "db/migrate/20230910021213_update_category_setting_approval_values.rb")))
```

spec/rubocop/cop/rspec/unexpected_requires_spec.rb
TheWudu: I ran it against the real-world-rspecs:

Prepare real-world-rspecs folder:
```
git clone git@github.com:pirj/real-world-rspec.git
cd real-world-spec
git submodule update --init --remote --checkout --single-branch --depth 1
find ./ -type f -name .rubocop.yml -exec rm -rf {} \;
```

```
cd rubocop-rspec
bundle exec rubocop --only RSpec/UnexpectedRequires ../real-world-rspec/
```

It finds many occurences ... but mostly: 
* pry
* helper
* test/unit
* active_support/*
* simplecov
* webmock
* rspec/support/*
* support/*
* tmpdir

and others. Which might be "Ok" to require in specs, but still i would more expect them to be required in spec_helper than in the individual files. 

So this does not really change my mind, that it should be avoided to add requires to spec-files. And as you can disable rubocops anyways when necessary, it still should be fine. 

I also saw this: 

```
require File.dirname(__FILE__) + "/../../spec_helper"
```

which could still be excluded in the checks .. but imo i would prefer "require 'spec_helper'" there anyways, not using a relative file-path .. but that is a different topic.

pirj: [This](https://github.com/rspec/rspec-core/blob/2ba0f10fe68948e5ee9addd569b4694f0245aa71/lib/rspec/core/project_initializer/spec/spec_helper.rb#L11) is interesting.

To me, however, it’s unclear if we should create a separate webmock_spec_helper.rb with a single line that just requires webmock.

> production … was not able to start up due to missing requires

And I’m uncertain how imposing such a structure will solve the original issue.

Also, I don’t quite understand the original issue. Isn’t Bundler supposed to load all requirements? Shouldn’t own code be eagerly autoloaded in production (assuming a Rails app)?

Please don’t get me wrong, I still recall a very-well covered feature that relied on a gem (json schema validator) that was only present in the `:test` and `:development` bundler groups, making it pass specs perfectly, but fail in production.
I can’t tell off the top of my head how `require: false` would affect this.

There are also non-Rails apps with, supposedly, explicit require structure. And non-eagerly autoloaded paths in Rails apps, too (support lib code for rake tasks; internal rubocop cops, …).

I have no certainty that just banning requires from specs would solve all this. It wouldn’t solve my past issue with the json schema validator, because it was `required: false` in the Gemfile.

pirj: `require File.dirname(__FILE__) + "/../../spec_helper"` is some legacy, and can be written with a `require_relative`, `__dir__`, using the `/` on the `Pathname`, etc.

But we have to keep in mind that you can run specs from a directory inside the spec directory, not necessarily the project root.

And if there’s a `gems/` directory where you keep code that is to be eventually extracted, with their own `spec/` directory?

TheWudu: We do not use rails anymore, as it was just to heavy back then. We require a separate require-structure, and it is not about bundler-requires, it is really about e.g. requiring the use-case class from within the worker-class or the api or whatever. 

And we sometimes had the issue that we started test-driven, implemented the spec, and obviously required the use-case-class there to be able to test it, and forgot to add it to the worker/api/.. then, but after deploying the api/worker failed due to missing require / unknown class. 

It did not happen often, but a few times already. So we thought: a easy rubocop would help us. And finally, after quite some time, we found the time to invest a bit of time here. 

I totally get to require stuff only where it is needed, as suggested in the comment you linked above, but imo this should only relate to "stuff needed for the test, and only the test", which tbh. we rarely have (and if, it is added to the spec_helper).

> I have no certainty that just banning requires from specs would solve all this. It wouldn’t solve my past issue with the json schema validator, because it was required: false in the Gemfile.

If it is not loaded via bundler, and not in your "config" somehow, the require in the spec-file would have covered it is not loaded on application start, so it might have solved your issue? Or did i get anything wrong?

pirj: > only the test", which tbh. we rarely have

By “we”, do you mean the whole community, or your project?
Because above you’ve provided a “mostly” list that mostly has test-related dependencies.

I kind of agree that it’s a cleaner approach to use helpers. But I can’t recall if I’ve seen more than a handful.
Putting all into one or two helpers is not always the best option. Everyone hates when specs load the whole thing for half a minute. And every small dependency contributes to this.

At this point, I’m not convinced that requires from specs (except for helpers) are inherently bad, and should be banned.

> If it is not loaded via bundler

It was loaded by bundler, but only in test and development. In production, with `bundle --without development,test` there was no way it would be loaded.
I can’t think of anything that could have solved it.

TheWudu: "We" is the projects I work on with others. One of them e.g. has this rspec run:

> Finished in 24.33 seconds (files took 7.11 seconds to load)
> 2989 examples, 0 failures

In this project we basically have everything either loaded in usual app code, or if only needed for specs, in spec_helper. Can't say it is slow tbh. ;-)

But nothing to discuss now. 

> It was loaded by bundler, but only in test and development. In production, with bundle --without development,test there was no way it would be loaded.
> I can’t think of anything that could have solved it.

True that, this would not have helped in that case. 

> At this point, I’m not convinced that requires from specs (except for helpers) are inherently bad, and should be banned.

I will discuss it a bit further with my team, and come back to you.

TheWudu: Ok , i talked to my team and we aligned on closing this PR, we might still add it for us as we still think it is valuable, but i get why it will not yet added to the public repo. 

Thanks for your feedback.

PR #1952, opened 2024-08-26
Fix: Fix false negative for `RSpec/Dialect` when specified Capybara-specific methods

### LINKED ISSUE #1951: RSpec/Dialect causes a false negative when using below setting mentioned in the documentation
RSpec/Dialect causes a false negative when using below setting mentioned in the documentation.

```yaml
RSpec/Dialect:
  PreferredMethods:
    background: :before
    scenario:   :it
    xscenario:  :xit
    given:      :let
    given!:     :let!
    feature:    :describe
```

cf. https://docs.rubocop.org/rubocop-rspec/cops_rspec.html#rspecdialect

## Expected behavior

The following code should not be passed:

```ruby
background do # 👮‍♂️ C: [Correctable] RSpec/Dialect: Prefer before over background.
  # some setup codes
end
```

## Actual behavior

The following code is passed:

```ruby
background do
  # some setup codes
end
```

In contrast, the following codes doesn't pass.

```ruby
RSpec.feature do
  # some codes
end
```

```ruby
scenario do
  # some codes
end
```

```ruby
xscenario do
  # some codes
end
```

I look like `rspec_method?` returns nil when `node.method_name` is `background`.

https://github.com/rubocop/rubocop-rspec/blob/e02576f1a8c39d9dc5b3ca37fea4596f7e75f493/lib/rubocop/cop/rspec/dialect.rb#L67-L80

## RuboCop version

```
$ bundle exec rubocop -V
1.65.1 (using Parser 3.3.4.2, rubocop-ast 1.32.0, running on ruby 3.2.3) +server [arm64-darwin21]
  - rubocop-capybara 2.21.0
  - rubocop-factory_bot 2.26.1
  - rubocop-performance 1.21.1
  - rubocop-rails 2.25.1
  - rubocop-rspec 3.0.4
  - rubocop-rspec_rails 2.30.0
```

sanfrecce-osaka:
fixed https://github.com/rubocop/rubocop-rspec/issues/1951

`RuboCop::Cop::RSpec::Dialect#rspec_method?` detects if node's second child is an RSpec block using `RuboCop::RSpec::Language::ALL.all`. However, `RuboCop::Cop::RSpec::Dialect#rspec_method?` returned nil because the following Capybara-specific methods were not set in config/default.yml.

- given
- given!
- background

I remove the line that checks if it’s a `rspec_method?` and add a `inside_example_group?` check.

cf. https://github.com/rubocop/rubocop-rspec/issues/1951#issuecomment-2296916136

Due to the above, it wouldn't filter out are calls with an explicit receiver where the receiver is not RSpec, eg `foo.describe`. so the following examples were removed.

- allows calling methods named xxx in examples

- [x] Added tests.
- [ ] Updated documentation.

### DIFF
--- a/spec/rubocop/cop/rspec/dialect_spec.rb
+++ b/spec/rubocop/cop/rspec/dialect_spec.rb
@@ -1,59 +1,186 @@
 # frozen_string_literal: true
 
 RSpec.describe RuboCop::Cop::RSpec::Dialect do
-  let(:cop_config) do
-    {
-      'PreferredMethods' => {
-        'context' => 'describe'
-      }
-    }
-  end
-
-  it 'allows describe blocks' do
-    expect_no_offenses(<<~RUBY)
-      describe 'display name presence' do
+  context 'with preferred methods' do
+    context 'when `describe` is preferred to `context`' do
+      let(:cop_config) do
+        {
+          'PreferredMethods' => {
+            'context' => 'describe'
+          }
+        }
       end
-    RUBY
-  end
 
-  it 'allows calling methods named context in examples' do
-    expect_no_offenses(<<~RUBY)
-      it 'tests common context invocations' do
-        expect(request.context).to be_empty?
+      it 'allows describe blocks' do
+        expect_no_offenses(<<~RUBY)
+          RSpec.describe 'context' do
+            describe 'display name presence' do
+            end
+          end
+        RUBY
       end
-    RUBY
-  end
 
-  it 'registers an offense for context blocks' do
-    expect_offense(<<~RUBY)
-      context 'display name presence' do
-      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ Prefer `describe` over `context`.
-      end
-    RUBY
+      it 'registers an offense for context blocks' do
+        expect_offense(<<~RUBY)
+          RSpec.describe 'context' do
+            context 'display name presence' do
+            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ Prefer `describe` over `context`.
+            end
+          end
+        RUBY
 
-    expect_correction(<<~RUBY)
-      describe 'display name presence' do
+        expect_correction(<<~RUBY)
+          RSpec.describe 'context' do
+            describe 'display name presence' do
+            end
+          end
+        RUBY
       end
-    RUBY
-  end
+    end
 
-  it 'registers an offense for RSpec.context blocks' do
-    expect_offense(<<~RUBY)
-      RSpec.context 'context' do
-      ^^^^^^^^^^^^^^^^^^^^^^^ Prefer `describe` over `context`.
-        it 'tests common context invocations' do
-          expect(request.context).to be_empty?
-        end
+    context 'when `describe` is preferred to `feature`' do
+      let(:cop_config) do
+        {
+          'PreferredMethods' => {
+            'feature' => 'describe'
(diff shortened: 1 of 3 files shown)


sanfrecce-osaka:
Should I also change the following document as it no longer checks for `RuboCop::RSpec::Language::ALL.all?`

https://github.com/rubocop/rubocop-rspec/blob/fa5da43962b53bffd3565087731e6a98666aac4b/lib/rubocop/cop/rspec/dialect.rb#L8-L23

pirj:
> change the following document

Maybe. I find it hard to understand what the “based” means in “dialect can be based on the following RSpec methods”, though.

Can you think of a simpler description?

spec/rubocop/cop/rspec/dialect_spec.rb
```diff
@@ -1,59 +1,186 @@
...
       end
-    RUBY
-  end
 
-  it 'registers an offense for RSpec.context blocks' do
-    expect_offense(<<~RUBY)
-      RSpec.context 'context' do
-      ^^^^^^^^^^^^^^^^^^^^^^^ Prefer `describe` over `context`.
-        it 'tests common context invocations' do
-          expect(request.context).to be_empty?
```
pirj, line 45: Those ones were here to ensure that `context` with an explicit receiver is not detected. Can you keep it in some way?

ydah:
@sanfrecce-osaka ping ;)

spec/rubocop/cop/rspec/dialect_spec.rb
sanfrecce-osaka: Sorry for the late reply 🙏 

> Those ones were here to ensure that `context` with an explicit receiver is not detected. Can you keep it in some way?

In fa5da43 the check using `rspec_method?` has been removed, so explicit receivers are also detected. To avoid detection, I can't think of any ideas other than to do the following.

```diff
          def on_send(node)
            return unless inside_example_group?(node)
+           return unless rspec?(node.children.first)
            return unless preferred_methods[node.method_name]
```

Do you have any good ideas?

sanfrecce-osaka:
https://github.com/rubocop/rubocop-rspec/pull/1952#issuecomment-2311194088



> > change the following document
> 
> Maybe. I find it hard to understand what the “based” means in “dialect can be based on the following RSpec methods”, though.
> 
> Can you think of a simpler description?

How about the following?

```diff
-       # Enforces custom RSpec dialects.
+       # Enforces custom RSpec dialects based on your config.
        #
-       # A dialect can be based on the following RSpec methods:
+       # `Dialects` are the following examples:
        #
        # - describe, context, feature, example_group
        # - xdescribe, xcontext, xfeature
        # - fdescribe, fcontext, ffeature
        # - shared_examples, shared_examples_for, shared_context
        # - it, specify, example, scenario, its
        # - fit, fspecify, fexample, fscenario, focus
        # - xit, xspecify, xexample, xscenario, skip
        # - pending
        # - prepend_before, before, append_before,
        # - around
        # - prepend_after, after, append_after
        # - let, let!
        # - subject, subject!
        # - expect, is_expected, expect_any_instance_of
        #
        # By default all of the RSpec methods and aliases are allowed. By setting
        # a config like:
        #
        #   RSpec/Dialect:
        #     PreferredMethods:
        #       context: describe
```

pirj:
I fail to grasp the essence of the change. Please feel free to re-open with a clear description.

PR #1953, opened 2024-08-26
Improve offense message for `RSpec/IndexedLet`

Earlopain:
Ran into this offense and thought:
> Where exactly does my name contain "index"?

Of course I got it eventually but I had to look up docs online. I believe this message format is more clear.

- [ ] Added tests.
- [ ] Updated documentation.

### DIFF
--- a/lib/rubocop/cop/rspec/indexed_let.rb
+++ b/lib/rubocop/cop/rspec/indexed_let.rb
@@ -48,8 +48,8 @@ module RuboCop
         include AllowedIdentifiers
         include AllowedPattern
 
-        MSG = 'This `let` statement uses index in its name. Please give it ' \
-              'a meaningful name.'
+        MSG = 'This `let` statement uses `%<index>s` in its name. ' \
+              'Please give it a meaningful name.'
 
         # @!method let_name(node)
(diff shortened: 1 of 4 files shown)


bquorning (approved):

pirj (approved):

lib/rubocop/cop/rspec/indexed_let.rb
```diff
@@ -48,8 +48,8 @@ class IndexedLet < Base
         include AllowedIdentifiers
         include AllowedPattern
 
-        MSG = 'This `let` statement uses index in its name. Please give it ' \
-              'a meaningful name.'
+        MSG = 'This `let` statement uses `%<index>s` in its name. ' \
```
pirj, line 51: Should we keep the work “index” eg “uses `3` index in its name”?

Earlopain: `index` doesn't seem like the right word to use. I know the cop is called `IndexedLet` but I wonder if that is a misnomer, `NumberedLet` (or similar) seems more accurate to what it does.

Personally I would leave it as is but overall not that important to me if you want it changed. Let me know.

ydah (approved):
Thank you!

PR #1955, opened 2024-09-04
CI: Use arm64 runners

bquorning:
CI: Use arm64 runners

We can now run on the arm based CI runners: https://github.blog/changelog/2025-01-16-linux-arm64-hosted-runners-now-available-for-free-in-public-repositories-public-preview/

Our CI generally runs pretty fast, so it's hard to see any real difference in runtime yet. However, the two relatively slow JRuby jobs finished 10-15% faster on this PR than they did on the latest CI run on the master branch ([master](https://github.com/rubocop/rubocop-rspec/actions/runs/12763800983/job/35574680857) vs [this PR](https://github.com/rubocop/rubocop-rspec/actions/runs/12857027580/job/35844409916)).

Links:

- GitHub’s documentation of [Standard GitHub-hosted runners for public repositories](https://docs.github.com/en/actions/using-github-hosted-runners/using-github-hosted-runners/about-github-hosted-runners#standard-github-hosted-runners-for-public-repositories)
- The Readme for the [Ubuntu 22.04 by Arm Limited](https://github.com/actions/partner-runner-images/blob/main/images/arm-ubuntu-22-image.md) runner image.

### DIFF
--- a/.github/workflows/codespell.yml
+++ b/.github/workflows/codespell.yml
@@ -7,7 +7,7 @@ concurrency:
 jobs:
   codespell:
     name: CodeSpell
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-24.04-arm
     steps:
       - uses: actions/checkout@v4
       - name: CodeSpell
--- a/.github/workflows/linting.yml
+++ b/.github/workflows/linting.yml
@@ -7,7 +7,7 @@ concurrency:
 jobs:
   yamllint:
     name: Yamllint
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-24.04-arm
     steps:
       - uses: actions/checkout@v4
       - name: Yamllint
@@ -20,7 +20,7 @@ jobs:
           GITHUB_ACCESS_TOKEN: ${{ secrets.GITHUB_TOKEN }}
   mdformat:
     name: Mdformat
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-24.04-arm
     steps:
       - uses: actions/checkout@v4
       - name: Mdformat
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -12,7 +12,7 @@ concurrency:
 
 jobs:
   confirm_config_and_documentation:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-24.04-arm
     name: Confirm config and documentation
     steps:
       - uses: actions/checkout@v4
@@ -23,7 +23,7 @@ jobs:
       - run: bundle exec rake confirm_config documentation_syntax_check confirm_documentation
 
   main:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-24.04-arm
     strategy:
       matrix:
         ruby:
@@ -48,7 +48,7 @@ jobs:
       - run: NO_COVERAGE=true bundle exec rake ${{ matrix.task }}
 
   coverage:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-24.04-arm
     name: "Test coverage"
     steps:
       - uses: actions/checkout@v4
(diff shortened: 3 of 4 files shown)


bquorning:
I missed this part from [the announcement](https://github.blog/changelog/2024-09-03-github-actions-arm64-linux-and-windows-runners-are-now-generally-available/):

> Arm64 runners are available to customers on our Team and Enterprise Cloud plans.

😞

pirj:
Maybe we should just remove JRuby jobs?

I'd let GitHub choose what they want to run our CI on, if arm64 is more effective, we don't mind (if they even need us to agree).

bquorning:
How do we let GitHub choose?

.github/workflows/codespell.yml
```diff
@@ -7,7 +7,7 @@ concurrency:
 jobs:
   codespell:
     name: CodeSpell
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-22.04-arm
```
pirj, line 10: Minor: Why not ubuntu-24.04-arm?

The difference between sensible packages (e.g. openssl) is negligible between 22.04 and 24.04

pirj (approved):
https://github.com/actions/runner-images/blob/af0df299674bc781d6a4f891d9cea62b438e1332/images/ubuntu/Ubuntu2404-Readme.md?plain=1#L8 doesn't explicitly tell it's an X86-64, except for a few (Java, Chromium) packages pointing to that. I don't see why GitHub wouldn't want to silently switch those for ARM runners sometimes in the future, considering cost savings. I believe such proactive users as us will be in the minority.

I don't mind the change, as we're unlikely to hit any weird CI issues. Also a good part of us devs are on ARM those days thanks to MacBooks.
Also, just for the curiosity's sake, it's quite interesting to how good CI run on ARM, and bring this experience to our primary jobs to save on ever-hungry CI.

.github/workflows/codespell.yml
bquorning: Mainly because we are currently running CI against Ubuntu 22.

bquorning:
I was hoping that GitHub would provide something like the `ubuntu-latest` runner, e.g. `ubuntu-arm-latest`, that points to different OS versions so we wouldn’t have to keep updating it manually. And who knows, maybe it will be added? This is still early days; this is still a “public preview” feature that is subject to change.

.github/workflows/codespell.yml
pirj: Isn’t the `-latest` same as 24?

bquorning: According to https://docs.github.com/en/actions/using-github-hosted-runners/using-github-hosted-runners/about-github-hosted-runners#standard-github-hosted-runners-for-public-repositories, there is an `ubuntu-latest`, `windows-latest`, and `macos-latest`, but no `-latest` are documented for the arm runners.

pirj: I mean, before we switched to arm runners, we used -latest, which is ubuntu 24.
But now, we switched to arm runners and ubuntu 22. Even though ubuntu 24 is available for arm, too. Why?

bquorning: Huh, that’s interesting. Our CI runs (e.g. [this](https://github.com/rubocop/rubocop-rspec/actions/runs/12870100230)) has an annotation at the bottom saying

> ubuntu-latest pipelines will use ubuntu-24.04 soon. For more details, see https://github.com/actions/runner-images/issues/10636

I understood that as “CI will run ubuntu 24, but _not yet_”. But looking closer at any CI run, I of course now see `Image: ubuntu-24.04`. So yes of course, this PR should target Ubuntu 24 as well.

pirj: Ha! Understood. I am sorry I didn’t when you first explained.
No objections then!

Issue #1136, opened 2021-03-23
double/instance_double argument type


bquorning:
I'm seeing double/instance_double/class_double/object_double/spy etc. being called with sometimes a string, sometimes a symbol (perhaps because the rspec-mocks documentation explicitly allows both), and of course sometimes a constant referring to a class or module.

I think mixing string and symbol arguments is a bit of a mess, so I would propose making a new cop that enforces either one or the other – with strings being the default enforced style. Why? Because `"MyClass"` looks better than `:MyClass` in those cases where you pass in a module name as a string/symbol.

pirj:
[`instance_double` accepts a class](https://github.com/rspec/rspec-mocks/blob/97c972be57f2c060a4a7fb8a3c5700a5ede693f0/lib/rspec/mocks/example_methods.rb#L38), and so does `class_double`/`instance_spy`. I see no good reasons not to use this. Except to avoid excessive class loading when `instance_double` and the class is referenced from globally-defined shared examples, and those shared examples are eagerly loaded from spec helper.

pirj:
I can also add that changing from a class literal to a string/symbol arguably makes the cop unsafe unless [`verify_doubled_constant_names`](https://github.com/rspec/rspec-mocks/blob/3-11-maintenance/lib/rspec/mocks/configuration.rb#L115) is turned on. It is [off by default in RSpec Mocks 3.11](https://github.com/rspec/rspec-mocks/blob/e4a7ececb851562f5f4ce64762e7976da8fe56f2/lib/rspec/mocks/configuration.rb#L8) and the upcoming [4.0, too](https://github.com/rspec/rspec-mocks/blob/4-0-dev/lib/rspec/mocks/configuration.rb#L8) for the reasons I struggle to remember. Related PR [turning on `verify_partial_doubles` by default](https://github.com/rspec/rspec-mocks/pull/1409).

corsonknowles:
We have a very simple Rubocop rule that autocorrects from string to Class for `instance_double`. It does not affect symbol declarations. Is there interest in us open-sourcing this? I'd be happy to push it up.

pirj:
That will be very kind of you, @corsonknowles ! A pr is welcome.

corsonknowles:
Happy to help! Here's a draft PR:

* https://github.com/rubocop/rubocop-rspec/pull/1957

We've run this on over 50,000 and had it in place for a few years. 

[...]

PR #1957, opened 2024-09-17
Add RSpec/StringAsInstanceDoubleConstant

corsonknowles:
Begins Addressing: #1136

Adds a cop which can autocorrect from String declarations for `instance_double` to Class declarations. Passing a variable -- which could hold a String or a Class -- is not affected by this cop.

Autocorrect is not marked as safe, because the correction will cause instantiation of the Class, which depending on test setup may or may not have been loaded yet. In addition, instantiating the Class before stubbing an `instance_double` means that RSpec will treat it as a verifying double. A string reference to a Class that is not loaded is treated as a normal non-verifying double.

We have been running this as safe autocorrect internally for years, and have not met with any issues as we use Zeitwerk and eager load once before running specs. In pushing this upstream, I don't want to assume that all test suites will have autoloading in place.

- [x] Added tests.
- [x] Updated documentation.


corsonknowles:
Getting

```
* generated /Users/david.corsonknowles/Documents/GitHub/rubocop-rspec/docs/modules/ROOT/pages//cops_rspec.adoc
rake aborted!
Please run `bundle exec rake generate_cops_documentation` and add docs/ to the commit.
/Users/david.corsonknowles/Documents/GitHub/rubocop-rspec/Rakefile:79:in `block in <top (required)>'
/Users/david.corsonknowles/.rbenv/versions/3.3.0/bin/ruby_executable_hooks:22:in `eval'
/Users/david.corsonknowles/.rbenv/versions/3.3.0/bin/ruby_executable_hooks:22:in `<main>'
Tasks: TOP => default => confirm_documentation
(See full trace by running task with --trace)
```

However, have run `rake generate_cops_documentation`

.rubocop.yml
```diff
@@ -101,34 +101,40 @@ RSpec:
...
 RSpec/MultipleExpectations:
   Max: 2
 
-# `expect_offense` does not use Kernel#format or String#%
-Style/FormatStringToken:
-  Exclude:
-    - spec/rubocop/**/*.rb
-
-Style/RequireOrder:
+RSpec/NoStringifiedInstanceDoubleConstant:
```
bquorning, line 119: Could you move this to the bottom of the file, please?

CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## Master (Unreleased)
 
+## 3.0.6 (2024-09-17)
+
```
```suggestion
```

lib/rubocop/cop/rspec/no_stringified_instance_double_constant.rb
```diff
@@ -0,0 +1,42 @@
...
+      #
+      class NoStringifiedInstanceDoubleConstant < Base
+        extend AutoCorrector
+
+        MSG = 'Do not use a string as `instance_double` constant.'
+        RESTRICT_ON_SEND = %i[instance_double].freeze
+
+        # @!method stringified_instance_double_const?(node)
+        def_node_matcher :stringified_instance_double_const?, <<~PATTERN
+          (send nil? :instance_double $...)
```
bquorning, line 23: The node pattern can also do the `str_type?` check, _and_ return just the string node instead of an array:
```suggestion
          (send nil? :instance_double $str ...)
```

spec/rubocop/cop/rspec/no_stringified_instance_double_constant_spec.rb
```diff
@@ -0,0 +1,37 @@
...
+               :config do
+  context 'when using a class for instance_double' do
+    let(:source) do
+      <<~RUBY
+        instance_double(Foo, failure_message, stubs)
+      RUBY
+    end
+
+    it { expect_no_offenses source }
+  end
```
bquorning, line 13: I much prefer to have specs written out, especially in simple cases like this.
```suggestion
  it 'finds no offense when using a class for instance_double' do
    expect_no_offenses(<<~RUBY)
      instance_double(Foo, failure_message, stubs)
    RUBY
  end
```

.rubocop.yml
```diff
@@ -101,34 +101,40 @@ RSpec:
...
 RSpec/ExampleLength:
   CountAsOne:
     - heredoc
   Max: 11
 
-RSpec/DescribeClass:
-  Exclude:
-    - spec/project/**/*.rb
+RSpec/ExpectInLet:
+  Enabled: true
```
bquorning, line 114: I’m not sure where this entry comes from, ExpectInLet is already enabled by default.yml 😅 

I’ll open a separate pull request with the other changes to .rubocop.yml.

bquorning:
> However, have run `rake generate_cops_documentation`

Did you also “add docs/ to the commit”? When I run the rake task I see a diff in the docs folder.

pirj:
There are some wip rubocop cop naming guidelines, and they insist on naming cops so that they highlight the problem. This should be “StringifiedInstabceDoubleConstant”

.rubocop.yml
bquorning: Done: #1958

spec/rubocop/cop/rspec/no_stringified_instance_double_constant_spec.rb
corsonknowles: Thanks!

lib/rubocop/cop/rspec/no_stringified_instance_double_constant.rb
corsonknowles: Fixed and refactored the cop to match. Thanks for the improvement!

corsonknowles:
> Did you also “add docs/ to the commit”? When I run the rake task I see a diff in the docs folder.

Looks like my rake problems were partly self-inflicted, I tried to add my own documentation while I was following the contribution guidelines.

technicalpickles (commented):
I am biased since it was extracted from our app, but in favor of this 👍🏻 

Interestingly, `instance_double` _does_ support a string as the name, at least [according to its rdoc](https://github.com/rspec/rspec-mocks/blob/134b1acf5d357b09df1f4516392d50c82634f109/lib/rspec/mocks/example_methods.rb#L38-L49). I am looking at some of the implementation ([`instance_double`](https://github.com/rspec/rspec-mocks/blob/134b1acf5d357b09df1f4516392d50c82634f109/lib/rspec/mocks/example_methods.rb#L57-L58) and [`ObjectReference.for(doubled_class)`](https://github.com/rspec/rspec-mocks/blob/134b1acf5d357b09df1f4516392d50c82634f109/lib/rspec/mocks/object_reference.rb#L7-L28) but I couldn't tell you what happens for classes vs strings.

bquorning (commented):
Two questions:

1. Could / should we make the same rule for `class_double` too?
2. Shouldn’t we also detect and replace symbols? As I understand [the source](https://github.com/rspec/rspec-mocks/blob/main/lib/rspec/mocks/example_methods.rb#L38-L59), calling `instance_double` with a symbol won’t even work.

corsonknowles:
I don't think we need to extend the cop to cover symbols because RSpec will raise like this:
```
     ArgumentError:
       Module or String expected, got :Task
```
I have updated the PR description to be clearer about that.

corsonknowles:
I think enforcing this for `class_double` would be another Cop, right?

spec/rubocop/cop/rspec/string_as_instance_double_constant_spec.rb
```diff
@@ -0,0 +1,33 @@
...
+    end
+  end
+
+  context 'when using a symbol for instance_double' do
+    it 'does not register an offense' do
+      expect_no_offenses(<<~RUBY)
+        instance_double(:Foo, failure_message, stubs)
+      RUBY
+    end
+  end
```
bquorning, line 19: Maybe we can remove this test, if RSpec would raise an ArgumentError for it anyway.

bquorning:
> I think enforcing this for `class_double` would be another Cop, right?

@pirj What’s your take on this?

config/default.yml
```diff
@@ -929,6 +929,13 @@ RSpec/SpecFilePathSuffix:
     - "**/spec/**/*"
   Reference: https://www.rubydoc.info/gems/rubocop-rspec/RuboCop/Cop/RSpec/SpecFilePathSuffix
 
+RSpec/StringAsInstanceDoubleConstant:
+  Description: Do not use a string as `instance_double` constant.
+  Enabled: pending
+  Safe: false
```
pirj, line 935: What are those unsafe cases?

spec/rubocop/cop/rspec/string_as_instance_double_constant_spec.rb
```diff
@@ -0,0 +1,33 @@
+# frozen_string_literal: true
+
+RSpec.describe RuboCop::Cop::RSpec::StringAsInstanceDoubleConstant,
+               :config do
+  context 'when using a class for instance_double' do
+    it 'does not register an offense' do
+      expect_no_offenses(<<~RUBY)
+        instance_double(Foo, failure_message, stubs)
```
pirj, line 8:
```suggestion
        instance_double(Shape, area: 12)
```
The second optional positional arg is a "name", but I don't see this frequently used in practice.
The last one, a hash, is what is stubbed.

Let's give here and in the cop doc examples that are more similar to real usage.

lib/rubocop/cop/rspec/string_as_instance_double_constant.rb
```diff
@@ -0,0 +1,40 @@
+# frozen_string_literal: true
+
+module RuboCop
+  module Cop
+    module RSpec
+      # Do not use a string as `instance_double` constant.
+      #
+      # @example
+      #   # bad
+      #   instance_double('Foo', failure_message, stubs)
```
pirj, line 10:
```suggestion
      #   instance_double('User', name: 'John')
```

config/default.yml
pirj: In real-world-rspec, I could find:
```
                                      ^^^^^^^^
/Users/pirj/source/real-world-rspec/engine/spec/views/locomotive/current_site/edit_spec.rb:44:39: C: [Correctable] RSpec/StringAsInstanceDoubleConstant: Do not use a string as instance_double constant.
      let(:_policy) { instance_double('policy', create?: true, point?: true, update_advanced?: true, destroy?: false) }
```

and
```
/Users/pirj/source/real-world-rspec/chef/spec/unit/provider/package/openbsd_spec.rb:56:31: C: [Correctable] RSpec/StringAsInstanceDoubleConstant: Do not use a string as instance_double constant.
              instance_double("shellout", stdout: "#{name}-#{version}\n")
                              ^^^^^^^^^^
```

But this is probably due to verified doubles being turned off.
I'd say this should become a regular `double`, or become a proper class constant like in:
```
/Users/pirj/source/real-world-rspec/chef/chef-config/spec/unit/config_spec.rb:930:33: C: [Correctable] RSpec/StringAsInstanceDoubleConstant: Do not use a string as instance_double constant.
          cmd = instance_double("Mixlib::ShellOut", exitstatus: 0, stdout: locales, error!: nil)
                                ^^^^^^^^^^^^^^^^^^
```

spec/rubocop/cop/rspec/string_as_instance_double_constant_spec.rb
```diff
@@ -0,0 +1,33 @@
...
+        instance_double('Foo', failure_message, stubs)
+                        ^^^^^ Do not use a string as `instance_double` constant.
+      RUBY
+
+      expect_correction <<~RUBY
+        instance_double(Foo, failure_message, stubs)
+      RUBY
+    end
+  end
+end
```
pirj, line 33: Please add a case when the first argument is a variable/method `instance_double(type_undetectable_in_static_analysis)`.

pirj (commented):
No errors whatsoever during a run against real-world-rspec. 499 offences look legit at a glance.

Thank you!

A few minor tweaks, and good to go.

config/default.yml
corsonknowles: Yes, I think we can't mark it as safe for autocorrect, because to do so would require too many assumptions to be met about the test setup. 

The other concern mentioned in the original repo issue here is that using the class to initialize requires actually loading the class (and again I'm not making any assumptions about Zeitwerk etc for the end user's test suite here). 

Although we can't be sure the class is loaded to safely autocorrect, much of the benefit of this Cop is enforcing that the `instance_double` really does work as intended (because this way the class is loaded in order to initialize)

From the docs:
>Constructs a test double against a specific class. If the given class name has been loaded, only instance methods defined on the class are allowed to be stubbed. In all other ways it behaves like a [double](https://rubydoc.info/github/rspec/rspec-mocks/double).
Source: https://rubydoc.info/github/rspec/rspec-mocks/RSpec%2FMocks%2FExampleMethods:instance_double

spec/rubocop/cop/rspec/string_as_instance_double_constant_spec.rb
corsonknowles: Thanks, done!

config/default.yml
corsonknowles: Do you want me to add a sentence like "This is not marked as safe for autocorrect, since the correction only works when the class is loaded" ?

corsonknowles: I'm hoping that I was on point with this, and I added it. Happy to amend as needed!

pirj: What are the unsafe cases?

pirj: I do get one.

> using the class to initialize requires actually loading the class

This makes sense. We certainly don’t want to enforce loading classes.

From the doc:
> If the given class name has been loaded, only instance methods defined on the class are allowed to be stubbed. In all other ways it behaves like a [double](https://rubydoc.info/github/rspec/rspec-mocks/double).

So the two legit usages from my perspective are:
 - instance_double with a class
 - double with a string when the intention is to NOT load a class

I’m struggling to quickly come up with an idea how to provide clear separation between those two for our ysers.

pirj: There’s a dedicated @satefy tag, it will be parsed and added to docs. Please keep the description short.

I apologise for brief response- I’m from my phone.

corsonknowles: Perfect. Added the @safety tag! I think this is a lot more clearly presented now, with just enough info folks won't need to hunt down the original issue or even the RSpec docs.

corsonknowles: Yes, I think the key thing is that this Cop will enforce that an `instance_double` really is a validating double. String instantiation "falls back" to a simple non-verifying double when the class has not been loaded. So the behavior for strings is ambiguous in that it may result in a validating double or a non-validating double. And if there are other setup issues, that could actually be a dynamic problem, and lead to flaky tests.

corsonknowles:
As an aside, while following the test coverage results, I noticed what appears to be a dead code method:
`top_level_group?`

https://github.com/search?q=repo%3Arubocop%2Frubocop-rspec%20top_level_group%3F&type=code

All specs pass without it. 

The only other uncovered line is `raise ::NotImplementedError`

I'd be happy to raise another PR after this, if its useful. I don't know if you prefer to handle all maintenance PRs internally.

bquorning:
> As an aside, while following the test coverage results, I noticed what appears to be a dead code method:
> `top_level_group?`

Yeah, it seems its last use was removed in 62763702395587ae6ea75d0df8abbcfa78d3aae2 / #977.

> I'd be happy to raise another PR after this, if its useful. I don't know if you prefer to handle all maintenance PRs internally.

You’re more than welcome to open a PR to remove that method 👍🏼

config/default.yml
pirj: Nice summary!

This makes me think is if the cop itself is safe, it's just the autocorrection to a literal class that is not. The `SafeAutocorrect: false` setting should do the job of highlighting the issue, and letting users decide whether they want to switch to a regular double and keep a string to prevent loading, or to benefit from what instance_double offers, and switch to the class. @bquorning WDYT?

corsonknowles:
Are we ready to merge here?

bquorning:
The original issue #1957 mentions all of `double`/`instance_double`/`class_double`/`object_double`/`spy` etc. Since the cop name includes the method name `instance_double`, I think before we merge we should talk a bit more about if this cop should also police some of the other methods, of we want to have separate cops for each.

corsonknowles:
Thanks @pirj --- so the problem with adding `class_double` is that it is a complete re-write. 
We need to: 
1. Rename the cop
2. Rewrite the description
3. Redo the matcher
4. Double the specs
5. Re-run the new rule on live codebases

Then I have an additional downstream problem where that rule doesn't exist in my codebase, so instead of merging this and then pulling the upstream, I need to go work through all `class_double`s at the same time. 

I guess what I'm saying is it's much more preferable for me to decouple these things. 

In comparison to all of the above, it's relatively trivial for anyone to duplicate this new Cop and adapt it to work for `class_double` -- it's just a find and replace exercise, very modular.

I can't think of any performance reasons to combine them either, with restrict on send in place. A combined rule would need some kind of OR matcher for both method names, or to leverage something to insert the method name in the pattern, if something like that exists.

pirj:
Makes total sense to me. Thanks for the explanation.

bquorning:
Yep, that makes sense. Thank you 👍🏼

bquorning (approved):

bquorning:
Hi @corsonknowles, I think you rebased the branch yesterday, but it looks like you have got a few extra/unintended changes in. Would you mind taking a look before we merge?

corsonknowles:
Thanks @bquorning, amended!

bquorning (approved):
Thank you for the contribution, and for your patience with our nitpicking 😅

docs/modules/ROOT/pages/cops_rspec.adoc
```diff
@@ -5493,6 +5493,41 @@ spec/models/user.rb       # shared_examples_for 'foo'
...
+
+=== Examples
+
+[source,ruby]
+----
+# bad
+instance_double('User', name: 'John')
+
+# good
+instance_double(User, name: 'John')
```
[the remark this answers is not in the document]
pirj: Hi!
Nice catch, we’ve missed this completely.

The new cop is essentially the old one, but with no option to select the style. And this is certainly the direction we choose to enforce.

Your input is valuable, why did you decide to turn the cop off?

In a nutshell, we’ve discovered that an instance_double with a string argument does not verify, and behaves identically to a regular `double`. It makes sense to make the usage explicit, and to use a `double` for abstract doubles, and `instant_double` for specific ones, and verify them.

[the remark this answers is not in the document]
lucthev: > why did you decide to turn the cop off

We didn't totally turn off `VerifiedDoubleReference`, just ignored a number of existing offences via `.rubocop_todo.yml` — it's a large, old codebase, but Rubocop was only added semi-recently. It's definitely a useful cop for the reasons you've outlined!

That being said, if this new cop is wholly redundant, then I'll likely turn it off 😅. Would it make sense to deprecate one or the other? For what it's worth, `VerifiedDoubleReference` [appears to support](https://github.com/rubocop/rubocop-rspec/blob/a95f2f0d7e785176cbcfea9e70c7fd2fa3d80dfd/lib/rubocop/cop/rspec/verified_double_reference.rb#L50) e.g. `class_double` as well, which I see was a concern raised in this PR.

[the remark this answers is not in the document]
pirj: Thanks for sharing.

Yes, it would make sense to deprecate one of the cops, and I’d remove the EnforcedStyle: string option from the old one if it is less intrusive.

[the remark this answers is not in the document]
corsonknowles: Ah, thank you for immediately reporting this Luc!

I'm happy to deprecate the new cop and focus on updating the original cop. 

So the action items are 
1. deprecate StringAsInstanceDoubleConstant -- is there a way to make it an alias of the other cop?
2. remove String style from `VerifiedDoubleReference` as it is not consistently verifying. Is there a way to deprecate this or do we simply remove it?

I'll also take a look if there's anything in the new implementation worth porting over to [VerifiedDoubleReference](https://docs.rubocop.org/rubocop-rspec/cops_rspec.html#rspecverifieddoublereference

I can probably get started on this on Monday. If anyone wants to pitch in or get started earlier, you are by all means welcome to.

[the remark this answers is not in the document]
corsonknowles: I guess we could have it inherit from `VerifiedDoubleReference ` and add the deprecation message, if that makes sense?

Issue #1954, opened 2024-08-28
RSpec/UnspecifiedException triggers for `not_to` within a `to` block


G-Rath:
Given the following:

```
expect do
  expect { described_class.new(report).populate! }.not_to raise_error
end.to change(ReportTicket, :count).by(3)
```

Before #1940 this was fine, but now it's being flagged as an error - if I replace the outer `to` with `not_to`, it passes again.

pirj:
Walking up the tree seems incorrect [here](https://github.com/rubocop/rubocop-rspec/blob/74d78371f4495099cf23bcf93b8af7ef711aec26/lib/rubocop/cop/rspec/unspecified_exception.rb#L65).
I recall making a very similar note in another issue [here](https://github.com/rubocop/rubocop-rspec/issues/1949#issuecomment-2285070419), and the fix is the same.

@rubocop/rubocop-rspec why do we have this cop? Doesn’t it fall into a non-goal category considering [this RSpec built-in warning](https://github.com/rspec/rspec-expectations/blob/2bf42aaa1621202f0fb71b371dbf6c3fec1f4291/lib/rspec/matchers/built_in/raise_error.rb#L189)?

Should we fix the cop, or deprecate it?

bquorning:
It looks like that feature [was added](https://github.com/rspec/rspec-expectations/commit/e70959a13a5a382dcfec2a810f75b6c3450d955a) in RSpec 3.3, over 9 years ago. I’d say we should deprecate this cop.

ydah:
It might still have been worthwhile to have this cop if it had supported autocorrection, but since it does not support autocorrection, this cop should be deprecated.

G-Rath:
hmm so on second thought, I'm less in favor of deprecating this cop now because it seems RSpec doesn't fail when printing this warning - it might be possible to configure RSpec to error, but overall it does seem slightly useful to be able to lint this in parallel with running your whole test suite.

It also sounds like there's someone working [to address this](https://github.com/rubocop/rubocop-rspec/issues/1949#issuecomment-2341284685) - I assume that while the team have expressed a wish to deprecate the rule, they still wouldn't outright reject a pull request fixing the underlying issue so I might just for now add an inline disable in my project for now and see how this unfolds in the community

PR #1960, opened 2024-09-21
RSpec/UnspecifiedException: break on non-send blocks to avoid false positives on functions named `raise_exception`

### LINKED ISSUE #1949: RSpec/UnspecifiedException appears to trigger if the subject function is named `raise_exception`
It seems like the `RSpec/UnspecifiedException` cop should not be triggered on the following code, but it is:

```ruby
require "spec_helper"

class Foo

  def raise_exception
    raise StandardError, "huh"
  end

end

describe Foo do
  let(:subject) { described_class.new }

  describe "#raise_exception" do
    it "throws a StandardError" do
      expect { subject.raise_exception }.to raise_error(StandardError)
    end
  end
end

``` 

In particular, the cop highlights `subject.raise_exception` as violating the cop, which doesn't make sense - the cop should be looking it how we _handle_ the exception, yeah?

aarestad:
Fixes a false positive in the `UnspecifiedException` cop that is triggered when a function under test is literally called `raise_exception` or `raise_error`.

Resolves #1949.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Master (Unreleased)
 
+- Fix false-positive for `RSpec/UnspecifiedException` when a method is literally named `raise_exception`. ([@aarestad])
+
 ## 3.0.5 (2024-09-07)
 
 - Fix false-negative and error for `RSpec/MetadataStyle` when non-literal args are used in metadata in `EnforceStyle: hash`. ([@cbliard])
@@ -899,6 +901,7 @@ Compatibility release so users can upgrade RuboCop to 0.51.0. No new features.
 
 <!-- Contributors (alphabetically) -->
 
+[@aarestad]: https://github.com/aarestad
 [@abrom]: https://github.com/abrom
 [@ahukkanen]: https://github.com/ahukkanen
 [@akiomik]: https://github.com/akiomik
--- a/lib/rubocop/cop/rspec/unspecified_exception.rb
+++ b/lib/rubocop/cop/rspec/unspecified_exception.rb
@@ -62,7 +62,10 @@ module RuboCop
         end
 
         def find_expect_to(node)
-          node.each_ancestor(:send).find do |ancestor|
+          node.each_ancestor.find do |ancestor|
+            break if ancestor.block_type?
+            next unless ancestor.send_type?
+
             expect_to?(ancestor)
           end
         end
--- a/spec/rubocop/cop/rspec/unspecified_exception_spec.rb
+++ b/spec/rubocop/cop/rspec/unspecified_exception_spec.rb
@@ -85,6 +85,30 @@ RSpec.describe RuboCop::Cop::RSpec::UnspecifiedException do
         }.to raise_error(my_exception)
       RUBY
     end
+
+    it 'allows a subject function to be named raise_error' do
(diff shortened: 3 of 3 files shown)


.gitignore
```diff
@@ -24,4 +24,7 @@ coverage
 # vscode generated
 .vscode
 
+# IntelliJ generated
+.idea
```
bquorning, line 28: This line belongs in your `~/.gitignore` and not here.

bquorning (approved):

lib/rubocop/cop/rspec/unspecified_exception.rb
```diff
@@ -62,7 +62,11 @@ def empty_exception_matcher?(node)
         end
 
         def find_expect_to(node)
-          node.each_ancestor(:send).find do |ancestor|
+          node.each_ancestor.find do |ancestor|
+            break if ancestor.block_type?
+
```
bquorning, line 67: Sorry, I’m nitpicking now :-)
```suggestion
```

bquorning:
Can I ask you for a review @pirj ?

CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 ## Master (Unreleased)
 
+- Fix false-positive for `RSpec/UnspecifiedException` when a function is literally named `raise_exception`. ([@aarestad])
```
pirj, line 5: Function -> method

is this the only problem this pr fixes?

pirj (approved):
Does it fix #1954, too? Can you please add a spec and a changelog entry, or remove this mention from the pr description?

I’d say the cop should be disabled and deprecated, but if you’re fine with fizing it, I have no objections, too.

Thank you!

CHANGELOG.md
aarestad: Yeah, I'll remove the reference to the other issue since I don't really know anything about it other than that comment.

aarestad, line 5:
```suggestion
- Fix false-positive for `RSpec/UnspecifiedException` when a method is literally named `raise_exception`. ([@aarestad])
```

G-Rath: just confirming once this is landed I'm happy to do a follow-up to get my issue closed by e.g. adding a spec, so I support you pretending it doesn't exist 🙂

spec/rubocop/cop/rspec/unspecified_exception_spec.rb
```diff
@@ -198,5 +198,29 @@
...
+      expect_no_offenses(<<~RUBY)
+        def raise_error
+          raise StandardError
+        end
+
+        expect {
+          raise_error
+        }.to raise_exception(StandardError)
+      RUBY
+    end
```
bquorning, line 224: I just noticed that this spec file has a context named `'with raise_error matcher'`. So this example should probably be moved to around line 88 or so.

bquorning (approved):
One final nitpick from me 😬

spec/rubocop/cop/rspec/unspecified_exception_spec.rb
bquorning: Oh, also please squash the commits (as I imagine we are getting close to merging now).

bquorning (approved):

PR #1968, opened 2024-10-13
Obsolete `StringAsInstanceDoubleConstant` in favor of `VerifiedDoubleReference` for constants only

corsonknowles:
This PR obsoletes our newly introduced `StringAsInstanceDoubleConstant` in favor of `VerifiedDoubleReference`.

This changes `VerifiedDoubleReference` to no longer support string style, as string references are not verifying when the class has not yet been loaded.

From the `RSpec` docs:
>No checking will happen if the underlying object or class is not defined, but when run with it present (either as a full spec run or by explicitly preloading collaborators) a failure will be triggered if an invalid method is being stubbed or a method is called with an invalid number of arguments.

* https://rspec.info/features/3-12/rspec-mocks/verifying-doubles/

This PR essentially reverts nearly all of 
* https://github.com/rubocop/rubocop-rspec/pull/1957 

This uses the Obsoletion feature to point leading edge branch users to a revised `VerifiedDoubleReference` instead. 

- [x] Added tests.
- [x] Updated documentation.


corsonknowles:
@pirj @bquorning Does this look like the right approach to you as outlined in the comments here?
*  https://github.com/rubocop/rubocop-rspec/pull/1957

@lucthev also tagging you as a courtesy since you identified the duplication here.

ydah (commented):
Since this cop has been released once, it might be better not to delete or rename it until the next major version update, but to encourage migration with warning and changelogs, soft deprecation, and proceed with deletion and renaming in the next major version update.

corsonknowles:
I rebased and prepped this so it can be ready for the Monthly release

bquorning:
The code looks good, but I am not sure if removing `string` from `SupportedStyles` in `RSpec/VerifiedDoubleReference` is enough of a breaking change that we should wait until v4? @pirj I think (/hope) you know the major/minor release rules better than I.

pirj:
I’d check if we can avoid errors by using the obsoletion config.
If removal of the option errors out, it should go to v4.
But even if the above would error out, we can soft-deprecate the option by ignoring it, or printing a warning if it’s detected to use the deprecated style. It’s a hack, but worth it to push this out earlier.

pirj:
So what happens if you add the following to .rubocop.yml?
```
RSpec/StringAsInstanceDoubleConstant:
  Enabled: false

RSpec/VerifiedDoubleReference:
  EnforcedStyle: string
```
and run rubocop?

lib/rubocop/cop/rspec/verified_double_reference.rb
```diff
@@ -5,14 +5,14 @@ module Cop
...
       # Checks for consistent verified double reference style.
       #
-      # Only investigates references that are one of the supported styles.
-      #
       # @see https://rspec.info/features/3-12/rspec-mocks/verifying-doubles
       #
-      # This cop can be configured in your configuration using the
-      # `EnforcedStyle` option and supports `--auto-gen-config`.
+      # @safety
+      #   This cop is unsafe because the correction requires loading the class.
```
pirj, line 11: Nice!

bquorning:
> So what happens if you add the following to .rubocop.yml?
> 
> ```
> RSpec/StringAsInstanceDoubleConstant:
>   Enabled: false
> 
> RSpec/VerifiedDoubleReference:
>   EnforcedStyle: string
> ```
> 
> and run rubocop?

I tried that and first got this error:

```
Error: The `RSpec/StringAsInstanceDoubleConstant` cop has been renamed to `RSpec/VerifiedDoubleReference`.
(obsolete configuration found in .rubocop.yml, please update it)
```

I removed the obsolete configuration and ran rubocop again, this time getting a successful output, but got a warning:
```
Warning: RSpec/VerifiedDoubleReference does not support EnforcedStyle parameter.

Supported parameters are:

  - Enabled
```

bquorning (approved):

pirj:
If we used the [changed_parameters](https://github.com/rubocop/rubocop/blob/566df8551b48ab98556ec42c005d66f1a23fad68/config/obsoletion.yml#L123), this would also result in an error, not a warning, right?

pirj (approved):
LGTM without the `renamed` in obsoletion conf.
If changed_parameters results in a warning, please add it.

Thank you!

bquorning:
Both `changed_parameters` and `changed_enforced_styles` can be configures with a `severity`, so we could do e.g.

```yml
changed_enforced_styles:
  - cops: RSpec/VerifiedDoubleReference
    parameters: EnforcedStyle
    value: string
    reason: String references are not verifying unless the class is loaded.
    severity: warning
```

Of course, users will now get _two_ warnings instead of one:

```
Warning: obsolete `EnforcedStyle: string` (for `RSpec/VerifiedDoubleReference`) found in .rubocop.yml
String references are not verifying unless the class is loaded.
Warning: RSpec/VerifiedDoubleReference does not support EnforcedStyle parameter.

Supported parameters are:

  - Enabled
```

corsonknowles:
Thanks @bquorning -- what about the `EnforcedStyle` parameter no longer being needed, regardless of value?

bquorning:
Ah, that is true. Perhaps the `changed_parameters` obsoletion config is better then.

config/obsoletion.yml
```diff
@@ -6,6 +6,13 @@
 
 # Cop parameters that have been changed
 # Can be treated as a warning instead of a failure with `severity: warning`
+changed_enforced_styles:
+  - cops: RSpec/VerifiedDoubleReference
+    parameters: EnforcedStyle
+    value: string
+    reason: String references are not verifying unless the class is loaded.
+    severity: warning
+
```
bquorning, line 15: Instead of these lines, I’d add this to `changed_parameters`:
```yaml
  - cops: RSpec/VerifiedDoubleReference
    parameters: EnforcedStyle
    reason: String references are not verifying unless the class is loaded.
```

config/obsoletion.yml
```diff
@@ -28,3 +35,6 @@ extracted:
   RSpec/Rails/*: rubocop-rspec_rails
   RSpec/FactoryBot/*: rubocop-factory_bot
   RSpec/Capybara/*: rubocop-capybara
+
+renamed:
+  RSpec/StringAsInstanceDoubleConstant: RSpec/VerifiedDoubleReference
```
bquorning, line 40: I’d remove this and add the following to `removed`:
```yaml
  RSpec/StringAsInstanceDoubleConstant:
    reason: Please use `RSpec/VerifiedDoubleReference` instead
```

bquorning (commented):
Two final changes to config/obsoletion.yml, rebase on main, and fix position of the changelog entry. Then I think it’s ready to merge.

Are you up for it, @corsonknowles? 🚀

corsonknowles:
Happy to fix, @bquorning -- thanks for the guidance. How does it look now?

bquorning (approved):
@pirj Do you have time for a review as well?

pirj:
Side observation: this wasn't trivial to check out locally, or is it just me.
```
git checkout corsonknowles-master
git pull git@github.com:corsonknowles/rubocop-rspec.git master
```
I wouldn't be certain how to push amendments to the PR.

pirj (approved):
What I did:
1. In an existing project that uses rubocop-rspec, added the following to `.rubocop.yml`:
```
RSpec/StringAsInstanceDoubleConstant:
  Enabled: true
```
2. Changed it to use rubocop-rspec from a path:
3. Ran `rubocop`, and ✅ 

Thank you, @corsonknowles! 🙌

bquorning:
I did the same steps as you @pirj, but also had
```
RSpec/VerifiedDoubleReference:
  Enabled: true
  EnforcedStyle: constant
```
And the deprecation warnings work perfectly :+1:

CHANGELOG.md
```diff
@@ -14,6 +14,7 @@
 - Fix `RSpec/VoidExpect` to only operate inside an example block. ([@corsonknowles])
 - Change `RSpec/ContextWording` cop to always report an offense when both `Prefixes` and `AllowedPatterns` are empty. ([@ydah])
 - Add support for `and` and `or` compound matchers to `RSpec/ChangeByZero` cop. ([@ydah])
+- Replace `RSpec/StringAsInstanceDoubleConstant` with `RSpecVerifiedDoubleReference` configured to only support constant class references. ([@corsonknowles])
```
bquorning, line 17:
```suggestion
- Replace `RSpec/StringAsInstanceDoubleConstant` with `RSpec/VerifiedDoubleReference` configured to only support constant class references. ([@corsonknowles])
```

bquorning: I’ll fix this after merging.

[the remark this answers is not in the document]
bquorning: I’ll fix this after merging.

corsonknowles: Thank you!

PR #1970, opened 2024-10-13
Complete line coverage for Metadata mixin

corsonknowles:
At this point, there are only 2 uncovered lines in the whole repo. This PR addresses one of them, in the Metadata mixin:
* it adds a final spec to cover an intentionally unimplemented method.

The spec  for `on_metadata` may be trivial, but it accomplishes the goal of having 100% line coverage, which means all subsequent committers will know immediately if any of their lines are uncovered. 

This will also unblock this repo from updating the existing test coverage CI check to maintain 100% coverage forever moving forward:

Covering the other missing line is split to this PR:
* https://github.com/rubocop/rubocop-rspec/pull/1976  

- [x] Added tests.
- [x] Updated documentation.

### DIFF
--- a/spec/rubocop/cop/rspec/mixin/metadata_spec.rb
+++ b/spec/rubocop/cop/rspec/mixin/metadata_spec.rb
@@ -1,0 +1,17 @@
+# frozen_string_literal: true
+
+RSpec.describe RuboCop::Cop::RSpec::Metadata do
+  describe '#on_metadata' do
+    subject(:on_metadata) do
+      stub_class.new.on_metadata(:symbol, {})
+    end
+
+    let(:stub_class) do
+      Class.new do
+        include RuboCop::Cop::RSpec::Metadata
+      end
+    end
+
+    it { expect { on_metadata }.to raise_error(NotImplementedError) }
+  end
+end


lib/rubocop/cop/rspec/mixin/top_level_group.rb
```diff
@@ -28,10 +28,6 @@ def on_top_level_example_group(_node); end
 
         def on_top_level_group(_node); end
 
-        def top_level_group?(node)
-          top_level_groups.include?(node)
-        end
-
```
ydah, line 34: Note that while we may not be using this method, it is a mixin and may be used by custom Cops that rely on the RuboCop RSpec. And since we are using semantic versioning, we do not recommend removing this method until the next major version update. I would suggest just printing a deprecation warning first.

corsonknowles: I found 3 users of this on Github:

https://github.com/search?q=%22top_level_group%3F%22&type=code

I will look into how to test this. 

I'm not even sure we should remove it from the public API?

Also if you have a quick example of why this needs to be defined so that a subclass can overwrite it (is that even what's going on?), it would definitely save me some time. The examples I saw don't call `super` so I feel like I'm missing something.

corsonknowles: I split these changes to:

* https://github.com/rubocop/rubocop-rspec/pull/1976

I will drop the `top_level_group?` change from this PR and update the description to match.

corsonknowles:
@pirj You've been part of my journey here. Thoughts on merging this one?

ydah (commented):
Looks good

ydah (approved):

PR #1971, opened 2024-10-13
Update .simplecov

corsonknowles:
Require full line coverage

Depends on both:
* https://github.com/rubocop/rubocop-rspec/pull/1970
and 
* https://github.com/rubocop/rubocop-rspec/pull/1976

______________________________________________________________________

Before submitting the PR make sure the following are checked:

- [x] Feature branch is up-to-date with `master` (if not - rebase it).
- [x] Squashed related commits together.
- [x] Added tests.
- [x] Updated documentation.
- [x] Added an entry to the `CHANGELOG.md` if the new code introduces user-observable changes.
- [x] The build (`bundle exec rake`) passes (be sure to run this locally, since it may produce updated documentation that you will need to commit).

If you have created a new cop:

- [ ] Added the new cop to `config/default.yml`.
- [ ] The cop is configured as `Enabled: pending` in `config/default.yml`.
- [ ] The cop is configured as `Enabled: true` in `.rubocop.yml`.
- [ ] The cop documents examples of good and bad code.
- [ ] The tests assert both that bad code is reported and that good code is not reported.
- [ ] Set `VersionAdded: "<<next>>"` in `default/config.yml`.

If you have modified an existing cop's configuration options:

- [ ] Set `VersionChanged: "<<next>>"` in `config/default.yml`.

PR #1972, opened 2024-10-13
Complete branch coverage for ContextWording

corsonknowles:
Completes branch coverage for the ContextWording cop. 

See related efforts:
* https://github.com/rubocop/rubocop-rspec/pull/1970
* https://github.com/rubocop/rubocop-rspec/pull/1975 (split from this PR)
* https://github.com/rubocop/rubocop-rspec/pull/1976 (split from PR#1970)

Before:

Screenshot 2024-10-14 at 11 24 47 AM

- [x] Added tests.
- [x] Updated documentation.

### DIFF
--- a/spec/rubocop/cop/rspec/context_wording_spec.rb
+++ b/spec/rubocop/cop/rspec/context_wording_spec.rb
@@ -214,4 +214,17 @@ RSpec.describe RuboCop::Cop::RSpec::ContextWording do
       end
     end
   end
+
+  context 'when `AllowedPatterns:` and `Prefixes:` are both empty' do
+    let(:cop_config) do
+      { 'Prefixes' => [], 'AllowedPatterns' => [] }
+    end
+
+    it 'skips any description' do
+      expect_no_offenses(<<~RUBY)
+        context 'arbitrary text' do
+        end
+      RUBY
+    end
+  end
 end


spec/rubocop/cop/rspec/context_wording_spec.rb
```diff
@@ -214,4 +214,23 @@
       end
     end
   end
+  context 'when `AllowedPatterns:` is empty' do
```
pirj, line 217:
```suggestion

  context 'when `AllowedPatterns:` is empty' do
```

spec/rubocop/cop/rspec/context_wording_spec.rb
```diff
@@ -214,4 +214,24 @@
       end
     end
   end
+
+  context 'when `AllowedPatterns:` is empty' do
+    let(:cop_config) do
+      { 'Prefixes' => [], 'AllowedPatterns' => [] }
+    end
+
+    it "skips any description" do
```
ydah, line 223:
```suggestion
    it 'skips any description' do
```

spec/rubocop/cop/rspec/context_wording_spec.rb
```diff
@@ -214,4 +214,24 @@
...
+      expect_no_offenses(<<~RUBY)
+        context 'this is valid patterns context' do
+        end
+      RUBY
+    end
+
+    it "skips no description" do
```
ydah, line 230:
```suggestion
    it 'skips no description' do
```

spec/rubocop/cop/rspec/context_wording_spec.rb
```diff
@@ -214,4 +214,24 @@
       end
     end
   end
+
+  context 'when `AllowedPatterns:` is empty' do
```
pirj, line 218: Why would anyone do that? What is not getting 100% coverage if we don’t add this spec?

spec/rubocop/cop/rspec/void_expect_spec.rb
```diff
@@ -44,4 +44,18 @@
...
     RUBY
   end
+
+  it 'ignores unrelated expect method' do
+    expect_no_offenses(<<~RUBY)
+      MyObject.expect
+    RUBY
+  end
+
+  describe '#void?' do
```
pirj, line 54: Is there another way to test this private method? :x:

corsonknowles: That was my question for you!

corsonknowles: This is the guard clause in the private method that is not being invoked:

Screenshot 2024-10-14 at 6 49 05 AM

pirj: So no spec fails if we remove this line from the cop? 😕

spec/rubocop/cop/rspec/context_wording_spec.rb
corsonknowles: That method name tricked me too -- it turns out the allowed_patterns method is a combination of the Allowed Patterns and Prefixes. The earlier spec has a prefix.

spec/rubocop/cop/rspec/void_expect_spec.rb
corsonknowles: No spec fails, but I don't think its advisable to remove it. 

I plastered this with tests and got the coverage. 

Also some of the tests seem to indicate behavior that might not be wanted (outside of it blocks, etc). But in this PR I see my role as documenting the existing implementation to unblock easier test coverage. 

Take a look and see if there's anything you'd want to amend in a subsequent PR.

corsonknowles:
@pirj @ydah Thanks for your help on this one!

I think it's good to go now, just simple descriptive specs with no stubbing for these 2 Cops. 

Proof of coverage:

Screenshot 2024-10-14 at 11 15 31 AM
Screenshot 2024-10-14 at 11 15 14 AM

spec/rubocop/cop/rspec/void_expect_spec.rb
corsonknowles: I fixed the clarity here by updating the Context name to include the Prefixes for this spec

spec/rubocop/cop/rspec/context_wording_spec.rb
```diff
@@ -214,4 +214,24 @@
...
+    it 'skips any description' do
+      expect_no_offenses(<<~RUBY)
+        context 'this is valid patterns context' do
+        end
+      RUBY
+    end
+
+    it 'skips no description' do
+      expect_no_offenses(<<~RUBY)
+        context '' do
```
pirj, line 232: It feels weird to allow an empty docstring. Do we need this example?

spec/rubocop/cop/rspec/context_wording_spec.rb
```diff
@@ -214,4 +214,24 @@
...
   end
+
+  context 'when `AllowedPatterns:` and `Prefixes:` are both empty' do
+    let(:cop_config) do
+      { 'Prefixes' => [], 'AllowedPatterns' => [] }
+    end
+
+    it 'skips any description' do
+      expect_no_offenses(<<~RUBY)
+        context 'this is valid patterns context' do
```
pirj, line 225:
```suggestion
        context 'arbitrary text' do
```
Nitpick: there are no patterns to validate against

spec/rubocop/cop/rspec/void_expect_spec.rb
```diff
@@ -44,4 +44,62 @@
...
+    expect_offense(<<~RUBY)
+      it 'something' do
+        expect
+        ^^^^^^ Do not use `expect()` without `.to` or `.not_to`. Chain the methods or remove it.
+      end
+    RUBY
+  end
+
+  context 'when expect has no parent node' do
+    it 'still registers an offense' do
```
pirj, line 66: This sounds weird, and doesn’t look correct at a glance. I believe the cop should not flag a top-level `expect`, and such statement is not valid RSpec, and should not have a spec for this case.

spec/rubocop/cop/rspec/void_expect_spec.rb
```diff
@@ -44,4 +44,62 @@
       end
     RUBY
   end
+
+  it 'ignores unrelated expect method in an it block' do
```
pirj, line 48:
```suggestion
  it 'ignores unrelated expect call in example block' do
```

spec/rubocop/cop/rspec/void_expect_spec.rb
```diff
@@ -44,4 +44,62 @@
...
+    expect_no_offenses(<<~RUBY)
+      it 'something' do
+        MyObject.expect
+      end
+    RUBY
+  end
+
+  it 'rejects bare `expect` in an it block' do
```
pirj, line 56:
```suggestion
  it 'flags bare `expect` in example block' do
```

spec/rubocop/cop/rspec/void_expect_spec.rb
```diff
@@ -44,4 +44,62 @@
...
+    end
+
+    it 'rejects bare `expect` in a block' do
+      expect_offense(<<~RUBY)
+        method { expect }
+                 ^^^^^^ Do not use `expect()` without `.to` or `.not_to`. Chain the methods or remove it.
+      RUBY
+    end
+
+    it 'accepts expect.to with no arguments' do
```
pirj, line 87: Is this example necessary?
I think ‘to’ expects at least one argument.
Why do we need to cover what is not valid syntax?

pirj (commented):
It now feels that the VoidExpect needs a slight rework to be more focused on specs.

I’d love to be able to skip specs for what is not valid RSpec.

spec/rubocop/cop/rspec/void_expect_spec.rb
corsonknowles: This isn't necessary. I was just showing that that Cop will reject `expect` on main but accept `expect.to` on main.

I can probably remove most of the variations on this with no impact on coverage.

pirj: Do you think it is reasonable to fix the cop to check ‘inside_example?’ somewhere and remove the “no parent” check from the cop and related spec contexts altogether?

Thanks for this yak shaving! And sorry for an attempt to drag you even further 😅

corsonknowles: I don't think it's a good fit for the branch coverage PR. But I do think it's worth looking into!

corsonknowles: It looks like an easy fix to copy that over:
Screenshot 2024-10-14 at 12 44 08 PM

But it breaks the branch coverage again.

corsonknowles: (the `on_send` copy of `inside_example?` is covered)

corsonknowles: Draft PR here: 
https://github.com/rubocop/rubocop-rspec/pull/1975

corsonknowles: This has been an interesting exercise. I'm a little bit of a branch coverage skeptic, and a huge line coverage fan. So it's interesting to see this has revealed an issue worth addressing.

spec/rubocop/cop/rspec/context_wording_spec.rb
corsonknowles: For this spec, I impoved the clarity by updating the Context name to include that both Prefixes and AllowedPatterns must be empty at the same time.

I don't think anyone should configure the cop this way (vs. simply turning it off in that case), but this does better describe the behavior of the Cop.

corsonknowles:
@pirj @ydah Should we merge this one? It's now just a very simple spec for the missing branch in `ContextWording`

ydah:
@corsonknowles Perhaps this problem can be resolved by the following refactoring:
https://github.com/rubocop/rubocop-rspec/pull/1981

corsonknowles:
Thanks @ydah, it's a little bit philosophical. 

There is a measurement problem we want to solve, which is just that it's much easier to know when new lines are uncovered with 100% coverage. Since that is my primary concern in this recent set of PRs, I'm happy with both solutions. 

And then there is another concern for this example, what do we mean when something has a spec? There's an edge case here for when the default patterns have been deliberately removed. What do we want the Cop to do? Currently, it runs and no-ops. The spec in this PR describes that behavior. But maybe we don't want to specify what it does? Maybe we want to allow it to raise? Or have it print a warning? Philosophically, I would suggest it is easier to do that when the behavior is already described. But some might take the position that if we leave it unspecified, it's easier to change. I take the former view, because it's harder to see that an edge case needs handling when it is not described. 

If we fix it at the code level, but don't add this spec, we still leave this edge case un-described.

bquorning (approved):
I like this. Describing edge case behavior in specs makes sure that _if_ we change it, it will be an intentional change (where we need to change the spec as well).

ydah (approved):
No objection to adding this spec.

spec/rubocop/cop/rspec/context_wording_spec.rb
```diff
@@ -214,4 +214,17 @@
       end
     end
   end
+
+  context 'when `AllowedPatterns:` and `Prefixes:` are both empty' do
```
ydah, line 218: memo) In this case, it is pointless to enable this cop because it is not always an offense. Should we occur warning or runtime error for this setting? WDYT? @rubocop/rubocop-rspec

bquorning: >  it is not always an offense

I don’t understand. Do you mean it’s not deterministic?

ydah: The following checks are always considered non-offenses:

https://github.com/rubocop/rubocop-rspec/blob/7592cecc0a16e2f73475e5b60c72d0baa13b2a29/lib/rubocop/cop/rspec/context_wording.rb#L88

This Cop sets valid wording in Prefixes and AllowedPatterns, but it does not always consider a offenses if there is no valid wording setting.

corsonknowles: Right, with these settings, there will be no offenses, regardless of what is or is not in the Context. It is always not an offense. (I'm not attached to it staying that way). 

I looked at the Obsoletion setup recently, and I think it raises runtime errors with the expectation that the user will want to fix those issues. That feels similar to how a user may want to fix the config or disable this cop instead of letting it run "on empty"

bquorning: It’s an unlikely combination of settings, so unless it’s very simple to implement, I don‘t think we need to warn our users about it.

ydah: Maybe if the `Prefixes` and `AllowedPatterns` are empty to begin with, there are no allowed Wording, so maybe it's better to make them all violations. That way, users would be able to notice any misconfigurations and the implementation wouldn't be that hard.

PR #1973, opened 2024-10-13
Update unspecified_exception_spec.rb

corsonknowles:
Some mild stubbing to cover an uncovered branch of code. 

Maybe there is a better way? Or a code change?

- [x] Added tests.
- [x] Updated documentation.

### DIFF
--- a/spec/rubocop/cop/rspec/unspecified_exception_spec.rb
+++ b/spec/rubocop/cop/rspec/unspecified_exception_spec.rb
@@ -254,5 +254,27 @@ RSpec.describe RuboCop::Cop::RSpec::UnspecifiedException do
         }.to raise_exception(StandardError)
       RUBY
     end
+
+    it 'skips ancestors that are not send types' do
+      node = parse_source(<<~RUBY).ast
+        expect {
+          raise StandardError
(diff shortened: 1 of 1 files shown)


pirj:
Can you make it green?

Please feel to ignore the ruby-head failure

spec/rubocop/cop/rspec/unspecified_exception_spec.rb
```diff
@@ -254,5 +254,27 @@ def raise_error
...
     end
+
+    it 'skips ancestors that are not send types' do
+      node = parse_source(<<~RUBY).ast
+        expect {
+          raise StandardError
+        }.to raise_error
+      RUBY
+
+      allow_any_instance_of(RuboCop::AST::Node)
```
bquorning, line 265: I think I would rather have missing code coverage than using `allow_any_instance_of` in our specs 😅 Let’s see if we can find another approach.

corsonknowles: That's fair. 

I'm going to close this PR, I may or may not circle back to this one as I don't have any other ideas on how to test this right now.

PR #1975, opened 2024-10-14
Require VoidExpect to operate inside an example block

corsonknowles:
This PR completes branch coverage for `VoidExpect`.

This PR copies the `inside_example?` method and re-uses it here to get more coherent behavior from the VoidExpect cop. 

Note that if we were thinking of copying `inside_example?` a 3rd time, I'd suggest factoring it out into a utility (Sandi Metz' rule of 3).

Note that I remove the `return true unless parent` guard clause, because we expect the `inside_example?` guard clause to fully cover that condition.

Split from:
* https://github.com/rubocop/rubocop-rspec/pull/1972

- [x] Added tests.
- [x] Updated documentation.

### DIFF
--- a/spec/rubocop/cop/rspec/void_expect_spec.rb
+++ b/spec/rubocop/cop/rspec/void_expect_spec.rb
@@ -44,4 +44,26 @@ RSpec.describe RuboCop::Cop::RSpec::VoidExpect do
       end
     RUBY
   end
+
+  it 'ignores unrelated method named expect in an example block' do
+    expect_no_offenses(<<~RUBY)
+      it 'something' do
+        MyObject.expect(:foo)
+      end
+    RUBY
+  end
+
+  context 'when expect has no parent node' do
+    it 'does not register an offense' do
+      expect_no_offenses(<<~RUBY)
+        expect(something)
+      RUBY
+    end
+
+    it 'does not register an offense for unrelated expect with block' do
+      expect_no_offenses(<<~RUBY)
+        expect { block_contents }
+      RUBY
+    end
+  end
 end
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Master (Unreleased)
 
+- Fix `RSpec/VoidExpect` to only operate inside an example block. ([@corsonknowles])
+
 ## 3.1.0 (2024-10-01)
 
 - Add `RSpec/StringAsInstanceDoubleConstant` to check for and correct strings used as instance_doubles. ([@corsonknowles])
--- a/lib/rubocop/cop/rspec/void_expect.rb
+++ b/lib/rubocop/cop/rspec/void_expect.rb
@@ -29,12 +29,14 @@ module RuboCop
 
         def on_send(node)
           return unless expect?(node)
+          return unless inside_example?(node)
 
           check_expect(node)
         end
 
         def on_block(node) # rubocop:disable InternalAffairs/NumblockHandler
           return unless expect_block?(node)
+          return unless inside_example?(node)
 
           check_expect(node)
         end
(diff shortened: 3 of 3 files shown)


corsonknowles:
@pirj Can you review 1 more time please?

I think I solved this, but it entails removing this guard clause
`return true unless parent`
 in favor of what I believe is the much more restrictive one:
`return unless inside_example?(node)`

At this point, I can probably drop all VoidExpect changes from the other PR, so it's just amending that 1 other spec file.

spec/rubocop/cop/rspec/void_expect_spec.rb
```diff
@@ -44,4 +44,47 @@
...
+      it 'something' do
+        MyObject.expect
+      end
+    RUBY
+  end
+
+  it 'flags bare `expect` in example block' do
+    expect_offense(<<~RUBY)
+      it 'something' do
+        expect
```
corsonknowles, line 59: This behavior is also debatable. I'm somewhat in favor of leaving it. 

I think this would only be possible if someone is shadowing (overwrote in some context) the `expect` method. In that case, ideally there would be an even more specific error message, but I also don't think it's a case that deserves its own cop. 

So the fact that there is a Linter violation here is warranted I think, even if the message is not directly addressing how strange this is.

pirj: Completely agree, a “bare” is [invalid](https://github.com/rspec/rspec-expectations/blob/838952dc1416c943e7933684c47249a77481e7f5/lib/rspec/expectations/expectation_target.rb#L39) in RSpec.

Do you think we can just remove those examples with “bare” expect calls?

spec/rubocop/cop/rspec/void_expect_spec.rb
```diff
@@ -44,4 +44,47 @@
       end
     RUBY
   end
+
+  it 'ignores unrelated method named expect in an example block' do
+    expect_no_offenses(<<~RUBY)
+      it 'something' do
+        MyObject.expect
```
pirj, line 51:
```suggestion
        MyObject.expect(:foo)
```

spec/rubocop/cop/rspec/void_expect_spec.rb
```diff
@@ -44,4 +44,47 @@
...
+  it 'flags bare `expect` in example block' do
+    expect_offense(<<~RUBY)
+      it 'something' do
+        expect
+        ^^^^^^ Do not use `expect()` without `.to` or `.not_to`. Chain the methods or remove it.
+      end
+    RUBY
+  end
+
+  context 'when expect has no parent node' do
```
pirj, line 65: I’d just removed all this except the first one.

spec/rubocop/cop/rspec/void_expect_spec.rb
corsonknowles: I guess the only reason to keep this and essentially document this behavior is if we thought it were worth changing. 

Since we both agree it's not worth changing, I will go ahead and delete it. 

I still feel like we "know less" about how the Cop works with out this -- in the sense that I learned something more about how it worked by writing this, but that's perhaps not a strong enough reason to commit it since it's covered by RSpec behavior (or the user did something very strange).

spec/rubocop/cop/rspec/void_expect_spec.rb
corsonknowles: We need the block version below to cover the new guard clause in `on_block` 

I'll delete the rest and get this merged in.

corsonknowles:
@pirj Feel free to merge at will 
(or reapprove and I will merge? however this works)

pirj:
We usually get consensus on things that might be user-facing, so I’ll pass on merging to the next reviewer.

May I kindly ask to add a short changelog note that the cop was fixed to only inspect code inside examples?

corsonknowles:
Thanks! CHANGELOG entry added.

corsonknowles:
@bquorning Would you be up for doing the 2nd review on this one?

ydah (approved):
Looks good!

bquorning (approved):
Yep, looks good 👍🏼

PR #1976, opened 2024-10-14
Deprecate top_level_group? and test it in a Cop class

corsonknowles:
In this PR, we add a deprecation message for `top_level_group?`, which has no callers in current `Rubocop/RSpec`. 

`top_level_group?` was confirmed unused by @bquorning here:
* https://github.com/rubocop/rubocop-rspec/pull/1957#issuecomment-2370935206
* code change: https://github.com/rubocop/rubocop-rspec/pull/977/files#diff-538fd5bdf99dca6200224146c399b06a85bc9297598c1c4e71e0c9eaead8770bL15

We test this method using some stubbing in a Cop class that includes the mixin. 

The specs can be removed along with the method in the next Major version update. 

Split from:
* https://github.com/rubocop/rubocop-rspec/pull/1970

This PR serves the goal of completing line coverage for this repo ~and unblocks~ helps unblock:

- [x] Added tests.
- [x] Updated documentation.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 - Fix `RSpec/VoidExpect` to only operate inside an example block. ([@corsonknowles])
 - Change `RSpec/ContextWording` cop to always report an offense when both `Prefixes` and `AllowedPatterns` are empty. ([@ydah])
 - Fix an error for `RSpec/ChangeByZero` when `change (...) .by (0)` and `change (...)`, concatenated with `and` and `or`. ([@ydah])
+- Deprecate `top_level_group?` method from `TopLevelGroup` mixin as all of its callers were intentionally removed from `Rubocop/RSpec`. ([@corsonknowles])
 
 ## 3.1.0 (2024-10-01)
 
--- a/lib/rubocop/cop/rspec/mixin/top_level_group.rb
+++ b/lib/rubocop/cop/rspec/mixin/top_level_group.rb
@@ -7,6 +7,9 @@ module RuboCop
       module TopLevelGroup
         extend RuboCop::NodePattern::Macros
 
+        DEPRECATED_MODULE_METHOD_WARNING =
+          'top_level_group? is deprecated and will be ' \
+          'removed in the next major version of rubocop_rspec.'
         def on_new_investigation
           super
 
@@ -28,7 +31,10 @@ module RuboCop
 
         def on_top_level_group(_node); end
 
+        # @private
+        # @deprecated All callers of this method have been removed.
         def top_level_group?(node)
+          warn DEPRECATED_MODULE_METHOD_WARNING, uplevel: 1
           top_level_groups.include?(node)
         end
 
--- a/spec/rubocop/cop/rspec/mixin/top_level_group_spec.rb
+++ b/spec/rubocop/cop/rspec/mixin/top_level_group_spec.rb
@@ -1,0 +1,24 @@
+# frozen_string_literal: true
+
+RSpec.describe RuboCop::Cop::RSpec::TopLevelGroup do
+  describe '#top_level_group?' do
+    let(:stub_class) do
+      Class.new do
+        include RuboCop::Cop::RSpec::TopLevelGroup
+
+        def initialize
+          @top_level_groups = []
+        end
+
+        def test_top_level_group
+          top_level_group?(nil)
+        end
+      end
+    end
+
+    it 'warns because it is deprecated' do
+      expect { stub_class.new.test_top_level_group }.to \
+        output(/warning: top_level_group\? is deprecated/).to_stderr
+    end
+  end
+end


lib/rubocop/cop/rspec/mixin/top_level_group.rb
```diff
@@ -29,6 +34,7 @@ def on_top_level_example_group(_node); end
         def on_top_level_group(_node); end
 
         def top_level_group?(node)
```
pirj, line 36:
```suggestion
         # @private
         # @deprecated <some description on how to replace>
        def top_level_group?(node)
```

This is more a reminder for us not to forget to remove in 4.0
Should we file a ticket?

CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 ## Master (Unreleased)
 
+- Deprecate `top_level_group?` private method on TopLevelGroup mixin. ([@corsonknowles])
```
pirj, line 5: It’s not exactly private, just like any “private” from includable modules.

lib/rubocop/cop/rspec/mixin/top_level_group.rb
```diff
@@ -7,6 +7,11 @@ module RSpec
       module TopLevelGroup
         extend RuboCop::NodePattern::Macros
 
+        DEPRECATED_PRIVATE_METHOD_WARNING =
+          'top_level_group? is deprecated and will be ' \
+          'removed in the next major version of Rubocop/RSpec.' \
+          'Please copy the method if you need it for your mixin.'
```
pirj, line 13: Is it a good advice to copy?
We’ve stopped using it for a reason?
Was that performance? Should we say to consider using the hook instead?

lib/rubocop/cop/rspec/mixin/top_level_group.rb
```diff
@@ -29,6 +34,7 @@ def on_top_level_example_group(_node); end
         def on_top_level_group(_node); end
 
         def top_level_group?(node)
+          warn DEPRECATED_PRIVATE_METHOD_WARNING
           top_level_groups.include?(node)
         end
```
pirj, line 40: Something feels off about the `node.children` line.
I vaguely recall that a single-statement block and a multi-statement block’s AST are different (yuck!). Will the result be correct in both cases?
Like
`foo { RSpec.describe { } }` vs `foo { RSpec.describe { }; RSpec.shared_examples { } }`?


Unrelated to this PR, just an observation.
Might as well be a false alarm, and I apologise in advance.

lib/rubocop/cop/rspec/mixin/top_level_group.rb
corsonknowles: I actually don't know what the hook is. Do you want to suggest language?

lib/rubocop/cop/rspec/mixin/top_level_group.rb
corsonknowles: Sounds like more specs are needed!

lib/rubocop/cop/rspec/mixin/top_level_group.rb
corsonknowles: Like an Issue? Or is there an internal ticketing system?

pirj:
You’ll need to rebase to fix the failing config/docs CI check.

corsonknowles:
Rebased and ready to go.

corsonknowles:
@ydah or @bquorning Would either of you like to be the 2nd reviewer on this?

lib/rubocop/cop/rspec/mixin/top_level_group.rb
corsonknowles: I removed the copying advice

lib/rubocop/cop/rspec/mixin/top_level_group.rb
corsonknowles: I added the `@deprecated` declaration, but I'm still not sure what filing a ticket entails.

bquorning: Yeah, we just use issues. I created a `4.0` milestone and opened #1986.

lib/rubocop/cop/rspec/mixin/top_level_group.rb
```diff
@@ -7,6 +7,9 @@ module RSpec
       module TopLevelGroup
         extend RuboCop::NodePattern::Macros
 
+        DEPRECATED_PRIVATE_METHOD_WARNING =
+          'top_level_group? is deprecated and will be ' \
+          'removed in the next major version of Rubocop/RSpec.'
```
bquorning, line 12:
```suggestion
          'removed in the next major version of rubocop_rspec.'

```

lib/rubocop/cop/rspec/mixin/top_level_group.rb
```diff
@@ -28,7 +31,10 @@ def on_top_level_example_group(_node); end
 
         def on_top_level_group(_node); end
 
+        # @private
+        # @deprecated All callers of this method have been removed.
         def top_level_group?(node)
+          warn DEPRECATED_PRIVATE_METHOD_WARNING
```
bquorning, line 37: In RuboCop [they use](https://github.com/rubocop/rubocop/blob/87740810904f57499e1255d492b80b4bd83d3a66/lib/rubocop/cop/mixin/configurable_max.rb#L12-L14) `Rainbow` to make yellow warnings. I’m a bit hesitant to add it as a dependency to our gem, but I think we should use the `uplevel` keyword argument that they also use:
```suggestion
          warn DEPRECATED_PRIVATE_METHOD_WARNING, uplevel: 1
```
That also prints the line that called the method, plus the word `warning`.

spec/rubocop/cop/rspec/describe_class_spec.rb
```diff
@@ -215,4 +215,23 @@
...
+        allow(cop_with_top_level_group_mixin).to receive(:warn)
+        allow(cop_with_top_level_group_mixin).to receive(:top_level_groups)
+          .and_return([])
+
+        cop_with_top_level_group_mixin.send(:top_level_group?, nil)
+        expect(cop_with_top_level_group_mixin).to have_received(:warn)
+        # rubocop:enable RSpec/SubjectStub
+      end
+    end
+  end
```
bquorning, line 236: We can add a new file testing only the module you changed – and that way we can (almost) avoid stubbing. E.g.:
```rb
# spec/rubocop/cop/rspec/mixin/top_level_group_spec.rb

RSpec.describe RuboCop::Cop::RSpec::TopLevelGroup do
  describe '#top_level_group?' do
    let(:stub_class) do
      Class.new do
        include RuboCop::Cop::RSpec::TopLevelGroup

        def initialize
          @top_level_groups = []
        end

        def test_top_level_group
          top_level_group?(nil)
        end
      end
    end

    it 'warns because it is deprecated' do
      expect { stub_class.new.test_top_level_group }.to \
        output(/warning: top_level_group\? is deprecated/).to_stderr
    end
  end
end
```

corsonknowles:
Unfortunately after rebasing, there's a new straggler:
Screenshot 2024-10-25 at 4 23 54 PM

corsonknowles:
This would have locked in coverage:

But I can't chase a moving target at this speed.

bquorning (approved):

CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 ## Master (Unreleased)
 
+- Deprecate `top_level_group?` method from `TopLevelGroup` mixin as all of its callers were intentionally removed from `Rubocop/RSpec`. ([@corsonknowles])
+- 
 ## 3.2.0 (2024-10-26)
```
bquorning:
```suggestion
- Deprecate `top_level_group?` method from `TopLevelGroup` mixin as all of its callers were intentionally removed from `Rubocop/RSpec`. ([@corsonknowles])

## 3.2.0 (2024-10-26)
```

lib/rubocop/cop/rspec/mixin/top_level_group.rb
```diff
@@ -7,6 +7,9 @@ module RSpec
       module TopLevelGroup
         extend RuboCop::NodePattern::Macros
 
+        DEPRECATED_MODULE_METHOD_WARNING =
+          'top_level_group? is deprecated and will be ' \
+          'removed in the next major version of rubocop_rspec.'
         def on_new_investigation
```
bquorning:
```suggestion
          'removed in the next major version of rubocop_rspec.'

        def on_new_investigation
```

bquorning (approved):

bquorning:
> Unfortunately after rebasing, there's a new straggler:

The straggler went away with #1985. I’ll rebase your other branch (#1971) and get it merged while we’re still at 100% ✨

PR #1979, opened 2024-10-16
Fix problem with `Open3.popen3` where `confirm_documentation` task never finishes waiting for child processes

ydah:
see: https://github.com/rubocop/rubocop-rspec/actions/runs/11360936578/job/31599655573

This PR is composed of the following three commits:

## 1. Fix problem with `Open3.popen3` where `confirm_documentation` task never finishes waiting for child processes 799df9101977ff41a36bba64ecf9d33385f29b1d

It freezes at confirm_documentation as follows:

```
❯ bundle exec rake confirm_documentation
Files:         123
Modules:        16 (    3 undocumented)
Classes:       116 (    0 undocumented)
Constants:     191 (  191 undocumented)
Attributes:      1 (    0 undocumented)
Methods:       321 (  252 undocumented)
 30.85% documented
* generated /Users/ydah/rubocop-rspec/docs/modules/ROOT/pages//cops_rspec.adoc
```

The documentation for `Open3.popen3` described the following:

Refs: https://www.rubydoc.info/stdlib/open3/Open3.popen3
>You should be careful to avoid deadlocks. Since pipes are fixed length buffers, `Open3.popen3("prog") {|i, o, e, t| o.read }` deadlocks if the program generates too much output on stderr. You should read stdout and stderr simultaneously (using threads or IO.select). However, if you don’t need stderr output, you can use `Open3.popen2`. If merged stdout and stderr output is not a problem, you can use `Open3.popen2e`. If you really need stdout and stderr output as separate strings, you can consider `Open3.capture3`.

Perhaps, but the following deadlock conditions were present.
- Child process: It was waiting to write beyond the buffer size to stderr output.
- Parent process: Waiting for the pipe to which the standard output of the child process was to be closed.

## 2. Improve confirm_documentation task to display out-of-sync documentation e6ccc3796eb365d617db6934f39735e8b4f825db

This is a change to output what is the difference when the `confirm_documentation` task fails and prompts you to run `rake generate_cops_documentation` but you have not made any changes that affect the documentation.
I actually got confused when using ruby 3.4.0dev because I forgot that the hash format had changed as follows
https://github.com/rubocop/rubocop-rspec/pull/1974/files

## 3. bundle exec rake generate_cops_documentation 9e2ecd9a4040f3023a7473be189bd07512f3d891

This is a change that only updates the documentation. using `rake generate_cops_documentation`
Without this, CI will fail.

- [ ] Added tests.
- [ ] Updated documentation.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -72,12 +72,15 @@ end
 
 desc 'Confirm documentation is up to date'
 task confirm_documentation: :generate_cops_documentation do
-  _, _, status =
+  stdout, _stderr, status =
     Open3.capture3('git diff --exit-code docs/')
 
   unless status.success?
-    raise 'Please run `rake generate_cops_documentation` ' \
-          'and add docs/ to the commit.'
+    warn 'Documentation is out of sync:'
+    warn stdout
+    warn 'Please run `rake generate_cops_documentation` ' \
+         'and add docs/ to the commit.'
+    exit 1
   end
 end
 
--- a/Rakefile
+++ b/Rakefile
@@ -72,10 +72,10 @@ end
 
 desc 'Confirm documentation is up to date'
 task confirm_documentation: :generate_cops_documentation do
-  _, _, _, process =
-    Open3.popen3('git diff --exit-code docs/')
+  _, _, status =
+    Open3.capture3('git diff --exit-code docs/')
 
-  unless process.value.success?
+  unless status.success?
     raise 'Please run `rake generate_cops_documentation` ' \
           'and add docs/ to the commit.'
   end
--- a/docs/modules/ROOT/pages/cops_rspec.adoc
+++ b/docs/modules/ROOT/pages/cops_rspec.adoc
@@ -6,6 +6,7 @@
 
 = RSpec
 
+[#rspecalignleftletbrace]
 == RSpec/AlignLeftLetBrace
 
 |===
@@ -20,6 +21,7 @@
 
 Checks that left braces for adjacent single line lets are aligned.
 
+[#examples-rspecalignleftletbrace]
 === Examples
 
 [source,ruby]
@@ -35,10 +37,12 @@ let(:baz)    { bar }
 let(:a)      { b }
 ----
 
+[#references-rspecalignleftletbrace]
 === References
 
 * https://www.rubydoc.info/gems/rubocop-rspec/RuboCop/Cop/RSpec/AlignLeftLetBrace
 
+[#rspecalignrightletbrace]
 == RSpec/AlignRightLetBrace
 
 |===
@@ -53,6 +57,7 @@ let(:a)      { b }
 
 Checks that right braces for adjacent single line lets are aligned.
 
+[#examples-rspecalignrightletbrace]
 === Examples
 
 [source,ruby]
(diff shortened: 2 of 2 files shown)


bquorning (commented):
I think I need some more explanation of each commit in this PR.

ydah:
@bquorning Updated PR description. How do you like it?

Rakefile
```diff
@@ -72,10 +72,12 @@ end
...
 task confirm_documentation: :generate_cops_documentation do
-  _, _, _, process =
-    Open3.popen3('git diff --exit-code docs/')
+  stdin, _, process =
+    Open3.capture3('git diff --exit-code docs/')
 
-  unless process.value.success?
+  unless process.to_i.zero?
+    puts 'Documentation is out of sync:'
+    puts stdin
```
bquorning, line 80:
```suggestion
    puts stdout
```

Rakefile
```diff
@@ -72,10 +72,12 @@ end
...
     raise 'Please run `rake generate_cops_documentation` ' \
           'and add docs/ to the commit.'
```
bquorning, line 82:
```suggestion
    puts 'Please run `rake generate_cops_documentation` ' \
          'and add docs/ to the commit.'
    exit 1
```

bquorning (commented):
I didn’t know about the deadlock problem with `popen3`, but it makes perfect sense now that you mention it.

There’s a few problems with your PR (stdin is in fact stdout, and we get a status object returned, not a process – which means we can call `#success?` on it.

I would also suggest that we don’t `raise` at the end of the task, but just `exit` with an error status.

Rakefile
```diff
@@ -72,12 +72,15 @@ end
...
-    Open3.popen3('git diff --exit-code docs/')
-
-  unless process.value.success?
-    raise 'Please run `rake generate_cops_documentation` ' \
-          'and add docs/ to the commit.'
+  stdout, _stderr, status =
+    Open3.capture3('git diff --exit-code docs/')
+
+  unless status.success?
+    puts 'Documentation is out of sync:'
```
pirj, line 79:
```suggestion
    warn 'Documentation is out of sync:'
```

Same as `warn` for us running on CI ? (Skipping the discussion on stdout vs stderr)


This line and the three below

pirj (approved):
One nitpick

Thank you!

I yet to dive into the pr description to understand why this suddenly started happening in the first place

bquorning:
> why this suddenly started happening in the first place

We run out of IO buffer and deadlock the process when there are _too_ many errors.

bquorning:
But yes, why did the task suddenly start failing? Did something change in rubocop master branch?

Rakefile
ydah: Fixed it 👍🏻  Thank you.

ydah:
> But yes, why did the task suddenly start failing? Did something change in rubocop master branch?

I think it's because the following change made the header links to be output, and there were a lot of diffs.
- https://github.com/rubocop/rubocop/commit/2e6254f516f6b3c2604543d065a8bef23e8116bd

pirj:
Thank you! ❤️

PR #1980, opened 2024-10-17
Add job for CI using Prism parser

bquorning:
Inspired by https://github.com/rubocop/rubocop-capybara/pull/112

- [ ] Added tests.
- [ ] Updated documentation.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -119,3 +119,24 @@ jobs:
           ruby-version: "3.3"
           bundler-cache: true
       - run: NO_COVERAGE=true bundle exec rake spec
+
+  prism:
+    runs-on: ubuntu-latest
+    name: Prism
+    steps:
+      - uses: actions/checkout@v4
+      - name: Use prism parser
+        run: |
+          cat << EOF > Gemfile.local
+            gem 'prism'
+          EOF
+      - name: set up Ruby
+        uses: ruby/setup-ruby@v1
+        with:
+          # Specify the minimum Ruby version 2.7 required for Prism to run.
+          ruby-version: "2.7"
+          bundler-cache: true
+      - env:
+          NO_COVERAGE: true
+          PARSER_ENGINE: parser_prism
+        run: bundle exec rake spec


pirj:
But ruby-head is prism already, isn’t this sufficient?

bquorning:
Ah yes, probably so. I'll keep this in draft and look at it again next week.

Earlopain:
My understanding is that it will still use the plain  parser gem, even with ruby-head/ruby-3.4
The prism default is runtime-only. Users need to set it in their config files still, even when targeting ruby 3.3 or higher.

pirj:
Thanks for the hint, @Earlopain !

I’m good with a separate ci job with prism.
Do we need to turn it on in RuboCop somehow? 🤔

Earlopain:
As a user, you would set `AllCops/ParserEngine`. For extensions, this should have you covered: https://github.com/rubocop/rubocop/blob/9f2ffa2270bbc174b2a622350d120b05672a1ce5/lib/rubocop/rspec/cop_helper.rb#L13. I'm not immediatly finding the place where you inherit this, but you probably do somewhere? Since you use the same assertion methods from `RuboCop::RSpec::ExpectOffense`.

For reference, here's the PR for `rubocop-rails` and it basically only sets the env variable. https://github.com/rubocop/rubocop-rails/pull/1245. The other changes are for places that construct `ProcessedSource` themselves but I don't believe that will be the case here.

.github/workflows/main.yml
```diff
@@ -119,3 +119,23 @@ jobs:
...
+          EOF
+      - name: set up Ruby
+        uses: ruby/setup-ruby@v1
+        with:
+          # Specify the minimum Ruby version 2.7 required for Prism to run.
+          ruby-version: "2.7"
+          bundler-cache: true
+      - env:
+          PARSER_ENGINE: parser_prism
+        run: NO_COVERAGE=true bundle exec rake
```
Earlopain, line 141: Docs generator is broken with Ruby 2.7, I openend https://github.com/rubocop/rubocop/pull/13388. Probably, you can just run `bundle exec rake spec` here regardless.

bquorning:
I changed it to run only `rake spec` and not all of `rake`. Does it look ok now?

ydah (approved):
Looks good! Thank you♥

PR #1981, opened 2024-10-18
Remove unnecessary early return in `RSpec/ContextWording`cop

ydah:
Checks if `allowed_patterns` is empty, but looking at the implementation of `matches_allowed_pattern?`, it refers to it as `allowed_patterns.any?`, so `return false if allowed_patterns .empty?` is a completely unnecessary early return.

https://github.com/rubocop/rubocop/blob/e7ccb0200d013d1c4b75b3face0dea0538620769/lib/rubocop/cop/mixin/allowed_pattern.rb#L29-L31

```
def matches_allowed_pattern?(line)
  allowed_patterns.any? { |pattern| Regexp.new(pattern).match?(line) }
end
```

also resolve:  https://github.com/rubocop/rubocop-rspec/pull/1972

image

- [-] Added tests.
- [-] Updated documentation.

### DIFF
--- a/lib/rubocop/cop/rspec/context_wording.rb
+++ b/lib/rubocop/cop/rspec/context_wording.rb
@@ -67,7 +67,7 @@ module RuboCop
 
         def on_block(node) # rubocop:disable InternalAffairs/NumblockHandler
           context_wording(node) do |context|
-            if bad_pattern?(context)
+            unless matches_allowed_pattern?(description(context))
               message = format(MSG, patterns: expect_patterns)
               add_offense(context, message: message)
             end
@@ -82,12 +82,6 @@ module RuboCop
 
         def prefix_regexes
           @prefix_regexes ||= prefixes.map { |pre| /^#{Regexp.escape(pre)}\b/ }
-        end
-
-        def bad_pattern?(node)
-          return false if allowed_patterns.empty?
-
-          !matches_allowed_pattern?(description(node))
         end
 
         def description(context)


lib/rubocop/cop/rspec/context_wording.rb
```diff
@@ -65,9 +65,17 @@ class ContextWording < Base
           (block (send #rspec? { :context :shared_context } $({str dstr xstr} ...) ...) ...)
         PATTERN
 
+        # @!method allowed?(node)
+        def_node_matcher :allowed?, <<~PATTERN
+          {
+            (str #matches_allowed_pattern?)
+            ({dstr xstr} (str #matches_allowed_pattern?) ...)
```
pirj, line 72: This reads like it accepts twoand more  arguments, and checks if it matches on the second one. Is this correct?

pirj:
I’m quite concerned if we had a method, and it was not fully covered with specs, and now we’re refactoring it to a node pattern where coverage is not calculated, can we say that it does the same thing with certainty?
I live node patterns, they are often more expressive due to their declarative nature.
But does this one read better than the method?

Node patterns support inline comments, but we rarely use that, and it’s not always useful.
Node pattern has this @!method YARD directive, but we often miss the opportunity to add examples of what it does match and what not.

With all this, I think in this case a node pattern is inferior to a method.
Especially since this change conceals a gap in our spec coverage.

ydah:
@pirj It is certainly not a good choice to replace it with a node pattern for which coverage is not calculated.
So I have updated this PR concern as removing unnecessary early returns. What do you think?

pirj (approved):
Nice refactoring!
Great to see it’s getting us to 100% coverage ☺️

pirj:
Another pair of eyes in case I’m missing something this early morning, @rubocop/rubocop-rspec ?

bquorning:
Isn’t the reason we didn’t have complete spec coverage before that we never hit a case where `allowed_patterns.empty?`? Should we add a `context 'when both Prefixes and AllowedPatterns are empty'` to the specs before refactoring?

corsonknowles:
@bquorning That spec is available here:
* https://github.com/rubocop/rubocop-rspec/pull/1972

ydah:
Close this PR because it was found to operate differently from the current specification.

PR #1985, opened 2024-10-25
Revert an change that would conceal unintentional RSpec syntax

pirj:
Specifically:

    expect { ... }.to change { ... }.by(...) and change { ... }.by(...)

Here the usage of `and` is incorrect, as RSpec does not (and can't reasonably) support it.
Compound should use the `and` method, not the operator, or the `&`. Same for `.or` and `|`.

Related: #1984 #1983

- [-] Added tests.
- [-] Updated documentation.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,8 +3,7 @@
 ## Master (Unreleased)
 
 - Change `RSpec/ContextWording` cop to always report an offense when both `Prefixes` and `AllowedPatterns` are empty. ([@ydah])
-- Fix an error for `RSpec/ChangeByZero` when `change (...) .by (0)` and `change (...)`, concatenated with `and` and `or`. ([@ydah])
-- Revert an change that would conceal unintentional RSpec syntax. ([@pirj])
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 
 - Change `RSpec/ContextWording` cop to always report an offense when both `Prefixes` and `AllowedPatterns` are empty. ([@ydah])
 - Fix an error for `RSpec/ChangeByZero` when `change (...) .by (0)` and `change (...)`, concatenated with `and` and `or`. ([@ydah])
+- Revert an change that would conceal unintentional RSpec syntax. ([@pirj])
```
bquorning, line 7: These two entries should probably be combined.

bquorning (approved):

CHANGELOG.md
pirj: Updated.

bquorning (approved):

ydah (approved):

ydah:
Thank you!

PR #1990, opened 2024-10-26
Complete branch coverage for more files

corsonknowles:
Increase branch coverage minimum threshold.

Modernize `.rubocop_todo.yml` with Rubocop 1.6.8

In the further pursuit of full coverage for `Rubocop/RSpec`, this solves for additional files.

> [!NOTE]
> We remove a safe operator from a method used in a node matcher here, since it should always get a node.

Reconciled with:
* https://github.com/rubocop/rubocop-rspec/pull/1989

- [x] Added tests.
- [x] Updated documentation.


bquorning:
Surprisingly little duplication between this and #1989 :yay:

corsonknowles:
@bquorning @ydah I reconciled this with #1989 and it should be ready to go!

spec/rubocop/cop/rspec/stubbed_mock_spec.rb
```diff
@@ -134,4 +134,23 @@
...
+    it 'returns "allow(subject)" for :is_expected' do
+      expect(cop.send(:replacement, :is_expected)).to eq('allow(subject)')
+    end
+
+    it 'returns "allow_any_instance_of" for :expect_any_instance_of' do
+      expect(cop.send(:replacement,
+                      :expect_any_instance_of)).to eq(:allow_any_instance_of)
+    end
+
+    it 'falls through silently and returns nil for unknown methods' do
```
corsonknowles, line 152: technically this is all that's needed in here to solve for the `case` statement

corsonknowles:
The Ruby head CI failure comes down to how rocket hashes are spaced in the parser.

bquorning:
> The Ruby head CI failure comes down to how rocket hashes are spaced in the parser.

See https://github.com/rubocop/rubocop-rspec/blob/00b6c0387652f8f5140465f92b3299a746e9d18d/spec/rubocop/rspec/hook_spec.rb#L77-L83 for inspiration on how to work around that.

bquorning (approved):
Thanks

PR #1991, opened 2024-10-30
Cover and then remove `ancestor.send_type?` branch in `UnspecifiedException`

corsonknowles:
Wraps some non-send type branches around `raise_error` specs

From:
Screenshot 2024-10-30 at 5 44 37 AM

To:
Screenshot 2024-10-30 at 5 39 32 AM

Then we remove this line, because all specs pass without it due to the node pattern covering both branches.

- [x] Added tests.
- [x] Updated documentation.

### DIFF
--- a/lib/rubocop/cop/rspec/unspecified_exception.rb
+++ b/lib/rubocop/cop/rspec/unspecified_exception.rb
@@ -64,7 +64,6 @@ module RuboCop
         def find_expect_to(node)
           node.each_ancestor.find do |ancestor|
             break if ancestor.block_type?
-            next unless ancestor.send_type?
 
             expect_to?(ancestor)
           end
--- a/spec/rubocop/cop/rspec/unspecified_exception_spec.rb
+++ b/spec/rubocop/cop/rspec/unspecified_exception_spec.rb
@@ -254,5 +254,12 @@ RSpec.describe RuboCop::Cop::RSpec::UnspecifiedException do
         }.to raise_exception(StandardError)
       RUBY
     end
+
+    it 'detects even when a non-send node is an ancestor' do
+      expect_offense(<<~RUBY)
+        expect { raise 'error' }.to (branch_conditional ? raise_error : handle_exception)
+                                                          ^^^^^^^^^^^ Specify the exception being captured
+      RUBY
+    end
   end
 end


spec/rubocop/cop/rspec/unspecified_exception_spec.rb
```diff
@@ -254,5 +254,22 @@ def raise_error
         }.to raise_exception(StandardError)
       RUBY
     end
+
+    it 'detects even when class definition is a direct ancestor' do
+      expect_offense(<<~RUBY)
+        expect { raise 'error' }.to class MyClass
+          raise_error
+          ^^^^^^^^^^^ Specify the exception being captured
+        end
```
bquorning, line 263: Is this even valid RSpec code?

corsonknowles: Oh no, these are monstrous constructions. I apologize if I undersold that in the description. 

The issue here is that to reach this condition:
```
        def find_expect_to(node)
          node.each_ancestor.find do |ancestor|
            break if ancestor.block_type?
            next unless ancestor.send_type?

            expect_to?(ancestor)
          end
        end
```

we have to put some kind of non-send, non-block node inside the `.to` statement of the expectation. 

We also only need 1 of the 2 examples I've presented here. 

So we could just go with the `if` block. Or anything else that you can think of that makes a more realistic example. 

All I am saying here is that when you type this into an editor, this describes how Rubocop will alert to it.

corsonknowles: Update: I was leaning into it being obvious bad code before. I got the behavior here down to using a ternary, and it is valid RSpec. My only concern is that it's slightly less obvious how problematic that is!

bquorning:
While looking at these PRs to improve our test/code coverage, I have been going a bit back and forth on whether 100% coverage should be a goal in itself. We should at least take care not to add nonsensical test cases, just to achieve better coverage. I think such tests may become a source of confusion for developers fixing another edge case, perhaps 6 months or 2 years from now. I think it’s better to look back through the commit history, find out why a particular piece of code was added, and use that information to inspire a more realistic test case.

On the other hand, having (and enforcing) a high coverage percentage will ensure that no new code is added without having a meaningful test case to complement it. It will be harder to introduce code that is not well explained by its tests. So yeah, let’s continue increasing the coverage toward the 100% 👍🏼

That said, even with your added test case, you could delete `next unless ancestor.send_type?` from the lib/ code and all specs still pass. I think because the type is also checked by the `expect_to?` node matcher.

corsonknowles:
Thanks @bquorning, I'm happy to close this PR if you want to delete that line instead!

corsonknowles:
Although, it might still be worth documenting this. I.e. that one cannot escape the Cop here by adding layers of nodes inside the `.to` clause.  Do you want to do both?

bquorning:
Yeah, let's do both. In one PR?

corsonknowles:
@aarestad Thoughts on this discovery? Are we missing anything here?

I see you added both guard clauses here:
* https://github.com/rubocop/rubocop-rspec/pull/1960

We now think only the 1st is needed, but let me know if I'm missing anything!

lib/rubocop/cop/rspec/unspecified_exception.rb
```diff
@@ -64,7 +64,6 @@ def empty_exception_matcher?(node)
         def find_expect_to(node)
           node.each_ancestor.find do |ancestor|
             break if ancestor.block_type?
-            next unless ancestor.send_type?
```
corsonknowles, line 67: The other thing is, does this give us any reasonable performance benefit?

spec/rubocop/cop/rspec/unspecified_exception_spec.rb
```diff
@@ -243,15 +243,10 @@ def raise_error
       RUBY
     end
 
-    it 'allows a subject function to be named raise_error' do
```
bquorning, line 246: Let’s keep this test case.

bquorning (approved):

spec/rubocop/cop/rspec/unspecified_exception_spec.rb
corsonknowles: I'll put it back. Just so you know, I replaced it because it would be very hard to break this spec without also breaking the spec above it. You'd have to specifically add special handling to do so for only one of the method names, which are not referenced outside `RESTRICT_ON_SEND`

bquorning (approved):

spec/rubocop/cop/rspec/unspecified_exception_spec.rb
bquorning: Ah, I see. Anyway, let’s keep both anyway.

PR #1992, opened 2024-10-30
Test derivative `case style` for `BeNil` cop

corsonknowles:
I was worried all the case statements with missing branch coverage would be like this, but some other ones at least are actually reachable code.

Solves for:

Screenshot 2024-10-30 at 2 45 27 PM

- [ ] Added tests.
- [ ] Updated documentation.


corsonknowles:
We should probably write up an Issue on the various causes of missing branch coverage, as well as the edge cases that we've solved for so far by filling out branch coverage. 

As a reminder, having full branch coverage allows us to require full branch coverage, which could be quite useful as a standard moving forward to mitigate the introduction of unintended behaviors.

PR #1994, opened 2024-11-02
Add an else condition to correct `StubbedMock` behavior

corsonknowles:
Fix edge case for `StubbedMock` where it could return a blank suggestion for unhandled Expectation types.

Completes branch coverage for this file.

- [x] Added tests.
- [ ] Updated documentation.

### DIFF
--- a/lib/rubocop/cop/rspec/stubbed_mock.rb
+++ b/lib/rubocop/cop/rspec/stubbed_mock.rb
@@ -14,8 +14,9 @@ module RuboCop
       #   expect(foo).to receive(:bar).with(42)
       #
       class StubbedMock < Base
-        MSG = 'Prefer `%<replacement>s` over `%<method_name>s` when ' \
+        MSG = 'Prefer %<replacement>s over `%<method_name>s` when ' \
               'configuring a response.'
+        RESTRICT_ON_SEND = %i[to].freeze
 
         # @!method message_expectation?(node)
         #   Match message expectation matcher
@@ -133,8 +134,6 @@ module RuboCop
           }
         PATTERN
 
-        RESTRICT_ON_SEND = %i[to].freeze
-
         def on_send(node)
           expectation(node) do |expectation, method_name, matcher|
             on_expectation(expectation, method_name, matcher)
@@ -155,19 +154,23 @@ module RuboCop
         end
 
         def msg(method_name)
-          format(MSG,
-                 method_name: method_name,
-                 replacement: replacement(method_name))
+          format(
+            MSG,
+            method_name: method_name,
+            replacement: replacement(method_name)
+          )
         end
 
         def replacement(method_name)
           case method_name
           when :expect
-            :allow
+            '`allow`'
           when :is_expected
-            'allow(subject)'
+            '`allow(subject)`'
           when :expect_any_instance_of
-            :allow_any_instance_of
+            '`allow_any_instance_of`'
+          else
+            'an allow statement'
           end
         end
       end
--- a/spec/rubocop/cop/rspec/stubbed_mock_spec.rb
+++ b/spec/rubocop/cop/rspec/stubbed_mock_spec.rb
@@ -126,12 +126,19 @@ RSpec.describe RuboCop::Cop::RSpec::StubbedMock do
     RUBY
   end
 
-  it 'tolerates passed arguments without parentheses' do
+  it 'flags even when passed arguments without parentheses' do
     expect_offense(<<~RUBY)
       expect(Foo)
       ^^^^^^^^^^^ Prefer `allow` over `expect` when configuring a response.
         .to receive(:new)
         .with(bar).and_return baz
+    RUBY
+  end
+
+  it 'flags `are_expected`' do
+    expect_offense(<<~RUBY)
+      are_expected.to receive(:bar).and_return(:baz)
+      ^^^^^^^^^^^^ Prefer an allow statement over `are_expected` when configuring a response.
     RUBY
   end
 end


corsonknowles:
Hmm, I think this spec passes both before and after the change, so there is some issue here:

```ruby
  it 'flags `are_expected`' do
    expect_offense(<<~RUBY)
      are_expected.to receive(:bar).and_return(:baz)
      ^^^^^^^^^^^^ Prefer `` over `are_expected` when configuring a response.
    RUBY
  end
```

lib/rubocop/cop/rspec/stubbed_mock.rb
```diff
@@ -157,18 +162,7 @@ def on_expectation(expectation, method_name, matcher)
...
-        def replacement(method_name)
-          case method_name
-          when :expect
-            :allow
-          when :is_expected
-            'allow(subject)'
-          when :expect_any_instance_of
-            :allow_any_instance_of
-          end
+                 replacement: REPLACEMENTS[method_name])
```
bquorning, line 165: Simplecov would complain if we used `#fetch` here, right?

corsonknowles: Simplecov is fine with using `fetch` -- it's slightly less performant than simple lookup. 

And without a fallback value it raises an error, which I think we shouldn't do without error handling. 

Not sure how to proceed here.

corsonknowles:
Okay, I pushed up a real change -- now it only flags on the methods it can correct.

Thoughts?

spec/rubocop/cop/rspec/stubbed_mock_spec.rb
```diff
@@ -126,12 +126,49 @@
...
+     skip: 'This passed before this commit. Handled like is_expected?' do
+    expect_offense(<<~RUBY)
+      are_expected.to receive(:bar).and_return(:baz)
+      ^^^^^^^^^^^^ Prefer `` over `are_expected` when configuring a response.
+    RUBY
+  end
+
+  it 'flags `should`', skip: 'Not implemented yet' do
+    expect_offense(<<~RUBY)
+      foo.should receive(:bar).and_return(:baz)
```
pirj, line 163: RSpec 2 legacy, deprecated a decade ago and removed in next major RSpec. We don’t want to support this.

spec/rubocop/cop/rspec/stubbed_mock_spec.rb
```diff
@@ -126,12 +126,49 @@
...
+    expect_offense(<<~RUBY)
+      foo.should_not_receive(:bar)
+      ^^^^^^^^^^^^^^^^^^^^^^ Prefer `allow(...).not_to receive` over `should_not_receive` when configuring a response.
+    RUBY
+  end
+
+  it 'flags `are_expected`',
+     skip: 'This passed before this commit. Handled like is_expected?' do
+    expect_offense(<<~RUBY)
+      are_expected.to receive(:bar).and_return(:baz)
```
pirj, line 156: Is this from rspec-its? It’s soft-maintained, and is unlikely can be named an integral part of RSpec. I’d rather documented how to add it to the co figuration if the cop was configurable.

lib/rubocop/cop/rspec/stubbed_mock.rb
```diff
@@ -157,18 +162,7 @@ def on_expectation(expectation, method_name, matcher)
         def msg(method_name)
           format(MSG,
                  method_name: method_name,
-                 replacement: replacement(method_name))
-        end
-
-        def replacement(method_name)
-          case method_name
```
pirj, line 164: I would prefer a `case`.
Why not?

bquorning: SimpleCov pointed out that we don’t have any test coverage for when `method_name` didn’t match any of the options. And correctly so.

I also prefer the `case` statement, but perhaps we should add an `else raise ArgumentError` or `KeyError` or similar, and add test coverage (where relevant).

pirj: Ha! We can actually [configure expectations](https://github.com/rubocop/rubocop-rspec/blob/5d96600242e849526a2aaa6e35ecf16c2982ff87/lib/rubocop/rspec/language.rb#L129) DSL, and add eg the mentioned `are_expected`, for which this cop won’t find a substitute.

But there is no substitute for eg are_expected. Can we skip correction in this case instead of throwing an argument error?

`fetch` would raise, and this is also an undesirable outcome, and a breaking change.

I was never a big fan of 100% coverage, but it really starts to manifest its benefits here!

pirj: Correcting myself. Dsl conf [link](https://github.com/rubocop/rubocop-rspec/blob/5d96600242e849526a2aaa6e35ecf16c2982ff87/config/default.yml#L62). are_expected, and should_* are included.

Quick suggestion: keep the `case`, but limit matching to just the three options we have substitutes for. Wdyt?

bquorning: > Quick suggestion: keep the `case`, but limit matching to just the three options we have substitutes for. Wdyt?

That is what we had before, no? Should we add an `else` case, and how should it respond?

pirj: Previously I think it was returning nil, and we auto-corrected to nonsense. Now, I think, it should just raise an offence and don’t autocorrect

spec/rubocop/cop/rspec/stubbed_mock_spec.rb
corsonknowles: Yes: https://github.com/search?q=org%3Arspec%20are_expected&type=code

My understanding is that it is included in the `Expectations.all` that the current Cop is using.

lib/rubocop/cop/rspec/stubbed_mock.rb
corsonknowles: > Now, I think, it should just raise an offence and don’t autocorrect

I'm not exactly sure how to accomplish this. 

Would you like me to close this PR so you can take it a new direction?

Or I'm happy to continue helping out here if you are willing to coach me through it.

pirj: My bad - this cop does not autocorrect 🙈

Since there’s no valid suggestion for the message:
1 add a new MSG_NO_RePLACEMENT with a different message
2 use it for anything that doesn’t have a built-in counterpart

I insist on keeping the case statement, as ‘fetch’ may conceal execution paths from our coverage detection

corsonknowles:
Okay, how's this approach?

bquorning (approved):

pirj:
Perfect, thank you, @corsonknowles !

PR #1998, opened 2024-11-20
Use Node extensions instead of *splat

bquorning:
Use more rubocop-ast and less `*` splatting.

- [ ] Added tests.
- [ ] Updated documentation.

### DIFF
--- a/lib/rubocop/cop/rspec/single_argument_message_chain.rb
+++ b/lib/rubocop/cop/rspec/single_argument_message_chain.rb
@@ -67,11 +67,10 @@ module RuboCop
         end
 
         def autocorrect_hash_arg(corrector, arg)
-          key, value = *arg.children.first
-
-          corrector.replace(arg, key_to_arg(key))
+          pair = arg.pairs.first
+          corrector.replace(arg, key_to_arg(pair.key))
           corrector.insert_after(arg.parent.loc.end,
-                                 ".and_return(#{value.source})")
+                                 ".and_return(#{pair.value.source})")
         end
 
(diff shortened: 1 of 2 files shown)


pirj (approved):
Thank you!

pirj:
With such a coverage now it's almost a blind approve 😄

bquorning:
@ydah It looks like mdformat checks are broken, probably related to https://github.com/executablebooks/mdformat/issues/480. Do you want to have a look at it?

lib/rubocop/cop/rspec/single_argument_message_chain.rb
```diff
@@ -67,11 +67,9 @@ def single_element_array?(node)
         end
 
         def autocorrect_hash_arg(corrector, arg)
-          key, value = *arg.children.first
-
-          corrector.replace(arg, key_to_arg(key))
+          corrector.replace(arg, key_to_arg(arg.keys.first))
```
Darhazer, line 70: how about
```ruby
pair = arg.pairs.first
```
and then using `key_to_arg(pair.key)` and `pair.value.source` respectively ?

Darhazer (commented):
Always good to see more node specific methods usage and less deconstruction

Darhazer (approved):

lib/rubocop/cop/rspec/single_argument_message_chain.rb
bquorning: Yep, that’s even better. Thank you.

PR #2014, opened 2024-12-24
Fix a false negative for `RSpec/PendingWithoutReason` cop when pending/skip without reason inside a numblock in a deeply nested example

ydah:
The following will cause a violation to be generated for skip/pending without reason inside examples.

```ruby
RSpec.describe Foo do
  context 'when something' do
    it 'does something' do
      _1
      skip
      ^^^^ Give the reason for skip.
      pending
      ^^^^^^^ Give the reason for pending.
      _1
    end
  end
end
```

- [x] Added tests.
- [-] Updated documentation.

### DIFF
--- a/spec/rubocop/cop/rspec/pending_without_reason_spec.rb
+++ b/spec/rubocop/cop/rspec/pending_without_reason_spec.rb
@@ -301,21 +301,43 @@ RSpec.describe RuboCop::Cop::RSpec::PendingWithoutReason, :ruby27 do
       RUBY
     end
 
-    context 'with a numblock' do
+    context 'with a numblock at the top level' do
       it 'registers offense' do
         expect_offense(<<~RUBY)
           RSpec.describe Foo do
+            _1
             pending
             ^^^^^^^ Give the reason for pending.
             skip
             ^^^^ Give the reason for skip.
             _1
+          end
+        RUBY
+      end
+    end
+
+    context 'with a numblock in a nested context' do
+      it 'registers offense' do
+        expect_offense(<<~RUBY)
+          RSpec.describe Foo do
             context 'when something' do
               _1
               pending
               ^^^^^^^ Give the reason for pending.
               skip
               ^^^^ Give the reason for skip.
+              _1
+            end
+          end
+        RUBY
+      end
+    end
+
+    context 'with a numblock in a deeply nested example' do
+      it 'registers offense' do
+        expect_offense(<<~RUBY)
+          RSpec.describe Foo do
+            context 'when something' do
               it 'does something' do
                 _1
                 skip
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Master (Unreleased)
 
+- Fix a false negative for `RSpec/PendingWithoutReason` cop when pending/skip without reason inside a numblock in a deeply nested example. ([@ydah])
+
 ## 3.3.0 (2024-12-12)
 
 - Deprecate `top_level_group?` method from `TopLevelGroup` mixin as all of its callers were intentionally removed from `Rubocop/RSpec`. ([@corsonknowles])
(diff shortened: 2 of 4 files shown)


spec/rubocop/cop/rspec/pending_without_reason_spec.rb
```diff
@@ -301,21 +301,43 @@
       RUBY
     end
 
-    context 'with a numblock' do
```
ydah, line 304: The reason for the split is that these would be syntax errors for prism and parse.y.

parse.y
```
❯ ruby --parser=parse.y -v -e '
[1].each do
  _1
  [1].each do
    _1
    [1].each do
      _1
      2
      _1
    end
  end
end
'
ruby 3.4.0rc1 (2024-12-12 master 29caae9991) [arm64-darwin23]
-e:5: numbered parameter is already used in
-e:3: outer block here
-e:7: numbered parameter is already used in
-e:3: outer block here
-e:9: numbered parameter is already used in
-e:3: outer block here
-e:8: warning: possibly useless use of a literal in void context
-e:3: warning: possibly useless use of a variable in void context
ruby: compile error (SyntaxError)
```

Prism:
```
❯ ruby --parser=prism -v -e '
[1].each do
  _1
  [1].each do
    _1
    [1].each do
      _1
      2
      _1
    end
  end
end
'
ruby 3.4.0rc1 (2024-12-12 master 29caae9991) +PRISM [arm64-darwin23]
-e:7: warning: possibly useless use of a variable in void context
-e:8: warning: possibly useless use of a literal in void context
-e:5: warning: possibly useless use of a variable in void context
-e:3: warning: possibly useless use of a variable in void context
-e: -e:7: syntax errors found (SyntaxError)
  3 |   _1
  4 |   [1].each do
> 5 |     _1
    |     ^~ numbered parameter is already used in outer block
  6 |     [1].each do
> 7 |       _1
    |       ^~ numbered parameter is already used in outer block
  8 |       2
> 9 |       _1
    |       ^~ numbered parameter is already used in outer block
  10 |     end
  11 |   end
```

pirj: But if this is in the code, it won't even run, will it? Should we account for that?

ydah: Each split code does not result in a syntax error. Also, the reason I had all these in one example was probably to test for the unexplained pending/skip in various places. And I found this false negative when I was splitting this test, so I'm assuming the test makes sense, but what do you think?

pirj: Even though I acknowledge the presence of a Ruby feature that adds `_1` (and, now in 3.4, `it`), but unless I'm missing something, that feature in the context of specs it has limited or no use.

To me, Prism's error looks sensible. Why would anyone use just the `_1`, not `_1.description` or something like this? 🤯 
And how calling the plain `_1` prior to `pending` fixes all this? 🤔 

But the problem I see here is different - it's the elevated cognitive load of us trying to keep those examples up-to-date.
I'd rather remove those examples with numblock rather than having a maintenance headache.

WDYT?

ydah: Um, just to note this as an aid in understanding, the Prism and parse.y error suggests that the nested block only allows one numbered parameter.

For example, the following case is OK

```ruby
[1].each { put _1; [1].each { put 'ok' }
[1].each { put 'ok'; [1].each { put _1 }
```

The following cases result in syntax errors
````ruby
[1].each { put _1; [1].each { put _1 }
```

This is indicated by the error message.
```ruby
❯ ruby -e '[1].each { put _1; [1].each { put _1 }
-e: -e:1: syntax error found (SyntaxError)
> 1 | ... _1 }
    | ^~ numbered parameter is already used in outer block
```

ydah: However, I reconsidered and change my mind to just remove these examples. Create another PR.

ydah: - https://github.com/rubocop/rubocop-rspec/pull/2017

PR #2017, opened 2024-12-25
Remove an example which is a syntax error in Ruby's parser

ydah:
This code generates a syntax error as follows:
```ruby
❯ ruby -e '[1].each { put _1; [1].each { put _1 } }'
-e: -e:1: syntax error found (SyntaxError)
> 1 | ... _1 } }
    |     ^~ numbered parameter is already used in outer block
```

When numbered parameters are used in nested blocks, it is not possible to use numbered parameters in multiple different hierarchies.

While it was considered to split these up, we discussed below. 
https://github.com/rubocop/rubocop-rspec/pull/2014#discussion_r1897371078

- [ ] Added tests.
- [ ] Updated documentation.


pirj (approved):
Thank you! ❤️

PR #2020, opened 2025-01-08
Add new `RedundantContext` cop

### LINKED ISSUE #2001: New cop to discourage `context` with only a single test
Relevant to #2000.

When the `context` block has only a single test, it makes little sense to actually use the `context` block and making the code unnecessary indented and a little more complicated.

```ruby
# bad
context 'when condition' do
  it 'tests something' do
  end
end

# good
it 'tests something when condition' do
end
```

tejasbubane:
To check context with single example.

Closes https://github.com/rubocop/rubocop-rspec/issues/2001

- [x] Added tests.
- [x] Updated documentation.


tejasbubane:
Seems like we have a bunch occurrences where we use single example inside context.

koic:
It’s doubtful whether this should be regarded as "redundant" or as an appropriate use of vocabulary, as it depend on individual judgment. The justification provided in #2001 doesn’t seem entirely sufficient, and at the least, this doesn’t appear to be a rule that should be enabled by default.

bquorning:
I like this cop. I have often manually changed specs like

```rb
context "when foo" do
  it "does bar" do
  end
end
```

into

```rb
it "does bar when foo" do
end
```

but I agree that it should probably be disabled by default.

lib/rubocop/cop/rspec/redundant_context.rb
```diff
@@ -0,0 +1,39 @@
...
+      #   it 'tests something when condition' do
+      #   end
+      #
+      class RedundantContext < Base
+        MSG = 'Redundant context with single example.'
+
+        # @!method redundant_context?(node)
+        def_node_matcher :redundant_context?, <<~PATTERN
+          (block
+            (send #rspec? :context _)
```
pirj, line 25: Just specifically `context` and `it`?
I’dd love to have at least balancing examples for describe/example aliases if it’s a deliberate design choice.

lib/rubocop/cop/rspec/redundant_context.rb
```diff
@@ -0,0 +1,39 @@
...
+            _
+            (block (send _ :it ...) ...))
```
pirj, line 27: I think the receiver should be nil here, not _

lib/rubocop/cop/rspec/redundant_context.rb
```diff
@@ -0,0 +1,39 @@
...
+        PATTERN
+
+        def on_block(node)
+          return unless redundant_context?(node)
+
+          add_offense(node)
+        end
+        alias on_numblock on_block
```
pirj, line 35: Will a numblock match with the current node pattern?

spec/rubocop/cop/rspec/redundant_context_spec.rb
```diff
@@ -0,0 +1,25 @@
...
+    expect_offense(<<~RUBY)
+      context 'when condition' do
+      ^^^^^^^^^^^^^^^^^^^^^^^^^^^ Redundant context with single example.
+        it 'does something' do
+        end
+      end
+    RUBY
+  end
+
+  it 'does not register offense when multiple examples inside context' do
```
pirj, line 14: It is more fair to say “when the example is the only statement inside the example group. Eg another it_behaves_like or a method def would cause the cop to ignore the group.

Are there any cases with more than one statement (but one example) that we still want to flag/correct?
If not - just ignore this note, just make the description more generic, and maybe add another example with an example and some other statement in an example group.

config/default.yml
```diff
@@ -794,6 +794,12 @@ RSpec/RedundantAround:
   VersionAdded: '2.19'
   Reference: https://www.rubydoc.info/gems/rubocop-rspec/RuboCop/Cop/RSpec/RedundantAround
 
+RSpec/RedundantContext:
+  Description: Detect redundant `context` hook.
+  Enabled: pending
```
pirj, line 799: I’m usually strongly against disabling cops, but this one, even though certainly serving some projects that strive to enforce this style, isn’t a good fit for the majority.

I support what @koic and @bquorning said.

pirj:
Fair enough, @ydah 
Btw be_empty is a nice one, but it’s probably unsafe for cases when the target is not an array.

@bquorning wdyt?

pirj:
Closing as stale.

PR #2025, opened 2025-01-27
Use node groups in node patterns to replace unions of types

dvandersluis:
`rubocop-ast` defines some node groups (https://github.com/rubocop/rubocop-ast/blob/85bfe84/lib/rubocop/ast/node.rb#L89-L116) that can be used in place of a union of node types.

rubocop-ast v1.38.0 is needed for `any_block`. I added it to the gemspec in this PR but if you'd prefer to handle it later, I can remove the `any_block` changes.

### DIFF
--- a/lib/rubocop/cop/rspec/be_eql.rb
+++ b/lib/rubocop/cop/rspec/be_eql.rb
@@ -45,7 +45,7 @@ module RuboCop
 
         # @!method eql_type_with_identity(node)
         def_node_matcher :eql_type_with_identity, <<~PATTERN
-          (send _ :to $(send nil? :eql {true false int float sym nil}))
+          (send _ :to $(send nil? :eql {boolean int float sym nil}))
         PATTERN
 
         def on_send(node)
--- a/lib/rubocop/cop/rspec/be_eq.rb
+++ b/lib/rubocop/cop/rspec/be_eq.rb
@@ -31,7 +31,7 @@ module RuboCop
 
         # @!method eq_type_with_identity?(node)
         def_node_matcher :eq_type_with_identity?, <<~PATTERN
-          (send nil? :eq {true false nil})
+          (send nil? :eq {boolean nil})
         PATTERN
 
         def on_send(node)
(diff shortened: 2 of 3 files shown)


dvandersluis:
CI failures are unrelated to this PR but due to https://github.com/rubocop/rubocop/pull/13719.

bquorning:
Yeah, I think v1.38.0 is a bit too recent for us to set as a hard requirement. Postponing the use of `any_block`, which version of rubocop-ast is then required?

bquorning:
> CI failures are unrelated to this PR but due to [rubocop/rubocop#13719](https://github.com/rubocop/rubocop/pull/13719).

The CI failures are fixed in #2026.

dvandersluis:
@bquorning I updated the PR to skip `{block numblock}` => `any_block` for now. It appears that the version you use as "Oldest RuboCop" is 1.61, which shouldn't have a problem handling the other node group types.

dvandersluis:
> The CI failures are fixed in #2026.

Thank you! I'll rebase once it's been merged.

lib/rubocop/cop/rspec/be_eql.rb
```diff
@@ -45,7 +45,7 @@ class BeEql < Base
 
         # @!method eql_type_with_identity(node)
         def_node_matcher :eql_type_with_identity, <<~PATTERN
-          (send _ :to $(send nil? :eql {true false int float sym nil}))
+          (send _ :to $(send nil? :eql {boolean int float sym nil}))
```
pirj, line 48:
```suggestion
          (send _ :to $(send nil? :eql {boolean numeric sym nil}))
```
This needs to be checked though, and the doc above needs to be adjusted.

I don’t mind leaving this out

pirj (approved):
Thanks!

pirj:
> Oldest RuboCop" is 1.61

we can update that together with bumping the dependency in the gemspec.

lib/rubocop/cop/rspec/be_eql.rb
dvandersluis: `numeric` is more than `int` and `float`, so up to you if you want to change behaviour. I was only making equivalent changes 😄

bquorning:
> Yeah, I think v1.38.0 is a bit too recent for us to set as a hard requirement. Postponing the use of `any_block`, which version of rubocop-ast is then required?

That was 3 weeks ago, and now I just merged #2042 which adds that dependency 😅 Would you mind rebasing this branch @dvandersluis, then I think we can get it into the next release.

PR #2026, opened 2025-01-27
Update .rubocop.yml, and disable OnSendWithoutOnCSend

bquorning:
A new cop InternalAffairs/OnSendWithoutOnCSend was added to RuboCop in https://github.com/rubocop/rubocop/pull/13719.

cc @dvandersluis


pirj:
My internal parser broke on `arg?.(1)`. What is a csend?

dvandersluis:
`csend` is safe navigation (`foo&.bar`).

I haven't looked through all the cops in here but it might be that the `OnSendWithoutOnCSend` just doesn't make sense for rubocop-rspec and could be wholesale disabled. I don't think there's a lot of instances where safe navigation would be used in tests, and anywhere that you're calling a method on a constant or as a macro doesn't really need safe navigation support.

bquorning:
I decided to follow @dvandersluis’ advice and outright disable the OnSendWithoutOnCSend cop.

I have update the PR to also include enabling of all pending cops up until the latest version.

ydah (approved):

PR #2027, opened 2025-02-05
Mark `RSpec/LetSetup` as unsafe

r7kamura:
For the reasons mentioned, I thought it might be better to consider this cop as unsafe.

For example, in cases like the following, `let!(:foo)` is used to override the original definition, and there is no code directly referencing `foo` within the surrounding AST. However, I believe this usage of `let!` is valid because it is difficult to substitute this with other methods, such as `before` or `let`.

```ruby
let!(:foo) { ... }

...

context 'when foo is something' do
  let!(:foo) { ... }

  it { is_expected.to be false }
end
```

What do you think?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Master (Unreleased)
 
+- Mark `RSpec/LetSetup` as unsafe. ([@r7kamura])
+
 ## 3.4.0 (2025-01-20)
 
 - Fix `RSpec/SortMetadata` cop to limit sorting to trailing metadata arguments. ([@cbliard])
--- a/config/default.yml
+++ b/config/default.yml
@@ -606,7 +606,9 @@ RSpec/LetBeforeExamples:
 RSpec/LetSetup:
   Description: Checks unreferenced `let!` calls being used for test setup.
   Enabled: true
+  Safe: false
   VersionAdded: '1.7'
+  VersionChanged: "<<next>>"
   Reference: https://www.rubydoc.info/gems/rubocop-rspec/RuboCop/Cop/RSpec/LetSetup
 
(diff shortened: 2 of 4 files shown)


bquorning:
I guess you are correct :+1:

@pirj What do you think?

pirj:
Can't the inner `let!` be a regular `let`?

A full-fledged example just for this case. I can't reproduce a difference when I change the inner `let!` to `let`.

```ruby
RSpec.describe 'A' do
  subject(:acc) { [] }

  def add(value)
    acc.unshift(value)
    value
  end

  let!(:foo) { add('original') }

  it { is_expected.to contain_exactly('original') }

  context 'when foo is something' do
    let(:foo) { add('overridden') }

    it { is_expected.to contain_exactly('overridden') }
  end
end
```

I agree that we can stretch our reasoning further, in such a way that the outer `let!` can be removed during a spec refactoring, and now specs fully rely on`let!(:foo)`s defined in several contexts that now each define them separately. This would break if they were written as non-bang `let`s.
My answer would be - use balancing examples.



Another thing is that I don't see `:foo` used anywhere in the provided example, and I don't see a good reason not to replace it with just a `before`.
Please provide an example of a spec where it is used, so we can reason about it without making potentially different mental images of the discussed subject.



So far, I'm not convinced at all that we should mark the cop as unsafe.

r7kamura:
It's true that in most situations, this kind of `let!` can be replaced by `let`. And I agree that I should provide an example that is easy to discuss.

One Rails project I work on has an example of complex test code where `let!` cannot be replaced by `let`, but it is difficult to cut it down to the smallest example, and this is an extremely rare situation...

I am going to withdraw this pull request for now until I can cut out a good example code for this. Thanks for your good input 🙏

PR #2030, opened 2025-02-09
Finish branch coverage for `RSpec/DescribedClass`

corsonknowles:
Finish branch coverage for `RSpec/DescribedClass`

Showing that the tested code evaluates:
```ruby
zenpayroll(dev)* class MyClass3
zenpayroll(dev)> end
=> nil
zenpayroll(dev)> ("MyClass" + "3").constantize
=> MyClass3

zenpayroll(dev)* class MyClass3; class MyClass; end;
zenpayroll(dev)> end
=> nil
zenpayroll(dev)> (("MyClass" + "3").constantize)::MyClass
=> MyClass3::MyClass
```

Although at this point the math is superfluous -- `(method_that_returns_a_class)::MyClass` will cover this branch while `method_that_returns_a_class::MyClass`  will not

Before:
Screenshot 2025-02-08 at 10 25 14 PM
After:
Screenshot 2025-02-08 at 10 25 31 PM

- [x] Added tests.
- [ ] Updated documentation.

### DIFF
--- a/spec/rubocop/cop/rspec/described_class_spec.rb
+++ b/spec/rubocop/cop/rspec/described_class_spec.rb
@@ -338,6 +338,14 @@ RSpec.describe RuboCop::Cop::RSpec::DescribedClass do
         RUBY
       end
     end
+
+    it 'ignores parenthesized namespaces - begin_types' do
+      expect_no_offenses(<<~RUBY)
+        describe MyClass do
+          subject { (MyNamespace)::MyClass }
+        end
+      RUBY
+    end
   end
 
   context 'when EnforcedStyle is :explicit' do


bquorning:
`(1 + 2)::MyClass` generates invalid Ruby, since a module cannot be named `3`. Can we can come up with a valid example that hits the `else` branch? Otherwise, if the code is unreachable, we can tell SimpleCov to ignore it with
```
          else
            # :nocov:
            nil
            # :nocov:
```

corsonknowles:
Thanks @bquorning -- it works with a little magic sprinkled in, and it documents the behavior I think.

corsonknowles:
It's possible there's another approach here to handle the `s(:begin,` block this example creates in the AST, but I don't know how it should handle that.

spec/rubocop/cop/rspec/described_class_spec.rb
```diff
@@ -338,6 +338,14 @@ module SomeGem
         RUBY
       end
     end
+
+    it 'ignores a class with a dynamic namespace' do
+      expect_no_offenses(<<~RUBY)
+        describe MyClass do
+          subject { (("MyClass" + "3").constantize)::MyClass }
```
bquorning, line 345: This example still seems a bit far fetched 😅

bquorning: I’ll see in real-world-rspec if there’s any code that hits the branch you’re looking at.

corsonknowles: Not just far fetched, a little hard to find! A good puzzle. 

I buried this in the updated description above, but the way we're getting past the case statement here is just by having a `:begin` node from the parentheses. 

I replaced the example with a simpler parenthesized method call. 

The weird thing is: 
` subject { method_that_returns_a_class::MyClass }` 
does get handled. 

I'm not certain how `method_that_returns_a_class` or `(method_that_returns_a_class)` ought to be corrected, but maybe they should be handled identically.

bquorning: `subject { method_that_returns_a_class::MyClass }` makes `namespace` a send type, whereas `subject { (method_that_returns_a_class)::MyClass }` makes `namespace` a begin type. So maybe the test case we want is just the existing `foo::CONST` one, but with parentheses around `foo`?

bquorning:
Btw I just found that changing line 85 from `let(:baz) { foo::CONST }` to `let(:baz) { @foo::CONST }` will raise an exception. Not sure if that is related at all, or should be a separate issue.

corsonknowles:
> Btw I just found that changing line 85 from let(:baz) { foo::CONST } to let(:baz) { @foo::CONST } will raise an exception. Not sure if that is related at all, or should be a separate issue.

Great find! Separate issue, but another perk of the exercise.

bquorning (commented):
Don’t forget to update .simplecov ✨

corsonknowles:
I didn’t want to start a conflict cascade. I think we can change it to 100%
when all of your PRs are merged?

bquorning:
There will be conflicts to resolve either way, and this PR is the first that is ready to be merged.

PR #2031, opened 2025-02-09
Config formatter coverage

bquorning:
`SUBDEPARTMENTS` was last used in ConfigFormatter before we released v3.0.0, before `RSpec/Capybara` was extracted. It’s being removed in this PR.

Also, initially wanting to improve spec branch coverage for ConfigFormatter, ~I found that `AMENDMENTS` was defined as a String, not an Array. And splatting the array into `#gsub` doesn't work if there is more than one element - we need to use `Regexp.union` instead.~ Actually I ended up removing `AMENDMENDS` altogether.

To avoid conflicts with other PRs improving code coverage, the .simplecov file is not updated here.

- [x] Added tests.
- [ ] Updated documentation.

### DIFF
--- a/spec/rubocop/rspec/config_formatter_spec.rb
+++ b/spec/rubocop/rspec/config_formatter_spec.rb
@@ -8,6 +8,12 @@ RSpec.describe RuboCop::RSpec::ConfigFormatter do
       'AllCops' => {
         'Setting' => 'forty two'
       },
+      'Metrics/BlockLength' => {
+        'Exclude' => [
+          '**/*_spec.rb',
+          '**/spec/**/*'
+        ]
+      },
       'RSpec/Foo' => {
         'Config' => 2,
         'Enabled' => true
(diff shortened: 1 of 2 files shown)


bquorning:
cc @corsonknowles

corsonknowles:
Thanks! I was about to raise an issue on this because of the String/Array mismatch. Glad you knew how to fix it!

corsonknowles (approved):

spec/rubocop/rspec/config_formatter_spec.rb
```diff
@@ -8,6 +8,12 @@
       'AllCops' => {
         'Setting' => 'forty two'
       },
+      'Metrics/BlockLength' => {
```
corsonknowles, line 11: Nice the spec explains the behavior now!

pirj (approved):
Thank you!

It was such a pain to make this work. I'm happy we don't need this code anymore.

image

PR #2032, opened 2025-02-09
Add eql/equal support to RSpec/PredicateMatcher

bquorning:
The `be_bool?` node matcher was already finding nodes using `eql` and `equal`, but then didn't properly handle them.

- [x] Added tests.
- [ ] Updated documentation.

### DIFF
--- a/lib/rubocop/cop/rspec/predicate_matcher.rb
+++ b/lib/rubocop/cop/rspec/predicate_matcher.rb
@@ -112,7 +112,7 @@ module RuboCop
 
         def true?(to_symbol, matcher)
           result = case matcher.method_name
-                   when :be, :eq
+                   when :be, :eq, :eql, :equal
                      matcher.first_argument.true_type?
                    when :be_truthy, :a_truthy_value
                      true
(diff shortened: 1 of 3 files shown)


bquorning:
cc @corsonknowles

corsonknowles (approved):
Great find! Another win for coverage.

lib/rubocop/cop/rspec/predicate_matcher.rb
```diff
@@ -106,7 +106,7 @@ def rewrite_matcher(corrector, predicate, matcher)
 
         def true?(to_symbol, matcher)
           result = case matcher.method_name
-                   when :be, :eq
+                   when :be, :eq, :eql, :equal
```
pirj, line 109: A few aliases to make this complete? https://github.com/rspec/rspec/blob/a80119c703d38f993c414490abe8c992154cf5d1/rspec-expectations/lib/rspec/matchers.rb#L578

I just went there just in case, I don't think I've seen those used once in my practice 🙈  So really up to you.

pirj (approved):
Thank you!

lib/rubocop/cop/rspec/predicate_matcher.rb
bquorning: Yeah, I have never seen those used either. I think there’s no need to add them now.

ydah (approved):

PR #2033, opened 2025-02-09
Refactor RSpec/Focus, increase branch coverage

bquorning:
The corrector logic for pair_type/str_type/sym_type seems like a "generic" implementation, so I move things around to let that be a "default" case in the if/else block.

To avoid conflicts with other PRs improving code coverage, the .simplecov file is not updated here.

- [ ] Added tests.
- [ ] Updated documentation.

### DIFF
--- a/lib/rubocop/cop/rspec/focus.rb
+++ b/lib/rubocop/cop/rspec/focus.rb
@@ -75,23 +75,27 @@ module RuboCop
         def on_send(node)
           return if node.chained? || node.each_ancestor(:def, :defs).any?
 
-          focus_metadata(node) do |focus|
-            add_offense(focus) do |corrector|
-              if focus.pair_type? || focus.str_type? || focus.sym_type?
-                corrector.remove(with_surrounding(focus))
-              elsif focus.send_type?
-                correct_send(corrector, focus)
-              end
+          if focused_block?(node)
+            on_focused_block(node)
+          else
+            metadata(node) do |focus|
+              on_metadata(focus)
             end
           end
         end
 
         private
 
-        def focus_metadata(node, &block)
-          yield(node) if focused_block?(node)
+        def on_focused_block(node)
+          add_offense(node) do |corrector|
+            correct_send(corrector, node)
+          end
+        end
 
-          metadata(node, &block)
+        def on_metadata(node)
+          add_offense(node) do |corrector|
+            corrector.remove(with_surrounding(node))
+          end
         end
 
         def with_surrounding(focus)


bquorning:
cc @corsonknowles

lib/rubocop/cop/rspec/focus.rb
```diff
@@ -77,10 +77,10 @@ def on_send(node)
 
           focus_metadata(node) do |focus|
             add_offense(focus) do |corrector|
-              if focus.pair_type? || focus.str_type? || focus.sym_type?
```
pirj, line 80: Focus can't be a str_type?, or can it?

lib/rubocop/cop/rspec/focus.rb
```diff
@@ -89,9 +89,11 @@ def on_send(node)
         private
 
         def focus_metadata(node, &block)
-          yield(node) if focused_block?(node)
-
-          metadata(node, &block)
+          if focused_block?(node)
```
pirj, line 92: What if inline this method? Won't we be able to remove one "if" statement?

lib/rubocop/cop/rspec/focus.rb
bquorning: I don’t think so, no. But the code has been there forever 🤷🏼

pirj: I love the gentle idea of dropping this condition by flipping else with elsif! 😄

lib/rubocop/cop/rspec/focus.rb
bquorning: `focused_block?` is a node matcher; I’m not sure how to inline it?

pirj:
```ruby
        def on_send(node)
          return if node.chained? || node.each_ancestor(:def, :defs).any?

          if focused_block?(node)
              correct_send(corrector, node)
            end
          else
            metadata(node) do |focus|
              add_offense(focus) do |corrector|
                corrector.remove(with_surrounding(focus))
              end
            end
          end
        end
```

pirj: This causes an offence, though `Metrics/MethodLength: Method has too many lines. [12/10]`.

To me, this improves the separation of the two cases:
 - fit/fdescribe/... focused example/example groups methods
 - examples/example groups focused with `focus` metadata

It just happens that the entry point to detect them is the same, `on_send`.

For those two cases:
1. node pattern to detect is different
2. the selection for offence is different
3. the correction is different

bquorning: Brilliant. Thanks.

pirj (approved):
Just perfect! Thank you 🙌

PR #2034, opened 2025-02-09
Tell SimpleCov to ignore unreachable code

bquorning:
In some cases we know that a case statement is covering all possible branches, e.g. `case [foo?, bar?]` has four branches (if the predicate methods return booleans, of course), and `case style` has a know number of branches, depending on `SupportedStyles` in config/default.yml.

So, when to use `else ... raise ArgumentError`, and when to change the last `elsif` or `when` into an `else`? I don't know.

~To avoid conflicts with other PRs improving code coverage, the .simplecov file is not updated here.~

- [ ] Added tests.
- [ ] Updated documentation.

### DIFF
--- a/.simplecov
+++ b/.simplecov
@@ -2,7 +2,7 @@
 
 SimpleCov.start do
   enable_coverage :branch
-  minimum_coverage line: 100, branch: 98.44
+  minimum_coverage line: 100, branch: 100
   add_filter '/spec/'
   add_filter '/vendor/bundle/'
 end
--- a/lib/rubocop/cop/rspec/implicit_subject.rb
+++ b/lib/rubocop/cop/rspec/implicit_subject.rb
@@ -97,6 +97,7 @@ module RuboCop
 
         private
 
+        # rubocop:disable Metrics/MethodLength
         def autocorrect(corrector, node)
           case node.method_name
           when :expect
@@ -107,8 +108,13 @@ module RuboCop
             corrector.replace(node.location.selector, 'expect(subject).to')
           when :should_not
             corrector.replace(node.location.selector, 'expect(subject).not_to')
+          else
+            # :nocov:
+            :noop
+            # :nocov:
           end
         end
+        # rubocop:enable Metrics/MethodLength
 
         def message(_node)
           case style
@@ -119,6 +125,7 @@ module RuboCop
           end
         end
 
+        # rubocop:disable Metrics/MethodLength
         def invalid?(node)
           case style
           when :require_implicit
@@ -129,8 +136,13 @@ module RuboCop
             implicit_subject_in_non_its_and_non_single_line?(node)
           when :single_statement_only
             implicit_subject_in_non_its_and_non_single_statement?(node)
+          else
+            # :nocov:
+            :noop
+            # :nocov:
           end
         end
+        # rubocop:enable Metrics/MethodLength
 
         def implicit_subject_in_non_its?(node)
           implicit_subject?(node) && !its?(node)
--- a/lib/rubocop/cop/rspec/predicate_matcher.rb
+++ b/lib/rubocop/cop/rspec/predicate_matcher.rb
@@ -110,6 +110,7 @@ module RuboCop
           )
         end
 
+        # rubocop:disable Metrics/MethodLength
         def true?(to_symbol, matcher)
           result = case matcher.method_name
                    when :be, :eq, :eql, :equal
@@ -118,9 +119,14 @@ module RuboCop
                      true
                    when :be_falsey, :be_falsy, :a_falsey_value, :a_falsy_value
                      false
+                   else
+                     # :nocov:
+                     :noop
+                     # :nocov:
                    end
           to_symbol == :to ? result : !result
         end
+        # rubocop:enable Metrics/MethodLength
       end
 
       # A helper for `explicit` style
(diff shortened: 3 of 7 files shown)


bquorning:
cc @corsonknowles

corsonknowles:
I like the use of `else` blocks!

I agree it's a good sign to be deliberate that it's unreachable code. 

At this point, I think we should probably add a contributor guideline to the README so folks know how to handle unreachable code.  

Some ways of using node patterns and then further matching in Ruby methods and case handling for config both lend themselves to that outcome, so it will likely come up.

corsonknowles:
Also FYI, I tried `in` blocks instead of `when` blocks for exhaustive pattern matching, and although they work great, they do not change the coverage issue:
https://docs.ruby-lang.org/en/3.0/syntax/pattern_matching_rdoc.html

bquorning:
> I like the use of `else` blocks!

`else` is easy, but e.g. `when :be_falsey, :be_falsy, :a_falsey_value, :a_falsy_value` or `when [false, false]` is more informative, almost serving as a kind of documentation. Which is why in some places I kept the list of possible values as a comment, and in other places chose to add an `else` branch that just raises an exception.

bquorning:
If I remember correctly, `case`/`in` blocks _must_ have an `else` branch. Or is it just that it raises an exception if none of the branches are matched?

corsonknowles:
It raises an exception NoMatchingPatternError. This essentially handles
that part for us (instead of ArgumentError), but it's not considered
exhaustive. It also doesn't give us a convenient place to put a `# :nocov`
block.

Not considering the `case ... in` block to be exhaustive may be a failing
in Simplecov that we should look into, since pattern matching is new but
not that new.

lib/rubocop/cop/rspec/implicit_subject.rb
```diff
@@ -107,8 +108,13 @@ def autocorrect(corrector, node)
             corrector.replace(node.location.selector, 'expect(subject).to')
           when :should_not
             corrector.replace(node.location.selector, 'expect(subject).not_to')
+          else
```
pirj, line 111: Why is this needed? In practice, this is unreachable, isn't it?

lib/rubocop/cop/rspec/implicit_subject.rb
```diff
@@ -129,8 +136,13 @@ def invalid?(node)
             implicit_subject_in_non_its_and_non_single_line?(node)
           when :single_statement_only
             implicit_subject_in_non_its_and_non_single_statement?(node)
+          else
+            # :nocov:
+            raise ArgumentError("Unsupported style :#{style}")
```
pirj, line 141: Same

lib/rubocop/cop/rspec/predicate_matcher.rb
```diff
@@ -110,7 +110,7 @@ def true?(to_symbol, matcher)
                      matcher.first_argument.true_type?
                    when :be_truthy, :a_truthy_value
                      true
-                   when :be_falsey, :be_falsy, :a_falsey_value, :a_falsy_value
+                   else # :be_falsey, :be_falsy, :a_falsey_value, :a_falsy_value
```
pirj, line 113: Is this becase we don't want an "else" here?
Or because we don't want a separate test case for all four variants?
Might be we just add those four cases in specs and keep the cop intact?

lib/rubocop/cop/rspec/predicate_matcher.rb
```diff
@@ -329,6 +329,10 @@ def on_send(node)
             check_inflected(node)
           when :explicit
             check_explicit(node)
+          else
+            # :nocov:
+            raise ArgumentError("Unsupported style :#{style}")
```
pirj, line 334: This is recurring. Even if RuboCop doesn't handle the inclusion validation of [SupportedStyles](https://github.com/rubocop/rubocop-rspec/blob/2be7260d25e78c740379ea56bd3c5ef0f8e21005/config/default.yml#L764) yet, we should open a ticket for it to do this.

Or do we still need those "else" clauses? Why?

bquorning: When we have a `case` or an `if` block without an `else` branch, it will _behave_ as if there was an empty branch:
```
else
  nil
end
```
SimpleCov know this, and marks the implicit `else` branch as not covered by tests. If we want to reach 100% branch coverage, we need to either change the last `elsif condition` into an `else`, or add an `else` branch.

I think you are right in the RuboCop itself will complain if a configured style isn’t in the listed available style options.

bquorning: I agree, and will shortly push up such changes.

However I think I’ll leave the two cases of `else; node.source` which seem like reasonable defaults for those case blocks.

.simplecov
```diff
@@ -2,7 +2,7 @@
 
 SimpleCov.start do
   enable_coverage :branch
-  minimum_coverage line: 100, branch: 98.44
+  minimum_coverage line: 100, branch: 100
```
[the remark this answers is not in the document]
pirj: 🤩👏

bquorning:
@pirj 👍🏼❓

pirj (approved):
Thank you! 🤩

ydah (approved):

.simplecov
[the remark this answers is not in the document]
corsonknowles: ❤️ :shipit: 🥇

PR #2035, opened 2025-02-09
Update `CONTRIBUTING.md` to include `Simplecov`

corsonknowles:
Add Simplecov instructions, common usage, and the repo link to our `CONTRIBUTING.md` guidelines. 

- [ ] Added tests.
- [x] Updated documentation.

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -51,6 +51,37 @@ We are running [mdformat](https://github.com/executablebooks/mdformat) for forma
 $ mdformat . --number
 ```
 
+### Test Coverage - Line and Branch
+
+We are using [Simplecov](https://github.com/colszowka/simplecov) to track test coverage.
+
+It is included and reported when you run `bundle exec rake` or `bundle exec rspec`.
+
+To view the coverage report, open the `coverage/index.html` file in your browser.
+
+E.g. on macOS:
+
+```console
+$ open coverage/index.html
+```
+
+If you have unreachable lines, you can add `# :nocov` around those lines. The code itself or a comment should explain why the line is unreachable.
+
+Example:
+
+```ruby
+# :nocov:
+raise ArgumentError("Unsupported style :#{style}")
+# :nocov:
+```
+
+This can happen for a few reasons, including:
+
+1. When you handle config with a case statement and there is no else block.
+2. When matching with a node pattern even when you handle all cases: all other node types will be excluded before reaching your handler, because the node pattern will not match them.
+
+You will need full line and branch coverage to merge. This helps detect edge cases and prevent errors.
+
 ## Creating new cops
 
 - Document examples of good and bad code in your cop.


corsonknowles:
@bquorning How does this look?

.github/CONTRIBUTING.md
```diff
@@ -51,6 +51,32 @@ We are running [mdformat](https://github.com/executablebooks/mdformat) for forma
 $ mdformat . --number
 ```
 
+### Test Coverage - Line and Branch
+
+We are using [Simplecov](https://github.com/colszowka/simplecov) to track test coverage.
+
+It is included and reported when you run `bundle exec rake`.
```
bquorning, line 58: also when you run `bundle exec rspec`.

.github/CONTRIBUTING.md
```diff
@@ -51,6 +51,32 @@ We are running [mdformat](https://github.com/executablebooks/mdformat) for forma
...
+
+To view the coverage report, open the `coverage/index.html` file in your browser. Here is a quick generic command to find and open the file on macOS:
+
+```console
+$ open $(find . -type f -path "*/coverage/index.html" -print -quit)
```
bquorning, line 63: This simpler version has never failed me:
```suggestion
$ open coverage/index.html
```
Plus, if the file doesn’t exist (before first spec run) you get a usable error message.

.github/CONTRIBUTING.md
```diff
@@ -51,6 +51,32 @@ We are running [mdformat](https://github.com/executablebooks/mdformat) for forma
...
+# :nocov:
+raise ArgumentError("Unsupported style :#{style}")
+# :nocov:
+```
+
+This can happen for a few reasons:
+1. When you handle config with a case statement and there is no else block
+2. When matching with a node pattern even when you handle all cases: all other node types will be excluded before reaching your handler, because the node pattern will not match them.
+
+You will need full line and branch coverage to merge. This has helped prevent and detect errors and edge cases.
```
bquorning, line 78:
```suggestion
You will need full line and branch coverage to merge. This helps detect edge cases and prevent errors.
```

.github/CONTRIBUTING.md
corsonknowles: Fair enough, I aliased this to work across nested repos at some point and never looked back.

`alias cov=...`

bquorning (approved):

.github/CONTRIBUTING.md
```diff
@@ -51,6 +51,33 @@ We are running [mdformat](https://github.com/executablebooks/mdformat) for forma
...
+
+Example:
+```ruby
+# :nocov:
+raise ArgumentError("Unsupported style :#{style}")
+# :nocov:
+```
+
+This can happen for a few reasons, including:
+1. When you handle config with a case statement and there is no else block
```
bquorning, line 76:
```suggestion
1. When you handle config with a case statement and there is no else block.
```

bquorning (commented):
Do you have mdformat available locally? `mdformat .` should fix the CI.

corsonknowles:
I can do that!

It will be a few hours until I am back online today.

corsonknowles:
Formatted!

bquorning:
Yeah, but… Maybe we should merge https://github.com/rubocop/rubocop-rspec/pull/2037 first?

corsonknowles:
Good call! 

Also the docs we're editing here calls this out and says to use `mdformat . --number` so this latest amendment should get it passing enough to merge in either order

bquorning (approved):

PR #2042, opened 2025-02-15
Pluginfy RuboCop RSpec

koic:
This PR adds support for RuboCop's Plugin feature: https://github.com/rubocop/rubocop/pull/13792

It replaces the ad-hoc `Inject` with RuboCop plugins introduced in RuboCop 1.72.

NOTE: Version 1.72.1 or later is specified, as it includes https://github.com/rubocop/rubocop/pull/13837, which removes the `Inject` from spec_helper.rb. While 1.72.0 is sufficient for runtime, but specifying the patch version 1.72.1 keeps things simple.

- [ ] Added tests.
- [x] Updated documentation.


bquorning (commented):
Our users will not be able to continue using `require` in their config file, right? They _need_ to start using `plugins` instead?

tasks/cops_documentation.rake
```diff
@@ -12,6 +12,8 @@ end
 
 desc 'Generate docs of all cops departments'
 task generate_cops_documentation: :yard_for_generate_documentation do
+  RuboCop::ConfigLoader.inject_defaults!("#{__dir__}/../config/default.yml")
```
bquorning, line 15: How come this line is necessary?

rubocop-rspec.gemspec
```diff
@@ -34,8 +34,10 @@ Gem::Specification.new do |spec|
...
     'changelog_uri' => 'https://github.com/rubocop/rubocop-rspec/blob/master/CHANGELOG.md',
     'documentation_uri' => 'https://docs.rubocop.org/rubocop-rspec/',
-    'rubygems_mfa_required' => 'true'
+    'rubygems_mfa_required' => 'true',
+    'default_lint_roller_plugin' => 'RuboCop::RSpec::Plugin'
   }
 
-  spec.add_dependency 'rubocop', '~> 1.61'
+  spec.add_dependency 'lint_roller', '~> 1.1'
+  spec.add_dependency 'rubocop', '~> 1.72.1'
```
bquorning, line 42: I think we want to allow rubocop 1.73+ as well:
```suggestion
  spec.add_dependency 'rubocop', '~> 1.72', '>= 1.72.1'
```

tasks/cops_documentation.rake
koic: It may eventually be removed, but it is still needed for merging the `config/default.yml` configuration used for documentation generation.

From a technical perspective, `Plugin` is called back from RuboCop core during runtime and test execution. However, documentation generation has not yet been integrated with this plugin callback mechanism. In any case, this logic is only required for the documentation generation process, while runtime works with plugins.

rubocop-rspec.gemspec
koic: Oops, I've fixed it 🙇

tasks/cops_documentation.rake
bquorning: Thanks

.rubocop.yml
```diff
@@ -1,10 +1,10 @@
 inherit_from: .rubocop_todo.yml
 
-require:
+plugins:
   - rubocop-performance
   - rubocop-rake
```
bquorning, line 5: These two don’t seem to be plugins just yet.

bquorning (commented):
CI is failing because module `RuboCop::RSpec::Plugin#about` never gets called. Would it make sense to make an assertion on e.g.
```
env = Struct.new(:config_store).new(RuboCop::ConfigStore.new))
RuboCop::Version.extension_versions(env)
```
including the string `"rubocop-rspec"`?

koic:
> Our users will not be able to continue using `require` in their config file, right? They _need_ to start using `plugins` instead?

For compatibility, the old `require` specification will still work. However, warning will be displayed to encourage updating to `plugins`. Therefore, users can replace it with `plugins` at any time.

.rubocop.yml
koic: RuboCop Performance 1.24 with plugin support has been released:
https://github.com/rubocop/rubocop-performance/releases/tag/v1.24.0

bquorning: 👍🏼 Thanks

koic:
> CI is failing because module `RuboCop::RSpec::Plugin#about` never gets called.

Sorry, `Plugin#about` is something that gets called back from the RuboCop engine, so to be honest, I’m not sure if this coverage is meaningful.

koic:
Note, a simple plugin migration guide is available for reference:
https://docs.rubocop.org/rubocop/1.72/plugin_migration_guide.html

lib/rubocop/rspec/plugin.rb
```diff
@@ -0,0 +1,35 @@
...
+    # A plugin that integrates RuboCop RSpec with RuboCop's plugin system.
+    class Plugin < LintRoller::Plugin
+      def about
+        LintRoller::About.new(
+          name: 'rubocop-rspec',
+          version: Version::STRING,
+          homepage: 'https://github.com/rubocop/rubocop-rspec',
+          description: 'Code style checking for RSpec files.'
+        )
+      end
```
bquorning, line 16: Let’s tell SingleCov to not care about this method:
```suggestion
      # :nocov:
      def about
        LintRoller::About.new(
          name: 'rubocop-rspec',
          version: Version::STRING,
          homepage: 'https://github.com/rubocop/rubocop-rspec',
          description: 'Code style checking for RSpec files.'
        )
      end
      # :nocov:
```

.github/workflows/main.yml
```diff
@@ -91,8 +91,8 @@ jobs:
       - uses: actions/checkout@v4
       - name: Use oldest RuboCop allowed by gemspec
         run: |
-          sed -nr "s/ *spec.add_dependency 'rubocop', '~> ([0-9\.]+)'/gem 'rubocop', '= \1'/p" \
-            rubocop-rspec.gemspec > Gemfile.local
+          sed -nr "s/ *spec.add_dependency 'rubocop', '~> ([0-9\.]+)'(, '>= ([0-9\.]+)')?/\
+                   gem 'rubocop', '= \3' || '= \1'/p" rubocop-rspec.gemspec > Gemfile.local
```
bquorning, line 95: I think we can ignore the `~>` requirement and only use the `>=` one, right?
```suggestion
          sed -nr "s/ *spec.add_dependency 'rubocop'.*'>= ([0-9\.]+)'/gem 'rubocop', '= \1'/p" \
            rubocop-rspec.gemspec > Gemfile.local
```

bquorning:
cc @pirj, @ydah

pirj (approved):
Nice!

I’m away from my computer for a few days and can’t test manually.

lib/rubocop/rspec/plugin.rb
koic: I've applied the suggestion.

.github/workflows/main.yml
koic: I considered reverting to `~>` in the future, but this method seems simpler. Thank you!

bquorning:
Thank you @koic 🙏🏼

ydah (approved):
Thank you!

PR #2046, opened 2025-02-16
Use new ast features

bquorning:
Now that we depend on a more recent rubocop-ast version, we can use the features that @dvandersluis mentioned in #2025:

- Use `any_block` in node patterns.
- Use `Node#type?` with multiple arguments.

- [ ] Added tests.
- [ ] Updated documentation.

### DIFF
--- a/lib/rubocop/cop/rspec/receive_messages.rb
+++ b/lib/rubocop/cop/rspec/receive_messages.rb
@@ -148,8 +148,7 @@ module RuboCop
         end
 
         def heredoc_or_splat?(node)
-          ((node.str_type? || node.dstr_type?) && node.heredoc?) ||
-            node.splat_type?
+          (node.type?(:str, :dstr) && node.heredoc?) || node.splat_type?
         end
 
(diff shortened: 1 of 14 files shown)


lib/rubocop/cop/rspec/receive_messages.rb
```diff
@@ -148,7 +148,7 @@ def item_range_by_whole_lines(item)
         end
 
         def heredoc_or_splat?(node)
-          ((node.str_type? || node.dstr_type?) && node.heredoc?) ||
+          ((node.type?(:str, :dstr)) && node.heredoc?) ||
```
pirj, line 151: Extra parens?
```suggestion
          (node.type?(:str, :dstr) && node.heredoc?) ||
```

pirj (approved):
Thanks!

lib/rubocop/cop/rspec/receive_messages.rb
bquorning: Fixed

PR #2049, opened 2025-02-19
Remove `Reference` from config

Earlopain:
On https://docs.rubocop.org/rubocop-rspec/3.5/cops_rspec.html, every cop has a reference section, linking to the same text on rubydoc. It also doesn't render very well, ` for example is just taken as is.

The reference section is intended for style guide urls and other extra reading for specific cops. I don't think linking to rubydoc provides much value, it just dilutes the actual references, so maybe we can just remove them?

- [ ] Added tests.
- [x] Updated documentation.

### DIFF
--- a/config/default.yml
+++ b/config/default.yml
@@ -116,33 +116,28 @@ RSpec/AlignLeftLetBrace:
   Description: Checks that left braces for adjacent single line lets are aligned.
   Enabled: false
   VersionAdded: '1.16'
-  Reference: https://www.rubydoc.info/gems/rubocop-rspec/RuboCop/Cop/RSpec/AlignLeftLetBrace
 
 RSpec/AlignRightLetBrace:
   Description: Checks that right braces for adjacent single line lets are aligned.
   Enabled: false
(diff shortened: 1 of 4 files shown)


bquorning:
The links were added in #430, arguing that 

> This is great for IDE/text editor integration and means that users don't have to copy/paste RSpec/CopName into Google to find documentation. Because you guys already use RubyDoc, we basically get this for free.

Perhaps we can keep the links, but just not render them in the HTML docs?

Earlopain:
The proper place for that nowadays is `Cop.documentation_url`. `ruby-lsp` for example uses it to link to first-party docs on offenses: 
image

You get this for free by defining `DocumentationBaseURL` here: https://github.com/rubocop/rubocop-rspec/blob/e8ac379f27bc09d561c4bd9c13769cd2e7470baa/config/default.yml#L5

I didn't find an option to output this link on cli for offenses but there probably is one and I missed it? There is `--show-docs-url` but that just shows the links without actually running it.

PR #2050, opened 2025-02-20
Only run publish workflow on upstream repo

Earlopain:
When syncing forks, they also try to publish (which doesn't make a lot of sense and doesn't work)

https://github.com/Earlopain/rubocop-rspec/actions/runs/13419581185/job/37488734637


bquorning (approved):
Thank you.

PR #2052, opened 2025-02-25
Always use `expect_correction` or `expect_no_corrections` when autocorrect is supported

Earlopain:
In https://github.com/rubocop/rubocop/pull/13183 I want to add a cop that enforces this automatically. This fixes offenses here in anticipation of that.

### DIFF
--- a/spec/rubocop/cop/rspec/example_wording_spec.rb
+++ b/spec/rubocop/cop/rspec/example_wording_spec.rb
@@ -329,6 +329,8 @@ RSpec.describe RuboCop::Cop::RSpec::ExampleWording do
       DESC
       end
     RUBY
+
+    expect_no_corrections
   end
 
   it 'flags an unclear description' do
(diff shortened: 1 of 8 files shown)


pirj (approved):
I expected much more offences after reading the pr’s description 😅

Nice cop, thank you.

spec/rubocop/cop/rspec/example_wording_spec.rb
```diff
@@ -329,6 +329,8 @@
       DESC
       end
     RUBY
+
+    expect_no_corrections
```
bquorning, line 333: Is this what we want, though? Does it discourage people from implementing autocorrect, or will it work fine as a “placeholder” until autocorrect is implemented.

Earlopain: This method only useful for cops that implement autocorrect but not for all cases. If someone doesn't implement autocorrect to begin with, there will also be no offense. There's no cop to check for `extend AutoCorrector` (;

In this specific case, I think autocorrect could theoretically be implemented but hasn't because heredocs are hard and nobody actually writes tests with heredoc text descriptions.

bquorning (approved):

PR #2054, opened 2025-03-03
Fix a false positive for `RSpec/DescribedClass` when `SkipBlocks` is true and numblocks are used

Earlopain:
I will probably make a few more of these. I don't particularly care about numblocks, this just future-proofs it for `itblock`.

- [x] Added tests.
- [ ] Updated documentation.

### DIFF
--- a/spec/rubocop/cop/rspec/described_class_spec.rb
+++ b/spec/rubocop/cop/rspec/described_class_spec.rb
@@ -10,7 +10,7 @@ RSpec.describe RuboCop::Cop::RSpec::DescribedClass do
       expect_offense(<<~RUBY)
         describe MyClass do
           controller(ApplicationController) do
-            bar = MyClass
+            self.bar = MyClass
           end
 
           before do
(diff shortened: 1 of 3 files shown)


spec/rubocop/cop/rspec/described_class_spec.rb
```diff
@@ -24,6 +24,26 @@
         end
       RUBY
     end
+
+    it 'ignores offenses within non-rspec numblocks' do
+      expect_offense(<<~RUBY)
+        describe MyClass do
+          controller(ApplicationController) do
+            _1
```
pirj, line 32: What is the real-world use case for this?

spec/rubocop/cop/rspec/described_class_spec.rb
```diff
@@ -24,6 +24,26 @@
...
+            bar = MyClass
```
pirj, line 33: Is this a setter? Doesn’t it need to be prefixed with `self.`?

Earlopain: I just moved this over from the test preceeding it. I added the prefix to both to make more sense

spec/rubocop/cop/rspec/described_class_spec.rb
Earlopain: This is just here to make the `block` node into a `numblock` node. Since this is a user-defined method it can yield as much or as little as it wants and the user can use a few ways to express that. It just aligns it to work for all types of blocks there are (currently `foo {}`, `foo { |bar| }` and `foo { _1 }`, but also `foo { it }` in the future from Ruby 3.4).

I made the spec a bit more aligned to how could look like in actual code.

pirj (approved):
Thank you!

bquorning (approved):

pirj:
Sorry to bother you with this, I’m away from my computer. Can you please resolve the conflict?

Earlopain:
Sure, no problem

PR #2061, opened 2025-03-14
Add explanation for ExpectChange unsafe autocorrection

lovro-bikic:
https://github.com/rubocop/rubocop-rspec/pull/1178 marked `RSpec/ExpectChange`'s autocorrection as unsafe, but it didn't update the docs. This PR adds an explanation for the cop's unsafe autocorrection. 

- [ ] Added tests.
- [x] Updated documentation.

### DIFF
--- a/lib/rubocop/cop/rspec/expect_change.rb
+++ b/lib/rubocop/cop/rspec/expect_change.rb
@@ -5,10 +5,29 @@ module RuboCop
     module RSpec
       # Checks for consistent style of change matcher.
       #
-      # Enforces either passing object and attribute as arguments to the matcher
-      # or passing a block that reads the attribute value.
+      # Enforces either passing a receiver and message as method arguments,
+      # or a block.
       #
       # This cop can be configured using the `EnforcedStyle` option.
+      #
+      # @safety
+      #   Autocorrection is unsafe because `method_call` style calls the
+      #   receiver *once* and sends the message to it before and after
+      #   calling the `expect` block, whereas `block` style calls the
+      #   expression *twice*, including the receiver.
+      #
+      #   If your receiver is dynamic (e.g., the result of a method call) and
+      #   you expect it to be called before and after the `expect` block,
+      #   changing from `block` to `method_call` style may break your test.
+      #
+      #   [source,ruby]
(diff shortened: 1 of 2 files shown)


lib/rubocop/cop/rspec/expect_change.rb
```diff
@@ -10,6 +10,15 @@ module RSpec
       #
       # This cop can be configured using the `EnforcedStyle` option.
       #
+      # @safety
+      #   Autocorrection is unsafe because `block` style evaluates the object
+      #   twice (before and after evaluating `expect` block), whereas
+      #   `method_call` style evaluates the object only once.
```
pirj, line 16: What do you mean by “object” here? Statement or a series of statements inside the block?

pirj: After looking at the original PR, it’s clear to me. Nice catch, and thanks for marking the cop, it’s totally justified.

Can you please provide a bit more context in safety docs to make it obvious for the reader?

bquorning: If I understand https://github.com/rubocop/rubocop-rspec/issues/1174 correctly, the issue is that the `method_call` style calls `message` twice on the same `receiver`, so to speak. But with the `block` style, the entire block is evaluated twice – including instantiating two different `receiver` objects.

In #1174, the `#time` method _is_ being called twice in each example, which confused me at first.

lovro-bikic: > What do you mean by “object” here? Statement or a series of statements inside the block?

I followed the terminology used in the previous section ("Enforces either passing object and attribute as...") which I think was in turn borrowed from the example in [RSpec docs](https://rspec.info/features/3-13/rspec-expectations/built-in-matchers/change/):
```ruby
expect { do_something }.to change(object, :attribute)
```

Instead of object/attribute, more precise terminology would be receiver/message, but I'm not sure if that is more approachable?

bquorning (approved):
Thank you ❤️ The explanation is so much clearer now.

ydah (approved):
Thank you❤️

PR #2062, opened 2025-03-25
Update `RSpec/SharedExamples` to allow specifying a preferred method to use for defining and including shared examples and context

dvandersluis:
Allows `RSpec/SharedExamples` to be configured with preferred methods for defining and including shared examples and context, so that consistency can be enforced.

Follows rubocop/rubocop#14033.

- [x] Added tests.
- [x] Updated documentation.

### DIFF
--- a/lib/rubocop/cop/rspec/shared_examples.rb
+++ b/lib/rubocop/cop/rspec/shared_examples.rb
@@ -3,11 +3,27 @@
 module RuboCop
   module Cop
     module RSpec
-      # Checks for consistent style for shared example names.
+      # Checks for consistent style for shared examples.
       #
-      # Enforces either `string` or `symbol` for shared example names.
+      # Enforces either `string` or `symbol` for shared example names. This
+      # can be configured using the `EnforcedStyle` option.
       #
-      # This cop can be configured using the `EnforcedStyle` option
+      # Can also be used to enforce the preferred method name to use for
+      # defining and including shared examples and contexts. Each can get a
+      # different value.
+      #
+      # [source,yaml]
+      # ----
+      # RSpec/SharedExamples:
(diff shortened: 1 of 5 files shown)


lib/rubocop/cop/rspec/shared_examples.rb
```diff
@@ -72,6 +92,37 @@ def offense?(ast_node)
...
+            cop_config['PreferredExamplesMethod']
+          elsif SharedGroups.context(method_name)
+            cop_config['PreferredContextMethod']
+          elsif Includes.examples(method_name)
+            cop_config['PreferredIncludeExamplesMethod']
+          elsif Includes.context(method_name)
+            # :nocov: - simplecov is not detecting the coverage of this branch!
+            cop_config['PreferredIncludeContextMethod']
+            # :nocov:
+          end
```
dvandersluis, line 123: I'm unclear as to why, but simplecov doesn't think this branch is covered, even though it is (line 270 of the spec).

dvandersluis:
Failing edge rubocop check is fixed in #2063.

pirj:
Nice!

Glancing over, it feels this would better be a separate cop. Do you have a specific reason to modify the existing one?

Once concern that comes to mind is that new cops are pending by default. We're releasing major versions more often, and in less than a year this cop will be enabled for everyone.

dvandersluis:
@pirj ha! I went back and forth on if it should enhance the existing cop or create a new one. I settled on updating because `RSpec/SharedExamples` seems like the right name for this and I wasn't sure what to name a new cop as. Do you have a suggestion?

dvandersluis:
How about for these changes `RSpec/PreferredSharedExampleMethods`?

dvandersluis:
I'll open a new PR with the new cop.

PR #2065, opened 2025-03-25
Add new `RSpec/PreferredSharedExampleMethods` cop to enforce consistency when defining shared examples

dvandersluis:
Allows RSpec/SharedExamples to be configured with preferred methods for defining shared examples and context, so that consistency can be enforced.

Follows https://github.com/rubocop/rubocop/pull/14033, and reimplementation of #2062.

- [x] Added tests.
- [x] Updated documentation.


.rubocop.yml
```diff
@@ -119,6 +119,13 @@ RSpec/ExampleLength:
 RSpec/MultipleExpectations:
   Max: 2
 
+RSpec/PreferredSharedExampleMethods:
+  Enabled: true
+  PreferredExamplesMethod: shared_examples
+  PreferredContextMethod: shared_context
+  PreferredIncludeExamplesMethod: include_examples
+  PreferredIncludeContextMethod: include_context
```
dvandersluis, line 127: I chose the most commonly used methods in the codebase:

* `shared_examples` was used 10 times and `shared_examples_for` was used 4 times.
* `include_examples` was used 42 times and `it_behaves_like` was never used.

lib/rubocop/cop/rspec/preferred_shared_example_methods.rb
```diff
@@ -0,0 +1,138 @@
...
+          if SharedGroups.examples(method_name)
+            cop_config['PreferredExamplesMethod']
+          elsif SharedGroups.context(method_name)
+            cop_config['PreferredContextMethod']
+          elsif Includes.examples(method_name)
+            cop_config['PreferredIncludeExamplesMethod']
+          elsif Includes.context(method_name)
+            # :nocov: - simplecov is not detecting the coverage of this branch!
+            cop_config['PreferredIncludeContextMethod']
+            # :nocov:
```
dvandersluis, line 132: This branch has coverage (line 114 of the spec), but simplecov doesn't recognize it. Is there anything else I can do?

bquorning: I think Simplecov really complains about there not being an `else` branch, and it not being covered by tests. In other places where if/elsif’s cover all cases, we have added a “noop” else branch:
```suggestion
            cop_config['PreferredIncludeContextMethod']
          else
            # :nocov:
            :noop
            # :nocov:
```

dvandersluis: @bquorning thanks! I need to use `false` instead of `:noop` but this works!

dvandersluis: (or I could use `nil` which would make more sense, but would require a disable of `Style/EmptyElse`).

dvandersluis: Never mind, the `else` branch can't actually be reached, `:noop` it is!

bquorning (approved):
@pirj, what do you think?

.rubocop.yml
lovro-bikic: Fyi, `include_examples` and `it_behaves_like` [don't do the same thing](https://rspec.info/features/3-13/rspec-core/example-groups/shared-examples/) (unlike `shared_examples` and `shared_examples_for` which are synonyms).

Most projects I worked on typically used both, `include_examples` for example can't be used if you want to do multiple includes which override each other's setups:
```ruby
RSpec.describe do
  shared_examples 'foo' do |var|
    subject { var }

    it { is_expected.to eq(var) }
  end

  include_examples 'foo', 1 # fails because the second include sets subject to 2
  include_examples 'foo', 2
end
```
so you need to use `it_behaves_like` to wrap each include in a context.

dvandersluis: So a few thoughts:
1. The easiest solution is probably just to leave `PreferredIncludeExamplesMethod` unset (which I should probably do here, because I didn't realize the distinction), unless you know you always want to use one or the other. 
2. It is also possible to do what I did in the spec for this cop and define a top level context within the shared examples, in which case `include_examples` still works since the examples have their own contexts. The changed `it_behaves_like` examples still work for the same reason, but it might be worthwhile to undo that change.
3. The `Language` hash doesn't distinguish between the two which is what this cop is based on. I'm not sure if it should, but that is outside the scope of this PR at the least.
4. Given that `include_examples` doesn't create a context, we should probably have a cop that looks for shared examples that set `subject` or `let` that then are called with `include_examples` multiple times in the same context. Again outside of the scope of this PR, but would be good to prevent a footgun.
5. The other solution here would be to remove the `PreferredIncludeExamplesMethod` configuration, or even not consider includes at all. I don't think it's a huge risk to keep it given that it defaults to unset (ie. all methods are acceptable). 
6. We could theoretically treat `include_examples` differently, but again because the different Includes are not distinguished in the `Language` hash, the problem could still exist if it is aliased.
7. At the very least we should document this distinction. 

/cc @bquorning @pirj

pirj:
As @lovro-bikic mentions, `include_examples` and `it_behaves_like` are different. Let's split the cop (again!).
One cop will enforce which alias to use.

The other - whether to include examples or include them into nested context, `include_examples` vs `it_behaves_like`.


There are use cases for `include_examples`, but in most cases it's a footgun. I suggest no configurable value, just the cop to flag `include_examples` as an offence. Those who understand the difference, and who deliberately want it, can disable the cop.
> a cop that looks for shared examples that set subject or let that then are called with

Methods as well. Also, `before`/`after` hooks can run multiple times if defined in a shared group and included multiple times. Too much of an effort to detect those side effects, and I'm uncertain if this can be done reliably. I can't recall major benefits of using `include_examples` to justify its usage in specs, and the effort to create a cop to ensure its safe usage.


There may be another cop that would detect shared contexts that contain examples, and yet another, that will detect no examples in a shared example group. We may have an issue to track this idea already.


I've grepped [real-world-rspec](https://github.com/pirj/real-world-rspec), and usage stats are like (rough `$ rg --no-ignore --hidden --glob *.rb --no-filename shared_examples_for | wc -l`, seems not to count blank lines):

aliases:
shared_context - 636
shared_examples - 6401
shared_examples_for - 1295

include_examples - 4450
include_context (does the same as above!)  - 2628

aliases:
it_behaves_like - 24441
it_should_behave_like - 562

`it_should_behave_like` will be removed in RSpec 4, other aliases will remain.

My proposal:
1. For the cop that enforces a preferred alias to use, not consider include_examples and it_behaves_like aliases
2. File an issue to add a cop to flag include_examples
3. File (unless it's there already) an issue to add a cop for enforcing semantic use of shared_context vs shared_examples

dvandersluis:
@pirj thank you for the feedback!

I have updated this PR to only apply to **defining** shared examples. 

> There are use cases for include_examples, but in most cases it's a footgun. I suggest no configurable value, just the cop to flag include_examples as an offence. Those who understand the difference, and who deliberately want it, can disable the cop.

I'll work on this next.

dvandersluis:
BTW I just realized that `it_behaves_like` vs `it_should_behave_like` is already covered by `RSpec/ItBehavesLike` 😄

pirj:
I am sorry to have missed that. We seem to be in the spot where we can’t remember all the cops we already have 😅

dvandersluis:
@pirj @bquorning anything else needed for this one?

spec/rubocop/cop/rspec/preferred_shared_example_methods_spec.rb
```diff
@@ -0,0 +1,97 @@
+# frozen_string_literal: true
+
+RSpec.describe RuboCop::Cop::RSpec::PreferredSharedExampleMethods, :config do
+  shared_examples 'preferred method' do |config_key, methods|
```
pirj, line 4: Nitpick: add  ‘preferred’ and ‘**others’ (or pass those as kwargs to indicate the intent when calling, too)

spec/rubocop/cop/rspec/preferred_shared_example_methods_spec.rb
```diff
@@ -0,0 +1,97 @@
...
+          end
+
+          it "does not register an offense for `RSpec.#{method}`" do
+            expect_no_offenses(<<~RUBY)
+              RSpec.#{method} 'desc'
+            RUBY
+          end
+        end
+
+        it 'does not register an offense for another method' do
```
pirj, line 29: This doesn’t check that the cop is not offended by ‘shared_state’/‘shared_scenario’, does it?

pirj:
I was always curious, and this question reveals each time when I think about directory-local .rubocop.yml

which could come handy in a case when there’s an alias defined

eg ‘shared_context’ is preferred in specs/, but shared_state in features/

How does it work with multi-threaded RuboCop, that nechanism that parallelizes execution? There should be separate ConfigLoaders (or whatever they are called).

How does rubocop-rspec work in this case? We do have a global Language module that keeps a reference to the config?

Can’t this be a source of flaky subtle bugs?

pirj:
The question above is totally unrelated to the PR’s code.
It just made me remember this again

dvandersluis:
@pirj I've addressed your requests now.

spec/rubocop/cop/rspec/preferred_shared_example_methods_spec.rb
dvandersluis: I added `opposing_methods` to list the methods from the other group (example methods for context and vice versa), and ensured that they do not register an offense.

dvandersluis:
> which could come handy in a case when there’s an alias defined
>
> eg ‘shared_context’ is preferred in specs/, but shared_state in features/

I tried this out and it _should_ work how you expect. You can create .rubocop.yml files in subdirectories, and they do get loaded (and can inherit from the main one too).

> How does it work with multi-threaded RuboCop, that nechanism that parallelizes execution? There should be separate ConfigLoaders (or whatever they are called).

There's a lot of layers to investigate here, but ultimately what happens in short is that each inspected file [mobilizes a team of cops](https://github.com/rubocop/rubocop/blob/596db4faf98187761fa815d304a36fb3014895c4/lib/rubocop/runner.rb#L345-L358), during which a config object is [created for that file](https://github.com/rubocop/rubocop/blob/596db4faf98187761fa815d304a36fb3014895c4/lib/rubocop/runner.rb#L374-L377). `ConfigStore.for_file` eventually calls `ConfigFinder.find_project_dotfile`, which calls [`FileFinder.find_file_upwards`](https://github.com/rubocop/rubocop/blob/f12f1e585b7817a60be3d0c3bb6727da13742776/lib/rubocop/file_finder.rb#L13-L18) which searches from the file directory upwards until a configuration file is found (and uses the first one found).

So each file gets the correct config (there's some caching involved as well) for the directory it's in. 

> We do have a global Language module that keeps a reference to the config?

Language has a class level attr_accessor `config`, but it is set [`on_new_investigation`](https://github.com/rubocop/rubocop-rspec/blob/d841fe4152d780e9fcbfdf0789dbb6a7a596ed1e/lib/rubocop/cop/rspec/base.rb#L19-L22) which means that each **cop** will reset the Language data each time it is run. A cop can be run one or more times, per file (depending on if there are autocorrections that cause another iteration), so this reset can happen a lot of times. 

Therefore, per-directory `Language` should work properly as well.

config/default.yml
```diff
@@ -775,6 +775,14 @@ RSpec/PredicateMatcher:
   StyleGuide: https://rspec.rubystyle.guide/#predicate-matchers
   Reference: https://www.rubydoc.info/gems/rubocop-rspec/RuboCop/Cop/RSpec/PredicateMatcher
 
+RSpec/PreferredSharedExampleMethods:
```
pirj, line 778: On a different note, I suddenly recalled we have RSpec/Dialect cop 🙈😭

I sincerely apologise if that other cop does what this one does with aliases, again, it slipped my mind as I’m less involved with the project lately.

config/default.yml
```diff
@@ -775,6 +775,14 @@ RSpec/PredicateMatcher:
...
+  Description: Checks for consistent method selection for shared examples.
+  Enabled: pending
+  VersionAdded: "<<next>>"
+  PreferredExamplesMethod: ~
```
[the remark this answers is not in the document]
dvandersluis: I think it's useful to enforce consistency, there are plenty of other examples of this across RuboCop of cops that are just meant to enforce one style or another when there isn't an actual difference between them. For instance, `Style/CollectionMethods` allows users to choose to enforce `map` vs. `collect`, etc. 

I didn't specify actual defaults because I don't think there is one answer (but we could if we want...).

> Why do we implement this cop in the first place? 

In a codebase with many contributors, there is no particular reason (IMO) to use one alias over the other so it becomes inconsistent. I noticed this in the main rubocop repo and wanted to fix it, and it was suggested to add a RuboCop::RSpec cop instead 😆 In the meantime, the inconsistency was fixed directly in https://github.com/rubocop/rubocop/pull/14034.

pirj:
Thank you for the in-depth explanation, @dvandersluis ! 

How does context switch happen, how do we load multiple cores? Do we have a pool of rubocop processes, each with its own set of class-level attributes? Or can a thread context switch happen before the cop finished the investigation, and some other cop would overwrite the cached language DSL config?
How does this isolation work?
Can we guarantee that if a cop finishes with this config being untouched?

config/default.yml
dvandersluis: Oh wow! I didn't know about it either. The tests for it are pretty sparse and the documentation is a bit confusing as to how to configure, but it does seem to do what I was looking to do.

```yaml
RSpec/Dialect:
  Enabled: true
  PreferredMethods:
    shared_examples_for: shared_examples
```

With this in mind, we can probably close this PR.

dvandersluis:
I'm going to close this PR given `RSpec/Dialect` already covers this. If there's anything you want me to extract from this (for instance, enabling `RSpec/Dialect` and replacing `shared_examples_for`, please let me know!)

PR #2066, opened 2025-03-26
Add new `RSpec/IncludeExamples` cop to prefer `it_behaves_like` over `include_examples`

dvandersluis:
Add cop to enforce using `it_behaves_like` over `include_examples` due to the potential footguns from `include_examples`. Users can disable this cop if they want to use `include_examples`.

Follows https://github.com/rubocop/rubocop-rspec/pull/2065#issuecomment-2753440476 action item 2.

- [x] Added tests.
- [x] Updated documentation.


spec/rubocop/cop/rspec/include_examples_spec.rb
```diff
@@ -0,0 +1,32 @@
...
+
+  it 'does not register an offense for `it_behaves_like`' do
+    expect_no_offenses(<<~RUBY)
+      it_behaves_like 'examples'
+    RUBY
+  end
+
+  it 'does not register an offense for `it_should_behave_like`' do
+    expect_no_offenses(<<~RUBY)
+      it_should_behaves_like 'examples'
```
pirj, line 23:
```suggestion
      it_should_behave_like 'examples'
```

pirj (approved):
Thank you! 🤩 

I was unaware that we use include_examples so extensively in our specs.

spec/rubocop/cop/rspec/include_examples_spec.rb
dvandersluis: Oops! Thank you, it's fixed now.

.rubocop.yml
```diff
@@ -116,6 +116,9 @@ RSpec/ExampleLength:
     - heredoc
   Max: 11
 
+RSpec/IncludeExamples:
+  Enabled: true
+
```
bquorning, line 121: We used to have a section at the bottom, but I see that it was removed in f41cd5a5f1349335bdcdf3f3a77a9c4d8b46db19 and 1aa987fa709c77e9ba068498e624c08ab2e5c9e8.

Could I get you to add this at the very bottom of this file instead?
```yaml

# Enable our own pending cops.

RSpec/IncludeExamples: {Enabled: true}
```

pirj (approved):

bquorning:
Learning from the fate of #2065, could we have use RSpec/Dialect instead of adding this new cop too?

PR #2068, opened 2025-03-29
Improve PredicateMatcher recommendations

### LINKED ISSUE #2067: PredicateMatcher recommends changing `#key?` to matcher `be_key`
> This is very old, but the cop still forces some odd changes by default:
> 
> ```diff
> - expect(hash.key?(:foo)).to be_truthy
> + expect(hash).to be_key(:foo)
> ```
> 
> From what I've seen, it's possible to configure matchers, but not to which methods it applies. Couldn't some of these methods common methods even be hardcoded? I don't think that anyone using `key?` would prefer `be_key` over `key?`. And if that'd be the case, with the current implementation it'd be possible to configure it using `AllowedExplicitMatchers` 

 _Originally posted by @dgmora in [#466](https://github.com/rubocop/rubocop-rspec/issues/466#issuecomment-2720691383)_

bquorning:
PredicateMatcher should not recommend changing `#key?` to `be_key`, but will now instead suggest using `have_key`.

Fixes #2067

- [x] Added tests.
- [ ] Updated documentation.

### DIFF
--- a/lib/rubocop/cop/rspec/predicate_matcher.rb
+++ b/lib/rubocop/cop/rspec/predicate_matcher.rb
@@ -69,18 +69,21 @@ module RuboCop
                  matcher_name: to_predicate_matcher(predicate.method_name))
         end
 
+        TO_PREDICATE_MATCHER_MAP = {
+          exist?:       'exist',
+          exists?:      'exist',
+          include?:     'include',
+          instance_of?: 'be_an_instance_of',
+          is_a?:        'be_a',
+          respond_to?:  'respond_to'
+        }.freeze
+        private_constant :TO_PREDICATE_MATCHER_MAP
+
         def to_predicate_matcher(name)
-          case name = name.to_s
-          when 'is_a?'
-            'be_a'
-          when 'instance_of?'
-            'be_an_instance_of'
-          when 'include?', 'respond_to?'
-            name[0..-2]
-          when 'exist?', 'exists?'
-            'exist'
-          when /\Ahas_/
-            name.sub('has_', 'have_')[0..-2]
+          if TO_PREDICATE_MATCHER_MAP.key?(name)
+            TO_PREDICATE_MATCHER_MAP.fetch(name)
+          elsif name.start_with?('has_')
+            name.to_s.sub('has_', 'have_')[0..-2]
           else
             "be_#{name[0..-2]}"
           end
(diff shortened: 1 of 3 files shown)


lib/rubocop/cop/rspec/predicate_matcher.rb
```diff
@@ -241,18 +245,25 @@ def move_predicate(corrector, actual, matcher, block_child)
...
+          be_instance_of:    'instance_of?',
+          be_kind_of:        'is_a?',
+          include:           'include?',
+          respond_to:        'respond_to?'
+        }.freeze
+        private_constant :TO_PREDICATE_METHOD_MAP
+
         def to_predicate_method(matcher)
-          case matcher = matcher.to_s
-          when 'be_a', 'be_an', 'be_a_kind_of', 'a_kind_of', 'be_kind_of'
```
pirj, line 246: Subjectively, this reads better than multiple hash items

spec/rubocop/cop/rspec/predicate_matcher_spec.rb
```diff
@@ -91,6 +91,8 @@
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ Prefer using `be_something` matcher over `something?`.
           expect(foo.has_key?('foo')).to be_truthy
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ Prefer using `have_key` matcher over `has_key?`.
+          expect(foo.key?('foo')).to be_truthy
+          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ Prefer using `have_key` matcher over `key?`.
```
pirj, line 95: Does this imply that the receiver is an instance of a Hash?
If it’s e.g. an SqlColumn, have_key? won’t fit, and it should be is_key? instead

pirj: Can we make a distinction at least if there’s an argument?

pirj: TIL key? is aliased as has_key? https://ruby-doc.org/3.4.1/Hash.html#method-i-include-3F

pirj: Proposal:
Only auto-correct expect(foo.key?(foo)) with a single argument.
Only raise an offence with zero or multiple args or a block

bquorning: To be honest, I don't really like this cop (or others) that make broad assumptions about types which can lot safely be deduced with static analysis. There will _always_ be false positives.

bquorning: > TIL key? is aliased as has_key?

And I thought one of them had been deprecated, but apparently that's not the case.

I also think this method is aliased as `member?` and `includes?` btw.

lib/rubocop/cop/rspec/predicate_matcher.rb
```diff
@@ -69,18 +69,22 @@ def message_inflected(predicate)
...
+          include?:     'include',
+          instance_of?: 'be_an_instance_of',
+          is_a?:        'be_a',
+          key?:         'have_key',
+          respond_to?:  'respond_to'
+        }.freeze
+        private_constant :TO_PREDICATE_MATCHER_MAP
+
         def to_predicate_matcher(name)
-          case name = name.to_s
```
pirj, line 73: Side note, unrelated to this PR. This list is incomplete, since RSpec has support for eg start_with/end_with. They work with Arrays, too. Also cover/exist/...

lib/rubocop/cop/rspec/predicate_matcher.rb
bquorning: I changed it to a hash lookup because one of the Metric cops said the method became too long. But I agree with you, and I will change it back when/if I get back to this PR.

pirj:
How do we know that the target is hash-like and has the `has_key?` method? What if it only has `key?` like e.g. [1](https://api.rubyonrails.org/classes/ActionDispatch/Http/Headers.html#method-i-key-3F), [2](https://api.rubyonrails.org/classes/ActionDispatch/Flash/FlashHash.html#method-i-key-3F), [3](https://api.rubyonrails.org/classes/ActionDispatch/Request.html#method-i-key-3F), [4](https://api.rubyonrails.org/classes/ActiveRecord/Result/IndexedRow.html#method-i-key-3F), [5](https://api.rubyonrails.org/classes/ActiveSupport/EncryptedFile.html#method-i-key-3F), ...?

Maybe we can fix the original problem
```diff
- expect(hash.key?(:foo)).to be_truthy
+ expect(hash).to be_key(:foo)
```
if we just ignored `key?` calls that accept arguments? The cop really has no way to know if it's a hash-like or not, if there's an accompanying `has_key?` to the original `key?` or not. Or does it have a way?

bquorning:
With static analysis, we have no way of knowing which methods are implemented on the object under test. The only thing we do know is, if users use the `have_key?` RSpec matcher, `RSpec::Matchers::BuiltIn::Has` will call `#has_key?` instead. And that is quite the assumption to make, I guess.

And you are right, we should probably not assume that `#has_key?` exists just because the `#key?` method exists.

PR #2070, opened 2025-04-04
RSpec/SpecFilePathFormat allow multiple values for same key in IgnoreMetadata

sriedel:
RSpec/SpecFilePathFormat allows ignoring certain values in rspec metadata through the IgnoreMetadata option, as given in the documentation:

```
      # @example `IgnoreMetadata: {type=>routing}` (default)
      #   # good
      #   whatever_spec.rb         # describe MyClass, type: :routing do; end
```

However the current implementation does not seem to allow supplying multiple values for the same key - e.g. if you have some specs marked `type: controller`, others as `type: :model`  etc, only one of `controller`, `model`, ... can be supplied to be ignored.

This is probably due to an oversight (or me not being able to figure out how to properly supply an array of values to be ignored).

The PR allows for either an array or a single value to be given for a key in the IgnoreMetadata option.

While it would also be desirable to give a `true` value to ignore all metadata, in the format that `IgnoreMethods` use, this should probably be part of a different PR.

### DIFF
--- a/config/default.yml
+++ b/config/default.yml
@@ -925,6 +925,7 @@ RSpec/SpecFilePathFormat:
   IgnoreMetadata:
     type: routing
   VersionAdded: '2.24'
+  VersionChanged: "<<next>>"
   Reference: https://www.rubydoc.info/gems/rubocop-rspec/RuboCop/Cop/RSpec/SpecFilePathFormat
 
 RSpec/SpecFilePathSuffix:
--- a/config/default.yml
+++ b/config/default.yml
@@ -925,7 +925,6 @@ RSpec/SpecFilePathFormat:
   IgnoreMetadata:
     type: routing
   VersionAdded: '2.24'
-  VersionChanged: "<<next>>"
   Reference: https://www.rubydoc.info/gems/rubocop-rspec/RuboCop/Cop/RSpec/SpecFilePathFormat
 
 RSpec/SpecFilePathSuffix:
--- a/lib/rubocop/cop/rspec/spec_file_path_format.rb
+++ b/lib/rubocop/cop/rspec/spec_file_path_format.rb
@@ -32,6 +32,11 @@ module RuboCop
       #   # good
       #   whatever_spec.rb         # describe MyClass, type: :routing do; end
       #
+      # @example `IgnoreMetadata: {type=>[routing,models]}` (default)
+      #   # good
+      #   whatever_spec.rb         # describe MyClass, type: :routing do; end
+      #   whatever_spec.rb         # describe MyClass, type: :models do; end
+      #
       class SpecFilePathFormat < Base
         include TopLevelGroup
         include Namespace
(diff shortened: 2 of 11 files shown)


spec/rubocop/cop/rspec/spec_file_path_format_spec.rb
```diff
@@ -281,4 +281,14 @@ class Foo
       RUBY
     end
   end
+
+  context 'when configured with `IgnoreMetadata: { "foo" => ["bar"] }`' do
```
bquorning, line 285: I think maybe you should write a spec for `IgnoreMetadata: { "foo" => ["bar", "baz"] }` and validate that it ignores specs with either value. And that it does not ignore a spec without another value for `foo`, or no `foo` metadata at all.

sriedel: Fair, are the amended specs what you had in mind? I needed to play around with the wording a bit, since I was over the line length limit by one and two characters respectively, by using the existing phrasing.

config/default.yml
```diff
@@ -925,6 +925,7 @@ RSpec/SpecFilePathFormat:
   IgnoreMetadata:
     type: routing
   VersionAdded: '2.24'
+  VersionChanged: "<<next>>"
```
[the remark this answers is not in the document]
sriedel: Sure, however I'd like to confirm that the revert would be correct, as the Pull Request template specifically states:

```

```

and this PR does change an existing cops configuration options.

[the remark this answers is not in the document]
ydah: According to my understanding, "modified an existing cop's configuration options" refers to instances such as when adding a new configuration option, changing a name, or for example, when adding autocorrect functionality and adding `Autocorrect: true`, you should add `VersionChanged: "<<next>>"`.

lib/rubocop/cop/rspec/spec_file_path_format.rb
```diff
@@ -73,7 +78,10 @@ def ensure_correct_file_path(send_node, class_name, arguments)
         def ignore_metadata?(arguments)
           arguments.any? do |argument|
             metadata_key_value(argument).any? do |key, value|
-              ignore_metadata.values_at(key.to_s).include?(value.to_s)
+              ignore_values = ignore_metadata[key.to_s]
+              ignore_values = [ignore_values] unless ignore_values.is_a?(Array)
```
bquorning, line 82: You can use `Array()` instead:
```suggestion
              ignore_values = Array(ignore_metadata[key.to_s])
```

bquorning (approved):

ydah:
Could you rebase to the latest master branch?

sriedel:
I seem to have fat-fingered the rebase. I'll close this PR and reopen a new one with the current state of changes tomorrow.

sriedel:
Clean rebase PR in #2076

PR #2074, opened 2025-04-04
Add new `RSpec/ChangeWithoutExpect` cop

ydah:
Follow up: https://github.com/rubocop/rubocop-rspec/pull/2071

- [x] Added tests.
- [x] Updated documentation.


spec/rubocop/cop/rspec/change_without_expect_spec.rb
```diff
@@ -0,0 +1,107 @@
...
+        change(Counter, :count).from(0).by(1)
+      end
+    RUBY
+  end
+
+  it 'does not register an offense for a simple `change` call ' \
+     'without chains' do
+    expect_no_offenses(<<~RUBY)
+      it 'changes the count' do
+        change(Counter, :count)
```
pirj, line 67: Why doesn’t it register an offence?

lib/rubocop/cop/rspec/change_without_expect.rb
```diff
@@ -0,0 +1,119 @@
...
+        MSG = 'Use `change` matcher within an `expect` block.'
+        RESTRICT_ON_SEND = [:change].freeze
+        SINGLE_RESTRICTED_METHODS = %i[by by_at_least by_at_most from].freeze
+
+        # @!method expectation?(node)
+        def_node_search :expectation?, <<~PATTERN
+          (send
+            {
+              (block (send nil? :expect ...) ...)
+              (send nil? :expect ...)
```
pirj, line 64: Why do we match the non-block form here?

lib/rubocop/cop/rspec/change_without_expect.rb
```diff
@@ -0,0 +1,119 @@
...
+      #   it 'changes the count' do
+      #     expect { subject }.to change(Counter, :count).by(1)
+      #   end
+      #
+      class ChangeWithoutExpect < RuboCop::Cop::Base
+        include RangeHelp
+
+        MSG = 'Use `change` matcher within an `expect` block.'
+        RESTRICT_ON_SEND = [:change].freeze
+        SINGLE_RESTRICTED_METHODS = %i[by by_at_least by_at_most from].freeze
```
pirj, line 57: There’s also “to” https://github.com/rspec/rspec/blob/d064fd0e9f00af834dfcfbe55e74b066a76e98e6/rspec-expectations/lib/rspec/matchers/built_in/change.rb#L36

lib/rubocop/cop/rspec/change_without_expect.rb
```diff
@@ -0,0 +1,119 @@
...
+      #   it 'changes the count' do
+      #     change(Counter, :count).from(0).by(1)
+      #   end
+      #
+      #   # good - within expect block
+      #   it 'changes the count' do
+      #     expect { subject }.to change(Counter, :count).by(1)
+      #   end
+      #
+      class ChangeWithoutExpect < RuboCop::Cop::Base
```
pirj, line 52: Even though this would detect a number of offences, wouldn’t it make more sense to make a more broad cop that would detect all built-in block matchers? “output”, “raise_error”, …

pirj, line 52: Also, why only block matchers?

spec/rubocop/cop/rspec/change_without_expect_spec.rb
```diff
@@ -0,0 +1,107 @@
...
+
+  it 'does not register an offense for `change` with chains ' \
+     'within an `expect` with not_to' do
+    expect_no_offenses(<<~RUBY)
+      it 'does not change the count' do
+        expect { subject }.not_to change(Counter, :count).by(1)
+      end
+    RUBY
+  end
+end
```
pirj, line 107: How does it behave with a case with two statements? Like “subject; change(:foo, :bar)”?

pirj, line 107: What it the matcher is assigned to a variable, and then used multiple times?
“m = change { sum }
expect { foo }.to m
expect { bar }.to m
“
This is a legit RSpec construct, and should work fine?

pirj:
This is a long-standing issue in RSpec, and I've seen misleading `it { validate_presence_of(:name) }` specs so many times.

Still, it feels that it would be easer to implement the detection in the runtime, e.g. by comparing a list of initialized matchers against those that were called `match?`, and warning on the difference.

The spectrum of the inspection surface for such a cop, on the contrary, is near-impossible to detect due to two reasons:
1) there can be different ways to "run" the matcher as in `expect_model_to validate_presence_of(:name)`
2) abundance of custom and third-party matchers

Our [NoExpectationExample](https://github.com/rubocop/rubocop-rspec/blob/dc9346274ea7674f32ba04f353a6c7e420f6e69a/lib/rubocop/cop/rspec/no_expectation_example.rb#L23) cop seems to cover some of those issues.

Related: https://github.com/rspec/rspec-core/issues/591 https://github.com/rspec/rspec-core/issues/404 https://github.com/rspec/rspec/issues/5 https://github.com/rspec/rspec-expectations/issues/47

What part of this is the PR covering, and where it plays better than other means?

PR #2076, opened 2025-04-10
RSpec/SpecFilePathFormat allow multiple values for same key in IgnoreMetadata

sriedel:
Retry of PR #2070  - I seem to have fat-fingered the rebase to master in the previous PR. So here is a fresh PR based off of the current master with all the change requests of the original included.

Description of original PR:

RSpec/SpecFilePathFormat allows ignoring certain values in rspec metadata through the IgnoreMetadata option, as given in the documentation:

```
      # @example `IgnoreMetadata: {type=>routing}` (default)
      #   # good
      #   whatever_spec.rb         # describe MyClass, type: :routing do; end
```

However the current implementation does not seem to allow supplying multiple values for the same key - e.g. if you have some specs marked `type: controller`, others as `type: :model`  etc, only one of `controller`, `model`, ... can be supplied to be ignored.

This is probably due to an oversight (or me not being able to figure out how to properly supply an array of values to be ignored).

The PR allows for either an array or a single value to be given for a key in the IgnoreMetadata option.

While it would also be desirable to give a `true` value to ignore all metadata, in the format that `IgnoreMethods` use, this should probably be part of a different PR.

### DIFF
--- a/lib/rubocop/cop/rspec/spec_file_path_format.rb
+++ b/lib/rubocop/cop/rspec/spec_file_path_format.rb
@@ -32,6 +32,11 @@ module RuboCop
       #   # good
       #   whatever_spec.rb         # describe MyClass, type: :routing do; end
       #
+      # @example `IgnoreMetadata: {type=>[routing,models]}` (default)
+      #   # good
+      #   whatever_spec.rb         # describe MyClass, type: :routing do; end
+      #   whatever_spec.rb         # describe MyClass, type: :models do; end
+      #
       class SpecFilePathFormat < Base
         include TopLevelGroup
         include Namespace
@@ -73,7 +78,9 @@ module RuboCop
         def ignore_metadata?(arguments)
           arguments.any? do |argument|
             metadata_key_value(argument).any? do |key, value|
-              ignore_metadata.values_at(key.to_s).include?(value.to_s)
+              ignore_values = Array(ignore_metadata[key.to_s])
+
+              ignore_values.include?(value.to_s)
             end
           end
         end
--- a/docs/modules/ROOT/pages/cops_rspec.adoc
+++ b/docs/modules/ROOT/pages/cops_rspec.adoc
@@ -5897,6 +5897,16 @@ my_class_spec.rb         # describe MyClass, '#method'
 whatever_spec.rb         # describe MyClass, type: :routing do; end
 ----
 
+[#_ignoremetadata_-_type___routing_models___-_default_-rspecspecfilepathformat]
+==== `IgnoreMetadata: {type=>[routing,models]}` (default)
+
+[source,ruby]
+----
+# good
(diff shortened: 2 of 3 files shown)


pirj (changes requested):
I left some feedback on the previous PR.
Can you please check?
I’ll move my comments to this PR when I het back to my computer.

pirj:
I think the current approach will lead to ambiguous configuration, and we should do this differently. I haven’t heard counter arguments, so I’m setting a block until we figure out how to proceed.

lib/rubocop/cop/rspec/spec_file_path_format.rb
```diff
@@ -32,6 +32,11 @@ module RSpec
       #   # good
       #   whatever_spec.rb         # describe MyClass, type: :routing do; end
       #
+      # @example `IgnoreMetadata: {type=>[routing,models]}` (default)
+      #   # good
+      #   whatever_spec.rb         # describe MyClass, type: :routing do; end
+      #   whatever_spec.rb         # describe MyClass, type: :models do; end
```
pirj, line 38: What if the metadata's value IS an array?
```ruby
RSpec.describe Airplane, prepare: [:fuel] do
```
How do we define IgnoreMetadata in YAML?

pirj, line 38: I propose:
```yaml
IgnoreMetadata:
  - prepare # ignore by presence of a key
  - type: model # ignore one value
  - type: routing # and another, too
  - skip: [database] # ignore when the value is an array of specific values
```

Luckily enough, YAML allows to mix all that, and to cover all ignored cases unambiguously.

pirj (changes requested):

sriedel:
> I think the current approach will lead to ambiguous configuration, and we should do this differently. I haven’t heard counter arguments, so I’m setting a block until we figure out how to proceed.

From my end the current approach is the simplest change to resolve the issue we are having with the status quo: that only a single value can be configured at all.

Any greater changes and refactorings are outside the scope that I'm seeing for this immediate ticket. While your objection is valid, that noone complained about it yet does imply that it's a rare edge case. 

Unfortunately I only have limited time availble for such fixes to external gems. If a non-resolving of the edge case is a showstopper, this is understandable. However since I cannot proceed in a timely fashion, I think a more effective solution on our end is to then close the PR and just disable the rule for our code base.

pirj:
> the current approach will lead to ambiguous configuration

By “current approach” I mean changes introduced in this PR. Not our existing code.
Thus, I request to change your code, not refactor the existing project code.
It’s totally up to you how to proceed with the PR. Please keep it open as a reminder that a draft solution exist. Someone may want to pick it up or use as a starting point if you have no intention to finalise the contribution.

PR #2077, opened 2025-04-11
Support `itblock`

Earlopain:
All cops that handle numblock now also handle itblock. `itblock` is relatively new and the required `rubocop-ast` version is required via RuboCop 1.75

- [x] Added tests.
- [ ] Updated documentation.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Master (Unreleased)
 
+- Add support for `itblock` from Ruby 3.4. ([@earlopain])
+
 ## 3.7.0 (2025-09-01)
 
 - Mark `RSpec/IncludeExamples` as `SafeAutoCorrect: false`. ([@yujideveloper])
--- a/lib/rubocop/cop/rspec/hooks_before_examples.rb
+++ b/lib/rubocop/cop/rspec/hooks_before_examples.rb
@@ -38,13 +38,11 @@ module RuboCop
           }
         PATTERN
 
-        def on_block(node)
+        def on_block(node) # rubocop:disable InternalAffairs/NumblockHandler
           return unless example_group_with_body?(node)
 
           check_hooks(node.body) if multiline_block?(node.body)
         end
-
-        alias on_numblock on_block
 
         private
 
--- a/lib/rubocop/cop/rspec/mixin/metadata.rb
+++ b/lib/rubocop/cop/rspec/mixin/metadata.rb
@@ -27,7 +27,7 @@ module RuboCop
           (send (lvar %) #Hooks.all _ $...)
         PATTERN
 
-        def on_block(node)
+        def on_block(node) # rubocop:disable InternalAffairs/NumblockHandler
           rspec_configure(node) do |block_var|
             metadata_in_block(node, block_var) do |metadata_arguments|
               on_metadata_arguments(metadata_arguments)
@@ -38,7 +38,6 @@ module RuboCop
             on_metadata_arguments(metadata_arguments)
           end
         end
-        alias on_numblock on_block
 
         def on_metadata(_symbols, _hash)
           raise ::NotImplementedError
--- a/rubocop-rspec.gemspec
+++ b/rubocop-rspec.gemspec
@@ -39,5 +39,5 @@ Gem::Specification.new do |spec|
   }
 
   spec.add_dependency 'lint_roller', '~> 1.1'
-  spec.add_dependency 'rubocop', '~> 1.72', '>= 1.72.1'
+  spec.add_dependency 'rubocop', '~> 1.75'
 end
--- a/spec/rubocop/cop/rspec/hook_argument_spec.rb
+++ b/spec/rubocop/cop/rspec/hook_argument_spec.rb
@@ -262,29 +262,59 @@ RSpec.describe RuboCop::Cop::RSpec::HookArgument do
     context 'when Ruby 2.7', :ruby27 do
       it 'does not flag :example for hooks' do
         expect_no_offenses(<<~RUBY)
-          around(:example) { _1 }
+          around(:example) { _1.run }
         RUBY
       end
 
       it 'detects :each for hooks' do
         expect_offense(<<~RUBY)
-          around(:each) { _1 }
+          around(:each) { _1.run }
           ^^^^^^^^^^^^^ Use `:example` for RSpec hooks.
         RUBY
 
         expect_correction(<<~RUBY)
-          around(:example) { _1 }
+          around(:example) { _1.run }
         RUBY
       end
 
       it 'detects hooks without default scopes' do
         expect_offense(<<~RUBY)
-          around { _1 }
+          around { _1.run }
           ^^^^^^ Use `:example` for RSpec hooks.
         RUBY
 
         expect_correction(<<~RUBY)
-          around(:example) { _1 }
+          around(:example) { _1.run }
+        RUBY
+      end
+    end
+
+    context 'when Ruby 3.4', :ruby34 do
+      it 'does not flag :example for hooks' do
+        expect_no_offenses(<<~RUBY)
+          around(:example) { it.run }
+        RUBY
+      end
+
+      it 'detects :each for hooks' do
+        expect_offense(<<~RUBY)
+          around(:each) { it.run }
+          ^^^^^^^^^^^^^ Use `:example` for RSpec hooks.
+        RUBY
+
+        expect_correction(<<~RUBY)
+          around(:example) { it.run }
+        RUBY
+      end
+
+      it 'detects hooks without default scopes' do
+        expect_offense(<<~RUBY)
+          around { it.run }
+          ^^^^^^ Use `:example` for RSpec hooks.
+        RUBY
+
+        expect_correction(<<~RUBY)
+          around(:example) { it.run }
         RUBY
       end
     end
--- a/spec/rubocop/cop/rspec/hooks_before_examples_spec.rb
+++ b/spec/rubocop/cop/rspec/hooks_before_examples_spec.rb
@@ -227,4 +227,65 @@ RSpec.describe RuboCop::Cop::RSpec::HooksBeforeExamples do
       RUBY
     end
   end
+
+  context 'when Ruby 3.4', :ruby34 do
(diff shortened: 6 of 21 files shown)


lib/rubocop/cop/rspec/hooks_before_examples.rb
```diff
@@ -38,14 +38,12 @@ class HooksBeforeExamples < Base
...
         PATTERN
 
-        def on_block(node)
+        def on_block(node) # rubocop:disable InternalAffairs/NumblockHandler
           return unless example_group_with_body?(node)
 
           check_hooks(node.body) if multiline_block?(node.body)
         end
 
-        alias on_numblock on_block
```
Earlopain, line 47: This had no effect, so I removed it. It relies on a previous block and then simply uses `next_sibling` together with `any_block` (which includes `itblock`). I only added tests to show it already works.

lib/rubocop/cop/rspec/mixin/metadata.rb
```diff
@@ -38,7 +38,6 @@ def on_block(node)
             on_metadata_arguments(metadata_arguments)
           end
         end
-        alias on_numblock on_block
```
Earlopain, line 41: This module is included in 4 cops and none have tests for numblocks. So I simply removed this one.

I don't think handling these blocks makes sense in context anyways and it was only added here because of the internal affairs cop

rubocop-rspec.gemspec
```diff
@@ -39,5 +39,5 @@ Gem::Specification.new do |spec|
   }
 
   spec.add_dependency 'lint_roller', '~> 1.1'
-  spec.add_dependency 'rubocop', '~> 1.72', '>= 1.72.1'
+  spec.add_dependency 'rubocop', '~> 1.75'
```
Earlopain, line 42: 1.75 transitively depends on `rubocop-ast >= 1.43.0` which is good enough

Earlopain:
I guess I should clarify: `itblock` is `foo { it }`, where `it` is the same as `_1` for `numblock`. It's a separate node type.

spec/spec_helper.rb
```diff
@@ -44,6 +44,11 @@ module SpecHelper
 
   config.include_context 'with default RSpec/Language config', :config
   config.include_context 'smoke test', type: :cop_spec
+
+  # whitequark/parser is syntax compatible with 2.0 to 3.3.
+  if ENV['PARSER_ENGINE'] != 'parser_prism'
```
Earlopain, line 49: The parser gem doesn't support 3.4-specific syntax. RuboCop now always uses prism for these versions, so some tests that use newer syntax may fail with `parser`.

Earlopain:
Hm, coverage is reduced because when using the `parser` gem `on_itblock` will never get called (rather, its interpreted as a plain `block` node). Any tips on how to solve this?

spec/rubocop/cop/rspec/hook_argument_spec.rb
```diff
@@ -288,5 +288,35 @@
         RUBY
       end
     end
+
+    context 'when Ruby 3.4', :ruby34, unsupported_on: :parser do
+      it 'does not flag :example for hooks' do
+        expect_no_offenses(<<~RUBY)
```
pirj, line 294: Can you please change this to ‘it.run’ not to offend the other cop?

spec/rubocop/cop/rspec/hooks_before_examples_spec.rb
```diff
@@ -227,4 +227,65 @@
       RUBY
     end
   end
+
+  context 'when Ruby 3.4', :ruby34 do
+    it 'flags `around` after `it`' do
+      expect_offense(<<~RUBY)
+        RSpec.describe User do
+          it { is_expected.to be_after_around_hook }
+          around { it }
```
pirj, line 236: I wonder if this would work anyway without making a special case for an it-block? Looks like a regular no-args block call with an ‘it’ call inside. No surprise, this may or may not be an it-block, hard to tell.

Side note, how does Ruby knows ‘describe { it { foo } }’ is not an it-block?

pirj: By what I propose here to keep the spec, and attempt to remove special handling in the cop. Chances are it will work anyway.

spec/spec_helper.rb
pirj: If we remove this, what fails?

The ultimate goal is to be running those new specs at least in one CI job. What job would it be?

We can still check against 3.4 syntax and use prism even if we run ‘rubocop’ in 3.1, right?

How do we set the “Ruby version of the code being investigated”? Same as the runtime Ruby?

Earlopain: > If we remove this, what fails?

I was a bit overeager with adding the excludes. There are only two failures for two cops, so I removed it from the others. The others pass by chance because `itblock` is simply `block` with zero block arguments from the perspective of `parser`.

> The ultimate goal is to be running those new specs at least in one CI job. What job would it be?

It will run only in the `prism` job from `workflows/main.yml` since it will fail on the `parser` one. It sets `PARSER_ENGINE=parser_prism` which is used by the specs.

> We can still check against 3.4 syntax and use prism even if we run ‘rubocop’ in 3.1, right?
> How do we set the “Ruby version of the code being investigated”? Same as the runtime Ruby?

These two points I don't understand what you are asking, can you elaborate?

spec/rubocop/cop/rspec/hook_argument_spec.rb
Earlopain: I copy-pasted the current numblock specs, will change these as well.

spec/rubocop/cop/rspec/hooks_before_examples_spec.rb
Earlopain: Yup, that's what I did: https://github.com/rubocop/rubocop-rspec/pull/2077#discussion_r2039275425

> Side note, how does Ruby knows ‘describe { it { foo } }’ is not an it-block?

`it` with block had special considerations from ruby itself, `it` only points to the block argument when `it` doesn't take a block itself. https://bugs.ruby-lang.org/issues/18980
> The following cases have been discussed:
>  *  it method, most famously in RSpec: You almost always pass a positional and/or block argument to RSpec's it, so the conflict is avoided with my proposal. You virtually never use a completely naked it

Earlopain:
Coverage is no longer an issue (https://github.com/rubocop/rubocop/pull/14119) but now there's a gnarly-looking ruby-head failure I'm 100% certain is not my fault (:

spec/spec_helper.rb
Earlopain: No longer needed, I removed this code

lib/rubocop/cop/rspec/mixin/metadata.rb
bquorning: I think it would be helpful to have some guidelines on when to add numblock/itblock aliases, and when it does not. It's too easy to appease toe internal affairs cop and just add a useless alias.

Earlopain: In this case it's for `describe`/`context`. You could handle numblocks but what would taking a block argument even do for those functions? There are a few other cases like this that I can think of like `prepend`/`include`/`extend` that take a block but never any arguments.

I guess as long as the block actually takes an argument. If a use can write `foo { |arg| }` and have the code make sense, then the same with numblock/itblock should also be possible.

pirj: I recall we had a PR where we were adding aliases everywhere, and then we just disabled that cop, and decided not to merge the pr. I yet to see any missed offence caused by lack of those aliases to justify their addition.

CHANGELOG.md
```diff
@@ -10,6 +10,7 @@
 - Change `RSpec/ScatteredSetup` to allow `around` hooks to be scattered. ([@ydah])
 - Fix an error `RSpec/ChangeByZero` cop when without expect block. ([@lee266])
 - Fix a false positive for `RSpec/DescribedClass` when `SkipBlocks` is true and numblocks are used. ([@earlopain])
+- Add support for `itblock` from Ruby 3.4. ([@earlopain])
```
bquorning: Oops, this line should be moved to the Unreleased section.

Earlopain: Oops indeed. I rebased.

lib/rubocop/cop/rspec/mixin/metadata.rb
Earlopain: I can see how that cop can be annoying. Personally I never use numblocks (and I think I'm not the only one, just doesn't look very nice) but I do actually write `it`.

pirj:
Understood, thanks for detailed explanation.

> not relevant when compared to the other 3 types

But some cop like “a blank line around hooks” may want to consider the blockpass form of hooks, too?

Is there a good reason why there is no `on_any_block` at least for those blocks with body? Would rubocop-rspec at least benefit from having it?
We can define it in our own code, and delegate from on_block/on_numpass/on_itblock.
Or do those cops that you’ve fixed need to handle those flavours of blocks differently?

Earlopain:
> But some cop like “a blank line around hooks” may want to consider the blockpass form of hooks, too?

Hm, yeah. I see. Yes, I think that may be the case. I usually see `on_send` with `RESTRICT_ON_SEND` used so you don't even have to really consider all the different block types but since everything is so dynamic here that makes it a but more difficult.

Sharing the implementation with `on_block` etc. is not feasable as they are quite different in shape.

> Is there a good reason why there is no on_any_block at least for those blocks with body?

I guess nobody really wanted it yet. Usually you can get away with reusing the same implementation for `on_block`/`on_numblock`/`on_itblock` but it's not always the case. Right now I aliases it 6 times and 2 times it has its own distinct but somewhat similar implementation.

I had the same though a few times as well. In rubocop itself this also happens repeatedly so it wouldn't make much sense to just add it it rubocop-rspec I think. Same also for `send`/`csend`. I guess eventually I'd like this to just be handled via some `InternalAffairs` cop.

Earlopain:
Let me just close this, it has been a while. Should this become relevant I'm happy to rebase, just let me know.

PR #2079, opened 2025-04-14
Consider prism in coverage results

Earlopain:
In https://github.com/rubocop/rubocop-rspec/pull/2077, I want to add prism-only logic that won't be executed by the `parser` gem. Coverage complains, because it thinks some code is unreachable.

This PR fixes this by making a few changes to CI, with the goal to merge the coverage results so it can get back to 100%.

I made the following changes:
* Move internal_investigation out of the matrix and run it only once. `internal_investigtion` is not really for catching `rubocop-rspec` errors, there's the test suite for that. So running it only once is ok (I made the same change in `rubocop` a while ago). Doing this simplifies merging since only a single task is executed.
* Move prism to the main workflow. There's nothing that needs to be considered separately anymore, `rubocop-ast` pulls prism in. Also simplifies merging coverage
* Add a new rake task to merge coverage results. These are uploaded per each matrix job, and when all is done a separate job merges them and checks for coverage drops etc.

If this is merged, I can rebase https://github.com/rubocop/rubocop-rspec/pull/2077 and CI should be green there. I tested this locally with coverage results from the different parser engines and it reports 100% coverage after merging and behaves like I would expect it to.

### DIFF
--- a/.github/workflows/linting.yml
+++ b/.github/workflows/linting.yml
@@ -5,6 +5,17 @@ concurrency:
   group: ${{ github.workflow }}-${{ github.head_ref || github.run_id }}
   cancel-in-progress: true
 jobs:
+  internal_investigation:
+    runs-on: ubuntu-latest
+    name: Internal Investigation
+    steps:
+      - uses: actions/checkout@v4
+      - uses: ruby/setup-ruby@v1
+        with:
+          ruby-version: ruby # Latest stable CRuby version
+          bundler-cache: true
+      - run: bundle exec rake internal_investigation
+
   yamllint:
     name: Yamllint
     runs-on: ubuntu-latest
@@ -18,6 +29,7 @@ jobs:
           yamllint_comment: true
         env:
           GITHUB_ACCESS_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
   mdformat:
     name: Mdformat
     runs-on: ubuntu-latest
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -35,17 +35,14 @@ jobs:
           - "3.4"
           - ruby-head
           - jruby-9.4
-        task:
-          - internal_investigation
-          - spec
-    name: "Ruby ${{ matrix.ruby }}: ${{ matrix.task }}"
+    name: "Ruby ${{ matrix.ruby }} - Spec"
     steps:
       - uses: actions/checkout@v4
       - uses: ruby/setup-ruby@v1
         with:
           ruby-version: "${{ matrix.ruby }}"
           bundler-cache: true
-      - run: NO_COVERAGE=true bundle exec rake ${{ matrix.task }}
+      - run: NO_COVERAGE=true bundle exec rake spec
 
   coverage:
     runs-on: ubuntu-latest
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -47,20 +47,33 @@ jobs:
         with:
           ruby-version: "${{ matrix.ruby }}"
           bundler-cache: true
-      - run: NO_COVERAGE=true bundle exec rake spec
+      - run: bundle exec rake spec
         env:
           PARSER_ENGINE: ${{ matrix.parser_engine }}
+      - name: Upload Coverage Artifact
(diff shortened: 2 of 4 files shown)


.github/workflows/linting.yml
```diff
@@ -5,6 +5,17 @@ concurrency:
...
+  internal_investigation:
+    runs-on: ubuntu-latest
+    name: Internal Investigation
+    steps:
+      - uses: actions/checkout@v4
+      - uses: ruby/setup-ruby@v1
+        with:
+          ruby-version: ruby # Latest stable CRuby version
+          bundler-cache: true
+      - run: NO_COVERAGE=true bundle exec rake internal_investigation
```
pirj, line 17: Internal investigation shouldn't be bothered by NO_COVERAGE, or why is it?

.github/workflows/main.yml
```diff
@@ -35,28 +35,46 @@ jobs:
...
       - uses: actions/checkout@v4
       - uses: ruby/setup-ruby@v1
         with:
           ruby-version: "${{ matrix.ruby }}"
           bundler-cache: true
-      - run: NO_COVERAGE=true bundle exec rake ${{ matrix.task }}
+      - run: bundle exec rake spec
+        env:
+          PARSER_ENGINE: ${{ matrix.parser_engine }}
+      - name: Upload Coverage Artifact
```
pirj, line 53: I think we're quite fine if it just fails.
Also, it's fine if it runs for just one Ruby version, which is done below in the "coverage" job (line 50)

@rubocop/rubocop-rspec WDYT?

.github/workflows/main.yml
```diff
@@ -35,28 +35,46 @@ jobs:
...
           - jruby-9.4
-        task:
-          - internal_investigation
-          - spec
-    name: "Ruby ${{ matrix.ruby }}: ${{ matrix.task }}"
+        parser_engine:
+          - parser_whitequark
+        include:
+          - ruby: "2.7" # Minimum version required for Prism to run.
+            parser_engine: parser_prism
```
pirj, line 42: https://github.com/rubocop/rubocop/blob/d8448a3790a78f16536e267ed7ea5ffc2e03561b/docs/modules/ROOT/pages/compatibility.adoc#L44

> Starting from RuboCop 1.75, the prism translation layer is used by default when analyzing Ruby 3.4+.

What is the point for us to run this job on 2.7, and tell the Prism to analyze the code as it was Ruby 3.4? Don't we have those 3.4-specific (eg it-block) specs disabled when running on Ruby 2.7?

Is it just my OCD, or we need two Prism jobs, one on 2.7, to ensure it runs on older rubies, and one on 3.4, to ensure cops do what we expect them to, including new features, with Prism?

tasks/coverage.rake
```diff
@@ -0,0 +1,10 @@
+# frozen_string_literal: true
+
+namespace :coverage do
+  desc 'Report Coverage from merged CI runs'
+  task :ci do
+    require 'simplecov'
+
+    SimpleCov.collate Dir['coverage-*/.resultset.json']
```
pirj, line 8: Will this combine results from different Ruby versions, and Parser results with Prism results?

.github/workflows/linting.yml
Earlopain: You're right, I had this from the main job where I extracted this from. Will remove

.github/workflows/main.yml
Earlopain: > I think we're quite fine if it just fails.

Yeah, I can do that. `if-no-files-found` was a bit nice during debuging but not really necessary anymore (plus jruby kinda needs that, or something similar)

tasks/coverage.rake
Earlopain: Yes. Different ruby versions aren't really needed since it will be at 100% with just two jobs right now (one prism, one parser) but limiting it in the workflow adds complexity which I don't think is really needed.

Just doing it will all runs seems ok to me for simplicity.

Earlopain:
This mess is no longer needed since 3.4 syntax is now always parsed via prism: https://github.com/rubocop/rubocop/pull/14119

PR #2080, opened 2025-04-17
Remove runtime dependency on lint_roller

bquorning:
rubocop-rspec should not need to have a runtime dependency on lint_roller. And since RuboCop v1.72.1 already has a dependency on lint_roller `~> 1.1.0`, it should be safe to remove from our gemspec.

- [ ] Added tests.
- [ ] Updated documentation.

### DIFF
--- a/rubocop-rspec.gemspec
+++ b/rubocop-rspec.gemspec
@@ -38,6 +38,5 @@ Gem::Specification.new do |spec|
     'default_lint_roller_plugin' => 'RuboCop::RSpec::Plugin'
   }
 
-  spec.add_dependency 'lint_roller', '~> 1.1'
   spec.add_dependency 'rubocop', '~> 1.72', '>= 1.72.1'
 end


bquorning:
cc @paracycle

pirj (approved):
No objections 
I wonder, if we had it, do we use it?
I’m a tad concerned if RuboCop drops or replaces it, will it still work for us?

bquorning:
I don't think RuboCop will stop it, then the whole plugin loading would stop working.

But I would like @koic's comment. We spoke briefly about it today at RubyKaigi.

bquorning:
I’ll move this PR to draft while I rethink some problems with lint_roller…

koic:
Unless something has been overlooked, declaring a runtime dependency on `lint_roller` in the gemspec would be a safer way to stabilize the dependency on the intermediate API layer provided by `lint_roller`.

Although the likelihood is low, if RuboCop were to depend on a future version such as `lint_roller 2.x` that introduces breaking changes, an incompatibility might arise between `lint_roller 2.x` (required by RuboCop) and `lint_roller 1.x` (required by a plugin, such as RuboCop RSpec).

By specifying the dependency in the gemspec, potential errors caused by an unexpected update to `lint_roller` during `bundle update` can be avoided. This is because both RuboCop and the plugin would remain dependent on the `lint_roller 1.x` API until the plugin explicitly resolves compatibility with `lint_roller 2.x`.

When removing the dependency on `lint_roller`, it is necessary to consider whether the plugin can accept the risk of version mismatch should such a situation occur.

bquorning:
I realize that currently [this line](https://github.com/rubocop/rubocop-rspec/blob/b0fda476f90aab690d68aed9f8276bb495ac6b8d/lib/rubocop-rspec.rb#L11) prevents us from removing lint_roller as a dependency – by always loading plugin.rb, lint_roller _is_ a direct dependency.

> Although the likelihood is low, if RuboCop were to depend on a future version such as `lint_roller 2.x` that introduces breaking changes, an incompatibility might arise between `lint_roller 2.x` (required by RuboCop) and `lint_roller 1.x` (required by a plugin, such as RuboCop RSpec).

If the version 2 API is incompatible with the current API, I would expect lint_roller to also change what needs to be entered into our [gemspec metadata](https://github.com/rubocop/rubocop-rspec/blob/b0fda476f90aab690d68aed9f8276bb495ac6b8d/rubocop-rspec.gemspec#L38). Perhaps they would use a `'default_lint_roller_v2_plugin'` key? That way we could ship _one_ gem with support for _two_ plugin systems.

> By the way, @bquorning, I was glad to meet you at RubyKaigi 2025!

Likewise 🙏🏼

bquorning:
As a hypothetical / abstract example, what if we wanted rubocop-rspec to also be able to be a plugin is another linter project which uses a different way to write plugins. Their plugin system adds not one, but _three_ runtime dependencies. Should we then allow rubocop-rspec to have 4 extra runtime dependencies (lint_roller, plus these 3 new ones)? I don’t think so.

I think if our gemspec metadata defined a path instead of a class name, rubocop would be able to load our plugin code, and rubocop-rspec would not have any knowledge (besides defining, but not loading, a `Plugin` class) of the plugin system.

PR #2085, opened 2025-05-04
CI: Add JRuby 10.0 and 10.1

bquorning:
Should we add JRuby 10.0 and 10.1 to the CI matrix?

JRuby 10.0 targets CRuby 3.4 support, and 10.1 targets CRuby 4.0 support.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -36,6 +36,8 @@ jobs:
           - "4.0"
           - ruby-head
           - jruby-9.4
+          - jruby-10.0
+          - jruby-10.1
         task:
           - internal_investigation
           - spec
--- a/Gemfile
+++ b/Gemfile
@@ -5,14 +5,13 @@ source 'https://rubygems.org'
 gemspec
 
 gem 'bump'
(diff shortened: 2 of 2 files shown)


pirj:
> CI / Ruby 3.4 Successful in 27s
> CI / Ruby jruby-9.4 Successful in 1m
> CI / Ruby jruby-10.0 Successful in **2m**

🥲

bquorning:
cc @ydah, @pirj

.github/workflows/main.yml
```diff
@@ -36,6 +36,8 @@ jobs:
           - "4.0"
           - ruby-head
           - jruby-9.4
+          - jruby-10.0
+          - jruby-10.1
```
pirj, line 40: Maybe keep just the latest? I recall that they recommend stick to the latest ones.

pirj (approved):
Thank you!

.github/workflows/main.yml
bquorning: I added both because they target different scrubby versions. On the JRuby download page, all three are listed so I assume they are all stable.

PR #2088, opened 2025-05-24
Recognize stubbing of `described_class` in `RSpec/SubjectStub`

lovro-bikic:
This PR makes stubbing of `described_class` register an offense for `RSpec/SubjectStub` cop. I think this fits under this cop since its purpose is to prevent stubbing the system under test, and `described_class` is just that.

To give a production code example offense: I recently noticed a spec for a module that had a `allow(described_class).to receive(:some_module_method)`.

- [x] Added tests.
- [x] Updated documentation.

### DIFF
--- a/lib/rubocop/cop/rspec/subject_stub.rb
+++ b/lib/rubocop/cop/rspec/subject_stub.rb
@@ -38,6 +38,16 @@ module RuboCop
       #     end
       #   end
       #
+      #   # bad - described_class stubs are recognized as well
+      #   describe Article do
+      #     it 'indicates that the author is unknown' do
+      #       article = double(Article, description: 'by an unknown author')
+      #       allow(described_class).to receive(:new).and_return(article)
+      #
+      #       expect(Article.new.description).to include('by an unknown author')
+      #     end
+      #   end
+      #
       #   # good
       #   describe Article do
       #     subject(:article) { Article.new(author: nil) }
@@ -141,7 +151,7 @@ module RuboCop
           subject_names = [*subject_names, *@explicit_subjects[node]]
           subject_names -= @subject_overrides[node] if @subject_overrides[node]
 
-          names = Set[*subject_names, :subject]
+          names = Set[*subject_names, :subject, :described_class]
           expectation_detected = message_expectation?(node, names)
           return yield(node) if expectation_detected
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Master (Unreleased)
 
+- Recognize stubbing of `described_class` in `RSpec/SubjectStub`. ([@lovro-bikic])
+
 ## 3.6.0 (2025-04-18)
 
 - Fix false positive in `RSpec/Pending`, where it would mark the default block `it` as an offense. ([@bquorning])
--- a/docs/modules/ROOT/pages/cops_rspec.adoc
+++ b/docs/modules/ROOT/pages/cops_rspec.adoc
@@ -6096,6 +6096,16 @@ describe Article do
   end
 end
 
+# bad - described_class stubs are recognized as well
+describe Article do
+  it 'indicates that the author is unknown' do
+    article = double(Article, description: 'by an unknown author')
+    allow(described_class).to receive(:new).and_return(article)
+
+    expect(Article.new.description).to include('by an unknown author')
+  end
+end
+
 # good
 describe Article do
   subject(:article) { Article.new(author: nil) }
--- a/spec/rubocop/cop/rspec/subject_stub_spec.rb
+++ b/spec/rubocop/cop/rspec/subject_stub_spec.rb
@@ -364,6 +364,17 @@ RSpec.describe RuboCop::Cop::RSpec::SubjectStub do
     RUBY
   end
 
+  it 'flags when described_class is mocked' do
+    expect_offense(<<~RUBY)
+      describe Foo do
+        it 'uses described_class' do
+          expect(described_class).to receive(:bar).and_return(baz)
+          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ Do not stub methods of the object under test.
+        end
+      end
+    RUBY
+  end
+
   it 'flags when there are several top level example groups' do
     expect_offense(<<~RUBY)
       RSpec.describe Foo do


lib/rubocop/cop/rspec/subject_stub.rb
```diff
@@ -38,6 +38,16 @@ module RSpec
       #     end
       #   end
       #
+      #   # bad - described_class stubs are recognized as well
+      #   describe Article do
+      #     it 'indicates that the author is unknown' do
+      #       article = double(Article, description: 'by an unknown author')
+      #       allow(described_class).to receive(:new).and_return(article)
```
pirj, line 45: In general, this is useful. Why is this bad?

how is this different from `allow(Article)`?

I see that in a spec dedicated to Article it may be strange to mock its methods, specifically the initializer, or other static ones.
However, how do you otherwise test static methods that call other static methods in isolation? (Considering those static methods arr the subject under test, noth the whole class).


Can you probably give a more realistic example of bad code that made you extend this cop in the forst place?

lovro-bikic: > how is this different from allow(Article)?

It's the same as `allow(Article)`, but note that there's another cop [RSpec/DescribedClass](https://docs.rubocop.org/rubocop-rspec/cops_rspec.html#rspecdescribedclass) which would autocorrect `Article` to `described_class`.

In fact, I think `allow(Article)` should be registered as an offense as well since it's also the subject under test (because not everyone might have RSpec/DescribedClass enabled). I can work on this as a separate PR.

> In general, this is useful. Why is this bad?

Let's say your subject is `subject(:article) { Article.new }`. This cop will currently register this:
```ruby
allow(article).to receive(:foo)
```
as an offense.

However, it won't register:
```ruby
allow_any_instance_of(described_class).to receive(:foo)
```
as an offense, which is quite similar to the above (except the former will stub only one instance, and the latter will stub any instance).

In both scenarios, we are stubbing a method of the object under test.

lovro-bikic: > Can you probably give a more realistic example of bad code that made you extend this cop in the forst place?

Let's say we have a simple module:
```ruby
module TokenGenerator
  extend self

  def generate
    { token: random_token }
  end

  private

  def random_token
    SecureRandom.uuid
  end
end
```

and a spec like:
```ruby
RSpec.describe TokenGenerator do
  before do
    allow(described_class).to receive(:random_token).and_return('123')
  end

  it 'creates a token' do
    expect(TokenGenerator.generate).to eq({ token: '123' })
  end
end
```

In my view, this is not a good spec because we're not letting the module execute the code we're supposedly testing. A better spec would be:
```ruby
RSpec.describe TokenGenerator do
  it 'creates a token' do
    expect(TokenGenerator.generate).to include({ token: be_a(String) })
  end
end
```
which goes through the whole code path of the module.

pirj: It’s an easy one.
1. Smells because of stubbing of private methods
2. `be_a(String)` won’t cut it - if someone mistakenly replaces the call to `uuid` with a static `"not really random"`, the test won’t catch it. You have to stub, but not the private method, but the external `uuid`.

Stubbing the class under test’s public methods though may be an indication of a smell or may not. I don’t feel we should always flag corresponding tests as offenses.


Do you have other examples, preferably real life ones?

lovro-bikic: This is a real-life example 😄 

If `TokenGenerator` were a class instead, this cop would register an offense:
```ruby
RSpec.describe TokenGenerator do
  subject(:generator) { TokenGenerator.new }

  before do
    allow(generator).to receive(:random_token).and_return('123')
#   ^^^^^^^^^^^^^^^^ RSpec/SubjectStub: Do not stub methods of the object under test.
  end
end
```

Can you explain to me why should this scenario be an offense under this cop, but the `module TokenGenerator` example shouldn't?

pirj: Fair enough!

But is it still the same "subject under test" if we make assertions on an instsnce, but stub class methods? 🤔

lovro-bikic: That's a good question. With this PR and also https://github.com/rubocop/rubocop-rspec/pull/2094 I've expanded the cop to register offenses for not just the subject's methods (which are typically an instance's methods), but also any method of the class/module for any instance. I guess this gives more responsibility to this cop, which might or might not be desirable.

The [style guide](https://rspec.rubystyle.guide/#dont-stub-subject) shows this:
```ruby
describe 'Article' do
  subject(:article) { Article.new(author: nil) }

  it 'indicates that the author is unknown' do
    expect(article.description).to include('by an unknown author')
  end
end
```
as a good example, but the test could still stub the `#author` method to return `nil` in any of the following ways:
```ruby
before do
  allow_any_instance_of(Article).to receive(:author).and_return(nil)
  allow_any_instance_of(described_class).to receive(:author).and_return(nil)
end
```
and these would not be offenses under any RSpec cop, even though they are in practice similar to:
```ruby
subject(:article) { Article.new }

before do
  allow(article).to receive(:author).and_return(nil)
end
```
which is an offense.

Personally, I think `(allow|expect)_any_instance_of` and `described_class` should register offenses _somewhere_, much like `RSpec/SubjectStub` does, but perhaps this should be a different cop (e.g. `RSpec/DescribedClassStub`)?

pirj: Your argument related to the usage of `*_any_instance_of` is convincing.

It’s tougher with `described_class`. I especially like [this comment](https://github.com/rspec/rspec/issues/31#issuecomment-2504679783).

lovro-bikic: > It’s tougher with described_class. I especially like https://github.com/rspec/rspec/issues/31#issuecomment-2504679783.

True (I also got bit by this when I once used `describe :symbol` foolishly), but then again, there's the [`RSpec/DescribedClass`](https://docs.rubocop.org/rubocop-rspec/cops_rspec.html#rspecdescribedclass) cop which enforces usage of `described_class`, including for stubs. Couldn't offenses for stubbing `described_class` follow the same logic as that cop?

pirj:
A few [real-world](https://github.com/pirj/real-world-rspec) examples that would offend the cop:

brew/Library/Homebrew/test/bundle/brew_services_spec.rb
```
RSpec.describe Homebrew::Bundle::BrewServices do
...
        allow(described_class).to receive(:started_services).and_return(%w[nginx])
        ...
        expect(described_class.stop("nginx")).to be(true)
        expect(described_class.started_services).not_to include("nginx")
```
bad? seems so.
```
      allow(described_class).to receive(:`).with("cargo install --list").and_return(<<~EOS)
...
      EOS
```
They probably meant to stub Kernel. Not possible when calling backticks directly from a class? Can't recall.

There are hundreds of offences in Homebrew. Pretty sure they'll just mute the cop.

solidus:
```
  describe ".from_product" do
    it "has an empty variants count" do
      allow(described_class).to receive(:new)

      described_class.from_product(instance_double(Spree::Product, total_on_hand: 123, variants: []))

      expect(described_class).to have_received(:new).with(
        on_hand: 123,
        variants_count: 0,
      )
    end
  end
```
IDK.

Ruby-Toolbox:
```
    it "calls perform on the import instance" do
      allow(described_class).to receive(:new).and_return(import)
      expect(import).to receive(:perform)
      described_class.perform "data"
    end
```

```
      it "returns client instance configured with the URL" do
        the_instance = instance_double described_class

        allow(described_class).to receive(:new)
          .with(url: configured_url)
          .and_return(the_instance)

        expect(described_class.client).to be the_instance
      end
```

RuboCop:
```
RSpec.describe RuboCop::PendingCopsReporter do
...
        it 'prints a warning' do
          expect(described_class).to receive(:warn_on_pending_cops)
          report_pending_cops
        end
```

Puppet (hundreds)
```
  describe "::new" do
    it "returns NoGemsSource if rubygems is not present" do
      expect(described_class).to receive(:has_rubygems?).and_return(false)
      expect(described_class.new).to be_kind_of(Puppet::Util::RubyGems::NoGemsSource)
    end

    it "returns Gems18Source if Gem::Specification responds to latest_specs" do
      expect(described_class).to receive(:has_rubygems?).and_return(true)
      expect(described_class.new).to be_kind_of(Puppet::Util::RubyGems::Gems18Source)
    end
  end
```

```
      it "returns the version hash if the package was found" do
        expect(described_class).to receive(:check_updates).with([], [], []).once.and_return(latest_versions)
        version = described_class.latest_package_version(name, [], [], [])
        expect(version).to eq(mypackage_version)
      end
```

Mastodon:
```
    context 'when Rails.cache does not exists' do
      before do
        allow(described_class).to receive(:default_settings).and_return(default_settings)

        Fabricate(:setting, var: key, value: 42) if save_setting

        Rails.cache.delete(cache_key)
      end
```

```
    it 'calls .find_or_create_by' do
      allow(described_class).to receive(:find_or_create_by)

      described_class.find_for_omniauth(auth)

      expect(described_class).to have_received(:find_or_create_by).with(uid: auth.uid, provider: auth.provider)
    end
```

GitLab:
```
        it 'logs completed row count and does not enqueue next batch' do
          allow(instance).to receive(:paginate).and_return(paginator)
          expect(instance).to receive(:log_extra_metadata_on_done).with(:result, status: :completed, updated_rows: 1)
          expect(described_class).not_to receive(:perform_async)

          perform
        end
```

Chatwoot:
```
      it 'returns static tool instance for non-custom tools' do
        scenario = create(:captain_scenario, assistant: assistant, account: account)
        allow(described_class).to receive(:resolve_tool_class).with('add_contact_note').and_return(
          Class.new do
            def initialize(_assistant); end
          end
        )

        tool_metadata = { id: 'add_contact_note' }
        tool_instance = scenario.send(:resolve_tool_instance, tool_metadata)
        expect(tool_instance).not_to be_nil
        expect(tool_instance).not_to be_a(Captain::Tools::HttpTool)
      end
```

Capybara:
```
    it 'should default to a proc that calls run_default_server' do
      mock_app = Object.new
      allow(described_class).to receive(:run_default_server).and_return(true)
      described_class.server.call(mock_app, 8000)
      expect(described_class).to have_received(:run_default_server).with(mock_app, 8000)
    end
```

Canvas-LMS:
```
        it "coerces truthy values to true" do
          yaml_content = <<~YAML
            overrides:
              account:
                #{global_account_id}: yes
          YAML
          allow(mock_dynamic_settings).to receive(:[]).and_return(yaml_content)
          expect(described_class).not_to receive(:rand)

          expect(described_class.use_graphql?(global_account_id, global_course_id)).to be true
        end
```

I just glanced over, but none of those hit me as horribly bad.

PR #2089, opened 2025-05-28
Mark `RSpec/IncludeExamples` as `SafeAutoCorrect: false`

### LINKED ISSUE #2083: RSpec/IncludeExamples is not safe to auto-correct
## Expected behavior

Rubocop should not auto-correct where the change could result in changes of behaviour unless the 'autocorrect all' option us used.

## Actual behavior

The `RSpec/IncludeExamples` cop is marked as safe to auto-correct. However, `include_examples` and `it_behaves_like` are not exactly equivalent, as is explained in [the documentation:](https://docs.rubocop.org/rubocop-rspec/cops_rspec.html#rspecincludeexamples)

* `include_examples` adds the content of the shared examples to the current context. This will include any memoized variables and the test suite may have been written to depend on this. This is precisely the poor design that the cop intends to prevent but this may be common in older-code bases.
* `it_behaves_like` creates a new context and so memoized variables are isolated.

A minimal example:

```ruby
RSpec.shared_examples 'test examples' do
  let(:var) { 99 }

  it { expect(var).to eq(99) }
end

RSpec.describe 'RSpec/IncludeExamples' do
  context 'with include_examples' do
    include_examples 'test examples'

    # This test will pass
    it { expect(var).to eq(99) }
  end

  context 'with it_behaves_like' do
    it_behaves_like 'test examples'

    # This test will fail
    it { expect(var).to eq(99) }
  end
end
```

## RuboCop RSpec version

```
% bundle exec rubocop -V
1.75.3 (using Parser 3.3.7.3, Prism 1.4.0, rubocop-ast 1.44.1, analyzing as Ruby 3.4, running on ruby 3.4.2) [arm64-darwin24]
  - rubocop-rspec 3.6.0
```

yujideveloper:
resolves #2083

- [ ] Added tests.
- [ ] Updated documentation.


bquorning (approved):

pirj:
All good, I support the change of marking the cop as unsafe for autocorrect.

yujideveloper:
@koic @pirj @bquorning 
Thank you for your review.
I've updated the document the safety concerns using `@safety`.

PR #2090, opened 2025-06-02
Let `RSpec/SpecFilePathFormat` leverage ActiveSupport inflections, if…

### LINKED ISSUE #740: Proposal: Avoid repetitive definitions of `CustomTransform` in `RSpec/FilePath`

corsonknowles:
… defined.

Fix #740

- [ ] Added tests.
- [ ] Updated documentation.


corsonknowles:
`bundle exec rake` is failing on the primary branch, unless I am missing something

corsonknowles:
Alernatively: We could remove all the handling to cache the lookup by rescuing and warning on a load error when the user has configured to us Inflectors but they aren't available.
The current approach fails gracefully when configured on, but unavailable.

config/default.yml
```diff
@@ -924,6 +924,8 @@ RSpec/SpecFilePathFormat:
   IgnoreMethods: false
   IgnoreMetadata:
     type: routing
+  InflectorPath: "./config/initializers/inflections.rb"
```
corsonknowles, line 942: most should never need to set this, but just in case

lib/rubocop/cop/rspec/spec_file_path_format.rb
```diff
@@ -32,13 +32,22 @@ module RSpec
...
+      #
       class SpecFilePathFormat < Base
         include TopLevelGroup
         include Namespace
         include FileHelp
 
         MSG = 'Spec path should end with `%<suffix>s`.'
 
+        # Class instance variable to cache ActiveSupport availability
+        @activesupport_available = nil
```
corsonknowles: `@@` was a violation

lib/rubocop/cop/rspec/spec_file_path_format.rb
```diff
@@ -57,6 +66,28 @@ def on_top_level_example_group(node)
...
+            return false unless File.exist?(inflections)
+
+            require 'active_support/inflector'
+            require inflections
+
+            true
+          rescue LoadError, NameError
+            false
+          end
+        rescue StandardError
```
corsonknowles: this separation in error handling is required by the Shadowing rule

bquorning: Yes, `NameError` is a subclass of `StandardError`. I think you can just change line 84 to `rescue LoadError, StandardError`.

lib/rubocop/cop/rspec/spec_file_path_format.rb
```diff
@@ -106,10 +137,29 @@ def expected_path(constant)
         end
 
         def camel_to_snake_case(string)
-          string
-            .gsub(/([^A-Z])([A-Z]+)/, '\1_\2')
-            .gsub(/([A-Z])([A-Z][^A-Z\d]+)/, '\1_\2')
-            .downcase
+          if activesupport_inflections_available?
```
bquorning: It was mentioned in #1266 that other libraries than RuboCop and ActiveSupport can be used to set inflections. I don’t say that we should implement that now (or later), but perhaps the code in this method should a bit more modular? E.g.
```rb
        module ActiveSupportInflector
          def self.call(string)
            ActiveSupport::Inflector.underscore(string)
          end
        end

        module DefaultInflector
          def self.call(string)
            string
              .gsub(/([^A-Z])([A-Z]+)/, '\1_\2')
              .gsub(/([A-Z])([A-Z][^A-Z\d]+)/, '\1_\2')
              .downcase
          end
        end
```
The `ActiveSupportInflector` module/class could also have an `#available?` method that is all the cop needs to call before setting an `@inflector` instance variable?

corsonknowles: I love the look of this. It took quite some re-doing for the specs.

corsonknowles: Should be good to go now!

lib/rubocop/cop/rspec/spec_file_path_format.rb
corsonknowles: Fixed!

config/default.yml
```diff
@@ -924,6 +924,8 @@ RSpec/SpecFilePathFormat:
...
+  UseActiveSupportInflections: false
```
bquorning: I am not saying that we should support more inflectors, but just in case: Should we rename this to e.g. `Inflector: (default|active_support)`? (And a `SupportedInflectors`?) Or would that be premature?

spec/rubocop/cop/rspec/spec_file_path_format_spec.rb
```diff
@@ -281,4 +281,272 @@ class Foo
...
+    around do |example|
+      described_class.reset_activesupport_cache!
+      example.run
+      described_class.reset_activesupport_cache!
+    end
+
+    before do
+      # rubocop:disable RSpec/VerifiedDoubleReference
+      inflector = instance_double('ActiveSupport::Inflector')
+      # rubocop:enable RSpec/VerifiedDoubleReference
```
bquorning: `RSpec/VerifiedDoubleReference` correctly complains that there’s no point in using `instance_double` with a String argument, because RSpec cannot verify which methods are available unless the class is loaded. And it won’t be 😅 Let’s just use a `double` instead.

```suggestion
      inflector = double('ActiveSupport::Inflector') # rubocop:disable RSpec/VerifiedDoubles
```

(Also, had the class been loaded, I think `class_double` would have been a better choice.)

spec/rubocop/cop/rspec/spec_file_path_format_spec.rb
```diff
@@ -281,4 +281,272 @@ class Foo
...
+  end
+
+  shared_context 'with ActiveSupport available' do
+    before do
+      allow(described_class::ActiveSupportInflector).to receive(:available?)
+        .and_return(true)
+    end
+  end
+
+  shared_context 'when using ActiveSupport integration' do
```
bquorning: It looks like this context is shared between (almost) all the specs you add. Should we try and add e.g. a `describe ActiveSupport integration` block around it all? I know that that would violate `RSpec/NestedGroups`, but let’s just disable that for now.

lib/rubocop/cop/rspec/spec_file_path_format.rb
```diff
@@ -57,8 +63,67 @@ def on_top_level_example_group(node)
           end
         end
 
+        # For testing and debugging
+        def self.reset_activesupport_cache!
```
bquorning: I am not crazy about this method. It feels like maybe we are memoizing on the wrong level somehow.

lib/rubocop/cop/rspec/spec_file_path_format.rb
```diff
@@ -57,8 +63,67 @@ def on_top_level_example_group(node)
...
+
         private
 
+        # Inflector module that uses ActiveSupport for advanced inflection rules
+        module ActiveSupportInflector
+          def self.call(string)
+            ActiveSupport::Inflector.underscore(string)
+          end
+
+          def self.available?(cop_config)
```
bquorning: I am not sure if this method should memoize without considering `cop_config`.

config/default.yml
corsonknowles: I had tried to explain this choice in a PR comment:
> I made the path configurable, but I didn't think an array of possible integrations in config was warranted at this point.
> In the event that someone does come along and adds an adapter to a new Inflector library, they could add a new top level config key and use the same InflectorPath key.
> I guess I just think it's unlikely we'll reach more than 3 custom inflectors at which point we could always switch to making users pick a style instead of having a boolean option.

Happy to change it if you want! 

But to reframe the above, I was on the "premature" side as I was implementing this.  I thought it would be more user-friendly to use a boolean than have folks look up two "magic strings"  to use this. They need both the key and the value if we abstract it, whereas they just need the key if it's a boolean.

Just let me know if you want it though.

spec/rubocop/cop/rspec/spec_file_path_format_spec.rb
corsonknowles: I think I had a double and it got autocorrected for another violation, at which point I disabled it. Good call, I will clean this up.

spec/rubocop/cop/rspec/spec_file_path_format_spec.rb
corsonknowles: Yes, that's what I started with when refactoring to use config. I still like it a lot more, I will turn it back into a NestedGroup!

lib/rubocop/cop/rspec/spec_file_path_format.rb
corsonknowles: I think I need the config, but I will see what I can do

corsonknowles: I mean even if I separate the availability from the config check, I still need the config to look up the `inflector_path` and only in the case where it is `available?` or something like it.

So config is somewhat coupled to availability

lib/rubocop/cop/rspec/spec_file_path_format.rb
corsonknowles: I kind of thought we should hoist this (the logic is in the module still), so we can use it more transparently. 

But open to alternatives...

spec/rubocop/cop/rspec/spec_file_path_format_spec.rb
corsonknowles: Even with 3 lines for the enable/disable and an explanatory comment, this still saves us 9 lines!

spec/rubocop/cop/rspec/spec_file_path_format_spec.rb
corsonknowles: Okay, fixed to violate `VerifiedDoubles` instead, and added back the explanatory comment inline in the specs

spec/rubocop/cop/rspec/spec_file_path_format_spec.rb
corsonknowles: I was able to remove all shared contexts

lib/rubocop/cop/rspec/spec_file_path_format.rb
corsonknowles: Best I can think of would be something like this, but that seems worse to me:

```ruby
        def active_support_available?
          return false unless cop_config.fetch(
            'UseActiveSupportInflections',
            false
          )

          ActiveSupportInflector.available?(cop_config)
        end

        def inflector
          @inflector ||= if active_support_available?
                           ActiveSupportInflector
                         else
                           DefaultInflector
                         end
        end
```        

Let me know whart you're aiming for here

bquorning: Thinking this though once more, I happened to ask myself this question: Do we expect people to give different values for `InflectorPath` in different configuration files? Would that even work? I think `ActiveSupport::Inflector` sets its inflections globally.

So, should `InflectorPath` really be a global configuration that cannot be overwritten?

corsonknowles: I would guess the thinking here is that someone using ActiveSupport but not Rails might have a customized directory structure. And monkey patchers could still use this. 

`ActiveSupport::Inflector` does seem pretty locked down against changes to the base file, and I imagine the same reasoning applies to not moving the config file around, so Rails itself should be pretty stable around which path it scaffolds the file to. 

Maybe something else with nested Rails apps or nested Rails Engines?

I do think this is mildly superfluous, but it was one of the comment requests on the original PR, and now that we have it, I'm hesitant to rip it out again since in theory we're making life easier for +1% of users.

IMO, the heavyweight thing is we have to require 2 files and fail gracefully. Path configurability is minor compared that that.

corsonknowles: If you think its worth the sacrifice, we could take the path out of config and then make it a constant on the cop class. 

I'm imagining somewhere someone will eventually have to type `ruby -e` and patch our cop as they run it from the command line to  get it working for their use case, but perhaps the tradeoff is worth it?

corsonknowles:
@bquorning I'm not really sure where to take this. Should I remove the file path config? I added it because it was asked for in the previous PR. 
Also, should we leverage this to make it more of a seamless default? 

https://github.com/search?q=repo%3Arubocop%2Frubocop%20ActiveSupportExtensionsEnabled&type=code

bquorning:
Regarding `ActiveSupportExtensionsEnabled`, I don’t think I have used that configuration option before. [It looks like](https://github.com/rubocop/rubocop/blob/7dc917524f115771db4e11852d3c16af6c409ccf/config/default.yml#L171-L172) it is intended to tell RuboCop to allow method like `many? {}` and `present?` on a few cops. I don’t think it is intended for things like our use case.

bquorning:
@corsonknowles If you want, I can take a stab at implementing those changes I suggested above?

lib/rubocop/cop/rspec/spec_file_path_format.rb
```diff
@@ -59,6 +65,53 @@ def on_top_level_example_group(node)
...
+
+            inflector_path = config.fetch('InflectorPath')
+
+            unless File.exist?(inflector_path)
+              raise "The configured `InflectorPath` #{inflector_path} does " \
+                    'not exist.'
+            end
+
+            require 'active_support/inflector'
+            require inflector_path
```
bquorning, line 87: I commented on Corson’s code, and now I can add the same comment on my own code 😅 It is unlikely, but not impossible, that different folders in a project use different inflector paths. We cannot blindly use the same `ActiveSupportInflector` instance for all, we need to use different ones per `cop_config` – and probably memoize the result somewhere.

bquorning: Hmm, the file in `inflector_path` is supposed to call `ActiveSupport::Inflector.inflections` with a block, which will configure global state

https://api.rubyonrails.org/classes/ActiveSupport/Inflector.html#method-i-inflections says

> Yields a singleton instance of `Inflector::Inflections` so you can specify additional inflector rules.

So I am not sure how we would handle multiple inflector paths 🤷🏼 Ideas are welcome.

corsonknowles: > So I am not sure how we would handle multiple inflector paths 🤷🏼 Ideas are welcome.

If it helps, I think the ActiveSupport inflector is in one of the most stable possible states software can be in --- it's not deprecated and doesn't look like it will ever be deprecated, and it's essentially Open-Closed -- Closed for Modification (to prevent breaking any legacy apps or forcing upgrades) and Open for Extension. 

With that context, I wonder if we have an example of a real world code base that needs to support multiple inflector paths? Because all I've ever seen is that the authoritative path is used, or possibly additional files are used byt they are included into it. 

I have not seen separate nesting across an application --- which I guess could happen in the case of engines being included in the same repo that don't all funnel their inflection into the base app (which I believe would be a choice folks would not necessarily need to stick to). Is that what we have in mind?

I guess what I'm suggesting rather than a technical solution is that we ship the 99.99% use case, comment and document it as needed, and let someone reach out if they want to support even further customization. 

Now, that goes out the window I guess if we have a real world use case in hand to work against right now, but it would help to know the details of that example to solve for it.

bquorning:
@pirj Would you have time to do the final review of this PR?

lib/rubocop/cop/rspec/spec_file_path_format.rb
```diff
@@ -59,6 +65,53 @@ def on_top_level_example_group(node)
 
         private
 
+        # Inflector module that uses ActiveSupport for advanced inflection rules
+        module ActiveSupportInflector
+          def self.call(string)
+            ActiveSupport::Inflector.underscore(string)
+          end
+
+          def self.prepare_availability(config)
```
pirj, line 74: Cop_config may differ across directories.
Does it make sense to cache the preparations?

lib/rubocop/cop/rspec/spec_file_path_format.rb
```diff
@@ -59,6 +65,53 @@ def on_top_level_example_group(node)
...
+              .gsub(/([^A-Z])([A-Z]+)/, '\1_\2')
+              .gsub(/([A-Z])([A-Z][^A-Z\d]+)/, '\1_\2')
+              .downcase
+          end
+        end
+
+        def inflector
+          case cop_config.fetch('EnforcedInflector')
+          when 'active_support'
+            ActiveSupportInflector.prepare_availability(cop_config)
```
pirj, line 104: In addition to the excessive caching across directories, there can be another problem - multiple threads. RuboCop can run in threads? Or are those processes?

In any case, can we init the inflector once per inspection?

Or this would be too expensive if the inflection file is large?

I can think of a monorepo with many engines each with its own set of inflections

lib/rubocop/cop/rspec/spec_file_path_format.rb
bquorning: Corson and I talked about this a couple of times earlier. The main problem is that loaded inflections are stored globally, in `ActiveSupport::Inflector::Inflections.instance`. So it would be very hard to separate the configurations, even if we tried.

I think that for most use cases, people will only have _one_ inflection configuration, so it’s not a problem. But perhaps we should document the issue for those who may run into the issue in the future?

pirj: Totally reasonable. Even though the anxiety of having config race conditions never left me, we haven’t seen any real reports for … five tears?

pirj:
Thank you!

PR #2094, opened 2025-07-23
Offend `allow_any_instance_of` and `expect_any_instance_of` in `RSpec/SubjectStub`

lovro-bikic:
Expands `RSpec/SubjectStub` to offend `allow_any_instance_of` and `expect_any_instance_of` on the subject (useful in case the subject is a class):
```ruby
allow_any_instance_of(subject).to receive(:foo)
expect_any_instance_of(subject).to receive(:bar)
```

- [x] Added tests.
- [x] Updated documentation.

### DIFF
--- a/lib/rubocop/cop/rspec/subject_stub.rb
+++ b/lib/rubocop/cop/rspec/subject_stub.rb
@@ -38,6 +38,17 @@ module RuboCop
       #     end
       #   end
       #
+      #   # bad
+      #   describe Article do
+      #     subject { Article }
+      #
+      #     it 'indicates that the author is unknown' do
+      #       allow_any_instance_of(subject).to receive(:author).and_return(nil)
+      #
+      #       expect(subject.new.description).to include('by an unknown author')
+      #     end
+      #   end
+      #
       #   # good
       #   describe Article do
       #     subject(:article) { Article.new(author: nil) }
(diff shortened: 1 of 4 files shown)


lib/rubocop/cop/rspec/subject_stub.rb
```diff
@@ -38,6 +38,17 @@ module RSpec
       #     end
       #   end
       #
+      #   # bad
+      #   describe Article do
+      #     subject { Article }
```
pirj, line 43: I can’t recall if I’ve seen such usage ever before. Is this common?

lib/rubocop/cop/rspec/subject_stub.rb
```diff
@@ -38,6 +38,17 @@ module RSpec
...
+      #
+      #     it 'indicates that the author is unknown' do
+      #       allow_any_instance_of(subject).to receive(:author).and_return(nil)
```
pirj, line 46: This is a consequence of usage of the above subject definition.

Do you feel that `allow_any_instance_of(Article).to receive(:author).and_return(nil)` is wrong, and why?

lib/rubocop/cop/rspec/subject_stub.rb
lovro-bikic: I've seen it a couple times, but I think this change makes even more sense in conjunction with this one https://github.com/rubocop/rubocop-rspec/pull/2088, which is part of the reason why I opened this PR.

`allow_any_instance_of(described_class).to receive(:author)` is common in my experience.

pirj:
Closing as stale.
The cop's detection can be worked around with using `described_class` or just the class itself.

Issue #795, opened 2019-08-01
RSpec/DescribedClass mis-identifies inside of rspec-rails anonymous controllers


benhutton:
Versions:
 - `rubocop` — 0.74.0
 - `rubocop-rspec` — 1.34.1

RSpec/DescribedClass will correct code like this:

```
RSpec.describe SomeConcern, type: :controller do
  controller(ApplicationController) do
    include SomeConcern
  end
  ...
end
```

to this:

```
RSpec.describe SomeConcern, type: :controller do
  controller(ApplicationController) do
    include described_class
  end
  ...
end
```

Resulting in this error:

```
An error occurred while loading ./spec/controllers/concerns/some_concern.rb.
Failure/Error: include described_class

NameError:
  undefined local variable or method `described_class' for #<Class:0x000056272150a908>
  Did you mean?  descendants
```

For reference, please see https://relishapp.com/rspec/rspec-rails/v/3-8/docs/controller-specs/anonymous-controller

pirj:
Thanks for reporting, I'll take a look.

pirj:
It works fine for some blocks, so I'd rather not make any code changes:
```ruby
class A
  def self.described_class
    'A'
  end

  def self.y(&block)
    block.call
  end

  y do
    puts self # => A
    puts described_class # => A
  end

  def self.x(&block)
    Class.new(&block)
  end

  x do
    puts self # => #<Class:0x00...
    puts described_class # fails
  end
end
```

There's no way to statically determine what kind of block it is.

It's [a known caveat](https://github.com/rubocop-hq/rubocop-rspec/blob/master/spec/rubocop/cop/rspec/described_class_spec.rb#L14), and there's a configuration option to work this around, `SkipBlocks`:

```yaml
RSpec/DescribedClass:
  SkipBlocks: true
```

I suggest you add a `.rubocop.yml` file with this snippet above only, and put it to a directory where the specs for controller concerns reside. That should solve the issue you're experiencing and will keep the detection for rest of the code base intact.

Please let me know if it works for you, @benhutton, @anthony-robin.

anthony-robin:
Thanks @pirj ! The `SkipBlocks: true` option solved offenses I had 👍

pirj:
@benhutton Did adding a directory-local RuboCop configuration file solve the issue you're experiencing?

benhutton:
[...]

PR #2096, opened 2025-07-25
Add explanation why RSpec/DescribedClass autocorrection is unsafe

lovro-bikic:
Adds missing safety info on `RSpec/DescribedClass`. This cop has [unsafe autocorrection](https://github.com/rubocop/rubocop-rspec/blob/47f03308fd1402d22a7e9b906c49852c062c1301/config/default.yml#L289).

The cop was made unsafe in https://github.com/rubocop/rubocop-rspec/commit/357488829d3e903a4f729b07b9215a2a5d633916. The reason for marking it as unsafe is documented in https://github.com/rubocop/rubocop-rspec/issues/795.

- [ ] Added tests.
- [ ] Updated documentation.

### DIFF
--- a/lib/rubocop/cop/rspec/described_class.rb
+++ b/lib/rubocop/cop/rspec/described_class.rb
@@ -13,6 +13,20 @@ module RuboCop
       # `OnlyStaticConstants` is only relevant when `EnforcedStyle` is
       # `described_class`.
       #
+      # There's a known caveat with rspec-rails's `controller` helper that
+      # runs its block in a different context, and `described_class` is not
+      # available to it. `SkipBlocks` option excludes detection in all
+      # non-RSpec related blocks.
+      #
+      # To narrow down this setting to only a specific directory, it is
+      # possible to use an overriding configuration file local to that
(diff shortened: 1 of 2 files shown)


lib/rubocop/cop/rspec/described_class.rb
```diff
@@ -47,15 +61,6 @@ module RSpec
...
       #   end
       #
-      # There's a known caveat with rspec-rails's `controller` helper that
-      # runs its block in a different context, and `described_class` is not
-      # available to it. `SkipBlocks` option excludes detection in all
-      # non-RSpec related blocks.
-      #
-      # To narrow down this setting to only a specific directory, it is
-      # possible to use an overriding configuration file local to that
-      # directory.
```
lovro-bikic, line 57: I moved this up because in the [generated documentation](https://docs.rubocop.org/rubocop-rspec/cops_rspec.html#rspecdescribedclass) it's also above examples:
Image

bquorning (approved):
Thank you 👍🏼

PR #2098, opened 2025-07-29
Don't register `RSpec/DescribedClass` offenses within `Data.define` blocks

### LINKED ISSUE #2086: Fix RSpec/DescribedClass false positive for Data objects
## Expected behavior

Hello. :wave: I'm running into a situation where `Struct` and `Data` behavior are not consistent when including a module. For a `Struct`, you don't get a `RSpec/DescribedClass` error but with a `Data` object, you do.

The reason is that you can't use `include described_class` as RuboCop tries to tell you because you'll get the following error:

``` 
NameError: undefined local variable or method 'described_class' for class #<Class:0x00000001048b2630>
```

## Actual behavior

The best way to see this is when comparing/contrasting behavior between a `Struct` and `Data` object. Example:

``` ruby
RSpec.describe Demo do
  let :struct do
    Struct.new(:name, :label) { include Demo }
  end

  let :data do
    Data.define(:name, :label) { include Demo }
  end
end
```

With a `Struct`, you won't get a `RSpec/DescribedClass` error but with `Data`, you will.

Here's a quick example (albeit contrived) to reproduce the error:

``` ruby
Demo = Module.new

RSpec.describe Demo do
  describe "#initialize" do
      let :implementation do
        Data.define(:name, :label) { include Demo }
      end

      it "answers Data type" do
        expect(implementation[name: "test", label: "Test"]).to be_a(Data)
      end
  end
end
```

## RuboCop RSpec version

```shell
1.75.5 (using Parser 3.3.8.0, Prism 1.4.0, rubocop-ast 1.44.1, analyzing as Ruby 3.4, running on ruby 3.4.3) [arm64-darwin24.4.0]
  - rubocop-capybara 2.22.1
  - rubocop-disable_syntax 0.2.0
  - rubocop-packaging 0.6.0
  - rubocop-performance 1.25.0
  - rubocop-thread_safety 0.7.2
  - rubocop-rake 0.7.1
  - rubocop-rspec 3.6.0
```

lovro-bikic:
Fixes #2086.

Changes `RSpec/DescribedClass` so it doesn't register offenses within a `Data.define` block. Previously, this:
```ruby
RSpec.describe Demo do
  let(:data) { Data.define(:foo) { include Demo } }
end
```
would have been an offense (with autocorrection to `include described_class`, which would raise an error because `described_class` isn't available within that block). With the patch, this code no longer produces an offense.

- [x] Added tests.
- [ ] Updated documentation.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 - Mark `RSpec/IncludeExamples` as `SafeAutoCorrect: false`. ([@yujideveloper])
 - Fix a false positive for `RSpec/LeakyConstantDeclaration` when defining constants in explicit namespaces. ([@naveg])
 - Add support for error matchers (`raise_exception` and `raise_error`) to `RSpec/Dialect`. ([@lovro-bikic])
+- Don't register offenses for `RSpec/DescribedClass` within `Data.define` blocks. ([@lovro-bikic])
 
 ## 3.6.0 (2025-04-18)
 
--- a/lib/rubocop/cop/rspec/described_class.rb
+++ b/lib/rubocop/cop/rspec/described_class.rb
@@ -83,7 +83,13 @@ module RuboCop
 
         # @!method common_instance_exec_closure?(node)
         def_node_matcher :common_instance_exec_closure?, <<~PATTERN
-          (block (send (const nil? {:Class :Module :Struct}) :new ...) ...)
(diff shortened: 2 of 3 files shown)


ydah (approved):
Thank you!

pirj:
Thank you!

Issue #2093, opened 2025-07-16
[FactoryBot] Wrong interpretation of attribute as example declaration


Mth0158:
## Expected behavior

When declaring my factories using FactoryBot gem, I want to be able to use an attribute called `example` without Rubocop complaining about expectations.

## Actual behavior

```ruby
FactoryBot.define do
  factory :word_function_meaning, class: "Word::Function::Meaning" do
    ...
    trait :with_admin_update_attributes do
      definition { Faker::Lorem.sentence(word_count: rand(4..8)) }

      example { definition }
    end
    ...
  end
end
```

```bash
Offenses:

spec/factories/word/function/meanings.rb:15:7: C: RSpec/NoExpectationExample: No expectation found in this example.
      example { definition }
      ^^^^^^^^^^^^^^^^^^^^^^
```

Here, I need to add `# rubocop:disable RSpec/NoExpectationExample` around the `example { definition }` line because I am getting a `RSpec/NoExpectationExample: No expectation found in this example.`.

## Steps to reproduce the problem

See above.

## RuboCop RSpec version

Include the output of `rubocop -V` or `bundle exec rubocop -V` if using Bundler.
If you see extension cop versions (e.g. `rubocop-performance`, `rubocop-rake`, and others)
output by `rubocop -V`, include them as well. Here's an example:

```shell
$ [bundle exec] rubocop -V
1.78.0 (using Parser 3.3.8.0, Prism 1.4.0, rubocop-ast 1.45.1, analyzing as Ruby 3.4, running on ruby 3.4.3) [arm64-darwin24]
  - rubocop-factory_bot 2.27.1
  - rubocop-performance 1.25.0
  - rubocop-rails 2.32.0
  - rubocop-rspec 3.6.0
  - rubocop-rspec_rails 2.31.0
  - rubocop-slim 0.3.0
```

pirj:
Thanks for reporting. This is an interesting case on the edge of usage of two tools.
I clearly understand the issue here.
It would make sense to skip several our cops by adding the Exclude directive with FactoryBot-specific paths. This would also benefit scanning speed.
But I’m more inclined to go through the list of our cops and make this change to all the cops that apply. I can’t tell off the top of my head, but I suspect it’s not all cops, and some may still apply as core RSpec c constructs may still be used inside factories? Or is that nonsense, and we should completely ignore?

pirj:
Hi @Mth0158 
[...]

PR #2099, opened 2025-07-31
Exclude all cops from inspecting factorybot files, except if explicit…

Mth0158:
Fixes https://github.com/rubocop/rubocop-rspec/issues/2093

@pirj as discussed

Reading the code, thanks to your help, I noticed that it might be preferable to exclude `factorybot` files from inspection. From my understanding, all the cops, except 2 (`Metrics/BlockLength` and `RSpec/RemoveConst`), are relevant for inspecting these files.

I have included the 2 that are relevant apply to `factorybot` with the include `inherit_method`.

Unsure about how to add tests for these? I read the test and decided it was not necessary to add some but let me know what you expect if so?

Let me know what you think.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,6 @@
 - Add support for error matchers (`raise_exception` and `raise_error`) to `RSpec/Dialect`. ([@lovro-bikic])
 - Don't register offenses for `RSpec/DescribedClass` within `Data.define` blocks. ([@lovro-bikic])
 - Add autocorrection support for `RSpec/IteratedExpectation` for single expectations. ([@lovro-bikic])
-- Exclude all cops from inspecting factorybot files, except if explicitly included ([@Mth0158])
 - Exclude all cops from inspecting factorybot files, except if explicitly included. ([@Mth0158])
 
 ## 3.6.0 (2025-04-18)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 - Add support for error matchers (`raise_exception` and `raise_error`) to `RSpec/Dialect`. ([@lovro-bikic])
 - Don't register offenses for `RSpec/DescribedClass` within `Data.define` blocks. ([@lovro-bikic])
 - Add autocorrection support for `RSpec/IteratedExpectation` for single expectations. ([@lovro-bikic])
+- Exclude all cops from inspecting factorybot files, except if explicitly included ([@Mth0158])
 
 ## 3.6.0 (2025-04-18)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
 - Don't register offenses for `RSpec/DescribedClass` within `Data.define` blocks. ([@lovro-bikic])
 - Add autocorrection support for `RSpec/IteratedExpectation` for single expectations. ([@lovro-bikic])
 - Exclude all cops from inspecting factorybot files, except if explicitly included ([@Mth0158])
+- Exclude all cops from inspecting factorybot files, except if explicitly included. ([@Mth0158])
 
 ## 3.6.0 (2025-04-18)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1028,6 +1028,7 @@ Compatibility release so users can upgrade RuboCop to 0.51.0. No new features.
 [@mlarraz]: https://github.com/mlarraz
 [@mockdeep]: https://github.com/mockdeep
 [@mothonmars]: https://github.com/MothOnMars
+[@Mth0158]: https://github.com/Mth0158
 [@mvz]: https://github.com/mvz
 [@naveg]: https://github.com/naveg
 [@nc-holodakg]: https://github.com/nc-holodakg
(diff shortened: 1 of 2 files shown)


config/default.yml
```diff
@@ -816,6 +829,15 @@ RSpec/RemoveConst:
   Enabled: true
   VersionAdded: '2.26'
   Reference: https://www.rubydoc.info/gems/rubocop-rspec/RuboCop/Cop/RSpec/RemoveConst
+  inherit_mode:
+    merge:
+      - Include
```
pirj: It seems unlikely to have remove_const in factories. Is this needed?

CHANGELOG.md
```diff
@@ -6,6 +6,7 @@
 - Fix a false positive for `RSpec/LeakyConstantDeclaration` when defining constants in explicit namespaces. ([@naveg])
 - Add support for error matchers (`raise_exception` and `raise_error`) to `RSpec/Dialect`. ([@lovro-bikic])
 - Don't register offenses for `RSpec/DescribedClass` within `Data.define` blocks. ([@lovro-bikic])
+- Exclude all cops from inspecting factorybot files, except if explicitly included ([@Mth0158])
 
 ## 3.6.0 (2025-04-18)
```
pirj, line 12: Please add 
```
       [@mth0158]: https://github.com/Mth0158
 ```
To the list in the bottom, alphabetically?

File: config/default.yml (hunk 1 above)
Comment (Mth0158) [reply config/default.yml]: I wasn't 100% sure of the use for this cop, let's remove it. I have never done a remove_const in Rspec

File: CHANGELOG.md [hunk 3]
```diff
@@ -1026,6 +1027,7 @@ Compatibility release so users can upgrade RuboCop to 0.51.0. No new features.
 [@mlarraz]: https://github.com/mlarraz
 [@mockdeep]: https://github.com/mockdeep
 [@mothonmars]: https://github.com/MothOnMars
+[@Mth0158]: https://github.com/Mth0158
```
pirj:
```suggestion
[@mth0158]: https://github.com/Mth0158
```
Appease markdown linter insisting on downcase link anchors

CHANGELOG.md
```diff
@@ -6,6 +6,7 @@
 - Fix a false positive for `RSpec/LeakyConstantDeclaration` when defining constants in explicit namespaces. ([@naveg])
 - Add support for error matchers (`raise_exception` and `raise_error`) to `RSpec/Dialect`. ([@lovro-bikic])
 - Don't register offenses for `RSpec/DescribedClass` within `Data.define` blocks. ([@lovro-bikic])
+- Exclude all cops from inspecting factorybot files, except if explicitly included ([@Mth0158])
```
pirj:
```suggestion
- Exclude all cops from inspecting factorybot files, except if explicitly included. ([@Mth0158])
```

config/default.yml
```diff
@@ -6,6 +6,12 @@ RSpec:
   Include:
     - "**/*_spec.rb"
     - "**/spec/**/*"
+  Exclude:
+    - "**/spec/factories.rb"
+    - "**/spec/factories/**/*.rb"
+    - "**/test/factories.rb"
+    - "**/test/factories/**/*.rb"
+    - "**/features/support/factories/**/*.rb"
```
bquorning: I am not sure that we need to exclude files that we haven’t asked RuboCop to include.

config/default.yml
```diff
@@ -110,7 +116,14 @@ RSpec:
...
   inherit_mode:
     merge:
+      - Include
       - Exclude
+  Include:
+    - "**/spec/factories.rb"
+    - "**/spec/factories/**/*.rb"
+    - "**/test/factories.rb"
+    - "**/test/factories/**/*.rb"
+    - "**/features/support/factories/**/*.rb"
```
bquorning: And here, I am not sure that we need to include files that we haven’t asked RuboCop to exclude. Maybe I understand it wrong?

config/default.yml
pirj: This roughly means “include spec/**/*, but not spec/factories”.

config/default.yml
pirj: You’re right. The logic was to “not affect RuboCop default settings of BlockLength on factories”, but this would mean that it will start triggering on long factories.

Our previous setting was to tell it to ignore factories, and we should keep it this way.

Please revert this part.

Mth0158: Done ;)
In the end, you were 100% right @pirj, removing the factorybot file inspection was the way to go :)

config/default.yml
bquorning: I understand why lines 10 and 11 are necessary, but not lines 12–14. We don’t configure RuboCop to include those files, right?

pirj: Fair enough regarding 12-13

But we do include spec/features eg  https://github.com/mastodon/mastodon/blob/dc7e2985e60c90e8efe82e56edb4fdd473bd1651/spec/features/profile_spec.rb

They may have spec/features/support/factories which we don’t want to expect.

bquorning: Good point.

config/default.yml
```diff
@@ -6,6 +6,12 @@ RSpec:
   Include:
     - "**/*_spec.rb"
     - "**/spec/**/*"
+  Exclude:
+    - "**/spec/factories.rb"
+    - "**/spec/factories/**/*.rb"
+    - "**/test/factories.rb"
+    - "**/test/factories/**/*.rb"
```
bquorning: Based on the comment below ↓
```suggestion
```

Mth0158: Done!

PR #2110, opened 2025-09-16
Fix CI on ruby 2.7

Earlopain:
Prism 1.5.0 contains an incorrect polyfill for `warn`, which causes it to always output with location information, even if no uplevel is provided.

I fixed this in https://github.com/ruby/prism/pull/3647 but it is not released yet.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -128,11 +128,6 @@ jobs:
     name: Prism
     steps:
       - uses: actions/checkout@v5
-      - name: Use prism parser
-        run: |
-          cat << EOF > Gemfile.local
-            gem 'prism'
-          EOF
       - name: set up Ruby
         uses: ruby/setup-ruby@v1
         with:
--- a/Gemfile
+++ b/Gemfile
@@ -13,5 +13,10 @@ gem 'rubocop-rake', '~> 0.7'
 gem 'simplecov', '>= 0.19'
 gem 'yard'
 
+# FIXME: Remove when the next prism version is released.
+if RUBY_VERSION < '3.0' || RUBY_ENGINE == 'jruby'
+  gem 'prism', '!= 1.5.0', '!= 1.5.1'
+end
+
 local_gemfile = 'Gemfile.local'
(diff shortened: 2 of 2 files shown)


.github/workflows/main.yml
```diff
@@ -128,11 +128,6 @@ jobs:
     name: Prism
     steps:
       - uses: actions/checkout@v5
-      - name: Use prism parser
```
Earlopain, line 131: This is no longer needed, it gets pulled in by `rubocop-ast` and also prevents me from specifying it in the gemfile since it then would appear twice.

Gemfile
```diff
@@ -13,5 +13,8 @@ gem 'rubocop-rake', '~> 0.7'
 gem 'simplecov', '>= 0.19'
 gem 'yard'
 
+# FIXME: Remove when the next prism version is released.
+gem 'prism', '< 1.5.0' if RUBY_VERSION < '3.0' || RUBY_ENGINE == 'jruby'
```
bquorning: Since your PR is merged, this could probably be `'!= 1.5.0', '!= 1.5.1'` instead?

bquorning (approved):

Gemfile
Earlopain: Sure, it can be. BTW, jruby is included in this list because it has an incompatible `UnboundMethod#parameters` implementation, which causes the polyfill to be also be applied even though it really shouldn't be

pirj:
Just thinking out loud.
Prism is opt-in with `PARSER_ENGINE=parser_prism` in `rubocop`.
Those who opt-in, most certainly are early adopters, and unlikely to run Ruby 2.7
We're kind of interested in how well Prism is with the oldest Rubies, but we're also interested how Prism would work with modern code, e.g. support the itblock on Ruby 3.4.

With this in mind, does it make sense to keep the Prism build on 2.7? I can't recall why we've [introduced it with 2.7](https://github.com/rubocop/rubocop-capybara/pull/112) in the first place.
Should we switch it to 3.4?

Please accept my apologies if I'm missing something obvious.

PR #2111, opened 2025-09-17
CI: Reduce duplication by having extra gemfiles

bquorning:


pirj:
@bquorning Do you want to continue on that?
I recall https://github.com/thoughtbot/appraisal was useful for having multiple Gemfiles to run locally and on CI.

bquorning:
Appraisal is useful for generating/maintaining multiple gemfiles, but as I recall it, doesn’t play very well with GitHub Actions and bundler caching etc.

And yes, Appraisal may be better than using the `eval_gemfile 'common.rb'` pattern. I’ll close this PR for now.

PR #2112, opened 2025-10-02
RSpec/VerifiedDoubles: fix detection of nameless doubles

ushi-as:
The detection of nameless doubles is broken, when methods are specified. This PR fixes it.

- [x] Added tests.
- [x] Updated documentation.

### DIFF
--- a/lib/rubocop/cop/rspec/verified_doubles.rb
+++ b/lib/rubocop/cop/rspec/verified_doubles.rb
@@ -78,7 +78,7 @@ module RuboCop
 
         def on_send(node)
           unverified_double(node) do |name, *_args|
-            return if name.nil? && cop_config['IgnoreNameless']
+            return if (name.nil? || hash?(name)) && cop_config['IgnoreNameless']
             return if symbol?(name) && cop_config['IgnoreSymbolicNames']
 
             add_offense(node)
@@ -89,6 +89,10 @@ module RuboCop
 
         def symbol?(name)
           name&.sym_type?
+        end
+
+        def hash?(arg)
+          arg.hash_type?
         end
       end
     end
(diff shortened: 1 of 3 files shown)


spec/rubocop/cop/rspec/verified_doubles_spec.rb
```diff
@@ -74,6 +74,14 @@
     RUBY
   end
 
+  it 'ignores doubles that have no name but methods specified' do
```
bquorning, line 86: Could you add a similar test in the block above (line 56–67) that we detect an offense on the same code when `IgnoreNameless` is false?

bquorning (commented):
Nice find

spec/rubocop/cop/rspec/verified_doubles_spec.rb
ushi-as: @bquorning done

bquorning:
Let me have a look at the failing CI. It seems unrelated to your changes.

lib/rubocop/cop/rspec/verified_doubles.rb
```diff
@@ -90,6 +90,10 @@ def on_send(node)
         def symbol?(name)
           name&.sym_type?
         end
+
+        def hash?(arg)
+          arg&.hash_type?
```
pirj: We’re checking for `nil?`, and safe navigation here is redundant.

pirj:
```suggestion
          arg.hash_type?
```

bquorning: > Could you please review the test coverage? Add tests as needed.

The thing missing test coverage is actually the `else` branch of this safe navigation operator.

ydah: 👍️

ushi-as: @pirj Thx for your review. I applied your suggestion.

ydah (approved):

Issue #2075, opened 2025-04-08
RSpec/LetSetup false positive with nested groups


dgmora:
I realize this is similar to https://github.com/rubocop/rubocop-rspec/issues/830, but this one here is happening without shared examples:

```ruby
# frozen_string_literal: true

describe User, type: :model do
  let!(:user) { create(:user) }

  it 'is valid' do
    expect(user).to be_valid
  end

  context 'with no direct usage' do
    let!(:user) { create(:user, :special) } # <-- offense, but shouldn't!

    it 'creates something' do
      expect(SomeModel.count).to eq(2)
    end
  end
end
```

The second let declaration is constrained by a higher level `let!`. The `let!` is part of the setup, so changing the second let to `before { ... }` may not work, since the `user` will already be created and you are now creating a second user. You can't also change the first one, because it is used in another test. 

I see that this could be refactored to `let(:user)` plus `before`. But it still looks like a false positive to me because is not actually possible to change that second `let!` to a before block. The first usage of `let!` is legit. I don't think you should refactor the whole test just to comply with `RSpec/LetSetup`

______________________________________________________________________

## Expected behavior

If the same `let!` is declared in a higher level, those on the lower level shouldn't be marked as an offense because it's not possible to determine if you can change it to a before or not.

## Actual behavior

The second let reports `RSpec/LetSetup`

## Steps to reproduce the problem

See above

## RuboCop RSpec version

```shell
$ bundle exec rubocop -V
1.74.0 (using Parser 3.3.7.2, rubocop-ast 1.38.1, analyzing as Ruby 3.3, running on ruby 3.3.6) [arm64-darwin23]
  - rubocop-performance 1.24.0
  - rubocop-factory_bot 2.27.1
  - rubocop-rails 2.30.3
  - rubocop-rspec 3.5.0
  - rubocop-rspec_rails 2.31.0
```

PR #2117, opened 2025-10-02
Fix a false positive for `RSpec/LetSetup` when `let!` used in outer scope

ydah:
Fixes: https://github.com/rubocop/rubocop-rspec/issues/2075

- [x] Added tests.
- [x] Updated documentation.

### DIFF
--- a/lib/rubocop/cop/rspec/let_setup.rb
+++ b/lib/rubocop/cop/rspec/let_setup.rb
@@ -25,6 +25,18 @@ module RuboCop
       #   it 'counts widgets' do
       #     expect(Widget.count).to eq(1)
       #   end
+      #
+      #   # good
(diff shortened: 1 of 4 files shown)


lib/rubocop/cop/rspec/let_setup.rb
```diff
@@ -68,6 +82,22 @@ def child_let_bang(node, &block)
             let_bang(let, &block)
           end
         end
+
+        def overrides_outer_let_bang?(node, method_name)
+          node.each_ancestor(:block).any? do |ancestor|
```
bquorning, line 87: I just thought, does rubocop-ast need an `any_ancestor?` method?

bquorning (approved):

PR #2118, opened 2025-10-02
Remove test for safety configuration of cops documented with `@safety` YARD tag

ydah:
This is because in some cases, even when a safety tag is present, `safe` may be `false` or `set_autocorrect` may be `true`.
- https://github.com/rubocop/rubocop-rspec/blob/e63ee9a79c17be296d41ac36dc006d94d6d12069/lib/rubocop/cop/rspec/example_wording.rb#L13-L25

- [-] Added tests.
- [-] Updated documentation.

### DIFF
--- a/spec/project/default_config_spec.rb
+++ b/spec/project/default_config_spec.rb
@@ -102,18 +102,4 @@ RSpec.describe 'config/default.yml' do
       )
     end
   end
-
-  it 'is expected that all cops documented with `@safety` are `Safe: false` ' \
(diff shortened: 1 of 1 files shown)


pirj:
No objections on merging as is. Though, isn’t this a helpful poka-yoke? Maybe we could exclude some cops by marking them as “@safety tentative” in their docs and filtering those from participating in this spec?

pirj (approved):
Thank you!

PR #2120, opened 2025-10-02
Fix false negative for `RSpec/RepeatedExampleGroupDescription` cop when include skip/pending

### LINKED ISSUE #2021: False negative for `RSpec/RepeatedExampleGroupDescription`
## Describe the solution you'd like

```ruby
RSpec.describe Screenshots::CreateInteractor do
  context 'when the request is valid', use_vcr_cassette: 'screenshots/valid_request' do
    pending 'add something'
  end

  context 'when the request is valid', use_vcr_cassette: 'screenshots/invalid_request' do
    pending 'add something'
  end
end
```

## RuboCop

```console
ydakuka@yauhenid:~/Work/project$ rubocop -V
1.70.0 (using Parser 3.3.6.0, rubocop-ast 1.37.0, analyzing as Ruby 2.7, running on ruby 2.7.8) [x86_64-linux]
  - rubocop-capybara 2.21.0
  - rubocop-factory_bot 2.26.1
  - rubocop-performance 1.23.1
  - rubocop-rails 2.28.0
  - rubocop-rake 0.6.0
  - rubocop-rspec 3.3.0
  - rubocop-rspec_rails 2.30.0
  - rubocop-thread_safety 0.6.0
```

ydah:
Fixes: https://github.com/rubocop/rubocop-rspec/issues/2021

- [x] Added tests.
- [-] Updated documentation.

### DIFF
--- a/lib/rubocop/cop/rspec/repeated_example_group_description.rb
+++ b/lib/rubocop/cop/rspec/repeated_example_group_description.rb
@@ -45,7 +45,7 @@ module RuboCop
       class RepeatedExampleGroupDescription < Base
         include SkipOrPending
 
-        MSG = 'Repeated %<group>s block description on line(s) %<loc>s'
+        MSG = 'Repeated %<group>s block description on line(s) %<loc>s.'
 
         # @!method several_example_groups?(node)
         def_node_matcher :several_example_groups?, <<~PATTERN
@@ -64,7 +64,9 @@ module RuboCop
           return unless several_example_groups?(node)
 
           repeated_group_descriptions(node).each do |group, repeats|
-            add_offense(group, message: message(group, repeats))
+            message = format(MSG, group: group.method_name,
+                                  loc: repeats.join(', '))
+            add_offense(group, message: message)
           end
         end
 
@@ -74,7 +76,6 @@ module RuboCop
           node
             .children
             .select { |child| example_group?(child) }
-            .reject { |child| skip_or_pending_inside_block?(child) }
             .reject { |child| empty_description?(child) }
             .group_by { |group| doc_string_and_metadata(group) }
             .values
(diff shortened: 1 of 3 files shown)


pirj:
I couldn’t find at a glance, why did we implement it in a way that it was skipping pending and skip?

ydah:
@pirj I think they were more guarded than necessary to prevent false positives below. No other discussion of skip was found in the PR when the cop was created.

```ruby
    context 'rejected' do
      skip
    end

    context 'processed' do
      skip
    end
```

refs: https://github.com/rubocop/rubocop-rspec/pull/860#pullrequestreview-346900434

spec/rubocop/cop/rspec/repeated_example_group_description_spec.rb
```diff
@@ -240,4 +240,180 @@
       end
     RUBY
   end
+
+  it 'does not register offense for example group with skip' do
```
pirj: This description is not exactly fair, is it?
We do not register an offence, but because context descriptions are different. Wouldn’t this pass before the fix, too?

spec/rubocop/cop/rspec/repeated_example_group_description_spec.rb
```diff
@@ -240,4 +240,180 @@
...
+        end
+
+        context 'when foo is false' do
+          skip
+        end
+      end
+    RUBY
+  end
+
+  it 'does not register offense for example group with pending' do
```
pirj: This one as well

lib/rubocop/cop/rspec/repeated_example_group_description.rb
```diff
@@ -74,7 +76,6 @@ def repeated_group_descriptions(node)
           node
             .children
             .select { |child| example_group?(child) }
-            .reject { |child| skip_or_pending_inside_block?(child) }
```
pirj, line 77: I think the decision not to flag pending/skip [was a deliberate decision](https://github.com/rubocop/rubocop-rspec/pull/860#pullrequestreview-346900434).

pirj (commented):
I'd rather not be doing this. 
I have no certainty that the issue https://github.com/rubocop/rubocop-rspec/issues/2021 had any focus on skip/pending, even though `pending` was in the description.
How I understand it, it was a difference in metadata that caused the issue.

Issue #2012, opened 2024-12-24
RedundantPredicateMatcher is not a safe autocorrect


luke-hill:
*Be clear, concise and precise in your description of the problem.
Open an issue with a descriptive title and a summary in grammatically correct,
complete sentences.*

*Use the template below when reporting bugs. Please, make sure that
you're running the latest stable RuboCop RSpec and that the problem you're reporting
hasn't been reported (and potentially fixed) already.*

*Before filing the ticket you should replace all text above the horizontal
rule with your own words.*

*In the case of false positive or false negative, please add the
corresponding cop name.*

______________________________________________________________________

## Expected behavior

When using RSpec with autocorrect I assumed it wouldn't break my codebase

## Actual behavior

In `cucumber-ruby-core` it replaced `be_match` with `match`. However these weren't correct, because the codebase had defined `match?` as a method. So the correct behaviour here is to detect that the method `match?` has been defined, and then not suggest an autocorrect or error here.

## Steps to reproduce the problem

https://github.com/cucumber/cucumber-ruby-core/compare/release/v13.0.3...main#diff-146e0d9fad5351a0fc7a77b7bb4e47a19b46a83e67e03c9330cd294077be02abL65)

## RuboCop RSpec version

```shell
RuboCop version incompatibility found, RuboCop server restarting...
RuboCop server starting on 127.0.0.1:36575.
1.69.2 (using Parser 3.3.6.0, rubocop-ast 1.37.0, analyzing as Ruby 3.0, running on ruby 3.1.6) +server [x86_64-linux]
  - rubocop-packaging 0.5.2
  - rubocop-rake 0.6.0
  - rubocop-rspec 3.0.5

```

pirj:
Do you think that using the `match` matcher would work incorrectly with your object that defines that method?
RSpec compares the actual and expected with `==`, then with `===`, and [then](https://github.com/rspec/rspec/blob/fdbedfac7ff4383a4c55818e798802f4e0b2851c/rspec-expectations/lib/rspec/matchers/built_in/match.rb#L55) with `match`. Do you expect any side effects or downsides here?
I recall a peculiar case when eg an object was changing its `hash` on comparison.

pirj:
Moving my comments from the pr here for better visibility. Those are just thoughts and personal opinions.

I agree that the `match` matcher works with objects defining `match?` coincidentally, and such usage is not documented in RSpec.

[...]

PR #2121, opened 2025-10-02
Add SupportedMethods configuration to `RSpec/RedundantPredicateMatcher` cop

ydah:
This PR add SupportedMethods configuration to `RSpec/RedundantPredicateMatcher` cop. 
Fixes: https://github.com/rubocop/rubocop-rspec/issues/2012

This change involves trade-offs. Previously, we could specify replaceable method names in a hard-coded manner, which allowed us to use RESTRICT_ON_SEND. However, now that the decision is based on values set in the config, RESTRICT_ON_SEND can no longer be used. Additionally, since `be_all` requires special internal processing, it's difficult to create this cop generically with Config, leaving special handling for `be_all` inside the implementation.

However, this change enables the `ResultPredictedMatcher` cop to determine which cops trigger violations through Config. For example, when using Cucumber, this allows you to exclude only `match` from violations. As mentioned earlier though, this change involves trade-offs, so if these trade-offs are unacceptable, we need to discard this change. I think it's fine to discard it if necessary.

- [x] Added tests.
- [x] Updated documentation.

### DIFF
--- a/config/default.yml
+++ b/config/default.yml
@@ -818,7 +818,20 @@ RSpec/RedundantAround:
 RSpec/RedundantPredicateMatcher:
   Description: Checks for redundant predicate matcher.
   Enabled: true
+  SupportedMethods:
+    be_all: all
+    be_cover: cover
+    be_end_with: end_with
+    be_eql: eql
+    be_equal: equal
+    be_exist: exist
+    be_exists: exist
+    be_include: include
+    be_match: match
+    be_respond_to: respond_to
+    be_start_with: start_with
   VersionAdded: '2.26'
+  VersionChanged: "<<next>>"
   Reference: https://www.rubydoc.info/gems/rubocop-rspec/RuboCop/Cop/RSpec/RedundantPredicateMatcher
 
 RSpec/RemoveConst:
--- a/lib/rubocop/cop/rspec/redundant_predicate_matcher.rb
+++ b/lib/rubocop/cop/rspec/redundant_predicate_matcher.rb
@@ -5,6 +5,9 @@ module RuboCop
     module RSpec
       # Checks for redundant predicate matcher.
       #
+      # This cop configures the mapping of redundant predicate matchers
+      # and their preferable alternatives via `SupportedMethods` option.
+      #
       # @example
       #   # bad
       #   expect(foo).to be_exist(bar)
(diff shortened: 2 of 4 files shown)


lib/rubocop/cop/rspec/redundant_predicate_matcher.rb
```diff
@@ -16,50 +19,55 @@ module RSpec
...
-        RESTRICT_ON_SEND =
-          %i[be_all be_cover be_end_with be_eql be_equal
-             be_exist be_exists be_include be_match
-             be_respond_to be_start_with].freeze
+        UNSUPPORTED_AUTOCORRECT_METHODS = %w[be_all].freeze
 
         def on_send(node)
-          return if node.parent.block_type? || node.arguments.empty?
-          return unless replaceable_arguments?(node)
+          return unless !node.parent&.block_type? && !node.arguments.empty?
```
bquorning: I find this line much harder to read than the original line. We could even change it to
```rb
return if node.parent.block_type?
return if node.arguments.empty?
```
However you added a `&.` so there’s special handling if `node.parent.nil?`

ydah: I fix it. Thank you!

bquorning (approved):

config/default.yml
```diff
@@ -818,7 +818,20 @@ RSpec/RedundantAround:
...
+  SupportedMethods:
+    be_all: all
+    be_cover: cover
+    be_end_with: end_with
+    be_eql: eql
+    be_equal: equal
+    be_exist: exist
+    be_exists: exist
+    be_include: include
+    be_match: match
```
pirj, line 830: Will our users have to configure the cop by copying over all this except the replacements they want to skip?

pirj (commented):
I’m uncertain about this, especially considering that the behaviour of the suggested `match` matcher is [not necessarily incorrect](https://github.com/rspec/rspec/blob/fdbedfac7ff4383a4c55818e798802f4e0b2851c/rspec-expectations/lib/rspec/matchers.rb#L666) even for objects that are not regexp/string/hash/array.

ydah:
@bquorning @pirj I showed a different approach. This might be better. How is it?
- https://github.com/rubocop/rubocop-rspec/pull/2127

PR #2125, opened 2025-10-05
Add new cop `RSpec/RedundantPending`

ydah:
This cop detects redundant `pending` or `skip` calls inside examples that are already marked as skipped. When an example is skipped using `xit`, `xspecify`, `xexample`, or skip/pending metadata, adding a `pending` or `skip` call inside the example body is redundant and creates unnecessary duplication.

It's common to see code like this:

```ruby
xspecify do
  pending 'Need to upgrade to the latest HTTP gem version'
  expect(pinger.call).to eq(204)
end

xit 'answers success status' do
  pending 'Need to upgrade HTTP gem before this will work again'
  expect(pinger.call).to eq(200)
end
```

In these cases:
- The example is already skipped via `xspecify`/`xit`
- The `pending` call inside is redundant and never executed
- This creates confusion about where the skip/pending is actually applied
- The reason message in `pending` is not displayed (the example is already skipped at method level)

This cop flags such redundant calls and suggests either:
1. Remove the redundant `pending`/`skip` call entirely, OR
2. Use a regular example method (`it`, `specify`) if the pending call is actually needed

The cop detects redundancy in the following cases:

1. Skipped example methods:
   - `xit`, `xspecify`, `xexample`

2. Skip/pending metadata:
   - Symbol metadata: `it 'test', :skip do`
   - Symbol metadata: `it 'test', :pending do`
   - Hash metadata: `it 'test', skip: true do`
   - Hash metadata: `it 'test', skip: 'reason' do`
   - Hash metadata: `it 'test', pending: 'reason' do`

3. Both `pending` and `skip` calls:
   - Detects both `pending 'reason'` and `skip 'reason'`

The cop does NOT flag these legitimate uses:

1. Regular examples with pending/skip: ```ruby it 'does something' do pending 'not yet implemented' expect(something).to be_truthy end ```

2. Conditional skip/pending: ```ruby xit 'does something' do setup_something skip 'setup failed' if setup_failed?  # Not first statement expect(something).to be_truthy end ```

3. Metadata with `false` value: ```ruby it 'does something', skip: false do pending 'conditional pending' expect(something).to be_truthy end ```

- Checks only the first statement in the example body
- Supports both regular blocks (`do...end`) and numblocks
- Handles examples with single or multiple statements
- Handles one-liner syntax: `xit('test') { pending 'reason' }`

- Reduces confusion about where examples are actually skipped
- Eliminates redundant code
- Makes skip/pending reasons visible (use metadata instead of body)
- Improves test suite maintainability

- [x] Added tests.
- [x] Updated documentation.


.rubocop.yml
```diff
@@ -294,3 +294,4 @@ Performance/ZipWithoutBlock: {Enabled: true}
 
 RSpec/IncludeExamples: {Enabled: true}
 RSpec/LeakyLocalVariable: {Enabled: true}
+RSpec/RedundantPending: {Enabled: false}
```
bquorning, line 297:
```suggestion
RSpec/RedundantPending: {Enabled: true}
```

lib/rubocop/cop/rspec/redundant_pending.rb
```diff
@@ -0,0 +1,106 @@
...
+
+        # @!method skipped_example?(node)
+        def_node_matcher :skipped_example?, <<~PATTERN
+          {
+            (any_block (send _ #Examples.skipped ...) ...)
+          }
+        PATTERN
+
+        # @!method skipped_by_metadata?(node)
+        def_node_matcher :skipped_by_metadata?, <<~PATTERN
```
bquorning, line 65: Can we reuse `skipped_in_metadata?` from the `SkipOrPending` mixin?

PR #2127, opened 2025-10-06
Add `AllowedIdentifiers` configuration option to `RSpec/RedundantPredicateMatcher`

ydah:
Fixes: #2012

Another solution of https://github.com/rubocop/rubocop-rspec/pull/2121

- [x] Added tests.
- [x] Updated documentation.


ydah:
If there's another case where this solution might be useful, it would be if there've redefined match themself, right? In that case, this might be usable. What do you think?

ydah:
Oops, what I was talking about was exactly the case in the Issue#2012. This cop offense follows standard usage, and we don't need to proceed with this right now. I'll close this for now.

Issue #2128, opened 2025-10-21
New cop to disallow printing from tests


kevinrobell-st:
## Is your feature request related to a problem? Please describe.
We noticed some print debugging statements in test output. I assumed they were already covered by existing cops but they weren't.

## Describe the solution you'd like
It'd be nice to have a cop similar to the existing `Rails/Output` cop that prevents print debugging in the test directory.

- https://docs.rubocop.org/rubocop-rails/cops_rails.html#railsoutput

## Describe alternatives you've considered
Writing a custom cop.

## Additional context
I could technically just change the `Include:` field on the `Rails/Output` cop to prevent this from happening in tests but the error message would then be unhelpful. I also assume this would be more generally applicable outside of my own project.

pirj:
Can you give an example? Is this debugging output coming from the spec code, or the production code? Did you have in mind to scope this to Rails projects, or do you think this should apply to any Ruby project that uses RSpec?

kevinrobell-st:
It's mostly to make sure that print debugging in tests never passes CI.

```diff
it "returns an empty response" do
  get endpoint, :params => params
-  puts response.parsed_body
  expect(response).to have_http_status(:ok)
end
```

It would only check for print statements in spec code and not in production code. I think it would be useful in general but maybe there's some project that relies on printing from tests that I don't know about.

bquorning:
In one project, I have redirected `$stdout` (and sometimes `$stderr`) to a `StringIO` initialized with an immutable string (`-""`), so all printing will cause an error. So yes, I think this cop will be useful 😅

However most of the cases I catch come from people using a `Logger` instance initialized with `$stdout`/`STDOUT`. And the cop suggested in #2129 will unfortunately _not_ catch such cases.

PR #2129, opened 2025-11-06
Add a new cop `RSpec/Output`

kevinrobell-st:
- Closes https://github.com/rubocop/rubocop-rspec/issues/2128

This is based on the `Rails/Output` cop with three minor changes.

1. Autocorrection is removed as the expectation is that the print statement will be removed by the user.
2. The message is changed.
3. The cop runs only on spec files.

- [x] Added tests.
- [x] Updated documentation.


kevinrobell-st:
`RSpec/Output` was chosen as the cop name to align with `Rails/Output` but potentially it'd be better to have more descriptive name here.

kevinrobell-st:
This is succeeding for me locally with no file changes. I wonder why.
- https://github.com/rubocop/rubocop-rspec/actions/runs/19120657194/job/54741707842?pr=2129

docs/modules/ROOT/pages/cops_rspec.adoc
```diff
@@ -6098,15 +6129,15 @@ whatever_spec.rb         # describe MyClass, type: :routing do; end
 | Array
 
 | CustomTransform
-| `{"RuboCop" => "rubocop", "RSpec" => "rspec"}`
+| `{"RuboCop"=>"rubocop", "RSpec"=>"rspec"}`
```
bquorning: Could you please regenerate the documentation, but using Ruby 3.4?

I’ve opened #2130 to mitigate this problem for future contributors.

kevinrobell-st: I installed Ruby 3.4 and then it formatted things differently so that must be it. I rebased and pushed those changes.

lib/rubocop/cop/rspec/output.rb
```diff
@@ -0,0 +1,66 @@
+# frozen_string_literal: true
+
+module RuboCop
+  module Cop
+    # NOTE: This is the same as the `Rails/Output` cop with minor changes.
```
ydah: Could you please delete this comment? It will diverge significantly from the actual Rails/Output cop in the future.

lib/rubocop/cop/rspec/output.rb
```diff
@@ -0,0 +1,66 @@
...
+        def on_send(node)
+          return if node.parent&.call_type? || node.block_node
+          return if !output?(node) && !io_output?(node)
+          return if node.arguments.any? { |arg| arg.type?(:hash, :block_pass) }
+
+          range = offense_range(node)
+
+          add_offense(range)
+        end
+        # rubocop:enable Metrics/CyclomaticComplexity
```
ydah: Could you please add the directive to the specific line where the offense occurs, instead of using a range?
```suggestion
        def on_send(node) # rubocop:disable Metrics/CyclomaticComplexity
          return if node.parent&.call_type? || node.block_node
          return if !output?(node) && !io_output?(node)
          return if node.arguments.any? { |arg| arg.type?(:hash, :block_pass) }

          range = offense_range(node)

          add_offense(range)
        end
```

lib/rubocop/cop/rspec/output.rb
kevinrobell-st:
```suggestion
    # NOTE: Originally based on the `Rails/Output` cop.
```

Maybe something like this. It still shows the history but doesn't indicate that the the changes will stay minor going forward.

ydah:
Could you please squashed related commits into one?

lib/rubocop/cop/rspec/output.rb
```diff
@@ -0,0 +1,64 @@
...
+        # @!method io_output?(node)
+        def_node_matcher :io_output?, <<~PATTERN
+          (send
+            {
+              (gvar #match_gvar?)
+              (const {nil? cbase} {:STDOUT :STDERR})
+            }
+            {:binwrite :syswrite :write :write_nonblock}
+            ...)
+        PATTERN
```
r7kamura: Regarding the fact that method names like `p` and `pp` appear twice, I believe there will be difficulties in maintainability if you want to add or remove corresponding methods in the future. So how about managing them collectively as constants for each group?

I understand that the implementation was brought from `Rails/Output`, but I still think there is room for improvement, so why not take this opportunity to make it better?

```suggestion

        KERNEL_METHODS = %i[
          ap
          p
          pp
          pretty_print
          print
          puts
        ].to_set.freeze
        private_constant :KERNEL_METHODS

        IO_METHODS = %i[
          binwrite
          syswrite
          write
          write_nonblock
        ].to_set.freeze
        private_constant :IO_METHODS

        RESTRICT_ON_SEND = (KERNEL_METHODS + IO_METHODS).to_a.freeze

        # @!method output?(node)
        def_node_matcher :output?, <<~PATTERN
          (send nil? KERNEL_METHODS ...)
        PATTERN

        # @!method io_output?(node)
        def_node_matcher :io_output?, <<~PATTERN
          (send
            {
              (gvar #match_gvar?)
              (const {nil? cbase} {:STDOUT :STDERR})
            }
            IO_METHODS
            ...)
        PATTERN
```

lib/rubocop/cop/rspec/output.rb
```diff
@@ -0,0 +1,64 @@
...
+
+        def on_send(node) # rubocop:disable Metrics/CyclomaticComplexity
+          return if node.parent&.call_type? || node.block_node
+          return if !output?(node) && !io_output?(node)
+          return if node.arguments.any? { |arg| arg.type?(:hash, :block_pass) }
+
+          range = offense_range(node)
+
+          add_offense(range)
```
r7kamura: I think `Rails/Output` is limiting the range of its offenses because its main point is "please use `Rails.logger.debug` instead." 

```ruby
# bad
puts 'A debug message'
^^^^

# good
Rails.logger.debug 'A debug message'
```

On the other hand, this cop is different in nature; it argues that the method call itself should be removed. 

```ruby
# bad
puts 'A debug message'
^^^^^^^^^^^^^^^^^^^^^^

# good
```

Therefore, shouldn't the entire node be the target of the offense?

```suggestion
          add_offense(node)
```

r7kamura: How about adding an autocorrect that removes the node while setting `AutoCorrect: contextual`?

```suggestion
          add_offense(node) do |corrector|
            corrector.remove(node)
          end
```

r7kamura:
I agree with adding this cop with `Enabled: pending`, as there should be sufficiently beneficial use cases.

Also, I believe the cop name `RSpec/Output` is appropriate. It should be easily understood by users, especially since there is already a well-known cop named `Rails/Output`.

As I commented inline in the code, I believe that contextual autocorrect is preferable, but this is a point where opinions may differ. What does everyone else think?

config/default.yml
```diff
@@ -758,6 +758,13 @@ RSpec/NotToNot:
   VersionAdded: '1.4'
   Reference: https://www.rubydoc.info/gems/rubocop-rspec/RuboCop/Cop/RSpec/NotToNot
 
+RSpec/Output:
+  Description: Checks for the use of output calls like puts and print in specs.
+  Enabled: pending
```
r7kamura, line 763: As I commented below, I think `AutoCorrect: contextual` should be set.

- https://github.com/rubocop/rubocop-rspec/pull/2129#discussion_r2587039982

This setting primarily relates to the behavior of autofix in editors like VSCode. If this is not set, as soon as a user writes `puts ...` for debugging and saves, that code will be erased by autofix, making debugging difficult.

```suggestion
  Enabled: pending
  AutoCorrect: contextual
```

lib/rubocop/cop/rspec/output.rb
```diff
@@ -0,0 +1,73 @@
+# frozen_string_literal: true
+
+module RuboCop
+  module Cop
+    # NOTE: Originally based on the `Rails/Output` cop.
+    module RSpec
+      # Checks for the use of output calls like puts and print in specs.
```
r7kamura, line 7: I believe setting `SafeAutoCorrect: false` is the correct approach.
Furthermore, when setting `SafeAutoCorrect: false`, it is customary to include its explanation with `@safety`, so it would be good to document it in this cop as well.

```suggestion
      # Checks for the use of output calls like puts and print in specs.

      # @safety
      #   ...
```

kevinrobell-st:
Okay, I added a `@safety` comment and marked the new cop as `AutoCorrect: contextual`. For some reason, I thought that happened automatically for unsafe cops so that's why I didn't add it before.

PR #2130, opened 2025-11-06
Add a .ruby-version file

bquorning:
In two recent pull requests, new contributors have had CI failing because they used Ruby 3.3 (or older) to run the documentation Rake tasks. (Ruby 3.4 formats hashes a bit differently, adding whitespace around hashrockets.)

By adding a .ruby-version file, most contributors will automatically be using the same Ruby version. Ideally we would not need to specify the patch version in .ruby-version, but I believe that the `major.minor.patch` format is what is compatible with the most Ruby version managers out there.

I found this discussion about writing a specification for .ruby-version, but it doesn't seem to be completed yet: https://github.com/ruby/rubygems/discussions/7074


pirj:
Thank you!

PR #2132, opened 2025-11-12
Enhances RSpec/MultipleExpectations message to suggest aggregate_failure

### LINKED ISSUE #2131: Suggest usage of aggregate_failures directly in RSpec/MultipleExpectation error message
## Is your feature request related to a problem? Please describe.

When encountering the RSpec/MultipleExpectation cop violation, developers often default to splitting specs into multiple examples to satisfy the "one expectation per example" rule, without realizing that aggregate_failures is a better alternative. I discovered the aggregate_failures option late in my refactoring process, which meant I had unnecessarily split many test cases that would have been better structured using aggregated failures. The problem is that the cop's error message doesn't educate users about aggregate_failures, so developers only learn about this option if they dig into the documentation. Since this cop is straightforward to "fix" by splitting specs, most users never consult the docs and miss out on a more appropriate solution.

## Describe the solution you'd like

Enhance the RSpec/MultipleExpectation cop error message to proactively suggest using aggregate_failures as an alternative to splitting specs. Similar to other RuboCop messages that use "Prefer using..." or "Consider using..." language, this would educate developers at the point of encounter. For example, the error message could be:

```
Example has too many expectations [2/1]. Consider using `aggregate_failures` if these expectations are logically related, or split into separate examples.
```
This would:
- Educate users about aggregate_failures when they first encounter the violation
- Help them make informed decisions about whether to aggregate or split
- Improve test suite structure by encouraging proper use of aggregate_failures
- Save significant time by preventing unnecessary spec splitting and later refactoring

## Describe alternatives you've considered

Not much here, maybe auto correction but this seems not viable since the cop can't determine whether expectations should be aggregated or split into separate examples

svgr-slth:
# Summary

This PR improves the RSpec/MultipleExpectations cop error message to proactively suggest using aggregate_failures as an alternative to splitting specs.

Before:
```Example has too many expectations [2/1].```

After:
```Example has too many expectations [2/1]. Consider using `aggregate_failures` if these expectations are logically related.```

# Motivation

Developers often encounter this cop and default to splitting specs without realizing aggregate_failures is available. This enhancement educates users at the point of encounter, improving both developer experience and test suite quality.

# Changes

Updated error message in RSpec::MultipleExpectations cop
Updated all test cases to expect the new message format

Fixes #2131

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
 - Fix a false positive for `RSpec/ReceiveNever` cop when `allow(...).to receive(...).never`. ([@ydah])
 - Fix detection of nameless doubles with methods in `RSpec/VerifiedDoubles`. ([@ushi-as])
 - Improve an offense message for `RSpec/RepeatedExample` cop. ([@ydah])
+- Improve `RSpec/MultipleExpectations` message to suggest using `aggregate_failures`. ([@svgr-slth])
 - Let `RSpec/SpecFilePathFormat` leverage ActiveSupport inflections when configured. ([@corsonknowles], [@bquorning])
 
 ## 3.7.0 (2025-09-01)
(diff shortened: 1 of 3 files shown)


ydah (commented):
Thank you for your work. But I believe this change is not appropriate, as the offense message does not apply when `aggregate_failures: false` is used.

bquorning (approved):
Looks good. Please rebase & squash commits.

bquorning:
Closing & reopening to trigger a CI run.

pirj:
See https://github.com/rubocop/rubocop-rspec/issues/2131#issuecomment-5651747110

PR #2138, opened 2025-11-24
Fix false positive in `RSpec/LetSetup` cop when `let!` variables are used in blocks declared in an outer context

### LINKED ISSUE #2136: `RSpec/LetSetup` false positive when used only on a block declared on an outer context
Similar to #2075, but on this case the false positive is caused when using the variable in a block on an outside context:

```ruby
# frozen_string_literal: true

describe 'False positive example', type: :system do
  after do
    authenticator.remove!
  end

  context 'that defines an authenticator' do
    let!(:authenticator) { add_virtual_authenticator_with_some_options } # <-- offense, but shouldn't!

    it 'does something without using the authenticator variable' do
      # ...      
    end
  end

  context 'that defines another authenticator' do
    let!(:authenticator) { add_virtual_authenticator_with_other_options } # <-- offense, but shouldn't!

    it 'does something without using the authenticator variable' do
      # ...
    end
  end
end
```

Note that if the `let!` is placed at the same level as the block that used the variable, there's no offense:

```ruby
# frozen_string_literal: true

describe 'False positive example', type: :system do
  let!(:authenticator) { add_virtual_authenticator } # <-- no offense!

  after do
    authenticator.remove!
  end

  context 'that defines an authenticator' do
    it 'does something without using the authenticator variable' do
      # ...      
    end
  end
end
```

## Expected behavior

No offense is emitted as the variables declared in the `let!`s are being used in the spec – as it happens when the `let!` is declared at the same level as the block that uses the variable.

## Actual behavior

The `let!`s report an `RSpec/LetSetup` offense.

## RuboCop RSpec version

```shell
$ bundle exec rubocop -V
1.81.7 (using Parser 3.3.9.0, rubocop-ast 1.48.0, analyzing as Ruby 2.7, running on ruby 3.4.4) [arm64-darwin24]
  - rubocop-rspec 3.8.0
  - rubocop-rails 2.34.0
```

ydah:
Fixes: #2136 

- [x] Added tests.
- [-] Updated documentation.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Master (Unreleased)
 
+- Fix false positive in `RSpec/LetSetup` cop when `let!` variables are used in blocks declared in an outer context. ([@ydah])
+
 ## 3.8.0 (2025-11-12)
 
 - Add new cop `RSpec/LeakyLocalVariable`. ([@lovro-bikic])
--- a/lib/rubocop/cop/rspec/let_setup.rb
+++ b/lib/rubocop/cop/rspec/let_setup.rb
@@ -72,8 +72,9 @@ module RuboCop
         def unused_let_bang(node)
           child_let_bang(node) do |method_send, method_name|
             next if overrides_outer_let_bang?(node, method_name)
+            next if method_called_in_scope?(node, method_name.to_sym)
 
-            yield(method_send) unless method_called?(node, method_name.to_sym)
+            yield(method_send)
           end
         end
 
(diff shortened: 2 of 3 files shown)


pirj:
In the case when there's code that is accesing the memoized helper in the outer context, there seems to be no apparent reason to use the bang-`let!` over regular `let`. The regular `let` won't trigger the cop. See https://github.com/rubocop/rubocop-rspec/issues/2136#issuecomment-5651795961

Closing, unless there's some evidence of usefulness of such a change.

PR #2140, opened 2025-11-27
Fix false negative for RSpec/DescribedClass with constant base

lovro-bikic:
Currently, `RSpec/DescribedClass` won't register offenses when described class uses a constant base and the reference doesn't, and vice versa. For example:
```ruby
RSpec.describe MyClass do
  before do
    MyClass.foo # RSpec/DescribedClass registers an offense
    ::MyClass.foo # RSpec/DescribedClass does not register an offense currently
  end
end
```

This PR adds handling of constant bases to this cop.

- [x] Added tests.
- [x] Updated documentation.

### DIFF
--- a/lib/rubocop/cop/rspec/described_class.rb
+++ b/lib/rubocop/cop/rspec/described_class.rb
@@ -50,6 +50,11 @@ module RuboCop
       #     subject { MyClass::CONSTANT }
       #   end
       #
+      #   # good
+      #   describe MyClass do
+      #     subject { described_class::CONSTANT }
+      #   end
+      #
       # @example `EnforcedStyle: explicit`
       #   # bad
       #   describe MyClass do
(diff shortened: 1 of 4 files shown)


lib/rubocop/cop/rspec/described_class.rb
```diff
@@ -50,6 +50,11 @@ module RSpec
       #     subject { MyClass::CONSTANT }
       #   end
       #
+      #   # good
+      #   describe MyClass do
+      #     subject { described_class::CONSTANT }
+      #   end
```
lovro-bikic, line 56: I added this because I think a "good" example was missing here

spec/rubocop/cop/rspec/described_class_spec.rb
```diff
@@ -438,6 +488,16 @@ module SomeGem
...
+
+    it 'ignores parenthesized namespaces within modules - begin_types' do
+      expect_no_offenses(<<~RUBY)
+        module Foo
+          describe MyClass do
+            subject { (MyNamespace)::MyClass }
+          end
+        end
+      RUBY
+    end
```
lovro-bikic, line 500: This test is currently failing on master

pirj: TIL this syntax is valid

spec/rubocop/cop/rspec/described_class_spec.rb
```diff
@@ -130,6 +130,56 @@
...
+        describe ::MyClass do
+          include described_class
+
+          subject { described_class.do_something }
+
+          before { described_class.do_something }
+        end
+      RUBY
+    end
+
```
pirj, line 182: Would the cop register an offence for:
```
module Foo
  RSpec.describe Bar do
    subject { ::Bar.new }
  end
end
```
?

PR #2144, opened 2025-12-08
Remove .ruby-version file

ydah:
This PR removes the `.ruby-version` file

## Why

The goal is to prevent forcing a specific Ruby version on contributors. Developers should be able to work on this repository regardless of their local Ruby version.

## Details

* Previously, the Ruby version was pinned solely to address inconsistencies in hash syntax during documentation generation, not due to internal logic issues.
* I believe pinning the version for this reason is not ideal.
* With this change, documentation generation now produce consistent results without relying on a fixed Ruby version.
* Contributors no longer need to worry about or strictly adhere to a specific Ruby version.

______________________________________________________________________

Before submitting the PR make sure the following are checked:

- [x] Feature branch is up-to-date with `master` (if not - rebase it).
- [x] Squashed related commits together.
- [-] Added tests.
- [-] Updated documentation.
- [-] Added an entry to the `CHANGELOG.md` if the new code introduces user-observable changes.
- [x] The build (`bundle exec rake`) passes (be sure to run this locally, since it may produce updated documentation that you will need to commit).

PR #2147, opened 2025-12-15
Fix a false positive for `RSpec/ScatteredSetup` when the hook is defined inside a class method

d4rky-pl:
This PR fixes a false positive in `RSpec/ScatteredSetup` by ignoring the hooks defined inside class methods.

- [x] Added tests.
- [x] Updated documentation.

### DIFF
--- a/lib/rubocop/cop/rspec/scattered_setup.rb
+++ b/lib/rubocop/cop/rspec/scattered_setup.rb
@@ -8,6 +8,7 @@ module RuboCop
       # Unify `before` and `after` hooks when possible.
       # However, `around` hooks are allowed to be defined multiple times,
       # as unifying them would typically make the code harder to read.
+      # Hooks defined in class methods are also ignored.
       #
       # @example
       #   # bad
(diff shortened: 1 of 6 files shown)


bquorning:
> I didn't want to chase this and figure out if this is Ruby version or some specific dependency getting silently updated due to unpinned version so I reverted those changes.

You should use Ruby 3.4 (as specified in .ruby-version), but an Ruby version agnostic alternative has been suggested in #2144. For now, reverting/ignoring the documentation change was the correct call. 👍🏼

lib/rubocop/cop/rspec/scattered_setup.rb
```diff
@@ -56,7 +65,7 @@ def on_block(node) # rubocop:disable InternalAffairs/NumblockHandler
         def repeated_hooks(node)
           hooks = RuboCop::RSpec::ExampleGroup.new(node)
             .hooks
-            .select { |hook| hook.knowable_scope? && hook.name != :around }
+            .select { |hook| repeated_hook?(hook) }
```
bquorning: The “repeated hook” is really the whole thing (old lines 56–62), not just what you extracted to a method. I would probably filter twice instead of extracting the method:
```suggestion
            .reject(&:inside_class_method?)
            .select { |hook| hook.knowable_scope? && hook.name != :around }
```

If RuboCop tells you there’s too many lines, or the code gets too complex, feel free to add an inline disable and we’ll refactor later.

bquorning (approved):
@ydah What is your opinion?

d4rky-pl:
@bquorning I had to push another fix because I've realized too late that it won't work with multiline `class << self` due to parser's implicit `begin`. It then took me forever to hit 100% coverage and 100% rubocop and have all the tests 😅 

The PR should be OK now, apologies for pushing too soon.

PR #2148, opened 2025-12-15
Fix RSpec/DescribedClass to ignore *_eval and *_exec blocks

sucicfilip:
Fix false positives for RSpec/DescribedClass inside eval blocks

`RSpec/DescribedClass` currently registers offenses for constant instantiations inside `class_eval` / `module_eval` / `instance_eval` blocks.

In such contexts, `described_class` is not available nor semantically correct, as the code is defining or operating on dynamically evaluated classes/modules.

This PR makes the cop ignore nodes inside eval blocks, preventing false positives and removing the need to disable the cop in valid RSpec setups.

- [x] Added tests.
- [x] Updated documentation.

### DIFF
--- a/spec/rubocop/cop/rspec/described_class_spec.rb
+++ b/spec/rubocop/cop/rspec/described_class_spec.rb
@@ -195,6 +195,34 @@ RSpec.describe RuboCop::Cop::RSpec::DescribedClass do
       RUBY
     end
 
+    it 'ignores class inside *_eval and *_exec blocks' do
+      expect_no_offenses(<<~RUBY)
+        RSpec.describe Foo do
+          before do
+            stub_const('Dummy', Class.new).class_eval do
+              Foo.new
+            end
+
+            stub_const('Dummy', Class.new).module_eval do
+              Foo.new
+            end
+
+            stub_const('Dummy', Class.new).instance_eval do
+              Foo.new
+            end
+
+            stub_const('Dummy', Class.new).class_exec do
+              Foo.new
+            end
+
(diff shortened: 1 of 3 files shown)


r7kamura:
> My plan would be to add :class_exec, :instance_exec, and :module_exec to the existing :common_instance_exec_closure? node matcher

Certainly. It seems good to add it there. I think it would be clearer if `instance_eval` and the like were also included there. There's no reason to distinguish it from `instance_exec`.

In fact, it's strange that `instance_exec` and the like haven't been written here until now, despite this name... I've read through some of the changelogs from when this was added, but I couldn't figure out the background. There might not be a deep reason for it.

- https://github.com/rubocop/rubocop-rspec/pull/147

> Should I include this in the current PR or open a separate one?

It seems appropriate to include it in the same pull request 👍 
In this project, there's also a rule to squash commits included in a pull request. So, please add that change, squash the commits together into one commit, and then modify the changelog, commit message, and pull request title accordingly.

By the way, the diff includes changes related to `VersionChanged`, which I believe are unnecessary. It would be good if you could remove them as well.

spec/rubocop/cop/rspec/described_class_spec.rb
```diff
@@ -195,6 +195,34 @@ module MyModule
       RUBY
     end
 
+    it 'ignores class inside *_eval and *_exec blocks' do
```
sucicfilip, line 198: Just wanted to edit this so it's more in tune with the existing specs :-)

r7kamura:

PR #2150, opened 2026-01-07
Fix CHANGELOG

kevinrobell-st:
The `RSpec/Output` cop was merged in on 2025/12/7 but I accidentally added it to the already released changelog entry.

- Related PR: https://github.com/rubocop/rubocop-rspec/pull/2129


bquorning (approved):
Good catch, thank you for the fix.

PR #2155, opened 2026-01-16
Extract helper module to detect repeated nodes

### LINKED ISSUE #996: Extract helper module for duplicate detection
#994 would be the 5th cop that deals with repeated usage, and all the cops share similar structure for the detection: node.children.select { ... }.group_by { .... }.values.reject(&:one?).flat_map { repeated_lines }

It's time to extract a module for this and DRY the code

Darhazer:
Fixes #996 

- [ ] ~Added tests.~


lib/rubocop/cop/rspec/mixin/repeated_items.rb
```diff
@@ -0,0 +1,36 @@
...
+        # @param items [Enumerable] the filtered collection to group
+        # @param key_proc [Proc] block returning the grouping key for each item
+        # @return [Array<Array>] array of groups containing more than one item
+        #   that share the same key and there are multiple items in the group
+        def find_repeated_groups(items, key_proc:)
+          items
+            .group_by(&key_proc)
+            .values
+            .reject(&:one?)
+        end
```
r7kamura, line 23: Wouldn't it be simpler to just use a block instead of passing a Proc instance as an argument?

```suggestion
        def find_repeated_groups(items, &block)
          items
            .group_by(&block)
            .values
            .reject(&:one?)
        end
```

r7kamura:
As a general policy, I agree with this change 👍

lib/rubocop/cop/rspec/mixin/repeated_items.rb
bquorning: I actually found the code a bit confusing, when I tried rewriting it with an implicit block just now. If we do do that however, could we rename the block argument from `&block` to e.g. `&signature_proc`?

r7kamura: That makes sense.

Without keyword arguments to label things, it can be hard to tell what's going on from the caller’s side. In that case, the original way of passing a Proc via keyword arguments definitely has its perks. Alternatively, it might improve things somewhat to change the method name to something like `find_repeated_groups_by`, which clearly implies that grouping logic is expected.

Since both approaches seem to have their pros and cons, I felt that leaving it as it is would also be fine 👌

r7kamura (approved):

bquorning (approved):

PR #2158, opened 2026-01-19
Add RSpec/HaveAttributes cop

### LINKED ISSUE #1057: Cop idea: HaveAttributes
If a test checks few attributes in a row, it probably deserve to use `have_attributes`
bad:
```ruby
expect(obj.foo).to eq(bar)
expect(obj.fu).to eq(bax)
expect(obj.name).to eq(baz)
```
good:
```
expect(obj).to have_attributes(
foo: bar,
fu: bax
name: baz
```
Things to consider:
the used matches. Initially, only eq could be used, as it could turn out to be quite complex to detect what can be converted to have_attributes and what not. We need a list of matching matchers.
An initial list:
```
be < 2 => a_value < 2
be > 2 => a_value > 2
be_an_instance_of => an_instance_of
be_within => a_value_within
contain_exactly => a_collection_containing_exactly
end_with => a_string_ending_with, ending_with
match => a_string_matching
start_with => a_string_starting_with
```

Check only direct method calls, no method chains
The method calls should be without arguments

Darhazer:
Fixes #1057

- [x] Added tests.
- [x] Updated documentation.

### DIFF
--- a/lib/rubocop/cop/rspec/have_attributes.rb
+++ b/lib/rubocop/cop/rspec/have_attributes.rb
@@ -158,14 +158,36 @@ module RuboCop
           attr_reader :sorted_nodes
 
           def build_attributes
-            sorted_nodes.map do |exp|
-              method_name = exp[:method]
-              matcher = exp[:matcher]
-              value = exp[:value]
+            method_groups = {}
+            sorted_nodes.each do |exp|
+              (method_groups[exp[:method]] ||= []) << exp
+            end
 
-              transformed_value = transform_value(matcher, value)
-              "#{method_name}: #{transformed_value}"
+            method_groups.map do |method_name, exps|
+              value_str = if exps.size == 1
+                            exp = exps.first
+                            transform_value(exp[:matcher], exp[:value])
+                          else
+                            combine_matchers(exps)
+                          end
+              "#{method_name}: #{value_str}"
             end.join(",\n    ")
+          end
+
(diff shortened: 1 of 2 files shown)


Darhazer:
@bquorning I have updated the cop to ignore non-consecutive expectations. I'm yet to check for false negatives against a real code base, but seems the false positives are fixed and I no longer have cases when the specs are broken due to the auto-correction. Would appreciate another look / test against your code base

bquorning:
Sorry for the delay. I ran this new cop on my code base and found no serious problems. I do have two comments, though:

1. When used with Rails, date comparisons don’t work exactly the same in `eq` and `have_attributes`. E.g.
   ```rb
   expect(foo.created_at).to eq("2025-01-27T15:53:13Z")
   ```
   ends up in [Time#==](https://github.com/rspec/rspec/blob/rspec-expectations-v3.13.5/rspec-expectations/lib/rspec/matchers/built_in/eq.rb#L39) but
   ```rb
   expect(foo).to have_attributes(
     created_at: "2025-01-27T15:53:13Z"
   )
   ```
   ends up in [String#==](https://github.com/rspec/rspec/blob/rspec-support-v3.13.7/rspec-support/lib/rspec/support/fuzzy_matcher.rb#L20-L23).
2. It would be nice to also treat `be_nil`, `be(nil)`, `be(true)`, and `be(false)` as attribute checks.

PR #2166, opened 2026-03-31
Add new RSpec/ControllerSpecDirectory and RSpec/ControllerSpecType cops

CariWest:

- [x] Added tests.
- [x] Updated documentation.


ydah:
Thank you for your work! I think these cops may fit better in rubocop-rspec_rails than in rubocop-rspec.

Both cops encode Rails-specific testing conventions: `spec/controllers`, `type: :controller`, and the recommendation to move to request specs. That feels more aligned with rubocop-rspec_rails, which is the Rails-specific extension and already owns adjacent behavior via `RSpecRails/InferredSpecType` and its default controller/request mappings.

I also checked for similar cops and couldn't find other controller/request-specific cops in rubocop-rspec itself. The closest things there seem to be more general cops, such as `RSpec/DescribeClass`, which only special-case Rails metadata rather than enforce Rails testing conventions.

So unless there is a deliberate reason to keep this kind of Rails-specific guidance in the core gem, would it make sense to move these cops to rubocop-rspec_rails?

CariWest:
> I think these cops may fit better in rubocop-rspec_rails than in rubocop-rspec.

@ydah That makes sense to me! I can close this and move there — thanks so much for the feedback.

PR #2169, opened 2026-04-30
Add new cop `RSpec/MatchWithSimpleRegex`

bquorning:
Adds a new cop that suggests using `include` matcher when `expect(...).to match()` is called with a regex that contains only literal characters without regex-specific features (anchors, character classes, quantifiers, alternations, metacharacters).

See also https://github.com/rubocop/rubocop-minitest/pull/348.

## Changes

- Adds detection for `match` calls with regexes that contain only literal characters.
- Suggests `include` as a more semantic and clearer alternative for simple substring matching.
- Makes `regexp_parser` a direct dependency (already a transitive dependency via `rubocop`) to properly parse and analyze regex patterns.
- Handles escaped literals in regexes (e.g., `/http:\/\/example\.com/` → `'http://example.com'`).
- Converts simple regexes to their string equivalents with proper quote handling.

## Examples

```ruby
# bad
expect('foobar').to match(/foo/)
expect(response.body).to match(/http:\/\/example\.com/)

# good
expect('foobar').to include('foo')
expect(response.body).to include('http://example.com')

# still uses match (regex features needed)
expect('foobar').to match(/^foo/)     # has anchor
expect('foobar').to match(/foo\d+/)   # has quantifier
expect('foobar').to match(/foo.bar/)  # has metacharacter
```

- [x] Added tests.
- [x] Updated documentation.

🤖 Generated with [Claude Code](https://claude.ai/claude-code)


pirj (approved):
Looks good, thank you!

A few thoughts out loud below.

Should the cop also consider ^$\A\Z as simple and suggest start_with/end_with?

Is there a difference between ‘match’ and ‘include’ for matches against multiline strings?

With the %r syntax, is quote escaping correct in auto-correction ?

bquorning:
> A few thoughts out loud below.
> 
> Should the cop also consider ^$\A\Z as simple and suggest start_with/end_with?

That’s a good idea. But let’s do it in a later PR.

> Is there a difference between ‘match’ and ‘include’ for matches against multiline strings?

I think that using a regex to match against a multiline string without using e.g. `.` or `\s` matchers would be odd. But I added a spec for exactly that case and it seems to work.

> With the %r syntax, is quote escaping correct in auto-correction ?

Can you give me a particular case you’re worried about?

pirj:
Checked with `%r{ }` with mixed quotes - all good, too. I don't think it's worth duplicating an example, you already added one with `/"'/`, it pretty much covers it.

Thanks again!

bquorning:
Specs will pass after https://github.com/rubocop/rubocop-rspec/pull/2170 is merged.

pirj:
@copilot resolve the merge conflicts in this pull request

pirj (approved):

PR #2171, opened 2026-05-03
Catch argumentless matchers in `RSpec/ExpectActual`

cvx:
Flags e.g. `expect(".foo").to be_present` and `expect(1).to be`.

- [x] Added tests.
- [x] Updated documentation.

### DIFF
--- a/docs/modules/ROOT/pages/cops_rspec.adoc
+++ b/docs/modules/ROOT/pages/cops_rspec.adoc
@@ -2178,7 +2178,9 @@ expect(pattern).to eq(/foo/)
 expect(name).to eq("John")
 
 # bad (not supported autocorrection)
+expect(42).to be_even
 expect(false).to eq(true)
+expect("user").to be_present
 ----
 
 [#configurable-attributes-rspecexpectactual]
--- a/lib/rubocop/cop/rspec/expect_actual.rb
+++ b/lib/rubocop/cop/rspec/expect_actual.rb
@@ -19,12 +19,15 @@ module RuboCop
       #   expect(name).to eq("John")
       #
       #   # bad (not supported autocorrection)
+      #   expect(42).to be_even
       #   expect(false).to eq(true)
+      #   expect("user").to be_present
       #
       class ExpectActual < Base
         extend AutoCorrector
 
         MSG = 'Provide the actual value you are testing to `expect(...)`.'
+        MSG_NO_ARG = 'Test a non-literal value with `expect(...)`.'
 
         RESTRICT_ON_SEND = Runners.all
 
@@ -65,9 +68,21 @@ module RuboCop
           )
         PATTERN
 
+        # @!method expect_literal_no_arg(node)
+        def_node_matcher :expect_literal_no_arg, <<~PATTERN
+          (send
+            (send nil? :expect $#literal?)
+            #Runners.all
+            $(send nil? $_)
+          )
+        PATTERN
+
         def on_send(node)
           expect_literal(node) do |actual, send_node, matcher, expected|
             register_offense(actual, send_node, matcher, expected)
+          end
+          expect_literal_no_arg(node) do |actual, send_node, matcher|
+            register_offense(actual, send_node, matcher, nil)
           end
         end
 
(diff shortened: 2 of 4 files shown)


spec/rubocop/cop/rspec/expect_actual_spec.rb
```diff
@@ -227,14 +227,17 @@
     RUBY
   end
 
-  it 'ignores `be` with no argument' do
-    expect_no_offenses(<<~RUBY)
+  it 'flags `be` with no argument' do
```
cvx, line 230: This case wasn't supported only to simplify implementation afaict (see #858)

docs/modules/ROOT/pages/cops_rspec.adoc
```diff
@@ -2168,11 +2168,13 @@ Autocorrection is performed when the expected is not a literal.
 expect(5).to eq(price)
 expect(/foo/).to eq(pattern)
 expect("John").to eq(name)
+expect(".my-css-class").to be_present
```
ydah: Could you rewrite this to use RSpec use cases as examples instead of Capybara-specific ones?

cvx: done!

lib/rubocop/cop/rspec/expect_actual.rb
```diff
@@ -12,11 +12,13 @@ module RSpec
       #   expect(5).to eq(price)
       #   expect(/foo/).to eq(pattern)
       #   expect("John").to eq(name)
+      #   expect("user").to be_present
       #
       #   # good
       #   expect(price).to eq(5)
       #   expect(pattern).to eq(/foo/)
       #   expect(name).to eq("John")
+      #   expect(user).to be_present
```
bquorning: The other 3 are autocorrections of the `bad` examples above, so let’s remove this one.
```suggestion
```

lib/rubocop/cop/rspec/expect_actual.rb
```diff
@@ -12,11 +12,13 @@ module RSpec
       #   expect(5).to eq(price)
       #   expect(/foo/).to eq(pattern)
       #   expect("John").to eq(name)
+      #   expect("user").to be_present
```
bquorning: I think this should be moved to the `bad (not supported autocorrection)` section below. Perhaps add e.g. `expect(42).to be_even` too?

spec/rubocop/cop/rspec/expect_actual_spec.rb
```diff
@@ -339,6 +342,21 @@
     expect_no_corrections
   end
 
+  it 'flags literal values with argumentless predicate matchers' do
+    expect_offense(<<~RUBY)
+      describe Foo do
+        it 'uses expect incorrectly' do
+          expect("user").to be_present
+                 ^^^^^^ Provide the actual value you are testing to `expect(...)`.
```
bquorning: Should we add a different offense message for these cases, telling people that they should test an object that is not a literal value?

bquorning (commented):
Thank you for the PR, I think it will be a useful addition. I have added a few comments.

Also, I would prefer if the first commit was extracting the method `#register_offense` (without `next if expected.nil?`) so that the 2nd commit clearly shows what the actual change is.

cvx:
All done!

bquorning (approved):

docs/modules/ROOT/pages/cops_rspec.adoc
```diff
@@ -2168,11 +2168,13 @@ Autocorrection is performed when the expected is not a literal.
 expect(5).to eq(price)
 expect(/foo/).to eq(pattern)
 expect("John").to eq(name)
+expect("user").to be_present
```
bquorning: Oh, it looks like you need to regenerate documentation.

cvx: ah, yup, my bad. fixed!

bquorning (approved):

bquorning:
@cvx Could I ask you to rebase on top of the latest master branch please?

cvx:
Done!

bquorning (approved):

pirj:
Are there real life examples of this? 
How those offences would be handled by the user?

bquorning:
My guess is that offenses would reveal typos, where people are testing literals instead of variables. E.g. they would change `expect(5).to be_odd` to `expect(foo).to be_odd`.

bquorning:
Just now, I used this cop to find an offense in a large project at work. This is a useful addition.

PR #2174, opened 2026-05-21
Fix a false positive for `RSpec/ContainExactly` when `contain_exactly` has multiple splat arguments

### LINKED ISSUE #2173: RSpec/ContainExactly vs multiple splats of ranges
We have this (convoluted yet real) spec:
```ruby
expect(ary).to contain_exactly(*0..9, *100..109, *200..209)
```

(Basically, checking that some simple helper fetches first 10 elements from each hundred.)

`RSpec/ContainExactly` says "Prefer match_array when matching array values", and autocorrect this to:
```ruby
match_array(0..9 + 100..109 + 200..209 + 300..309)
```
...which is not only _not better_ (IMO), but also plain incorrect, even syntactically :)

I believe that the best course of action is **to limit suggestions to exactly one splat**. Because:
* Ranges are not the only class that can be `*`-unpacked, but don't sum like arrays;
* I believe that, in any case, `contain_exactly(*vals)` vs `match_array(vals)` is a definite improvement, but `contain_exactly(*ary1, *ary2, *ary3)` vs `match_array(ary1 + ary2 + ary3)` is questionable and shouldn't be a suggestion

## RuboCop RSpec version

Include the output of `rubocop -V` or `bundle exec rubocop -V` if using Bundler.
If you see extension cop versions (e.g. `rubocop-performance`, `rubocop-rake`, and others)
output by `rubocop -V`, include them as well. Here's an example:

```shell
$ [bundle exec] rubocop -V
1.86.1 (using Parser 3.3.11.1, Prism 1.9.0, rubocop-ast 1.49.1, analyzing as Ruby 3.3, running on ruby 3.3.8) [x86_64-linux]
  - rubocop-rspec 3.9.0
  - test-prof 1.5.2
  - rubocop-rails 2.34.3
  - rubocop-performance 1.26.1
  - rubocop-rake 0.7.1
  - rubocop-capybara 2.22.1
  - rubocop-rspec_rails 2.32.0
  - rubocop-factory_bot 2.28.0
  - rubocop-graphql 1.1.1
```

ydah:
Fixes: https://github.com/rubocop/rubocop-rspec/issues/2173

- [x] Added tests.
- [x] Updated documentation.

### DIFF
--- a/lib/rubocop/cop/rspec/contain_exactly.rb
+++ b/lib/rubocop/cop/rspec/contain_exactly.rb
@@ -12,10 +12,13 @@ module RuboCop
       #
       # @example
       #   # bad
-      #   it { is_expected.to contain_exactly(*array1, *array2) }
+      #   it { is_expected.to contain_exactly(*array) }
       #
       #   # good
-      #   it { is_expected.to match_array(array1 + array2) }
+      #   it { is_expected.to match_array(array) }
+      #
+      #   # good
+      #   it { is_expected.to contain_exactly(*array1, *array2) }
       #
       #   # good
       #   it { is_expected.to contain_exactly(content, *array) }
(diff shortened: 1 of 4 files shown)


lib/rubocop/cop/rspec/contain_exactly.rb
```diff
@@ -35,20 +38,19 @@ def on_send(node)
         private
 
         def check_populated_collection(node)
-          return unless node.each_child_node.all?(&:splat_type?)
+          return unless node.arguments.one? && node.first_argument.splat_type?
```
bquorning, line 33: I feel like `node.arguments.one?` makes the `node.arguments.empty?` check in `#on_send` unnecessary. Maybe we should just write everything in one method now, since the logic just got a lot simpler.

```rb
def on_send(node)
  return unless node.arguments.one? && node.first_argument.splat_type?

    array = node.first_argument.children.first
    corrector.replace(node, "match_array(#{array.source})")
  end
end
```

ydah: Addressed in 61123e83 by inlining the check and autocorrection into `on_send`. Thanks!

bquorning (approved):

PR #2177, opened 2026-05-26
Use simplecov v1

bquorning:
The major new feature (for our purposes) is that Simplecov can now be configured to ignore the "implicit else" branches that previously needed 1. to exist and 2. have their no-op contents wrapped with a `:nocov:` comment.


ydah (approved):
🚀

PR #2183, opened 2026-06-05
Ignore interpolation in MatchWithSimpleRegex

### LINKED ISSUE #2182: RSpec/MatchWithSimpleRegex autocorrect drops string interpolation

Run bundle exec rubocop --only RSpec/MatchWithSimpleRegex -A on the following code:

```
# frozen_string_literal: true

RSpec.describe 'repro' do
  it 'matches an interpolated regex' do
    id = 42
    other = 43
    expect(html).to match(/data-feature_#{id},data-feature_#{other}/)
  end
end
```
## Expected behavior

The cop should not register an offense when the regexp literal contains interpolation or the correction should include the interpolation.

## Actual behavior

The interpolation is dropped and the resulting code looks like:

```
# frozen_string_literal: true

RSpec.describe 'repro' do
  it 'matches an interpolated regex' do
    id = 42
    other = 43
    expect(html).to include('data-feature_,data-feature_')
  end
end
```

## RuboCop RSpec version

```shell
$ [bundle exec] rubocop -V
1.87.0 (using Parser 3.3.11.1, Prism 1.9.0, rubocop-ast 1.49.1, analyzing as Ruby 3.4, running on ruby 3.4.8) [arm64-darwin23]
  - rubocop-capybara 2.23.0
  - rubocop-factory_bot 2.28.0
  - rubocop-performance 1.26.1
  - rubocop-rails 2.35.3
  - rubocop-rspec 3.10.0
  - rubocop-rspec_rails 2.32.0
```

bquorning:
The cop should not register an offense when the regexp literal contains interpolation.

Fixes #2182.

- [x] Added tests.
- [ ] Updated documentation.

### DIFF
--- a/lib/rubocop/cop/rspec/match_with_simple_regex.rb
+++ b/lib/rubocop/cop/rspec/match_with_simple_regex.rb
@@ -54,6 +54,8 @@ module RuboCop
         private
 
         def simple_regexp?(node)
+          return false if node.interpolation?
+
           parsed = Regexp::Parser.parse(node.content)
           parsed.expressions.all? { |expr| simple_expression?(expr) }
         end
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Master (Unreleased)
 
+## 3.10.1 (2026-06-05)
+
 - Add `Strict` option to `RSpec/SharedContext` to flag `shared_context` whenever it contains examples, even alongside setup code. ([@Darhazer])
 - Add `NegatedMatcher` configuration option `RSpec/ExpectChange`. ([@Darhazer])
 - Fix `RSpec/MatchWithSimpleRegex` to ignore regular expressions with interpolations. ([@bquorning])
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 - Add `Strict` option to `RSpec/SharedContext` to flag `shared_context` whenever it contains examples, even alongside setup code. ([@Darhazer])
 - Add `NegatedMatcher` configuration option `RSpec/ExpectChange`. ([@Darhazer])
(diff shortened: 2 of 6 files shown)


lib/rubocop/cop/rspec/match_with_simple_regex.rb
```diff
@@ -40,6 +40,7 @@ class MatchWithSimpleRegex < Base
 
         def on_send(node)
           match_with_regexp?(node) do |regexp|
+            next if regexp.interpolation?
```
Darhazer: minor question:
what do you think to have it as a `return false if regexp.interpolation?` in the `simple_regexp?`. Does it belong there (regexp with interpolation is not simple) or does it belong here (we can't even say whether it's a simple regexp)

Darhazer (approved):

lib/rubocop/cop/rspec/match_with_simple_regex.rb
bquorning: Yeah, good point 👍🏼 I’ll move this check into `#simple_regexp?`

bquorning:
Should I add the necessary changes for a patch release in a commit on this branch before merging?

Darhazer (approved):

Darhazer:
> Should I add the necessary changes for a patch release in a commit on this branch before merging?

The rest of the merged changes are opt-in, do not affect anyone without a change in config, so apart that I'm not sure if that counts for 3.10.1 or 3.11, I don't see a problem releasing together following the normal release process. If we're to release just this fix, then yes, I guess we'll have to do it from this branch

Darhazer:
But since you already did the bump here, 👍

bquorning:
Yeah I know, it’s just a bit of a hassle to separate out the changes to a separate branch, especially now that we have a workflow set to publish automatically when version.rb changes on `master` branch.

PR #2188, opened 2026-06-06
Consider regexp with options a non-simple regexp

### LINKED ISSUE #2185: RSpec/MatchWithSimpleRegex incorrectly flags/corrects regexes with `/i`
/i modifier on regexes indicates case-insensitivity. The contents of the regex should not be extracted into a string for an include comparison.

```ruby
expect("A").to_not(match(/a/i))
```

## Expected behavior

The cop should not autocorrect or flag this as being replaceable with `include` since they're not equivalent.

## Actual behavior

This is flagged as equivalent to `include('a')` 

## RuboCop RSpec version

```shell
1.87.0 (using Parser 3.3.11.1, Prism 1.9.0, rubocop-ast 1.49.1, analyzing as Ruby 4.0, running on ruby 4.0.2) [arm64-darwin25]
  - rubocop-factory_bot 2.28.0
  - rubocop-performance 1.26.1
  - rubocop-rails 2.35.3
  - rubocop-rspec 3.10.0
  - rubocop-rspec_rails 2.32.0
  - rubocop-sorbet 0.12.0
  - rubocop-thread_safety 0.7.3
```

bquorning:
Fixes #2185

- [x] Added tests.
- [ ] Updated documentation.

### DIFF
--- a/spec/rubocop/cop/rspec/match_with_simple_regex_spec.rb
+++ b/spec/rubocop/cop/rspec/match_with_simple_regex_spec.rb
@@ -90,6 +90,17 @@ RSpec.describe RuboCop::Cop::RSpec::MatchWithSimpleRegex, :config do
     RUBY
   end
 
+  it 'does not register an offense when using match with regex options' do
+    expect_no_offenses(<<~RUBY)
+      expect('foobar').to match(/foo/x)
+      expect('foobar').to match(/foo/i)
+      expect('foobar').to match(/foo/m)
+      expect('foobar').to match(/foo/n)
+      expect('foobar').to match(/foo/u)
+      expect('foobar').to match(/foo/o)
+    RUBY
+  end
+
   it 'does not register an offense when using match with string ' \
      'instead of regex' do
     expect_no_offenses(<<~RUBY)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Master (Unreleased)
 
+- Fix `RSpec/MatchWithSimpleRegex` to ignore regular expressions with options. ([@bquorning])
+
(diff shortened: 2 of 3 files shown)


bquorning:
cc @Darhazer

spec/rubocop/cop/rspec/match_with_simple_regex_spec.rb
```diff
@@ -90,6 +90,23 @@
...
+      expect('foobar').to match(/foo/i)
+      expect('foobar').to match(/foo/m)
+      expect('foobar').to match(/foo/n)
+      expect('foobar').to match(/foo/u)
+    RUBY
+  end
+
+  it 'registers an offense when using match with regex option o' do
+    expect_offense(<<~RUBY)
+      expect('foobar').to match(/foo/o)
```
Darhazer, line 100: Why is that? Is it because it does not have an effect if there is no {} in the regexp, and if there is it would be ignored by another condition? I’ve never used /o so I had to google it.
For spec completeness though we need to have /o as not the single option

bquorning: That’s just because of [this detail in rubocop-ast](https://github.com/rubocop/rubocop-ast/blob/ca769d5582f47a21ad2bc17dd417ea9eda6ee990/lib/rubocop/ast/node/regexp_node.rb#L15). But you are right, `/o` regexes cannot safely be converted to `include` specs either.

```
>> a = /./
=> /./
>> /#{a}/o
=> /(?-mix:.)/
>> /#{a}/o.match?("x")
=> true
```

I’ll need to _also_ check for `node.single_interpolation?`.

Darhazer (approved):

PR #2190, opened 2026-06-09
Document custom matchers and helper modules

Darhazer:
Since my colleagues started writing internal rspec cops, I've noticed they are doing it as if it was an ordinary rubocop cop. Completely unaware of some DSLs that we add, not using the Language module, etc. So I've asked Claude to add those to the development documentation. It is a bit more extensive that I expected (downside: we may commit to support all that as it's more likely people will start relying on it, upside: now that it's visible we can also take some explicit decisions on the design of those helpers), but I believe it will be useful for the cop-writing community

### DIFF
--- a/docs/modules/ROOT/pages/development.adoc
+++ b/docs/modules/ROOT/pages/development.adoc
@@ -45,9 +45,239 @@ for good and bad examples (old name is on the left, new name on the right).
 
 == Base class
 
-The `RuboCop::Cop::RSpec::Base` class includes convenient https://docs.rubocop.org/rubocop-ast/node_pattern.html[node pattern DSL] matchers that will automatically account for any xref:usage.adoc#rspec-dsl-configuration[custom RSpec DSL configuration].
+The `RuboCop::Cop::RSpec::Base` class includes the `RuboCop::RSpec::Language` module, which provides convenient https://docs.rubocop.org/rubocop-ast/node_pattern.html[node pattern DSL] matchers that will automatically account for any xref:usage.adoc#rspec-dsl-configuration[custom RSpec DSL configuration].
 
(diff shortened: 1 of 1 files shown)


docs/modules/ROOT/pages/development.adoc
```diff
@@ -45,10 +45,240 @@ for good and bad examples (old name is on the left, new name on the right).
...
+The `RuboCop::Cop::RSpec::Base` class includes the `RuboCop::RSpec::Language` module, which provides convenient https://docs.rubocop.org/rubocop-ast/node_pattern.html[node pattern DSL] matchers that will automatically account for any xref:usage.adoc#rspec-dsl-configuration[custom RSpec DSL configuration].
 
 For example, if the project defines https://github.com/test-prof/test-prof/blob/master/docs/recipes/let_it_be.md[`let_it_be`] as a `Helper`, then all cops will find `let_it_be` when using the `let?` matcher.
 
+=== Language module and node pattern DSL
+
+When writing matchers for standard RSpec constructs, always use the `Language` module's matchers rather than hardcoding method names. The `Language` module reads from the `RSpec/Language` section of the RuboCop configuration, so matchers automatically adapt to user-defined aliases.
+
+`RuboCop::Cop::RSpec::Base` includes `RuboCop::RSpec::Language` automatically. If you write a standalone module that needs these matchers (e.g. a mixin), include `RuboCop::RSpec::Language` explicitly:
```
bquorning:
```suggestion
`RuboCop::Cop::RSpec::Base` includes `RuboCop::RSpec::Language`. But if you write a standalone module that needs these matchers (e.g. a mixin), you should include `RuboCop::RSpec::Language` explicitly:
```

bquorning (approved):
Do we have this documentation also in each helper source file (and should we?)

PR #2193, opened 2026-06-22
Fix incorrect autocorrection for RSpec/ExampleWording with percent literals

ydah:
Fix incorrect autocorrection for `RSpec/ExampleWording` when rewriting example descriptions that use percent literals or escaped string content.

Before this change, the cop calculated the docstring range by trimming one character from each side of `source_range`. For `%q[should do something]`, that included the `%q[` opener in the offense and replacement range. For escaped strings, autocorrection used the decoded string value, which could drop required escapes.

```ruby
it %q[should do something] do
end

it "should include \"quotes\"" do
end
```

After this change, delimited strings use parser delimiter locations for the docstring range, and replacements use the original source between delimiters when available. Existing fallback behavior is kept for heredocs and backslash-concatenated strings.

- [x] Added tests.
- [ ] Updated documentation.

### DIFF
--- a/lib/rubocop/cop/rspec/example_wording.rb
+++ b/lib/rubocop/cop/rspec/example_wording.rb
@@ -106,6 +106,12 @@ module RuboCop
         end
 
         def docstring(node)
+          return source_without_outer_chars(node) unless delimited_string?(node)
+
+          node.loc.begin.end.join(node.loc.end.begin)
+        end
+
+        def source_without_outer_chars(node)
           expr = node.source_range
 
           Parser::Source::Range.new(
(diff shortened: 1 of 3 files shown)


lib/rubocop/cop/rspec/example_wording.rb
```diff
@@ -106,6 +106,16 @@ def add_wording_offense(node, message)
         end
 
         def docstring(node)
+          return source_without_outer_chars(node) unless delimited_string?(node)
+
+          Parser::Source::Range.new(
+            node.source_range.source_buffer,
+            node.loc.begin.end_pos,
+            node.loc.end.begin_pos
+          )
```
koic: For reference, this can be simplified as follows:

```suggestion
          node.loc.begin.end.join(node.loc.end.begin)
```

ydah: Thank you! I fixed it.

bquorning (approved):

PR #2195, opened 2026-06-23
Fix `RSpec/MatchWithSimpleRegex` to ignore match nested inside include matchers

ydah:
Fix `RSpec/MatchWithSimpleRegex` to ignore `match` when it is already nested inside `include` matcher arguments.

Before this change, `include(match(/message/))` was detected and autocorrected to `include(include('message'))`. The nested `include` form is valid matcher composition, but it is not clearer than the original expression and goes against the cop's purpose.

This keeps the direct `expect(body).to match(/message/)` correction, while leaving nested `include(match(...))` unchanged.

- [x] Added tests.
- [ ] Updated documentation.

### DIFF
--- a/spec/rubocop/cop/rspec/match_with_simple_regex_spec.rb
+++ b/spec/rubocop/cop/rspec/match_with_simple_regex_spec.rb
@@ -48,6 +48,14 @@ RSpec.describe RuboCop::Cop::RSpec::MatchWithSimpleRegex, :config do
     RUBY
   end
 
+  it 'does not register an offense when using match within an ' \
+     'include matcher' do
+    expect_no_offenses(<<~RUBY)
+      expect(errors).to include(match(/message/))
+      expect(errors).to include(match(/message/), match(/warning/))
+    RUBY
+  end
+
   it 'does not register an offense when using match with anchor at start' do
     expect_no_offenses(<<~RUBY)
       expect('foobar').to match(/^foo/)
--- a/spec/rubocop/cop/rspec/match_with_simple_regex_spec.rb
+++ b/spec/rubocop/cop/rspec/match_with_simple_regex_spec.rb
@@ -76,6 +76,16 @@ RSpec.describe RuboCop::Cop::RSpec::MatchWithSimpleRegex, :config do
     RUBY
   end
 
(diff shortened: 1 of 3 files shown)


spec/rubocop/cop/rspec/match_with_simple_regex_spec.rb
```diff
@@ -12,6 +12,18 @@
     RUBY
   end
 
+  it 'registers an offense when using match with simple string regex ' \
+     'without a parent node' do
+    expect_offense(<<~RUBY)
+      match(/foo/)
+      ^^^^^^^^^^^^ Prefer using `include('foo')` when the regex is a simple string literal.
```
bquorning: This is wrong behavior, right? We should only match inside examples I think.

It's not something you added by the way. The cop was implemented wrong (by me) from the beginning.

ydah: follow up: https://github.com/rubocop/rubocop-rspec/pull/2196

spec/rubocop/cop/rspec/match_with_simple_regex_spec.rb
```diff
@@ -48,6 +48,14 @@
     RUBY
   end
 
+  it 'does not register an offense when using match within an ' \
+     'include matcher' do
+    expect_no_offenses(<<~RUBY)
+      expect(errors).to include(match(/message/))
+      expect(errors).to include(match(/message/), match(/warning/))
```
bquorning: What is a valid use case for an example like this?

bquorning: 🤦🏼 I see it now. The `errors` collection should contain at least one element matching `/message/`, and at least one element matching `/warning/`.

bquorning:
> Before this change, `include(match(/message/))` was detected and autocorrected to `include(include('message'))`. The nested `include` form is valid matcher composition, but it is not clearer than the original expression and goes against the cop's purpose.

Ah, now I see, and I agree. A clearer replacement would perhaps be:
```rb
include(a_string_including('message'))
```
but I think that’s a separate concern. Let’s get this PR merged.

PR #2196, opened 2026-06-23
Fix `RSpec/MatchWithSimpleRegex` to ignore `match` outside examples

ydah:
Fix `RSpec/MatchWithSimpleRegex` to ignore `match` calls outside RSpec examples.

Before this change, the cop inspected any bare `match(/literal/)` call. That could register an offense outside examples, even though this cop is about RSpec expectation matcher style.

This change limits the cop to nodes inside examples, and updates the existing examples in the spec so they continue to cover behavior inside `it` blocks.

- [x] Added tests.
- [ ] Updated documentation.


bquorning (approved):
Thank you 🙏🏼

PR #2197, opened 2026-06-25
Pin GitHub Actions to commit hashes

ydah:
Pin GitHub Actions workflow dependencies to full commit hashes.

This replaces tag and branch references in workflow `uses:` entries with the commit SHA currently resolved from each upstream action. This makes CI and publish workflows less dependent on mutable refs such as `v1`, `master`, and `main`.

The changed workflows are `main.yml`, `codespell.yml`, `linting.yml`, and `publish.yml`.

- [-] Added tests.
- [-] Updated documentation.

- [-] Added the new cop to `config/default.yml`.
- [-] The cop is configured as `Enabled: pending` in `config/default.yml`.
- [-] The cop is configured as `Enabled: true` in `.rubocop.yml`.
- [-] The cop documents examples of good and bad code.
- [-] The tests assert both that bad code is reported and that good code is not reported.
- [-] Set `VersionAdded: "<<next>>"` in `default/config.yml`.

- [-] Set `VersionChanged: "<<next>>"` in `config/default.yml`.

### DIFF
--- a/.github/workflows/codespell.yml
+++ b/.github/workflows/codespell.yml
@@ -9,9 +9,9 @@ jobs:
     name: CodeSpell
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v7
+      - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0
       - name: CodeSpell
-        uses: codespell-project/actions-codespell@master
+        uses: codespell-project/actions-codespell@3abb875e3aa9713e40eed5aea082672a42f7f95c
         with:
           check_filenames: true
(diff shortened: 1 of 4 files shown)


Darhazer (approved):

bquorning:
But why? (Security, I guess)

It _does_ make it harder to figure out which version we are using, and if there are new versions available.

Darhazer:
> But why? (Security, I guess)
> 
> It _does_ make it harder to figure out which version we are using, and if there are new versions available.

The reason is security indeed. Dependabot bumps hashes as well, so the new versions should not be concern. And where I work we left the exact version used as an inline comment

PR #2209, opened 2026-07-30
[Fix #2145] Fix RSpec/LeadingSubject crash on itblock/numblock groups

pcbeingused333:
Fixes #2145.

`RSpec/LeadingSubject` raised `NoMethodError: undefined method 'body' for nil` when the enclosing example group is parsed as an `itblock`/`numblock` node — i.e. Ruby 3.4's implicit `it` block parameter under the Prism parser. `#parent` used `node.each_ancestor(:block).first.body`, and since `each_ancestor(:block)` doesn't match `itblock`/`numblock` nodes, `.first` returned `nil`.

This widens the ancestor lookup to `:any_block` so it also matches `numblock`/`itblock` groups, resolving the crash. Added a regression spec under a `:ruby34` context.
______________________________________________________________________

Before submitting the PR make sure the following are checked:

- [X] Feature branch is up-to-date with `master` (if not - rebase it).
- [X] Squashed related commits together.
- [X] Added tests.
- [x] Updated documentation.
- [X] Added an entry to the `CHANGELOG.md` if the new code introduces user-observable changes.
- [X] The build (`bundle exec rake`) passes (be sure to run this locally, since it may produce updated documentation that you will need to commit).

If you have created a new cop:

- [ ] Added the new cop to `config/default.yml`.
- [ ] The cop is configured as `Enabled: pending` in `config/default.yml`.
- [ ] The cop is configured as `Enabled: true` in `.rubocop.yml`.
- [ ] The cop documents examples of good and bad code.
- [ ] The tests assert both that bad code is reported and that good code is not reported.
- [ ] Set `VersionAdded: "<<next>>"` in `default/config.yml`.

If you have modified an existing cop's configuration options:

- [ ] Set `VersionChanged: "<<next>>"` in `config/default.yml`.

PR #2213, opened 2026-08-02
Support Ruby 3.4 itblock/numblock across RSpec cops

pcbeingused333:
Follow-up to the spot check discussed in #2209.

Under Ruby 3.4 / Prism, a bare `it` (pending example) or a numbered
parameter makes the enclosing block parse as an `itblock`/`numblock`
instead of a `block`. Several cops assume `:block` and so silently stop
flagging (and in a couple of cases risk a false positive) when an example
or example group takes that form.

This makes block-type handling uniform across five spots:

1. **Traversal** — `each_ancestor(:block)` / `each_descendant(:block)` → `:any_block`.
2. **Node patterns** — `example?`, `example_group_with_body?` (also relaxing its fixed `args` child, which an `itblock`/`numblock` doesn't have), `shared_group?`, and per-cop matchers → `any_block`.
3. **`each_child_node(…, :block, …)`** → `:any_block`.
4. **Callback registration** — `on_block`-only cops → add `on_numblock`/`on_itblock` (following `AroundBlock`/`IteratedExpectation`).
5. **Type dispatch** — `case node.parent.type; when :block` in `DiscardedMatcher#void_value?`.

Adds `:ruby34` regression specs to each affected cop. Full suite is green locally (2466 examples, 0 failures); `bundle exec rake internal_investigation` is clean.

Opening as a **draft**: it touches shared `Language` patterns and several cops, so I'm happy to split/stage it however you prefer (cf. #2209).

- [X] Added tests.
- [ ] Updated documentation.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 - Fix incorrect autocorrection for `RSpec/DescribedClass` when using nested example groups with `EnforcedStyle: explicit`. ([@ydah])
 - Fix `RSpec/MatchWithSimpleRegex` to ignore `match` outside examples. ([@ydah])
 - Fix incorrect autocorrection for `RSpec/ExampleWording` with percent literal and escaped example descriptions. ([@ydah])
+- Fix `RSpec/LeadingSubject` to not crash on `itblock`/`numblock` example groups with Ruby 3.4 and Prism. ([@pcbeingused333])
 
 ## 3.10.2 (2026-06-06)
 
@@ -1095,6 +1096,7 @@ Compatibility release so users can upgrade RuboCop to 0.51.0. No new features.
 [@oshiro3]: https://github.com/oshiro3
 [@patrickomatic]: https://github.com/patrickomatic
 [@paydaylight]: https://github.com/paydaylight
+[@pcbeingused333]: https://github.com/pcbeingused333
 [@philcoggins]: https://github.com/PhilCoggins
 [@pirj]: https://github.com/pirj
 [@pocke]: https://github.com/pocke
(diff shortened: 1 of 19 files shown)


pirj:
Thanks for the PR!

You mention [here](https://github.com/rubocop/rubocop-rspec/pull/2209#issuecomment-5156218874) that two cops can flip to a false positive (InstanceVariable, NamedSubject). There’s no new specs for those two cops. Do we need something to prevent regression? Can you show those cases when this may happen?

pcbeingused333:
Noted — I dug in, and those two shouldn't be in this PR.

The InstanceVariable case (verified):

```ruby
RSpec.describe Foo do
  subject { Class.new { @foo; it } }  # Class.new parses as an itblock
end
```

`@foo` should be exempt (it's inside `Class.new`), but the cop flags it. The catch: my change is **inert** for this — the exemption is gated by `dynamic_class?`/`custom_matcher?`, which are `(block …)` patterns. Widening only the traversal doesn't help; you'd also have to widen those patterns, which is exactly the block-syntax completeness you'd rather not chase.

So instead of pattern-widening + speculative specs for a contrived case, I've reverted `InstanceVariable` and `NamedSubject` — and, on the same principle, the other cops whose change wasn't backed by a regression spec. The PR now only touches cops that have a demonstrated `:ruby34` spec.

pirj:
> a bare it (pending example) or a numbered
parameter

It may happen, but is it realistic that someone would place a bare (no docstring) `it` in specs for whatever reason?

Also, do you think that there is a place for numbered-parameter blocks in specs?

pcbeingused333:
Yes, I think it's best I close this PR 😬

PR #2214, opened 2026-08-03
Fix RSpec/LeadingSubject to not move a subject above another subject

### LINKED ISSUE #1835: `RSpec/MultipleSubjects` and `RSpec/LeadingSubject` autocorrect removes subject used in tests
Given this code:

```rb
describe "Something" do
  let(:something_unrelated) { "i exist" }

  subject(:named_subject) { 1 }
  subject do
    2
  end

  it "should be 2" do
    expect(subject).to eq(2)
  end
end
```

And this reduced example command (assuming a config that requires `rubocop-rspec`):

```
bundle exec rubocop --only RSpec/MultipleSubjects,RSpec/LeadingSubject spec/example_spec.rb -a -d
```

This autocorrects to:

```rb
describe "Something" do
  subject(:named_subject) { 1 }
  let(:something_unrelated) { "i exist" }


  it "should be 2" do
    expect(subject).to eq(2)
  end
end
```

Output:

```
Inspecting 1 file
C

Offenses:

spec/example_spec.rb:2:3: C: [Corrected] RSpec/MultipleSubjects: Do not set more than one subject per example group
  subject do ...
  ^^^^^^^^^^
spec/example_spec.rb:4:3: C: [Corrected] RSpec/LeadingSubject: Declare subject above any other let declarations.
  subject(:named_subject) { 1 }
  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
spec/example_spec.rb:4:3: C: [Corrected] RSpec/MultipleSubjects: Do not set more than one subject per example group
  subject(:named_subject) { 1 }
  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
spec/example_spec.rb:5:3: C: [Corrected] RSpec/LeadingSubject: Declare subject above any other let declarations.
  subject do ...
  ^^^^^^^^^^
```

I'm guessing the sequence of events is something like this:

1. `RSpec/MultipleSubjects` no-ops. Ordinarily, if executed alone it would turn the named `subject` into a `let`.
2. `RSpec/LeadingSubject` moves the named subject to the start of the block.
3. `RSpec/LeadingSubject` moves the unnamed subject to the start of the block:

```rb
describe "Something" do
  subject do
    2
  end
  subject(:named_subject) { 1 }
  let(:something_unrelated) { "i exist" }


  it "should be 2" do
    expect(subject).to eq(2)
  end
end
```

4. `MultipleSubjects` removes the unnamed `subject` because it will be overridden by the declaration of the `named_subject`.

This makes the test fail.

---

This was based on some relatively recent code (albeit a misuse of subjects). I agree with the autocorrect behaviour of `MultipleSubjects` in concept, but I'm not sure what should be done to prevent the conflict of these two cops.

pcbeingused333:
Fixes #1835.

`RSpec/LeadingSubject` moved every subject above preceding `let`s/hooks without regard for other subjects, so with multiple subjects it could swap their relative order. Combined with `RSpec/MultipleSubjects` autocorrect, that removed the subject actually used in the example.

Following the discussion on #1835, `LeadingSubject` now treats another subject as a boundary — a subject still moves above preceding `let`s/hooks, but never above another subject — so the subjects keep their relative order. With the order preserved, `MultipleSubjects` renames only the overridden subject and the example keeps working.

Added a regression spec.

- [X] Added tests.
- [ ] Updated documentation.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Master (Unreleased)
 
+- Fix `RSpec/LeadingSubject` to not move a subject above another subject, avoiding an autocorrect clash with `RSpec/MultipleSubjects` that could remove a subject used in tests. ([@pcbeingused333])
 - Fix false positives for `RSpec/Pending` when using SimpleCov 1.x `skip` filters. ([@gee-forr])
 - Fix false positives for `RSpec/SpecFilePathFormat` when matching spec partials. ([@ydah])
 - Fix incorrect autocorrection for `RSpec/DescribedClass` when using nested example groups with `EnforcedStyle: explicit`. ([@ydah])
@@ -1095,6 +1096,7 @@ Compatibility release so users can upgrade RuboCop to 0.51.0. No new features.
 [@oshiro3]: https://github.com/oshiro3
 [@patrickomatic]: https://github.com/patrickomatic
 [@paydaylight]: https://github.com/paydaylight
+[@pcbeingused333]: https://github.com/pcbeingused333
 [@philcoggins]: https://github.com/PhilCoggins
 [@pirj]: https://github.com/pirj
 [@pocke]: https://github.com/pocke
--- a/lib/rubocop/cop/rspec/leading_subject.rb
+++ b/lib/rubocop/cop/rspec/leading_subject.rb
@@ -47,7 +47,7 @@ module RuboCop
         private
 
         def check_previous_nodes(node)
-          offender = first_offender(node)
+          offender = preceding_offender(node)
           return unless offender
 
           msg = format(MSG, offending: offender.method_name)
@@ -57,21 +57,24 @@ module RuboCop
           end
         end
 
-        def first_offender(node)
+        # The declaration that makes this `subject` non-leading. Used only to
+        # build the offense message and to decide whether to report an offense;
+        # where the `subject` actually moves is decided by `move_target`.
+        def preceding_offender(node)
           preceding_siblings(node).find { |sibling| offending?(sibling) }
         end
 
-        # A subject may move above preceding `let`s/hooks, but never above
-        # another subject. Returns the earliest node to move in front of, or
-        # `nil` when a preceding subject blocks it (a later pass handles it).
+        # A `subject` may move above preceding `let`s/hooks, but never above
+        # another `subject`. Walking preceding siblings in reverse, we stop at
+        # the nearest `subject` and return the topmost offender after it, or
+        # `nil` when a `subject` blocks the move (a later pass moves this one
+        # once the `subject` above it has moved).
         def move_target(node)
           target = nil
-          preceding_siblings(node).each do |sibling|
-            if subject?(sibling)
-              target = nil
-            elsif target.nil? && offending?(sibling)
-              target = sibling
-            end
+          preceding_siblings(node).reverse_each do |sibling|
+            break if subject?(sibling)
+
+            target = sibling if offending?(sibling)
           end
           target
         end
--- a/lib/rubocop/cop/rspec/leading_subject.rb
+++ b/lib/rubocop/cop/rspec/leading_subject.rb
@@ -47,26 +47,39 @@ module RuboCop
         private
 
         def check_previous_nodes(node)
-          offending_node(node) do |offender|
-            msg = format(MSG, offending: offender.method_name)
-            add_offense(node, message: msg) do |corrector|
-              autocorrect(corrector, node, offender)
-            end
+          offender = first_offender(node)
+          return unless offender
+
+          msg = format(MSG, offending: offender.method_name)
+          add_offense(node, message: msg) do |corrector|
+            target = move_target(node)
+            autocorrect(corrector, node, target) if target
           end
         end
 
-        def offending_node(node)
-          offender = nil
-          parent(node).each_child_node do |sibling|
-            break if sibling.equal?(node)
+        def first_offender(node)
+          preceding_siblings(node).find { |sibling| offending?(sibling) }
+        end
 
+        # A subject may move above preceding `let`s/hooks, but never above
+        # another subject. Returns the earliest node to move in front of, or
+        # `nil` when a preceding subject blocks it (a later pass handles it).
+        def move_target(node)
+          target = nil
+          preceding_siblings(node).each do |sibling|
             if subject?(sibling)
-              offender = nil
-            elsif offender.nil? && offending?(sibling)
-              offender = sibling
+              target = nil
+            elsif target.nil? && offending?(sibling)
+              target = sibling
             end
           end
-          yield offender if offender
+          target
+        end
+
+        def preceding_siblings(node)
+          parent(node).each_child_node.take_while do |sibling|
+            !sibling.equal?(node)
+          end
         end
 
         def parent(node)
--- a/lib/rubocop/cop/rspec/leading_subject.rb
+++ b/lib/rubocop/cop/rspec/leading_subject.rb
@@ -56,11 +56,17 @@ module RuboCop
         end
 
         def offending_node(node)
-          parent(node).each_child_node.find do |sibling|
+          offender = nil
+          parent(node).each_child_node do |sibling|
             break if sibling.equal?(node)
 
-            yield sibling if offending?(sibling)
+            if subject?(sibling)
+              offender = nil
+            elsif offender.nil? && offending?(sibling)
+              offender = sibling
+            end
           end
+          yield offender if offender
         end
 
         def parent(node)
--- a/spec/rubocop/cop/rspec/leading_subject_spec.rb
+++ b/spec/rubocop/cop/rspec/leading_subject_spec.rb
@@ -294,10 +294,10 @@ RSpec.describe RuboCop::Cop::RSpec::LeadingSubject do
     expect_offense(<<~RUBY)
       RSpec.describe User do
         let(:foo) { bar }
-
         subject(:a) { first }
         ^^^^^^^^^^^^^^^^^^^^^ Declare `subject` above any other `let` declarations.
         subject(:b) { second }
+        ^^^^^^^^^^^^^^^^^^^^^^ Declare `subject` above any other `let` declarations.
       end
     RUBY
 
@@ -306,7 +306,6 @@ RSpec.describe RuboCop::Cop::RSpec::LeadingSubject do
         subject(:a) { first }
         subject(:b) { second }
         let(:foo) { bar }
-
       end
     RUBY
   end
--- a/spec/rubocop/cop/rspec/leading_subject_spec.rb
+++ b/spec/rubocop/cop/rspec/leading_subject_spec.rb
@@ -289,4 +289,25 @@ RSpec.describe RuboCop::Cop::RSpec::LeadingSubject do
       end
     RUBY
   end
+
+  it 'does not move a subject above another subject' do
+    expect_offense(<<~RUBY)
+      RSpec.describe User do
+        let(:foo) { bar }
+
+        subject(:a) { first }
+        ^^^^^^^^^^^^^^^^^^^^^ Declare `subject` above any other `let` declarations.
+        subject(:b) { second }
+      end
+    RUBY
+
+    expect_correction(<<~RUBY)
+      RSpec.describe User do
+        subject(:a) { first }
+        subject(:b) { second }
+        let(:foo) { bar }
+
+      end
+    RUBY
+  end
 end
--- a/spec/rubocop/cop/rspec/leading_subject_spec.rb
+++ b/spec/rubocop/cop/rspec/leading_subject_spec.rb
@@ -309,4 +309,37 @@ RSpec.describe RuboCop::Cop::RSpec::LeadingSubject do
       end
     RUBY
   end
+
+  it 'reports but does not move a subject blocked by a preceding subject' do
+    # `subject(:b)` is offending (a `before` precedes it), but on the first
+    # pass `move_target` is `nil` because `subject(:a)` blocks the move. Once
+    # `subject(:a)` has moved above the hook, a later pass moves `subject(:b)`.
+    expect_offense(<<~RUBY)
+      RSpec.describe User do
+        before { prepare }
+        subject(:a) { first }
+        ^^^^^^^^^^^^^^^^^^^^^ Declare `subject` above any other `before` declarations.
+        subject(:b) { second }
+        ^^^^^^^^^^^^^^^^^^^^^^ Declare `subject` above any other `before` declarations.
+      end
+    RUBY
+
+    expect_correction(<<~RUBY)
+      RSpec.describe User do
+        subject(:a) { first }
+        subject(:b) { second }
+        before { prepare }
+      end
+    RUBY
+  end
+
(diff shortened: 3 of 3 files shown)


spec/rubocop/cop/rspec/leading_subject_spec.rb
```diff
@@ -289,4 +289,25 @@ module Support
...
   end
+
+  it 'does not move a subject above another subject' do
+    expect_offense(<<~RUBY)
+      RSpec.describe User do
+        let(:foo) { bar }
+
+        subject(:a) { first }
+        ^^^^^^^^^^^^^^^^^^^^^ Declare `subject` above any other `let` declarations.
+        subject(:b) { second }
```
pirj, line 299: Those are both offenders.  It’s just the order that should remain stable.

Do you think it would work practically to only fix autocorrection?

I believe the real runs are more complicated, as the corrected file is passed through other cops after correction, and this cuases multiple passes.

pcbeingused333: For the autocorrect: both subjects are still flagged, I only changed *where* a subject gets moved. Now it moves above preceding `let`s/hooks but stops at another subject — so it never moves one subject above another. If a subject is blocking, it's just not moved on that pass; on the next pass the order is already settled, so it moves then and stays stable.

lib/rubocop/cop/rspec/leading_subject.rb
```diff
@@ -47,19 +47,38 @@ def on_block(node) # rubocop:disable InternalAffairs/NumblockHandler, InternalAf
...
-          parent(node).each_child_node.find do |sibling|
-            break if sibling.equal?(node)
+        def first_offender(node)
+          preceding_siblings(node).find { |sibling| offending?(sibling) }
+        end
+
+        # A subject may move above preceding `let`s/hooks, but never above
+        # another subject. Returns the earliest node to move in front of, or
+        # `nil` when a preceding subject blocks it (a later pass handles it).
+        def move_target(node)
```
pirj, line 72: Would it work if this used `find` on preceeding siblings in reverse order, and would break on first occurrence of `subject`.

I imagine this to be simpler, and slightly more straightforward.

lib/rubocop/cop/rspec/leading_subject.rb
```diff
@@ -47,19 +47,38 @@ def on_block(node) # rubocop:disable InternalAffairs/NumblockHandler, InternalAf
...
-            add_offense(node, message: msg) do |corrector|
-              autocorrect(corrector, node, offender)
-            end
+          offender = first_offender(node)
+          return unless offender
+
+          msg = format(MSG, offending: offender.method_name)
+          add_offense(node, message: msg) do |corrector|
+            target = move_target(node)
+            autocorrect(corrector, node, target) if target
```
pirj, line 56: Can you please add a test to cover the nil target?

lib/rubocop/cop/rspec/leading_subject.rb
```diff
@@ -47,19 +47,38 @@ def on_block(node) # rubocop:disable InternalAffairs/NumblockHandler, InternalAf
         private
 
         def check_previous_nodes(node)
-          offending_node(node) do |offender|
-            msg = format(MSG, offending: offender.method_name)
-            add_offense(node, message: msg) do |corrector|
-              autocorrect(corrector, node, offender)
-            end
+          offender = first_offender(node)
```
pirj: What used to be a good approach doesn’t work with the new one.
Prevoiusly, we scanned all blocks, and if it was a subject definition, we checked its enclosing parent, and checked all subjects that were not at the top of the enclosing block and moved them up.

With finding just the first one, I believe the methods should be at least named accordingly. It is a first offender, but semantically it’s the only one, because we intend to skip on the other ones, and deal with them in another pass.

lib/rubocop/cop/rspec/leading_subject.rb
```diff
@@ -47,19 +47,38 @@ def on_block(node) # rubocop:disable InternalAffairs/NumblockHandler, InternalAf
...
+          msg = format(MSG, offending: offender.method_name)
+          add_offense(node, message: msg) do |corrector|
+            target = move_target(node)
+            autocorrect(corrector, node, target) if target
           end
         end
 
-        def offending_node(node)
-          parent(node).each_child_node.find do |sibling|
-            break if sibling.equal?(node)
```
pirj, line 60: Frankly, looking at this, I now have doubts that it used to work the way I imagined it.

If it was scanning for all previous subjects only, it means it used to deal with one at once, not many.

lib/rubocop/cop/rspec/leading_subject.rb
pirj: Alternatively, we can do this all in one pass, but the problem would be the reversing that happens due to inserting at the top of the previously corrected ones.

My take: change the iteration to autocorrect in reverse order.

lib/rubocop/cop/rspec/leading_subject.rb
pcbeingused333: Switched to `reverse_each` with a `break` at the first `subject` (`6729f785`). A plain `find` can't quite express it (we need to stop at the boundary *and* keep the topmost offender before it), but the reverse walk reads much cleaner than the reset-to-nil version.

lib/rubocop/cop/rspec/leading_subject.rb
pcbeingused333: `'Reports but does not move a subject blocked by a preceding subject'`. There `subject(:b)` is reported but `move_target` returns `nil` on the first pass (blocked by `subject(:a)`), so the `if target` guard is covered. Worth noting the `nil` target is transient — it only shows up mid-convergence, since the blocking subject always moves up on an earlier pass.

lib/rubocop/cop/rspec/leading_subject.rb
pcbeingused333: Right. The old `offending_node` used `find` with a block that `yield`ed on *every* offending sibling (`add_offense` returns falsy, so `find` never short-circuited), so across passes it hoisted the subject above all preceding lets/hooks — ignoring other subjects as boundaries. That's exactly the bug: with two subjects it lifted both above the lets and could flip their order, and `MultipleSubjects` then dropped the one still in use. The boundary rule fixes that.

lib/rubocop/cop/rspec/leading_subject.rb
pcbeingused333: Renamed to `preceding_offender` with a doc comment (`6729f785`). I kept it separate from `move_target` on purpose: we still need to report a subject even when its `move_target` is `nil` (blocked by a preceding subject).

I stayed with the multi-pass approach — RuboCop already re-runs to convergence, and it keeps the correction free of manual bookkeeping of shifting insertion points as siblings move. The subject-as-boundary rule preserves relative order across passes. Please tell me if you want to switch to single-pass reverse-order version if you'd prefer one pass, just say the word.

pcbeingused333:
Heads-up on CI here, in case it looks like this branch was never verified: the workflow runs for my last two commits — `26d0e370` and `6729f785`, the ones addressing the review — have been sitting in `action_required` since Aug 4 and never started. That's the fork-contributor gate, not a failure. The first push (`62d5b41e`) ran green before the follow-ups landed, so the checks currently shown belong to an older state of the branch.

Would one of you mind clicking *Approve and run workflows* when you get a chance? Locally on `6729f785`: `rake spec` 2461 examples, 0 failures, 100% line and branch coverage, `rake internal_investigation` clean across 290 files. Happy to rebase or push again if that is easier than approving.

bquorning:
Hi @pcbeingused333, could you rebase once more, please?

pcbeingused333:
Rebased onto `master`, and squashed the three commits into one while I was in there, since the PR template asks for it.

The one conflict was against #2209, which landed in this same spec file earlier today — resolved by keeping both, with the new specs ahead of the `:ruby34` context. The changelog entry needed no contributor link this time, that PR already added it.

Locally on `289e1cd4`: `rake spec` 2471 examples / 0 failures, line and branch coverage both 100%, and `internal_investigation`, `confirm_config` and `confirm_documentation` all clean with nothing regenerated. Checks are running now — thanks for approving the workflows on the previous push, that is what let me see it was green before rebasing.

pcbeingused333:
Gentle nudge on this one. The rebase and squash landed on `289e1cd4` a week ago and the 34 checks have been green since; `MERGEABLE` / `CLEAN`, no drift against master. Happy to rebase again if it's gone stale — otherwise it's ready whenever you have a moment.

lib/rubocop/cop/rspec/leading_subject.rb
```diff
@@ -47,19 +47,41 @@ def on_block(node) # rubocop:disable InternalAffairs/NumblockHandler, InternalAf
...
+        # The declaration that makes this `subject` non-leading. Used only to
+        # build the offense message and to decide whether to report an offense;
+        # where the `subject` actually moves is decided by `move_target`.
+        def preceding_offender(node)
+          preceding_siblings(node).find { |sibling| offending?(sibling) }
+        end
+
+        # A `subject` may move above preceding `let`s/hooks, but never above
```
pirj: specs should explain this in runnable way

lib/rubocop/cop/rspec/leading_subject.rb
```diff
@@ -47,19 +47,41 @@ def on_block(node) # rubocop:disable InternalAffairs/NumblockHandler, InternalAf
...
+          add_offense(node, message: msg) do |corrector|
+            target = move_target(node)
+            autocorrect(corrector, node, target) if target
           end
         end
 
-        def offending_node(node)
-          parent(node).each_child_node.find do |sibling|
-            break if sibling.equal?(node)
+        # The declaration that makes this `subject` non-leading. Used only to
```
pirj: The method name preceding_offender says it all

spec/rubocop/cop/rspec/leading_subject_spec.rb
```diff
@@ -290,6 +290,59 @@ module Support
...
+        subject(:a) { first }
+        subject(:b) { second }
+        let(:foo) { bar }
+      end
+    RUBY
+  end
+
+  it 'reports but does not move a subject blocked by a preceding subject' do
+    # `subject(:b)` is offending (a `before` precedes it), but on the first
+    # pass `move_target` is `nil` because `subject(:a)` blocks the move. Once
```
pirj: Why don’t subject(:a) move up first? B could move next. I think that’s the reason we’ve changed the order to reverse

lib/rubocop/cop/rspec/leading_subject.rb
pcbeingused333: Removed the comment. The behaviour is in the specs now: does not move a subject above another subject, moves several subjects above a preceding hook, keeping their order (three subjects, so the order preservation is visible), and keeps the order of subjects separated by a let.

lib/rubocop/cop/rspec/leading_subject.rb
pcbeingused333: Removed.

spec/rubocop/cop/rspec/leading_subject_spec.rb
pcbeingused333: It does — subject(:a) moves on the first pass and subject(:b) on the next; the expect_correction block is the converged result, both above the hook in their original order. The old name described only the first pass, which was the confusing part, so I renamed it.

The reverse walk in move_target is what makes subject(:a) move first: it breaks at the first preceding subject, so subject(:b) is boundaried by subject(:a) and gets nil until subject(:a) has cleared the slot ahead of it. Moving both in one pass runs into the insert-at-top reversal you flagged on Aug 3 — two subjects moving above the same anchor in a single corrector pass clash. The boundary rule plus RuboCop's convergence loop avoids that: one subject clears the anchor per pass, order preserved. Checked with before; subject(:a); subject(:b); subject(:c) → all three settle above the hook in order.

pcbeingused333:
Rebased onto master (past #2227) while addressing the review.

The Mdformat check is red, but it's failing in the action's own Docker build, not on content: ydah/mdformat-action's Dockerfile runs apt-get upgrade -y on python:3.10-bullseye, and that now 404s on a batch of packages (libperl5.32…deb11u5, libc-dev-bin…deb11u14, linux-libc-dev_5.10.262-1, …) after Debian's latest bullseye point release pruned the old versions from the mirror. It dies before inspecting any file. CHANGELOG.md passes mdformat --check locally and the other 33 checks are green. A re-run once the mirror settles should clear it — I've stopped pushing so as not to burn CI on something the PR can't fix.

bquorning:
> The Mdformat check is red, but it's failing in the action's own Docker build

I opened a PR to fix it: https://github.com/ydah/mdformat-action/pull/15 – cc @ydah

lib/rubocop/cop/rspec/leading_subject.rb
pirj: Is there a test that will fail if we remove ‘if target’? Short yes or no.

Is it a dedicated test? Or would other tests fail due to this side effect?

Short yes or no, no explanation. One sentence.

CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 
 - Fix an error for `RSpec/ImplicitExpect` when the runner is separated from `is_expected` by whitespace. ([@viralpraxis])
 - Fix false positives for `RSpec/InstanceVariable` when an instance variable is used inside a `class_eval` or `module_eval` block. ([@corsonknowles])
+- Fix `RSpec/LeadingSubject` to not move a subject above another subject, avoiding an autocorrect clash with `RSpec/MultipleSubjects` that could remove a subject used in tests. ([@pcbeingused333])
```
pirj, line 7: “Keep a stable subject order”?

pirj (approved):
I’m out of quick ideas how to simplify the code. Maybe you could spend a bit of resources to dig into this? I’d love to see this to be a bit more expressive.
Otherwise, looks good enough, thanks!

pcbeingused333:
@pirj

Yes. Not dedicated — removing `if target` crashes two existing behavior tests (`does not move a subject above another subject`, `moves several subjects above a preceding hook, keeping their order`) with `NoMethodError: undefined method 'loc' for nil`; the nil branch is a side effect of asserting the final subject order, not something tested in isolation. Verified by pulling the guard locally and running the specs.

On the CHANGELOG line: leaving the current wording. It names the actual failure mode — an autocorrect clash with `RSpec/MultipleSubjects` that could silently drop a subject used in tests — which is what matters to someone deciding whether to run `-A` before reading the diff. Can shorten to your version if you'd still rather have it.

On simplifying further: nothing that keeps `move_target`'s single backward pass. The two-phase result (`subject(:b)` landing correctly) comes from RuboCop re-investigating after `subject(:a)`'s correction is applied, not from anything the method itself tracks — encoding that directly would mean duplicating state the framework already gives for free across passes. Leaving it as is unless you see a concrete cut.

Thanks for the review.

pcbeingused333:
@bquorning your ydah/mdformat-action#15 was merged (2026-09-08), so a re-run of the Mdformat job here should now pass — it was failing in the action's Docker build, not on `CHANGELOG.md`. The other 33 checks are green and the branch is up to date with master.

bquorning (approved):

PR #2203, opened 2026-07-17
Lazy loading of cops and mixins

bquorning:
Now that https://github.com/rubocop/rubocop/issues/14983 is completed, I thought it would be fun to change also rubocop-rspec to lazy-load its cops.

The feature requires RuboCop v1.89.0.

______________________________________________________________________

Before submitting the PR make sure the following are checked:

- [x] Feature branch is up-to-date with `master` (if not - rebase it).
- [x] Squashed related commits together.
- [ ] Added tests.
- [ ] Updated documentation.
- [x] Added an entry to the `CHANGELOG.md` if the new code introduces user-observable changes.
- [x] The build (`bundle exec rake`) passes (be sure to run this locally, since it may produce updated documentation that you will need to commit).

PR #2215, opened 2026-08-04
Load cops lazily

koic:
Follow-up to https://github.com/rubocop/rubocop/pull/15436.

Replace the 115 cop requires in `lib/rubocop/cop/rspec_cops.rb` with an `RSpec` department module (`lib/rubocop/cop/rspec.rb`) that registers every cop with the global registry through `RuboCop::Cop::LazyLoader#register_cop` without loading the cop files. A cop's file is loaded only when the cop class is needed, which for an inspection run means only the enabled cops. This follows the same mechanism RuboCop core adopted in rubocop/rubocop#15436 and requires RuboCop 1.89.0+.

The `register_cop` directives keep the former require order, because the registration order defines the cop execution order. The mixins, `RuboCop::Cop::RSpec::Base`, `Language`, and the other shared infrastructure stay eagerly required from the gem entry point ahead of the department module, as before, so a lazily loaded cop finds its superclass and mixins already defined. The load-time augmentation of `Layout/ExtraSpacing` keeps working as is; its references to the two `AlignLet*Brace` cops resolve at call time through the autoload.

`lib/rubocop/cop/rspec_cops.rb` is kept as a deprecated compatibility shim for code that requires it directly. The `new_cop` rake task switches from `inject_require` to `inject_registration`, and the development documentation is updated accordingly.

A new `spec/project/lazy_loading_spec.rb` requires every cop file in `lib/rubocop/cop/rspec` to be registered exactly once and asserts the lazy behavior itself in subprocesses: no cop files loaded after `require 'rubocop-rspec'`, and no double registration when a cop file is required directly.

- [x] Added tests.
- [x] Updated documentation.


bquorning:
This looks like a duplicate of #2203, although there are slight differences.

lib/rubocop/cop/rspec.rb
```diff
@@ -0,0 +1,127 @@
+# frozen_string_literal: true
+
+module RuboCop
+  module Cop
+    # Cops for the `RSpec` department. The department's cops are
+    # registered for lazy loading and their files are loaded on demand.
+    module RSpec
```
bquorning:
```suggestion
    module RSpec # rubocop:disable Metrics/ModuleLength
```

.rubocop.yml
```diff
@@ -64,6 +65,11 @@ Metrics/BlockLength:
 Metrics/MethodLength:
   Max: 15
 
+Metrics/ModuleLength:
+  Exclude:
+    # The department module is a generated list of `register_cop` directives.
+    - lib/rubocop/cop/rspec.rb
+
```
bquorning: I think I’d prefer inline disable instead.
```suggestion

```

bquorning (commented):
What’s your opinion on lazy loading the mixins as well, as suggested in #2203?

koic:
> What’s your opinion on lazy loading the mixins as well, as suggested in https://github.com/rubocop/rubocop-rspec/pull/2203?

Looks good to me. I've updated this PR.

bquorning (approved):

pirj (approved):
II’m wondering how does RedundantDisable cop works now when inline disabled cops are not loaded. Or does it run inspections of the code to do its job?

Otherwise looks good. Thank you!

bquorning:
> I’m wondering how does RedundantDisable cop works now when inline disabled cops are not loaded. Or does it run inspections of the code to do its job?

As far as I can tell, `RedundantCopDisableDirective` compares against the registered cops, not the actually loaded cops – nor only _enabled_ cops. So I think it’ll continue working as before.

ydah (approved):
Thank you!

Issue #830, opened 2019-10-07
Bug: false positive of RSpec/LetSetup with context and outer shared_examples


dmytro-savochkin:
This issue produces false positives of Rspec/LetSetup with the latest rubocop-rspec (1.36.0) when we put `let!` inside some context and we also have `shared_examples` referencing that `let!` outside the context.

Something like this:
```
shared_examples 'some shared example' do
  it 'checks something' do
    is_expected.to eq user
  end
end

context 'when inside some context' do
  let!(:user) { create :user }

  include_examples 'some shared example'
end
```

If we are to run rubocop-rspec on the code above it would produce the following:
```
spec/test_spec.rb:8:3: C: RSpec/LetSetup: Do not use let! to setup objects not referenced in tests.
  let!(:user) { create :user }
  ^^^^^^^^^^^

1 file inspected, 1 offense detected
```

Darhazer:
This is a limitation of the static analysis. The shared examples could be defined in a separate file as well, and Rubocop does not cross-reference files. I would suggest to explicitly pass `user` in this well, like
```
include_examples 'some shared example', user
```
and
```
shared_examples 'some shared example' do |user|
```

This way you will not have a hidden dependency in the code.

typhoon2099:
I have just tried the proposed solution and it causes failures in my tests, as the `user` created by FactoryBot is reused between tests (since it is created outside of a test context). You also cannot declare `user` in a let above the include_examples and use it as a parameter for the same reason.

pirj:
How about
```ruby
  include_examples 'some shared example' do
    let!(:user) { create :user }
  end
```

typhoon2099:
That works, but it still raises a RSpec/LetSetup warning.

Darhazer:
I guess the cop should really ignore `let!` inside `include_examples` blocks, as it can't determine whether the let is used in the shared examples or not

pirj:
Agree. Another exception would be an example group that is including examples, since it's impossible to determine if it's referenced or not.

StuartHadfield:
I'm fairly certain this isn't limited to shared examples.

If you have a top-level subject and create an object inside a context with a `let!`, you will get the same false positive.

I.e.

```ruby
subject(:foo) do
  post "url",
    params: {
      id: my_object.id
    }
end

context "when calling the method" do
  let!(:my_object) { FactoryBot.create(object) } # you will get a false positive here
  
  it { .. }
end
```

[...]

PR #2218, opened 2026-08-06
Update local RuboCop configuration, start using Rubydex

bquorning:
https://metaredux.com/posts/2026/08/05/rubocop-1-89.html tells how RuboCop can now use Rubydex to have knowledge of more than one file at a time during inspection, which allows for some new cop features. They require RuboCop v1.89 or newer, which was recently set in gemspec when switching to lazy loading of cops.

The first commit updates our local RuboCop configuration in .rubocop.yml to enable all pending cops up to the latest version of RuboCop and RuboCop Performance.

- [ ] Added tests.
- [ ] Updated documentation.


pirj:
> have knowledge of more than one file at a time during inspection

I vaguely recall we had a case to make use of this feature. @bquorning do you happen to remember the details?

@copilot please find a supporting discussion in prs.

bquorning:
Could it be https://github.com/rubocop/rubocop-rspec/issues/830#issuecomment-539638311 @pirj ?

pirj:
That’s an impressive dig up! Must be it.

I’m sceptical if the shared examples can be defined in a gem and won’t be loaded by Rubydex. Might be possible with some configuration and by vendoring gems, but won’t work out of the box.

No objections on merging this in, though. Might come handy.

pirj (approved):
If we happen to use Rubydex/Cypher in some of our cops, would our users have to enable `UseProjectIndex` in their configs? Just something to keep in mind when we introduce such a cop.

bquorning:
This PR is only about how we lint this repository, there are no new features or configuration for our users.

If we later want to add a rubydex-powered cop, I'm actually unsure if we need to change any logic, or if our users optin in using the AllCops -> UseProjectIndex config is enough.

bquorning:
See #2219 where I try using Rubydex in a cop.

PR #2219, opened 2026-08-07
Use project index to fix RSpec/LetSetup shared-code false positives

### LINKED ISSUE #830: Bug: false positive of RSpec/LetSetup with context and outer shared_examples

bquorning:
`RSpec/LetSetup` has long flagged a `let!` as unused when its only reference lives in a `shared_examples`/`shared_context` block defined in another file, since single-file analysis can't see the reference (#830).

RuboCop 1.89's Rubydex-backed project index resolves references across files. For a `let!` connected to shared code, the cop now consults the project-wide index and skips the offense if the name is referenced anywhere in the project. The index is consulted only for these shared-code cases, not for every `let!`, since common `let` names recur across spec files and a blanket check would suppress nearly every offense. Without `AllCops/UseProjectIndex` enabled, behavior is unchanged.

Fixes #830.

- [x] Added tests.
- [x] Updated documentation.

### DIFF
--- a/lib/rubocop/cop/rspec/let_setup.rb
+++ b/lib/rubocop/cop/rspec/let_setup.rb
@@ -5,6 +5,17 @@ module RuboCop
     module RSpec
       # Checks unreferenced `let!` calls being used for test setup.
       #
+      # A `let!` that is defined in a `shared_context`/`shared_examples` block,
+      # or in an example group that includes shared examples, may be referenced
+      # from the shared code rather than from the example group itself. Whether
+      # such a reference is present cannot be determined from a single file,
+      # because the shared code is often defined elsewhere. When
+      # `AllCops/UseProjectIndex` is enabled and the `rubydex` gem is installed,
+      # the cop consults the project-wide index for these `let!` calls and does
+      # not flag one whose name is referenced anywhere in the project. Without
+      # the index it falls back to the file-local check, which may report such a
+      # `let!` as unused.
+      #
       # @example
       #   # bad
       #   let!(:my_widget) { create(:widget) }
@@ -38,6 +49,8 @@ module RuboCop
       #   end
       #
       class LetSetup < Base
+        include SharedCodeReference
+
         MSG = 'Do not use `let!` to setup objects not referenced in tests.'
 
         # @!method example_or_shared_group_or_including?(node)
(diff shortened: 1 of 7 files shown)


bquorning:
@pirj I let Claude try fixing the issue we talked about in #2218. Does this look like the right direction?

config/default.yml
```diff
@@ -636,6 +636,7 @@ RSpec/LetSetup:
   Description: Checks unreferenced `let!` calls being used for test setup.
   Enabled: true
   VersionAdded: '1.7'
+  VersionChanged: "<<next>>"
```
pirj: Do we need this? I was under the impression this is only when we add/change config options.

bquorning: You’re right, we don’t need that.

lib/rubocop/cop/rspec/let_setup.rb
```diff
@@ -72,11 +106,41 @@ def on_block(node) # rubocop:disable InternalAffairs/NumblockHandler, InternalAf
...
+              includes_shared_code?(ancestor)
           end
         end
 
+        def project_index_reference_names
+          self.class.reference_names_cache[project_index] ||=
+            project_index.method_references.to_set do |reference|
+              reference.name.delete_suffix('()')
+            end
+        end
```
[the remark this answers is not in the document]
bquorning: Fixed

pirj (commented):
It's my fourth attempt at reviewing this change. And even though I really like the idea, and the implementation seems to cover edge cases.

Still, I can't shake off the feeling that the added complexity doesn't pay off the added value.
What is your own impression of building this?

Would it detect shared contexts included using metadata matching as defined in spec_helper? How would it know which spec_helper is used if there are several (spec_helper, rails_helper, custom_cop_helper etc)?

PR #2221, opened 2026-08-20
Fix false positives for subject detection nested in another block

corsonknowles:
`RSpec/MultipleSubjects` reports a same-named DSL method as a duplicate subject, and its autocorrect then rewrites the DSL call into something that raises:

```ruby
let(:definition) do
  AbilityDefinition.define do
    subject(Model, test: ['value']) { can :read }
    subject(Model, other: ['value']) { can :write }
  end
end
```

```diff
-      subject(Model, test: ['value']) { can :read }
+      let(Model, test: ['value']) { can :read }
```

On a real codebase that gave `NoMethodError`, 3 failures out of 7 examples.

## Rewritten to @pirj's suggestion

> What if the argument is a method call/var?
>
> what would be the consequences if we checked that subject is directly in the example group, not wrapped in another block instead?

Both points were right, so this now takes the second approach.

My first attempt constrained the argument shape to zero-or-one symbol/string. That regresses legitimate dynamic names, which I confirmed:

| snippet | first attempt | this version |
| --- | --- | --- |
| `subject { 1 }` | detected | detected |
| `subject(:name) { 1 }` | detected | detected |
| `subject(name) { 1 }` | **missed** | detected |
| `subject(some_method) { 1 }` | **missed** | detected |
| `subject(SOME_CONST) { 1 }` | **missed** | detected |

`ExampleGroup#subjects` now stops descending into blocks that are not example groups, so a `subject` declared inside some other block is not counted. No argument-shape guessing.

This also matches how the cop already treats declarations it cannot rely on — compare `ScatteredSetup`'s `reject(&:inside_class_method?)`. A `subject` inside an arbitrary block is not a declaration the example group makes.

## Consequences

The one behaviour change beyond the fix: subjects declared inside an iterator are no longer counted, so

```ruby
%i[a b].each { |name| subject(name) { 1 } }
```

reports nothing. That is pinned by a spec so it is a decision rather than an accident. If you would rather keep detecting that, I think it needs an explicit allowance for iterator blocks, which felt like more machinery than the case warrants — happy to add it if you disagree.

## Specs

Six cases: the DSL shape; the DSL shape alongside a real group-level subject (still counted once); and dynamic names by local variable, method call and constant — the three my first attempt broke and which no existing spec covered, which is why its suite went green.

`bundle exec rake` — 2469 examples, 0 failures, 100% line and branch coverage, 293 files inspected, no offenses.

Related: #2222 applies the same "not unconditionally declared" reasoning to `ScatteredSetup`.

### DIFF
--- a/lib/rubocop/rspec/example_group.rb
+++ b/lib/rubocop/rspec/example_group.rb
@@ -22,7 +22,7 @@ module RuboCop
       end
 
       def subjects
-        find_all_in_scope(node, :subject?)
+        find_all_in_scope(node, :subject?, skip_nested_blocks: true)
       end
 
       def examples
@@ -47,20 +47,26 @@ module RuboCop
       # @param predicate [Symbol] method to call with node as argument
       #
       # @return [Array<RuboCop::AST::Node>] discovered nodes
-      def find_all_in_scope(node, predicate)
+      def find_all_in_scope(node, predicate, skip_nested_blocks: false)
         node.each_child_node.flat_map do |child|
-          find_all(child, predicate)
+          find_all(child, predicate, skip_nested_blocks: skip_nested_blocks)
         end
       end
 
-      def find_all(node, predicate)
-        if public_send(predicate, node)
-          [node]
-        elsif scope_change?(node) || example?(node)
-          []
-        else
-          find_all_in_scope(node, predicate)
-        end
+      def find_all(node, predicate, skip_nested_blocks: false)
+        return [node] if public_send(predicate, node)
+        # No attempt is made to tell an RSpec DSL block from any other: a
+        # declaration inside *any* block is not one the group makes
+        # unconditionally, so stop rather than identify the block.
+        return [] if skip_nested_blocks && node.block_type?
+        return [] if scope_change?(node)
+        return [] if example?(node)
+
+        find_all_in_scope(
+          node,
+          predicate,
+          skip_nested_blocks: skip_nested_blocks
+        )
       end
     end
   end
--- a/spec/rubocop/cop/rspec/multiple_subjects_spec.rb
+++ b/spec/rubocop/cop/rspec/multiple_subjects_spec.rb
@@ -80,4 +80,101 @@ RSpec.describe RuboCop::Cop::RSpec::MultipleSubjects do
       end
     RUBY
   end
+
+  it 'registers an offense for subjects named by a local variable' do
+    expect_offense(<<~RUBY)
+      describe 'hello there' do
+        name = :dynamic
+        subject(name) { 1 }
+        ^^^^^^^^^^^^^^^^^^^ Do not set more than one subject per example group
+        subject(:other) { 2 }
+      end
+    RUBY
+  end
+
+  it 'registers an offense for subjects named by a method call' do
+    expect_offense(<<~RUBY)
+      describe 'hello there' do
+        subject(subject_name) { 1 }
+        ^^^^^^^^^^^^^^^^^^^^^^^^^^^ Do not set more than one subject per example group
+        subject(:other) { 2 }
+      end
+    RUBY
+  end
+
+  it 'registers an offense for subjects named by a constant' do
+    expect_offense(<<~RUBY)
+      describe 'hello there' do
+        subject(SUBJECT_NAME) { 1 }
+        ^^^^^^^^^^^^^^^^^^^^^^^^^^^ Do not set more than one subject per example group
+        subject(:other) { 2 }
+      end
+    RUBY
+  end
+
+  it 'ignores a same-named method called inside another block' do
+    expect_no_offenses(<<~RUBY)
+      describe 'hello there' do
+        let(:definition) do
+          AbilityDefinition.define do
+            subject(Model, test: ['value']) { can :read }
+            subject(Model, other: ['value']) { can :write }
+          end
+        end
+      end
+    RUBY
+  end
+
+  it 'counts the group own subject beside one nested in a block' do
+    expect_no_offenses(<<~RUBY)
+      describe 'hello there' do
+        let(:definition) do
+          AbilityDefinition.define do
+            subject(Model, test: ['value']) { can :read }
+          end
+        end
+
+        subject(:real_one) { described_class.new }
+      end
+    RUBY
+  end
+
+  it 'autocorrects the group own subjects, not a nested DSL call' do
+    expect_offense(<<~RUBY)
+      describe 'hello there' do
+        let(:definition) do
+          AbilityDefinition.define do
+            subject(Model, test: ['value']) { can :read }
+          end
+        end
+
+        subject(:first) { 1 }
+        ^^^^^^^^^^^^^^^^^^^^^ Do not set more than one subject per example group
+        subject(:second) { 2 }
+      end
+    RUBY
+
+    expect_correction(<<~RUBY)
+      describe 'hello there' do
+        let(:definition) do
+          AbilityDefinition.define do
+            subject(Model, test: ['value']) { can :read }
+          end
+        end
+
+        let(:first) { 1 }
+        subject(:second) { 2 }
+      end
+    RUBY
+  end
+
+  it 'does not register an offense for subjects declared inside an iterator' do
+    expect_no_offenses(<<~RUBY)
+      describe 'hello there' do
+        %i[a b].each do |name|
+          subject(name) { 1 }
+        end
+      end
+    RUBY
+  end
 end
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 - Fix incorrect autocorrection for `RSpec/DescribedClass` when using nested example groups with `EnforcedStyle: explicit`. ([@ydah])
 - Fix `RSpec/MatchWithSimpleRegex` to ignore `match` outside examples. ([@ydah])
 - Fix incorrect autocorrection for `RSpec/ExampleWording` with percent literal and escaped example descriptions. ([@ydah])
+- Fix false positives for cops that detect `subject` when a same-named method is called inside another block, such as a DSL evaluated in a `let`. ([@corsonknowles])
 - Fix `RSpec/SpecFilePathFormat` raising when `EnforcedInflector: active_support` and RuboCop is invoked from outside the project root. ([@corsonknowles])
 - Fix `RSpec/LeadingSubject` to not crash on `itblock`/`numblock` example groups with Ruby 3.4 and Prism. ([@pcbeingused333])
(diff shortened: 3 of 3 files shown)


lib/rubocop/rspec/example_group.rb
```diff
@@ -47,21 +47,30 @@ def hooks
...
 
-      def find_all(node, predicate)
+      def find_all(node, predicate, skip_nested_blocks: false)
+        # A declaration nested in some other block is not one the group makes
+        # unconditionally, and may not be RSpec's method at all, so callers can
+        # ask to stop at those.
         if public_send(predicate, node)
           [node]
-        elsif scope_change?(node) || example?(node)
+        elsif nested_block?(node, skip_nested_blocks) ||
```
pirj: How do you feel about this?
```suggestion
        elsif skip_nested_blocks && nested_block?(node)
          []
        elsif scope_change?(node) || example?(node)
          []
```
not to mix the two, and not to bake the flag to check with the check itself

corsonknowles: Applied, with one adjustment:

the suggestion trips this repo's `Lint/DuplicateBranch`, because the two `elsif` bodies are both `[]`.

Guard clauses get both of your points — the flag checked by the caller, the two checks not mixed — without the duplicate branch:

```ruby
def find_all(node, predicate, skip_nested_blocks: false)
  return [node] if public_send(predicate, node)
  return [] if skip_nested_blocks && nested_block?(node)
  return [] if scope_change?(node) || example?(node)

  find_all_in_scope(node, predicate, skip_nested_blocks: skip_nested_blocks)
end

# A declaration nested in some other block is not one the group makes
# unconditionally, and may not be RSpec's method at all.
def nested_block?(node)
  node.block_type?
end
```

`nested_block?` is now a plain predicate about the node, and the explanatory comment moved onto it, where the reasoning belongs.

On your earlier question — *what if the argument is a method call/var?* — that's exactly why this checks position rather than inspecting the argument. 

Whatever the argument is, a `subject` declared inside another block isn't one the group makes unconditionally, so the cop shouldn't count it. That avoids having to reason about the argument at all.

lib/rubocop/rspec/example_group.rb
```diff
@@ -47,20 +47,25 @@ def hooks
...
 
-      def find_all(node, predicate)
-        if public_send(predicate, node)
-          [node]
-        elsif scope_change?(node) || example?(node)
-          []
-        else
-          find_all_in_scope(node, predicate)
-        end
+      def find_all(node, predicate, skip_nested_blocks: false)
```
pirj, line 56: I’d rather keep the `if` as less dense unless we’re hitting some “method too long” offense.

lib/rubocop/rspec/example_group.rb
```diff
@@ -47,20 +47,25 @@ def hooks
...
+        return [node] if public_send(predicate, node)
+        return [] if skip_nested_blocks && nested_block?(node)
+        return [] if scope_change?(node) || example?(node)
```
pirj: Those two seem to handle unrelated cases. Leaving it up to you if you want to keep this as is, or to separate them - it was together before this PR, so it's unrelated.

lib/rubocop/rspec/example_group.rb
```diff
@@ -47,20 +47,25 @@ def hooks
...
+
+        find_all_in_scope(node, predicate,
+                          skip_nested_blocks: skip_nested_blocks)
+      end
+
+      # A declaration nested in some other block is not one the group makes
+      # unconditionally, and may not be RSpec's method at all.
```
pirj: I’m struggling to understand at a glance how the distinction between an RSpec DSL block and not is detected

lib/rubocop/rspec/example_group.rb
pirj: Sorry, I missed the above Lint/DuplicateBranch note.

lib/rubocop/rspec/example_group.rb
corsonknowles: Right, it *isn't* detected — nothing distinguishes an RSpec DSL block from any other, by design. The old `nested_block?` name implied otherwise, which is probably what made it hard to read.

I've inlined it and stated the intent at the decision point:

```ruby
return [node] if public_send(predicate, node)
# No attempt is made to tell an RSpec DSL block from any other: a
# declaration inside *any* block is not one the group makes
# unconditionally, so stop rather than identify the block.
return [] if skip_nested_blocks && node.block_type?
return [] if scope_change?(node) || example?(node)
```

The reasoning: identifying which blocks are RSpec's means reasoning about the receiver and the argument, which is the rabbit hole your earlier question about method calls/vars pointed at. Position is sufficient — a `subject` inside any block isn't declared unconditionally, so it shouldn't count either way.

lib/rubocop/rspec/example_group.rb
corsonknowles: Separated them:

```ruby
return [] if skip_nested_blocks && node.block_type?
return [] if scope_change?(node)
return [] if example?(node)
```

You are right that they were unrelated — a scope change means a different example group owns what follows, whereas an example means the declaration belongs to a single example rather than the group. Worth one line each.

Worth noting for anyone reading the earlier thread: as guard clauses these do *not* trip `Lint/DuplicateBranch`, which only looks at `if`/`elsif`/`else` bodies. That cop was the only reason I moved away from your original suggestion, so splitting is free here.

`rake internal_investigation` is clean and the full suite is 2469 examples, 0 failures.

spec/rubocop/cop/rspec/multiple_subjects_spec.rb
```diff
@@ -80,4 +80,101 @@
...
+    RUBY
+  end
+
+  it 'ignores a same-named method called inside another block' do
+    expect_no_offenses(<<~RUBY)
+      describe 'hello there' do
+        let(:definition) do
+          AbilityDefinition.define do
+            subject(Model, test: ['value']) { can :read }
+            subject(Model, other: ['value']) { can :write }
```
pirj, line 121:
```suggestion
            subject(:foo) { can :read }
            subject(:bar) { can :write }
```

I know that it's breaking the demonstrational purpose, and the whatever it is the gem DSL is behind `AbilitityDefinition`.
My concern is that if we ever decide to stricten our [`subject?` matcher](https://github.com/rubocop/rubocop-rspec/blob/394a695a09c0f1078c32fd7a7fc209ab44deabec/lib/rubocop/rspec/language.rb#L73) so that it only accepts _one_ optional argument like [the RSpec DSL's `subject` does](https://github.com/rspec/rspec/blob/df877b73ab2a01bc7e3b5a39e7e252e8d9bb67a3/rspec-core/lib/rspec/core/memoized_helpers.rb#L371), this example may start passing for the wrong reason.

spec/rubocop/cop/rspec/multiple_subjects_spec.rb
```diff
@@ -80,4 +80,101 @@
...
+      end
+    RUBY
+  end
+
+  it 'counts the group own subject beside one nested in a block' do
+    expect_no_offenses(<<~RUBY)
+      describe 'hello there' do
+        let(:definition) do
+          AbilityDefinition.define do
+            subject(Model, test: ['value']) { can :read }
```
pirj, line 133:
```suggestion
            subject(:foo) { can :read }
```
same

spec/rubocop/cop/rspec/multiple_subjects_spec.rb
```diff
@@ -80,4 +80,101 @@
...
+      end
+    RUBY
+  end
+
+  it 'autocorrects the group own subjects, not a nested DSL call' do
+    expect_offense(<<~RUBY)
+      describe 'hello there' do
+        let(:definition) do
+          AbilityDefinition.define do
+            subject(Model, test: ['value']) { can :read }
```
pirj, line 147:
```suggestion
            subject(:foo) { can :read }
```

spec/rubocop/cop/rspec/multiple_subjects_spec.rb
```diff
@@ -80,4 +80,101 @@
...
+        ^^^^^^^^^^^^^^^^^^^^^ Do not set more than one subject per example group
+        subject(:second) { 2 }
+      end
+    RUBY
+
+    expect_correction(<<~RUBY)
+      describe 'hello there' do
+        let(:definition) do
+          AbilityDefinition.define do
+            subject(Model, test: ['value']) { can :read }
```
pirj, line 161:
```suggestion
            subject(:foo) { can :read }
```

lib/rubocop/rspec/example_group.rb
```diff
@@ -22,7 +22,7 @@ def lets
       end
 
       def subjects
-        find_all_in_scope(node, :subject?)
+        find_all_in_scope(node, :subject?, skip_nested_blocks: true)
```
pirj, line 25: This symptomatic fix introduces a new class of problems - there's now a distinction between how we treat `let`s,  `subject`s and all other RSpec DSLs in those helpers.

This comes from some third-party gem's DSL that has an overlapping method. But collaterally, it affects that now subjects can be defined with `each`.

I beg my pardon for another steering request, but I'd prefer this change to:
1) affect all `find_all_in_scope` equally
2) only filter out obvious non-generic and non-RSpec DSL

That way, we don't introduce toleration to `each`, and we keep parity between detection of `let` and `subject`.

lib/rubocop/rspec/example_group.rb
```diff
@@ -47,20 +47,26 @@ def hooks
...
-          []
-        else
-          find_all_in_scope(node, predicate)
-        end
+      def find_all(node, predicate, skip_nested_blocks: false)
+        return [node] if public_send(predicate, node)
+        # No attempt is made to tell an RSpec DSL block from any other: a
+        # declaration inside *any* block is not one the group makes
+        # unconditionally, so stop rather than identify the block.
+        return [] if skip_nested_blocks && node.block_type?
```
pirj, line 61: How about we unconditionally skip `AbilityDefinition.define`-like alien blocks?

lib/rubocop/rspec/example_group.rb
```diff
@@ -47,20 +47,26 @@ def hooks
...
+        return [] if scope_change?(node)
+        return [] if example?(node)
```
pirj, line 63: Or, alternatively, add a check that would skip all incompatible nesting, like `subject` inside `let` etc.

Memoized helpers, hooks, example groups and examples can be nested under an example group. Nothing can be nested under example, hook, memoized helper. Is that right?

It deserves a dedicated commit with an explanation.

And that automatically solves the
```ruby
let(:foo) do
  AD.define do
    subject(:bar) { ... }
  end
end
```
case. No special treatment needed.

PR #2226, opened 2026-08-24
Fix `RSpec/SpecFilePathFormat` raising when run from outside the project root

corsonknowles:
## Problem

`InflectorPath` is a path parameter, but it is the only one RuboCop does not resolve against the config's base directory. Both the existence check and the `require` see the raw configured value:

```ruby
inflector_path = config.fetch('InflectorPath')

unless File.exist?(inflector_path)
  raise "The configured `InflectorPath` #{inflector_path} does not exist."
end
```

So the default `./config/initializers/inflections.rb` is resolved against the **process working directory**. With `EnforcedInflector: active_support`, that means the cop raises whenever RuboCop is invoked from anywhere but the project root — an editor integration that runs it from the file's own folder, or a Git worktree. It surfaces as a RuboCop error with no offense reported, which reads like a bug in the cop rather than a path that could not be found.

## Fix

Resolve it with `config.base_dir_for_path_parameters` — the same anchor `Include` and `Exclude` already use, so `InflectorPath` now behaves like every other path parameter. Relative paths are anchored to the directory of the `.rubocop.yml` that set them; absolute paths are unaffected, since `File.expand_path` is a no-op for them.

`prepare_availability` now takes the resolved path rather than the whole `cop_config`, which also narrows its interface to the one value it actually used.

## Tests

A new example pins the behaviour: with the config's base directory pointing somewhere other than the working directory, the cop checks `<config dir>/config/initializers/inflections.rb`. It fails on `master` and passes with this change.

The existing `InflectorPath` examples stubbed `File.exist?` with the raw relative value, so they are updated to the resolved path.

## Verification

- `bundle exec rspec spec/rubocop/cop/rspec/spec_file_path_format_spec.rb` — 60 examples, 0 failures
- `bundle exec rubocop` — clean on both changed files
- Full suite: 2466 examples, 2 failures — both in `context_wording_spec.rb`, and both reproduce on an unmodified `master` in my environment (they involve Japanese `とき` patterns and look locale-related), so they are unrelated to this change.

### DIFF
--- a/lib/rubocop/cop/rspec/spec_file_path_format.rb
+++ b/lib/rubocop/cop/rspec/spec_file_path_format.rb
@@ -74,12 +74,10 @@ module RuboCop
             ActiveSupport::Inflector.underscore(string)
           end
 
-          def self.prepare_availability(config)
+          def self.prepare_availability(inflector_path)
             return if @prepared
 
             @prepared = true
-
-            inflector_path = config.fetch('InflectorPath')
 
             unless File.exist?(inflector_path)
               raise "The configured `InflectorPath` #{inflector_path} does " \
@@ -104,11 +102,18 @@ module RuboCop
         def inflector
           case cop_config.fetch('EnforcedInflector')
           when 'active_support'
-            ActiveSupportInflector.prepare_availability(cop_config)
+            ActiveSupportInflector.prepare_availability(inflector_path)
             ActiveSupportInflector
           when 'default'
             DefaultInflector
           end
+        end
+
+        def inflector_path
+          File.expand_path(
+            cop_config.fetch('InflectorPath'),
+            config.base_dir_for_path_parameters
+          )
         end
 
         def ensure_correct_file_path(send_node, class_name, arguments)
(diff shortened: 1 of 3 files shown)


lib/rubocop/cop/rspec/spec_file_path_format.rb
```diff
@@ -104,13 +102,25 @@ def self.call(string)
...
           when 'default'
             DefaultInflector
           end
         end
 
+        # `InflectorPath` is a path parameter, so resolve it the way RuboCop
+        # resolves every other one: against the directory of the config that
+        # set it, rather than the process working directory. Otherwise the
+        # cop raises whenever RuboCop is invoked from anywhere but the
+        # project root, such as an editor integration or a Git worktree.
```
bquorning: I think this comment is mostly useful during code review.
```suggestion
```

bquorning (approved):

lib/rubocop/cop/rspec/spec_file_path_format.rb
corsonknowles: Agreed

bquorning (approved):

PR #2229, opened 2026-09-09
Fix `RSpec/IteratedExpectation` crash on a receiverless `each`

pcbeingused333:
## Summary

`RSpec/IteratedExpectation` raises `NoMethodError: undefined method 'source' for nil` during autocorrection when the iterated `each` is called without an explicit receiver:

```ruby
it 'validates users' do
  each { |user| expect(user).to be_valid }
end
```

`single_expectation_replacement` does `node.receiver.source` to build `expect(<collection>).to all(...)`. With no receiver, `node.receiver` is `nil` and `rubocop -a` crashes on this file.

## Fix

Skip the correction when there is no receiver. Without a collection there is nothing to move into `expect(...)`, so the offense is still registered but left uncorrected — the same behaviour the cop already has for the other non-correctable shapes (error message argument, block argument used in the matcher).

## Details

- One guard line in `check_offense`'s corrector block: `next if node.receiver.nil?`
- Regression spec added; fails on `master` with the `NoMethodError`, passes with the fix
- `bundle exec rake` green: 2471 examples / 0 failures, `internal_investigation` clean (293 files), `confirm_config` / `confirm_documentation` / `documentation_syntax_check` regenerate nothing

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01GtYgwYNfJ3wHrw9xrooVoB


lib/rubocop/cop/rspec/iterated_expectation.rb
```diff
@@ -74,6 +74,7 @@ def check_offense(node, argument)
           if single_expectation?(node.body, argument)
             add_offense(node.send_node) do |corrector|
               next unless node.body.arguments.one?
+              next if node.receiver.nil?
```
pirj: Nitpick: why not just next unless node.receiver?
Isn’t there a cop for this? Why it’s not offended?

pirj (approved):
No objections. Still, the receiverless each feels wrong . I’d rather used explicit. Tell us more about your case

PR #2231, opened 2026-09-09
Fix RSpec/ContainExactly to not crash on a nested `contain_exactly` splat

pcbeingused333:
## Summary

`RSpec/ContainExactly` raises `Parser::ClobberingError` when the splatted argument is itself a `contain_exactly` call:

```ruby
is_expected.to contain_exactly(*contain_exactly(*array))
```

## Problem

Both calls match `node.arguments.one? && node.first_argument.splat_type?`, so both get an offense. The inner one replaces itself with `match_array(array)`; the outer one replaces its whole node (which includes the inner call's source) with `match_array(<inner source>)`. The outer's replacement range fully contains the inner one — combining both corrections in the same pass clobbers.

## Fix

Skip the correction when the splatted expression is itself a `contain_exactly` call — it is already offended and will be replaced by its own correction. The offense is still reported. On the next autocorrect pass, with the inner correction already applied, the splatted expression is a `match_array` call rather than `contain_exactly`, so the guard no longer applies and the outer correction goes through — both end up fully corrected across two passes.

## Tests

- New `does not crash when the splat arg is itself a \`contain_exactly\` call`.
- Verified in red: reverting the fix reproduces the crash (`Parser::ClobberingError`), confirmed both via the spec and via `rubocop -a` on a throwaway file (this cop's autocorrection isn't marked unsafe, so it reproduces without `-A`).
- `bundle exec rspec`: 2471 examples, 0 failures, 100% line/branch coverage.
- `bundle exec rake internal_investigation`: 293 files, no offenses.
- `bundle exec rake confirm_config confirm_documentation documentation_syntax_check`: clean, nothing regenerated.
- `bundle exec rspec spec/project`: 20 examples, 0 failures (changelog format).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01PjzCQWswbH69CXJ2nCa9Zk


spec/rubocop/cop/rspec/contain_exactly_spec.rb
```diff
@@ -56,4 +56,16 @@
       it { is_expected.to contain_exactly() }
     RUBY
   end
+
+  it 'does not crash when the splat arg is itself a `contain_exactly` call' do
+    expect_offense(<<~RUBY)
+      it { is_expected.to contain_exactly(*contain_exactly(*array)) }
```
bquorning: I’m having a hard time wrapping my head around the use case for this “nested” spec. Could you please expand it with an explicit subject and set a value for the `array` lvar?

PR #2235, opened 2026-09-14
Fix an incorrect autocorrect for `RSpec/SharedExamples` cop

viralpraxis:
An MRE:

```bash
$ bundle exec rubocop --debug --stdin /dev/stdin -a --config <(cat <<'YAML'
AllCops:
  DisabledByDefault: true
  SuggestExtensions: false
plugins:
  - rubocop-rspec
RSpec:
  Include:
    - '**/*'
RSpec/SharedExamples:
  Enabled: true
  EnforcedStyle: symbol
YAML
) <<'RUBY'
RSpec.shared_examples "a `b`" do
end
RUBY
configuration from /proc/self/fd/11
Inspecting 1 file
Scanning /dev/stdin
F

Offenses:

/dev/stdin:1:23: C: [Corrected] RSpec/SharedExamples: Prefer :a_b over "a b" to symbolize shared examples.
RSpec.shared_examples "a `b`" do
                      ^^^^^^^
/dev/stdin:1:26: F: Lint/Syntax: unexpected token tXSTRING_BEG
RSpec.shared_examples :a_`b` do
                         ^
```

There are two ways to handle strings which are **not** valid symbols after replacing space with `_`:
1. do not register any offenses at all
2. correct e.g. `"a-b"` to `:"a-b"`

This patch chooses the second option.

- [X] Added tests.
- [X] Updated documentation.


pirj (approved):
Thank you! 🤩

bquorning (approved):

